PR #3, opened 2019-06-25
Fix bug #65269: SplFileObject::__construct() throws LogicException as well

BenMorel:
Fix for [bug 65269](https://bugs.php.net/bug.php?id=65269).


nikic:
@php-pulls merge

PR #5, opened 2019-07-13
Fix typos

carusogabriel:
Found with https://github.com/vlajos/misspell-fixer

### DIFF
--- a/reference/ev/watcher-callbacks.xml
+++ b/reference/ev/watcher-callbacks.xml
@@ -12,7 +12,7 @@
   a bitmask of received events.
  </para>
  <para>
-  Watcher callbacks are passed to the watcher contructors(the classes derived
+  Watcher callbacks are passed to the watcher constructors (the classes derived
   from
   <classname>EvWatcher</classname>
   -
--- a/reference/mssql/ini.xml
+++ b/reference/mssql/ini.xml
@@ -47,7 +47,7 @@
      <entry></entry>
     </row>
     <row>
-     <entry>mssql.compatability_mode</entry>
+     <entry>mssql.compatibility_mode</entry>
      <entry>"0"</entry>
      <entry>PHP_INI_ALL</entry>
      <entry></entry>
--- a/reference/parle/parle.parser.xml
+++ b/reference/parle/parle.parser.xml
@@ -77,7 +77,7 @@
     <fieldsynopsis>
      <modifier>const</modifier>
      <type>integer</type>
-     <varname linkend="parle-parser.constants.error-unkown-token">Parle\Parser::ERROR_UNKOWN_TOKEN</varname>
+     <varname linkend="parle-parser.constants.error-unknown-token">Parle\Parser::ERROR_UNKNOWN_TOKEN</varname>
      <initializer>2</initializer>
     </fieldsynopsis>
     
@@ -156,8 +156,8 @@
      </listitem>
     </varlistentry>
 
-    <varlistentry xml:id="parle-parser.constants.error-unkown-token">
-     <term><constant>Parle\Parser::ERROR_UNKOWN_TOKEN</constant></term>
+    <varlistentry xml:id="parle-parser.constants.error-unknown-token">
+     <term><constant>Parle\Parser::ERROR_UNKNOWN_TOKEN</constant></term>
      <listitem>
       <para></para>
      </listitem>
--- a/reference/yaf/yaf-route-map.xml
+++ b/reference/yaf/yaf-route-map.xml
@@ -57,7 +57,7 @@
     </fieldsynopsis>
     <fieldsynopsis>
      <modifier>protected</modifier>
-     <varname linkend="yaf-route-map.props.delimeter">_delimeter</varname>
+     <varname linkend="yaf-route-map.props.delimiter">_delimiter</varname>
     </fieldsynopsis>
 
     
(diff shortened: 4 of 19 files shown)


reference/ev/watcher-callbacks.xml
```diff
@@ -12,7 +12,7 @@
   a bitmask of received events.
  </para>
  <para>
-  Watcher callbacks are passed to the watcher contructors(the classes derived
+  Watcher callbacks are passed to the watcher constructors(the classes derived
```
nikic:
```suggestion
  Watcher callbacks are passed to the watcher constructors (the classes derived
```

reference/mssql/ini.xml
```diff
@@ -47,7 +47,7 @@
      <entry></entry>
     </row>
     <row>
-     <entry>mssql.compatability_mode</entry>
+     <entry>mssql.compatibility_mode</entry>
```
nikic: To make sure, did you check that there is no typo in the ini name of the extension?

reference/mysqlnd_qc/quickstart.xml
```diff
@@ -1367,7 +1367,7 @@ array(4) {
   array(9) {
     ["query"]=>
     string(25) "DROP TABLE IF EXISTS test"
-    ["occurences"]=>
+    ["occurrences"]=>
```
nikic: I'd assume that this is a typo in the implementation -- if we want to change the output here, the implementation needs to change first.

reference/parle/parle.parser.xml
```diff
@@ -77,7 +77,7 @@
     <fieldsynopsis>
      <modifier>const</modifier>
      <type>integer</type>
-     <varname linkend="parle-parser.constants.error-unkown-token">Parle\Parser::ERROR_UNKOWN_TOKEN</varname>
+     <varname linkend="parle-parser.constants.error-unknown-token">Parle\Parser::ERROR_UNKNOWN_TOKEN</varname>
```
nikic: Once again, what is the actual name of the constant in the extension? Does it have the typo or not?

reference/sybase/ini.xml
```diff
@@ -53,7 +53,7 @@
      <entry></entry>
     </row>
     <row>
-     <entry>sybase.compatability_mode</entry>
+     <entry>sybase.compatibility_mode</entry>
```
nikic: Same question here.

reference/yaf/yaf-route-map.xml
```diff
@@ -57,7 +57,7 @@
     </fieldsynopsis>
     <fieldsynopsis>
      <modifier>protected</modifier>
-     <varname linkend="yaf-route-map.props.delimeter">_delimeter</varname>
+     <varname linkend="yaf-route-map.props.delimiter">_delimiter</varname>
```
nikic: Same here.

reference/mssql/ini.xml
cmb69: Both names are supported: https://github.com/php/php-src/blob/05fb10165081cb4fb55a9943b44cab62c627db7a/ext/mssql/php_mssql.c#L250-L251

reference/mysqlnd_qc/quickstart.xml
cmb69: Indeed a typo in the implementation: http://svn.php.net/viewvc/pecl/mysqlnd_qc/trunk/mysqlnd_qc_logs.c?revision=331414&view=markup#l125

reference/parle/parle.parser.xml
cmb69: See https://github.com/weltling/parle/issues/14

reference/sybase/ini.xml
cmb69: Implementation typo: https://github.com/php/php-src/blob/PHP-5.2/ext/sybase/php_sybase_db.c#L271

reference/yaf/yaf-route-map.xml
cmb69: See https://github.com/laruence/yaf/commit/e366da2b99df147d66b790bdea7c982cef74875b#diff-ee82af77d5310c2a2fa9060fdcc2e16f.

reference/mysqlnd_qc/quickstart.xml
carusogabriel: Should we change for `PHP-7.4+` or is a bug fix, and should enter `PHP-7.2+`?

reference/sybase/ini.xml
carusogabriel: Should we change for `PHP-7.4+` or is a bug fix, and should enter `PHP-7.2+`?

reference/mssql/ini.xml
nikic: Okay, then I think it's fine to only mention the fixed name.

reference/sybase/ini.xml
nikic: This extension does not exist in PHP 7, so this change should just be reverted.

reference/mysqlnd_qc/quickstart.xml
nikic: The last commit to this extension is 5 years old, so I don't think it makes sense to change this.

nikic:
@php-pulls merge

PR #6, opened 2019-07-13
Fix $revisions number

carusogabriel:


nikic:
Do these files have the necessary svn:keywords property?

carusogabriel:
@nikic Sorry, but what does `svn:keywords` means?

I've made this PR based on the fact that these files were the only ones that have a revision number, following https://github.com/php/doc-en/commit/564b0eb4c9b26c4b5bcab9cfb27368094aa4b130#diff-649f9459e06d2296dcac4935ebc209eeR2

nikic:
@carusogabriel It took me a while to find a reference for this... see the very end of http://doc.php.net/tutorial/faq.php. These SVN properties are not set for the parallel docs. Once they are set, the `$Revision$` keyword should work correctly.

carusogabriel:
Done via ed0f1b9795919290fc53faa727d9b5fcced93b25.

PR #8, opened 2019-09-21
Little improvement

lex111:
cc @salathe


php-pulls:
**Comment on behalf of salathe at php.net:**

Applied via fed0bd59. Thanks!

PR #13, opened 2019-09-22
Migration to PHP 7.4: use special markup for configure option

lex111:
cc @salathe


php-pulls:
**Comment on behalf of salathe at php.net:**

Applied as 3ad6446a. Thanks!

PR #17, opened 2019-11-14
Add missing MB_CASE_UPPER_SIMPLE constant

BenMorel:
This constant appears in the [7.3 migration page](https://www.php.net/manual/en/migration73.constants.php#migration73.constants.mbstring), but for some reason was forgotten here.


nikic:
@php-pulls merge

PR #18, opened 2019-11-25
Fix mysqli_get_client_info() parameter list

kocsismate:
Based on the discussion at https://github.com/php/php-src/pull/4878#pullrequestreview-309797488, the signature of `mysqli_get_client_info()` should be fixed in the documentation.

### DIFF
--- a/reference/mysqli/mysqli/get-client-info.xml
+++ b/reference/mysqli/mysqli/get-client-info.xml
@@ -19,7 +19,7 @@
   <para>&style.procedural;</para>
   <methodsynopsis>
    <type>string</type><methodname>mysqli_get_client_info</methodname>
-   <methodparam><type>mysqli</type><parameter>link</parameter></methodparam>
+   <methodparam><type>mysqli</type><parameter>link</parameter><initializer>NULL</initializer></methodparam>
   </methodsynopsis>
   <para>
    Returns a string that represents the MySQL client library version.


nikic:
cc @cmb69 

Not clear from the discussion whether this is quite the right change -- is the object oriented variant a static method?

cmb69:
No, the method is not static, but it lacks any ZPP. Same with `mysqli::thread_safe()`. However, the former is documented to expect a parameter, while the latter is document to expect void.

kocsismate:
Yes, my reasoning behind this change was to correct the error in the documentation first, and then we can decide later if we still want to disallow any arguments, or want to be more lenient with the ZPP, and if we want to make their method counterparts static.

cmb69:
Well, given that the *function* accepts arbitrary arguments, and that the related *method* is not documented to be static, it might make sense to document that it accepts a single *optional* argument.

kocsismate:
@cmb69 Makes sense, I updated the PR.

kocsismate:
@cmb69 If you have a little bit of available time, could you please have a look at if this PR is good to be merged now?

kocsismate:
@php-pulls merge

nikic:
@php-pulls merge

Should work for you as well now, but keep in mind that everything will still get committed under my name, as SVN has no author/committer distinction.

kocsismate:
Thanks! It's fine for me.

PR #19, opened 2019-11-27
Add array destructuring example to language.types.array

rask:
I'm not sure if PRs are accepted here, but here goes:

Simple example of the PHP 7.1 array destructuring feature.

There seems to be no official manual page/section for this, so adding it here with the rest of the array syntax-y things.


nikic:
@salathe Is that a reasonable place to put this?

salathe:
> @salathe Is that a reasonable place to put this?

I would prefer a new page on the topic of destructuring (though I'm not sure the best place for that), which can then be referenced from the array-related pages and other pages like `list()`.

However, in the interest of just having this written down somewhere (other than the migration guide, or user contributed notes) being better than nowhere, including the proposed section in the array page would be a good starting point.

rask:
Agreed, the topic is large enough to warrant a separate page in my opinion as well. This PR is mainly to just have it visible somewhere as stated. My inspiration was a PHP veteran using latest PHP saying "wtf since when has that been possible" when I showed them an example, and they said no amount of search returned results in the manual for it.

Is the position on the page OK or would it work better somewhere else?

salathe:
@rask this PR has gotten all out of shape due to the forced pushes on the php/doc-en master branch.  Is this still something you want to pursue, or should we just lay this PR to rest?

rask:
Hello. Sorry, the new GitHub notification center is seemingly broken and I can't see any notifications anywhere. Or maybe I just don't understand how it works haha. I stumbled on this on the GitHub frontpage, which I see maybe once a year.

I'll see if I can remold this to fit the new base.

rask:
You know what, the base has changed so much a simple master to master rebase is giving me tons and tons of merge conflcts for no good reason. I shall refork and redo the commit manually and create a new PR.

rask:
New one at #145.

PR #20, opened 2020-01-17
Fix: Remove $serverctrls parameter from ldap_bind()

localheinz:
This PR

* [x] removes the non-existent `$serverctrls` parameter from the documentation for `ldap_bind()`

💁‍♂ When looking at (for example)

* https://github.com/php/php-src/blob/php-7.3.11/ext/ldap/ldap.c#L1142-L1194
* https://github.com/php/php-src/blob/php-7.4.1/ext/ldap/ldap.c#L1117-L1169

it's not apparent that `ldap_bind()` accepts 4 arguments. It does not have a `$serverctrls` parameter.


OskarStark:
Maybe it was mixed up with `ldap_bind_ext()`

nikic:
@php-pulls merge

localheinz:
Hmm, awesome!

localheinz:
Thank you, @nikic, @OskarStark, and @salathe!

OskarStark:
Awesome 😎🥳

PR #23, opened 2020-01-23
Fix #79133 - Replace <literal> with <code>

SjonHortensius:
For output reasons - a single literal pipe will be printed italic - looking like a `/`

### DIFF
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -249,7 +249,7 @@ class MyException extends Exception
    </para>
    <para>
     In PHP 7.1 and later, a &catch; block may specify multiple exceptions
-    using the pipe (<literal>|</literal>) character. This is useful for when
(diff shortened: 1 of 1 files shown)


salathe:
IMO this should be fixed at the [PhD](https://github.com/php/phd) (to use something other than plain `<em>` HTML elements for the `<literal>` DocBook elements, maybe add a class to it?) and/or [web-php](https://github.com/php/web-php) (to apply a better CSS style to the elements) layers.

I don't think it is a good idea to change the DocBook markup for this one character (and only in this one place, of several where it occurs).

SjonHortensius:
Thanks, while I do agree there are a lot more possible issues that may surface when making that change, while the referenced problem is small and easily fixed.

I don't mind fixing this globally - I'll look into that and reference this issue with your feedback where applicable

PR #24, opened 2020-01-23
Clarify pseudo C preprocessor support

cmb69:
Based on a patch provided by Larry Garfield.

### DIFF
--- a/reference/ffi/ffi/load.xml
+++ b/reference/ffi/ffi/load.xml
@@ -36,6 +36,7 @@
      <para>
       The header file <emphasis>should</emphasis> contain a <literal>#define</literal> statement for the
       <literal>FFI_SCOPE</literal> variable, e.g.: <code>#define FFI_SCOPE "MYLIB"</code>.
+      Refer to the <link linkend="ffi.intro">class introduction</link> for details.
      </para>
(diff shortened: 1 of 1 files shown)


Crell:
I wasn't quite sure how to file this information, frankly.  But I spent >2 days figuring out the path needs, so it seemed prudent to document it so that others don't have to waste 2 days on it.  I'm open to suggestions on how better to document these facts.

Crell:
How do we move this forward?

cmb69:
Thanks for the ping! In my opinion, the PR is okay (maybe wording could be improved; suggestions welcome!)

nikic:
Possibly we can refer to https://www.php.net/manual/en/class.ffi.php, which explains what the scope stuff is for? The added note here doesn't really make it clear what it does, apart from the fact that you can use it.

cmb69:
I fixed the broken PR, and added a commit with a link to the class intro.

Crell:
Thanks, @cmb69!  I think this is ready for whoever does the magic from here.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349215.

PR #25, opened 2020-01-24
Fixed an incorrect return value

mmcev106:


nikic:
@php-pulls merge

Checked the code, indeed `false`. Thanks!

PR #27, opened 2020-01-30
Document missing curl constants added in 7.3.x

ekinhbayar:
FYI, I was looking at the following blame and curl's symbol versions reference for these changes:
[blame](https://github.com/php/php-src/blame/494615fcb8c1fb5984e0e7d666e51a2dfc6bee55/UPGRADING#L570)
[symbols in versions](https://github.com/curl/curl/blob/master/docs/libcurl/symbols-in-versions)

There are two commits of interest from the blame:
[Add constants from curl 7.50 to 7.55](https://github.com/php/php-src/commit/02b2dbb7242cba4324ceebe27fe8ad1e5eedc818) - merged into 7.3.0RC3
[Add constants from curl 7.56 to 7.61](https://github.com/php/php-src/commit/64881a1e64b47cb612d358f2ebf234defe314e90) - merged into 7.3.0RC3

### DIFF
--- a/reference/curl/constants.xml
+++ b/reference/curl/constants.xml
@@ -1894,6 +1894,39 @@
     </simpara>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curl-version-asynchdns">
+   <term>
+    <constant>CURL_VERSION_ASYNCHDNS</constant>
+    (<type>integer</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Available since PHP 7.3.0 and cURL 7.10.7
+    </simpara>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curl-version-brotli">
+   <term>
+    <constant>CURL_VERSION_BROTLI</constant>
+    (<type>integer</type>)
+   </term>
+   <listitem>
+    <simpara>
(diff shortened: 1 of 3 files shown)


cmb69:
Thanks for the PR! The `CURLINFO_*` constants should also listed on the [`curl_getinfo` page](https://www.php.net/curl-getinfo), and the `CURLOPT_*` constants on the [`curl_setopt` page](https://www.php.net/curl-setopt).

ekinhbayar:
> Thanks for the PR! The `CURLINFO_*` constants should also listed on the [`curl_getinfo` page](https://www.php.net/curl-getinfo), and the `CURLOPT_*` constants on the [`curl_setopt` page](https://www.php.net/curl-setopt).

Cool, will look at that page as well. @cmb69, should these two files be included in this PR or separately?

cmb69:
If you like to do it, it would be nice to have all relevant changes for these constants bundled in one PR.

ekinhbayar:
@cmb69, I think I've got them all listed in both pages as well now.

ekinhbayar:
@PeeHaa thanks! I think I've covered all the requested changes now.

nikic:
@php-pulls merge

nikic:
Closing manually, as this has landed in SVN, but won't be picked up in the Git repo right away due to the pending mirroring switchover.

PR #30, opened 2020-02-03
Remove whitespace

OskarStark:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349109.

PR #31, opened 2020-02-03
Fix line length

OskarStark:


cmb69:
I would leave that as is, because otherwise that file would be flagged as being out-dated in all translations, and each translation would have to update the `EN-Revision`.

OskarStark:
Ok didn’t know about that

PR #32, opened 2020-02-03
Remove whitespace

OskarStark:


nikic:
As @cmb69 mentioned on the other PR, invisible whitespace changes like this one are generally not worthwhile, because they need to be carried over into all translations afterwards. It's something that can be fixed when the file is touched anyway, but causes unnecessary churn as an independent change.

salathe:
> It's something that can be fixed when the file is touched anyway,

Please do *not* mix whitespace-only commits with content changes. 

Whitespace, formatting, etc. changes should be separate commit messages with clear commit messages (e.g. "whitespace", "WS" or similar) to make it clear to translators that the change is not adding or removing content.

salathe:
As for this particular PR, double spaces between sentence is acceptable.

OskarStark:
Thank you all for the feedback

PR #33, opened 2020-02-05
Add missed `deprecated` attributes in versions.xml of info/ and image/

berezuev:
Added `deprecated` tags to some functions, because they should be placed in a '**Deprecated**' section of menu.
image

### DIFF
--- a/reference/image/versions.xml
+++ b/reference/image/versions.xml
@@ -90,14 +90,14 @@
  <function name="imagepalettetotruecolor" from="PHP 5 &gt;= 5.5.0, PHP 7"/>
  <function name="imagepng" from="PHP 4, PHP 5, PHP 7"/>
  <function name="imagepolygon" from="PHP 4, PHP 5, PHP 7"/>
- <function name="imagepsbbox" from="PHP 4, PHP 5"/>
- <function name="imagepscopyfont" from="PHP 5"/>
- <function name="imagepsencodefont" from="PHP 4, PHP 5"/>
- <function name="imagepsextendfont" from="PHP 4, PHP 5"/>
- <function name="imagepsfreefont" from="PHP 4, PHP 5"/>
- <function name="imagepsloadfont" from="PHP 4, PHP 5"/>
- <function name="imagepsslantfont" from="PHP 4, PHP 5"/>
- <function name="imagepstext" from="PHP 4, PHP 5"/>
+ <function name="imagepsbbox" from="PHP 4, PHP 5" deprecated="PHP 7"/>
+ <function name="imagepscopyfont" from="PHP 5" deprecated="PHP 7"/>
+ <function name="imagepsencodefont" from="PHP 4, PHP 5" deprecated="PHP 7"/>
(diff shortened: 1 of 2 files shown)


cmb69:
To my knowledge, the `imageps*` functions have never been formally deprecated; the are just removed as of PHP 7.0.0. Not sure if those functions should be marked as deprecated in the manual.

berezuev:
@cmb69 AFAIK, other functions from current PR have never been deprecated. But, I think, they shouldn't be in a main list of functions. Is there other way to separate them?

nikic:
Making deprecated version = removed version seems like a reasonable convention for functions that were not deprecated prior to removal.

berezuev:
@nikic I've seen this in docs:
```xml
<function name="HTTP_RAW_POST_DATA" from="PHP 4, PHP 5" deprecated="PHP 5.6.0" removed="PHP 7"/>
```
It's the only place, where `removed` have been used. But, how about using this in my case?
```xml
<function name="imagepsbbox" from="PHP 4, PHP 5" deprecated="PHP 7.0.0" removed="PHP 7.0.0"/>
```

cmb69:
Maybe we should change the warning for the `imageps*` function then, according to [php_logo_guid](https://www.php.net/php_logo_guid) ("deprecated and removed" instead of "removed" only).

nikic:
@berezuev I think that usage of `removed` is incorrect. At least I can't find any handling related to it from a quick grep of PhD. For reference, `deprecated` is handled in https://github.com/php/phd/blob/50576cfd1ebb2f9a22bae5c2c8e3ac22cfcecbf8/phpdotnet/phd/Package/PHP/XHTML.php#L211-L246.

nikic:
Probably the "Deprecated" header on php.net should be renamed to "Deprecated or Removed"... Don't know where that is in the code though.

cmb69:
I thinks it's [here](https://github.com/php/web-php/blob/b789e70559b0978c19d52fdc98832607d8b11d16/include/footer.inc#L40). However, the manual is supposed to document PHP 5 as well, so marking these functions as removed in the TOC might be confusing.

nikic:
@cmb69 Which is why the "or". Whether it's deprecated or removed depends on the version, but it can be either one of those.

berezuev:
@nikic @cmb69 updated image/versions.xml.

I can propose pull request to [this file](https://github.com/php/web-php/blob/master/include/footer.inc#L40) after finishing this PR.

cmb69:
@berezuev, it seems that php-pulls messed up this PR; could you please fix this up? Also I think we should at first make the change to web-php, and only afterwards make respective changes to versions.xml.

nikic:
I've fixed up the PR after the mirroring change broke it, please run `git fetch origin && git reset --hard origin/master` before pushing new commits.

cmb69:
As time went by, the manual no longer documents PHP 5, and these functions have already been removed from the manual altogether, so this PR is obsolete. Thanks nonetheless for having worked on this!

PR #34, opened 2020-02-05
Remove hint to security purpose of disable_functions

jsmmo:
Because this feature shouldnt be used for security purposes (its zero percent effective) this hint is misleading.

### DIFF
--- a/appendices/ini.core.xml
+++ b/appendices/ini.core.xml
@@ -353,8 +353,7 @@
       </term>
       <listitem>
        <para>
-        This directive allows you to disable certain functions for 
-        <link linkend="security">security</link> reasons. It takes 
+        This directive allows you to disable certain functions. It takes 
         on a comma-delimited list of function names. disable_functions
         is not affected by <link linkend="ini.safe-mode">Safe Mode</link>.
        </para>
--- a/appendices/ini.core.xml
+++ b/appendices/ini.core.xml
@@ -354,8 +354,7 @@
       <listitem>
        <para>
         This directive allows you to disable certain functions. It takes 
-        on a comma-delimited list of function names. disable_functions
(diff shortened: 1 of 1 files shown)


cmb69:
*If* this will be changed here, it should be changed in [php.ini-development](https://github.com/php/php-src/blob/b3f17ea5d55f5c541f07bf28fda95d54bbf958e0/php.ini-development#L309) and php.ini-production as well.

nikic:
This change looks fine to me (with the corresponding ini changes, yes).

jsmmo:
Please ping me if you want me to prepare that PR too.

nikic:
As a side note, we should probably not be referencing "safe mode" here anymore, which has been gone for a very long time.

appendices/ini.core.xml
```diff
@@ -353,8 +353,7 @@
       </term>
       <listitem>
        <para>
-        This directive allows you to disable certain functions for 
-        <link linkend="security">security</link> reasons. It takes 
+        This directive allows you to disable certain functions. It takes
```
nikic: The same wording change should also be applied to "disable_classes" below.

jsmmo:
Related php-src PR: https://github.com/php/php-src/pull/5150

hopeseekr:
Why is it "zero percent effective"? I thought this was -the- way to prevent shenanigans like `popen` before we shifted to docker?

fiskie:
> Why is it "zero percent effective"? I thought this was -the- way to prevent shenanigans like `popen` before we shifted to docker?

If the qualm is security, and an attacker is far enough in that they can execute arbitrary PHP code on a server, then it is too late. There are far more ways to harm a service than doing system calls.

nikic:
@php-pulls merge

PR #36, opened 2020-02-06
Remove the incorrect pattern "..." in glob()

colshrapnel:
According to the linked answer on Stack Overflow, and to my actual experience, the ... special character does nothing, returning anything but a recursive list of sub-directories. Looks like it should be removed.
https://stackoverflow.com/questions/58852914/glob-pattern-does-not-match-anything

### DIFF
--- a/reference/filesystem/functions/glob.xml
+++ b/reference/filesystem/functions/glob.xml
@@ -53,11 +53,6 @@
         </listitem>
         <listitem>
          <simpara>
-          <literal>...</literal> - Matches all the subdirectories, recursively.
-         </simpara>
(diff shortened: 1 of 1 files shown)


cmb69:
cc @Girgias

nikic:
There might be a `**` pattern though, dunno if it's supported.

Girgias:
It is not as the C glob() function doesn't support it

nikic:
@php-pulls merge

nikic:
Landed in SVN, closing manually as it won't show up here until the new mirror is running.

PR #37, opened 2020-02-07
Add deprecation of money_format

BackEndTea:
I hope this is correct, i copy pasted this from other functions

### DIFF
--- a/reference/strings/functions/money-format.xml
+++ b/reference/strings/functions/money-format.xml
@@ -249,8 +249,7 @@
        <entry>7.4.0</entry>
        <entry>
         This function has been deprecated. Instead, 
-        use <function>numberformatter.formatcurrency</function>.
-       </entry>
+        use <methodname>NumberFormatter::formatCurrency</methodname>.
        </entry>
       </row>
      </tbody>
--- a/reference/strings/functions/money-format.xml
+++ b/reference/strings/functions/money-format.xml
@@ -6,6 +6,10 @@
   <refpurpose>Formats a number as a currency string</refpurpose>
  </refnamediv>
  
+ <refsynopsisdiv>
+   &warn.deprecated.function-7-4-0;
+ </refsynopsisdiv>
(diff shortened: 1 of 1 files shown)


reference/strings/functions/money-format.xml
```diff
@@ -229,6 +233,30 @@
...
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>7.4.0</entry>
+       <entry>
+        This function has been deprecated.
```
nikic: I think we should give some indication as to what you should use instead -- in this case that would be the number formatting functionality in intl.

reference/strings/functions/money-format.xml
```diff
@@ -229,6 +233,32 @@
...
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>7.4.0</entry>
+       <entry>
+        This function has been deprecated. Instead, 
+        use <function>numberformatter.formatcurrency</function>.
```
nikic: I believe the correct syntax is `<methodname>NumberFormatter::formatCurrency</methodname>`

reference/strings/functions/money-format.xml
```diff
@@ -229,6 +233,32 @@
...
+       </entry>
+       </entry>
```
nikic: There is a stray `</entry>` here.

nikic:
Small ping, did you see my last two comments?

BackEndTea:
I saw them, and then comletely forgot 😅. I fixed them now

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349216.

BackEndTea:
Thanks @cmb69 & @nikic

PR #38, opened 2020-02-10
Fix the table of integer values on setopt page

ekinhbayar:
CURLOPT_SOCKS5_AUTH has one extra entry tag, so it's making the table have 4 columns.
This PR makes the extra entry another para tag to fix the table.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349197.

PR #39, opened 2020-02-13
Clarification of return values descriptions for mysqi fetch functions

colshrapnel:
The first intention was to remove "of strings" as factually incorrect. In mysqli, a null, int or float value can be returned in the resulting array or object depends on the configuration. 

Besides, "string" being sort of a synonym for "row", made an ambiguity which resulted in the disastrous  Russian translation where it now says "returns an array of rows".

Then some minor unification and clarification followed.

### DIFF
--- a/reference/mysqli/mysqli_result/fetch-assoc.xml
+++ b/reference/mysqli/mysqli_result/fetch-assoc.xml
@@ -39,7 +39,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an associative array of strings representing the fetched row in the result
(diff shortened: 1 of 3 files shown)


afilina:
Looks like there might be some overlap with PR #207.

kamil-tekiela:
[My PR](https://github.com/php/doc-en/pull/207) is closed now. I didn't know how to resolve the conflict between this PR and mine. However, I believe this PR still has good improvements, so if you could rebase it and address the conflicts it would be great. Sorry for creating a conflicting PR.

salathe:
Superseded by #841

PR #42, opened 2020-02-16
Fix #79256: Generator::next not behaving as documented

ekinhbayar:
Attempting to use the return value of a void function will evaluate to NULL, hence calling next() has the same effect as send(null) https://3v4l.org/PsDrI9 so that sentence is valid.

This change is just to not say 'it is equivalent', as they aren't fully the same.


cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=349409.

PR #43, opened 2020-02-16
Fix #79168: Missing `__construct` method documentation in SPL Exceptions

carusogabriel:
Generalizing the `Exception::__construct` documentation to `methodsynopsis`
instead of `constructorsynopsis` will include it in SPL Exception queries.


carusogabriel:
@cmb69 So the action point here is to add `<xi:include />` of `constructorsynopsis` to all SPL Exception with an `<xi:fallback />`?

cmb69:
Yes. Basically like http://svn.php.net/viewvc?view=revision&revision=345155.

carusogabriel:
ping @cmb69 just to make sure everything is okay and we can merge this one :)

PR #45, opened 2020-02-16
Fix #78130: Add SplPriorityQueue::isCorrupted version

carusogabriel:
Introduced via php/php-src@c54045a1f949, documented via php/doc-en@5f6c0c7994f3 but it was never added to `versions.xml`.

Thanks to @nikic to point how to fix this one 👍


carusogabriel:
@php-pulls merge

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=349233.

PR #46, opened 2020-02-18
file_get_contents() throws error for directories

ADmad:


reference/filesystem/functions/file-get-contents.xml
```diff
@@ -114,7 +114,8 @@
   &reftitle.errors;
   <para>
    An <constant>E_WARNING</constant> level error is generated if <parameter>filename</parameter> cannot be found, <parameter>maxlength</parameter>
-   is less than zero, or if seeking to the specified <parameter>offset</parameter> in the stream fails.
+   is less than zero, or if seeking to the specified <parameter>offset</parameter> in the stream fails. As of PHP 7.4 an <constant>E_NOTICE</constant>
+   level error is generated if <parameter>filename</parameter> points to a directory.
```
cmb69: Well, on Windows an E_WARNING is raised (for earlier versions as well). Also, if we document this, we should document this for all relevant file functions.

cmb69: See also https://bugs.php.net/bug.php?id=79166

afilina:
Pinging @ADmad in case this fell through the cracks.

reference/filesystem/functions/file-get-contents.xml
ADmad: Sorry but it's not possible for me to check the warnings/notices for various PHP versions on different OSes. Not can I check for which other related functions the behaviour has changed. 

So someone more knowledge can use the description and/or close PR.

afilina: @cmb69 Do we have tooling available to be able to verify it on various OS/PHP combinations, or do we need to set up our own? I'd like to give it a stab if it's feasible.

cmb69: PHP 7.4 CI is currently running on Linux, macOS and Windows, so a draft PR with one or more test cases could be used to verify the behavior.

afilina: I crafted a PR, but there is no Windows job in the [pipeline][azure]. The first three are Ubuntu 24.04.1 and the last one is OSX 10.15.7.

[azure]:https://dev.azure.com/phpazuredevops/PHP/_build/results?buildId=14080&view=logs&j=0d3730b8-0d59-574b-ab4a-5e189a74d53b

afilina: Additionally, the pipeline took longer for an unknown reason and got automatically cancelled. I don't know how to restart it, trigger a run or where to ask questions about this.

afilina: Confirmed difference on Windows with a test: 
```
001+ Warning: file_get_contents(C:\projects\php-src\ext\standard\tests\file): Failed to open stream: Permission denied in C:\projects\php-src\ext\standard\tests\file\file_get_contents_error_folder.php on line 2
001- Notice: file_get_contents(): Read of %d bytes failed with errno=21 Is a directory in %s on line %d
```

Now to write a Windows-specific test and update the docs. I'll also open PRs for other 7.0+ versions to check whether there are differences in how it's handled by Windows on different PHP versions. I suspect not, but worth putting the extra effort.

afilina: I'll just document it for this function for the sake of the PR, as to do otherwise would be too big a bite.

cmb69: I think checking with PHP 7.3 and 7.4 should be sufficient. There has been a relevant change between these versions, IIRC.

afilina: Alright, first PR is in: https://github.com/php/php-src/pull/6600

cmb69:
I'm going with a somewhat different wording.

PR #48, opened 2020-02-20
Fix \Ds\Map::values() example

enumag:


reference/ds/ds/map/values.xml
```diff
@@ -46,12 +46,12 @@ var_dump($map->values());
    &example.outputs.similar;
    <screen>
 <![CDATA[
-object(Ds\Map)#2 (3) {
-  [0]=>
+class Ds\Vector#2 (3) {
```
nikic: This does not look like a standard var_dump. Maybe you have XDebug enabled?

cmb69:
DS 1.2.9 produces for me:
````
object(Ds\Vector)#2 (3) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
}
````
So just the class name would have to fixed in the docs.

nikic:
@php-pulls merge

enumag:
@nikic You're right, it was because of xdebug. I had no idea it changed var_dump format.

PR #49, opened 2020-02-24
var_export: Mention 7.3.0 change for stdClass

hikari-no-yume:
Based in part on the text already in migration73


hikari-no-yume:
(See https://github.com/php/php-src/pull/2420 if interested in the PHP change this documents!)

reference/var/functions/var-export.xml
```diff
@@ -73,11 +73,16 @@
      </thead>
      <tbody>
       <row>
+       <entry>7.3.0</entry>
+       <entry>
+        Now exports <classname>stdClass</classname> objects as an array cast to
+        an object (<code>(object) array( ... )</code>), rather than using the
+        nonexistent method <methodname>stdClass::__setState</methodname>.
+       </entry>
```
nikic: Could also add an example below :)

hikari-no-yume: That is not a bad idea. I will do that.

reference/var/functions/var-export.xml
```diff
@@ -135,7 +143,31 @@ echo $v;
   </para>
   <para>
    <example>
-    <title>Exporting classes since PHP 5.1.0</title>
+    <title>Exporting stdClass (since PHP 7.1.0)</title>
```
hikari-no-yume: @nikic, do you like it? :)

reference/var/functions/var-export.xml
```diff
@@ -228,8 +260,9 @@ object(A)#2 (2) {
    <para>
     To be able to evaluate the PHP generated by <function>var_export</function>,
     all processed objects must implement the magic <link
-    linkend="object.set-state">__set_state</link> method. Note that StdClass
-    does not implement __set_state().
+    linkend="object.set-state">__set_state</link> method. The only exception is
+    <classname>stdClass</classname>, which is exported using an array cast to
```
hikari-no-yume: stdClass is an interesting class because while it _is_ mentioned in the PHP manual, it doesn't have its own page, it's just listed on https://www.php.net/manual/en/reserved.classes.php

nikic:
@php-pulls merge

hikari-no-yume:
oops… this is why I miss the online editor's preview function, back when it still worked

old and new content on same row

hikari-no-yume:
Also, oops, I put 7.1.0 in the example title. Will make a follow-up PR to fix both

cmb69:
@hikari-no-yume [done](http://svn.php.net/viewvc?view=revision&revision=349260) :)

PR #50, opened 2020-02-25
Add preg_last_error_msg documentation

nicoSWD:
This PR adds the documentation for `preg_last_error_msg`, added in [php-src/pull/5185](https://github.com/php/php-src/pull/5185)

Here's a rendered version of the manual page
Screenshot

### DIFF
--- a/reference/pcre/functions/preg-last-error-msg.xml
+++ b/reference/pcre/functions/preg-last-error-msg.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<refentry xml:id="function.preg-last-error-msg" xmlns="https://docbook.org/ns/docbook">
+<refentry xml:id="function.preg-last-error-msg" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>preg_last_error_msg</refname>
   <refpurpose>Returns the error message of the last PCRE regex execution</refpurpose>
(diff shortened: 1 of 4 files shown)


nikic:
@salathe @cmb69 What's our policy on docs for functions in not yet released PHP versions?

cmb69:
Usually, we document only after features have been released (GA).

nicoSWD:
@nikic @cmb69 fair enough. Should I close and re-submit this PR once it's released?

nikic:
As we tend to do lag behind when it comes to updating the manual for new PHP versions, maybe it would make sense to include the documentation right away, but add a warning box along the lines of "This function is only available in pre-release versions of PHP" or so?

cmb69:
Another option would be to keep such PRs open, but tag them with something like "PHP 8.0", so they can be merged when the migration guide is written.

salathe:
> Another option would be to keep such PRs open, but tag them with something like "PHP 8.0", so they can be merged when the migration guide is written.

For what it's worth, this is my preference.

cmb69:
I've did that now, but actually a PHP 8.0 milestone might be more apt.

divinity76:
can't we just have a PHP8 branch where we put all the PHP8 stuff and merge on release?

cmb69:
Thanks for the fine patch, @nicoSWD! I have applied it as http://svn.php.net/viewvc?view=revision&revision=350809.

PR #51, opened 2020-02-25
Allow only bool|float|int|string for sprintf $args

simPod:
Other types for $args than `bool|float|int|string` won't work with `sprintf()`.


cmb69:
Thanks for the PR! However, that is not quite correct; at least resources can be passed to sprintf() as well (see https://3v4l.org/04gHY).

Anyhow, we do not yet fully support union types in the documentation (see php/phd#22), so this would have to wait.

nikic:
I don't think this makes sense anyway. Semantically, arguments can be anything, and the specific accepted types are determined by the printf modifiers.

simPod:
Understand, thanks

PR #52, opened 2020-02-28
#79273: Document nullable properties on DOMDocument and DOMNode

beberlei:
https://bugs.php.net/bug.php?id=79273

### DIFF
--- a/reference/dom/domdocument.xml
+++ b/reference/dom/domdocument.xml
@@ -66,7 +66,7 @@
     </fieldsynopsis>
     <fieldsynopsis>
      <modifier>public</modifier>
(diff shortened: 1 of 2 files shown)


cmb69:
Thanks for the PR! The changes match the implementation, except for `Node::prefix` which appears to always return a string. I think the docs should reflect the implementation, or any erroneous implementation should be fixed as soon as possible.

cmb69:
Ugh, but see the discussion in PR #54 and php/phd#22.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351078.

PR #53, opened 2020-03-02
Fix #74284: Update runtime configuration page for memcached

jboulen:
I based my work on [php-memcached/memcached.ini](https://github.com/php-memcached-dev/php-memcached/blob/master/memcached.ini).

It should also [fix this issue](https://bugs.php.net/bug.php?id=74284).

### DIFF
--- a/reference/memcached/ini.xml
+++ b/reference/memcached/ini.xml
@@ -19,81 +19,177 @@
     <tbody>
      <row>
       <entry><link linkend="ini.memcached.sess-locking">memcached.sess_locking</link></entry>
-      <entry>1</entry>
+      <entry>On</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 0.1.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-consistent-hash">memcached.sess_consistent_hash</link></entry>
-      <entry>0</entry>
+      <entry>On</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 2.1.0. Default value is On since memcached 3.0.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-binary">memcached.sess_binary</link></entry>
-      <entry>0</entry>
+      <entry>Off</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 2.0.0. Replaced by memcached.sess_binary_protocol since memcached 3.0.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-lock-wait">memcached.sess_lock_wait</link></entry>
       <entry>150000</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 0.1.0. Removed since memcached 3.0.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-prefix">memcached.sess_prefix</link></entry>
       <entry>memc.sess.key.</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 0.1.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-number-of-replicas">memcached.sess_number_of_replicas</link></entry>
       <entry>0</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 2.1.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-randomize-replica-read">memcached.sess_randomize_replica_read</link></entry>
-      <entry>0</entry>
+      <entry>Off</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 2.1.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.sess-remove-failed">memcached.sess_remove_failed</link></entry>
-      <entry>0</entry>
+      <entry>On</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 2.1.0. Replaced by memcached.sess_remove_failed_servers since memcached 3.0.0.</entry>
      </row>
      <row>
       <entry><link linkend="ini.memcached.compression-type">memcached.compression_type</link></entry>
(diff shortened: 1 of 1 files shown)


reference/memcached/ini.xml
```diff
@@ -85,14 +139,32 @@
      </row>
      <row>
       <entry><link linkend="ini.memcached.serializer">memcached.serializer</link></entry>
-      <entry>php</entry>
+      <entry>igbinary</entry>
```
TysonAndre: This should continue to be `php` - PHP is the default. igbinary can only be used if the ini setting is changed, and the igbinary extension is installed.

jboulen: The comment on [default memcached.ini](https://github.com/php-memcached-dev/php-memcached/blob/master/memcached.ini#L127) indicates : 
> ; The default is igbinary if available, then msgpack if available, then php otherwise.
> ;memcached.serializer = "igbinary"

So I will update [this line](https://github.com/php/doc-en/pull/53/files#diff-e55e997e08f9c83bde010c4b38f96056L300) too.
> The default is igbinary if available and php otherwise.

But what default value should we display in documentation ?

cmb69: I think using "php" as default, but explaining that it depends on compile time (options), might be best.

cmb69:
I can't really comment on memcached, but please note that the PHP manual is supposed to document also older versions. So if some of these changes would have been introduced only in recent versions, respective changelog entries should be added (or be documented otherwise).

reference/memcached/ini.xml
jboulen: Can I use _changelog_ column or it's better to let explanation in [settings description](https://www.php.net/manual/en/memcached.configuration.php#ini.memcached.serializer) ?

jboulen:
I will try to fill _changelog_ column. Thank you for feedback.

reference/memcached/ini.xml
TysonAndre: I'm sorry. I was mistaken, this was right -- I didn't expect igbinary and msgpack to be the default and thought it'd be opt-in.

If igbinary is available, it's igbinary, if msgpack is available, it's msgpack, and otherwise, it's php.

```
#if defined(HAVE_MEMCACHED_IGBINARY)
# define SERIALIZER_DEFAULT SERIALIZER_IGBINARY
# define SERIALIZER_DEFAULT_NAME "igbinary"
#elif defined(HAVE_MEMCACHED_MSGPACK)
# define SERIALIZER_DEFAULT SERIALIZER_MSGPACK
# define SERIALIZER_DEFAULT_NAME "msgpack"
#else
# define SERIALIZER_DEFAULT SERIALIZER_PHP
# define SERIALIZER_DEFAULT_NAME "php"
#endif /* HAVE_MEMCACHED_IGBINARY / HAVE_MEMCACHED_MSGPACK */
```

TysonAndre: https://github.com/php-memcached-dev/php-memcached/commit/abd0cae8de0016ef3482a65bda9c4488a75115be#diff-af2b8e0bfbc900b4a13b10a7af197df3 - It seems like igbinary's been the default if available since 1.0.1, as far as I can tell (commit was December 29th, 2009) - https://pecl.php.net/package/memcached

jboulen:
I tried to update information about availability or deprecation of each variable.

If someone can review this PR, it will be great !

cmb69:
@sodabrew, since you seem to be the current lead maintainer, could you please review?

afilina:
Pinging @sodabrew since this PR seems to have fell through the cracks.

sodabrew:
Thanks for the nudge, this is great. Needs to be rebased, I'll take care of that.

sodabrew (approved):
To the maintainers: please do rebase/merge this, it looks good!

reference/memcached/ini.xml
```diff
@@ -19,81 +19,177 @@
     <tbody>
      <row>
       <entry><link linkend="ini.memcached.sess-locking">memcached.sess_locking</link></entry>
-      <entry>1</entry>
+      <entry>On</entry>
       <entry>PHP_INI_ALL</entry>
-      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
+      <entry>Available since memcached 0.1.0.</entry>
```
Girgias, line 24:
```suggestion
      <entry>Available as of memcached 0.1.0.</entry>
```
Please do the following change for all other entries

reference/memcached/ini.xml
```diff
@@ -104,38 +200,32 @@
 
  <para>
   <variablelist>
-
    <varlistentry xml:id="ini.memcached.sess-locking">
      <term>
       <parameter>memcached.sess_locking</parameter>
-      <type>integer</type>
+      <type>boolean</type>
```
Girgias, line 206:
```suggestion
      <type>bool</type>
```

reference/memcached/ini.xml
```diff
@@ -104,38 +200,32 @@
...
        Use session locking. Valid values: On, Off, the default is On.
       </para>
      </listitem>
     </varlistentry>
-
     <varlistentry xml:id="ini.memcached.sess-consistent-hash">
      <term>
       <parameter>memcached.sess_consistent_hash</parameter>
-      <type>integer</type>
+      <type>boolean</type>
```
Girgias, line 217:
```suggestion
      <type>bool</type>
```

reference/memcached/ini.xml
```diff
@@ -104,38 +200,32 @@
...
+       If set to On, consistent hashing (libketama) is used for session handling. When consistent hashing is used, one can add or remove cache node(s) without messing up too much with existing keys. Default is On.
       </para>
      </listitem>
     </varlistentry>
-
     <varlistentry xml:id="ini.memcached.sess-binary">
      <term>
       <parameter>memcached.sess_binary</parameter>
-      <type>integer</type>
+      <type>boolean</type>
```
Girgias, line 228:
```suggestion
      <type>bool</type>
```

reference/memcached/ini.xml
```diff
@@ -172,31 +260,28 @@
...
+       Write data to a number of additional memcached servers. This is "poor man's HA" as libmemcached calls it. If this value is positive and sess_remove_failed_servers is enabled when a memcached server fails the session will continue to be available from a replica. However, if the failed memcache server becomes available again it will read the session from there which could have old data or no data at all. Default is 0.
       </para>
      </listitem>
     </varlistentry>
-
     <varlistentry xml:id="ini.memcached.sess-randomize-replica-read">
      <term>
       <parameter>memcached.sess_randomize_replica_read</parameter>
-      <type>integer</type>
+      <type>boolean</type>
```
Girgias, line 277:
```suggestion
      <type>bool</type>
```

reference/memcached/ini.xml
```diff
@@ -314,7 +402,192 @@
...
+       Sets the default memcached protocol for new connections. (To configure the memcached protocol for connections used by sessions, use memcached.sess_binary_protocol instead)
 
+       If set to On, the memcached binary protocol is used by default. If set to Off, the memcached text protocol is used. Default is Off
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry xml:id="ini.memcached.default-connect-timeout">
+     <term>
+      <parameter>memcached.default_connect_timeout</parameter>
+      <type>integer</type>
```
Girgias, line 421:
```suggestion
      <type>int</type>
```Apply this change to everything below

Girgias (changes requested):
Types should be ``int`` or ``bool`` and user ``as of`` instead of ``since``

cmb69:
Thank you!

PR #54, opened 2020-03-04
Fix #79341: XmlWriter docs does not say $uri is nullable

cmb69:
@salathe, is it okay to use nullable type markers (`?`)? Should that be clarified in https://www.php.net/manual/en/about.prototypes.php?

### DIFF
--- a/reference/xmlwriter/functions/xmlwriter-start-attribute-ns.xml
+++ b/reference/xmlwriter/functions/xmlwriter-start-attribute-ns.xml
@@ -54,6 +54,8 @@
      <listitem>
       <para>
        The namespace URI.
+       If <parameter>uri</parameter> is &null;, the namespace declaration will be
+       omitted.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/xmlwriter/functions/xmlwriter-start-attribute-ns.xml
+++ b/reference/xmlwriter/functions/xmlwriter-start-attribute-ns.xml
@@ -14,7 +14,7 @@
    <type>bool</type><methodname>XMLWriter::startAttributeNs</methodname>
    <methodparam><type>string</type><parameter>prefix</parameter></methodparam>
    <methodparam><type>string</type><parameter>name</parameter></methodparam>
-   <methodparam><type>string</type><parameter>uri</parameter></methodparam>
+   <methodparam><type>?string</type><parameter>uri</parameter></methodparam>
   </methodsynopsis>
   <para>&style.procedural;</para>
   <methodsynopsis>
--- a/reference/xmlwriter/functions/xmlwriter-start-attribute-ns.xml
+++ b/reference/xmlwriter/functions/xmlwriter-start-attribute-ns.xml
@@ -22,7 +22,7 @@
    <methodparam><type>resource</type><parameter>xmlwriter</parameter></methodparam>
    <methodparam><type>string</type><parameter>prefix</parameter></methodparam>
    <methodparam><type>string</type><parameter>name</parameter></methodparam>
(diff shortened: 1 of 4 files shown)


salathe:
> @salathe, is it okay to use nullable type markers (?)?

Only if PhD supports them and gives a sensible rendering.

jtojnar:
Oh, I stopped reading the description after “Experimental mirror”, otherwise I would send a PR myself.

Maybe we could also mention what will happen when null is provided? Something like:

> The namespace URI. If not `&null;`, `xmlns:$prefix` attribute with `$uri` value will be added when processing of the opening tag ends.

-----

Also, weirdly, passing `$prefix = null` and `$uri = 'foo'` acts as if `$prefix = ''` was passed. But perhaps that is a separate bug.

<details>
<summary>example</summary>

```
<?php

$writer = new XMLWriter;
$writer->openMemory();
$writer->setIndent(true);
$writer->startElement('foo');
$writer->writeAttribute('xmlns:example', 'https://www.example.com/');
$writer->startElementNS('example', 'meta', null);
$writer->writeAttributeNS(null, 'spout', 'https://www.example.com/', 'test');
$writer->endElement();
$writer->endElement();
echo $writer->outputMemory();
```
</details>

jtojnar:
By the way, the same problem occurs in the procedural variants. And there is also the change of  `resource` to `XmlWriter` you did in https://github.com/php/php-src/commit/d61bc63c4da9484fb320435ff62101c2291ed621#diff-00774916617141196399be5feceabe9b

cmb69:
> Only if PhD supports them and gives a sensible rendering.

Good point. I thought it would, but that's [not really the case](https://www.php.net/manual/en/class.commonmark-node-codeblock.php). See php/phd#22.

> Also, weirdly, passing $prefix = null and $uri = 'foo' acts as if $prefix = '' was passed.

I think this is intentional (although I don't understand why these parameters are nullable).

> By the way, the same problem occurs in the procedural variants.

Indeed, missed these. Will update the PR.

> And there is also the change of resource to XmlWriter […]

The manual is not supposed to document PHP 8 yet.

jtojnar:
> I think this is intentional (although I don't understand why these parameters are nullable).

It makes sense to have `$uri` nullable since that signifies it should inherit the URI assigned to the prefix in some parent node. Having `$prefix` nullable only makes sense if you have the elements serialized in some other format (e.g. JSON) and want to write them as XML without switching the used function – but that does seem to work for attributes which behave as if `$prefix = ''` was passed, producing non-well-formed XMLs:

```php
<?php

$writer = new XMLWriter;
$writer->openMemory();
$writer->setIndent(true);
$writer->startElement('foo');
$writer->writeAttribute('xmlns:example', 'https://www.example.com/');
$writer->startElementNS(null, 'meta', null);
$writer->writeAttributeNS(null, 'spout', null, 'test');
$writer->endElement();
$writer->endElement();
echo $writer->outputMemory();
```

```xml
<foo xmlns:example="https://www.example.com/">
 <meta :spout="test"/>
</foo>
```

jtojnar:
Opened a separate issue about the `$prefix = null` bug: https://bugs.php.net/bug.php?id=79344&thanks=4

jtojnar:
Until it is fixed, we should probably not document `$prefix` as nullable for `writeAttributeNS` and `startAttributeNS` or add a warning to those two functions.

jtojnar:
Also forgot to mention `$prefix = null` for elements corresponds to the default namespace set in `xmlns` attribute.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351079.

PR #55, opened 2020-03-06
Add `...$args` name to sprintf

simPod:


cmb69:
Did you check how that would render? I presume that comes out as `$...args` instead of the PHP syntax conforming `...$args`. If that's the case, [PhD](https://github.com/php/phd) should be modified first, before making these changes in the manual (and we certainly would want to change all occurances at once).

simPod:
Thank you for feedback, will look into it

cmb69:
This is now handled [somewhat differently](http://svn.php.net/viewvc?view=revision&revision=351129), but still this PR can be closed.

PR #56, opened 2020-03-06
Removed misleading example from rowCount()

kamil-tekiela:
The example given was confusing and misleading. If the purpose of it is to show how to get the count of matching records from the database then the example can be much shorter. `query()` should not be checked with `if` statement and there should not be 2 separate queries to the database if the actual count is not needed. If the number of fetched records is needed then you can simply use PHP's `count()` function.


reference/pdo/pdostatement/rowcount.xml
```diff
@@ -82,43 +82,22 @@ Deleted 9 rows.
...
-    }
-}
-
-$res = null;
-$conn = null;
-?>
+print "There are " .  $count . " matching records.";
 ]]>
     </programlisting>
     &example.outputs;
```
cmb69: Unless this example is supposed to be run against some prepared table, this should better be: `&example.outputs.similar;`

cmb69 (approved):
Thanks! Looks good to me.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351083.

PR #57, opened 2020-03-07
Change definition for varargs

simPod:
This would allow to properly render all arguments definitions

- arguments
- arguments by reference
- varible-length argument lists
- varible-length argument lists passed by reference

See https://github.com/php/phd/pull/24 and https://github.com/salathe/phpdoc-base/pull/2

_This is an example of what would have been needed to be done_

image


cmb69:
This is now handled [somewhat differently](http://svn.php.net/viewvc?view=revision&revision=351129), but still this PR can be closed.

PR #58, opened 2020-03-10
Added note about existing nodes being moved in the dom

PeeHaa:
Make it clear that existing nodes are being moved in the DOM when calling `\DOMNode::appendChild()`.

### DIFF
--- a/reference/dom/domnode/appendchild.xml
+++ b/reference/dom/domnode/appendchild.xml
@@ -20,6 +20,11 @@
    <methodname>DOMDocument::createTextNode</methodname> etc. or simply by using
    any other node.
   </para>
+  <note>
+   <simpara>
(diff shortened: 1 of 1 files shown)


cmb69:
Ah, thanks, good catch! I'm not sure that marking this as `<note>` is beneficial, though. In any way, I think this info should also be added to `DOMNode::insertBefore()`.

PeeHaa:
> Ah, thanks, good catch! I'm not sure that marking this as `<note>` is beneficial, though. In any way, I think this info should also be added to `DOMNode::insertBefore()`.

I will just put it in the para than and also look at `insertBefore`

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351088.

PR #59, opened 2020-03-10
Added changelog documentation to mb_convert_case()

thg2k:
Since I stumbled upon this problem, I think it's worth to mention in the changelog that 7.3.0 added the constants MB_CASE_FOLD and MB_CASE_*_SIMPLE

### DIFF
--- a/reference/mbstring/functions/mb-convert-case.xml
+++ b/reference/mbstring/functions/mb-convert-case.xml
@@ -41,8 +41,8 @@
        <constant>MB_CASE_LOWER</constant>, 
        <constant>MB_CASE_TITLE</constant>,
        <constant>MB_CASE_FOLD</constant>,
-       <constant>MB_CASE_LOWER_SIMPLE</constant>,
        <constant>MB_CASE_UPPER_SIMPLE</constant>,
+       <constant>MB_CASE_LOWER_SIMPLE</constant>,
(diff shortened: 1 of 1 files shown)


reference/mbstring/functions/mb-convert-case.xml
```diff
@@ -82,6 +82,35 @@
...
+     <tbody>
+      <row>
+       <entry>7.3.0</entry>
+       <entry>
+        <constant>MB_CASE_FOLD</constant>,
+        <constant>MB_CASE_UPPER_SIMPLE</constant>,
+        <constant>MB_CASE_LOWER_SIMPLE</constant>,
+        <constant>MB_CASE_TITLE_SIMPLE</constant>, and
+        <constant>MB_CASE_FOLD_SIMPLE</constant>,
+        <parameter>mode</parameter> were added.
```
nikic: Grammar is a bit off here, the `, $mode` is too much. Or should be something like `for $mode`. Now it sounds like the mode parameter itself was added.

thg2k: I agree with you but I actually copy-pasted it from [json-encode.xml](https://www.php.net/json_encode) (point 5.3.0), while trying to look for the exact "syntax" used in the rest of the docs...

Would you please take over the change? I can cancel the pull request.

nikic:
@php-pulls merge

PR #61, opened 2020-03-27
Operator Precedence - Add `print`, fix associativity of `instanceof` and unary operators

guilliamxavier:
See:
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4893#discussion_r398776216
- https://github.com/php/php-src/commit/2a04efe0e42c9d7ac150bc918e16edb96cf4f219 (part of)

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -266,6 +266,11 @@
         </entry>
        </row>
        <row>
+        <entry>right</entry>
+        <entry><literal>print</literal></entry>
+        <entry><function>print</function></entry>
+       </row>
+       <row>
         <entry>left</entry>
         <entry><literal>and</literal></entry>
         <entry>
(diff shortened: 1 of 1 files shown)


language/operators.xml
```diff
@@ -109,14 +109,14 @@
         </entry>
        </row>
        <row>
-        <entry>non-associative</entry>
+        <entry>left</entry>
```
cmb69: Shouldn't this be `(n/a)` as well?

guilliamxavier: `$a instanceof $b instanceof $c` is evaluated as
`($a instanceof $b) instanceof $c` (i.e. will always result in `false` but still valid syntax)
hence left-associative

cmb69: Ah, yes, you're right. Thanks!

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=349661.

PR #62, opened 2020-03-27
Return value of mb_regex_set_options() has changed

cmb69:
Cf. <https://github.com/php/php-src/pull/5264>.


cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=350810.

PR #64, opened 2020-04-01
Typo fix.

Crell:
I have no idea why there were no spaces here.  Now there are.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349581.

PR #65, opened 2020-04-01
Fix the return value of \zip_open()

LeSuisse:
\zip_open() can also return false [0].

[0] https://github.com/php/php-src/blob/php-7.4.4/ext/zip/php_zip.c#L1160

### DIFF
--- a/reference/zip/functions/zip-open.xml
+++ b/reference/zip/functions/zip-open.xml
@@ -36,8 +36,8 @@
   <para>
    Returns a resource handle for later use with
    <function>zip_read</function> and <function>zip_close</function>
-   or returns the number of error if <parameter>filename</parameter> does not
-   exist or in case of other error.
+   or returns either &false; or the number of error if <parameter>filename</parameter>
+   does not exist or in case of other error.
   </para>
  </refsect1>
  <refsect1 role="seealso">


cmb69:
I think this is [a case of](https://www.php.net/manual/en/functions.internal.php):

> If the parameters given to a function are not what it expects, such as passing an array where a string is expected, the return value of the function is undefined. In this case it will likely return NULL but this is just a convention, and cannot be relied upon.

We do not document these return values explicitly.

LeSuisse:
Fine by me, I can close the PR.

It's however a bit different of the quoted section since here the parameter might be a valid file path.

ret394:
don't fully get the code. if function return type can be anything other than `resource`, please try to document it in PHP manual when you can, it would help more than harm, even if against tradition.

>parameter might be a valid file path

I assume, `ZSTR_LEN(filename)` evaluates to be non-zero then and the if clause evaluates to `FALSE`?

thanks. have a nice day.

[ret394](https://github.com/ret394)

cmb69:
Given that PHP 8 has support for union types, and well defined function stubs, it makes sense to properly document failure return types in the manual. I would exclude returns for ZPP failures, though (which is mostly `null`). As such this PR is valid, but I'd like to have the return type in the `<methodsynopsis>` fixed as well. Please use the [dedicated union type notation](https://github.com/php/doc-base/commit/82d74e69032e89cd6c8f624f7baf9780c639ec71). Note that union types [cannot be rendered yet](https://github.com/php/phd/pull/30), though.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351061

LeSuisse:
Looks it was applied as https://svn.php.net/viewvc?view=revision&revision=351062

Thanks for the hint about the new union type notation and sorry did not got the time to get back to the PR :)

PR #67, opened 2020-04-01
Adding documentation for mb_str_split

TZubiri:
According to [this item](https://bugs.php.net/bug.php?id=79425) in the issue tracker, I looked into the new function and drafted a corresponding doc page. It's a pretty straightforward function.

Notes:

1. I couldn't test this due to configure.php followed by php .manual.xml yielding a parse error :/
```

php .manual.xml 
PHP Parse error:  syntax error, unexpected ']', expecting end of file in /home/tomas/p/php/doc-base/.manual.xml on line 9262

```

2.  I tried to add a couple of \<link linkend="reference.mbstring.supported-encodings"\> but I received errors on these, so I removed them. A link to the book and supported encodings on the See Also section would be nice.

### DIFF
--- a/reference/mbstring/functions/mb-str-split.xml
+++ b/reference/mbstring/functions/mb-str-split.xml
@@ -65,7 +65,7 @@
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member><function>str-split</function></member>
+    <member><function>str_split</function></member>
    </simplelist>
   </para>
  </refsect1>
--- a/reference/mbstring/functions/mb-str-split.xml
+++ b/reference/mbstring/functions/mb-str-split.xml
@@ -57,7 +57,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   <function>mb_str_split</function> returns an array of strings. In case of an error, returns False.
+   <function>mb_str_split</function> returns an array of strings, &return.falseforfailure;.
   </para>
  </refsect1>
 
--- a/reference/mbstring/functions/mb-str-split.xml
+++ b/reference/mbstring/functions/mb-str-split.xml
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>array</type><methodname>mb_str_split</methodname>
    <methodparam><type>string</type><parameter>string</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>split_length</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>split_length</parameter><initializer>1</initializer></methodparam>
    <methodparam choice="opt"><type>string</type><parameter>encoding</parameter><initializer>mb_internal_encoding()</initializer></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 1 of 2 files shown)


reference/mbstring/functions/mb-str-split.xml
```diff
@@ -0,0 +1,94 @@
...
+  <para>
+   <function>mb_str_split</function> returns an array of strings. In case of an error, returns False.
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>str-split</function></member>
```
cmb69, line 68: This should use the proper name of the function:
```suggestion
    <member><function>str_split</function></member>
```

reference/mbstring/functions/mb-str-split.xml
```diff
@@ -0,0 +1,94 @@
...
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   <function>mb_str_split</function> returns an array of strings. In case of an error, returns False.
```
cmb69, line 60: You could use the &return.falseforfailure; entity ref here:
```suggestion
   <function>mb_str_split</function> returns an array of strings, &return.falseforfailure;.
```

reference/mbstring/functions/mb-str-split.xml
```diff
@@ -0,0 +1,94 @@
...
+  <refname>mb_str_split</refname>
+  <refpurpose>Given a multibyte string, return an array of its characters.</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>array</type><methodname>mb_str_split</methodname>
+   <methodparam><type>string</type><parameter>string</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>split_length</parameter></methodparam>
```
cmb69, line 14: This parameter defaults to 1, so:
```suggestion
   <methodparam choice="opt"><type>int</type><parameter>split_length</parameter><initializer>1</initializer></methodparam>
```

cmb69:
Thanks for the PR! 

> I couldn't test this due to configure.php followed by php .manual.xml yielding a parse error :/

Just do `php doc-base/configure.php` :)

cmb69:
Ah, forgot: please also add an entry to [versions.xml](http://svn.php.net/viewvc/phpdoc/en/trunk/reference/mbstring/versions.xml?view=markup): `<function name="mb_str_split" from="PHP 7 &gt;= 7.4.0"/>`

TZubiri:
Thanks for looking into this, I still was not able to test the doc, I don't know what to do after php configure.php

Here's what I ran, $ are commands, > is output, // are comments.

> $ php doc-base/configure.php
> // redacted
> > All good. Saving .manual.xml... done.
> > All you have to do now is run 'phd -d /home/tomas/p/php/doc-base/.manual.xml'
> > If the script hangs here, you can abort with ^C.
> //  redacted
> $ php -d /home/tomas/p/php/doc-base/.manual.xml
> //nothing interesting happens, the shell just reads input until I use ctrl C, as if I were to run a bare php command, which makes sense since -d just defines a variable.

cmb69:
> I don't know what to do after php configure.php

If you just want to validate the docs, this is sufficient (if there are errors, these are displayed instead of "All good. Saving .manual.xml... done.")

If you want to actually build the manual, you need to [install and run phd](http://doc.php.net/tutorial/local-setup.php).

cmb69:
Thanks! Applied with minor tweaks as http://svn.php.net/viewvc?view=revision&revision=349604.

PR #68, opened 2020-04-05
Use consistent modifiers order "public" before "readonly"

thg2k:
Before this change:
 - public/readonly: 50 cases
 - readonly/public: 30 cases (mostly in DOM)

Reordered to be always public/readonly, in this order.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349623.

PR #69, opened 2020-04-13
List of Parser Tokens - Sort, update (PHP 7.4, versions)

guilliamxavier:
Cf https://3v4l.org/Q6FEn


cmb69:
Very nice, thanks!

Applied as http://svn.php.net/viewvc?view=revision&revision=349654.

PR #70, opened 2020-04-13
List of Keywords - Add fn (PHP 7.4)

guilliamxavier:


appendices/reserved.xml
```diff
@@ -139,17 +139,20 @@
        <entry>
         <link linkend="language.exceptions">finally</link> (as of PHP 5.5)
        </entry>
+       <entry>
+        <link linkend="migration74.new-features.core.arrow-functions">fn</link> (as of PHP 7.4)
```
guilliamxavier, line 143: I couldn't find a better link (see also #69)

cmb69: Probably, we should add a section about arrow functions to https://www.php.net/manual/en/functions.anonymous.php.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349701. (I have changed the link to point to the new section introduced via PR #71.)

PR #71, opened 2020-04-26
Document PHP 7.4 arrow functions

TysonAndre:
Fixes https://bugs.php.net/bug.php?id=79303
PHP 7.4 has been out since December.

Examples were taken from https://wiki.php.net/rfc/arrow_functions_v2
to illustrate places where arrow functions differed from anonymous
functions.

This was written to be similar to
https://www.php.net/manual/en/functions.anonymous.php

Related to #70 - it felt cleaner to introduce arrow functions after
anonymous functions to teach anonymous functions and their caveats first,
but it's easy enough to move this to a section of functions.anonymous.

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1737,6 +1737,170 @@ Warning: Cannot bind an instance to a static closure in %s on line %d
    </sect2>
 
   </sect1>
+  <sect1 xml:id="functions.arrow">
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349700.

PR #72, opened 2020-04-26
Link to the new arrow functions section

TysonAndre:
This is on the master branch, but not on php.net at the time of writing


cmb69:
It's already available on http://docs.php.net/manual/en/functions.arrow.php (which builds 4 or 6 times a day).

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349702.

PR #74, opened 2020-04-26
Fix other miscellaneous typos

TysonAndre:
Detected by codespell


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349709.

PR #76, opened 2020-04-26
Fix other typos outside of php module docs

TysonAndre:
https://en.wikipedia.org/wiki/Integral


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349711.

PR #77, opened 2020-04-29
fix return values documentation on linkinfo() function

redreinard:
The return values on this page are not correctly documented. I have updated them to more closely explain what the function returns.
Proof: https://3v4l.org/O8IbZ

### DIFF
--- a/reference/filesystem/functions/linkinfo.xml
+++ b/reference/filesystem/functions/linkinfo.xml
@@ -43,7 +43,7 @@
   <para>
    <function>linkinfo</function> returns the <literal>st_dev</literal> field
    of the Unix C stat structure returned by the <literal>lstat</literal>
-   system call. Returns 0 or &false; in case of error.
+   system call. Returns 0 or a positive integer on success or -1 in case of error.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/linkinfo.xml
```diff
@@ -43,7 +43,7 @@
   <para>
    <function>linkinfo</function> returns the <literal>st_dev</literal> field
    of the Unix C stat structure returned by the <literal>lstat</literal>
-   system call. Returns 0 or &false; in case of error.
+   system call. Returns 0 or a positive integer on success or -1 in case of error.
```
cmb69, line 46: The function returns `false` on an `open.base_dir` violation, though. Also instead of "0 or a positive integer on success", maybe "a non-negative integer on success" might be more succinct.

reference/filesystem/functions/linkinfo.xml
```diff
@@ -66,6 +66,14 @@
         2008 or greater).
        </entry>
       </row>
+      <row>
```
cmb69, line 69: We do not add documentation fixes to the changelog.

reference/filesystem/functions/linkinfo.xml
redreinard: I have submitted and improved description of the return values:

reference/filesystem/functions/linkinfo.xml
redreinard: I have removed this row from the changelog.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=349726.

PR #78, opened 2020-05-02
Document other primes for `opcache.max_accelerated_files`

TysonAndre:
See https://github.com/php/php-src/blob/master/ext/opcache/zend_accelerator_hash.c

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -325,7 +325,7 @@
       The maximum number of keys (and therefore scripts) in the OPcache hash
       table. The actual value used will be the first number in the set of
       prime numbers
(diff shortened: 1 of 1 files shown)


cmb69:
Shouldn't the values < 223 be documented as well?

TysonAndre:
> The actual value used **will be the first number in the set of
> prime numbers that is greater than or equal to the configured value.... The minimum value is 200.** 

The minimum value a user can set this to is 200. The smallest prime in that set greater than or equal to 200 is 223.

cmb69:
Ah, thanks!

Applied as http://svn.php.net/viewvc?view=revision&revision=349735.

PR #79, opened 2020-05-04
Remove unnecessary "Sometimes"

thutter:
The statement(s) nested in `while` won't run if the conditional evaluates to `FALSE`. Not even "Sometimes".


cmb69:
Thanks for the PR! ("sometimes" is not unnecessary, but plain wrong here)

Applied as http://svn.php.net/viewvc?view=revision&revision=349746.

PR #80, opened 2020-05-05
Update flag --with-onig information

demiurge-ash:


nikic:
This should also say what it has been replaced with (pkg-config).

reference/mbstring/configure.xml
```diff
@@ -41,6 +41,8 @@
       <productname>Oniguruma</productname> is already installed on the system,
       <option role="configure">--with-onig[=DIR]</option> can be specified to
       use the installed library.
+      As of PHP 7.4.0 <option role="configure">--with-onig</option> has been removed
+      and now pkg-config is used to detect libonig library.
```
nikic, line 45:
```suggestion
      and pkg-config is now used to detect the libonig library.
```

nikic:
@php-pulls merge

PR #82, opened 2020-05-07
Remove useless declare statement from example

SjonHortensius:
declare(encoding) is only needed when including UTF-8 directly in the code, which it isn't in this example. It even produces additional errors, see https://3v4l.org/fWYRr while it works fine in https://3v4l.org/ePTAk


nikic:
@php-pulls merge

PR #83, opened 2020-05-10
IntlCalendar::setMinimalDaysInFirstWeek: fix procedural function name

jrfnl:
See: https://github.com/php/php-src/blob/b452d5923de3bfef3268bcea289d59d6bc789437/ext/intl/calendar/calendar.stub.php#L145


nikic:
@php-pulls merge

PR #84, opened 2020-05-10
FDF: extension was moved to PECL / no longer bundled with PHP as of PHP 5.3

jrfnl:
As per: https://www.php.net/manual/en/fdf.installation.php

It is unclear to me what the current version nr is of the PECL extension, but showing the functions as if they are available in PHP > 5.4 seems very wrong either way.

### DIFF
--- a/reference/fdf/versions.xml
+++ b/reference/fdf/versions.xml
@@ -4,41 +4,41 @@
   Do NOT translate this file
 -->
 <versions> 
- <function name='fdf_add_doc_javascript' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
- <function name='fdf_add_template' from='PHP 4, PHP 5, PHP 7'/>
- <function name='fdf_close' from='PHP 4, PHP 5, PHP 7'/>
- <function name='fdf_create' from='PHP 4, PHP 5, PHP 7'/>
- <function name='fdf_enum_values' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
- <function name='fdf_errno' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
- <function name='fdf_error' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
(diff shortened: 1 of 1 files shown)


reference/fdf/versions.xml
```diff
@@ -4,41 +4,41 @@
...
- <function name='fdf_set_flags' from='PHP 4 &gt;= 4.0.2, PHP 5, PHP 7'/>
- <function name='fdf_set_javascript_action' from='PHP 4 &gt;= 4.0.2, PHP 5, PHP 7'/>
- <function name='fdf_set_on_import_javascript' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
- <function name='fdf_set_opt' from='PHP 4 &gt;= 4.0.2, PHP 5, PHP 7'/>
- <function name='fdf_set_status' from='PHP 4, PHP 5, PHP 7'/>
- <function name='fdf_set_submit_form_action' from='PHP 4 &gt;= 4.0.2, PHP 5, PHP 7'/>
- <function name='fdf_set_target_frame' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
- <function name='fdf_set_value' from='PHP 4, PHP 5, PHP 7'/>
- <function name='fdf_set_version' from='PHP 4 &gt;= 4.3.0, PHP 5, PHP 7'/>
+ <function name='fdf_add_doc_javascript' from='PHP 4 &gt;= 4.3.0, PHP 5 &lt; 5.3.0, PECL/fdf'/>
```
nikic, line 7:
```suggestion
 <function name='fdf_add_doc_javascript' from='PHP 4 &gt;= 4.3.0, PHP 5 &lt; 5.3.0, PECL fdf SVN'/>
```

I think. At least that's what ming, hwapi, iisfunc do.

jrfnl: Done. As far as I remember, the annotations used by the various PECL extensions are inconsistent, but anything is better than what it was now.

nikic:
@php-pulls merge

PR #85, opened 2020-05-10
Ming: extension was moved to PECL / no longer bundled with PHP as of PHP 5.3

jrfnl:
As per: https://www.php.net/manual/en/ming.install.php

A number of functions/classes/methods still showed as if they were available in PHP 7.


nikic:
@php-pulls merge

PR #86, opened 2020-05-10
XLSWriter: extension is not bundled with PHP

jrfnl:
As per: https://www.php.net/manual/en/xlswriter.installation.php

So the version information shouldn't give the impression that the classes/methods are available in PHP itself. They are only available in the PECL extension.


nikic:
@php-pulls merge

PR #87, opened 2020-05-10
Swoole: extension is not bundled with PHP

jrfnl:
As per: https://www.php.net/manual/en/swoole.installation.php

So the version information shouldn't give the impression that the classes/methods are available in PHP itself. They are only available in the PECL extension.

### DIFF
--- a/reference/swoole/versions.xml
+++ b/reference/swoole/versions.xml
@@ -6,895 +6,895 @@
 
 <versions>
  <!-- Functions -->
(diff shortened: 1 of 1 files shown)


carusogabriel:
@php-pulls merge

cmb69:
Thanks for the PR! Applied as http://svn.php.net/viewvc?view=revision&revision=349772.

jrfnl:
You're welcome ;-)

PR #88, opened 2020-05-10
SPLObjectStorage: fix class version nr

jrfnl:
If the class wasn't introduced until PHP 5.3.0, how can it be that certain methods would have existed before PHP 5.3 ?

_Updated to reflect @nikic's feedback - not the method version nrs were wrong, but the class version nr._

### DIFF
--- a/reference/spl/versions.xml
+++ b/reference/spl/versions.xml
@@ -977,7 +977,7 @@
  <function name="splsubject::detach" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
  <function name="splsubject::notify" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
 
- <function name="splobjectstorage" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
+ <function name="splobjectstorage" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
  <function name="splobjectstorage::addall" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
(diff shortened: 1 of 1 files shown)


nikic:
According to https://github.com/php/php-src/blob/PHP-5.2/ext/spl/spl_observer.c SplObjectStorage was in 5.2 already.

jrfnl:
> According to https://github.com/php/php-src/blob/PHP-5.1/ext/spl/spl_observer.c SplObjectStorage was in 5.1 already.

In that case, the class version nr was wrong. I will update the PR.

nikic:
@php-pulls merge

reference/spl/versions.xml
```diff
@@ -977,7 +977,7 @@
  <function name="splsubject::detach" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
  <function name="splsubject::notify" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
 
- <function name="splobjectstorage" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
+ <function name="splobjectstorage" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
  <function name="splobjectstorage::addall" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
  <function name="splobjectstorage::attach" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
  <function name="splobjectstorage::contains" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
```
carusogabriel, line 983: @jrfnl I this we should fix `SPLObjectStorage`'s methods versions as well :)

PR #89, opened 2020-05-12
PHP 7.4: PCRE: document new $flags param for preg_replace_callback[_array]()

jrfnl:
The PHP 7.4 `$flags` parameter as added to the `preg_replace_callback()` and `preg_replace_callback_array()` functions was not yet included in the documentation.

In this PR, I'm
* Adding the parameter to the function signature.
* Adding minimal parameter documentation with a reference to the more detailed explanation in the parameter documentation of the [`preg_match()`](https://www.php.net/manual/en/function.preg-match.php) function.
* Adding a changelog entry for the parameter to each function.

:crossed_fingers: I hope I've got the xml syntax right for this change. I'm not 100% familiar yet with it, so please bear with me if I've got it wrong. (and just tell me what I've got wrong & I'll fix it promptly)

Refs:
* https://github.com/php/php-src/pull/3958
* https://github.com/php/php-src/blob/2f1398dad934086b605073c51af3118c8eff28b1/UPGRADING#L304-L308

### DIFF
--- a/reference/pcre/functions/preg-replace-callback-array.xml
+++ b/reference/pcre/functions/preg-replace-callback-array.xml
@@ -14,6 +14,7 @@
    <methodparam><type>mixed</type><parameter>subject</parameter></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>limit</parameter><initializer>-1</initializer></methodparam>
    <methodparam choice="opt"><type>int</type><parameter role="reference">count</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer>0</initializer></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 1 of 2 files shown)


jrfnl:
Anything I need to update for this PR to be considered ?

cmb69:
Thanks for the ping! The patch is good, but there is no need to explicitly link functions; just marking them up as `<function>func_name</function>` is sufficient (links are then inserted automatically if the function is documented).

I've applied that as http://svn.php.net/viewvc?view=revision&revision=349965.

jrfnl:
> there is no need to explicitly link functions; just marking them up as <function>func_name</function> is sufficient

Thanks, this was one of the first PRs I send in. I'm slowly learning how it all works and hopefully getting better at it ;-)

PR #90, opened 2020-05-12
list(): add changelog entry for reference assignments

jrfnl:
~~... and link both the new entry as the entry for keyed lists to the relevant migration guide information (same as the other entries of this particular changelog).~~

++... and remove the links in the changelog.++

Ref: https://www.php.net/manual/en/function.list

### DIFF
--- a/reference/array/functions/list.xml
+++ b/reference/array/functions/list.xml
@@ -76,6 +76,12 @@
      </thead>
      <tbody>
       <row>
+       <entry>7.3.0</entry>
+       <entry>
+        Support for reference assignments in array destructuring was added.
+       </entry>
+      </row>
+      <row>
        <entry>7.1.0</entry>
        <entry>
         It is now possible to specify keys in <function>list</function>. This
@@ -85,27 +91,21 @@
       <row>
        <entry>7.0.0</entry>
        <entry>
-        <link linkend="migration70.incompatible.variable-handling.list.order">
-         The order that the assignment operations are performed in has
-         changed.
-        </link>
+        The order that the assignment operations are performed in has
+        changed.
        </entry>
       </row>
       <row>
        <entry>7.0.0</entry>
        <entry>
-        <link linkend="migration70.incompatible.variable-handling.list.empty">
-         <function>list</function> expressions can no longer be completely
-         empty.
-        </link>
+        <function>list</function> expressions can no longer be completely
+        empty.
        </entry>
       </row>
       <row>
        <entry>7.0.0</entry>
        <entry>
-        <link linkend="migration70.incompatible.variable-handling.list.string">
-         Strings can no longer be unpacked.
-        </link>
+        Strings can no longer be unpacked.
        </entry>
       </row>
      </tbody>


reference/array/functions/list.xml
```diff
@@ -75,11 +75,21 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>7.3.0</entry>
+       <entry>
+        <link linkend="migration73.new-features.core.destruct-reference">
+         Added support for reference assignments in array destructuring.
+        </link>
```
cmb69, line 83: I don't think that linking full sentences is a good idea, even though there is precedence on that page. Also, I don't see much use in linking back to the migration guide at all, since it doesn't really provide more info.

jrfnl: The migration guide does offer a code sample of what is meant by it, so there's that.

What would you suggest for this ? Remove the links altogether for all these changelog entries ? Keep the links, but have them only on a select part of the sentence where it makes the most sense ?

cmb69: I would prefer to remove these links altogether, but rather to have detailed information on the list() manual page. I've just seen that this entry could be linked to [example #6](https://www.php.net/manual/en/function.list.php#example-6331) instead. And I'd link to that example from the respective section of the migration guide (maybe just `See <link>example below</link>` or so).

@salathe, thoughts?

salathe: I echo @cmb69's comments, we should not be adding links back to the migration guide.   If there is anything useful in the migration guide that is missing from the manual page proper, then the latter should be updated.

jrfnl: That would mean this PR needs a lot more changes as reference assignments in lists aren't mentioned on the [`list()`](https://www.php.net/manual/en/function.list.php) page at all.

I currently don't have inspiration for that. For now, I will just update the PR to remove the links (including for the older entries).

salathe: > That would mean this PR needs a lot more changes as reference assignments in lists aren't mentioned on the [`list()`](https://www.php.net/manual/en/function.list.php) page at all.
> 
> I currently don't have inspiration for that. For now, I will just update the PR to remove the links (including for the older entries).

There's no need to make lots of changes at once.  Incremental improvements are still improvements, and are very welcome! 👍

reference/array/functions/list.xml
```diff
@@ -75,37 +75,37 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>7.3.0</entry>
+       <entry>
+         Support for reference assignments in array destructuring was added.
```
cmb69, line 81: Nit: we prefer indentation to be 1 space. (same below)

jrfnl: Darn. I actually knew that and forgot to adjust it after removing the link tags. Fixed now.

jrfnl:
Anything I can do to move this PR forward ?

cmb69:
LGTM now, thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350044.

PR #91, opened 2020-05-12
Filter: add missing constants

jrfnl:
Add a number of constants missing from the Predefined Constants page of the Filter extension.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349855.

PR #92, opened 2020-05-13
DOMImplementation: fix typo

jrfnl:
If I read the source correctly, `DOMImplementation` is a class, not an interface.

Ref: https://github.com/php/php-src/blob/5322de1ba8428c0231c972586217f9b1b705c45d/ext/dom/domimplementation.stub.php


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349859.

PR #93, opened 2020-05-13
SimpleXML: add missing changelog entries for $is_prefix

jrfnl:
According to the PHP 5.2 migration guide, the `$is_prefix` parameter was added to three methods and two functions in PHP 5.2, but only two of those had a changelog entry about this change.

Ref: https://www.php.net/manual/en/migration52.parameters.php


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349881.

I also reordered the sections in a [separate commit](http://svn.php.net/viewvc?view=revision&revision=349882) (__construct had it also wrong).

PR #95, opened 2020-05-13
Recode: update installation instructions

jrfnl:
As of PHP 7.4, the extension is no longer bundled with PHP, but moved to PECL.

The way I've annotated this emulates the installation instructions for [POSIX regex](https://www.php.net/manual/en/regex.installation.php).


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349879.

PR #96, opened 2020-05-13
Mcrypt/installation: mention "move" version nr

jrfnl:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349868.

I think that you should apply for a [php.net doc account](https://www.php.net/git-php.php). :)

jrfnl:
> I think that you should apply for a php.net doc account. :)

@cmb69 I appreciate the whuffies ;-)

What would be the advantage of me having a php.net doc account ?
I'm generally used to working in a pull-review flow (i.e. don't merge your own PRs, always have someone review), so I'd still be sending in PRs in that case.
I also feel like I'm still very much finding my way in how the docs-setup works, so really do appreciate a second opinion for the time being.

Either way, I definitely appreciate the encouragement. This current round of PRs is the result of updates I'm making to [PHPCompatibility](https://github.com/PHPCompatibility/PHPCompatibility/) and discrepancies I noticed because of that.

Going through the docs for this update round, I also noticed a lot of inconsistencies in page layouts, phrasing and extension page organisation.
Would PRs to make various docs more consistent/readable be welcome ?
Example of something I think might be useful: https://twitter.com/jrf_nl/status/1259877705698418689

cmb69:
> I'm generally used to working in a pull-review flow […]

A full pull-review workflow certainly has advantages, but obvious improvements/fixes (such as this PR or PR #98) could be committed directly; after all, that saves some time (especially regarding the SVN/Git dichotomy we currently have to live with, and also because CI does not (yet) produce rendered docs). Whenever there are doubts regarding a particular change, a PR (still) is fine.

> This current round of PRs is the result of updates I'm making to PHPCompatibility and discrepancies I noticed because of that.

This is very much appreciated!

> Would PRs to make various docs more consistent/readable be welcome ?

Generally, yes. Regarding the "table vs def list" question: in my opinion, tables are mostly bad for the purposes they're used in the current docs; in this particular case, a `variablelist` makes much more sense semantically, although the rendering/presentation should definitely be improved. But there are worse cases, e.g. [tables which overflow](https://www.php.net/manual/en/filter.filters.sanitize.php) on moderately sized browser windows, or even ["presentational tables"](https://www.php.net/manual/en/reserved.keywords.php) which yield [terrible diffs](https://github.com/php/doc-en/pull/70/files) when they have to be amended.

PR #97, opened 2020-05-13
XMLWriter: fix version nrs of a few functions

jrfnl:
Per https://www.php.net/manual/en/migration52.functions.php, these functions were added in PHP 5.2 and looking at the commits, that seems correct.

Ref: https://github.com/php/php-src/commit/3cd6518f392db926d439c0a277b11de7035ab959


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349880.

salathe:
These were introduced in PHP 5.2.1, not 5.2.0.  Also, don't forget to update the corresponding XMLWriter methods.

cmb69:
@salathe, thanks! Done with http://svn.php.net/viewvc?view=revision&revision=349884.

jrfnl:
> These were introduced in PHP 5.2.1, not 5.2.0. Also, don't forget to update the corresponding XMLWriter methods.

Does that also apply to the other two PHP 5.2 functions/methods ? `xmlwriter_full_end_element()` and `xmlwriter_write_raw()` ? Or were those introduced in `5.2.0` ?

salathe:
> Does that also apply to the other two PHP 5.2 functions/methods ? `xmlwriter_full_end_element()` and `xmlwriter_write_raw()` ? Or were those introduced in `5.2.0` ?

Those were introduced in 5.2.0, so they're okay.

PR #100, opened 2020-05-20
DBX/installation: extension was moved to PECL

jrfnl:
Updating: https://www.php.net/manual/en/dbx.installation.php

The extension has not been bundled with PHP since 5.1.0.

Ref: https://www.php.net/manual/en/intro.dbx.php


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349953.

PR #101, opened 2020-05-20
Frontbase/installation: extension was moved to PECL

jrfnl:
Updating: https://www.php.net/manual/en/fbsql.installation.php

The extension has not been bundled with PHP since 5.3.0.

As I couldn't find the extension on PECL anymore, I have not included a link to the PECL package.

Ref: https://www.php.net/manual/en/intro.fbsql.php

----

[Edit] I've added a second commit as the function version nrs appear incorrect as well.

#### Frontbase: fix function version nrs

The extension has not been bundled with PHP since 5.3.0, so shouldn't give the impression that the functions are available in all PHP 5 versions, as well as PHP 7.

As I couldn't find the extension on PECL anymore, I have not included "PECL/fbsql SVN" in the version nr.

### DIFF
--- a/reference/fbsql/configure.xml
+++ b/reference/fbsql/configure.xml
@@ -1,7 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<section xml:id="fbsql.installation" xmlns="http://docbook.org/ns/docbook">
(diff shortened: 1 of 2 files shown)


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349948.

PR #102, opened 2020-05-20
fix ugly bug in sample code

divinity76:
imagine if 2 handles finish in the same multi_exec(), before exec() $active is 2 and after the call $active is 0, then we read the first message, and the the message from the 2nd handle will be lost/ignored!

this has caused real bugs in the wild, see https://stackoverflow.com/questions/61920359/count-how-many-curl-multi-requests-have-been-made/61923372


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349952.

PR #103, opened 2020-05-21
filePro: extension was moved to PECL / no longer bundled with PHP as of PHP 5.2

jrfnl:
As per: https://www.php.net/manual/en/intro.filepro.php

It is unclear to me what the current version nr is of the PECL extension, but showing the functions as if they are available in PHP > 5.2 seems wrong either way.

Includes adding a mention of when the extension was moved to PECL on the "Installing/Configuring" page.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349949.

PR #104, opened 2020-05-21
Msql: extension was removed as of PHP 5.3

jrfnl:
As per: https://www.php.net/manual/en/migration53.removed-extensions.php

AFAICS the extension was *not* moved to PECL.

With that in mind, I've not used the macros, but have used similar phrasing as used in the macros.

Adjustments include:
* Adding a note about the removal on the [introduction](https://www.php.net/manual/en/intro.msql.php) page.
* Adding a note about the removal on the [installation](https://www.php.net/manual/en/msql.installation.php) page.
* Adjusting the version numbers for all functions.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349950

PR #105, opened 2020-05-21
Firebird/Interbase: extension was moved to PECL as of PHP 7.4

jrfnl:
As per: https://www.php.net/manual/en/migration74.removed-extensions.php

Adjustments include:
* Adding a note about the move on the [introduction](https://www.php.net/manual/en/intro.ibase.php) page.
* Adding a note about the move on the [installation](https://www.php.net/manual/en/ibase.installation.php) page.
* Adjusting the version numbers for all functions.

### DIFF
--- a/reference/ibase/book.xml
+++ b/reference/ibase/book.xml
@@ -30,6 +30,9 @@
   </para>
   <note>
    <para>
+    &pecl.moved-ver;7.4.0
+   </para>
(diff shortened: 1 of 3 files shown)


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349951.

PR #106, opened 2020-05-22
DB++/installation: extension was moved to PECL

jrfnl:
Based on the version nrs used for the functions, this extension has not been bundled with PHP since 4.1.0.

Ref: https://www.php.net/manual/en/ref.dbplus.php


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349962.

PR #107, opened 2020-05-22
MaxDB/installation: extension is PECL only

jrfnl:
Updating: https://www.php.net/manual/en/maxdb.installation.php

This extension is not bundled with PHP, so people need to download the PECL extension before they can use the extension.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349964.

PR #108, opened 2020-05-23
Informix/installation: extension was moved to PECL

jrfnl:
Updating: https://www.php.net/manual/en/ifx.installation.php

The Informix extension was moved to PECL as of PHP 5.2.1.

Ref: https://www.php.net/manual/en/book.ifx.php


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349961.

PR #110, opened 2020-05-24
OCI8: various tweaks to the function version numbers

jrfnl:
Various updates to the docs of the MySQLi extensions.

I've split the PR into logical commits to allow for easier reviewing and, if necessary, partial committing to SVN.

Critical review of this PR is warranted and appreciated.

### OCI8: various tweaks to the function version numbers [1]

[`oci_set_call_timeout()`](https://www.php.net/manual/en/function.oci-set-call-timout.php)

I'm questioning whether the current version numbering: "(PHP 7 >= 7.2.14, PHP 7.3.1 > 7.3 PECL OCI8 >= 2.2.0)" is correct.
This would imply that the function was added in PHP 7.2.14 and removed in PHP 7.3.1.

I believe this this should be showing that the function was added in both PHP 7.2.14 and PHP 7.3.1 and is currently still available.

### OCI8: various tweaks to the function version numbers [2]

[`oci_set_db_operation()`](https://www.php.net/manual/en/function.oci-set-db-operation.php)

I'm questioning whether the current version numbering: "(PHP 7 >= 7.2.14, PHP 7.3.1 > 7.3 PECL OCI8 >= 2.2.0)" is correct.
This would imply that the function was added in PHP 7.2.14 and removed in PHP 7.3.1.

I believe this this should be showing that the function was added in both PHP 7.2.14 and PHP 7.3.1 and is currently still available.

### OCI8: various tweaks to the function version numbers [3]

[`oci_unregister_taf_callback()`](https://www.php.net/manual/en/function.oci-unregister-taf-callback.php)

I'm questioning whether the current version numbering: "(PHP 7.0 >= 7.0.23, PHP 7 >= 7.1.9, PHP 7.2, PECL OCI8 >= 2.1.7)" is correct.
This would imply that the function was added in PHP 7.0.23 and 7.1.9 and removed in PHP 7.3.0 (not available in 7.3/7.4).

I believe this this should be showing that the function was added in both PHP 7.0.23 and 7.1.9 and is currently still available.

### OCI8: various tweaks to the function version numbers [4]

* [`ocicloselob()`](https://www.php.net/manual/en/function.ocicloselob.php)
* [`ocicollassign()`](https://www.php.net/manual/en/function.ocicollassign.php)
* [`ociwritetemporarylob()`](https://www.php.net/manual/en/function.ociwritetemporarylob.php)

Each of these functions has a warning showing that the function was deprecated in PHP 5.4.0, but according to the version numbers "(PHP 4 >= 4.0.6, PECL OCI8 1.0)", the function didn't exist in PHP 5 at all.

I'm presuming this is an oversight in the version numbering and have fixed those.
Alternatively, the deprecation warning should be removed/replaced with a removal warning.

### OCI8: various tweaks to the function version numbers [5]

In line with what I saw in various other `versions.xml` files, adding the `deprecated` attribute to various functions for which it seems to apply.

### DIFF
--- a/reference/oci8/versions.xml
+++ b/reference/oci8/versions.xml
@@ -96,9 +96,9 @@
 
  <function name="ocibindbyname" from="PHP 4, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
  <function name="ocicancel" from="PHP 4, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
- <function name="ocicloselob" from="PHP 4 &gt;= 4.0.6, PECL OCI8 1.0"/>
+ <function name="ocicloselob" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 1.0"/>
  <function name="ocicollappend" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
- <function name="ocicollassign" from="PHP 4 &gt;= 4.0.6, PECL OCI8 1.0"/>
+ <function name="ocicollassign" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 1.0"/>
  <function name="ocicollassignelem" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
  <function name="ocicollgetelem" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
  <function name="ocicollmax" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
@@ -145,7 +145,7 @@
  <function name="ocisetprefetch" from="PHP 4, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
  <function name="ocistatementtype" from="PHP 4, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
  <function name="ociwritelobtofile" from="PHP 4, PHP 5, PHP 7, PECL OCI8 &gt;= 1.0.0"/>
- <function name="ociwritetemporarylob" from="PHP 4 &gt;= 4.0.6, PECL OCI8 1.0"/>
+ <function name="ociwritetemporarylob" from="PHP 4 &gt;= 4.0.6, PHP 5, PHP 7, PECL OCI8 1.0"/>
 </versions>
 <!-- Keep this comment at the end of the file
 Local variables:
--- a/reference/oci8/versions.xml
+++ b/reference/oci8/versions.xml
@@ -86,7 +86,7 @@
  <function name="oci_set_call_timeout" from="PHP 7.2 &gt;= 7.2.14, PHP 7 &gt;= 7.3.1, PECL OCI8 &gt;= 2.2.0"/>
  <function name="oci_set_client_info" from="PHP 5 &gt;= 5.3.2, PHP 7, PECL OCI8 &gt;= 1.4.0"/>
  <function name="oci_set_client_identifier" from="PHP 5 &gt;= 5.3.2, PHP 7, PECL OCI8 &gt;= 1.4.0"/>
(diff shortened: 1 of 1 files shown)


reference/oci8/versions.xml
```diff
@@ -83,69 +83,69 @@
  <function name="oci_rollback" from="PHP 5, PHP 7, PECL OCI8 &gt;= 1.1.0"/>
  <function name="oci_server_version" from="PHP 5, PHP 7, PECL OCI8 &gt;= 1.1.0"/>
  <function name="oci_set_action" from="PHP 5 &gt;= 5.3.2, PHP 7, PECL OCI8 &gt;= 1.4.0"/>
- <function name="oci_set_call_timeout" from="PHP 7 &gt;= 7.2.14, PHP 7.3.1 &gt; 7.3 PECL OCI8 &gt;= 2.2.0"/>
+ <function name="oci_set_call_timeout" from="PHP 7.2 &gt;= 7.2.14, PHP 7 &gt;= 7.3.1 PECL OCI8 &gt;= 2.2.0"/>
```
cmb69, line 86: Nit:
```suggestion
 <function name="oci_set_call_timeout" from="PHP 7.2 &gt;= 7.2.14, PHP 7 &gt;= 7.3.1, PECL OCI8 &gt;= 2.2.0"/>
```

reference/oci8/versions.xml
```diff
@@ -83,69 +83,69 @@
...
  <function name="oci_set_client_info" from="PHP 5 &gt;= 5.3.2, PHP 7, PECL OCI8 &gt;= 1.4.0"/>
  <function name="oci_set_client_identifier" from="PHP 5 &gt;= 5.3.2, PHP 7, PECL OCI8 &gt;= 1.4.0"/>
- <function name="oci_set_db_operation" from="PHP 7 &gt;= 7.2.14, PHP 7.3.1 &gt; 7.3 PECL OCI8 &gt;= 2.2.0"/>
+ <function name="oci_set_db_operation" from="PHP 7 &gt;= 7.2.14, PHP 7 &gt;= 7.3.1 PECL OCI8 &gt;= 2.2.0"/>
```
cmb69, line 89: Same as above
```suggestion
 <function name="oci_set_db_operation" from="PHP 7 &gt;= 7.2.14, PHP 7 &gt;= 7.3.1, PECL OCI8 &gt;= 2.2.0"/>
```

reference/oci8/versions.xml
jrfnl: Fixed up in the original commit

jrfnl:
> I'm not sure how to handle the deprecation of the aliases. It seems most (all?) are only soft deprecations, and the hard deprecations are pending. Maybe the docs should distinguish between these cases?

I need a little more know-how on how the docs works to have an opinion.

From what I can see, the `deprecated="PHP 5.4.0"` in the `versions.xml` files doesn't actually _do_ much in the docs at this moment.

For a warning to be displayed, it looks like the individual function would still need to use a `&warn.deprecated.function...` macro. Similarly for this deprecation to be displayed in the "Changelog" such text would also need to be hard-coded into the individual function xml file.

The only thing I could find which I suspect is related to the `deprecated="..."` attribute is that for _global_ functions (but not class methods), they look to be categorized in a "Deprecated" list at the bottom of the function menu on the right of the manual page, though this doesn't hold true for the functions list in the main panel on the Extension "Functions" page.

So based on that, the `deprecated="..."` attributes currently have a very limited function or impact.

Either way, it is the reason why I used separate commits, so I/you can just remove/ignore the last commit if the attribute should not (yet) be added.

cmb69:
Yes, your observations are correct (any yes, there is certainly room for improvement regarding PhD). Anyhow, my point regarding the soft vs. hard deprecation is that this may also be confusing for the doc  team. If we mark the aliases as deprecated as of 5.4.0, somebody might later "fix" this when they get hard deprecated. I'm not sure whether it's worth the trouble to just document them as "using this function is discouraged as of 5.4.0" (or something like that) for now. I don't know whether there is precedent for something like that.

jrfnl:
> Anyhow, my point regarding the soft vs. hard deprecation is that this may also be confusing for the doc team. If we mark the aliases as deprecated as of 5.4.0, somebody might later "fix" this when they get hard deprecated.

I've removed that specific commit.

> I'm not sure whether it's worth the trouble to just document them as "using this function is discouraged as of 5.4.0" (or something like that) for now. I don't know whether there is precedent for something like that.

I seem to remember seeing comments like that in various places, especially with alias functions.
I suggest leaving that for now until https://github.com/php/php-src/pull/5589 is resolved.

cmb69:
Yes, we can leave that for now. :)

Applied the PR as http://svn.php.net/viewvc?view=revision&revision=349988. Thanks!

PR #111, opened 2020-05-24
PostgreSQL: add missing constants & annotate since version

jrfnl:
Updating: https://www.php.net/manual/en/pgsql.constants.php

* Add a few constants which weren't listed, but were added in PHP 5.6 / PHP 7.3.
* Annotate the `since` version for various constants.

Refs:
* https://www.php.net/manual/en/migration56.constants.php#migration56.constants.pgsql
* https://www.php.net/manual/en/migration73.constants.php#migration73.constants.pgsql

### DIFF
--- a/reference/pgsql/constants.xml
+++ b/reference/pgsql/constants.xml
@@ -87,9 +87,31 @@
     </simpara>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.pgsql-connection-auth-ok">
+   <term>
+    <constant>PGSQL_CONNECTION_AUTH_OK</constant>
+    (<type>integer</type>)
+   </term>
+   <listitem>
+    <simpara>
+      Available since PHP 5.6.0.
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=349989

cmb69:
PS: the docs *should* use "available as of PHP …" instead of "available since PHP …" (see Chronology section on http://doc.php.net/tutorial/style.php), but since that page (like many others) already uses "since", I think it's okay to stick with that wording for now.

jrfnl:
@cmb69 i remember reading that and then got confused by the prior art. Might be worth doing a doc-wide search for `since PHP` and fixing it all up in one go ?

cmb69:
I'm not sure, because mass updates are always hard for translators as well.

jrfnl:
Understood.

PR #112, opened 2020-05-24
add info about accessing typed property

tiffany-taylor:
In reference to [Bug #79620](https://bugs.php.net/bug.php?id=79620), I added information explaining that a typed property cannot be accessed before initialization, otherwise a fatal error occurs.

### DIFF
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -134,7 +134,7 @@ EOT;
   
   <para>
    As of PHP 7.4.0, property definitions can include a type declaration, with the exception of
-   the <literal>callable</literal> type.
+   the <literal>callable</literal> type. Typed properties must be initialized before accessing. If a typed property is accessed prior to initialization, a fatal error occurs.
    <example>
     <title>Example of typed properties</title>
     <programlisting role="php">
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -4,10 +4,10 @@
   <title>Properties</title>
 
   <para>
-   Class member variables are called "properties". You may also see
-   them referred to using other terms such as "attributes" or
-   "fields", but for the purposes of this reference we will use
-   "properties". They are defined by using one of the
+   Class member variables are called <emphasis>properties</emphasis>. You may also see
+   them referred to using other terms such as <emphasis>attributes</emphasis> or
+   <emphasis>fields</emphasis>, but for the purposes of this reference we will use
+   <emphasis>properties</emphasis>. They are defined by using one of the
    keywords <literal>public</literal>, <literal>protected</literal>,
    or <literal>private</literal>, optionally followed by a type declaration,
    followed by a normal variable declaration. This declaration may
@@ -100,7 +100,7 @@ EOD;
     the <link linkend="ref.classobj">Class/Object Functions</link>.
    </para>
   </note>
-
+  <sect2 xml:id="language.oop5.properties.heredoc-nowdoc">
   <para>
    As of PHP 5.3.0
    <link linkend="language.types.string.syntax.heredoc">heredocs</link> and 
@@ -131,12 +131,14 @@ EOT;
     Nowdoc and Heredoc support was added in PHP 5.3.0.
    </para>
   </note>
+  </sect2>
   
-  <para>
-   As of PHP 7.4.0, property definitions can include a type declaration. 
-   <example>
-    <title>Example of typed properties</title>
-    <programlisting role="php">
+  <sect2 xml:id="language.oop5.properties.typed-properties">
+   <para>
+    As of PHP 7.4.0, property definitions can include a type declaration. 
+    <example>
+     <title>Example of typed properties</title>
+     <programlisting role="php">
 <![CDATA[
 <?php
 
@@ -159,23 +161,23 @@ var_dump($user->name);
 
 ?>
 ]]>
-    </programlisting>
-    &example.outputs;
-    <screen>
+     </programlisting>
+     &example.outputs;
+     <screen>
 <![CDATA[
 int(1234)
 NULL
 ]]>
-    </screen>
-   </example>
-  </para>
+     </screen>
+    </example>
+   </para>
 
-  <para>
-   Typed properties must be initialized before accessing, otherwise an
-   <classname>Error</classname> is thrown.
-   <example>
-    <title>Accessing properties</title>
-    <programlisting role="php">
+   <para>
+    Typed properties must be initialized before accessing, otherwise an
+    <classname>Error</classname> is thrown.
+    <example>
+     <title>Accessing properties</title>
+     <programlisting role="php">
 <![CDATA[
 <?php
 
@@ -217,9 +219,9 @@ var_dump($circle->getName());
 var_dump($circle->getNumberOfSides());
 ?>
 ]]>
-    </programlisting>
-    &example.outputs;
-    <screen>
+     </programlisting>
+     &example.outputs;
+     <screen>
 <![CDATA[
 string(8) "triangle"
 int(3)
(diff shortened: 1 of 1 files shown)


language/oop5/properties.xml
```diff
@@ -134,7 +134,8 @@ EOT;
   
   <para>
    As of PHP 7.4.0, property definitions can include a type declaration, with the exception of
-   the <literal>callable</literal> type.
+   the <literal>callable</literal> type. Typed properties must be initialized before accessing. 
+   If a typed property is accessed prior to initialization, a fatal error occurs.
```
salathe, line 138: It's just `Error`, not fatal (unless uncaught).

Aside: It would be worth clarifying more precisely which types are/aren't allowed (see also [Function arguments: type declarations](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration)); e.g.`resource` also isn't allowed.

tiffany-taylor: Regarding error: https://3v4l.org/AsSO3 it says `Fatal error`, what is the difference?

IMSoP: There's two different meanings of "error" in that one message (unhelpfully). The "Fatal error" is that something was thrown never handled in a `catch` block; the `Uncaught Error` is the thing that was thrown. You'd more commonly see something like "Fatal error: Uncaught FileNotFotundException" or whatever.

So the documentation should point out that an `Error` is thrown, and it's up to the user to catch it if they don't want a fatal error: https://3v4l.org/4h7u9

language/oop5/properties.xml
```diff
@@ -133,8 +133,12 @@ EOT;
...
   
   <para>
-   As of PHP 7.4.0, property definitions can include a type declaration, with the exception of
-   the <literal>callable</literal> type.
+   As of PHP 7.4.0, property definitions can include a type declaration. 
+   Any <link linkend="functions.arguments.type-declaration">valid type 
+   declaration</link>, with the exception of the <literal>callable</literal> 
+   type. Typed properties must be initialized before accessing. If a typed 
+   property is accessed prior to initialization, an <link linkend="class.error">
+   Error</link> is thrown.
```
cmb69, line 141:
```suggestion
   property is accessed prior to initialization, an
   <classname>Error</classname> is thrown. 
```

language/oop5/properties.xml
```diff
@@ -133,8 +133,12 @@ EOT;
   </note>
   
   <para>
-   As of PHP 7.4.0, property definitions can include a type declaration, with the exception of
-   the <literal>callable</literal> type.
+   As of PHP 7.4.0, property definitions can include a type declaration. 
+   Any <link linkend="functions.arguments.type-declaration">valid type 
+   declaration</link>, with the exception of the <literal>callable</literal>
```
salathe, line 138: The linked page lists the valid types for *parameter type declarations*.  It would be better to not reference that list at all: a) saying "things in this list <reader has to click away>, but not these things <reader has to refer back to this page>" is convoluted for the reader, b) having a list of valid types for typed properties in the typed properties section means we don't need the "with the exception of" list.

Suggestion: add a table to this section for valid types for typed properties, similar to the one for parameter type declarations.

language/oop5/properties.xml
```diff
@@ -133,8 +133,12 @@ EOT;
   </note>
   
   <para>
-   As of PHP 7.4.0, property definitions can include a type declaration, with the exception of
-   the <literal>callable</literal> type.
+   As of PHP 7.4.0, property definitions can include a type declaration. 
+   Any <link linkend="functions.arguments.type-declaration">valid type 
+   declaration</link>, with the exception of the <literal>callable</literal> 
+   type. Typed properties must be initialized before accessing. If a typed
```
salathe, line 139: Mentioning that typed properties must be initialized before accessing, and the consequences of not doing that, should be a separate paragraph and would benefit from a separate example.  

Suggestion: keep the basic introduction and example together, then move on to details like initializing before accessing,.

language/oop5/properties.xml
salathe: 👍

language/oop5/properties.xml
tiffany-taylor: Something like this?
(not sure if this will show up correctly, giving it a try)
```
+------+-------------------------------------------+-------------------------+
| Type | Description                               | Minimum PHP Version     |
+------+-------------------------------------------+-------------------------+
| bool | The property must be a boolean value.     | PHP 7.0.0               |
+------+-------------------------------------------+-------------------------+
```

language/oop5/properties.xml
tiffany-taylor: Should I create two examples: one that shows what happens if the properties are accessed prior to initialization, and another which shows how it's correctly done? (that is, setting the property _somewhere_ before reading it)

cmb69: Maybe give one example with two properties (one used correctly, the other used incorrectly)?

tiffany-taylor:
Please review new changes.

language/oop5/properties.xml
```diff
@@ -144,33 +143,190 @@ EOT;
...
+int(3)
+string(6) "circle"
+
+Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be accessed before initialization
+]]>
+    </screen>
+   </example>
+  </para>
+
+  <title>Valid property types</title>
```
cmb69, line 234: I think this `<title>` is invalid here. Maybe use a [`<table>`](https://tdg.docbook.org/tdg/5.0/cals.table.html) instead?

tiffany-taylor: Could that be a reason why the other valid types table is surrounded by a different <sect> tag?

cmb69: Yes, that might be the reason.

language/oop5/properties.xml
```diff
@@ -217,115 +219,118 @@ var_dump($circle->getName());
...
-      <entry>?type</entry>
-      <entry>
-       The property may optionally be nullable.
-      </entry>
-      <entry>PHP 7.1.0</entry>
-     </row>
-    </tbody>
-   </tgroup>
-  </informaltable>
+  <sect3 xml:id="language.oop5.properties.typed-properties.valid-types">
```
tiffany-taylor, line 236: I'm not sure if `typed-properties.valid-types` is correct for the id.

cmb69: Looks good to me. :)

language/oop5/properties.xml
```diff
@@ -131,45 +131,206 @@ EOT;
...
+        <entry>
+         The property must be either an <type>array</type> or an &instanceof; 
+         <classname>Traversable</classname>.
+        </entry>
+        <entry>PHP 7.1.0</entry>
+       </row>
+       <row>
+        <entry><literal>self</literal></entry>
+        <entry>
+         The property must be an &instanceof; the same class as the same class
```
salathe, line 301: The repetition of "the same class" makes this read very awkwardly.
Suggestion: something like `The property must be an &instanceof; the same class that the property is defined on.`

language/oop5/properties.xml
```diff
@@ -131,45 +131,206 @@ EOT;
...
+         The property must be an &instanceof; the same class as the same class 
+         in which it is defined.
+        </entry>
+        <entry>PHP 5.0.0</entry>
+       </row>
+       <row>
+        <entry><literal>parent</literal></entry>
+        <entry>
+         The property must be a parent of the class in which it is defined.
+         Optionally, the class may be a sub-class of the parent instance.
```
salathe, line 310: This is not optional.  It must be an instance of the parent or a descendant of that parent.

Suggestion: keep the `must be an &instanceof; ...` pattern used elsewhere. E.g. `The property must be an &instanceof; the parent class of the class that the property is defined on.`

language/oop5/properties.xml
```diff
@@ -131,45 +131,206 @@ EOT;
...
+        <entry>
+         The property must be an &instanceof; the given class or interface
+         name.
+        </entry>
+        <entry>PHP 5.0.0</entry>
+       </row>
+       <row>
+        <entry>?type</entry>
+        <entry>
+         The property may optionally be nullable.
```
salathe, line 325: It's not optionally nullable, it *is* nullable. 

Suggestion: `The property must be the specified type, or &null;.`

language/oop5/properties.xml
```diff
@@ -100,7 +100,7 @@ EOD;
     the <link linkend="ref.classobj">Class/Object Functions</link>.
    </para>
   </note>
-
+  <sect2 xml:id="language.oop5.properties.heredoc-nowdoc">
```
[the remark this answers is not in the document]
tiffany-taylor: I tried correcting the indentation, but I'm unsure if it's fixed.

salathe:
Thank you for being very patient with the suggested changes!  When going to merge this, I noticed the version numbers in the table were wrong (copied from the function arguments page), so I updated those and fixed a couple of whitespace issues.

Merged now as [r350015](http://svn.php.net/viewvc/?view=revision&revision=350015) / a4365511

Thanks again!

PR #118, opened 2020-06-09
Add note on behavior of ArrayIterator::offsetUnset when used on current iteration index

alexdowad:
Bug report #55157 (which is not really a "bug") is based on a misunderstanding of the behavior
of `offsetUnset` as regards iteration. Although not all users will read the documentation, it
is better to have documentation addressing this issue, since it is obviously a source of
confusion.

### DIFF
--- a/reference/spl/arrayiterator/offsetunset.xml
+++ b/reference/spl/arrayiterator/offsetunset.xml
@@ -16,8 +16,15 @@
   <para>
    Unsets a value for an offset.
   </para>
-
-  &warn.undocumented.func;
+  <para>
+   If iteration is in progress, and <methodname>ArrayIterator::offsetUnset</methodname> is used to
+   unset the current index of iteration, the iteration position will be advanced to the next index.
+   Since the iteration position is also advanced at the end of a
+   <link linkend="control-structures.foreach"><literal>foreach</literal></link> loop body, use of
+   <methodname>ArrayIterator::offsetUnset</methodname> inside a
+   <link linkend="control-structures.foreach"><literal>foreach</literal></link> loop may result in
+   indices being skipped.
+  </para>
 
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/spl/arrayiterator/offsetunset.xml
```diff
@@ -16,8 +16,9 @@
   <para>
    Unsets a value for an offset.
   </para>
-
-  &warn.undocumented.func;
+  <para>
+    If iteration is in progress, and <function>offsetUnset</function> is used to unset the current index of iteration, the iteration position will be advanced to the next index. Since the iteration position is also advanced at the end of a <literal>foreach</literal> loop body, use of <function>offsetUnset</function> inside a <literal>foreach</literal> loop may result in indices being skipped.
```
salathe, line 20: Minor suggestions:

- `<function>offsetUnset</function>` should be `<methodname>ArrayIterator::offsetUnset</methodname>`
- `<literal>foreach</literal>` should be `<link linkend="control-structures.foreach"><literal>foreach</literal></link>`
- Indentation for this line should be 3 spaces but is 4 spaces.
- Several narrow lines of text (usually wrapping around 100 characters or so) is preferred over one very long line.

alexdowad: Wow, thank you! Working on these...

alexdowad: Just pushing the fixes... next time I will take note of these points from the start...

reference/spl/arrayiterator/offsetunset.xml
```diff
@@ -16,8 +16,15 @@
   <para>
    Unsets a value for an offset.
   </para>
-
-  &warn.undocumented.func;
+  <para>
+   If iteration is in progress, and <methodname>ArrayIterator::offsetUnset</methodname> is used to
+   unset the current index of iteration, the iteration position will be advanced to the next index.
+   Since the iteration position is also advanced at the end of a
+   <link linkend="control-structures.foreach"><literal>foreach</literal></link> loop body, use of
```
cmb69, line 23: You could also use a `&foreach;` entity reference instead of the `<link>` element:
```suggestion
   &foreach; loop body, use of
```

salathe:
Applied as [r350025](http://svn.php.net/viewvc/?view=revision&revision=350025) / bbe604e. Thanks!

PR #119, opened 2020-06-10
fix Phar::offsetGet return type

gharlan:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350026.

PR #120, opened 2020-06-10
Clarify handling of 3rd parameter for assert callbacks

alexdowad:
The handling of the 3rd parameter for assert callbacks set via `assert_options`
changed with the implementation of the expectations RFC, and changed again in
PHP 8. Update the documentation accordingly.

This one seems to be a cause for confusion; witness Bug [#79602](https://bugs.php.net/bug.php?id=79602).

### DIFF
--- a/reference/info/functions/assert-options.xml
+++ b/reference/info/functions/assert-options.xml
@@ -124,8 +124,7 @@
          <term><parameter>assertion</parameter></term>
          <listitem>
           <simpara>
-           The assertion that has been passed to <function>assert</function>,
-           converted to a string.
+           Prior to PHP 8, the assertion which has been passed to <function>assert</function>, but only when the assertion is given as a string. (If the assertion is a boolean condition, this parameter will be an empty string.) From PHP 8, this parameter is always &null;.
           </simpara>
          </listitem>
         </varlistentry>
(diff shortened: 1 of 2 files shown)


alexdowad:
Pinging maintainers.

cmb69:
Thanks for the PR (and sorry for the late reply)!

We do not want to document PHP 8 specific behavior yet. If you like feel free to provide a separate PR, which can be labelled with "PHP 8.0", and applied later.

We do still want to document PHP 5 behavior, though. It seems to me that the behavioral change between PHP 5 and 7 can be clarified better (I may be wrong here). Also note that we prefer the wording "as of PHP x.y.z" and "prior to PHP x.y.z" to be as detailed as possible.

cmb69:
As time passed, we moved on: PHP 5 specific docs are no longer desired; PHP 8 specific docs are good to go.

afilina:
@alexdowad Here is an example of how to remove references to pre-PHP 7 behavior: #277

afilina:
Pinging @alexdowad.

alexdowad:
@afilina Thank you for the follow-up, and for linking to PR #277. I looked through it and am not very sure what the salient point from that PR is.

cmb69:
Yes, you did @afilina. However, I'm not sure whether "up to PHP 7" means "prior to PHP 7.0.0" or "prior to PHP 8.0.0" here.

afilina:
@cmb69 I'm not sure. I think this should be clarified by @alexdowad.
- If "prior to PHP 7.0.0", then that information should not be included, as we no longer document behavior prior to PHP 7.
- If "prior to PHP 8.0.0", then it should be stated as such.

afilina:
Pinging @alexdowad to address the last blocking point.

alexdowad:
@afilina Please see if this looks better now.

PR #122, opened 2020-06-23
SplFileObject: fix links in the seealso section

villfa:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350079.

PR #123, opened 2020-07-11
Fully document behavior of negative arguments to mb_strcut

alexdowad:
When reading the source code, I noticed there is some special handling of
negative 'start' and 'length' arguments which is not mentioned in the
manual.


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350143.

PR #124, opened 2020-07-11
Clarify what the 'width' in mb_strimwidth means; it takes East Asian width into account

alexdowad:
If a link to an external resource can be included, this might be a good one:
http://www.unicode.org/reports/tr11/tr11-38.html

### DIFF
--- a/reference/mbstring/functions/mb-strimwidth.xml
+++ b/reference/mbstring/functions/mb-strimwidth.xml
@@ -17,7 +17,7 @@
    <methodparam choice="opt"><type>string</type><parameter>encoding</parameter><initializer>mb_internal_encoding()</initializer></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks! It certainly is reasonable to clarify this for `mb_strimwidth()`, but I think we should better match with the description of `mb_strwidth()`. Not sure whether we should duplicate the explantion, or have it only for `mb_strwidth()` and link to that from `mb_strimwidth()`.

Adding a link to TR11 makes also sense; you could write that as `<link xlink:href="&url.unicode;reports/tr11/">` (might be reasonable to define an own entity in doc-base/entities/global.ent, but that is not possible for a PR yet).

Also, it might be good to add an example (either for `mb_strwidth()` or `mb_strimwidth()`, or both?)  showing halfwidth and fullwidth character behavior.

afilina:
Pinging @alexdowad in case this fell through the cracks.

PR #126, opened 2020-07-27
Add exception_ignore_args

martinstuecklschwaiger:
It was added to PHP 7.4:
https://www.php.net/manual/en/migration74.other-changes.php#migration74.other-changes.ini

### DIFF
--- a/appendices/ini.core.xml
+++ b/appendices/ini.core.xml
@@ -100,7 +100,7 @@
         <entry>Available since PHP 7.1.0.</entry>
        </row>
        <row>
-        <entry><link linkend="ini.zend.exception_ignore_args">zend.exception_ignore_args</link></entry>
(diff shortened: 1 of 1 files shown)


appendices/ini.core.xml
```diff
@@ -447,6 +453,18 @@
       </listitem>
      </varlistentry>
      
+     <varlistentry xml:id="ini.zend.exception_ignore_args">
```
cmb69, line 456: For some reason, underscores are replaced by dashes for the IDs. I suggest to stick with that.

martinstuecklschwaiger: I completely missed that. I looked at `zend.multibyte` which didn't have any underscores or dashes.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=350231. Thanks!

PR #127, opened 2020-07-29
Document configuring opcache preloading.

Crell:
Yes, the compile_file vs include logic really is that bizarre.  We may want to consider fixing the _once behavior so preloading shares an inclusion cache.

### DIFF
--- a/reference/opcache/preload.xml
+++ b/reference/opcache/preload.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<chapter xml:id="features.preloading" xmlns="http://docbook.org/ns/docbook">
-    <title>Preloading</title>
+<chapter xml:id="opcache.preloading" xmlns="http://docbook.org/ns/docbook">
+<title>Preloading</title>
 
     <para>
         As of PHP 7.4, PHP can be configured to preload scripts into the opcache when the engine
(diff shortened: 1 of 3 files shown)


cmb69:
Thanks for the PR! I'm generally in favor on having a dedicated page about preloading, but I think it should be part of the OPcache "book".

Crell:
Page moved, indentation adjusted.

reference/opcache/preload.xml
```diff
@@ -0,0 +1,119 @@
...
+ <programlisting role="ini">
+ <![CDATA[
+opcache.preload=preload.php
+]]>
+ </programlisting>
+
+ <para>
+  <filename>preload.php</filename> is an arbitrary file that will run once at server startup
+  (PHP-FPM, mod_php, etc.) and load code into persistent memory.  If PHP will be run as
+  root (not recommended), the <link linkend="opcache.configuration.php#ini.opcache.preload_user">opcache.preload_user</link>
```
Crell, line 45: Travis doesn't like this line.  What's the right syntax for an anchor linkend?

cmb69: Use `linkend="ini.opcache.preload-user"`.

cmb69:
Thanks @Crell!

I've applied the patch (with a few tweaks) as http://svn.php.net/viewvc?view=revision&revision=350245.

PR #128, opened 2020-08-03
The 2nd argument to mb_parse_str is not optional

alexdowad:
Using mb_parse_str without the 2nd argument was deprecated in PHP 7.2. Nikita Popov made it a required argument in ff780feca4f1b2dd5c1b7cf692781070d49df7cb.

### DIFF
--- a/reference/mbstring/functions/mb-parse-str.xml
+++ b/reference/mbstring/functions/mb-parse-str.xml
@@ -5,13 +5,13 @@
   <refname>mb_parse_str</refname>
   <refpurpose>Parse GET/POST/COOKIE data and set global variable</refpurpose>
  </refnamediv>
-   
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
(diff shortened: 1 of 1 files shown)


cmb69:
Ah, good catch. Thanks! There should be changelog entries, though; one for the PHP 7.2.0 deprecation, and one for the parameter being mandatory as of PHP 8.0.0. Could you please add that? (see https://github.com/php/doc-base/blob/master/RFC/skeletons/function.xml for the format and position of the changelog)

alexdowad:
> Ah, good catch. Thanks! There should be changelog entries, though; one for the PHP 7.2.0 deprecation, and one for the parameter being mandatory as of PHP 8.0.0. Could you please add that? (see https://github.com/php/doc-base/blob/master/RFC/skeletons/function.xml for the format and position of the changelog)

OK. Have a look now.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350280 (I changed the order of the changelog entries; newer entries should be at the top).

PR #129, opened 2020-08-05
Add editorconfig file.

Crell:
If these are the expected formatting rules, let's let the tooling do the work for us.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,4 +9,3 @@ indent_size = 1
 indent_style = space
 end_of_line = lf
 insert_final_newline = true
-trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
 ; This file is for unifying the coding style for different editors and IDEs.
-; More information at http://editorconfig.org
+; More information at https://editorconfig.org
 
 root = true
 
@@ -10,6 +10,3 @@ indent_style = space
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
-
-[Makefile]
-indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://editorconfig.org
+
+root = true
+
+[*.xml]
+charset = utf-8
+indent_size = 1
+indent_style = space
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[Makefile]
+indent_style = tab


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+
+root = true
+
+[*.xml]
+charset = utf-8
+indent_size = 1
+indent_style = space
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
cmb69, line 12: I don't think this (trim_trailing_whitespace) would be appropriate; I'd rather prefer trailing whitespace removal in separate commits (if really necessary at all).

carusogabriel: Let's generate that PR, similar to https://github.com/php/php-src/pull/2995

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+
+[Makefile]
```
salathe, line 14: We don't have `Makefile` in this repo.

.editorconfig
salathe, line 12: While, generally, trailing whitespace is something we do not want; before enabling this, are you sure it won't have undesirable consequences?  I'm thinking particularly of things like example code and their output screens with significant whitespace (if such a thing exists).

.editorconfig
```diff
@@ -0,0 +1,15 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://editorconfig.org
```
salathe, line 2: Suggestion: `https://`

Crell:
Related, #131 is what stripping all trailing whitespace would look like if we did it in one go.  It's a lot. :smile: So far in my visual spot checking I don't see anything that it breaks.

Whether we trim whitespace all in one go or allow it to happen a little at a time over time, I don't much care, but IMO we should move to that.  It's very standard across nearly every OSS project I know.

salathe:
As #131 shows, many, many files in the EN tree would be affected by the `trim_trailing_whitespace = true` option.

WS changes should be committed separately to content changes, if they're committed at all, and should be very clearly labelled as WS changes in the commit log.  Please don't introduce anything that makes that more difficult than it already is.

Again, my suggestion is to drop that specific option.

salathe:
> So far in my visual spot checking I don't see anything that it breaks.

Please pay particular attention to example code and associated output screens.  When we say "The above example will output:", please make sure that is as true after the whitespace change as it was before.

KalleZ:
I personally prefer it without the option available. This is a lot of changes + the Git stuff for translators to do in one go, we can always add it in batches later and it will make reviewing the changes actually possible. I find this especially important since you are new to the workflow of docs

Crell:
Well, there seems to be little interest in addressing trailing whitespace right now one way or another, so for the time being here the editorconfig file without the whitespace trimming directive.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350335,

PR #130, opened 2020-08-08
Updated hash_algos doc for PHP 7.4

GrahamCampbell:
Support for crc32c was added in PHP 7.4: https://3v4l.org/NCnDY.


cmb69:
Thanks for the PR! Applied as http://svn.php.net/viewvc?view=revision&revision=350291.

GrahamCampbell:
🍻

PR #131, opened 2020-08-08
Strip trailing whitespace.

Crell:
Per #129, this is the result of running the following command over the whole repository:

```
find . -name "*.xml"  -type f -print0 | xargs -0 sed -i 's/[ \t]*$//'
```

It's mainly to see if trailing-whitespace-trimming would break anything, although we certainly can merge it as is if we wish to.


Girgias:
> we certainly can merge it as is if we wish to.

Hard disagree: this is is going to mark all files as out-dated for translations so until there is a mechanisms to make translations ignore a commit this shouldn't land as is.

Crell:
If all it does is validate that #129 wouldn't break anything to add the whitespace trimming rule for IDEs, I'm cool with that.

Whenever we remove extraneous trailing whitespace it will cause some speed bumps.  Whether that happens all at once or little by little I don't much care, but we really ought to.

Crell:
Forced pushed an updated version that skips image files; oopsies.

cmb69:
Ugh, that's way more trailing whitespace than I'd had expected. I don't think we can do that wrt. translations like @Girgias already mentioned.

salathe:
Closing. This is not something we want to do at this time.

PR #132, opened 2020-08-12
Fix Locale#getKeywords style inconsistencies

BafS:


cmb69:
Thanks for the PR! I agree that the coding style of that example is uncommon, but I'm not convinced that changing it useful, unless the docs adopt some particular coding style.

BafS:
But it should, at least, be consistent in the same file right ?
Otherwise I agree but it's a much bigger job, what about cleaning all the file to use PSR-2 ? If yes I can open another PR

salathe:
> Thanks for the PR! I agree that the coding style of that example is uncommon, but I'm not convinced that changing it useful, unless the docs adopt some particular coding style.

The current expected coding standard for examples is PEAR's coding standards.

References:
* http://doc.php.net/tutorial/style.php
* https://pear.php.net/manual/en/standards.php


> But it should, at least, be consistent in the same file right ?

Agreed.

> what about cleaning all the file to use PSR-2 ? If yes I can open another PR

I don't think that the cost of changing the coding standards for examples is worth it at this point.

BafS:
Thanks for the update.

> I don't think that the cost of changing the coding standards for examples is worth it at this point.

I'm not sure if that's complicated with a script, if you guys agree I could do something, I never used the PEAR's standards but it seems pretty close to PSR-2 and phpcbf could do most of the job.

Also the documentation is read by thousand of devs each day, I think it's worth having a good quality.

Girgias:
> 
> 
> Thanks for the update.
> 
> > I don't think that the cost of changing the coding standards for examples is worth it at this point.
> 
> I'm not sure if that's complicated with a script, if you guys agree I could do something, I never used the PEAR's standards but it seems pretty close to PSR-2 and phpcbf could do most of the job.
> 
> Also the documentation is read by thousand of devs each day, I think it's worth having a good quality.

The problem is the added burden given to translators, until translations have a mechanisms to ignore a certain commit which doesn't impact on the "is up to date" status of the page, this is the same reason why #131 can't land until *at least* the translations are moved from SVN to git.

cmb69:
> The current expected coding standard for examples is PEAR's coding standards.

Ah, thanks!

salathe:
@php-pulls merge

PR #133, opened 2020-08-14
Number of bits is an integer, not a GMP number

BenMorel:
Source: https://github.com/php/php-src/blob/e208cb23c6bda00737a9bfd227b82beaba350fa7/ext/gmp/gmp.c#L1738


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350328.

PR #134, opened 2020-08-17
array_map() tweaks

Crell:
These are the only changes I could see worth making based on comments.  I've purged a bunch of outdated comments, and can likely remove another 1 or 2 with this merged.

### DIFF
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -232,7 +232,7 @@ Array
   </para>
   <para>
    If <parameter>callback</parameter> expects more parameters than the <function>array_map</function>
-   invocation provides, an <function>ArgumentCountError</function> will be thrown.
+   invocation provides, an <classname>ArgumentCountError</classname> will be thrown.
   </para>
   <para>
    <example>
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -230,10 +230,6 @@ Array
    as the name of the callback function
   </para>
   <para>
-   If <parameter>callback</parameter> expects more parameters than the <function>array_map</function>
-   invocation provides, an <classname>ArgumentCountError</classname> will be thrown.
-  </para>
-  <para>
    <example>
     <title>Performing a zip operation of arrays</title>
     <programlisting role="php">
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -34,7 +34,7 @@
      <term><parameter>callback</parameter></term>
      <listitem>
       <para>
-       Callback function to run for each element in each array.
+       Callback callable to run for each element in each array.
       </para>
       <para>
        &null; can be passed as a value to <parameter>callback</parameter>
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -48,7 +48,8 @@
      <term><parameter>array1</parameter></term>
      <listitem>
       <para>
-       An array to run through the <parameter>callback</parameter> function.
+       An array to run through the <parameter>callback</parameter> function. Only actual
+       arrays are supported, not iterables.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -17,7 +17,7 @@
   <para>
    <function>array_map</function> returns an &array; containing
    the results of applying the <parameter>callback</parameter>
-   function to the corresponding index of <parameter>array1</parameter>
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-map.xml
```diff
@@ -34,7 +34,7 @@
      <term><parameter>callback</parameter></term>
      <listitem>
       <para>
-       Callback function to run for each element in each array.
+       Callback callable to run for each element in each array.
```
chx, line 37: Just "Callback to run for each element in each array." is enough. https://www.php.net/manual/en/language.types.callable.php clearly says Callback and callable are synonyms however "Callables to run" sounds terrible.

reference/array/functions/array-map.xml
```diff
@@ -229,6 +230,10 @@ Array
    which can be easily performed by using &null;
    as the name of the callback function
   </para>
+  <para>
+   If the callback function expects more parameters than the <function>array_map</function>
```
chx, line 234: Just "callback" is enough , you yourself made it clear it can be more than just functions.

reference/array/functions/array-map.xml
```diff
@@ -229,6 +230,10 @@ Array
...
+   invocation provides, an <function>ArgumentCountError</function> will be thrown.
```
chx, line 235: Please see the similar warning in `reference/array/functions/array-walk.xml` and copy it:

1. `ArgumentCountError` is not a function, the proper notation would be `<classname>ArgumentCountError</classname>`
2. This was added in PHP 7.1.0.

reference/array/functions/array-map.xml
```diff
@@ -48,7 +48,8 @@
      <term><parameter>array1</parameter></term>
      <listitem>
       <para>
-       An array to run through the <parameter>callback</parameter> function.
+       An array to run through the <parameter>callback</parameter> function. Only actual
+       arrays are supported, not iterables.
```
salathe, line 52: > Only actual arrays are supported, not iterables.

This is redundant. We already make it abundantly clear that the parameter must be an array.

reference/array/functions/array-map.xml
```diff
@@ -229,6 +230,10 @@ Array
    which can be easily performed by using &null;
    as the name of the callback function
   </para>
+  <para>
+   If the callback function expects more parameters than the <function>array_map</function>
+   invocation provides, an <function>ArgumentCountError</function> will be thrown.
+  </para>
```
salathe, line 236: This paragraph is badly positioned.  It comes between a sentence introducing using `null` as the callback and the associated example of doing that.

reference/array/functions/array-map.xml
salathe: This is often the case but not always.  For user-defined functions, I believe it always throws. For internal functions, I believe it mostly (always?) doesn't throw but rather issues an E_WARNING for PHP 7 >= 7.1.0, and throws as of PHP 8.0.0.

Note: this behaviour is not special for array_walk() callbacks and applies to function calls generally.  I'm not sure how useful it is to describe it on this page.

reference/array/functions/array-map.xml
Crell: I would agree, but there were a number of notes calling that out so apparently lots of people don't read.

reference/array/functions/array-map.xml
Crell: Should "callback" link to that page?  (There were a whole lot of comments about format different callable varieties, which is why I tried to clarify it.)

cmb69: Just use `<type>callable</type>` (that creates the link automatically). We could also add an entity definition in https://github.com/php/doc-base/blob/master/entities/global.ent#L659 for that.

reference/array/functions/array-map.xml
Crell: Oh delightful.

I mentioned it here because it was called out in the comments.  Based on @salathe's comments I'm unclear what the best way to proceed is.  (Other than correcting the element name, which I'll do momentarily.)

salathe: My suggestion would be to not add this paragraph.

reference/array/functions/array-map.xml
salathe: I don't think that justifies this change.

cmb69: Especially, if they don't read anyway, it makes no sense to write more. :)

salathe (approved):

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350490.

PR #135, opened 2020-08-17
array_filter() tweaks

Crell:
A few small tweaks based on the comment thread.  Nothing major, but these were the bits I felt were worth promoting.

### DIFF
--- a/reference/array/functions/array-filter.xml
+++ b/reference/array/functions/array-filter.xml
@@ -19,11 +19,11 @@
    passing them to the <parameter>callback</parameter> function.
    If the <parameter>callback</parameter> function returns &true;, the
    current value from <parameter>array</parameter> is returned into
-   the result &array;. Array keys are preserved.</para>
-   <para>
-    The resulting array therefore may have gaps in its key sequence. If numeric
-    keys are relevant, use <function>array_values</function> on the result to
-    reindex them.
+   the result &array;.
+  </para>
+  <para>
+   Array keys are preserved, and may result in gaps if the <parameter>array</parameter> was indexed.
+   The result &array; can be reindexed using the <function>array_values</function> function.
   </para>
 </refsect1>
 
--- a/reference/array/functions/array-filter.xml
+++ b/reference/array/functions/array-filter.xml
@@ -48,10 +48,9 @@
        The callback function to use
       </para>
       <para>
-       If no <parameter>callback</parameter> is supplied, all entries of
-       <parameter>array</parameter> equal to &false; (including 0, empty string, and similar, see
-       <link linkend="language.types.boolean.casting">converting to
-        boolean</link>) will be removed.
+       If no <parameter>callback</parameter> is supplied, all empty entries of
+       <parameter>array</parameter> will be removed. See <function>empty</function>
+       for how PHP defines empty in this case.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/array/functions/array-filter.xml
+++ b/reference/array/functions/array-filter.xml
@@ -19,7 +19,11 @@
    passing them to the <parameter>callback</parameter> function.
    If the <parameter>callback</parameter> function returns &true;, the
    current value from <parameter>array</parameter> is returned into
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-filter.xml
```diff
@@ -45,7 +49,7 @@
       </para>
       <para>
        If no <parameter>callback</parameter> is supplied, all entries of
-       <parameter>array</parameter> equal to &false; (see
+       <parameter>array</parameter> equal to &false; (including 0, empty string, and similar, see
```
chx, line 52: We have an excellent `empty` page with all the examples, let's link to it instead of inlining. Something like "all empty entries of <parameter>array</parameter> will be removed. See <function>empty</function> for more."

reference/array/functions/array-filter.xml
```diff
@@ -19,7 +19,11 @@
    passing them to the <parameter>callback</parameter> function.
    If the <parameter>callback</parameter> function returns &true;, the
    current value from <parameter>array</parameter> is returned into
-   the result &array;. Array keys are preserved.
+   the result &array;. Array keys are preserved.</para>
+   <para>
```
salathe, line 23: Please fix the indentation of the `</para>` tag.

reference/array/functions/array-filter.xml
```diff
@@ -19,7 +19,11 @@
...
+    The resulting array therefore may have gaps in its key sequence. If numeric
+    keys are relevant, use <function>array_values</function> on the result to
+    reindex them.
```
salathe, line 26: The wording here is not great. Suggestion: move the `Array keys are preserved` into this new paragraph, and reword as something along the lines of `Array keys are preserved, and may result in gaps if the <parameter>array</parameter> was indexed. The result &array; can be reindexed using the <function>array_values</function> function.`

reference/array/functions/array-filter.xml
salathe: :+1:

Crell:
All adjustments made.

reference/array/functions/array-diff.xml
```diff
@@ -91,18 +91,52 @@ Array
     </screen>
    </example>
   </para>
+
+  <para>
+   <example>
+    <title><function>array_diff</function> example with non-matching types</title>
+    <para>
+     Two elements are considered equal if and only if
+     <literal>(string) $elem1 === (string) $elem2</literal>. That is,
```
cmb69, line 100: Maybe:
```suggestion
     <code>(string) $elem1 === (string) $elem2</code>. That is,
```

cmb69 (approved):
LGTM. Thanks!

reference/array/functions/array-diff.xml
```diff
@@ -91,18 +91,52 @@ Array
     </screen>
    </example>
   </para>
+
```
salathe, line 94: Is changing array-diff.xml supposed to be in this PR, or was that accidental, given the change is also in #136?

reference/array/functions/array-filter.xml
```diff
@@ -19,7 +19,10 @@
    passing them to the <parameter>callback</parameter> function.
    If the <parameter>callback</parameter> function returns &true;, the
    current value from <parameter>array</parameter> is returned into
-   the result &array;. Array keys are preserved.
+   the result &array;.</para>
```
salathe, line 22: Nit: incorrect placement of `</para>`. It should be on its own line like all of the others in this file.

reference/array/functions/array-diff.xml
Crell: Accidental. Will remove.

salathe: The array-diff.xml change is still present.

Crell: Oh blargh, I didn't see it was the full change.  Rebased the branch to remove that entirely and update against master.

salathe (approved):

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350489.

PR #136, opened 2020-08-18
array_diff() cleanup

Crell:
A little bit more intensive cleanup on this one, but it let me kill a whole bunch of comments.

### DIFF
--- a/reference/array/functions/array-diff.xml
+++ b/reference/array/functions/array-diff.xml
@@ -133,7 +133,7 @@ $result = array_diff($source, $filter);
 ]]>
     </programlisting>
     <para>
-     If you want to use an alternate comparison function, see <function><link linkend="function.array-udiff">array_udiff</link></function>.
+     To use an alternate comparison function, see <function>array_udiff</function>.
     </para>
    </example>
   </para>
--- a/reference/array/functions/array-diff.xml
+++ b/reference/array/functions/array-diff.xml
@@ -131,6 +131,9 @@ $result = array_diff($source, $filter);
 ?>
 ]]>
     </programlisting>
+    <para>
+     If you want to use an alternate comparison function, see <function><link linkend="function.array-udiff">array_udiff</link></function>.
+    </para>
    </example>
   </para>
  </refsect1>
--- a/reference/array/functions/array-diff.xml
+++ b/reference/array/functions/array-diff.xml
@@ -100,7 +100,7 @@ Array
      Two elements are considered equal if and only if
      <literal>(string) $elem1 === (string) $elem2</literal>. That is,
      when the string representation is the same. If elements cannot be
-     converted to a string a type-appropriate Error or Notice will be thrown.
+     converted to a &string; a type-appropriate Error or Notice will be thrown.
     </para>
     <programlisting role="php">
      <![CDATA[
--- a/reference/array/functions/array-diff.xml
+++ b/reference/array/functions/array-diff.xml
@@ -99,8 +99,7 @@ Array
     <para>
      Two elements are considered equal if and only if
      <literal>(string) $elem1 === (string) $elem2</literal>. That is,
-     when the string representation is the same. If elements cannot be
-     converted to a &string; a type-appropriate Error or Notice will be thrown.
+     when the <link linkend="language.types.string.casting">string representation</link> is the same.
     </para>
     <programlisting role="php">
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-diff.xml
```diff
@@ -91,23 +92,60 @@ Array
...
+    <title><function>array_diff</function> example with non-matching types</title>
+    <para>
+     Two elements are considered equal if and only if
+     <literal>(string) $elem1 === (string) $elem2</literal>. That is,
+     when the string representation is the same. If elements cannot be
+     converted to a string a type-appropriate Error or Notice will be thrown.
+    </para>
+    <programlisting role="php">
+     <![CDATA[
+<?php
```
chx, line 107: I doubt we need to inline every time a string castable object is mentioned, we could just mention `See <link linkend="language.types.string.casting">Converting to string</link>` for more.

reference/array/functions/array-diff.xml
```diff
@@ -10,13 +10,13 @@
   &reftitle.description;
   <methodsynopsis>
    <type>array</type><methodname>array_diff</methodname>
-   <methodparam><type>array</type><parameter>array1</parameter></methodparam>
-   <methodparam><type>array</type><parameter>array2</parameter></methodparam>
+   <methodparam><type>array</type><parameter>source</parameter></methodparam>
+   <methodparam><type>array</type><parameter>filter1</parameter></methodparam>
```
chx, line 14: Someone much much higher than me will need to approve of this rename. My gut feeling is if we want to do this we need to go over every diff.

reference/array/functions/array-diff.xml
salathe: > Two elements are considered equal ... when the string representation is the same.

This  is used on several pages.  Rather than copy/paste, I suggest moving that to a language snippet entity.  

> If elements cannot be converted to a string a type-appropriate Error or Notice will be thrown.

Use `&string;` instead of the bare word; the entity will link to the page for the [string](https://www.php.net/manual/en/language.types.string.php) type.  

What does "a type-appropriate Error or Notice will be thrown" mean?

reference/array/functions/array-diff.xml
```diff
@@ -91,23 +92,60 @@ Array
...
+$filter = [new S('b'), new S('c'), new S('d')];
+
+$result = array_diff($source, $filter);
+
+// $result now contains one instance of S('a');
+?>
+]]>
+    </programlisting>
+    <para>
+     If you want to use an alternate comparison function, see <function><link linkend="function.array-udiff">array_udiff</link></function>.
```
salathe, line 136: Please do not use "you". See http://doc.php.net/tutorial/style.php

Using `<function>array_udiff</function>` is sufficient, the `<link>` element is not needed.

reference/array/functions/array-diff.xml
Crell: ... Ah crap, good point.  I made that a separate commit so we can easily back that out.  Leaving for now but I can rebase it out if we decide against it.

reference/array/functions/array-diff.xml
Crell: "Type appropriate error or notice" is me not being sure of the full scope.  With an array, you get "PHP Notice:  Array to string conversion".  I didn't test to see what an object or resource or whatever would give you.  I'm not sure how we want to word that.

reference/array/functions/array-diff.xml
cmb69: For now, the old names [match the implementation](https://github.com/php/php-src/blob/4514afc1875bedf4dc07cb457f8ef5c986a7ca55/ext/standard/basic_functions.stub.php#L203); unless that changes, we should not rename in the docs.

reference/array/functions/array-diff.xml
salathe: >  I'm not sure how we want to word that.

On the `array_diff()` page, I wouldn't worry about it.

Instead, I would suggest linking to [language.types.string.casting](https://www.php.net/manual/en/language.types.string.php#language.types.string.casting).  If we were to document possible errors or side effects resulting from casting values to strings, it would be there.

salathe (approved):

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350488.

PR #137, opened 2020-08-18
ucwords() cleanup

Crell:
About half the comments I removed were variations on working around the lack of delimiters, especially for complex or non-English names.  So make those examples more robust.

### DIFF
--- a/reference/strings/functions/ucwords.xml
+++ b/reference/strings/functions/ucwords.xml
@@ -123,7 +123,7 @@ $baz = ucwords($foo, "|");        // Hello|World!
     <programlisting role="php">
      <![CDATA[
 <?php
-$foo = 'mike o'hara';
+$foo = "mike o'hara";
 $bar = ucwords($foo);             // Mike O'hara
 
 $baz = ucwords($foo, " \t\r\n\f\v'");        // Mike O'Hara
--- a/reference/strings/functions/ucwords.xml
+++ b/reference/strings/functions/ucwords.xml
@@ -24,11 +24,6 @@
    after any character listed in the <parameter>delimiters</parameter> parameter
    (By default these are: space, form-feed, newline, carriage return, horizontal tab, and vertical tab).
   </para>
-  <para>
-   This function assumes the string is a basic ASCII string, and is therefore binary safe. If you need
-   to handle strings in non-ASCII character sets (such as UTF-8), use
-   <function><link linkend="function.mb-convert-case">mb_convert_case</link></function> instead.
-  </para>
  </refsect1>
 
  <refsect1 role="parameters">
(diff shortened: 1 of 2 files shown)


reference/strings/functions/ucwords.xml
```diff
@@ -24,6 +24,11 @@
    after any character listed in the <parameter>delimiters</parameter> parameter
    (By default these are: space, form-feed, newline, carriage return, horizontal tab, and vertical tab).
   </para>
+  <para>
+   This function assumes the string is a basic ASCII string, and is therefore binary safe. If you need
+   to handle strings in non-ASCII character sets (such as UTF-8), use
+   <function><link linkend="function.mb-convert-case">mb_convert_case</link></function> instead.
+  </para>
```
cmb69, line 31: Ah, that's pretty important information that is also [missing elsewhere](https://bugs.php.net/bug.php?id=79986). From a quick look at the implementation, and some quick checks, this function seems to be locale dependent (it uses tolower() and toupper()), so does not only work for ASCII strings; but since it's working on the single bytes, it likely fails miserably on multibyte and widechar locales. Not sure how that could be worded, but perhaps introducing an entity for that (or is there already something?) makes sense.

Regarding "binary safe": my understanding of that term in the manual is that it's about NUL bytes in strings. As such `ucwords` is binary safe, but that info might be confusing in this context. Also there is some pre-defined entity for binary safe functions(?)

reference/strings/functions/ucwords.xml
```diff
@@ -105,13 +110,29 @@ $bar = ucwords(strtolower($bar)); // Hello World!
    <example>
     <title><function>ucwords</function> example with custom delimiter</title>
     <programlisting role="php">
-<![CDATA[
+     <![CDATA[
```
salathe, line 113: Please undo this whitespace change.

reference/strings/functions/ucwords.xml
salathe: :+1:

Crell:
Here's an updated version that tries to break the single-byte warning out to an entity reference.  The text likely could still use work.  GitHub suggestions welcome.

reference/strings/functions/ucwords.xml
```diff
@@ -112,6 +112,22 @@ $bar = ucwords($foo);             // Hello|world!
...
+   </example>
+  </para>
+
+  <para>
+   <example>
+    <title><function>ucwords</function> example with additional delimiters</title>
+    <programlisting role="php">
+     <![CDATA[
+<?php
+$foo = 'mike o'hara';
```
salathe, line 126: Syntax error 😊

Crell: Bah! Fixed.

Crell:
I think this covers everything that's outstanding.

cmb69:
Thanks, @Crell! Applied as http://svn.php.net/viewvc?view=revision&revision=350369.

PR #138, opened 2020-08-18
Add `AddressInfo` and `Socket` resources

Ayesh:
Adds documentation for the `Socket` and `AddressInfo` resource types from the `sockets` extension.
There resource types are migrated to [AddressInfo and Socket objects in PHP 8.0](https://php.watch/versions/8.0/sockets-sockets-addressinfo), but this documentation will help for older PHP versions.

### DIFF
--- a/appendices/resources.xml
+++ b/appendices/resources.xml
@@ -27,6 +27,21 @@
      </thead>
      <tbody>
    <row>
+   <entry>AddressInfo</entry>
+    <entry>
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350334.

Ayesh:
Thanks a lot @cmb69 👍🏼.

PR #139, opened 2020-08-22
Document new com.dotnet_version INI setting

cmb69:
Cf. <https://github.com/php/php-src/pull/5949>.


cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=350811.

PR #140, opened 2020-08-25
Document JIT related OPcache INI settings

cmb69:
Any help on filling out the missing descriptions is appreciated!

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -1144,7 +1144,7 @@
      </para>
     </listitem>
    </varlistentry>
-   <varlistentry xml:id="ini.opcache.jit-max-loops-unroll">
+   <varlistentry xml:id="ini.opcache.jit-max-loop-unrolls">
     <term>
      <parameter>opcache.jit_max_loops_unroll</parameter>
      <type>OnUpdateUnrollL</type>
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -336,7 +336,7 @@
       <entry>Available as of PHP 8.0.0</entry>
      </row>
      <row>
-      <entry><link linkend="ini.opcache.jit-max-loops-unroll">opcache.jit_max_loops_unroll</link></entry>
+      <entry><link linkend="ini.opcache.jit-max-loop-unrolls">opcache.jit_max_loop_unrolls</link></entry>
       <entry>"8"</entry>
       <entry>PHP_INI_ALL</entry>
       <entry>Available as of PHP 8.0.0</entry>
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -1008,18 +1008,20 @@
     </term>
     <listitem>
      <para>
-
+      The amount of shared memory to reserve for compiled JIT code. A zero value disables the JIT.
      </para>
+     &ini.shorthandbytes;
     </listitem>
    </varlistentry>
    <varlistentry xml:id="ini.opcache.jit-debug">
     <term>
      <parameter>opcache.jit_debug</parameter>
-     <type>UNKNOWN</type>
+     <type>int</type>
     </term>
     <listitem>
      <para>
-
+      A bit mask specifying which JIT debug output to enable.
+      For possible values, please consult <filename>zend_jit.h</filename>.
      </para>
     </listitem>
    </varlistentry>
@@ -1030,7 +1032,8 @@
     </term>
     <listitem>
      <para>
-
+      Debugging option that disables JIT compilation after compiling a certain number of functions.
+      This may be helpful to bisect the source of a JIT miscompilation.
      </para>
     </listitem>
    </varlistentry>
@@ -1041,7 +1044,10 @@
     </term>
     <listitem>
      <para>
-
+      When using the "profile on first request" trigger mode, this threshold determines whether a
+      function is considered hot. The number of calls to the function divided by the number of calls
+      to all functions must be above the threshold. For example, a threshold of 0.005 means that
+      functions that made up more than 0.5% of all calls will be JIT compiled.
      </para>
     </listitem>
    </varlistentry>
@@ -1052,7 +1058,7 @@
     </term>
     <listitem>
      <para>
-
+      Maximum number of root traces.
      </para>
     </listitem>
    </varlistentry>
@@ -1063,7 +1069,7 @@
     </term>
     <listitem>
      <para>
-
+      Maximum number of side traces a root trace may have.
      </para>
     </listitem>
    </varlistentry>
@@ -1074,7 +1080,8 @@
     </term>
     <listitem>
      <para>
-
+      Maximum number of side trace exit counters. This limits the total number of side traces
+      there may be, across all root traces.
      </para>
     </listitem>
    </varlistentry>
@@ -1085,7 +1092,7 @@
     </term>
     <listitem>
      <para>
-
+      After how many iterations a loop is considered hot.
      </para>
     </listitem>
    </varlistentry>
@@ -1096,7 +1103,7 @@
     </term>
     <listitem>
      <para>
-
+      After how many calls a function is considered hot.
      </para>
     </listitem>
    </varlistentry>
@@ -1107,7 +1114,7 @@
     </term>
     <listitem>
      <para>
-
+      After how many returns a return is considered hot.
      </para>
     </listitem>
    </varlistentry>
@@ -1118,7 +1125,7 @@
     </term>
     <listitem>
      <para>
-
+      After how many exits a side exit is considered hot.
      </para>
     </listitem>
    </varlistentry>
@@ -1129,7 +1136,7 @@
     </term>
     <listitem>
      <para>
-
+      Maximum number of times the compilation of a root trace is attempted before it is blacklisted.
      </para>
     </listitem>
    </varlistentry>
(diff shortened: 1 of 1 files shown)


talotabi040:
NRS

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    <listitem>
+     <para>
+
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-buffer-size">
+    <term>
+     <parameter>opcache.jit_buffer_size</parameter>
+     <type>integer</type>
```
nikic, line 970: Drive-by note: We should canonicalize `<type>integer</type>` to `<type>int</type>` and `<type>boolean</type>` to `<type>bool</type>`. We do this correctly for function signatures, but apparently not for many other usages of `<type>`.

cmb69: See also https://github.com/php/doc-en/pull/174#discussion_r515692003.

cmb69: See https://github.com/php/doc-en/pull/176 and https://github.com/php/doc-base/pull/2.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-prof-threshold">
+    <term>
+     <parameter>opcache.jit_prof_threshold</parameter>
+     <type>float</type>
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1007: When using the "profile on first request" trigger mode, this threshold determines whether a function is considered hot. The number of calls to the function divided by the number of calls to all functions must be above the threshold. For example, a threshold of `0.005` means that functions that made up more than `0.5%` of all calls will the JIT compiled.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-loops-unroll">
+    <term>
+     <parameter>opcache.jit_max_loops_unroll</parameter>
+     <type>OnUpdateUnrollL</type>
```
nikic, line 1113:
```suggestion
     <type>int</type>
```

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1117: Maximum number of loops that can be unrolled within one trace.

(TBH I'm not totally sure what that means, is this about distinct loops or loop iterations?)

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-recursive-calls">
+    <term>
+     <parameter>opcache.jit_max_recursive_calls</parameter>
+     <type>OnUpdateUnrollL</type>
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1128: Maximum number of recursive calls that can be included within one trace.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-recursive-calls">
+    <term>
+     <parameter>opcache.jit_max_recursive_calls</parameter>
+     <type>OnUpdateUnrollL</type>
```
nikic, line 1124:
```suggestion
     <type>int</type>
```

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    <listitem>
+     <para>
+
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-recursive-return">
+    <term>
+     <parameter>opcache.jit_max_recursive_returns</parameter>
+     <type>OnUpdateUnrollL</type>
```
nikic, line 1135:
```suggestion
     <type>int</type>
```

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1139: Maximum number of recursive returns that may be included within one trace.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-polymorphic-calls">
+    <term>
+     <parameter>opcache.jit_max_polymorphic_calls</parameter>
+     <type>integer</type>
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1150: Maximum number of polymorphic (dynamic) calls to include within one trace.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-blacklist-root-trace">
+    <term>
+     <parameter>opcache.jit_blacklist_root_trace</parameter>
+     <type>integer</type>
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1095: Maximum number of times the compilation of a root trace is attempted before it is blacklisted.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,204 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-blacklist-side-trace">
+    <term>
+     <parameter>opcache.jit_blacklist_side_trace</parameter>
+     <type>integer</type>
+    </term>
+    <listitem>
+     <para>
+
```
nikic, line 1106: Maximum number of times the compilation of a side trace is attempted before it is blacklisted.

reference/opcache/ini.xml
cmb69: Maybe @dstogov can clarify?

cmb69:
Thanks a lot, @nikic!

reference/opcache/ini.xml
```diff
@@ -845,6 +953,278 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-polymorphic-calls">
+    <term>
+     <parameter>opcache.jit_max_polymorphic_calls</parameter>
+     <type>integer</type>
+    </term>
+    <listitem>
+     <para>
+      Maximum number of polymorphic (dynamic) calls to include within one trace.
```
dstogov, line 1224: Maximum number of attempts to inline polymorphic (dynamic or method) call. Calls above this limit treated as megamorphic and not inlined.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,278 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-recursive-return">
+    <term>
+     <parameter>opcache.jit_max_recursive_returns</parameter>
+     <type>int</type>
+    </term>
+    <listitem>
+     <para>
+      Maximum number of recursive returns that may be included within one trace.
```
dstogov, line 1213: Maximum number of unrolled recursive return loops.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,278 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-recursive-calls">
+    <term>
+     <parameter>opcache.jit_max_recursive_calls</parameter>
+     <type>int</type>
+    </term>
+    <listitem>
+     <para>
+      Maximum number of recursive calls that can be included within one trace.
```
dstogov, line 1202: Maximum number of unrolled recursive call loops.

reference/opcache/ini.xml
```diff
@@ -845,6 +953,278 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-loop-unrolls">
+    <term>
+     <parameter>opcache.jit_max_loops_unroll</parameter>
+     <type>int</type>
+    </term>
+    <listitem>
+     <para>
+      Maximum number of loops that can be unrolled within one trace.
```
dstogov, line 1191: Maximum number of attempts to unroll a loop in a side trace, trying to reach the root trace and close the outer loop.

reference/opcache/ini.xml
nikic: Just to check, this limits the number of side-traces from a single polymorphic call, right?

dstogov: Human language is not the best one :)

For example, with "jit_max_polymorphic_calls=2", "$obj->foo()" is going to be inlined for the first occurred class of $obj, then if we get object of another class, a side trace with another inlined method is going to be compiled, then if we get object of third class, another side trace starts, but method is not going to be inlined anymore. Trace is terminated by the transfer to the called method (goto *opline->handler).

cmb69:
Thanks, @dstogov! I have incorporated your suggestions. 

I think this good to be merged now.

reference/opcache/ini.xml
```diff
@@ -815,6 +923,280 @@
...
+     <para>
+      A bit mask specifying which JIT debug output to enable.
+      For possible values, please consult <filename>zend_jit.h</filename>.
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-bisect-limit">
+    <term>
+     <parameter>opcache.jit_bisect_limit</parameter>
+     <type>integer</type>
```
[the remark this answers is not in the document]
nikic: Here and in a few more places.

reference/opcache/ini.xml
```diff
@@ -815,6 +923,280 @@
...
+   </varlistentry>
+   <varlistentry xml:id="ini.opcache.jit-max-polymorphic-calls">
+    <term>
+     <parameter>opcache.jit_max_polymorphic_calls</parameter>
+     <type>integer</type>
+    </term>
+    <listitem>
+     <para>
+      Maximum number of attempts to inline polymorphic (dynamic or method) calls.
+      Calls above this limit treated as megamorphic and are not inlined.
```
nikic, line 1196:
```suggestion
      Calls above this limit are treated as megamorphic and are not inlined.
```

nikic (approved):

cmb69:
Thanks Nikita!

Applied as http://svn.php.net/viewvc?view=revision&revision=351205.

PR #141, opened 2020-08-25
Improve docs of round()

PhrozenByte:
Passing a negative precision with an absolute value greater than or equal to the number of digits is NOT treated specially, `round(987, 3)` yields the correct result of `float(1000)`, even though the docs incorrectly imply `float(0)`. See https://3v4l.org/HjU7W

### DIFF
--- a/reference/math/functions/round.xml
+++ b/reference/math/functions/round.xml
@@ -215,6 +215,7 @@ var_dump(round(9.5, 0, PHP_ROUND_HALF_DOWN));
 var_dump(round(9.5, 0, PHP_ROUND_HALF_EVEN));
 var_dump(round(9.5, 0, PHP_ROUND_HALF_ODD));
 
+echo PHP_EOL;
 echo 'Rounding modes with 8.5' . PHP_EOL;
 var_dump(round(8.5, 0, PHP_ROUND_HALF_UP));
 var_dump(round(8.5, 0, PHP_ROUND_HALF_DOWN));
@@ -231,6 +232,7 @@ float(10)
 float(9)
 float(10)
 float(9)
+
 Rounding modes with 8.5
 float(9)
 float(8)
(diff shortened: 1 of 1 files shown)


cmb69:
But see https://3v4l.org/bLtdo.

PhrozenByte:
The docs read as if `round()` treated this somehow special, but it doesn't. `round(987, -4)` simply means "round to the nearest multiple of 10,000" - and the nearest multiple of 10,000 for 987 is indeed 0. Nothing special about that. However, explicitly mentioning this implies that there's something special about it, causing confusion.

Girgias:
Having written that piece of documentation, I think adding the specific example you gave to the docs would be good. Because I got personally confused when I was trying to document the negative precision as I don't think the rounding to ``10^(n)`` where ``n = |precision|`` is the most obvious. So I think rewriting to something like that might be better than just dropping the erroneous information.

PhrozenByte:
I've just updated the PR, what do you guys think about that? I assume expressing it mathematically is no option?

cmb69:
> means "round to the nearest multiple of 10,000"

Thanks! This is a very concise description, I'd like to see in the docs (generalized to `10^^$precision`). We should also change the description of the modes, which talk of "decimal places" (but that doesn't need to be done in this PR).

PhrozenByte:
> Thanks! This is a very concise description, I'd like to see in the docs (generalized to `10^^$precision`).

Done :+1: 

> We should also change the description of the modes, which talk of "decimal places" (but that doesn't need to be done in this PR).

Done, too :+1: 

I'm not sure whether we should add a note that "rounding up" (and down) in programming is different from "rounding up" (and down) in mathematics. Mathematically speaking "rounding up" is defined as "rounding towards positive infinity", causing 1.5 to get 2 and -1.5 to get 1. However, `PHP_ROUND_HALF_UP` is implemented as "rounding away from zero", causing 1.5 to get 2 and -1.5 to get -2. This is often called "commercial rounding". Same for `PHP_ROUND_HALF_DOWN`. *This is not a bug!* Many languages do the same (e.g. Java and Python), it's pretty common. I'm just thinking about whether we should emphasis this or not.

Girgias:
> I'm not sure whether we should add a note that "rounding up" (and down) in programming is different from "rounding up" (and down) in mathematics. Mathematically speaking "rounding up" is defined as "rounding towards positive infinity", causing 1.5 to get 2 and -1.5 to get 1. However, `PHP_ROUND_HALF_UP` is implemented as "rounding away from zero", causing 1.5 to get 2 and -1.5 to get -2. This is often called "commercial rounding". Same for `PHP_ROUND_HALF_DOWN`. _This is not a bug!_ Many languages do the same (e.g. Java and Python), it's pretty common. I'm just thinking about whether we should emphasis this or not.

Can't harm to be more descriptive :)

PhrozenByte:
> > I'm not sure whether we should add a note that "rounding up" (and down) in programming is different from "rounding up" (and down) in mathematics. Mathematically speaking "rounding up" is defined as "rounding towards positive infinity", causing 1.5 to get 2 and -1.5 to get 1. However, `PHP_ROUND_HALF_UP` is implemented as "rounding away from zero", causing 1.5 to get 2 and -1.5 to get -2. This is often called "commercial rounding". Same for `PHP_ROUND_HALF_DOWN`. _This is not a bug!_ Many languages do the same (e.g. Java and Python), it's pretty common. I'm just thinking about whether we should emphasis this or not.
> 
> Can't harm to be more descriptive :)

Yeah, I'm not so sure about that :thinking: Don't you think a note like "Please note that 'rounding up' in programming is different from 'rounding up' in mathematics." causes confusion because one would ask how and why it is different? Not sure whether a non-mathematician even notices that, I'd assume most people assume commercial rounding for the term "rounding up", and not the mathematically correct "rounding towards positive infinity".

salathe:
A plain description of the behaviour in PHP would suffice.  More description certainly can be more harmful than good.

PhrozenByte:
Great, I just didn't want to leave this unnoticed and give a heads up. I just removed some examples from the "mode with precision examples" section which don't really improve understandability and would say that this can be merged now :+1:

salathe (approved):

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350529.

PR #142, opened 2020-09-02
Reflection can now get defaults for internals

SjonHortensius:
https://github.com/php/php-src/pull/5353

### DIFF
--- a/reference/reflection/reflectionparameter/getdefaultvalue.xml
+++ b/reference/reflection/reflectionparameter/getdefaultvalue.xml
@@ -47,8 +47,8 @@
       <row>
        <entry>8.0.0</entry>
        <entry>
-        This function now allows getting the default value of built-in functions or
-        methods of built-in classes. Previously, a <classname>ReflectionException</classname> was thrown.
+        This method now allows getting the default value of parameters of built-in functions and
+        built-in class methods. Previously, a <classname>ReflectionException</classname> was thrown.
        </entry>
       </row>
      </tbody>
--- a/reference/reflection/reflectionparameter/getdefaultvalue.xml
+++ b/reference/reflection/reflectionparameter/getdefaultvalue.xml
@@ -48,7 +48,7 @@
        <entry>8.0.0</entry>
        <entry>
         This function now allows getting the default value of built-in functions or
-        methods of built-in classes.
+        methods of built-in classes. Previously, a <classname>ReflectionException</classname> was thrown.
        </entry>
       </row>
      </tbody>
--- a/reference/reflection/reflectionparameter/getdefaultvalue.xml
+++ b/reference/reflection/reflectionparameter/getdefaultvalue.xml
@@ -14,7 +14,7 @@
    <void />
   </methodsynopsis>
   <para>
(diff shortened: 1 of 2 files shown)


reference/reflection/reflectionparameter/getdefaultvalue.xml
```diff
@@ -70,17 +95,6 @@ Default value: baz
   </para>
  </refsect1>
  
- <refsect1 role="notes">
-  &reftitle.notes;
-  <note>
-   <para>
-    Due to implementation details, it is not possible to get the default value
```
cmb69, line 77: Maybe leave this note, but replace "Due to implementation details," with "Prior to PHP 8.0.0,"?

SjonHortensius: Isn't that what the changelog is for; to prevent the manual from becoming an archive of all the things prior PHP versions did?

cmb69: Well, yes. But we should keep documentation how it behaved in previous versions. So maybe add something like `Previously, a <classname>ReflectionException</classname> was thrown.`

SjonHortensius: I respectfully disagree. Has this ever been done on other pages?

cmb69: I'd say, that is the common approach wrt. behavioral changes; see https://www.php.net/manual/en/doc.changelog.php. The basic idea is that the documentation is supposed to cover several PHP versions, and so anybody who is interested in the behavior of a specific version should be able to find all the relevant info.

salathe: The *"Previously, ..."* sentence should be appended to the changelog entry.  Then, this note can be deleted.

reference/reflection/reflectionparameter/getdefaultvalue.xml
```diff
@@ -32,6 +32,31 @@
...
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function now allows getting the default value of built-in functions or
```
salathe, line 50: "This function" should be "This method".   It might also be better to rephrase the latter part, to something like "now allows getting the default value of parameters of built-in functions and built-in class methods."

salathe:
Note that the description section still says, *"Gets the default value of the parameter for a user-defined function or method."*

The page for `ReflectionParameter::getDefaultValueConstantName()` should also be updated, due to the same change.

reference/reflection/reflectionparameter/getdefaultvalue.xml
SjonHortensius: I used the text from the original note, but I don't mind rephrasing it

salathe (approved):

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350519.

PR #144, opened 2020-09-06
Clarify preloading and constant behavior.

Crell:
Nikita said in a [bug report](https://bugs.php.net/bug.php?id=79956) that constants aren't covered and won't be, which is suboptimal but this at least captures the facts on the ground more accurately.

### DIFF
--- a/reference/opcache/preload.xml
+++ b/reference/opcache/preload.xml
@@ -5,7 +5,7 @@
 
  <para>
   As of PHP 7.4.0, PHP can be configured to preload scripts into the opcache when the engine
-  starts.  Any symbols (functions, classes, etc.) in those files will then become
+  starts.  Any functions, classes, interfaces, or traits (but not constants) in those files will then become
   globally available for all requests without needing to be explicitly included.  That trades
   convenience and performance (because the code is always available) for baseline memory usage.  It also
   requires restarting the PHP process to clear pre-loaded scripts, meaning this feature is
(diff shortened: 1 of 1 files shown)


cmb69:
AFAIK, other entities such as classes *may* not be preloaded as well. Maybe it's better to have the docs be less explicit about what exactly happens (perhaps some details of the current behavior may also change in the future)?

Crell:
When would classes not preload?  That's the first I've heard of it anything like that...

Girgias:
> 
> 
> When would classes not preload? That's the first I've heard of it anything like that...

Pretty sure when a class extends an internal one it cannot be preloaded.

nikic:
It depends on the used preloading method. `require` will preload all declared classes, `opcache_compile_file()` is best-effort.

Crell:
Oh geez.  Can you define what "best effort" means in practice?

It sounds like preloading can't be relied on to ever actually work as intended, so the code needs to still be able to self-load everything.  That's fine for autoloading (there's basically zero cost to having an autoloader enabled that never does anything), but for function includes I don't know what the performance trade-off is or how to determine that.

Also, if require() is more reliable, then you need an autoloader enabled for preloading so that you don't need to manually handle class dependency order resolution.  Yuck. :-/

afilina:
Pinging maintainers in case this fell through the cracks.

afilina:
Looks like there's still a lot of clarification required for this PR, but there's a lack of bandwidth. Should we have a "blocked" tag or something to more easily sort open PRs that won't be resolved any time soon?

cmb69:
I've just added a "Waiting on Review" label (we have that for php-src as well).

nikic (approved):
If nothing else, the description is more correct than before...

nikic:
> It sounds like preloading can't be relied on to ever actually work as intended, so the code needs to still be able to self-load everything. That's fine for autoloading (there's basically zero cost to having an autoloader enabled that never does anything), but for function includes I don't know what the performance trade-off is or how to determine that.

Functions get loaded reliably, as long as you're not declaring them conditionally. As of PHP 8.1, classes also get loaded reliably as long as their dependencies are available (we obviously can't preload B extends A if A doesn't exist).

PR #145, opened 2020-09-08
Add array destructuring documentation to array syntax section

rask:
This redoes the original PR I made against this repo, but which got stale: #19 

Adds basic syntax documentation for array destructuring.

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -488,10 +488,45 @@ Array
 )
 ]]>
      </screen>
-    </informalexample>       
+    </informalexample>
 
    </note>
 
+  </sect3>
+
+  <sect3 xml:id="language.types.array.syntax.destructuring">
+   <title>Array destructuring</title>
+
+   <note>
+    <para>
+     Array destructuring is available only in PHP versions 7.1 and newer.
+    </para>
+   </note>
+
+   <para>
+    Arrays can be destructured symmetrically. What this means is you can
+    define local scope variables using values from an existing array.
+   </para>
+
+   <para>
+    This feature is similar to what the <function>list</function> construct offers.
+   </para>
+
+   <informalexample>
+    <programlisting role="php">
+ <![CDATA[
+ <?php
+
+ $source_array = ['foo', 'bar', 'baz'];
+
+ [$foo, $bar, $baz] = $source_array;
+
+ assert($foo === 'foo');
+ assert($bar === 'bar');
+ assert($baz === 'baz');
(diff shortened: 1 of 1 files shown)


language/types/array.xml
```diff
@@ -488,11 +488,46 @@ Array
...
    </note>
 
   </sect3>
+
+  <sect3 xml:id="language.types.array.syntax.destructuring">
+   <title>Array destructuring</title>
+
+   <note>
+    <para>
+     Array destructuring is available only in PHP versions 7.1 and newer.
```
Girgias, line 502:
```suggestion
     Array destructuring is available as of PHP 7.1.0.
```

language/types/array.xml
```diff
@@ -488,11 +488,46 @@ Array
...
+    </para>
+   </note>
+
+   <para>
+    Arrays can be destructured symmetrically. What this means is you can
+    define local scope variables using values from an existing array.
+   </para>
+
+   <para>
+    This feature is similar to what the <function>list</function> construct offers.
```
Girgias, line 512:
```suggestion
    This feature is equivalent to the <function>list</function> construct.
```

Girgias (changes requested):
I'm concerned that the array type pages is maybe not the best place as it's already a massive and dense page, but this has nothing to do with the PR really.

afilina:
Pinging @rask to address suggestions. Looks like this PR fell through the cracks.

cmb69 (changes requested):
IMO, we should not duplicate the description of `list()` (array destructoring is just syntax sugar for that, like literal arrays are syntax sugar for `array()`.  Instead we should clarify on the list man page, that the shorthand syntax is supported as of PHP 7.1.0 and maybe give some examples. It may make sense to add a short note to the array docs, telling users that literal arrays used as lvalues are actually destructuring and link to list (just in case somebody is mislead by that construct).

gallogit (changes requested):
Corregir errores del sistema para licencias en México multriples

gallogit (approved):
Corregir error

afilina:
The comments were not addressed by the author in over a year, despite a reminder in January. Should we consider this PR abandoned and close it? @rask

kamil-tekiela:
I have opened a new PR recently which tried to address the open points in this PR and add even more information to the page. I believe my PR supersedes this one.

cmb69:
Oh, right, there is PR #796, so I'm closing this one.

PR #147, opened 2020-09-27
ReflectionMethod accepts string|object, not mixed

BenMorel:
ReflectionMethod's constructor currently uses `mixed` for its `$class` parameter, when it actually accepts either a `string` or an `object`.

There is precedence in the docs to use union types notation (see [this](https://www.php.net/manual/fr/resourcebundle.get.php) for example), so is it OK if I convert more `mixed` to union types in the future?

### DIFF
--- a/reference/reflection/reflectionmethod/construct.xml
+++ b/reference/reflection/reflectionmethod/construct.xml
@@ -11,7 +11,7 @@
   &reftitle.description;
   <methodsynopsis>
    <modifier>public</modifier> <methodname>ReflectionMethod::__construct</methodname>
-   <methodparam><type>mixed</type><parameter>class</parameter></methodparam>
+   <methodparam><type>string|object</type><parameter>class</parameter></methodparam>
    <methodparam><type>string</type><parameter>name</parameter></methodparam>
   </methodsynopsis>
   <methodsynopsis>


cmb69:
We need some clean way to represent union types for docbook, and to have PhD properly render these, before we should add union types to the manual. See https://github.com/php/phd/pull/25#issuecomment-669513888.

BenMorel:
Thanks for the pointer. In the meantime, can't we follow the footsteps of the doc pages that have already adopted union types in the following form?

```xml
<type>bool|array</type>
```

That would allow us to start working on better type documentation right now, and should be trivial to replace programmatically once a suitable syntax has been made available.

There are quite a few doc pages already using this syntax: https://pastebin.com/q3Yu5GPd

Girgias:
> 
> 
> Thanks for the pointer. In the meantime, can't we follow the footsteps of the doc pages that have already adopted union types in the following form?
> 
> ```
> <type>bool|array</type>
> ```
> 
> That would allow us to start working on better type documentation right now, and should be trivial to replace programmatically once a suitable syntax has been made available.
> 
> There are quite a few doc pages already using this syntax: https://pastebin.com/q3Yu5GPd

This just makes more work for translations, because we need to do 2 passes instead of just 1. Moreover, a lot of arguments in PHP 7.x do actually accept "mixed" due to legacy behaviour which only got removed with PHP 8.0.

As we need to update the prototypes *anyway* due to named parameters it would be better to combine all of this into one single change/commit by basing ourselves on the stubs when they are finalized.

BenMorel:
Just stumbled upon another one here: https://www.php.net/manual/en/function.getmyuid.php

The doc says `getmyuid ( void ) : int`; that's a mistake, it should be `: int|false`. What should we do for this one?

- leave as is: it's a documentation bug, probably not a good idea, even if the description does mention `false`
- change for `: mixed`: I hate the idea
- or again, use `<type>int|false</type>`?

cmb69:
> leave as is: it's a documentation bug, probably not a good idea, even if the description does mention false

IMO, leave it as is for PHP 7 docs, but do it properly for PHP 8 docs. It seems to me that it might be benefical to have separate signatures for PHP 7/8, at least for some functions. The PHP 8 [signatures should be generated (semi-)automatically](https://github.com/php/php-src/pull/5895).

> change for : mixed: I hate the idea

That would even be counter productive, since `false` signals failure, i.e. the rather exceptional situation.

> or again, use <type>int|false</type>?

Clear no from me. That [should be](https://github.com/php/doc-en/pull/151) `<type class="union"><type>int</type><type>false</type></type>` instead (what is still [not supported by PhD](https://github.com/php/phd/pull/30)), "even though that triples the size of thy methodsynopses and produce aches in thy typing fingers, for if thou thinkest 'this is too verbose', the gods shall surely punish thee for thy arrogance." I mean, we use DocBook for a reason. :) And, anyway, that type could be shortened to `<type class="union">&integer;&false;</type>` if desired.

BenMorel:
> IMO, leave it as is for PHP 7 docs, but do it properly for PHP 8 docs.

Will PHP 7 & PHP 8 have different doc pages?

cmb69:
> Will PHP 7 & PHP 8 have different doc pages?

That's not planned.

BenMorel:
So I didn't get your comment, sorry?

> IMO, leave it as is for PHP 7 docs, but do it properly for PHP 8 docs.

cmb69:
Sorry, pretty unclear explanation from side, indeed. What I meant is that we should focus on (semi-)automatic generation of the PHP 8 function signatures. If these are identical to those for PHP 7, just replace them (that would be the case for `getmyuid()`). If the signatures are different (e.g. because some error returns have been replaced by exceptions in PHP 8), we could have two different signatures (one for PHP 8, and one for PHP 7).

However, thinking more about it, there's nothing wrong with fixing the PHP 7 signatures as soon as we have support for union types.

cmb69:
I have applied this (with proper union type markup) as http://svn.php.net/viewvc?view=revision&revision=351060.

Wrt. getmyuid() returning `int|false`: I don't see much point in fixing individual functions. There are likely very many which may return `false`, but that is not yet reflected in the signature. It seems to me that a script which checks for `&return.falseforfailure;` in the "return values" section, could insert that automatically to the return type.

PR #150, opened 2020-10-03
Note case-sensitivity of the parameter name.

Crell:
One of the comments said this was the case.  No sense leaving it to a comment.  (I'll delete the comment afterward.)


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350755.

PR #151, opened 2020-10-06
Properly mark up nullable and union types

cmb69:
We change all existing nullable and union types to use the dedicated DocBook markup for union types. This requires support in PhD for proper rendering; see https://github.com/php/phd/pull/30.


cmb69:
Given that this PR would remove the `array|object` docs, general union type docs should be added.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351075, http://svn.php.net/viewvc?view=revision&revision=351076 and http://svn.php.net/viewvc?view=revision&revision=351077.

PR #153, opened 2020-10-11
parse_ini_file comment cleanup

Crell:
All of these updates are based on comments on the page, most of which I have since deleted as they'll either be incorporated here or we'll decide they just don't matter.

### DIFF
--- a/reference/filesystem/functions/parse-ini-file.xml
+++ b/reference/filesystem/functions/parse-ini-file.xml
@@ -108,7 +108,7 @@ urls[git] = "http://git.php.net"
      may also be parsed
      in the ini file so if you define a constant as an ini value before
      running <function>parse_ini_file</function>, it will be integrated into
-     the results.  Only ini values are evaluated.  For example:
+     the results.  Only ini values are evaluated, and the value must be just the constant.  For example:
     </para>
     <programlisting role="php">
 <![CDATA[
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/parse-ini-file.xml
```diff
@@ -32,7 +32,8 @@
      <term><parameter>filename</parameter></term>
      <listitem>
       <para>
-       The filename of the ini file being parsed.
+       The filename of the ini file being parsed. If a relative path is used,
+       it is evaluated relative to the current include path, then the current working directory.
```
cmb69, line 36: Is this true? [`include`](https://www.php.net/manual/en/function.include.php) is documented to work differently, and I'm not even sure if the `include_path` is used by `parse_ini_file` at all.

Crell: Frankly I'm not even sure.  That's what one of the comments said.  I've not tested it myself.

Crell: OK, just tested, and this isn't quite right.

* It *does* read from the include path.
* The cwd is read first.  If a file is in cwd and in the include path, the cwd file gets used.

I don't understand why it was written this way, but that's how it seems to work, so... :man_shrugging: 

I'll update the text momentarily.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350855.

PR #63, opened 2020-04-01
Document the __serialize/__unserialize magic in PHP 7.4.

Crell:
I'm not sure if there's anywhere else they need to be cross-linked or mentioned.  If you know of one, let me know and I'll add it.

PR #154, opened 2020-10-12
Link to __serialize()/__unserialize() from Serializable

TysonAndre:
Those magic methods were introduced in php 7.4 to resolve some of the
limitations with __wakeup() or Serializable

Related to https://github.com/php/doc-en/pull/63


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350854.

PR #155, opened 2020-10-17
Fix #78142: Function prototype return value does not contain type in …

cmb69:
…case of failure

---

This is blocked by missing support for union type rendering in PhD (see https://github.com/php/phd/pull/30).


cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351059.

PR #156, opened 2020-10-18
Remove mention of PHP 5 in Language Reference section

Girgias:
This does not handle the OOP section nor the Error in PHP 7 page/section (https://www.php.net/manual/en/language.errors.php)

Also I'm not sure what to do about a couple of pages from the Predefined Variables sections, such as https://www.php.net/manual/en/reserved.variables.get.php which have a `$HTTP_GET_VARS [deprecated]`, have these been removed or not in PHP 7?

Other than those case, everything in https://www.php.net/manual/en/langref.php should be covered.

This also exists as a patch on the online editor: http://edit.php.net/?patchID=21&project=PHP

One other question is if https://www.php.net/manual/en/reserved.variables.httprawpostdata.php should be removed?


language/namespaces.xml
```diff
@@ -767,7 +748,7 @@ function toGreenlandic()
   <sect2 xml:id="language.namespaces.importing.group">
    <title>Group <literal>use</literal> declarations</title>
    <para>
-    From PHP 7.0 onwards, classes, functions and constants being imported from
+    Cclasses, functions and constants being imported from
```
cmb69, line 751:
```suggestion
    classes, functions and constants being imported from
```

appendices/tokens.xml
```diff
@@ -266,7 +266,7 @@
      <entry><constant>T_ELLIPSIS</constant></entry>
      <entry>...</entry>
      <entry>
-      <link linkend="functions.variable-arg-list.new">function arguments</link>
+      <link linkend="functions.variable-arg-list">function arguments</link>
       (available since PHP 5.6.0)
```
cmb69, line 270:
```suggestion
```

cmb69:
> […] which have a $HTTP_GET_VARS [deprecated], have these been removed or not in PHP 7?

The have already been [removed in PHP 5.4.0](https://www.php.net/manual/en/ini.core.php#ini.register-long-arrays) (php-src confirms)

> One other question is if https://www.php.net/manual/en/reserved.variables.httprawpostdata.php should be removed?

IMO, yes.

language/predefined/closure/construct.xml
```diff
@@ -29,6 +29,7 @@
 
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
+  <!-- TODO Check: this doesn't seem to be true -->
```
cmb69, line 32: Indeed, it is an `Error` as of PHP 7.0.0 (see https://3v4l.org/U16Hd).

cmb69 (commented):
Looks pretty good to me, but a few issue needs to be addressed.

language/predefined/closure/construct.xml
Girgias: Should I remove the section?

cmb69: I think yes, since constructors never have a return value. It might be appropriate to add an `<refsect1 role="errors">`, though.

language/predefined/closure/construct.xml
```diff
@@ -29,9 +29,9 @@
...
   <para>
-   This method has no return value; it simply emits an error
-   (of type <constant>E_RECOVERABLE_ERROR</constant>).
+   This method has no return value; it simply throws an <class>Error</class>.
```
cmb69, line 34:
```suggestion
   This method has no return value; it simply throws an <classname>Error</classname>.
```

language/basic-syntax.xml
```diff
@@ -125,18 +96,9 @@ echo "Last statement";
...
-    <code>&lt;script language="php"&gt; &lt;/script&gt;</code>, are always
-    available. There is also the short echo tag <code>&lt;?= ?&gt;</code>,
-    which is always available in PHP 5.4.0 and later.
+    There is also the short echo tag <code>&lt;?= ?&gt;</code>.
    </para>
    <para>
-    The other two are short tags and <productname>ASP</productname> style
-    tags. As such, while some people find short tags and
-    <productname>ASP</productname> style tags convenient, they are less
-    portable, and generally not recommended.
```
cmb69, line 139: I'd keep this section (modulo the ASP style tags).

Girgias: Even if there is already a section below about them? Doesn't that duplicate the docs?

cmb69: Sorry, missed that. All fine.

language/predefined/closure/construct.xml
cmb69: Well, that was bad advice, since that error is not particularly special.

PR #157, opened 2020-10-18
Remove migration guides for PHP 5

Girgias:
@salathe, @cmb69 can I have your opinion on this?


cmb69:
I'm not sure about that. The migration guides might be helpful for users wanting to upgrade to PHP 7.0, and they may still be on PHP 5.3 or so. On the other hand, at least keeping the guides for early PHP 5 versions may not make any sense at all.

cmb69:
Given that [legacy docs](https://php-legacy-docs.zend.com/manual/php5/en/index) are available now, I think we can remove the PHP 5 migration guides (maybe keep 56 nonetheless). Could you please fix the merge conflicts?

afilina:
@Girgias I can help by fixing the merge conflicts in a separate PR if you're busy.

cmb69:
Thank you!

eusonlito:
I don't undestand why we need to remove old docs. It's always a good reference and is not adding noise to appendices page because new versions are added always on page top. Developers working with projects in different PHP versions is a good entry point to check which features are availables on project PHP version.

I have created this php changes cheatsheet automated from appendices page to simplify new/deprecated features search across different versions https://eusonlito.github.io/php-changes-cheatsheet/index.html

Why references to previous migrations are not usefull in docs?

Thanks!

Girgias:
> 
> 
> I don't undestand why we need to remove old docs. It's always a good reference and is not adding noise to appendices page because new versions are added always on page top. Developers working with projects in different PHP versions is a good entry point to check which features are availables on project PHP version.
> 
> I have created this php changes cheatsheet automated from appendices page to simplify new/deprecated features search across different versions https://eusonlito.github.io/php-changes-cheatsheet/index.html
> 
> Why references to previous migrations are not usefull in docs?
> 
> Thanks!

Because it puts a burden on maintainers and translations, we *did* need to edit theses pages by removing PHP 5 only extensions, and if it's present on the official docs they *need* to be up to date, we've fixed some content on one of the PHP 5 migration pages due to something incorrect.

And the reason for the removal is that we do *not* document PHP 5 any more only PHP 7 and 8, after the storm of PHP 8 has calmed down we're going to go through all the docs and remove mentions and specific information relating to PHP 5

sgolemon:
> Because it puts a burden on maintainers and translations

Another option we might have considered was to put a language-snippet on the older pages to the effect of:

> This document is legacy content and its translation may not be up to date.  Please reference the English version if possible.

Translate that one snippet to our half a dozen languages and re-use it anywhere we need to sunset legacy content.  The English version of the snippet would then be a no-op as it's the canonical document.

Something like that would preserve that older knowledge in a way that's accessible and easily surfaceable without burdening translators.

PR #158, opened 2020-10-18
Don't use FQNs for types

cmb69:
Using fully qualified names for identifiers in the manual is uncommon;
at least identifiers of the global namespace are (almost) always
written unqualified.  Furthermore, FQNs are not supported by the auto-
linking of PhD.

---

This PR covers Componere, but it is actually more about establishing a convention. If other prefer fully qualified names, we should go that route, but at least be consistent in the rendered manual (i.e. add proper support for FQNs, and render possibly missing leading backslashes).


cmb69:
@krakjoe might be interested in this, because he's the lead maintainer of several extensions which have namespaced entities.

nikic (approved):

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350956 (I've "fixed" other occurences of `<type>\` right away).

PR #159, opened 2020-10-18
array_rand updates based on comments

Crell:
Comments to delete once this is merged:

https://www.php.net/manual/en/function.array-rand.php#120180
https://www.php.net/manual/en/function.array-rand.php#121699


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350939. Comments deleted.

PR #160, opened 2020-10-21
Fix #77809: getStats() may return false

cmb69:


fgm:
Is there a way I can check the result of the PR in an actual doc page ? Or at least find the documentation for the XML format used ?

cmb69:
@fgm, no, there is no preview available. It might make sense to add a build step to CI.

The documentation is written in DocBook, but besides general DocBook documentation, there are no specific docs.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351058.

PR #161, opened 2020-10-22
Document ArrayAccess in PHP-DS

BenMorel:
The following classes/interfaces in PHP-DS implement `ArrayAccess`:

- `Map`
- `Stack`
- `Sequence` (`Deque`, `Vector`)

But this is not currently documented.

---

Script to check which classes/interfaces implement `ArrayAccess`:

```php
$classes = array_merge(get_declared_interfaces(), get_declared_classes());

foreach ($classes as $class) {
    if (substr($class, 0, 3) === 'Ds\\') {
        if ((new ReflectionClass($class))->implementsInterface(ArrayAccess::class)) {
            echo "$class\n";
        }
    }
}
```

```
Ds\Sequence
Ds\Vector
Ds\Deque
Ds\Stack
Ds\Map
```

### DIFF
--- a/reference/ds/ds.deque.xml
+++ b/reference/ds/ds.deque.xml
@@ -113,6 +113,29 @@
   </section>
 <!-- }}} -->
 
+  <section role="changelog">
+   &reftitle.changelog;
+   <para>
+    <informaltable>
+     <tgroup cols="2">
+      <thead>
+       <row>
+        <entry>&Version;</entry>
+        <entry>&Description;</entry>
+       </row>
+      </thead>
+      <tbody>
+       <row>
+        <entry>1.3.0</entry>
+        <entry>
+         The class now implements <classname>ArrayAccess</classname>.
+        </entry>
+       </row>
+      </tbody>
+     </tgroup>
+    </informaltable>
+   </para>
+  </section>
 
  </partintro>
 
--- a/reference/ds/ds.deque.xml
+++ b/reference/ds/ds.deque.xml
@@ -126,7 +126,7 @@
       </thead>
       <tbody>
        <row>
-        <entry>1.3.0</entry>
+        <entry>PECL ds 1.3.0</entry>
         <entry>
          The class now implements <classname>ArrayAccess</classname>.
         </entry>
--- a/reference/ds/ds.map.xml
+++ b/reference/ds/ds.map.xml
@@ -59,6 +59,9 @@
      <oointerface>
       <interfacename>Ds\Collection</interfacename>
      </oointerface>
+     <oointerface>
+      <interfacename>ArrayAccess</interfacename>
+     </oointerface>
     </classsynopsisinfo>
 <!-- }}} -->
     <classsynopsisinfo role="comment">Constants</classsynopsisinfo>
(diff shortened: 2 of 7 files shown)


cmb69:
Thanks for the PR! Implementing ArrayAccess seems to be a very recent addition for 1.3.0 (https://github.com/php-ds/ext-ds/commit/b297e8f37f4a6f5988579b0f6250aad32f141df7), and that should be mentioned in the manual, possibly as a changelog entry.

BenMorel:
Indeed, this deserves an entry in the changelog! Actually there are 2 more classes that should implement `ArrayAccess` (see https://github.com/php-ds/ext-ds/pull/163), so I'll finish this PR once the related PR gets approved.

In the meantime, could you please point me to where I can learn how to compile the PHP docs to HTML? I'm new to the XML format in use, and would like to be able to test my changes locally.

cmb69:
See the "Build the documentation" section at http://doc.php.net/tutorial/local-setup.php.

BenMorel:
Thank you, @cmb69! All done now, ready for review.

But please wait for @rtheunissen's feedback on https://github.com/php-ds/ext-ds/pull/163 before considering merging!

reference/ds/ds.deque.xml
```diff
@@ -113,6 +113,29 @@
...
+     <tgroup cols="2">
+      <thead>
+       <row>
+        <entry>&Version;</entry>
+        <entry>&Description;</entry>
+       </row>
+      </thead>
+      <tbody>
+       <row>
+        <entry>1.3.0</entry>
```
cmb69, line 129:
```suggestion
        <entry>PECL ds 1.3.0</entry>
```
Otherwise we get confusing version info on https://www.php.net/manual/en/doc.changelog.php.

BenMorel: Ah, sorry about that. I did think that this might bring confusion, but saw other extensions doing the same so thought that was the norm:

https://github.com/php/doc-en/blob/057d46abe671f7775ac3bf6739f18110288f7787/reference/mongo/mongolog.xml#L309

BenMorel: Fixed in fcc6727fb0901f834f50109b4c777c001a25f023.

cmb69: Sadly, the manual is pretty inconsistent in this regard.

BenMorel: I can write a quick script to identify these and fix them, if you wish.

(Basically, looking for any `<entry>` version < 4.x which would mean that we're not talking about a PHP version, as this manual does not have any changelog for PHP < 4.x, if I'm not mistaken. That would maybe not flag *all* of them, but *most* of them).

cmb69: Sure, that would be very nice! :)

rtheunissen:
I have merged the array access changes, which will be released soon alongside a few other fixes on 1.3.0

BenMorel:
@rtheunissen Thanks for your responsiveness! :+1: 

@cmb69 Ready to merge then!

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350981.

BenMorel:
Thank you! Is there still a work in progress to move the repo to Git?

cmb69:
Unfortunately, that migration has stalled.

PR #162, opened 2020-10-23
Add PECL in front of version numbers in changelogs

BenMorel:
As discussed in https://github.com/php/doc-en/pull/161#discussion_r511020631 with @cmb69, this PR adds the PECL package name in front of the PECL versions in the changelogs, to avoid confusing them with PHP versions.

The changes in this PR were generated using this script:
https://gist.github.com/BenMorel/19499bf5c918cd0ea0f0b3f27a5a8254

Open question: is `/reference/<name>/` always a PECL package name?

### DIFF
--- a/reference/apcu/functions/apcu-cache-info.xml
+++ b/reference/apcu/functions/apcu-cache-info.xml
@@ -62,13 +62,13 @@
      </thead>
      <tbody>
       <row>
-       <entry>3.0.11</entry>
+       <entry>PECL apcu 3.0.11</entry>
        <entry>
         The <parameter>limited</parameter> parameter was introduced.
        </entry>
       </row>
       <row>
-       <entry>3.0.16</entry>
+       <entry>PECL apcu 3.0.16</entry>
        <entry>
         The "<literal>filehits</literal>" option for the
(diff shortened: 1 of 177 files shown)


cmb69:
> Open question: is /reference/<name>/ always a PECL package name?

No, it is not. It can be a bundled extension as well (e.g. ldap), but it also can be something made up (e.g. var, image). That is something that bugs me for the automatic signature generation, too.

cmb69:
It might be possible to determine whether a package is a PECL package, by looking at the reference/*/version.xml files. Or, maybe better anyway, just have a manually compiled list.

BenMorel:
Here is a list of the unique packages affected by this PR:

- apcu
- event
- expect
- geoip
- imagick
- memcached
- mongo
- mongodb
- oauth
- rar
- sdo
- solr
- sphinx
- stomp
- sync
- win32service
- xhprof
- yaml

From a quick search, it looks like all of them are PECL packages! :+1:

cmb69:
Indeed! Thanks for checking, and the PR in general! Applied as http://svn.php.net/viewvc?view=revision&revision=350988.

BenMorel:
Thanks! :+1:

PR #165, opened 2020-10-25
Remove PHP 5 from static keyword page

tiffany-taylor:
I'll start working on removal of PHP 5 from OOP section, this was a file I encountered while browsing the manual. I'll add more files as I go through.

### DIFF
--- a/language/oop5/static.xml
+++ b/language/oop5/static.xml
@@ -39,16 +39,9 @@
     not available inside the method declared as static.
    </para>
 
-   <caution>
-    <simpara>
-     In PHP 5, calling non-static methods statically generates an
-     <constant>E_STRICT</constant> level warning.
-    </simpara>
-   </caution>
-
    <warning>
     <simpara>
-     In PHP 7, calling non-static methods statically is deprecated, and will
+     Calling non-static methods statically is deprecated, and will
      generate an <constant>E_DEPRECATED</constant> warning. Support for
      calling non-static methods statically may be removed in the future.
     </simpara>
@@ -67,7 +60,7 @@ class Foo {
 
 Foo::aStaticMethod();
 $classname = 'Foo';
-$classname::aStaticMethod(); // As of PHP 5.3.0
+$classname::aStaticMethod();
 ?> 
 ]]>
      </programlisting>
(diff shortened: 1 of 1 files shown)


tiffany-taylor:
Changing this to just the static keyword page as I probably won't have the time to go over all of the OOP5 pages.

language/oop5/static.xml
```diff
@@ -84,14 +77,13 @@ $classname::aStaticMethod(); // As of PHP 5.3.0
 
    <para>
     Like any other PHP static variable, static properties may only be
-    initialized using a literal or constant before PHP 5.6; expressions are
-    not allowed. In PHP 5.6 and later, the same rules apply as &const;
-    expressions: some limited expressions are possible, provided they can be
-    evaluated at compile time.
+    initialized using a literal or constant; expressions are not allowed. 
+    The same rules apply as &const; expressions: some limited expressions 
+    are possible, provided they can be evaluated at compile time.
```
cmb69, line 82: That doesn't look quite right. The initialization restriction to "literal or constant" only applies prior to PHP 5.6.

cmb69 (changes requested):
Thanks, very nice, except for the commented issue.

language/oop5/static.xml
tiffany-taylor: Does this mean static properties can be initialized with an expression as well? Trying to see how I should wordsmith it.

cmb69: With a constant expression, i.e. "some limited expressions are possible, provided they can be evaluated at compile time"

tiffany-taylor: > Static variables have limited expression support, provided they can be evaluated at compile time.
How is that?

tiffany-taylor: Or maybe:
> Static variables have limited support for expressions, provided they can be evaluated at compile time.

cmb69: Well, actually I think we should define "constant expression" in a central place, and then just link to it. Currently, the most fitting definition (but IMO still a bit unspecific) is in https://www.php.net/manual/en/language.oop5.constants.php, but probably should be in https://www.php.net/manual/en/language.constants.php, but that page doesn't even mention global `const` definitions, which are explained in https://www.php.net/manual/en/language.constants.syntax.php. Total mess :(

afilina: Pinging @tiffany-taylor in case this fell through the cracks.

Girgias:
Closing as this should have been covered by #277

PR #166, opened 2020-10-28
Fix mysqli commit examples

kamil-tekiela:
PHP Manual doesn't show a working example of SQL transactions for mysqli. `commit()` page has a completely misleading example. 
While `rollback()` page was explaining the behavior of `rollback` it has been removed at the request of CMB to keep the examples aligned. 
A new example was provided for `autocommit()` page presenting how it can be used to control transactions.

Additionally `savepoint()` and `release_savepoint()` function received a simple description.

### DIFF
--- a/reference/mysqli/mysqli/commit.xml
+++ b/reference/mysqli/mysqli/commit.xml
@@ -59,6 +59,16 @@
   </para>
  </refsect1>
 
+ <refsect1 role="notes">
+  &reftitle.notes;
+  <note>
+   <para>
+    This function doesn't work with non transactional table types (like
+    MyISAM or ISAM).
+   </para>
+  </note>
+ </refsect1>
+
  <refsect1 role="changelog">
   &reftitle.changelog;
   <para>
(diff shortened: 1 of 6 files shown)


reference/mysqli/mysqli/commit.xml
```diff
@@ -86,76 +96,7 @@
...
-    exit();
-}
-
-/* close connection */
-mysqli_close($link);
-?>
-]]>
-   </programlisting>
-  </example>
+   See the <link linkend="mysqli.begin-transaction.example.basic">example</link>
```
cmb69, line 99: I think you need to enclose this in a  `<para>` element:
```suggestion
  <para>
   See the <link linkend="mysqli.begin-transaction.example.basic">example</link>.
  </para>
```
The same in rollback.xml.

cmb69:
Thanks! Applied with some minor tweaks (mostly regarding indentation; 1 space for XML; 4 spaces for code) as http://svn.php.net/viewvc?view=revision&revision=351087.

PR #167, opened 2020-10-28
MySQLi updating the documentation: There is no mysqli_warning_next

kamil-tekiela:
Removed invalid line from versions.xml


cmb69:
Well, actually [there is](https://www.php.net/manual/en/mysqli-warning.next.php). :)

cmb69:
Oops!

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351029.

kamil-tekiela:
> Well, actually [there is](https://www.php.net/manual/en/mysqli-warning.next.php). :)

That's `mysqli_warning::next` which is a method of the class and I mentioned a function. But I assume you realized that already :)

PR #168, opened 2020-10-29
Generate methodysnopses based on stubs for Zend functions and methods

kocsismate:


cmb69:
Thank you for working on this! Highly appreciated.

reference/apache/functions/apache-child-terminate.xml
```diff
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>bool</type><methodname>apache_child_terminate</methodname>
+   <type>void</type><methodname>apache_child_terminate</methodname>
```
cmb69, line 13: First change – first shock: such changes would need a changelog entry. Do you see any chance that this could be (partially) automated?

kocsismate: Haha, this is called scope creep :D But sure, it's a very good idea to also automate this! And I'm fairly certain that I can generate at least a basic entry. Can you tell me what kind of messages do you expect to see?

cmb69: If there is already a changelog section, something like the following might be a useful bare minimum:
````.xml
     <row>
      <entry>8.0.0</entry>
      <entry>
       The function/method signature has changed.
      </entry>
     </row>
````
In this particular case, the changelog entry could be "The return type is `void` now. Previously, it has been `bool`." And since the return type has changed, the [return values section](https://www.php.net/manual/en/function.apache-child-terminate.php#refsect1-function.apache-child-terminate-returnvalues) also needed to be changed to `&return.void;` (and possibly to clarify how it behaved previously). That might be a lot of work.

If there is no changelog section yet, than one would have to be inserted according to the [template file](http://svn.php.net/viewvc/phpdoc/doc-base/trunk/RFC/skeletons/function.xml?view=markup).

File: reference/errorfunc/functions/trigger-error.xml [hunk 2]
```diff
@@ -1,17 +1,17 @@
...
+<refentry xmlns="http://docbook.org/ns/docbook" xml:id="function.trigger-error">
  <refnamediv>
   <refname>trigger_error</refname>
   <refpurpose>Generates a user-level error/warning/notice message</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
   &reftitle.description;
-  <methodsynopsis>
+  <methodsynopsis role="procedural">
```
nikic, line 11: Why `role="procedural"`? There's an alias here, but it's between two functions, not a function and a method.

nikic:
I scrolled through this, and it looks good to me! Apart from the `role="procedural"`, I didn't spot any errors. @cmb69 Any thoughts?

I think we might want to do a post-processing str_replace to preserve that newline before the final comment, to reduce the amount of spurious diff.

reference/errorfunc/functions/trigger-error.xml
kocsismate: Yeah, this is only because I don't currently check if the alias is method. I'm not sure anymore why I didn't do this, but I can add this condition to the generator.

kocsismate:
> I think we might want to do a post-processing str_replace to preserve that newline before the final comment, to reduce the amount of spurious diff.

I'll try to add this to the code.

reference/errorfunc/functions/trigger-error.xml
cmb69: I'm not sure whether `role="procedural"` is used anywhere. Maybe drop it completely for now?

language/predefined/closure/bind.xml
```diff
@@ -12,11 +11,10 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <modifier>static</modifier> <type class="union"><type>Closure</type><type>false</type></type><methodname>Closure::bind</methodname>
```
cmb69, line 15: Hmm, these spaces (e.g. between public and static) might be necessary.

language/predefined/closure/bind.xml
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-
-<refentry xml:id="closure.bind" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+<refentry xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="closure.bind">
```
cmb69, line 3: It would be nice to keep the xml:id as first attribute, because it is the most important attribute when manually reviewing the file.

language/predefined/closure/bind.xml
cmb69: Ugh, they are:
Screenshot 2020-11-09 115315

kocsismate: Gosh, I wouldn't have imagined that this is the reason. :S Can't rendering be fixed instead? Or this is a completely normal behaviour?

cmb69: Well, there is some special handling to add whitespace between type and function name, but otherwise not rendering additional whitespace is normal. If it's too hard to add the spaces here, we may consider to add special cases in PhD. OTOH, whitespace makes the XML slightly better readable.

kocsismate: No, it's not difficult at all, I was just surprised that rendering is whitespace-dependent. That said, I'll re-add them in my next commit.

reference/errorfunc/functions/trigger-error.xml
kocsismate: I've dropped `role="procedural"`, but realized when testing against "ext/date", that we don't only use `oop` and `procedural` as role: e.g. https://www.php.net/manual/en/datetime.format.php uses the define the class names as roles as well.

language/predefined/closure/bind.xml
kocsismate: I've just readded the missing whitespaces.

language/predefined/closure/bind.xml
kocsismate: I used `preg_replace` to achieve this.

kocsismate:
> I think we might want to do a post-processing str_replace to preserve that newline before the final comment, to reduce the amount of spurious diff.

I'm afraid this is not that easy. Some pages do the opposite, and doesn't already have a newline before/after the root element tags, so if I tried `str_replace()`, I would add some new spurious diff. Arguably, much less than currently.

Do you insist on this change, or is it ok if these documents are normalized to get rid of the whitespaces in question?

Girgias:
> 
> 
> > I think we might want to do a post-processing str_replace to preserve that newline before the final comment, to reduce the amount of spurious diff.
> 
> I'm afraid this is not that easy. Some pages do the opposite, and doesn't already have a newline before/after the root element tags, so if I tried `str_replace()`, I would add some new spurious diff. Arguably, much less than currently.
> 
> Do you insist on this change, or is it ok if these documents are normalized to get rid of the whitespaces in question?

The issue is mostly with translations, as spurious diffs still forces translations to update the EN-Revision tag. So the less the better IMHO

kocsismate:
> Just looked at the current diff, as it seems files already need to be updated the whitespace diffs can be ignored by translations

Exactly! :)

reference/errorfunc/functions/trigger-error.xml
cmb69: I don't think that any of these rules are used by PhD, and I doubt that they're really useful for anything, but I maybe missing something.

kocsismate: I skimmed through php/phd, and neither I could find any use-case for the role attribute. Does that mean that I can just remove that part from the generator?

cmb69: Yeah, remove the `role`s. :)

nikic:
Is this one ready to go now?

language/predefined/closure/bind.xml
```diff
@@ -12,11 +11,10 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <modifier>static</modifier> <type class="union"><type>Closure</type><type>false</type></type><methodname>Closure::bind</methodname>
+   <modifier>public</modifier> <modifier>static</modifier> <type class="union"><type>Closure</type><type>null</type></type><methodname>Closure::bind</methodname>
```
cmb69, line 14: The returnvalues section has a `&return.falseforfailure;` which appears to be wrong.

language/predefined/closure/construct.xml
```diff
@@ -9,10 +8,10 @@
 
  <refsect1 role="description">
   &reftitle.description;
-  <methodsynopsis>
+  <constructorsynopsis>
```
cmb69, line 11: This is correct, but `class.closure` doesn't xinclude constructors, so we'd need something like:
````.diff
Index: closure.xml
===================================================================
--- closure.xml	(revision 351366)
+++ closure.xml	(working copy)
@@ -48,6 +48,9 @@
 <!-- }}} -->
     
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.closure')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[not(@role='procedural')])">
+     <xi:fallback />
+    </xi:include>
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.closure')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[1])" />
    </classsynopsis>
 <!-- }}} -->
````

File: language/predefined/errorexception/construct.xml [hunk 7]
```diff
@@ -13,10 +12,10 @@
...
-   <methodparam choice="opt"><type>int</type><parameter>severity</parameter><initializer>E_ERROR</initializer></methodparam>
-   <methodparam choice="opt"><type>string</type><parameter>filename</parameter><initializer>__FILE__</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>lineno</parameter><initializer>__LINE__</initializer></methodparam>
-   <methodparam choice="opt"><type>Exception</type><parameter>previous</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>severity</parameter><initializer><constant>E_ERROR</constant></initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>filename</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>line</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>Throwable</type><type>null</type></type><parameter>previous</parameter><initializer>&null;</initializer></methodparam>
```
cmb69, line 18: Assuming these 3 params have not been nullable before 8.0, this would need to be documented (maybe even separate sysnopses for PHP 7/8?).

cmb69 (commented):
Pretty good job, @kocsismate! :)

Whether we want to merge as is – well, how thorough do we like to be *right* *now*? I left some comments for consideration.

nikic:
> Whether we want to merge as is – well, how thorough do we like to be right now? I left some comments for consideration.

Please, let's not do any changes, apart from the automated signature changes and anything that is required to keep it building properly (the constructorsynopsis change in this case). We'll get completely bogged down otherwise.

kocsismate:
@cmb69 Can you please add this PR a label? Or alternatively, give me right to do so ^^

kocsismate:
I've just rebased to master, and added a few change log entries where doing so seemed necessary. I also noticed that the manual entries are yet to be added for `Stringable` and `InternalIterator`.

sgolemon (commented):
Generally okay with this (and appreciate the effort!), but in a few spots the parameter names are changed.  Some of these are clearly for the better, some are questionable or even for the worse.  We should probably update the stubs to make the names match documentation in most cases since the docs were (presumably) written by someone who put editorial thought into the name.

kocsismate:
@sgolemon Together with Nikita we made quite a lot of effort to improve parameter names. :) Which ones are you not satisfied with?

Speaking about doing parameter name changes in the stubs: unless a name is really broken, I don't think we can do any such changes because the arginfo structures are generated from them, so it would be a BC break for named arguments.

sgolemon:
> Which ones are you not satisfied with?

The one that stood out to me was `handle` -> `resource`.  But there are some like `class_name` to `class` which I found a little suspect.  `resource` and `class` are types and I have a small worry that they conflate themselves by default.

> Unless a parameter name is really broken, I don't think we can change them in the stubs because the arginfo structures are generated from the stubs, so we would introduce a BC break for named arguments.

Argh... good point. Fine... Leave as-is then. :(

sgolemon (approved):

kocsismate:
I know it's a large one.. But it would be very useful for users. Can you please review it, @cmb69 @kamil-tekiela ?

cmb69:
Thanks everybody!

PR #169, opened 2020-10-29
Delete remaining mysqli RPL docs removed in PHP 5.3

kamil-tekiela:
Taking the footsteps of commit https://github.com/php/doc-en/commit/faa2178d532d7cd2a97129b8a20ac8e3e0db9421 this PR removes remaining mysqli RPL docs which were all removed in PHP 5.3


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351045.

cmb69:
And with http://svn.php.net/viewvc?view=revision&revision=351046.

Would be really great to migrate to Git soon. :)

PR #172, opened 2020-10-31
Proper variadics instead of pseudo variadics

cmb69:
We markup variadic parameters with the `rep=repeat` standard DocBook
attribute of `<methodparam>`, and use proper variable names instead of
using the old pseudo variable name `...`.

---

This does not yet render as desired, but would rather need something like https://github.com/php/phd/pull/29.

### DIFF
--- a/reference/array/functions/array-udiff-assoc.xml
+++ b/reference/array/functions/array-udiff-assoc.xml
@@ -10,7 +10,7 @@
   <methodsynopsis>
    <type>array</type><methodname>array_udiff_assoc</methodname>
    <methodparam><type>array</type><parameter>array</parameter></methodparam>
-   <methodparam rep="repeat"><type>array</type><parameter>rest</parameter></methodparam>
+   <methodparam rep="repeat"><type>array</type><parameter>arrays</parameter></methodparam>
    <methodparam><type>callable</type><parameter>value_compare_func</parameter></methodparam>
   </methodsynopsis>
   <para>
@@ -38,7 +38,7 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>rest</parameter></term>
+     <term><parameter>arrays</parameter></term>
      <listitem>
       <para>
        Arrays to compare against.
--- a/reference/array/functions/array-udiff-assoc.xml
+++ b/reference/array/functions/array-udiff-assoc.xml
@@ -9,9 +9,8 @@
   &reftitle.description;
   <methodsynopsis>
    <type>array</type><methodname>array_udiff_assoc</methodname>
-   <methodparam><type>array</type><parameter>array1</parameter></methodparam>
-   <methodparam><type>array</type><parameter>array2</parameter></methodparam>
-   <methodparam choice="opt"><type>array</type><parameter>...</parameter></methodparam>
+   <methodparam><type>array</type><parameter>array</parameter></methodparam>
+   <methodparam rep="repeat"><type>array</type><parameter>rest</parameter></methodparam>
    <methodparam><type>callable</type><parameter>value_compare_func</parameter></methodparam>
   </methodsynopsis>
   <para>
@@ -31,7 +30,7 @@
   <para>
    <variablelist>
     <varlistentry>
-     <term><parameter>array1</parameter></term>
+     <term><parameter>array</parameter></term>
      <listitem>
       <para>
        The first array.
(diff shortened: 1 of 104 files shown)


reference/array/functions/array-udiff-assoc.xml
```diff
@@ -9,9 +9,8 @@
   &reftitle.description;
   <methodsynopsis>
    <type>array</type><methodname>array_udiff_assoc</methodname>
-   <methodparam><type>array</type><parameter>array1</parameter></methodparam>
-   <methodparam><type>array</type><parameter>array2</parameter></methodparam>
-   <methodparam choice="opt"><type>array</type><parameter>...</parameter></methodparam>
+   <methodparam><type>array</type><parameter>array</parameter></methodparam>
+   <methodparam rep="repeat"><type>array</type><parameter>rest</parameter></methodparam>
```
nikic, line 13: For the udiff/uintersect style functions, I'd also use `<parameter>arrays</parameter>`. The reason why it's called `...$rest` in stubs, it because it also includes the `$value_compare_func` parameter -- but docs allow us to actually keep it separate, so we can use `...$array`.

reference/array/functions/array-uintersect-uassoc.xml
```diff
@@ -40,6 +40,14 @@
...
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term><parameter>arrays</parameter></term>
+     <listitem>
+      <para>
+       Further arrays.
+      </para>
+     </listitem>
+    </varlistentry>
```
nikic, line 50: We should drop `$array2` here in favor of `$arrays`, for consistency with the other functions.

reference/array/functions/list.xml
```diff
@@ -9,8 +9,8 @@
   &reftitle.description;
   <methodsynopsis>
    <type>array</type><methodname>list</methodname>
-   <methodparam><type>mixed</type><parameter>var1</parameter></methodparam>
-   <methodparam rep="repeat" choice="opt"><type>mixed</type><parameter>...</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter>var</parameter></methodparam>
+   <methodparam rep="repeat" choice="opt"><type>mixed</type><parameter>vars</parameter></methodparam>
```
nikic, line 13: Hm ... this abuse of using methodsynopsis to document language features is problematic. For arrays at least `array(...$values)` is valid syntax, but `list($var, ...$vars)` is not.

reference/math/functions/max.xml
```diff
@@ -59,7 +59,7 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>...</parameter></term>
+     <term><parameter>values</parameter></term>
      <listitem>
       <para>
        Any <link linkend="language.operators.comparison">comparable</link>
```
nikic, line 65: In the line below value -> values?

reference/var/functions/debug-zval-dump.xml
```diff
@@ -28,6 +28,14 @@
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term><parameter>variables</parameter></term>
+     <listitem>
+      <para>
+
```
nikic, line 35: Missing description.

reference/array/functions/array-udiff-assoc.xml
cmb69: Ah, right!

reference/array/functions/array-uintersect-uassoc.xml
cmb69: Looks like there was a bug in the manual.

reference/var/functions/debug-zval-dump.xml
cmb69: Ah, right!

reference/math/functions/max.xml
cmb69: Thanks!

reference/array/functions/list.xml
cmb69: Agreed ([echo](https://www.php.net/manual/en/function.echo.php) is even worse, since it requires a paragraph to describe the use of parentheses). We should probably change that for `eval`, `echo`, `exit`, `print`, `empty`, `isset` and `unset` as well. Not sure how exactly, though.

nikic: In this particular case we should probably add a separate "array destructuring" page for both the list() and [] syntaxes, and then just leave a stub here. But that's an independent problem...

nikic (approved):
LG (once the phd change lands).

reference/array/functions/list.xml
cmb69: See https://github.com/php/doc-en/pull/145.

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351129.

PR #173, opened 2020-11-01
Follow up on #156

Girgias:
A follow up to #156 and some rewording/restructuring for some pages

### DIFF
--- a/language/basic-syntax.xml
+++ b/language/basic-syntax.xml
@@ -19,12 +19,39 @@
    </para>
 
    <para>
-    PHP also allows for short open tag <literal>&lt;?</literal> (which is
-    discouraged since it is only available if enabled using the
-    <link linkend="ini.short-open-tag">short_open_tag</link> &php.ini;
-    configuration file directive, or if PHP was configured with the
-    <option>--enable-short-tags</option> option).
+    <example>
+     <title>PHP Opening and Closing Tags</title>
+     <programlisting role="php">
+<![CDATA[
+1.  <?php echo 'if you want to serve PHP code in XHTML or XML documents,
+                use these tags'; ?>
+
+2.  You can use the short echo tag to <?= 'print this string' ?>.
+    It's equivalent to <?php echo 'print this string' ?>.
+
+3.  <? echo 'this code is within short tags, but will only work '.
+            'if short_open_tag is enabled'; ?>
+]]>
+     </programlisting>
+    </example>
    </para>
+
+   <para>
+    Short tags (example three) are available by default but can be disabled
+    either via the <link linkend="ini.short-open-tag">short_open_tag</link>
+    &php.ini; configuration file directive, or are disabled by default if
+    PHP is built with the <option>--disable-short-tags</option> configuration.
+   </para>
+   <para>
+    <note>
+     <para>
+      As short tags can be disabled it is recommened to only use the normal
+      tags (<code>&lt;?php ?&gt;</code> and <code>&lt;?= ?&gt;</code>) to
+      maximise compatibility.
+     </para>
+    </note>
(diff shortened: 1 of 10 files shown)


language/constants.xml
```diff
@@ -57,8 +58,8 @@ define("__FOO__", "something");
   </note>
 
   <simpara>
-   Like &link.superglobals;, the scope of a constant is global.  You 
-   can access constants anywhere in your script without regard to scope.  
+   Like &link.superglobals;, the scope of a constant is global.
+   Constants can be accessed from anywhere in a script without regard to scope.
    For more information on scope, read the manual section on
    <link linkend="language.variables.scope">variable scope</link>.
```
Girgias, line 64: This isn't totally correct due to visibility on class constants, but I suppose that's kinda irrelevant here?

language/basic-syntax.xml
```diff
@@ -95,52 +123,17 @@ echo "Last statement";
...
-    PHP is built with the <option>--disable-short-tags</option> configuration.
-   </para>
    <para>
     <note>
      <para>
-      As short tags can be disabled it is recommened to only use the normal
-      tags (<code>&lt;?php ?&gt;</code> and <code>&lt;?= ?&gt;</code>) to
-      maximise compatibility.
+      If PHP is embeded within XML or XHTML the normal PHP
+      <code>&lt;?php ?&gt;</code> must be used to remain complient
```
cmb69, line 130:
```suggestion
      <code>&lt;?php ?&gt;</code> must be used to remain compliant
```

language/constants.xml
```diff
@@ -8,7 +8,7 @@
    suggests, that value cannot change during the execution of the
    script (except for <link linkend="language.constants.predefined">
    magic constants</link>, which aren't actually constants).
-   A constant is case-sensitive by default. By convention, constant 
+   Constant are case-sensitive. By convention, constant
```
cmb69, line 11: Well, in PHP 7 it is still possible to have case-insensitive constants (`define(…, …, true)` and internal constants can be as well). Also should be "constants are" (plural).

language/constants.xml
cmb69: That paragraph might not have been updated regarding constant visibility specifiers. Might make sense to add a note ("(if visible)" or such).

language/types/pseudo-types.xml
```diff
@@ -30,7 +31,7 @@
   </para>
 
  </sect2>
-
+<!-- TODO: Proper page? -->
```
cmb69, line 34: Yes; but that can be another commit/PR.

language/constants.xml
Girgias: I would have assumed the note below to indicate that prior to PHP 8 constants defined using ``define()`` can be case insensitive would be sufficient, but can reword if needed.

cmb69: Ah, I missed that note. All fine then. :)

Girgias: I moved the note up as it was on the next page in theory. :)

language/constants.xml
Girgias: Note added below, is that ok?

language/constants.xml
```diff
@@ -303,8 +327,7 @@ echo ANIMALS[1]; // outputs "cat"
        <row xml:id="constant.coloncolonclass">
         <entry><constant>ClassName::class</constant></entry>
```
cmb69, line 328: Maybe
```suggestion
        <entry><constant><replaceable>ClassName</replaceable>::class</constant></entry>
```
Not sure, though, whether that would "work" with PhD.

cmb69 (approved):
Thanks! Looks good to me.

PR #174, opened 2020-11-01
Split type declaration into its own page

Girgias:
Also adds a section about union types.


language/types/declarations.xml
```diff
@@ -0,0 +1,754 @@
...
+     <row>
+      <entry><literal>mixed</literal></entry>
+      <entry>
+       The value can be any value.
+      </entry>
+      <entry>PHP 8.0.0</entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
```
cmb69, line 129: Why are some of the types in this table marked up as `<type>` and others as `<literal>`? Doesn't look reasonable to me. (and yes, I'm aware that this table has just been moved)

language/types/declarations.xml
```diff
@@ -0,0 +1,754 @@
...
+   <title>false pseudo-type</title>
+   <para>
+    The <literal>false</literal> literal type is supported as part of unions,
+    and is included as for historical reasons many internal functions return
+    <literal>false</literal> instead of <literal>null</literal> for failures.
+    A classic example of such a function is <function>strpos</function>.
+    such that <literal>T1|T2|null</literal> can be used to create a nullable union.
+    The existing <literal>?T</literal> notation is considered a shorthand
+    for the common case of <literal>T|null</literal>.
+   </para>
```
cmb69, line 370: The 3 last lines appears to superfluous (copy-paste issue?)

language/types/declarations.xml
```diff
@@ -0,0 +1,754 @@
...
+     As such, all of <literal>false</literal>, <literal>false|null</literal>
+     and <literal>?false</literal> are not permitted.
+    </simpara>
+   </warning>
+   <warning>
+    <simpara>
+     The <literal>true</literal> literal type does <emphasis>not</emphasis>
+     exist.
+    </simpara>
+   </warning>
```
cmb69, line 385: Maybe `<warning>` is a bit too strong for both cases?

language/types/declarations.xml
```diff
@@ -0,0 +1,754 @@
...
+    <itemizedlist>
+     <listitem>
+      <simpara>
+       Each name-resolved type may only occur once. Types such as
+       <literal>int|string|INT</literal> result in an error.
+      </simpara>
+     </listitem>
+     <listitem>
+      <simpara>
+       If <literal>bool</literal> is used, <literal>false</literal> cannot be used additionally.
```
cmb69, line 404:
```suggestion
       If <type>bool</type> is used, <type>false</type> cannot be used additionally.
```
PhD support for `false` should be added explicitly, but this should work reasonably well right now.

cmb69 (commented):
Thanks for working on this! Moving the type declaration documentation to an appropriate place is long overdue, and having union types documented help to simplify the [PHP 8.0 migration guide](https://github.com/php/doc-en/pull/170) a bit.

language/types/declarations.xml
```diff
@@ -0,0 +1,754 @@
...
+    </informalexample>
+   </para>
+  </sect3>
+
+ </sect2>
+
+ <sect2 xml:id="language.types.declarations.return-only">
+  <title>Return only types</title>
+
+  <sect3 xml:id="language.types.declarations.void">
```
cmb69, line 461: Would it make sense to somehow unify/merge this section with https://github.com/php/doc-en/blob/cdaea0421544885f02ff3d36bd203dc01b78299e/language/types/pseudo-types.xml#L34-L43

and have `void`'s own section (`&language.types.void;`)? I'd really like to drop the pseudo-types section; `mixed` is a built-in type as well now, and `number` could be replaced with the respective union type; and variadics are being worked on already.

cmb69:
Also, before actually removing a page from the manual ([language/oop5/typehinting.xml](https://www.php.net/manual/en/language.oop5.typehinting.php) in this case), it might be sensible to cater to existing user notes (i.e. integrate or remove).

language/types/declarations.xml
Girgias: I indeed copy pasted this from the Union type RFC, don't know if it's superfluous, but might make more sense as a note.

language/types/declarations.xml
Girgias: Should I use ``<caution>`` instead?

language/types/declarations.xml
Girgias: Seems reasonable, I suppose I could add a note which goes along the lines:
```xml
<note>
 <para>
   The PHP documentation uses <literal>void</literal> in an argument list to convey that a function doesn't accept any arguments.
 </para>
</note>
```

Girgias:
> 
> 
> Also, before actually removing a page from the manual ([language/oop5/typehinting.xml](https://www.php.net/manual/en/language.oop5.typehinting.php) in this case), it might be sensible to cater to existing user notes (i.e. integrate or remove).

The notes all seems updated and irrelevant to some extend by just being benchmarks (I noted that Nikita also removed a bunch just now)

language/types/declarations.xml
cmb69: IMO, yes. :)

language/types/declarations.xml
```diff
@@ -0,0 +1,817 @@
...
+  <para>
+   Type declarations were also known as type hints in previous versions of PHP.
+  </para>
+ </note>
+
+ <para>
+  To specify a type declaration, the type name should be added before the
+  parameter name. The declaration can be made to accept &null; values if
+  the default value of the parameter is set to &null;.
+ </para>
```
Girgias, line 23: This should probably dropped now and/or integrated in some other paragraph

language/types/declarations.xml
```diff
@@ -0,0 +1,817 @@
...
+    function test(boolean $param) {}
+    test(true);
+?>
+]]>
+     </programlisting>
+     &example.outputs;
+     <screen>
+<![CDATA[
+    Fatal error: Uncaught TypeError: Argument 1 passed to test() must be an instance of boolean, boolean given, called in - on line 1 and defined in -:1
+]]>
```
Girgias, line 154: I think the test output needs to be updated, as I think it now says "bool given"

language/types/declarations.xml
```diff
@@ -0,0 +1,817 @@
...
+    such that <literal>T1|T2|null</literal> can be used to create a nullable union.
+    The existing <literal>?T</literal> notation is considered a shorthand
+    for the common case of <literal>T|null</literal>.
+   </para>
+   
+   <warning>
+    <simpara>
+     <literal>null</literal> cannot be used as a standalone type.
+    </simpara>
+   </warning>
```
Girgias, line 390: Should I mark this as caution too now that I made `false` a caution section?

language/types/declarations.xml
```diff
@@ -0,0 +1,817 @@
...
+  </sect3>
+ </sect2>
+
+ <sect2 xml:id="language.types.declarations.strict">
+  <title>Strict typing</title>
+
+  <para>
+   By default, PHP will coerce values of the wrong type into the expected
+   scalar type declaration if possible. For example, a function that is given
+   an <type>integer</type> for a parameter that expects a <type>string</type>
```
Girgias, line 520:
```suggestion
   an <type>int</type> for a parameter that expects a <type>string</type>
```
Should this use the canonical form or not?

language/types/declarations.xml
cmb69: IMO, yes.

language/types/declarations.xml
cmb69: IMO, yes.

language/types/declarations.xml
nikic: Maybe we should drop that as well? Using `void` to indicate no arguments is a C-ism, I don't think there's any value in doing that in the PHP manual.

cmb69: See https://github.com/php/phd/pull/32

language/types/declarations.xml
```diff
@@ -0,0 +1,817 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<sect1 xml:id="language.types.declarations">
+ <title>Type declarations</title>
+
+ <para>
+  Type declarations can be added to function arguments, return values,
+  and, as of PHP 7.4.0, class properties. And they ensure that the value
```
nikic, line 8:
```suggestion
  and, as of PHP 7.4.0, class properties. They ensure that the value
```

language/types/declarations.xml
```diff
@@ -0,0 +1,817 @@
...
+  is of the specified type at call time, otherwise a
+  <classname>TypeError</classname> is thrown.
+ </para>
+
+ <note>
+  <para>
+   Type declarations were also known as type hints in previous versions of PHP.
+  </para>
```
nikic, line 16: Let's make Levi happy and drop this sentence :)

cmb69: Or may be add that the term "type hints" is frowned upon?

language/types/declarations.xml
```diff
@@ -0,0 +1,819 @@
...
+]]>
+    </programlisting>
+   </example>
+  </para>
+
+  <note>
+   <para>
+    Prior to PHP 7.1.0, it was possible to achieve nullable arguments by making
+    <literal>null</literal> the default value.
+    This is not recommended as this breaks during inheritance.
```
Girgias, line 338: Should an example about this be added?

language/types/declarations.xml
```diff
@@ -0,0 +1,819 @@
...
+42.1  --> 42          // float compatible with int
+1e100 --> "1.0E+100"  // float too large for int type, fall back to string
+INF   --> "INF"       // float too large for int type, fall back to string
+true  --> 1           // bool compatible with int
+[]    --> TypeError   // array not compatible with int or string
+
+// int|float|bool
+"45"    --> 45        // int numeric string
+"45.0"  --> 45.0      // float numeric string
+"45X"   --> 45 + Notice: Non well formed numeric string
```
Girgias, line 719: That wouldn't work any more right @nikic ?

Maybe this should be made into "proper" examples

Girgias:
So I still don't know what to do with the two examples in the temporary "Misc" section, any ideas?

language/types/declarations.xml
cmb69: Yes, could go to language.types.declarations.nullable.

language/types/declarations.xml
cmb69: I don't think that's necessary.

language/types/declarations.xml
cmb69: "Proper" examples (using example.outputs and friends) can be hard to read for a lot of output. Maybe split out the cases which have different behavior for different PHP versions?

language/types/declarations.xml
```diff
@@ -0,0 +1,818 @@
...
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+int(3)
+Error: sum(): Argument #1 ($a) must be of type int, float given, called in - on line 10
+]]>
+    </screen>
+   </example>
```
cmb69, line 794: These two examples are PHP 8 only; at least that should be mentioned. Also use `&example.outputs.80;` (needs to be added to language-snippets.ent).

language/types/declarations.xml
Girgias: Well as it's about union types it can only be PHP 8.0.

cmb69: Good point!

language/types/declarations.xml
```diff
@@ -0,0 +1,812 @@
...
+   type under PHPs existing type checking semantics, then the type is chosen.
+   Otherwise the next type is tried.
+  </para>
+
+  <caution>
+   <para>
+    As an exception, if the value is a string and both int and float are part
+    of the union, the preferred type is determined by the existing
+    “numeric string” semantics.
+    For example, for <literal>“42”</literal> <type>int</type> is chosen,
```
cmb69, line 679: Not quite sure about using proper quotes in the text, but the `“42”` is definitely misleading.

Girgias: Oh yeah I don't know how that slipped in (probably from the copy and paste). I'll fix this before merging.

PR #175, opened 2020-11-01
Replace "number" type with built-in types

cmb69:
There is really no need for a pseudo type here; we can either use the
`int|float` union, or spell out the abbreviation, instead.

---

See also https://github.com/php/doc-en/pull/174#discussion_r515682273.

Oh, and of course, if we wanted, we could define a `type.number` entity for brevity.

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -1573,8 +1573,8 @@ echo $a <=> $b; // 1
        </entry>
       </row>
       <row>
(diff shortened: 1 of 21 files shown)


nikic (approved):
Nice!

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351123.

PR #176, opened 2020-11-02
Use canonical type names

cmb69:


nikic (approved):

cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351133.

PR #177, opened 2020-11-03
Fix return types

cmb69:
Functions which may return `false` according to the return values
section, should also reflect that in the signature.


nikic:
Wouldn't this be better covered by the sync with the stubs?

cmb69:
> Wouldn't this be better covered by the sync with the stubs?

I'm a bit worried that this sync may take a lot of manual work, so doing some upfront work might be reasonable. And there are a lot of extensions which are not bundled, and not all such extension may have stubs.

PR #178, opened 2020-11-04
Updating mysqli docs: mysqli::__construct

kamil-tekiela:
This is yet another PR attempting to modernize and explain better mysqli manual. 
Changes:
1. Provided new code examples.  
First, examples with exception mode enabled and setting the correct charset (the typo in output was fixed, but also changed to "or similar"). Then, a modernized example of extending mysqli class. Followed by the examples of manual error checking for completeness. 
2. Added a new sentence to the first parameter "The TCP/IP protocol is used if a host name and port number are provided together e.g. `localhost:3308`"
3. Return section now split up for all 3 "variants". Constructor only returns an instance, function returns an instance or false, and the method returns null or false. 
4. Added errors section. I was not sure about this, but I decided that it is very relevant for people who would like to catch the exception and are looking for its name. 
5. Changed the wording of some sentences for easier reading.
6. Removed the outdated and unnecessary example from `mysqli_connect` page. We should either keep a separate documentation page for each variant (e.g. like we do with `mysqli_real_connect`) or maintain all of them together and keep `mysqli_connect` page as an alias stub page only. However, given that the only difference is the return value I added a note pointing it out.

### DIFF
--- a/reference/mysqli/functions/mysqli-connect.xml
+++ b/reference/mysqli/functions/mysqli-connect.xml
@@ -11,44 +11,13 @@
   <para>
    &info.function.alias; <methodname>mysqli::__construct</methodname>
   </para>
-  <para>
-   Although the <methodname>mysqli::__construct</methodname> documentation also includes
-   procedural examples that use the <function>mysqli_connect</function> function, here
-   is a short example:
-  </para>
- </refsect1>
-
- <refsect1 role="examples">
-  &reftitle.examples;
-  <example>
-   <title><function>mysqli_connect</function> example</title>
-   <programlisting role="php">
-<![CDATA[
-<?php
-$link = mysqli_connect("127.0.0.1", "my_user", "my_password", "my_db");
-
-if (!$link) {
-    echo "Error: Unable to connect to MySQL." . PHP_EOL;
-    echo "Debugging errno: " . mysqli_connect_errno() . PHP_EOL;
-    echo "Debugging error: " . mysqli_connect_error() . PHP_EOL;
-    exit;
-}
-
-echo "Success: A proper connection to MySQL was made! The my_db database is great." . PHP_EOL;
-echo "Host information: " . mysqli_get_host_info($link) . PHP_EOL;
-
-mysqli_close($link);
-?>
-]]>
-   </programlisting>
-   &examples.outputs.similar;
-   <screen>
-<![CDATA[
-Success: A proper connection to MySQL was made! The my_db database is great.
-Host information: localhost via TCP/IP
-]]>
-   </screen>
-  </example>
+  <note>
+   <para>
+    If mysqli exception mode is not enabled and a connection fails 
+    then <function>mysqli_connect</function> returns &false; instead of an object. 
+    <function>mysqli_connect_error</function> function can be used to fetch the connection error.
+   </para>
+  </note>
  </refsect1>
 </refentry>
 
--- a/reference/mysqli/functions/mysqli-connect.xml
+++ b/reference/mysqli/functions/mysqli-connect.xml
@@ -13,7 +13,7 @@
   </para>
   <note>
    <para>
-    If mysqli exception mode is not enabled and a connection fails 
+    If mysqli exception mode is not enabled and a connection fails, 
     then <function>mysqli_connect</function> returns &false; instead of an object. 
     <function>mysqli_connect_error</function> function can be used to fetch the connection error.
    </para>
@@ -41,4 +41,3 @@ vim600: syn=xml fen fdm=syntax fdl=2 si
 vim: et tw=78 syn=sgml
 vi: ts=1 sw=1
 -->
-
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -158,7 +158,7 @@ mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 
 $mysqli = new mysqli('localhost', 'my_user', 'my_password', 'my_db');
 
-/* Setting the proper charset after establishing a connection */
+/* Set the desired charset after establishing a connection */
 $mysqli->set_charset('utf8mb4');
 
 printf("Success... %s\n", $mysqli->host_info);
@@ -174,7 +174,7 @@ mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 
 $mysqli = mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
 
-/* Setting the proper charset after establishing a connection */
+/* Set the desired charset after establishing a connection */
 mysqli_set_charset($mysqli, 'utf8mb4');
 
 printf("Success... %s\n", mysqli_get_host_info($mysqli));
@@ -207,7 +207,7 @@ $db = new MyMysqli('localhost', 'my_user', 'my_password', 'my_db', 3306, null, '
   </example>
   <example>
    <title>Manual error handling</title>
-   <para>If error reporting is disabled, the developer is responsible for failure checking and their proper handling</para>
+   <para>If error reporting is disabled, the developer is responsible for checking and handling failures</para>
    <para>&style.oop;</para>
    <programlisting role="php">
 <![CDATA[
@@ -219,7 +219,7 @@ if ($mysqli->connect_errno) {
 	exit;
 }
 
-/* Setting the proper charset after establishing a connection */
+/* Set the desired charset after establishing a connection */
 $mysqli->set_charset('utf8mb4');
 if ($mysqli->errno) {
 	error_log($mysqli->error);
@@ -238,7 +238,7 @@ if (mysqli_connect_errno()) {
 	exit;
 }
 
-/* Setting the proper charset after establishing a connection */
+/* Set the desired charset after establishing a connection */
 mysqli_set_charset($mysqli, 'utf8mb4');
 if (mysqli_errno($mysqli)) {
 	error_log(mysqli_error($mysqli));
@@ -254,8 +254,8 @@ if (mysqli_errno($mysqli)) {
   &mysqli.charset.note;
   <note>
    <para>
-    &style.oop; only: If the connection fails an object is still returned. To check 
-    if the connection failed, use either the 
+    &style.oop; only: If the connection fails, an object is still returned. To check 
+    whether the connection failed, use either the 
     <function>mysqli_connect_error</function> function or the <link 
     linkend="mysqli.connect-error">mysqli->connect_error</link> property as in 
     the preceding examples.
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -193,7 +193,7 @@ Success... localhost via TCP/IP
 <![CDATA[
 <?php
 
-class MyMysqli extends mysqli {
+class FooMysqli extends mysqli {
 	public function __construct($host, $user, $pass, $db, $port, $socket, $charset) {
 		mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 		parent::__construct($host, $user, $pass, $db, $port, $socket);
@@ -201,7 +201,7 @@ class MyMysqli extends mysqli {
 	}
 }
 
-$db = new MyMysqli('localhost', 'my_user', 'my_password', 'my_db', 3306, null, 'utf8mb4');
+$db = new FooMysqli('localhost', 'my_user', 'my_password', 'my_db', 3306, null, 'utf8mb4');
 ]]>
    </programlisting>
   </example>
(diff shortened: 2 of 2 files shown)


reference/mysqli/functions/mysqli-connect.xml
```diff
@@ -11,44 +11,13 @@
...
-   Although the <methodname>mysqli::__construct</methodname> documentation also includes
-   procedural examples that use the <function>mysqli_connect</function> function, here
-   is a short example:
-  </para>
- </refsect1>
-
- <refsect1 role="examples">
-  &reftitle.examples;
-  <example>
-   <title><function>mysqli_connect</function> example</title>
```
afilina, line 24: There is indeed no point in duplicating the example if this method is an alias.

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
- * compatibility with PHP versions prior to 5.2.9 and 5.3.0.
- */
-if (mysqli_connect_error()) {
-    die('Connect Error (' . mysqli_connect_errno() . ') '
-            . mysqli_connect_error());
-}
+$mysqli = new mysqli('localhost', 'my_user', 'my_password', 'my_db');
 
-echo 'Success... ' . $mysqli->host_info . "\n";
+/* Setting the proper charset after establishing a connection */
```
afilina, line 161:
```suggestion
/* Set the desired charset after establishing a connection */
```

I would avoid referring to it as the "proper" charset because it depends, but it's good to nudge people towards it in the example.

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
-    }
-}
+/* You should enable error reporting for mysqli before attempting to make a connection */
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 
-$db = new foo_mysqli('localhost', 'my_user', 'my_password', 'my_db');
+$mysqli = mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
 
-echo 'Success... ' . $db->host_info . "\n";
+/* Setting the proper charset after establishing a connection */
```
afilina, line 177:
```suggestion
/* Set the desired charset after establishing a connection */
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
 
-mysqli_close($link);
-?>
+$mysqli = @new mysqli('localhost', 'my_user', 'my_password', 'my_db');
+if ($mysqli->connect_errno) {
+	error_log($mysqli->connect_error);
+	exit;
+}
+
+/* Setting the proper charset after establishing a connection */
```
afilina, line 222:
```suggestion
/* Set the desired charset after establishing a connection */
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
-Success... MySQL host info: localhost via TCP/IP
+<?php
+
+$mysqli = @mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
+if (mysqli_connect_errno()) {
+	error_log(mysqli_connect_error());
+	exit;
+}
+
+/* Setting the proper charset after establishing a connection */
```
afilina, line 241:
```suggestion
/* Set the desired charset after establishing a connection */
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
+   <title>Extending mysqli class</title>
    <programlisting role="php">
 <![CDATA[
 <?php
-$link = mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
 
-if (!$link) {
-    die('Connect Error (' . mysqli_connect_errno() . ') '
-            . mysqli_connect_error());
+class MyMysqli extends mysqli {
```
afilina, line 196:
```suggestion
class FooMysqli extends mysqli {
```

"MyMy" can be a bit confusing to read.

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
+	public function __construct($host, $user, $pass, $db, $port, $socket, $charset) {
+		mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
+		parent::__construct($host, $user, $pass, $db, $port, $socket);
+		$this->set_charset($charset);
+	}
 }
 
-echo 'Success... ' . mysqli_get_host_info($link) . "\n";
+$db = new MyMysqli('localhost', 'my_user', 'my_password', 'my_db', 3306, null, 'utf8mb4');
```
afilina, line 204:
```suggestion
$db = new FooMysqli('localhost', 'my_user', 'my_password', 'my_db', 3306, null, 'utf8mb4');
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
    <programlisting role="php">
 <![CDATA[
 <?php
-$link = mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
 
-if (!$link) {
-    die('Connect Error (' . mysqli_connect_errno() . ') '
-            . mysqli_connect_error());
+class MyMysqli extends mysqli {
+	public function __construct($host, $user, $pass, $db, $port, $socket, $charset) {
```
afilina, line 197: I don't think that this is the best example. We shouldn't change the signature of the method that we're overloading. It's not a good practice. We can "overload" a constructor using a static constructor. Example: https://3v4l.org/37ToN

If static constructors make the example too complex for beginners, perhaps we can add an example that doesn't require overloading of the constructor.

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
 }
 
-echo 'Success... ' . mysqli_get_host_info($link) . "\n";
+$db = new MyMysqli('localhost', 'my_user', 'my_password', 'my_db', 3306, null, 'utf8mb4');
+]]>
+   </programlisting>
+  </example>
+  <example>
+   <title>Manual error handling</title>
+   <para>If error reporting is disabled, the developer is responsible for failure checking and their proper handling</para>
```
afilina, line 210:
```suggestion
   <para>If error reporting is disabled, the developer is responsible for checking and handling failures</para>
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
+<![CDATA[
+<?php
 
-mysqli_close($link);
-?>
+$mysqli = @new mysqli('localhost', 'my_user', 'my_password', 'my_db');
+if ($mysqli->connect_errno) {
+	error_log($mysqli->connect_error);
+	exit;
+}
```
afilina, line 220:
```suggestion
mysqli_reporting(MYSQLI_REPORT_OFF);
$mysqli = new mysqli('localhost', 'my_user', 'my_password', 'my_db');
if ($mysqli->connect_errno) {
	throw new RuntimeException('mysqli connection error: ' . $mysqli->connect_error);
}
```

1. Explicitly turn off error reporting instead of nudging people towards`@`. It also makes the example more similar to the others where `mysqli_reporting` is used.
2. Throw an exception instead of calling `exit;`, which avoids all sorts of problems down the line, especially with unit tests.

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
+
+/* Setting the proper charset after establishing a connection */
+$mysqli->set_charset('utf8mb4');
+if ($mysqli->errno) {
+	error_log($mysqli->error);
+	exit;
```
afilina, line 226:
```suggestion
 throw new RuntimeException('mysqli error: ' . $mysqli->error);
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
+   <programlisting role="php">
 <![CDATA[
-Success... MySQL host info: localhost via TCP/IP
+<?php
+
+$mysqli = @mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
+if (mysqli_connect_errno()) {
+	error_log(mysqli_connect_error());
+	exit;
+}
```
afilina, line 239:
```suggestion
mysqli_reporting(MYSQLI_REPORT_OFF);
$mysqli = mysqli_connect('localhost', 'my_user', 'my_password', 'my_db');
if (mysqli_connect_errno()) {
	throw new RuntimeException('mysqli connection error: ' . mysqli_connect_error());
}
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -137,79 +152,100 @@
...
+
+/* Setting the proper charset after establishing a connection */
+mysqli_set_charset($mysqli, 'utf8mb4');
+if (mysqli_errno($mysqli)) {
+	error_log(mysqli_error($mysqli));
+	exit;
```
afilina, line 245:
```suggestion
	throw new RuntimeException('mysqli error: ' . mysqli_error($mysqli));
```

reference/mysqli/mysqli/construct.xml
```diff
@@ -218,8 +254,8 @@ Success... MySQL host info: localhost via TCP/IP
   &mysqli.charset.note;
   <note>
    <para>
-    OO syntax only: If a connection fails an object is still returned. To check 
-    if the connection failed then use either the 
+    &style.oop; only: If the connection fails an object is still returned. To check 
+    if the connection failed, use either the
```
afilina, line 258:
```suggestion
    &style.oop; only: If the connection fails, an object is still returned. To check 
    whether the connection failed, use either the 
```

afilina (changes requested):
Thank you very much for all the improvements!
I made multiple suggestions for your consideration.

reference/mysqli/mysqli/construct.xml
```diff
@@ -52,10 +52,10 @@
...
      <listitem>
       <para>
-       Can be either a host name or an IP address. Passing the &null; value
-       or the string "localhost" to this parameter, the local host is
-       assumed. When possible, pipes will be used instead of the TCP/IP
-       protocol.
+       Can be either a host name or an IP address. The local host is 
+       assumed when passing the &null; value or the string "localhost" to this parameter. 
+       When possible, pipes will be used instead of the TCP/IP protocol. 
+       The TCP/IP protocol is used if a host name and port number are provided together e.g. <literal>localhost:3308</literal>.
```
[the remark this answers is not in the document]
kamil-tekiela: Sure, it doesn't matter much here. The idea was just to show the difference. 3307/3308 is also a commonly used port e.g. on my own server, the default port is actually 3307 for MariaDB. I understand though that PHP has port 3306 set as the default port which means that if users like me want to connect to a server listening on port 3307 or 3308 they have to provide the port number which makes this a common scenario.

reference/mysqli/mysqli/construct.xml
```diff
@@ -124,9 +124,24 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an object which represents the connection to a MySQL Server,
+   <methodname>mysqli::__construct</methodname> always returns an object which represents the connection to a MySQL Server,
```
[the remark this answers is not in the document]
kamil-tekiela: There are 3 methods/functions described on this page. It needs to be explicitly stated how each of them differs. Each one has different possible return values.

reference/mysqli/mysqli/construct.xml
kamil-tekiela: This is a good suggestion, but the idea here was to silence all errors including warnings. I think many people are used to this kind of error suppression, although I agree this is a terrible practice. 

If we are to make it more explicit, then do you mind if I prepend `error_reporting(0);` to the example? It will make it clear that PHP errors go hand in hand with mysqli errors.

reference/mysqli/mysqli/construct.xml
kamil-tekiela: I don't like my example, but I am not a big fan of yours, either. I don't think it's a good example of polymorphism. I would rather do what you suggested at the end. We should come up with a better example of extending mysqli class that doesn't involve changing constructor's signature. I haven't got anything in mind at the moment. 

The main thing is that both my and your example is much better than the existing one. :)

Another thing we could do is simply remove that example, but I would prefer to avoid that. We could instead use the [example I have been using for teaching purposes](https://stackoverflow.com/a/64884939/1839439) or something similar. I am open to ideas.

reference/mysqli/mysqli/construct.xml
afilina: Yes, that's a reasonable alternative.

afilina:
I reviewed the latest changes. LGTM.

cmb69:
Thank you!

PR #179, opened 2020-11-05
add CRYPT_SHA256 and CRYPT_SHA512 constants

webmaster777:
While the description is available at the `crypt()` function since a3e8136e4e9ed855110b15a65d26f52bb34780cc, it was never added to the list of predefined constants.

### DIFF
--- a/reference/strings/constants.xml
+++ b/reference/strings/constants.xml
@@ -22,7 +22,7 @@
    </term>
    <listitem>
     <simpara>
-
+     Indicates whether standard DES-based hashes are supported in <function>crypt</function>. Always <literal>1</literal>.
     </simpara>
    </listitem>
   </varlistentry>
@@ -33,7 +33,7 @@
    </term>
    <listitem>
     <simpara>
-
+     Indicates whether extended DES-based hashes are supported in <function>crypt</function>. Always <literal>1</literal>.
     </simpara>
    </listitem>
   </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/strings/constants.xml
```diff
@@ -59,6 +59,28 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.crypt-sha256">
+   <term>
+    <constant>CRYPT_SHA256</constant>
+     <type>int</type>
+   </term>
+   <listitem>
+    <simpara>
+
```
cmb69, line 69:
```suggestion
     To be used with <function>crypt</function>.
```
Also for the other `CRYPT_*` constants.

cmb69 (commented):
Thanks! Very nice.

reference/strings/constants.xml
nikic: These constants are not actually used with crypt(). Rather they indicate whether a certain algorithm is available.

To make matters worse, all CRYPT_* algorithms are always available, so these constants are completely useless really :)

cmb69: Ah, right, thanks! I think we still should document that; something like:
```suggestion
Indicates whether SHA-256 hashes are supported. Always <literal>1</literal>.
```
And deprecate these constants for 8.1.

webmaster777: I don't know where the deprecation should be issued. Can you take care of that @nikic, @cmb69 ?

Girgias: Deprecation would need to go through the RFC process first so documenting that is not possible atm.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351278.

PR #180, opened 2020-11-06
Updating mysqli: fixing mysqli_report

kamil-tekiela:
Couple of much needed fixes:

1. Changed the wording and the explanation to make it clearer what this function actually does.  
Remove "A function helpful in improving queries during code development and testing." which might be true but is also unnecessary and misleading. Besides, in OO style it is not a function...
2. This function [never returns `false`.](https://github.com/php/php-src/blob/master/ext/mysqli/mysqli_report.c#L34) 
3. Fixed broken examples. Come on, this is just a disgrace that a PHP manual has broken examples. 
4. Added a third example showing the most common use case of enabling errors without bad index errors. 
5. Removed links to `mysqli_debug` and `mysqli_dump_debug_info`. They're not really related and having them in See also can be quite confusing.

### DIFF
--- a/reference/mysqli/mysqli_driver/report-mode.xml
+++ b/reference/mysqli/mysqli_driver/report-mode.xml
@@ -99,7 +99,7 @@ $driver = new mysqli_driver();
 $driver->report_mode = MYSQLI_REPORT_ALL;
 
 try {
-	/* if the connection fails an exception will be thrown */
+	/* if the connection fails, a mysqli_sql_exception will be thrown */
 	$mysqli = new mysqli("localhost", "my_user", "my_password", "my_db");
 
 	/* this query should report an error */
@@ -146,7 +146,7 @@ try {
 mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 
 try {
-	/* if the connection fails an exception will be thrown */
+	/* if the connection fails, a mysqli_sql_exception will be thrown */
 	$mysqli = new mysqli("localhost", "my_user", "my_password", "my_db");
 
 	/* this query should report an error */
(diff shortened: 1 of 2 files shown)


reference/mysqli/mysqli_driver/report-mode.xml
```diff
@@ -158,8 +166,6 @@ mysqli_close($link);
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member><function>mysqli_debug</function></member>
-    <member><function>mysqli_dump_debug_info</function></member>
```
afilina, line 162: I'd argue that those are still related, because error reporting and debugging are within the same area of interests.

afilina (changes requested):
Added a few suggestions. Otherwise, LGTM.

reference/mysqli/mysqli_driver/report-mode.xml
kamil-tekiela: I can see where you are coming from, but that is actually something that I would like to prevent. I don't want people confusing these two functions with error reporting. `mysqli_debug` is meant for debugging of mysqli internal code, not userland code. `mysqli_dump_debug_info` triggers a debug function in the MySQL C API, which can be helpful to debug problems but is not part of error reporting. What I am proposing is not removing them from the manual, just removing a link from the `mysqli_report` page. I want to keep the links in the "See also" section relevant only for error reporting. MySQLi debugging is described on other pages in the manual.

afilina (approved):

cmb69:
Thank you!

Please note that we do not document the return value in case of failing zend_parse_parameters(); this is already covered by the note on https://www.php.net/manual/en/functions.internal.php, and as of PHP 8.0.0 a TypeErrror is thrown in this case.

Also note that code examples in the manual should be indented with four spaces instead of tabs.

PR #181, opened 2020-11-06
Updating mysqli: Remove embedded leftovers

kamil-tekiela:
The support for mysqli embedded server has been completely removed from PHP 8 and has not been functional for quite some time. 
See https://github.com/php/php-src/commit/7be0e06b482e5bf14eaf4fe9a882d8b72d644038

### DIFF
--- a/reference/mysqli/mysqli_driver.xml
+++ b/reference/mysqli/mysqli_driver.xml
@@ -54,7 +54,7 @@
     <fieldsynopsis>
      <modifier>public</modifier>
      <modifier>readonly</modifier>
-     <type>string</type>
+     <type>bool</type>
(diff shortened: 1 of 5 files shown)


cmb69:
Thanks! 

Since the manual is supposed to document PHP 7.0.0 and newer, removing all that info is not appropriate. From a quick look, it seems to me that only `mysqli_driver_embedded_server_end` and `mysqli_driver_embedded_server_start` should be removed from versions.xml, and a note about mysqli-driver.props.embedded should be added that it always is `&false;` as of PHP 7.x.0 (not sure about x).

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351293.

PR #182, opened 2020-11-07
Updating mysqli: bring overview up to date

kamil-tekiela:
Removed the old MySQL extension information from mysqli overview page.


cmb69:
Very nice, thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351329.

PR #183, opened 2020-11-07
Updating mysqli api support for transactions

kamil-tekiela:
Primary goal: Fix broken "see also" links and remove manual error checking


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351328.

PR #184, opened 2020-11-11
Updating mysqli: Improving Quick Start guide

kamil-tekiela:
This PR aims to fix Quick Start guide in mysqli by removing manual error checking where possible, using OO style for function names and in the code (to encourage developers to use it more in new code), and fixing See also links. 

Other changes:
- Fixed grammar and typos in some places
- Changed the wording on Multiple Statements page. Previous text sounded as if SQL injection is only possible with multiple statements. 
- See also links were removed/added in some place. Although there seems to be no consistency or logic in them.
- Closing tags were removed from examples. They were not there in the beginning and adding them is a bad coding practice.
- It is possible to open TCP/IP connection using `localhost:3306`. That sentence has been changed to reflect what can be done rather than what can't be done.
- Some examples were removed as they were unnecessary and only bloated the page.
- Variable `$res` was renamed to `$result`
- Examples of executing queries were simplified. 
- Call to `$stmt->more_results()` in SP examples was removed. It was confusing and not needed in this context.
- `$stmt->store_result()` was added to the last example in SP to avoid the "Out of sync" error.
- As part of the move from procedural to OO style `mysqli_init()` was replaced with `new mysqli()`. A PR was created to deprecate `mysqli::init()`

Hopefully, this will make the page more useful to people who are only starting with mysqli. Having simple examples showing the best practices is very important.

### DIFF
--- a/reference/mysqli/quickstart.xml
+++ b/reference/mysqli/quickstart.xml
@@ -168,19 +168,14 @@ Possible but bad style.
     <programlisting role="php">
 <![CDATA[
 <?php
+
 $mysqli = new mysqli("localhost", "user", "password", "database");
-if ($mysqli->connect_errno) {
-    echo "Failed to connect to MySQL: (" . $mysqli->connect_errno . ") " . $mysqli->connect_error;
-}
+
 echo $mysqli->host_info . "\n";
 
 $mysqli = new mysqli("127.0.0.1", "user", "password", "database", 3306);
-if ($mysqli->connect_errno) {
-    echo "Failed to connect to MySQL: (" . $mysqli->connect_errno . ") " . $mysqli->connect_error;
-}
 
 echo $mysqli->host_info . "\n";
-?>
 ]]>
     </programlisting>
     &example.outputs;
--- a/reference/mysqli/quickstart.xml
+++ b/reference/mysqli/quickstart.xml
@@ -1263,6 +1263,47 @@ do {
 } while ($stmt->next_result());
 ]]>
     </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+---
+array(3) {
+  [0]=>
+  array(1) {
+    [0]=>
+    int(1)
+  }
+  [1]=>
+  array(1) {
+    [0]=>
+    int(2)
+  }
+  [2]=>
+  array(1) {
+    [0]=>
+    int(3)
+  }
+}
+---
+array(3) {
+  [0]=>
+  array(1) {
+    [0]=>
+    int(2)
+  }
+  [1]=>
+  array(1) {
+    [0]=>
+    int(3)
+  }
+  [2]=>
+  array(1) {
+    [0]=>
+    int(4)
+  }
+}
+]]>
+    </screen>
    </example>
   </para>
   <para>
@@ -1299,6 +1340,17 @@ do {
 } while ($stmt->next_result());
 ]]>
     </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+id = 1
+id = 2
+id = 3
+id = 2
+id = 3
+id = 4
+]]>
+    </screen>
    </example>
   </para>
   <para>
--- a/reference/mysqli/quickstart.xml
+++ b/reference/mysqli/quickstart.xml
@@ -1528,14 +1528,12 @@ $mysqli->commit();
     <programlisting role="php">
 <![CDATA[
 <?php
+
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 $mysqli = new mysqli("example.com", "user", "password", "database");
-if ($mysqli->connect_errno) {
-    echo "Failed to connect to MySQL: (" . $mysqli->connect_errno . ") " . $mysqli->connect_error;
-}
 
 $res = $mysqli->query("SELECT 1 AS _one, 'Hello' AS _two FROM DUAL");
 var_dump($res->fetch_fields());
-?>
 ]]>
     </programlisting>
     &example.outputs;
@@ -1619,11 +1617,14 @@ array(2) {
     <programlisting role="php">
 <![CDATA[
 <?php
+
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
+$mysqli = new mysqli("example.com", "user", "password", "database");
+
 $stmt = $mysqli->prepare("SELECT 1 AS _one, 'Hello' AS _two FROM DUAL");
 $stmt->execute();
 $res = $stmt->result_metadata();
 var_dump($res->fetch_fields());
-?>
 ]]>
     </programlisting>
    </example>
--- a/reference/mysqli/quickstart.xml
+++ b/reference/mysqli/quickstart.xml
@@ -120,7 +120,7 @@ $mysqli = new mysqli("example.com", "user", "password", "database");
 $res = mysqli_query($mysqli, "SELECT 'Possible but bad style.' AS _msg FROM DUAL");
 
 if ($row = $res->fetch_assoc()) {
-	echo $row['_msg'];
+    echo $row['_msg'];
 }
 ]]>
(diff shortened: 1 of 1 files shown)


reference/mysqli/quickstart.xml
```diff
@@ -763,20 +694,6 @@ array(4) {
...
-   <example>
-    <title>Less round trips using multi-INSERT SQL</title>
-    <programlisting role="php">
-<![CDATA[
-<?php
-if (!$mysqli->query("INSERT INTO test(id) VALUES (1), (2), (3), (4)")) {
-    echo "Multi-INSERT failed: (" . $mysqli->errno . ") " . $mysqli->error;
-}
-?>
-]]>
```
afilina, line 776: The example has been removed, but the text above still refers to multi-inserts. I don't see such an example elsewhere on the page either. I think that it's a valuable example.

reference/mysqli/quickstart.xml
```diff
@@ -1052,7 +932,7 @@ array(2) {
    The API does not include emulation for client-side prepared statement emulation.
   </para>
   <para>
-   <emphasis role="bold">Quick prepared - non-prepared statement comparison</emphasis>
+   <emphasis role="bold">Quick prepared/non-prepared statement comparison</emphasis>
```
afilina, line 935:
```suggestion
   <emphasis role="bold">Quick comparison of prepared and non-prepared statements</emphasis>
```

reference/mysqli/quickstart.xml
```diff
@@ -1438,26 +1275,29 @@ do {
...
-if (!$stmt->execute()) {
-    echo "Execute failed: (" . $stmt->errno . ") " . $stmt->error;
-}
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
+$mysqli = new mysqli("example.com", "user", "password", "database");
 
-do {
+$mysqli->query("DROP TABLE IF EXISTS test");
+$mysqli->query("CREATE TABLE test(id INT)");
+$mysqli->query("INSERT INTO test(id) VALUES (1), (2), (3)");
```
afilina, line 1284: The example being replaced does not have all these statements. I suppose that they were inferred from the previous example. Perhaps we can get rid of them here and jump straight to `$stmt = $mysqli->prepare("CALL p()")` to make this example more focused.

reference/mysqli/quickstart.xml
```diff
@@ -1469,22 +1309,21 @@ do {
...
  </section>
 
  <section xml:id="mysqli.quickstart.multiple-statement">
   <title>Multiple Statements</title>
   <para>
-   MySQL optionally allows having multiple statements in one statement string.
-   Sending multiple statements at once reduces client-server
-   round trips but requires special handling.
+   MySQL optionally allows having multiple statements in one statement string, 
+   but it requires special handling.
```
afilina, line 1322: I kinda prefer the original wording because it mentions the advantages. It's up to you though.

afilina (changes requested):

reference/mysqli/quickstart.xml
```diff
@@ -338,7 +320,7 @@ mysqli.default_socket=/tmp/mysql.sock
   <para>
    <simplelist>
     <member><methodname>mysqli::__construct</methodname></member>
-    <member><methodname>mysqli::init</methodname></member>
+    <member><function>mysqli_init</function></member>
```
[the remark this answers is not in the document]
kamil-tekiela: `mysqli::init` and `mysqli_init` are two completely different functions. The PR is to deprecate `mysqli::init()` which serves almost no purpose. `mysqli_init()` is still a useful function in the procedural set of mysqli functions. While `new mysqli()` can replace `mysqli_init()`, the function is just a procedural variant of creating a mysqli object without establishing a connection. 

See also RFC I have created https://wiki.php.net/rfc/improve_mysqli

kamil-tekiela:
@afilina Thank you very much for your reviews, Anna! I really appreciate someone taking the time to go over the changes and submit improvement suggestions. I have already accepted some of your proposals, but I will need more time to review the other ones and commit them to my PRs.

reference/mysqli/quickstart.xml
```diff
@@ -270,9 +252,9 @@ mysqli.default_socket=/tmp/mysql.sock
   <para>
    For setting a connection option, the connect operation has to be
    performed in three steps: creating a connection handle with
-   <function>mysqli_init</function>, setting the requested options using
-   <function>mysqli_options</function>, and establishing the network
-   connection with <function>mysqli_real_connect</function>.
+   <function>mysqli_init</function> or <methodname>mysqli::__construct</methodname>, 
+   setting the requested options using <methodname>mysqli::options</methodname>, 
+   and establishing the network connection with <methodname>mysqli::real_connect</methodname>.
```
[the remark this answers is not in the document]
kamil-tekiela: I prefer not to add any more examples to this PR. I might suggest a new PR with an example later.

reference/mysqli/quickstart.xml
```diff
@@ -675,70 +619,57 @@ if (!$stmt->execute()) {
...
-if (!$stmt->execute()) {
-    echo "Execute failed: (" . $stmt->errno . ") " . $stmt->error;
+$data = [
+    1 => 'PHP',
+    2 => 'Java',
+    3 => 'C++'
+];
+foreach ($data as $id => $label) {
+    $stmt->execute();
 }
```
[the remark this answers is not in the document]
kamil-tekiela: I don't see any reason why would we want that. The binding should happen only once, because the variables bound do not change within the loop.

reference/mysqli/quickstart.xml
kamil-tekiela: I removed that line on purpose. I do not want people to think that there is an advantage of using `multi_query()` over prepared statements. In fact, I still haven't found any good reason to use that function. Let's not give people any ideas that it somehow improves performance, at the cost of security and usability. Besides, I haven't been able to understand what it meant by "reduces client-server". Don't we need to make the same number of trips to the server (1 multi_query + N next_result)?  Unless it was meant to be all UPSERT queries where we discard the results. Either way, that sentence didn't make much sense to me.

reference/mysqli/quickstart.xml
kamil-tekiela: True, but I wanted all examples to be self-contained. You can copy the code and run it to produce the shown results (which are missing in this case). I propose that we make all examples self-contained and this is why I just now added the output too.

reference/mysqli/quickstart.xml
kamil-tekiela: Ok, if you think the example was valuable then I will try to add it back. 

The reason why I removed it is because it was neither here nor there. That page talks about prepared statements and how to bind parameters and then out of a sudden we have an example of a non-prepared query with constant values.  How does it apply to prepared statements? How can I replace the constant literals with placeholders? It might be a nice piece of advice, but it doesn't belong here. Unless we should how to bind parameters to multi-INSERT query then this example shouldn't be there. One can always check the MySQL manual for an example of such syntax.

reference/mysqli/quickstart.xml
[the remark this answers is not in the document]
afilina: Right, thanks for clarifying this for me.

reference/mysqli/quickstart.xml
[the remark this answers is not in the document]
afilina: My bad, I didn't realize that `$stmt->bind_param` receives arguments by reference.

reference/mysqli/quickstart.xml
afilina: I'm okay with making them self-contained in the quickstart guide, but not everywhere. If that is also what you had in mind, then we're in agreement.

afilina (approved):

afilina:
Looks like there is nothing preventing this from being merged.

cmb69:
Thank you both!

PR #185, opened 2020-11-12
Generate ext/date methodsynopses based on stubs

kocsismate:


salathe (changes requested):
Please don't remove the `role` attributes from the `*synopsis` elements. Those are used to decide what to include in the class synopses (for example, the list of methods for each class).

cmb69:
Thanks @salathe! I completely missed [the xincludes](https://github.com/php/doc-base/blob/498fdd8f5425606ead8ad32dd0ff7df0b4b4c95f/RFC/skeletons/classname.xml#L69).

kocsismate:
I've reintroduced the role generator capability of gen_stub.php. As there are a few edge case `role` values in this extension, I had to manually remove some unintended changes from the git change set.

cmb69:
Thank you!

PR #186, opened 2020-11-14
Fix return type documentation of gmp_hamdist

TysonAndre:
It's an integer since the function was added in php 5.5+.
See https://3v4l.org/DgIdi


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351392.

PR #189, opened 2020-11-17
mysqli: fixes from PhpStorm Stubs

voku:
via: https://github.com/voku/php-doc-fixer

```php bin/phpdocfixer run --auto-fix="true" ../doc-en/reference/mysqli```

---
This change is [Reviewable](https://reviewable.io/reviews/php/doc-en/189)

### DIFF
--- a/reference/mysqli/mysqli_result/lengths.xml
+++ b/reference/mysqli/mysqli_result/lengths.xml
@@ -13,7 +13,7 @@
   <fieldsynopsis><type>array</type><varname linkend="mysqli-result.lengths">mysqli_result->lengths</varname></fieldsynopsis>
   <para>&style.procedural;</para>
   <methodsynopsis>
-   <type>array</type><methodname>mysqli_fetch_lengths</methodname>
+   <type class="union"><type>int[]</type><type>false</type></type><methodname>mysqli_fetch_lengths</methodname>
    <methodparam><type>mysqli_result</type><parameter>result</parameter></methodparam>
   </methodsynopsis>
   <para>
--- a/reference/mysqli/functions/mysqli-get-client-stats.xml
+++ b/reference/mysqli/functions/mysqli-get-client-stats.xml
@@ -9,8 +9,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>array</type><methodname>mysqli_get_client_stats</methodname>
-   <void/>
+   <type class="union"><type>array</type><type>false</type></type><methodname>mysqli_get_client_stats</methodname>
+   <void />
   </methodsynopsis>
   <para>
    Returns client per-process statistics.
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -21,7 +21,7 @@
    <methodparam choice="opt"><type>string</type><parameter>socket</parameter><initializer>ini_get("mysqli.default_socket")</initializer></methodparam>
   </constructorsynopsis>
   <methodsynopsis role="oop">
-   <modifier>public</modifier> <type>void</type><methodname>mysqli::connect</methodname>
+   <modifier>public</modifier> <type class="union"><type>mysqli</type><type>false</type></type><methodname>mysqli::connect</methodname>
    <methodparam choice="opt"><type>string</type><parameter>host</parameter><initializer>ini_get("mysqli.default_host")</initializer></methodparam>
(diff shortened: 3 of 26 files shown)

### CODE REVIEW

**Comment by cmb69** [comment]
Interesting! Note that there is [work in progress](https://github.com/php/php-src/pull/6367) to update the docs from the stub.php files. /cc @kocsismate

File: reference/mysqli/mysqli_result/lengths.xml [hunk 1]
```diff
@@ -13,7 +13,7 @@
   <fieldsynopsis><type>array</type><varname linkend="mysqli-result.lengths">mysqli_result->lengths</varname></fieldsynopsis>
   <para>&style.procedural;</para>
   <methodsynopsis>
-   <type>array</type><methodname>mysqli_fetch_lengths</methodname>
+   <type class="union"><type>int[]</type><type>false</type></type><methodname>mysqli_fetch_lengths</methodname>
```
Girgias, line 16: I am seriously not keen on introducing such a hack when we have no official syntax for array shapes. Moreover, you can pass it an array and it will certainly not type error, at "best" it will ValueError in PHP 8.0, at "worst" it will ignore the miss match of types and will coerce them implicitly. As the internal behaviour of extensions dealing with shapes is *widely* inconsistent this is a -1 in my eyes.

kocsismate:
It's an unfortunate coincidence that you also recently created a similar tool to php-src's gen_stub.php :/

IMO the main problem with this PR is that phpstorm stubs are not the definite source of truth, unlike php-src's stubs.

voku:
@kocsismate I hacked that in one / two afternoons, so no problem if we can't use the output. For me, ```array``` is not that much better than ```mixed``` because in fact ```array``` is ```array<mixed,mixed>``` and you will have no clue what you will get.

cmb69:
> For me, array is not that much better than mixed because in fact array is array<mixed,mixed> and you will have no clue what you will get.

I agree that `array` doesn't convey much, but I wouldn't like to introduce pseudo-types again, at least not before we could settle on a convention which ones to support, and how to denote these (including support for [PhD](https://github.com/php/phd) rendering).

voku:
I don't know what the "documentation mailing list" is, can you send me a link? 😊

We could also add the type information as sentences, if you do not want to use the phpdoc (https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html#arrays) syntax.

cmb69:
> I don't know what the "documentation mailing list" is, can you send me a link?

Sure: phpdoc@lists.php.net

> We could also add the type information as sentences, […]

Not sure I understand. In any case, if array is specified as type, the shape of the array should be described in the manual. In this case (`mysqli_result::$length`) the return values section already states: "An array of integers representing the size of each column […]"

voku:
> > I don't know what the "documentation mailing list" is, can you send me a link?
> 
> Sure: [phpdoc@lists.php.net](mailto:phpdoc@lists.php.net)

Thanks.

> 
> > We could also add the type information as sentences, […]
> 
> Not sure I understand. In any case, if array is specified as type, the shape of the array should be described in the manual. In this case (`mysqli_result::$length`) the return values section already states: "An array of integers representing the size of each column […]"

Instead of a complex description, I thought about something more generic, so you know where to look if you want to know the types.:  "An array with ```int``` as key and ```int``` as value." ... but ```array<int, int>``` is even shorter and better for reading, or?

cmb69:
Ah, you want to write `array<int, int>` or `int[]` in the running text! Well, that would still require to explain that notation somewhere (not every reader of the PHP manual is supposed to know these "extended" type notations).

voku:
> Ah, you want to write `array<int, int>` or `int[]` in the running text! Well, that would still require to explain that notation somewhere (not every reader of the PHP manual is supposed to know these "extended" type notations).

This is why I suggested a generic sentence like: ```An array with "int" as key and "int" as value.``` but as already written I think ```array<int, int>``` or ```int[]``` is more readable.

PS: I updated my little script, so that we can check stub files from different sources and here is the result from ```php-src/**/*.stub.php``` ⇒ https://github.com/php/doc-en/compare/master...voku:php-src-stubs?expand=1 ... maybe we can use it to verify the current "based on stubs" fixes, or something like that. :-)

cmb69:
Having a relatively quick way to double check appears to be a good idea!

voku:
Only for info: At least in my small Tech-Twitter Bubble it looks like ```int[]``` is more known then ```array<int, int>```. And @Chemaclass mentioned that ```list<int>``` is even more correct. ;)

https://twitter.com/suckup_de/status/1328871300727762945

PR #193, opened 2020-11-17
Generate json methodsynopses based on stubs

kocsismate:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351406.

reference/json/jsonserializable/jsonserialize.xml
```diff
@@ -10,8 +9,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>abstract</modifier> <modifier>public</modifier> <type>mixed</type><methodname>JsonSerializable::jsonSerialize</methodname>
-   <void />
+   <modifier>public</modifier> <type>mixed</type><methodname>JsonSerializable::jsonSerialize</methodname>
```
cmb69, line 12: Should probably mention this: the script swallowed the `<modifier>abstract</modifier>`. I've kept it, but I'm not even sure it makes much sense, since this is an interface, and apparently the docs are not consistent in this regard anyway.

kocsismate: Yeah, I introduced this behaviour in https://github.com/php/php-src/pull/6367/commits/7b4f6d9ea7703135480df12510190677b2eafb9c in fact, because I noticed that the `abstract` method modifier is used pretty inconsistently in this case, when I was working on generating the methodsynapses for ext/date.

I'm fine though if it stays as before! Thanks for the update!

cmb69: Well, I think we should address this inconsistency at some point, but for now, a lot of other things seem to be more important. :)

PR #194, opened 2020-11-17
Generate opcache methodsynopses based on stubs

kocsismate:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351408.

PR #198, opened 2020-11-18
Generate exif methodsynopses from stubs

kocsismate:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351452.

reference/exif/functions/exif-thumbnail.xml
```diff
@@ -9,11 +9,11 @@
...
-   <type>string</type><methodname>exif_thumbnail</methodname>
-   <methodparam><type>mixed</type><parameter>stream</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter role="reference">width</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter role="reference">height</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter role="reference">imagetype</parameter></methodparam>
+   <type class="union"><type>string</type><type>false</type></type><methodname>exif_thumbnail</methodname>
+   <methodparam><type class="union"><type>resource</type><type>string</type></type><parameter>file</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter role="reference">width</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter role="reference">height</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter role="reference">image_type</parameter><initializer>&null;</initializer></methodparam>
```
cmb69, line 16: I was already finished with review and adding changelog entries, as I noticed the `null` initializers. Since I don't consider https://github.com/php/doc-en/pull/200#discussion_r526476787 to be resolved, I just removed them for now. We can still readd later. :)

cmb69: Re-added with http://svn.php.net/viewvc?view=revision&revision=351454.

reference/exif/functions/exif-read-data.xml
```diff
@@ -9,11 +9,11 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>array</type><methodname>exif_read_data</methodname>
-   <methodparam><type>mixed</type><parameter>stream</parameter></methodparam>
-   <methodparam choice="opt"><type>string</type><parameter>sections</parameter><initializer>&null;</initializer></methodparam>
-   <methodparam choice="opt"><type>bool</type><parameter>arrays</parameter><initializer>&false;</initializer></methodparam>
-   <methodparam choice="opt"><type>bool</type><parameter>thumbnail</parameter><initializer>&false;</initializer></methodparam>
+   <type class="union"><type>array</type><type>false</type></type><methodname>exif_read_data</methodname>
+   <methodparam><type class="union"><type>resource</type><type>string</type></type><parameter>file</parameter></methodparam>
```
salathe, line 13: The `resource|string` type declaration for the `file` parameter is not correct.  The `string` part is "anything that can be converted to a string"; so `float`s, `int`s, `object`s implementing `__toString`, etc. are all allowed types for this parameter.

/cc @cmb69 @kocsismate

nikic: I believe this is in line with existing documentation practice of specifying approximate types where useful. For stubs we are very pedantic about this (see https://github.com/php/php-src/blob/38b80aaf8574a5297e01d9a7ade5baf1637401d7/ext/exif/exif.stub.php#L7 which only specifies it as a documentation hint), but for docs purposes, specifying `resource|string` is more useful to the reader.

kocsismate: Yes, it's correct what you've just said, but all the other parameters of type `string` have a very similar behaviour in weak mode. The distinction between a real `string` parameter and one which is "anything that can be converted to a string" is only notable in strict mode, therefore I think it makes sense to report `string` in the manual.

salathe: I don't disagree on any particular point, it's yet another case where the manual lies for the greater good.

nikic: FWIW I do think we have a problem here, because at least for methods the type that is actually specified in the signature is important for inheritance purposes. But I don't see an obvious way how to combine the "real" type with the "documented" one. Especially for method returns we usually only document a type (to avoid BC), but I think docs would be much less useful if we dropped return types from methods for that reason.

cmb69: IMHO, this is an example of perfect being the enemy of good. Yes, `string` is not really correct – it neither matches the semantics of strict nor of weak typing in PHP 8.0.0. Still, I think it is more useful to specify `resource|string` than `mixed`, because the latter does not convey anything. The manual even specifies types for the individual INI settings, although all of these are strings, and I'm not aware that this causes great confusion, nor that anybody did complain about that. And there are sometimes descriptions of the shapes of arrays (parameters or return types), and their handling is sometimes very special for different functions; for instance, the `disposition` element of the `$body` parameter of `imap_mail_compose()` accepts arbitrary values, but silently ignores anything but arrays. Should that be documented? Well, maybe in a perfect world.

PR #199, opened 2020-11-18
Generate fileinfo methodsynopses based on stubs

kocsismate:


reference/fileinfo/finfo/buffer.xml
```diff
@@ -10,17 +9,16 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis role="oop">
-   <modifier>public</modifier> <type>string</type><methodname>finfo::buffer</methodname>
+   <modifier>public</modifier> <type class="union"><type>string</type><type>false</type></type><methodname>finfo::buffer</methodname>
    <methodparam><type>string</type><parameter>string</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>options</parameter><initializer>FILEINFO_NONE</initializer></methodparam>
-   <methodparam choice="opt"><type>resource</type><parameter>context</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer><constant>FILEINFO_NONE</constant></initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>resource</type><type>null</type></type><parameter>context</parameter><initializer>&null;</initializer></methodparam>
```
kocsismate, line 15: I've just realized that I'm using types according to the comment in question at https://github.com/php/doc-base/pull/4#discussion_r526184630. :) And IMO this fact warrants the capitalization change of default values :)

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351435. I have added changelog entries to `finfo-buffer.xml` and `finfo-file.xml`, but not for `finfo/buffer.xml` and `finfo/file.xml`. As I understand it, the latter are only useful to have menu entries (and maybe for the search), and as such shouldn't even have a `<refsect1 role="description">`  (or maybe xinclude that from the respective alias file).

PR #204, opened 2020-11-20
Generate xmlreader methodsynopses based on stubs

kocsismate:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351468.

For some reason, XMLReader::xml() had no changes in this PR, although the stubs appear to be correct (method is now static, and $options renamed to $flags); I've added these.

kocsismate:
Nice catch! I managed to track down the issue, and the problem was that this method is called `XMLReader::XML()` in the stubs, but it's referenced as `XMLReader::xml()` in the manual (in the `methodname` element).

PR #205, opened 2020-11-20
Add ValueError page

Girgias:


cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351496.

PR #207, opened 2020-11-21
Updating mysqli: fetch functions

kamil-tekiela:
Pushed by this comment https://stackoverflow.com/questions/64933761/how-to-return-integer-of-count-mysql-query-result-in-php/64933840?noredirect=1#comment114809131_64933840 I updated the mysqli docs to show the best practices usage

Fixes:
- Aligned examples. OO and procedural examples where using different variable names. An example of iterator was using different style and data
- Improved descriptions. Some words were confusing and even wrong. 
- Aligned notes. All notes are in the same place and follow similar grammar
- Removed manual error checking from examples
- Removed unnecessary LIMIT 50,5
- Added .similar to the example results

### DIFF
--- a/reference/mysqli/mysqli_result/fetch-array.xml
+++ b/reference/mysqli/mysqli_result/fetch-array.xml
@@ -73,7 +73,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an array of strings that corresponds to the fetched row or &null; if there
+   Returns an array of values that corresponds to the fetched row or &null; if there
    are no more rows in resultset.
   </para>
  </refsect1>
--- a/reference/mysqli/mysqli_result/fetch-array.xml
+++ b/reference/mysqli/mysqli_result/fetch-array.xml
@@ -22,15 +22,12 @@
   </methodsynopsis>
   <para>
    Returns an array that corresponds to the fetched row or &null; if there
-   are no more rows for the resultset represented by the
-   <parameter>result</parameter> parameter.
+   are no more rows for the result set.
   </para>
   <para>
-   <function>mysqli_fetch_array</function> is an extended version of the
-   <function>mysqli_fetch_row</function> function. In addition to storing the
-   data in the numeric indices of the result array, the 
-   <function>mysqli_fetch_array</function> function can also store the data
-   in associative indices, using the field names of the result set as keys.
+   In addition to storing the data in the numeric indices of the result array, 
+   this function can also store the data in associative indices 
+   by using the field names of the result set as keys.
   </para>
   &database.field-case;
   &database.fetch-null;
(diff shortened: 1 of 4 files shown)


reference/mysqli/mysqli_result/fetch-array.xml
```diff
@@ -22,15 +22,12 @@
...
   </para>
   <para>
-   <function>mysqli_fetch_array</function> is an extended version of the
-   <function>mysqli_fetch_row</function> function. In addition to storing the
-   data in the numeric indices of the result array, the 
-   <function>mysqli_fetch_array</function> function can also store the data
-   in associative indices, using the field names of the result set as keys.
+   In addition to storing the data in the numeric indices of the result array, 
+   this function can also store the data in associative indices 
+   by using the field names of the result set as keys.
```
afilina, line 30: This really eases the comprehension of the sentence.

reference/mysqli/mysqli_result/fetch-array.xml
```diff
@@ -73,87 +70,64 @@
...
 <?php
-$mysqli = new mysqli("localhost", "my_user", "my_password", "world");
 
-/* check connection */
-if ($mysqli->connect_errno) {
-    printf("Connect failed: %s\n", $mysqli->connect_error);
-    exit();
-}
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
+$mysqli = new mysqli("localhost", "my_user", "my_password", "world");
```
afilina, line 88: I'm definitely in favor of using the built-in error reporting and avoiding any occurrences of `exit()`, which is a terrible practice to promote.

reference/mysqli/mysqli_result/fetch-array.xml
```diff
@@ -73,87 +70,64 @@
...
 $row = $result->fetch_array(MYSQLI_BOTH);
-printf ("%s (%s)\n", $row[0], $row["CountryCode"]);
-
-/* free result set */
-$result->free();
-
-/* close connection */
-$mysqli->close();
-?>
+printf("%s (%s)\n", $row[0], $row["CountryCode"]);
```
afilina, line 103: I agree with removing unnecessary code from the example to keep it focused.

reference/mysqli/mysqli_result/fetch-assoc.xml
```diff
@@ -128,42 +103,48 @@ Santa Clara (USA)
    </screen>
   </example>
   <example xml:id="mysqli-result.example.iterator">
-   <title>A <classname>mysqli_result</classname> example comparing <classname>iterator</classname> usage</title>
+   <title>Comparison of <classname>mysqli_result</classname> <classname>iterator</classname> and <methodname>mysqli_result::fetch_assoc</methodname> usage</title>
+   <para><classname>mysqli_result</classname> can be iterated using &foreach;. The result set will always be iterated from the beginning until the end.</para>
```
afilina, line 107:
```suggestion
   <para><classname>mysqli_result</classname> can be iterated using &foreach;. The result set will always be iterated from the first element, regardless of the current position.</para>
```

afilina (changes requested):
LGTM.
I added an optional suggestion, as well as my feedback for the maintainers.

reference/mysqli/mysqli_result/fetch-assoc.xml
kamil-tekiela: "first element" -> "first row" maybe? I have been struggling with this sentence myself actually, so I do appreciate the suggestion.

afilina: I took the idea for "first element" from here.

image

I don't mind either way since there's no standard regarding this in the docs. In some places, these are called "entries". "Row" seems well suited for this context.

afilina:
LGTM.

kamil-tekiela:
Should I merge PR #39 into this PR? I believe the other PR has some good suggestions that I didn't notice before.

cmb69:
Thank you!

PR #215, opened 2020-11-23
Add an appendix about numeric strings

Girgias:
Add a dedicated appendix explaining numeric strings, drive-by fixes.

### DIFF
--- a/appendices/numeric-strings.xml
+++ b/appendices/numeric-strings.xml
@@ -1,0 +1,137 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<appendix xml:id="numeric-strings" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <title>Numeric strings</title>
+ <para>
+  A PHP <type>string</type> is considered numeric if it can be interpreted as
+  an <type>int</type> or a <type>float</type>.
+ </para>
+
+ <para>
+  Formally as of PHP 8.0.0:
+ </para>
+
+ <informalexample>
+  <programlisting>
+   <![CDATA[
+WHITESPACES      \s*
+LNUM             [0-9]+
+DNUM             ([0-9]*)[\.]{LNUM}) | ({LNUM}[\.][0-9]*)
+EXPONENT_DNUM    (({LNUM} | {DNUM}) [eE][+-]? {LNUM})
+INT_NUM_STRING   {WHITESPACES} [+-]? {LNUM} {WHITESPACES}
+FLOAT_NUM_STRING {WHITESPACES} [+-]? {EXPONENT_DNUM} {WHITESPACES}
+NUM_STRING       ({INT_NUM_STRING} | {FLOAT_NUM_STRING})
+]]>
+  </programlisting>
+ </informalexample>
+
+ <para>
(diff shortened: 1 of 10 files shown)


cmb69:
Hmm, where is the appendix? :)

appendices/numeric-strings.xml
```diff
@@ -0,0 +1,137 @@
...
+ </informalexample>
+
+ <para>
+  PHP also has a concept of <emphasis>leading</emphasis> numeric strings.
+  This is simply a string which starts like a numeric string followed by
+  any characters.
+ </para>
+
+ <section xml:id="numeric-string.conversion">
+  <title>Strings used in numeric contextes</title>
```
cmb69, line 35: Nit:
```suggestion
  <title>Strings used in numeric contexts</title>
```

appendices/numeric-strings.xml
```diff
@@ -0,0 +1,137 @@
...
+      The <type>string</type> is not numeric, throw a
+      <classname>TypeError</classname>.
+     </simpara>
+    </listitem>
+   </orderedlist>
+  </para>
+ </section>
+
+ <section xml:id="numeric-string.prior">
+  <title>Behaviour prior to PHP 8.0.0</title>
```
cmb69, line 73: Nit:
```suggestion
  <title>Behavior prior to PHP 8.0.0</title>
```

cmb69 (commented):
I wonder whether the numeric-strings.xml file should better go into language/types/?

Girgias:
> 
> 
> I wonder whether the numeric-strings.xml file should better go into language/types/?

I was thinking about that but it's not really a type, so I thought an appendix might be better. But that's easily changeable.

cmb69:
Well, neither declarations nor type-juggling are types. :)

Girgias:
> 
> 
> Well, neither declarations nor type-juggling are types. :)

That's true... let me rework it

language/types/float.xml
```diff
@@ -81,23 +81,35 @@ EXPONENT_DNUM (({LNUM} | {DNUM}) [eE][+-]? {LNUM})
...
+    For values of other types, the conversion is performed by converting the
+    value to <type>int</type> first and then to <type>float</type>. See
+    <link linkend="language.types.integer.casting">Converting to integer</link>
+    for more information.
    </para>
-  </note>
 
+   <note>
+    <para>
+     A notice is thrown if an <type>object</type> is converted to <type>float</type>.
```
Girgias, line 109: I wonder why this note exits, as the same applies for integers but there it's just declared undefined behaviour (which would also apply for objects...)

cmb69: See https://github.com/php/doc-en/pull/156/files#diff-4af4f41802cf4b045bb14a24dfac1bea5f6bf1682797f0bf22c98925b8af0a60L92-R99. Apparently, that was not the case in PHP 4; could probably drop that note.

language/types/type-juggling.xml
```diff
@@ -74,8 +74,9 @@ examples:
 
  <note>
   <para>
-   The behaviour of an automatic conversion to <type>array</type> is currently
-   undefined.
+   An <type>array</type> cannot be coerced and will throw a
+   <classname>TypeError</classname> as of PHP 8.0.0.
+   Previously the behaviour was undefined.
```
Girgias, line 79: Just realised that I probably misunderstood this note, probably should revert this.

PR #218, opened 2020-11-23
PDOStatement::fetchObject() returns object|false

BenMorel:
https://www.php.net/manual/en/pdostatement.fetchobject.php

> Returns an instance of the required class with property names that correspond to the column names or FALSE on failure.

### DIFF
--- a/reference/pdo/pdostatement/fetchobject.xml
+++ b/reference/pdo/pdostatement/fetchobject.xml
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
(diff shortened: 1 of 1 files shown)


cmb69:
Hmm, the [stub says the return type is mixed](https://github.com/php/php-src/blob/ce18899b44852c29ef80ea41d609df0c47acb1d0/ext/pdo/pdo_stmt.stub.php#L43-L44), so if this PR is correct, the stub should be fixed first.

BenMorel:
PR opened! https://github.com/php/php-src/pull/6449

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=351575.

PR #226, opened 2020-11-24
IntlCalendar::set() - Move month sequence description to month parameter

othercorey:
The month sequence information is relevant to the month parameter not the dayOfMonth parameter.

The `FIELD_MONTH` constant has a slightly different description. Don't know if they should be the same.


cmb69:
Good catch, thanks! The CI failure is unrelated to this PR, which has been applied as http://svn.php.net/viewvc?view=revision&revision=351574.

PR #227, opened 2020-11-24
Added documentation for the get_resource_id function

emmanix2002:
Added the documentation for `get_resource_id` according to what we have here: https://php.watch/versions/8.0/get_resource_id and the implementation in https://github.com/php/php-src/pull/5427/files (only needed this to confirm the parameter name)


emmanix2002:
@Girgias & @cmb69 here's a PR for the `get_resource_id` documentation

cmb69:
Thank you! 

*Minor* nit: please avoid unrelated whitespace changes, since that makes it harder for the translators.

emmanix2002:
> Thank you!
> 
> _Minor_ nit: please avoid unrelated whitespace changes, since that makes it harder for the translators.

Noted!

PR #228, opened 2020-11-24
rename misspelled methods

flow-control:
Hey there,

by going through https://github.com/php/php-tasks/issues/26 I found that these two methods actually are named `registerProgressCallback` and `registerCancelCallback` instead of `setProgressCallback` and `setCancelCallback`.

/Flo

### DIFF
--- a/appendices/migration80/new-features.xml
+++ b/appendices/migration80/new-features.xml
@@ -560,12 +560,12 @@ array_intersect(...$arrays);
    </listitem>
    <listitem>
     <para>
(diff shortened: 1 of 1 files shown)


cmb69:
We also should fix that in https://github.com/php/php-src/blob/PHP-8.0/UPGRADING

remicollet (approved):
LGTM

flow-control:
> We also should fix that in https://github.com/php/php-src/blob/PHP-8.0/UPGRADING

:+1: 

@cmb69 created a pr at https://github.com/php/php-src/pull/6452

cmb69:
Thanks for the PRs!

PR #229, opened 2020-11-24
Order the JIT configuration options in definition order.

Crell:
The diff looks weird because Git, but all this does is change the order to CTRO, since that's what everything else in the section uses as the order.  Why they were listed backwards, I have no idea.

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -948,15 +948,21 @@
       For advanced usage, this option accepts a 4-digit integer <literal>CRTO</literal>, where the digits mean:
       <variablelist>
        <varlistentry>
-        <term><literal>O</literal> (optimization level)</term>
+        <term><literal>C</literal> (CPU-specific optimization flags)</term>
         <listitem>
          <simplelist>
-          <member><literal>0</literal>: No JIT.</member>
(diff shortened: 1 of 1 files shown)


afilina:
LGTM.

afilina:
LGTM.

cmb69:
We had some quick discussion about this elsewhere, and finally it is an endianess issue. Unless we're assuming there is only one proper endianess, there is no correct order. Otherwise, the existing order is correct (since little-endian is surely the Right Way ;).

Crell:
The internal endianness has no bearing on the user experience, no?  This is just about the documentation listing definitions in the right order.  The order of the digits themselves is definitive, so the descriptions should match the digits.

cmb69:
Okay, I'm fine to change this. Is it useful to clarify that leading zeroes can be ommitted? (currently it is states that it must be a four digit string)

Crell:
Honestly I'd almost rather not at that point.  It's not a number, it's a string with 4 flags that should never have been put together like this in the first place.  Allowing removal of leading zeros implies that it's behaving in a number-ish fashion, when that's really not the case.

cmb69:
Okay, let's go with this then. Thank you for the pull request!

PR #231, opened 2020-11-24
Generate curl methodsynopses based on stubs

kocsismate:
There is a page for `CURLFile::__wakeup()` (https://www.php.net/manual/en/curlfile.wakeup.php), but the stubs doesn't define such method... Should we do anything in this case?

### DIFF
--- a/reference/curl/curlfile.xml
+++ b/reference/curl/curlfile.xml
@@ -51,6 +51,9 @@
 
     
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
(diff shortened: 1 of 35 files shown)


cmb69:
The only really relevant info on that page is: "CURLFile is not serializable", and this is still the case (as of PHP 7.4.0, the serialize handlers is set to `zend_class_serialize_deny`, and the method removed). And, yes, `__wakeup` is about unserialization. ;) IMO that page should better be removed, and the info that `CURLFile` is not (un)serializable should be given on the class page.

cmb69:
[`::__wakeup()` page removed](http://svn.php.net/viewvc?view=revision&revision=351623)

PR #237, opened 2020-11-25
OCI8 class renaming for PHP 8

cjbj:
This is the change discussed on the doc mail list.  Untested for various reasons.  Let's see if it builds.
It has the id renames, since that's what I currently have along with the other changes.


reference/oci8/OCILob/read.xml
```diff
@@ -1,15 +1,15 @@
...
-  <refname>OCI-Lob::read</refname>
+  <refname>OCILob::read</refname>
   <refpurpose>Reads part of the large object</refpurpose>
  </refnamediv>
  
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis role="oop">
-   <type class="union"><type>string</type><type>false</type></type><methodname>OCI-Lob::read</methodname>
+   <type>string</type><methodname>OCILob::read</methodname>
```
cmb69, line 12: According to the return values section, this method returns false on failure; that should either be reflected in the in the methodsynopsis, or get fixed in the return values section. If that change is only relevant as of PHP 8.0.0, it could be listed as a changelog entry.

reference/oci8/OCICollection.xml
```diff
@@ -1,45 +1,50 @@
...
+  <section xml:id="OCICollection.intro">
    &reftitle.intro;
    <para>
     OCI8 Collection functionality.
    </para>
+   <note>
+   <para>
+    The OCI-Collection class was renamed to OCICollection in PHP 8 and PECL OCI8 3.0.0 to align with PHP naming standards.
+   </para>
+   </note>
```
cmb69, line 19: I would prefer to have changelog entries instead:
```suggestion
 <refsect1 role="changelog">
  &reftitle.changelog;
  <informaltable>
   <tgroup cols="2">
    <thead>
     <row>
      <entry>&Version;</entry>
      <entry>&Description;</entry>
     </row>
    </thead>
    <tbody>
     <row>
      <entry>8.0.0, PECL OCI8 3.0.0</entry>
      <entry>
       The <classname>OCI-Collection</classname> class was renamed to
       <classname>OCICollection</classname> to align with PHP naming standards.
      </entry>
     </row>
    </tbody>
   </tgroup>
  </informaltable>
 </refsect1>
```

appendices/aliases.xml
```diff
@@ -343,7 +343,7 @@
      </row>
      <row>
       <entry>ociassignelem</entry>
-      <entry><link linkend="oci-collection.assignelem">OCICollection::assignElem</link></entry>
+      <entry><link linkend="ocicollection.assignelem">OCICollection::assignElem</link></entry>
```
cmb69, line 346: I'm not 100% convinced that changing the IDs and renaming files is a good idea right now. While it is certainly correct, it causes more work for the translation teams, and these are likely to be under pressure anyway.

cmb69 (commented):
Thanks for the PR! :)

reference/oci8/OCILob/read.xml
cjbj: I see this is because in between me initially working on the changes, you had corrected the synopsis.  I will pick up your changes.

reference/oci8/OCICollection.xml
cjbj: That seems logical.  It's easy in the individual method files which already `<refentry>` sections but would require a bit of a rewrite for the OCI-Collection.xml and OCI-Lob.xml; I will defer that nicety.

appendices/aliases.xml
cjbj: Agree.

cjbj:
I'm going to update from a reset repo; and will abandon this PR.

PR #238, opened 2020-11-25
Exception updates

Crell:
* Empty catch statements
* Throw as an expression
* Various other cleanup while I was there to remove comments

I also deleted a crapload of comments.  It's so satisfying.

### DIFF
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -204,7 +204,7 @@ echo "\n\n";
    </para>
    <para>
     As of PHP 8.0.0, the &throw; keyword is an expression and may be used in any expression
-    context. In prior versions it was a statement and must be on its own line.
+    context. In prior versions it was a statement and was required to be on its own line.
    </para>
 
   </simplesect>
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -196,13 +196,17 @@ echo "\n\n";
     global scope without encountering a matching &catch; block, the program will
     terminate with a fatal error.
    </para>
-
    <para>
     The thrown object must be an instance of the
     <classname>Exception</classname> class or a subclass of
     <classname>Exception</classname>. Trying to throw an object that is not
     will result in a PHP Fatal Error.
    </para>
+   <para>
+    As of PHP 8.0.0, the &throw; keyword is an expression and may be used in any expression
+    context. In prior versions it was a statement and must be on its own line.
+   </para>
+
   </simplesect>
 
   <simplesect xml:id="language.exceptions.catch">
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -194,7 +194,7 @@ echo "\n\n";
     function until it finds a matching &catch; block.  All &finally; blocks it encounters
     along the way will be executed. If the call stack is unwound all the way to the
     global scope without encountering a matching &catch; block, the program will
-    terminate with a fatal error.
+    terminate with a fatal error unless a global exception handler has been set.
    </para>
    <para>
     The thrown object must be an instance of the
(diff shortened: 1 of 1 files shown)


language/exceptions.xml
```diff
@@ -188,22 +188,43 @@ echo "\n\n";
...
+
   </simplesect>
 
   <simplesect xml:id="language.exceptions.catch">
    <title><literal>catch</literal></title>
+   <para>
+    A &catch; block defines how to respond to a thrown exception.  A &catch;
+    block defines one or more types of exception or error it can handle, and
+    optionally a variable to which to assign the exception. (The variable was
+    required prior to PHP 8.0.0.)  The first &catch; block a thrown exception
```
Girgias, line 218:
```suggestion
    required prior to PHP 8.0.0.) The first &catch; block a thrown exception
```
Nit:

language/exceptions.xml
```diff
@@ -188,22 +188,43 @@ echo "\n\n";
...
     &catch; or &finally; block.
    </para>
-
+   <para>
+    If an exception is thrown and its current function scope has no &catch;
+    block, the exception will "bubble up" the call stack to the calling
+    function until it finds a matching &catch; block.  All &finally; blocks it encounters
+    along the way will be executed. If the call stack is unwound all the way to the
+    global scope without encountering a matching &catch; block, the program will
+    terminate with a fatal error.
```
cmb69, line 197: maybe add something about exception handlers?

language/exceptions.xml
```diff
@@ -188,22 +188,43 @@ echo "\n\n";
...
+   </para>
    <para>
     The thrown object must be an instance of the
     <classname>Exception</classname> class or a subclass of
     <classname>Exception</classname>. Trying to throw an object that is not
     will result in a PHP Fatal Error.
    </para>
+   <para>
+    As of PHP 8.0.0, the &throw; keyword is an expression and may be used in any expression
+    context. In prior versions it was a statement and must be on its own line.
```
cmb69, line 207: Tense doesn't match ("was" vs "must"), from what I can tell.

Girgias (approved):
LGTM

language/exceptions.xml
Crell: Double spaces is the correct, god-fearing way to write sentences.  Those heathens who have been corrupted by HTML's whitespace folding are in violation of the lord's design.

language/exceptions.xml
```diff
@@ -435,6 +461,52 @@ string(11) "MyException"
...
+   <example>
+    <title>Omitting the caught variable</title>
+    <para>Only permitted in PHP 8.0.0 and later.</para>
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+class SpecificException extends Exception {}
+
+function test() {
```
cmb69, line 473: According to the [manual's style guide](http://doc.php.net/tutorial/style.php), examples should adhere to PEAR CS. That might be worthwhile to discuss on the doc ML, but for now, maybe stick with it. I don't have any hard feelings about that, though.

cmb69 (commented):
Thank you! Overall very nice work. :)

language/exceptions.xml
Crell: Isn't that PEAR style?  I thought I already did that.

language/exceptions.xml
DaveRandom: Be gonne, foule fiend! Take thy additional whitespace and caste it oute! The power of kerning compels you.

language/exceptions.xml
cmb69: IIRC, PEAR CS want's function body braces on new lines, and also doesn't want a new line after try braces just before catch/finally. Should be same for PSR 2.

cmb69: Ah, other functions in this file are not K&R either, so this is fine.

PR #239, opened 2020-11-25
Document the ::class constant on objects.

Crell:
Not much else to say here.

### DIFF
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -413,7 +413,7 @@ a default value
    <title>::class</title>
 
    <para>
-    Since PHP 5.5, the <literal>class</literal> keyword is also used for class
+    As of PHP 5.5, the <literal>class</literal> keyword is also used for class
     name resolution. You can get a string containing the fully qualified name
     of the <literal>ClassName</literal> class by using
     <literal>ClassName::class</literal>. This is particularly useful with
@@ -467,7 +467,7 @@ Some\Class\Does\Not\Exist
     </example>
    </note>
    <para>
-    Since PHP 8.0.0, the <literal>::class</literal> constant may also be used on
+    As of PHP 8.0.0, the <literal>::class</literal> constant may also be used on
     objects. This resolution happens at runtime, not compile time. Its effect is
     the same as calling <function>get_class</function> on the object.
    </para>
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -449,6 +449,22 @@ NS\ClassName
      are expanded even if the class does not exist. No error is issued in
      that case.
     </para>
+    <example xml:id="language.oop5.basic.class.class.name">
+     <title>Missing class name resolution</title>
+     <programlisting role="php">
+<![CDATA[
+<?php
+print Some\Class\DoesNot\Exist::class;
+?>
+]]>
+     </programlisting>
+     &example.outputs;
+     <screen>
+<![CDATA[
+Some\Class\Does\Not\Exist
+]]>
+     </screen>
+    </example>
    </note>
    <para>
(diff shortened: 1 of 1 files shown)


language/oop5/basic.xml
```diff
@@ -450,6 +450,32 @@ NS\ClassName
      that case.
     </para>
    </note>
+   <para>
+    Since PHP 8.0.0, the <literal>::class</literal> constant may also be used on
```
Girgias, line 454:
```suggestion
    As of PHP 8.0.0, the <literal>::class</literal> constant may also be used on
```

language/oop5/basic.xml
```diff
@@ -450,6 +450,32 @@ NS\ClassName
...
+    objects. This resolution happens at runtime, not compile time. Its effect is
```
Girgias, line 455: Might want to clarify that i acts on variables? Not sure how important that is tho.

Girgias (commented):
Nit and a comment

language/oop5/basic.xml
Crell: The paragraph before it used "Since", which is why I copied that.  Should we change that one, too?

language/oop5/basic.xml
Crell: I think the example makes it sufficiently clear how to use it.

Girgias: That's true

language/oop5/basic.xml
Girgias: Would make the most sense.

language/oop5/basic.xml
```diff
@@ -449,7 +449,49 @@ NS\ClassName
      are expanded even if the class does not exist. No error is issued in
      that case.
     </para>
+    <example xml:id="language.oop5.basic.class.class.name">
```
cmb69, line 452: Maybe here (and also below):
```suggestion
    <informalexample xml:id="language.oop5.basic.class.class.name">
```
That should make configure happy. :)

Crell: What's the difference?

language/oop5/basic.xml
```diff
@@ -449,7 +449,49 @@ NS\ClassName
...
+?>
+]]>
+     </programlisting>
+     &example.outputs;
+     <screen>
+<![CDATA[
+Some\Class\Does\Not\Exist
+]]>
+     </screen>
+    </example>
```
Crell, line 467:
```suggestion
    </informalexample>
```

language/oop5/basic.xml
```diff
@@ -413,7 +413,7 @@ a default value
    <title>::class</title>
 
    <para>
-    Since PHP 5.5, the <literal>class</literal> keyword is also used for class
+    As of PHP 5.5, the <literal>class</literal> keyword is also used for class
```
Girgias, line 416: We don't actually need the As of part as the docs are PHP 7+ only now.

Girgias:
A quick `git grep` search shows that the XML ID is already in use here:
``language/oop5/basic.xml:    <example xml:id="language.oop5.basic.class.class.name">`` and is the only usage of it (no references) so renaming that example ID could be a solution.

language/oop5/basic.xml
cmb69: Indeed; I'm going to remove that. There are more references to PHP 5 in that file, but I'll leave these alone for now.

cmb69:
@Crell, the main difference between `<example>` and `<informalexample>` is that the former requires a `<title>` while the latter must not have any (there *may* also be differences how that is handled by PhD). As such my suggestion was bad. The actual problem have been the duplicated IDs (as pointed out by @Girgias).

PR #241, opened 2020-11-25
Function trailing commas

Crell:
* Trailing commas in function parameter list.
* Trailing commas in use clause.
* Minor language change along the way.
* I have no idea what the difference is between para and simplepara...

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -200,7 +200,31 @@ function takes_array($input)
 ]]>
      </programlisting>
     </example>
-   </para>   
+   </para>
+   <para>
+    As of PHP 8.0.0, the list of function arguments may include a trailing comma, which
+    will be ignored.  That is particularly useful in cases where the list of arguments is
+    long or contains long variable names, making it convenient to list arguments vertically.
+   </para>
+   <example>
+    <title>Passing arrays to functions</title>
+    <programlisting role="php">
+     <![CDATA[
+<?php
+function takes_many_args(
+    $first_arg,
+    $second_arg,
+    $a_very_long_argument_name,
+    $arg_with_default = 5,
+    $again = 'a default string', // This trailing comma is optional as of PHP 8.0.0
+)
+{
+    // ...
+}
+?>
+]]>
+    </programlisting>
+   </example>
    <sect2 xml:id="functions.arguments.by-reference">
     <title>Passing arguments by reference</title>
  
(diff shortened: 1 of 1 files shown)


language/functions.xml
```diff
@@ -200,7 +200,31 @@ function takes_array($input)
...
+    <title>Passing arrays to functions</title>
+    <programlisting role="php">
+     <![CDATA[
+<?php
+function takes_many_args(
+    $first_arg,
+    $second_arg,
+    $a_very_long_argument_name,
+    $arg_with_default = 5,
+    $again = 'a default string', // This trailing comma is optional as of PHP 8.0.0
```
Girgias, line 219: I feel how this comment is written it means that prior to PHP 8.0.0 the trailing comma was mandatory

cmb69:
`<simpara>` is somewhat like a HTML `<p>`, but `<para>` is more like a `<div>`. See https://tdg.docbook.org/tdg/4.5/para.html#d0e133584 for a more specific explanation.

That said, I think that currently PhD treats both pretty similar, and in practise, the documentation mostly uses `<para>`.

Girgias:
> 
> 
> `<simpara>` is somewhat like a HTML `<p>`, but `<para>` is more like a `<div>`. See https://tdg.docbook.org/tdg/4.5/para.html#d0e133584 for a more specific explanation.
> 
> That said, I think that currently PhD treats both pretty similar, and in practise, the documentation mostly uses `<para>`.

I think the only important distinction is when it's inside a `<note>`, `<warning>`, etc. element. With a simpara you'll get something like:
```
Note: simpara content
```
whereas with a para you get:
```
Note:

Para content
```

Crell:
Updated the comment.

Girgias (approved):
LGTM :)

cmb69:
Thanks!

It seems that the manual is generally sloppy regarding the term "argument", so I left it as is, but in the long run we may consider to distinguish between "parameter" and "argument" (or "formal parameter" and "actual parameter").

Crell:
What even is the difference between an argument and a parameter?  I've generally seen/used them interchangeably.

cmb69:
See https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter

Girgias:
Huh, I've been getting the proper order wrong then apparently, oh well TIL

PR #242, opened 2020-11-26
Update ext/ldap signatures to match PHP-8

MCMic:
Used gen_stub.php to update information in method synopses.
Had to fix ldap_connect by hand because of the deprecated signature
 information.

### DIFF
--- a/reference/ldap/functions/ldap-connect.xml
+++ b/reference/ldap/functions/ldap-connect.xml
@@ -9,8 +9,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>resource</type><methodname>ldap_connect</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>ldap_uri</parameter><initializer>&null;</initializer></methodparam>
+   <type class="union"><type>resource</type><type>false</type></type><methodname>ldap_connect</methodname>
+   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>uri</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <warning>
    <simpara>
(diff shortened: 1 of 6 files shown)


MCMic:
What is the situation for this merge request and others tagged methodsynopses, is someone reviewing/merging them?

cmb69:
Yes, we're [working on that](https://github.com/php/doc-en/pulls?q=label%3Amethodsynopses+is%3Aclosed); it's just a bit much for the doc team, currently.

cmb69:
I have applied (and amendent) most of the changes, but left out `ldap_connect()`, `ldap_list()`, `ldap_read()` and `ldap_search()` because it is not quite clear to me which changes have been applied compared to PHP 7 (and these should be documented). This needs closer review. Could you please rebase to make that easier?

MCMic:
> I have applied (and amendent) most of the changes, but left out `ldap_connect()`, `ldap_list()`, `ldap_read()` and `ldap_search()` because it is not quite clear to me which changes have been applied compared to PHP 7 (and these should be documented). This needs closer review. Could you please rebase to make that easier?

I rebased.

It’s really hard to answer what changed, because what the documentation said, what the code was doing and what the code was supposed to do were 3 different things. So I tried my best to fix that and to have sane default values for all attributes so that named parameters can be used in calls to ldap functions.

cmb69:
Yeah, it's quite difficult. I made some amendments, the best I could.

MCMic:
> Yeah, it's quite difficult. I made some amendments, the best I could.

Thank you.

The pages still list «(PHP 4, PHP 5, PHP 7)» in the version list (here for ldap_search).
How is this list handled, will it be automatically updated?

cmb69:
There are versions.xml files, where PHP 8 support will have to be added. I think we can do that with a script.

PR #243, opened 2020-11-26
Argument of sapi_windows_cp_get is optional

TysonAndre:
I'm a bit confused on how to document the rest of this since I havent used it myself. "If the value is the empty string (or an unrecognized value), this returns the current code page without restricting it to ansi or oem categories"? 

An example use of `sapi_windows_cp_get()` with no args would be to fetch the old code page, set it to run a command, then restore the old code page https://github.com/search?l=PHP&p=2&q=sapi_windows_cp_get&type=Code

```c
// win32/codepage.c
	if (kind_len == sizeof("ansi")-1 && !strncasecmp(kind, "ansi", kind_len)) {
		RETURN_LONG(GetACP());
	} else if (kind_len == sizeof("oem")-1 && !strncasecmp(kind, "oem", kind_len)) {
		RETURN_LONG(GetOEMCP());
	} else {
		const struct php_win32_cp *cp = php_win32_cp_get_current();
		RETURN_LONG(cp->id);
	}
```

### DIFF
--- a/reference/misc/functions/sapi-windows-cp-get.xml
+++ b/reference/misc/functions/sapi-windows-cp-get.xml
@@ -11,7 +11,7 @@
   &reftitle.description;
   <methodsynopsis role="procedural">
    <type>int</type><methodname>sapi_windows_cp_get</methodname>
-   <methodparam><type>string</type><parameter>kind</parameter></methodparam>
+   <methodparam choice="opt"><type>string</type><parameter>kind</parameter><initializer>""</initializer></methodparam>
   </methodsynopsis>
   <para>
    Get the identifier of the codepage of the current process.


cmb69:
Thanks , @TysonAndre! Looks like this API is a bit hard to document; I hope my improvements are sufficient.

TysonAndre:
As a single sentence, I first confused it with "Returns X if (Y, or Z)...". I forget if there's a style recommendation or typical style for documenting enums.

`If <parameter>kind</parameter> is <literal>'ansi'</literal>, this returns the current <acronym>ANSI</acronym> code page of the operating system.`
`If <parameter>kind</parameter> is <literal>'oem'</literal>, this returns the current <acronym>OEM</acronym> code page of the operating system.`
`Otherwise, the current codepage of the process is returned.`

instead of

```
   Returns the current <acronym>ANSI</acronym> code page of the operating system,
   if <parameter>kind</parameter> is <literal>'ansi'</literal>,
   or the current <acronym>OEM</acronym> code page of the operating system,
   if <parameter>kind</parameter> is <literal>'oem'</literal>;
   otherwise, the current codepage of the process is returned.
```

cmb69:
Yes, that's probably somewhat better. There's not really some general style for such return values, I think. Neither a list won't fit, nor would distinct paragraphs. Applied as http://svn.php.net/viewvc?view=revision&revision=351712.

PR #244, opened 2020-11-26
Add nullsafe arrow.

Crell:
I included a Note giving guidance on when to use it, because I keep fielding "but doesn't this just encourage bad error handling? questions from people. I figure this is a case where some usage guidance is warranted.

### DIFF
--- a/appendices/tokens.xml
+++ b/appendices/tokens.xml
@@ -589,7 +589,7 @@
     </row>
     <row>
      <entry><constant>T_NULLSAFE_OBJECT_OPERATOR</constant></entry>
-     <entry>-&gt;</entry>
+     <entry>?-&gt;</entry>
      <entry><link linkend="language.oop5">classes and objects</link></entry>
     </row>
     <row>
--- a/appendices/tokens.xml
+++ b/appendices/tokens.xml
@@ -588,6 +588,11 @@
      <entry><link linkend="language.oop5">classes and objects</link></entry>
     </row>
     <row>
+     <entry><constant>T_NULLSAFE_OBJECT_OPERATOR</constant></entry>
+     <entry>-&gt;</entry>
+     <entry><link linkend="language.oop5">classes and objects</link></entry>
+    </row>
+    <row>
      <entry><constant>T_OPEN_TAG</constant></entry>
      <entry>&lt;?php, &lt;? or &lt;%</entry>
      <entry><link linkend="language.basic-syntax.phpmode">escaping
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -451,7 +451,54 @@ NS\ClassName
     </para>
    </note>
   </sect2>
- </sect1>
+ <sect2 xml:id="language.oop5.basic.nullsafe">
+  <title>Nullsafe methods and properties</title>
+  <para>
+   As of PHP 8.0.0, properties and methods may also be accessed with the
+   "nullsafe" arrow instead: <literal>?-></literal>. The nullsafe arrow work the same as property
+   or method access as above, except that if the object being dereferenced
+   is null then null will be returned rather than an exception thrown.
+   If the dereference is part of a chain, the rest of the chain is skipped.
+  </para>
+  <para>
+   The effect is similar to wrapping each access in an <funtion>is_null</funtion>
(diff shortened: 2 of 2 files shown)


Girgias:
I think the token page and the language/operator page should also be update to include a mention/link to this page.

Other than that LGTM

cmb69:
May prefer the term "nullsafe operator" over "nullsafe arrow"?

language/oop5/basic.xml
```diff
@@ -451,7 +451,54 @@ NS\ClassName
     </para>
    </note>
   </sect2>
- </sect1>
+ <sect2 xml:id="language.oop5.basic.nullsafe">
+  <title>Nullsafe methods and properties</title>
+  <para>
+   As of PHP 8.0.0, properties and methods may also be accessed with the
+   "nullsafe" arrow instead: <literal>?-></literal>. The nullsafe arrow work the same as property
```
kamil-tekiela, line 458: The `->` is called "object operator". We should call `?->` the "nullsafe object operator" or just "nullsafe operator". Don't call it an arrow as this is very ambiguous.

iluuu1994:
We do call it "nullsafe operator" and "nullsafe object operator" internally. I don't have a preference but I'd stick with one of the two.

Crell:
I went with nullsafe operator for conciseness, and added it to the token list.

The -> operator is not in the operators page as is (one of the comments notes it), and it's not in the precedence list in zend_language_parser.y either, much to my surprise.  That seems like a larger question to resolve than I want to get into here.

appendices/tokens.xml
```diff
@@ -587,6 +587,11 @@
      <entry>-&gt;</entry>
      <entry><link linkend="language.oop5">classes and objects</link></entry>
     </row>
+    <row>
+     <entry><constant>T_NULLSAFE_OBJECT_OPERATOR</constant></entry>
+     <entry>-&gt;</entry>
```
kamil-tekiela, line 592: `?-&gt;` instead of `-&gt;`

kamil-tekiela (changes requested):

appendices/tokens.xml
Crell: Sigh. Thanks.

cmb69:
> The -> operator is not in the operators page as is (one of the comments notes it), and it's not in the precedence list in zend_language_parser.y either, much to my surprise. That seems like a larger question to resolve than I want to get into here.

I agree that this can be resolved another time.

language/oop5/basic.xml
```diff
@@ -451,7 +451,55 @@ NS\ClassName
...
    </note>
   </sect2>
- </sect1>
+ <sect2 xml:id="language.oop5.basic.nullsafe">
+  <title>Nullsafe methods and properties</title>
+  <para>
+   As of PHP 8.0.0, properties and methods may also be accessed with the
+   "nullsafe" operator instead: <literal>?-></literal>. The nullsafe operator
+   works the same as property or method access as above, except that if the
+   object being dereferenced is <literal>null</literal> then <literal>null</literal>
```
cmb69, line 460: When referring to the `null` value, it's usually best to denote it as `&null;`.

language/oop5/basic.xml
```diff
@@ -451,7 +451,55 @@ NS\ClassName
...
+   will be returned rather than an exception thrown. If the dereference is part of a
+   chain, the rest of the chain is skipped.
+  </para>
+  <para>
+   The effect is similar to wrapping each access in an <function>is_null</function>
+   check first, but more compact.
+  </para>
+  <para>
+   <example>
+    <title>Class name resolution</title>
```
cmb69, line 470: Interesting example for "class name resolution". ;)

language/oop5/basic.xml
```diff
@@ -451,7 +451,55 @@ NS\ClassName
...
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+// This line in PHP 8.0:
```
cmb69, line 475: This gets confusing when PHP 8.1.0 will be released: does it work only for PHP 8.0?

cmb69 (approved):
Thank you! I left a few commits and applied with respective changes.

PR #245, opened 2020-11-27
Updating mysqli: Remove PHP 5 mentions from options and fix bug #68923

kamil-tekiela:
This PR partially fixes https://bugs.php.net/bug.php?id=68923 but listing all possible options would be too much. The options given there are the most common options given as an example. That list is not exhaustive. 

I was not sure whether the note _"mysqli_options() should be called after mysqli_init() and before mysqli_real_connect()"_ should be reworded. This is not entirely true as some options can be changed at any given time e.g. `MYSQLI_OPT_INT_AND_FLOAT_NATIVE`

### DIFF
--- a/reference/mysqli/mysqli/options.xml
+++ b/reference/mysqli/mysqli/options.xml
@@ -57,24 +57,29 @@
          <tbody>
           <row>
            <entry><constant>MYSQLI_OPT_CONNECT_TIMEOUT</constant></entry>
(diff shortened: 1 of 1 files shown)


Girgias (approved):

cmb69:
In the long run, we should strive to document (or at least, list) *all* options, but this is already a nice improvement. Thanks!

I'm not sure about that note, either. If in doubt, just leave it as is. We can still improve later.

PR #247, opened 2020-11-28
Updating mysqli: Remove pre-deprecation notice

kamil-tekiela:
After giving it some thought I decided that there is absolutely no reason to have a confusing pre-deprecation notice in the documentation for this constructor. There are no plans to deprecate it, it works as designed, and there are no technical downsides to using it. If this function were going to be deprecated then it would have happened in the past 15 years. 
The only reason why someone might not want to use it is a personal preference. Telling people not to use something because **it might be** deprecated is a terrible practice. 
Also, using procedural style functions over OO style is arguably a worse suggestion.

### DIFF
--- a/reference/mysqli/mysqli_stmt/construct.xml
+++ b/reference/mysqli/mysqli_stmt/construct.xml
@@ -17,17 +17,6 @@
   <para>
    This method constructs a new <classname>mysqli_stmt</classname> object.
   </para>
-  <note>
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you! I mostly agree with your reasoning, so this PR is fine for me.

If anybody has objections, please raise them! :)

PR #248, opened 2020-11-28
Remove mysqli_warning::__construct

kamil-tekiela:
This feature was never completed. The constructor was never accessible from userland. It is defined as protected on a final class. 
[Since PHP 8.0 the constructor was made private.](https://github.com/php/php-src/commit/8138ed73d73209024ad8500a4ab565b84259a1bf)
Documenting the existence of an inaccessible constructor serves no purpose. This is just a redundant page in the documentation IMHO.


cmb69 (approved):
Good catch., thank you!

cmb69:
Amended with http://svn.php.net/viewvc?view=revision&revision=351738.

PR #250, opened 2020-11-28
Removed literal tags from title for readability

kamil-tekiela:
After recent changes the headers for control structures look really small. I removed the code highlighted which seems pretty useless in the title to improve readability.


cmb69:
Ah, yes, this is because `<literal>` is now translated to `<code>` where it formerly was `<em>`.

I don't see much value in having the markup in the `<title>`, so removing is fine for me.

PR #251, opened 2020-11-29
Document named arguments

ekinhbayar:
Initial draft for documenting named arguments on the functions page. 

Feedback is most welcomed! This PR could also use a quick fact-check and I'd appreciate pointing out what's missing that needs documentation. 

This PR also documents deprecation of optional args after mandatory ones.

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -228,8 +228,8 @@ function takes_many_args(
    <para>
     As of PHP 8.0.0, passing optional arguments after mandatory arguments
     is deprecated. This can generally be resolved by dropping the default value.
-    One exception to this rule are parameters of the form
-    Type $param = &null;, where the &null; default makes the type implicitly
+    One exception to this rule are arguments of the form
+    <code>Type $param = &null;</code>, where the &null; default makes the type implicitly
     nullable. This usage remains allowed, though it is recommended to use an
     explicit nullable type instead.
    </para>
@@ -633,7 +633,7 @@ array_fill(start_index: 0, num: 100, value: 50);
     </para>
 
     <example>
-     <title>Same as above example with a different order of parameters</title>
+     <title>Same example as above with a different order of parameters</title>
      <programlisting role="php">
       <![CDATA[
 <?php
@@ -644,7 +644,7 @@ array_fill(value: 50, num: 100, start_index: 0);
     </example>
 
     <para>
-     You can combine named arguments with positional arguments. In this case,
+     Named arguments can be combined with positional arguments. In this case,
      the named arguments must come after the positional arguments.
      It is also possible to specify only some of the optional arguments of a
      function, regardless of their order.
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -225,6 +225,29 @@ function takes_many_args(
 ]]>
     </programlisting>
    </example>
+   <para>
+    As of PHP 8.0.0, passing optional arguments after mandatory arguments
+    is deprecated. This can generally be resolved by dropping the default value.
+    One exception to this rule are parameters of the form
+    Type $param = &null;, where the &null; default makes the type implicitly
+    nullable. This usage remains allowed, though it is recommended to use an
+    explicit nullable type instead.
+   </para>
+   <example>
+    <title>Passing optional arguments after mandatory arguments</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function foo($a = [], $b) {} // Before
+function foo($a, $b) {}      // After
+
+function bar(A $a = null, $b) {} // Still allowed
+function bar(?A $a, $b) {}       // Recommended
+?>
+]]>
+    </programlisting>
+   </example>
+
    <sect2 xml:id="functions.arguments.by-reference">
     <title>Passing arguments by reference</title>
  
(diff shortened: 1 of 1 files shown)


language/functions.xml
```diff
@@ -225,6 +225,29 @@ function takes_many_args(
 ]]>
     </programlisting>
    </example>
+   <para>
+    As of PHP 8.0.0, passing optional arguments after mandatory arguments
+    is deprecated. This can generally be resolved by dropping the default value.
+    One exception to this rule are parameters of the form
```
Girgias, line 231:
```suggestion
    One exception to this rule are arguments of the form
```
To keep this consistent with the beginning of the paragraph?

language/functions.xml
```diff
@@ -225,6 +225,29 @@ function takes_many_args(
...
+    Type $param = &null;, where the &null; default makes the type implicitly
```
Girgias, line 232:
```suggestion
    <code>Type $param = &null;</code>, where the &null; default makes the type implicitly
```
Maybe? @cmb69 might want to confirm is that would work

language/functions.xml
```diff
@@ -556,6 +579,111 @@ echo sum(1, 2, 3, 4);
...
 
    </sect2>
+
+   <sect2 xml:id="functions.named-arguments">
+    <title>Named Arguments</title>
+
+    <para>
+     PHP 8.0.0 introduced named arguments as an extension of the existing
+     positional parameters. Named arguments allow passing arguments to a
+     function based on the parameter name, rather than the parameter position.
```
Girgias, line 589:
```suggestion
     function based on the argument name, rather than the argument position.
```
Again for consistency with the beginning of the paragraph

language/functions.xml
```diff
@@ -556,6 +579,111 @@ echo sum(1, 2, 3, 4);
...
+     This makes the meaning of the argument self-documenting, makes the
+     arguments order-independent and allows skipping default values arbitrarily.
+    </para>
+
+    <para>
+     Named arguments are passed by prefixing the value with the parameter name
```
Girgias, line 595:
```suggestion
     Named arguments are passed by prefixing the value with the argument name
```

language/functions.xml
```diff
@@ -556,6 +579,111 @@ echo sum(1, 2, 3, 4);
...
+      <![CDATA[
+<?php
+array_fill(value: 50, num: 100, start_index: 0);
+?>
+]]>
+     </programlisting>
+    </example>
+
+    <para>
+     You can combine named arguments with positional arguments. In this case,
```
Girgias, line 647:
```suggestion
     Named arguments can be combined with positional arguments. In this case,
```
Usage of "you" in the docs is to be limited to the tutorial section for the most part

language/functions.xml
```diff
@@ -556,6 +579,111 @@ echo sum(1, 2, 3, 4);
...
+]]>
+     </programlisting>
+    </example>
+
+    <para>
+     The order in which the named arguments are passed does not matter.
+    </para>
+
+    <example>
+     <title>Same as above example with a different order of parameters</title>
```
Girgias, line 636:
```suggestion
     <title>Same example as above with a different order of parameters</title>
```

Girgias (changes requested):
Looks good, other than the mix of parameter and arguments, which we might need to figure out when to use which in what context and try to normalize this in the whole docs :(

language/functions.xml
Girgias: So I might have been getting the correctness wrong looking at: https://github.com/php/doc-en/pull/241#issuecomment-735455648, so maybe ignore these comments.

ekinhbayar: In this specific example, I think parameters is the correct wording indeed since it's about the function signature which is the place to use the word parameter, though I wouldn't be surprised if I had gotten it wrong elsewhere in the PR :-)

language/functions.xml
ekinhbayar: I'd be happy to do this, I wasn't sure if that'd undo the use of `&null;` tbh. I could totally use `null` instead, if that would be the case.

language/functions.xml
ekinhbayar: I think this one is a legit use of argument, as it's again talking about the name vs position of the parameter at the function definition.

language/functions.xml
ekinhbayar: Same as above. If you do agree would you please mark these cases as resolved?

ekinhbayar:
@Girgias  Feel free to create an issue for `argument` vs `parameter` usage throughout the docs and even assign it to me. I can work on this over time in a single PR. Happy to make sure we keep this consistent across the docs :-) I'll commit the other changes shortly.

language/functions.xml
cmb69: It seems to me that `<code><replaceable>Type</replaceable> $param = &null;</code>` might be the proper markup, and might render as desired as well (fingers crossed!)

ekinhbayar:
Yup, by the looks of the failure, I do need to use null explicitly in the code block.

language/functions.xml
Girgias: *well I can't*

ekinhbayar:
@cmb69 @Girgias, is there anything I can do here to help closing/completing this one?

PR #252, opened 2020-11-29
utf8 has been deprecated for past 10 years

kamil-tekiela:
Following MySQL's recommendation from here https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html I changed the examples to talk about `utf8mb4` instead. 
As we know, MySQL deprecated the ill-conceived `utf8` charset in 2010.  Some explanation can be found in this document https://mysqlserverteam.com/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/


cmb69:
Thank you!

PR #253, opened 2020-11-29
Remove PHP 5 from mysqli, PDO_MySQL and mysqlnd

kamil-tekiela:
I was not sure what to do about https://www.php.net/manual/en/mysqli.installation.php

### DIFF
--- a/reference/mysqli/book.xml
+++ b/reference/mysqli/book.xml
@@ -60,6 +60,12 @@
  &reference.mysqli.mysqli-sql-exception;
  &reference.mysqli.reference;
 
+<appendix xmlns="http://docbook.org/ns/docbook" xml:id="changelog.mysqli">
+ <title>&ChangelogListingTitle;</title>
+ <para>&ChangelogListingDescription;</para>
+ <?phpdoc generate-changelog-for="ref.mysqli class.mysqli class.mysqli-stmt class.mysqli-result class.mysqli-driver class.mysqli-warning class.mysqli-sql-exception"?>
+</appendix>
+
 </book>
 <!-- Keep this comment at the end of the file
 Local variables:
--- a/reference/mysqli/book.xml
+++ b/reference/mysqli/book.xml
@@ -60,12 +60,6 @@
  &reference.mysqli.mysqli-sql-exception;
  &reference.mysqli.reference;
 
-<appendix xmlns="http://docbook.org/ns/docbook" xml:id="changelog.mysqli">
- <title>&ChangelogListingTitle;</title>
- <para>&ChangelogListingDescription;</para>
- <?phpdoc generate-changelog-for="ref.mysqli class.mysqli class.mysqli-stmt class.mysqli-result class.mysqli-driver class.mysqli-warning class.mysqli-sql-exception"?>
-</appendix>
-
 </book>
 <!-- Keep this comment at the end of the file
 Local variables:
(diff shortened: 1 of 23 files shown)


afilina (changes requested):
Besides one nitpicky suggestions, LGTM.

reference/mysqli/constants.xml
```diff
@@ -68,15 +66,14 @@
...
-      Only valid for mysqlnd. Available since PHP 5.3.0.
+      Only valid for mysqlnd.
      </para>
     </listitem>
    </varlistentry>
    <varlistentry xml:id="constantmysqli-opt-ssl-verify-server-cert">
     <term><constant>MYSQLI_OPT_SSL_VERIFY_SERVER_CERT</constant></term>
     <listitem>
      <para>
-      Available since PHP 5.3.0. (MySQL 5.1.10 and up)
```
cmb69, line 79: Sadly, PHP 7.0.0 is supposed to [work with MySQL >= 4.1.13 and MySQL >= 5.0.7](https://github.com/php/php-src/blob/php-7.0.0/ext/mysqli/config.m4#L87), so please keep the relevant info for MySQL.

reference/mysqli/book.xml
```diff
@@ -60,12 +60,6 @@
  &reference.mysqli.mysqli-sql-exception;
  &reference.mysqli.reference;
 
-<appendix xmlns="http://docbook.org/ns/docbook" xml:id="changelog.mysqli">
- <title>&ChangelogListingTitle;</title>
- <para>&ChangelogListingDescription;</para>
- <?phpdoc generate-changelog-for="ref.mysqli class.mysqli class.mysqli-stmt class.mysqli-result class.mysqli-driver class.mysqli-warning class.mysqli-sql-exception"?>
-</appendix>
```
cmb69, line 67: Even if there are no respective changelogs *currently*, I think we should leave that appendix (may comment it out).

cmb69 (commented):
Thanks! Overall a very nice cleanup. Please check my review comments.

reference/mysqli/book.xml
kamil-tekiela: Are you sure about this? I was really surprised when I discovered [this section](https://www.php.net/manual/en/changelog.mysqli.php). I believe that all changes should be logged under the functions that they belong too. Not many people will go looking through the manual to find some hidden page that contains a list of all the changes. Besides, it looks like nobody was maintaining it anyway. If we leave it there, how will we make sure that people update them when they update the manual? I haven't found such section in other books in the manual.

reference/mysqli/constants.xml
kamil-tekiela: I'll review it over the weekend and add it back in. Thanks!

reference/mysqli/book.xml
cmb69: That page is autogenerated by the `<?php generate-changelog-for?>` processing instruction. IMO a nice feature, that can be helpful to get an overview.

reference/mysqli/constants.xml
kamil-tekiela: Added the note + 2 other back in.

reference/mysqli/book.xml
kamil-tekiela: Reverted the commit

afilina:
Checked the new commits. Looks like all suggestions were addressed.

cmb69:
Thank you!

PR #255, opened 2020-11-30
Change instances of oop5 to oop

tiffany-taylor:
This changes instances of oop5 to oop.
This includes text within XML files, renaming the directory language/oop5/ to language/oop/, and renaming oop5.xml to oop.xml.
This PR is dependent on a change to the doc-base repo which also updates oop5 to oop. I will make a separate PR for that.


language/operators.xml
```diff
@@ -1369,7 +1369,7 @@ Expression: 0 = -4 << 62
     &reftitle.seealso;
     <para>
      <simplelist>
-<!-- <link linkend="language.oop5.basic.class.class">::class</link> -->
+<!-- <link linkend="language.oop.basic.class.class">::class</link> -->
```
Girgias, line 1372: This should just be removed, that's a leftover from one of my copy/pastes

Girgias (commented):
LGTM if this builds with the doc-base changes, as this is basically just a search and replace.

tiffany-taylor:
Something that occurred to me, since some of these changes will be unnecessary when the other PHP 5 removal PRs are merged, some of the files I've changed in this PR (e.g. migration guides) will be removed. Should I remove those changes from this PR?

afilina:
I believe that this PR is mergeable as long as it's rebased to take care of conflicts and recently removed files.

cmb69:
Translators will hate us for merging this.

afilina:
Can't we reproduce the same on the translation repos? It's the kind of change that doesn't require actually speaking the language.

Girgias:
> 
> 
> Can't we reproduce the same on the translation repos? It's the kind of change that doesn't require actually speaking the language.

The issue is tracking the revision changes. As an automatic "fix" which consists of moving files when the translation of said files is not up to date is basically useless, so one would need to ensure all translations have all files in oop5/ up to date.

And then there is the story about changing the IDs, that can be done but one cannot update the revision/commit hash for the same reasons as above

afilina:
Pinging @tiffany-taylor in case this fell through the cracks

tiffany-taylor:
> Pinging @tiffany-taylor in case this fell through the cracks

RL is currently taking most of my attention. I will try to come back to this in a few weeks.

tiffany-taylor:
- [ ] Set up redirection for oop5 to oop in php/web-php

tiffany-taylor:
With regards to https://github.com/php/doc-en/pull/255#issuecomment-757009800, if a translation doc repo is out of date, what is the best course of action for me to mitigate the translation maintainers workload with this change?

afilina:
This PR seems ambitious and has many merge conflicts after 2 years. How valuable is this change compared to the effort and possible disruption for translators? I feel like something so big requires a strategy before we make the change. What are your thoughts?

Girgias:
> This PR seems ambitious and has many merge conflicts after 2 years. How valuable is this change compared to the effort and possible disruption for translators? I feel like something so big requires a strategy before we make the change. What are your thoughts?

I still think this is a valuable change, but as with every QA change of that sort, leaving it to be done around spring is usually the best as *hopefully* translations will be up to date in regards to the oop sections as not that much changes should happen now that most of the updates for 8.2 should have landed.

But will need some coordinations with the translations (RU: @saundefined JP: @mumumu FR: myself, and the other languages I don't really know who's in charge)

afilina:
I'll set myself a reminder for March 15th :)

Girgias:
> @tiffany-taylor Is this still something that you want to try to complete? It's been sitting here for some time. Perhaps the sunk cost fallacy is at play here :)

It's not a very difficult change to enact, it's just taxing on translations. And trying to find the correct moment for this is always a bit tricky

PR #257, opened 2020-12-02
Add createFromInterface() for DateTime objects.

Crell:
Basically the same file for both.  Hopefully I caught all the places to reverse the class names... :smiley:

### DIFF
--- a/reference/datetime/datetime/createfrominterface.xml
+++ b/reference/datetime/datetime/createfrominterface.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+
+<refentry xml:id="datetime.createfrominterface" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>DateTime::createFromInterface</refname>
+  <refpurpose>Returns new DateTime object encapsulating the given DateTimeInterface object</refpurpose>
+ </refnamediv>
+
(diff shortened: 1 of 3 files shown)


reference/datetime/versions.xml
```diff
@@ -12,6 +12,7 @@
  <function name="datetime::add" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
  <function name="datetime::createfromformat" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
  <function name="datetime::createfromimmutable" from="PHP 7 &gt;= 7.3.0"/>
+ <function name="datetime::createfrominterface" from="PHP 8 &gt;= 8.0.0"/>
```
cmb69, line 15: No need to add any x.0.0 revision:
```suggestion
 <function name="datetime::createfrominterface" from="PHP 8"/>
```

reference/datetime/datetime/createfrominterface.xml
```diff
@@ -0,0 +1,85 @@
...
+$date = new DateTime("2014-06-20 11:45 Europe/London");
+$also_mutable = DateTime::createFromInterface($date);
+?>
+]]>
+    </programlisting>
+   </example>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
```
cmb69, line 58: The returnvalues sections should be placed before the examples section (I've noticed that this is not the case for some other date related pages, what should ideally be fixed sometime).

cmb69 (approved):
Thank you!

I left two comments regarding changes which I did right away. :)

PR #258, opened 2020-12-02
Add fdiv()

Crell:


reference/math/functions/fdiv.xml
```diff
@@ -0,0 +1,114 @@
...
+   </example>
+   <example>
+    <title>Using <function>fdiv</function> with possible zero divisor</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$result = match($res = fdiv(5, $user_data)) {
+  INF, -INF, NAN => 0, // This may be correct in context, or throw an exception.
+  default => $res,
+};
```
nikic, line 76: This example doesn't look very useful. If you wanted to treat INF/NAN specially, you'd just use normal division.

I think it would be more helpful to show different results here, e.g.:

```
var_dump(fdiv(5.7, 1.3)); // float(4.384615384615385)
var_dump(fdiv(4, 2)); // float(2)
var_dump(fdiv(1.0, 0.0)); // float(INF)
var_dump(fdiv(-1.0, 0.0)); // float(-INF)
var_dump(fdiv(0.0, 0.0)); // float(NAN)
```

Crell: Replaced the first example with that.

I still think it's useful to show how to handle different returns from fdiv.  Folding to 0 may not be the best answer, but just letting an INF or NAN flow through later math and do who knows what seems like a bad idea.

reference/math/functions/fdiv.xml
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.fdiv" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>fdiv</refname>
+  <refpurpose>Returns the division of two numbers</refpurpose>
```
cmb69, line 6: I think "division" is not the right term here; "quotient" appears to be more suitable. However, why not simply
```suggestion
  <refpurpose>Divides two numbers</refpurpose>
```

reference/math/versions.xml
```diff
@@ -25,6 +25,7 @@
  <function name="expm1" from="PHP 4 &gt;= 4.1.0, PHP 5, PHP 7"/>
  <function name="floor" from="PHP 4, PHP 5, PHP 7"/>
  <function name="fmod" from="PHP 4 &gt;= 4.2.0, PHP 5, PHP 7"/>
+ <function name="fdiv" from="PHP 8 &gt;= 8.0"/>
```
cmb69, line 28: I think this should be sorted alphabetically.

reference/math/functions/fdiv.xml
cmb69: IMO this example should throw.

cmb69 (commented):
Thanks for the PR!

reference/math/functions/fdiv.xml
nikic: I still don't think the match example should be here at all. It does not reflect typical usage, and doesn't do what you probably intended it to do (keep in mind that NAN === NAN is false, so NAN will always go to the default case).

Crell: ... Because of course NAN would not do anything logical. If it did, it wouldn't be NAN. :smile:

PR #259, opened 2020-12-02
Fix Imagick::getImageChannelMean() return value

BenMorel:


cmb69:
cc @Danack

cmb69:
Thank you!

PR #262, opened 2020-12-02
ImagickPixel::getColor() takes an int, not a bool

BenMorel:
Verified in source:

https://github.com/Imagick/imagick/blob/06116aa24b76edaf6b1693198f79e6c295eda8a9/imagickpixel_class.c#L605

(This should definitely be a float, though! Not sure if this can be changed now.)

### DIFF
--- a/reference/imagick/imagickpixel/getcolor.xml
+++ b/reference/imagick/imagickpixel/getcolor.xml
@@ -26,8 +26,46 @@
      <term><parameter>normalized</parameter></term>
      <listitem>
       <para>
-       Normalize the color values
+       Normalize the color values. Possible values are 0, 1 or 2.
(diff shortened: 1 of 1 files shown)


cmb69:
From looking at the implementation, the `normalization` parameter (docs call it `normalized`) is an integer enum value (0, 1 and 2 are supported; other values just return an empty array what appears to be an implementation glitch).

Could you please add the information from https://github.com/Imagick/imagick/blob/06116aa24b76edaf6b1693198f79e6c295eda8a9/imagickpixel_class.c#L591-L596 to the docs?

BenMorel:
@cmb69 Done!

cmb69:
Thank you!

PR #265, opened 2020-12-03
Generate string methodsynopses based on stubs

kocsismate:


cmb69:
Thanks, @kocsismate! I have applied most of the changes, but left some unresolved for now.

PR #266, opened 2020-12-03
Generate array methodsynopses based on stubs

kocsismate:


cmb69:
Thank you for the PR!

I'm going to merge everything that doesn't use `$rest` parameters. I'm not convinced that the proper PHP 8 signature is an improvement regarding the docs. It seems to me the proper solution is to move the callback to the beginning of the parameter list, i.e. introduce new functions with the "proper" parameter order, and let the old ones slowly fade out (RFC!) And yeah, there's `array_multisort()` …

There is also an issue regarding `array_walk()` (and maybe others): there is no diff in this PR, altough the docs say 1st param is `array`, but stubs say first param is `array|object`.

Girgias:
I'm also not convinced about the stub version for these special functions, maybe it's better to have a comment about how they are unsuable with named params?

Girgias:
I think other than ``array_multisort()`` using ``rest`` is OK, WDYT @cmb69 ?

cmb69:
> I think other than array_multisort() using rest is OK, […]

IMO, no. For instance, `array_diff_ukey()` is currently `array_diff_ukey(array $array, array ...$arrays, callable $key_compare_func): array`, but would change to `array_diff_ukey(array $array, array|callable ...$rest): array`. That is less specific, and even confusing (can I pass further arrays after the callback?) The price of having a "proper" signature would be the need for more explanatory text.

I'd rather exclude such functions from `gen_stub.php --generate-methodsynopses` altogether (and in the long run from PHP; the parameter order is just wrong).

Girgias:
I forgot the whole trailing callback issue >_>

kocsismate:
I'm closing this PR.

PR #267, opened 2020-12-03
Add new IntlDateFormatter constants.

Crell:
I'm not certain of the "right" way to denote the version for these.  I'd think a Changelog block, but that looks non-small to add from a quick glance at other pages.

### DIFF
--- a/reference/intl/dateformatter-constants.xml
+++ b/reference/intl/dateformatter-constants.xml
@@ -52,6 +52,50 @@
      <simpara>Most abbreviated style, only essential data (12/13/52 or 3:30pm)</simpara>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="intldateformatter.constants.relative_full">
+    <term>
+     <constant>IntlDateFormatter::RELATIVE_FULL</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>The same as <constant>IntlDateFormatter::FULL</constant>, but yesterday, today, and tomorrow
+      show as <literal>yesterday</literal>, <literal>today</literal>, and <literal>tomorrow</literal>,
+      respectively.</simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="intldateformatter.constants.relative_long">
+    <term>
+     <constant>IntlDateFormatter::RELATIVE_LONG</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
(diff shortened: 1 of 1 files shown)


cmb69:
Usually, for constants there is just a "Available as of PHP x.y.z." sentence at the end of the description.

Crell:
Works for me.  Updated.

reference/intl/dateformatter-constants.xml
```diff
@@ -52,6 +52,50 @@
...
    </varlistentry>
+   <varlistentry xml:id="intldateformatter.constants.relative_full">
+    <term>
+     <constant>IntlDateFormatter::RELATIVE_FULL</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>The same as <constant>IntlDateFormatter::FULL</constant>, but yesterday, today, and tomorrow
+      show as <literal>yesterday</literal>, <literal>today</literal>, and <literal>tomorrow</literal>,
+      respectively.  Available as of PHP 8.0.0</simpara>
```
cmb69, line 63: This doesn't look right; see [IntlDateFormatter::FULL](https://3v4l.org/3cFHM) vs. [IntlDateFormatter::RELATIVE_FULL](https://3v4l.org/jENvu). Do you have a link to the RFC or PR which introduced this?

Crell: There was no RFC AFAIK.  I based this description off of the test file in https://github.com/php/php-src/pull/5105/files

cmb69: Thanks! Maybe @kocsismate can clarify?

kocsismate: I think the issue with the example is that `RELATIVE_*` constants can't be provided as the time format. This is what the documentation of the `timeStyle` parameter of `udat_open()` also suggests for me: https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/udat_8h.html#a4261e9fd5382197c3d86a00211c48079

cmb69: Thanks! https://3v4l.org/BqjZB confirms that. @Crell, could you add the info that these constants are only supported for *date* formats?

kocsismate: @cmb69 Thanks for noticing this gotcha, I forgot about this a long-long time ago :) And thank you, Larry, for providing the documentation :)

Crell: Pushed an update to further clarify.

PR #270, opened 2020-12-05
fixed description about passing [mandatory|optional] parameter order.

mumumu:
It is inconsistent with migration guide (*1).

(*1) https://www.php.net/manual/en/migration80.deprecated.php#migration80.deprecated.core


cmb69 (approved):
Good catch, thanks!

PR #271, opened 2020-12-05
Fix attribute code formatting and a typo in yaf

kamil-tekiela:
Also, fixes typo mentioned in the chat https://chat.stackoverflow.com/transcript/message/51088431

### DIFF
--- a/language/attributes.xml
+++ b/language/attributes.xml
@@ -54,43 +54,44 @@ class SetUp {}
 
 class CopyFile implements ActionHandler
 {
-      public string $fileName;
-      public string $targetDirectory;
+    public string $fileName;
+    public string $targetDirectory;
 
-      #[SetUp]
-      public function fileExists()
-      {
-          if (!file_exists($this->fileName)) {
-              throw new RuntimeException("File does not exist");
-          }
-      }
+    #[SetUp]
+    public function fileExists()
+    {
+        if (!file_exists($this->fileName)) {
+            throw new RuntimeException("File does not exist");
+        }
+    }
 
-      #[SetUp]
-      public function targetDirectoryExists()
-      {
-          mkdir($this->targetDirectory);
-      }
+    #[SetUp]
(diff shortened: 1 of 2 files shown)


language/attributes.xml
```diff
@@ -44,53 +44,50 @@
...
+    #[SetUp]
+    public function fileExists() {
+        if (!file_exists($this->fileName)) {
+            throw new RuntimeException("File does not exist");
+        }
+    }
+
+    #[SetUp]
+    public function targetDirectoryExists() {
+        mkdir($this->targetDirectory);
```
cmb69, line 69: Not sure if the `@` is used deliberately; maybe an already existing targetDirectory is fine?

cmb69 (commented):
Thank you! I have two concerns, but otherwise this looks fine.

language/attributes.xml
kamil-tekiela: Then it should be wrapped in `if (!file_exists($this->targetDirectory))`. An error suppression operator should almost always be avoided. If used, then a comment should be present explaining why this particular line can be implicitly trusted to be always right.

language/attributes.xml
```diff
@@ -44,53 +44,50 @@
      <programlisting role="php">
 <![CDATA[
 <?php
-interface ActionHandler
-{
+
+interface ActionHandler {
```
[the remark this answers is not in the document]
kamil-tekiela: Fixed. I changed the style to PEAR CS. FWIW the page was not consistent before and I was not sure which one is more appropriate so I went for the less verbose style.

language/attributes.xml
cmb69: Fair enough.

language/attributes.xml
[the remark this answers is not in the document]
cmb69: Ah, indeed, was inconsistent before; thanks for fixing!

PR #272, opened 2020-12-07
Add PhpToken reference

jiripudil:


reference/tokenizer/phptoken.xml
```diff
@@ -0,0 +1,135 @@
...
+  </section>
+
+  <section xml:id="phptoken.props">
+   &reftitle.properties;
+   <variablelist>
+    <varlistentry xml:id="phptoken.props.id">
+     <term><varname>id</varname></term>
+     <listitem>
+      <para>
+       One of the T_* constants, or an integer &lt; 256 representing a single-char token.
```
Girgias, line 80:
```suggestion
       One of the T_* constants, or an ASCII codepoint representing a single-char token.
```
Might be better? (I'm assuming that's where the 256 comes from)

jiripudil: Yes, definitely sounds better, thanks!

cmb69:
Very nice, thank you very much! The only change I'm going to apply on top of this is the use of `?string`; nullable types should be written as union types; it would be nice, though, if PhD would render them with the question mark.

PR #274, opened 2020-12-09
Fix instances of siginfo and signinfo being used interchangably on pcntl_signal page

lpd-au:
As far as I can tell, `siginfo` and `siginfo_t` are correct and `signinfo` and `signinfo_t` are typos. This simple PR corrects that.

For example, at https://www.php.net/manual/en/function.pcntl-signal.php, clicking `signo` in the handler callback synopsis jumps you to the anchor on line 57, whereas clicking `signinfo` does nothing (it should jump to line 65).

According to the RFC, the intended struct is https://www.mkssoftware.com/docs/man5/siginfo_t.5.asp.


cmb69:
Thank you!

PR #275, opened 2020-12-10
json_decode: better explanation and example of 'depth' parameter

IMSoP:
- Talk about "nesting depth" not "recursion depth", because
  whether or not JSON is recursive isn't really relevant here.
- Make clear that the example document has a depth of 4, which
  wasn't obvious to me looking at it.
- Use json_last_error_msg() to simplify example; it presumably
  didn't exist when this example was written.
- Remove confusing foreach loop in favour of two hard-coded calls;
  it took me ages to realise that range(4,3,-1) just meant [4,3]

### DIFF
--- a/reference/json/functions/json-decode.xml
+++ b/reference/json/functions/json-decode.xml
@@ -52,7 +52,7 @@
      <term><parameter>depth</parameter></term>
      <listitem>
       <para>
-       User specified recursion depth.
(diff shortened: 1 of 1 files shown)


cmb69:
Very nice, thank you!

It might be reasonable to improve the error message "Maximum stack depth exceeded" as well; should better be "Maximum nesting depth exceeded."

PR #276, opened 2020-12-10
set_error_handler: clarify callback parameters and PHP 8.0 changes

IMSoP:
Link to reference page from migration guide, and update that
reference page with the changes applied in 8.0.

Additionally, the wording of the callback description was slightly
confusing regarding parameters being "optional" and "removed". This
patch aims to make clear that it is PHP providing the values, and
the user is just choosing whether to accept them.

This was inspired by [this Stack Overflow question](https://stackoverflow.com/q/65201307/157957).

### DIFF
--- a/appendices/migration80/incompatible.xml
+++ b/appendices/migration80/incompatible.xml
@@ -116,7 +116,8 @@
     </listitem>
     <listitem>
      <para>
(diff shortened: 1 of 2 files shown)


cmb69:
Thank you!

PR #277, opened 2020-12-15
Remove PHP 5 mentions from OO section

Girgias:


language/oop5/interfaces.xml
```diff
@@ -216,7 +209,7 @@ class b implements a
      </programlisting>
    </example>
    <para>
-     An interface, together with type-hinting, provides a good way to make sure
+     An interface, together with type declaring, provides a good way to make sure
```
afilina, line 212:
```suggestion
     An interface, together with type declarations, provides a good way to make sure
```

language/oop5/properties.xml
```diff
@@ -4,16 +4,15 @@
...
-   <emphasis>fields</emphasis>, but for the purposes of this reference we will use
-   <emphasis>properties</emphasis>. They are defined by using one of the
-   keywords <literal>public</literal>, <literal>protected</literal>,
-   or <literal>private</literal>, optionally followed by a type declaration,
-   followed by a normal variable declaration. This declaration may
-   include an initialization, but this initialization must be a constant
-   value--that is, it must be able to be evaluated at compile time and
-   must not depend on run-time information in order to be evaluated.
+   Class member variables are called <emphasis>properties</emphasis>.
+   They may be referred to using other terms such as <emphasis>fields</emphasis>,
```
afilina, line 8: Nice catch! For those who don't know why the mention of `attributes` was removed here, it's because that now refers to a different concept and can no longer be used interchangeably: https://wiki.php.net/rfc/attributes_v2

language/oop5/properties.xml
```diff
@@ -23,21 +22,11 @@
...
-    issue an <constant>E_STRICT</constant> warning, but since PHP
-    5.1.3 it is no longer deprecated and does not issue the warning.
-   </para>
-   <para>
-    If you declare a property using <literal>var</literal> instead of
-    one of <literal>public</literal>, <literal>protected</literal>,
-    or <literal>private</literal>, then PHP 5 will treat the property
-    as if it had been declared as <literal>public</literal>.
+    An alternate, and not recommended way as it is to maintain backward
+    compatibility with PHP 4, of declaring class properties is by using
```
afilina, line 26:
```suggestion
    An alternative and not recommended way of declaring class properties, as it is to maintain backward
    compatibility with PHP 4, is by using
```

- Alternate != alternative.
- "An alternative and not recommended way of declaring class properties" should be together to ease comprehension.
- Moved the comma to properly delimit the non-essential clause regarding backwards compatibility.

language/oop5/properties.xml
```diff
@@ -101,44 +86,12 @@ EOD;
    </para>
   </note>
 
-  <sect2 xml:id="language.oop5.properties.heredoc-nowdoc">
-   <title>Heredoc and Nowdoc</title>
```
afilina, line 105: Was this section removed due to duplication? https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc
If that is the case, can we link to that page from here?

language/oop5/properties.xml
```diff
@@ -235,108 +188,9 @@ Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be ac
      </screen>
     </example>
    </para>
-
-   <sect3 xml:id="language.oop5.properties.typed-properties.valid-types">
-    <title>Valid property types</title>
```
afilina, line 240: Was this section removed because of duplication with this? https://www.php.net/manual/en/language.types.php
If that is the case, can we at least link to that page from here?

language/oop5/references.xml
```diff
@@ -3,14 +3,14 @@
...
-   One of the key-points of PHP 5 OOP that is often mentioned is that 
+   One of the key-points of PHP OOP that is often mentioned is that 
    "objects are passed by references by default". This is not completely true. 
    This section rectifies that general thought using some examples.
   </para>
 
   <para>
    A PHP reference is an alias, which allows two different variables to write
-   to the same value. As of PHP 5, an object variable doesn't contain the object
+   to the same value. In PHP an object variable doesn't contain the object
```
afilina, line 13:
```suggestion
   to the same value. In PHP, an object variable doesn't contain the object
```

language/oop5/traits.xml
```diff
@@ -3,7 +3,7 @@
  <sect1 xml:id="language.oop5.traits" xmlns="http://docbook.org/ns/docbook">
   <title>Traits</title>
   <para>
-   As of PHP 5.4.0, PHP implements a method of code reuse called Traits.
+   PHP implements a method of code reuse called Traits.
```
afilina, line 6:
```suggestion
   PHP implements a way to reuse code called Traits.
```

Might as well avoid ambiguity with "methods" :)

language/types/declarations.xml
```diff
@@ -46,50 +46,59 @@
        method is defined on.
        Can only be used in classes.
       </entry>
-      <entry></entry>
+      <entry/>
+     </row>
+     <row>
+      <entry><type>parent</type></entry>
+      <entry>
+       The property must be an &instanceof; the parent class of the class
```
afilina, line 54: Maybe a bit nitpicky, but multiple occurrences of "class" makes the sentence harder to understand. Perhaps we can assume that "parent" == "parent class" and simplify.

```suggestion
       The property must be an &instanceof; the parent of the class
```

afilina (changes requested):
Amazing work!
I have some suggestions and questions.

language/oop5/constants.xml
```diff
@@ -96,39 +68,13 @@ namespace foo {
 ]]>
   </programlisting>
  </example>
- <example>
-  <title>Constant expression example</title>
```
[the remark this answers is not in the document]
Girgias: Not really as I though this would be covered by the link to the constant page.
I can reintroduce it

language/oop5/properties.xml
Girgias: This might need to be moved up to a constant expression part (and we might want to introduce a separate page talking about this as this concept comes up quite often)

language/oop5/properties.xml
Girgias: The ``<xref linkend="language.types.declarations" />`` above should link to the page, might expanding on what you explicitly want?

language/oop5/constants.xml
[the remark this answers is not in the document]
afilina: Ah, I see. Yeah, there is an example of constant expression on this page: https://www.php.net/manual/en/language.constants.syntax.php

I really like the example on this page though, as it's specific to _class_ constants and makes this page more complete. It's just a personal preference though and I'm not married to it. It's really up to you.

language/oop5/properties.xml
afilina: Is there a place where we can log this as a TODO ticket? I'd like to take a stab at it at some point. In the mean time, we can definitely remove this duplicate example.

cmb69: I think filing a respective ticket at https://bugs.php.net could be useful.

afilina: Looks like it's already a thing: https://bugs.php.net/bug.php?id=68040
Bookmarked.

afilina:
LGTM.

language/oop5/constants.xml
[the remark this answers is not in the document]
Girgias: Reintroduced the example with some amendments, wdyt?

language/oop5/interfaces.xml
```diff
@@ -29,14 +29,6 @@
...
     if desired by separating each interface with a comma.
    </para>
-   <note>
-    <para>
-     Prior to PHP 5.3.9, a class could not implement two interfaces that
-     specified a method with the same name, since it would cause ambiguity.
-     More recent versions of PHP allow this as long as the duplicate methods
-     have the same signature.
-    </para>
-   </note>
```
Girgias, line 39: This note should probably be maintained in some sense, maybe as a Warning

Girgias:
I wonder if I shouldn't split the method signature section into a separate PR.

Girgias:
@cmb69 @afilina could you please review once again?

language/oop5/basic.xml
```diff
@@ -46,27 +46,14 @@ class SimpleClass
...
-    As of PHP 7.0.0 calling a non-static method statically from an incompatible
-    context results in $this being undefined inside the method. Calling a
-    non-static method statically from an incompatible context has been
-    deprecated as of PHP 5.6.0. As of PHP 7.0.0 calling a non-static method
-    statically has been generally deprecated (even if called from a compatible
-    context). Before PHP 5.6.0 such calls already triggered a strict notice.
+    method is called from within an object context.
+    <varname>$this</varname> is a reference to the calling object.
+    Prior to PHP 8.0.0, calling a non-static method statically would generate
+    a deprecation notice.
```
cmb69, line 52: It seems to me that we're swallowing to much relevant info. E.g. is calling a non-static method statically allowed as of PHP 8.0.0? What's the value of $this in this case prior to PHP 8.0.0?

afilina:
@Girgias No objections from me.

language/oop5/basic.xml
Girgias: Indeed, restored information.

language/oop5/references.xml
```diff
@@ -3,14 +3,14 @@
...
+   One of the key-points of PHP OOP that is often mentioned is that 
    "objects are passed by references by default". This is not completely true. 
    This section rectifies that general thought using some examples.
   </para>
 
   <para>
    A PHP reference is an alias, which allows two different variables to write
-   to the same value. As of PHP 5, an object variable doesn't contain the object
+   to the same value. In PHP, an object variable doesn't contain the object
    itself as value anymore. It only contains an object identifier which allows
```
cmb69, line 14:
```suggestion
   itself as value. It only contains an object identifier which allows
```

cmb69 (approved):
Thank you! A added a nit comment; otherwise LGTM (I'm fine with committing the TODO markers).

PR #278, opened 2020-12-16
Update SQL to use parameterised queries

craigfrancis:
It's really bad practice to concatenate user supplied variables into an SQL string. While the original did cast the variable to an int, it's very easy to make a mistake.

### DIFF
--- a/reference/mysqli/examples.xml
+++ b/reference/mysqli/examples.xml
@@ -93,7 +93,10 @@ echo "Sometimes I see " . $actor['first_name'] . " " . $actor['last_name'] . " o
 // Now, let's fetch five random actors and output their names to a list.
 // We'll add less error handling here as you can do that on your own now
 $sql = "SELECT actor_id, first_name, last_name FROM actor ORDER BY rand() LIMIT 5";
-if (!$result = $mysqli->query($sql)) {
+if ($stmt = $mysqli->prepare($sql)) {
+    $stmt->execute();
+    $result = $stmt->get_result();
+} else {
     echo "Sorry, the website is experiencing problems.";
     exit;
 }
--- a/reference/mysqli/examples.xml
+++ b/reference/mysqli/examples.xml
@@ -21,10 +21,8 @@
 <![CDATA[
 <?php
 // Let's pass in a $_GET variable to our example, in this case
-// it's aid for actor_id in our Sakila database. Let's make it
-// default to 1, and cast it to an integer as to avoid SQL injection
-// and/or related security problems. Handling all of this goes beyond
-// the scope of this simple example. Example:
+// it's aid for actor_id in our Sakila database. With a default
+// of 1. Example:
 //   http://example.org/script.php?aid=42
 if (isset($_GET['aid']) && is_numeric($_GET['aid'])) {
     $aid = (int) $_GET['aid'];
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR!

See also the related [discussion on the mailing list](https://marc.info/?l=phpdoc&m=160658331129934&w=2).

reference/mysqli/examples.xml
```diff
@@ -89,7 +93,10 @@ echo "Sometimes I see " . $actor['first_name'] . " " . $actor['last_name'] . " o
 // Now, let's fetch five random actors and output their names to a list.
 // We'll add less error handling here as you can do that on your own now
 $sql = "SELECT actor_id, first_name, last_name FROM actor ORDER BY rand() LIMIT 5";
-if (!$result = $mysqli->query($sql)) {
+if ($stmt = $mysqli->prepare($sql)) {
+    $stmt->execute();
+    $result = $stmt->get_result();
```
nikic, line 98: I don't think it makes sense to use prepare for a query without placeholders.

craigfrancis: I go on the basis that the query I write today will be edited by someone else later, and if they aren't familiar with prepared statements, by using `prepare()` for everything, there is a much better chance of a good outcome (that said, I wish prepare/bind_param/execute was easier to write).

cmb69:
Frankly, I don't see the point in making this example even more complex than it is now. Of course, building queries with user supplied input using string interpolation/concatenation is bad practice, but I see this example as some kind of started (but never completed) tutorial for using MySQLi. Thus I still think that we should either drop this example section altogether, or that we should expand it to a tutorial. A first step might be to keep the example as is (except some of the comments), and then follow up pointing out serious issues with the example, and showing better solutions (like using prepared statements or [proper output escaping](https://github.com/php/doc-en/pull/279).

craigfrancis:
Unfortunately writing safe mysqli code is complex, and while I want that API to change, it won’t happen any time soon.

Today I see a page on the official PHP website, that new developers are copying, that teaches them really bad coding practices.

If you want to delete that page, fine, but do it now.

If you want to do a re-write, go for it, but if you’re going down this route, at least apply this small patch now.

There are comments in the code already, and they *will* be ignored (I’ve been in the unfortunate position of telling far too many programmers that the code they have written/copied is fundamentally flawed, and I’m really disappointed to find this unsafe example code here).

reference/mysqli/examples.xml
kamil-tekiela: Ideally, everyone should use prepared statements 100% of the time. The only downside is that you have to add 2 more lines over `mysqli_query`. I agree that in a tutorial an example should be simple, but mysqli is not simple. If you want something simple then go for PDO. IMO that example fails to teach the proper mysqli usage. I already said that I am either way for improving or removing it altogether, but I would really recommend to remove it and the whole section from the manual.

cmb69:
> If you want to delete that page, fine, but do it now.

I think that is the best solution. Can I please have a straw poll (👍 or 👎)?

craigfrancis:
Thanks @cmb69, I think it looks pretty definitive (6 to delete).

I've also started a thread on PHP internals, and an RFC to hopefully address the difficulty in using parameterised queries in mysqli, https://news-web.php.net/php.internals/112618

cmb69:
Thanks for the reminder. I have [removed the example](http://svn.php.net/viewvc?view=revision&revision=352184).

kamil-tekiela:
Bug report updated. [#77531](https://bugs.php.net/bug.php?id=77531)

PR #279, opened 2020-12-16
Use HTML and URL encoding (minor tweak)

craigfrancis:
Considering the rest of the script assumes HTML output, we should escape the values as appropriate (I would prefer some kind of templating system that understands the appropriate context, but this will do for a simple example).

### DIFF
--- a/reference/mysqli/examples.xml
+++ b/reference/mysqli/examples.xml
@@ -76,7 +76,8 @@ if ($result->num_rows === 0) {
     // Oh, no rows! Sometimes that's expected and okay, sometimes
     // it is not. You decide. In this case, maybe actor_id was too
     // large? 
-    echo "We could not find a match for ID $aid, sorry about that. Please try again.";
+    echo "We could not find a match for ID " . htmlspecialchars($aid) . ". ";
+    echo "Sorry about that. Please try again.";
     exit;
 }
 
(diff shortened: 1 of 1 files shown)


reference/mysqli/examples.xml
```diff
@@ -97,8 +99,10 @@ if (!$result = $mysqli->query($sql)) {
 // Print our 5 random actors in a list, and link to each actor
 echo "<ul>\n";
 while ($actor = $result->fetch_assoc()) {
-    echo "<li><a href='" . $_SERVER['SCRIPT_FILENAME'] . "?aid=" . $actor['actor_id'] . "'>\n";
-    echo $actor['first_name'] . ' ' . $actor['last_name'];
+    $url = "./?aid=" . urlencode($actor['actor_id']);
+    $name = $actor['first_name'] . " " . $actor['last_name'];
+    echo "<li><a href='" . htmlspecialchars($url) . "'>";
```
cmb69, line 104: This is prone to XSS (if $url may contain single-quotes), because the default $flags of htmlspecialchars() are ENT_COMPAT, so single-quotes would not be escaped.

craigfrancis: Dam, you’re right, that will teach me for not checking how the core functions work (if you could fix, that would be great, otherwise I’ll do it tomorrow when I get back to my computer)... out of interest, why is ENT_COMPAT the default?

cmb69: > […] why is ENT_COMPAT the default?

I guess for backwards compatibility with decisions made long ago. `&apos;` is not HTML 4 conformant, and numeric character references may have been used only in later PHP versions – or something like that.

craigfrancis: Thanks @cmb69. As it looks like this page will be deleted, I'll leave this as-is. That said, I have started a thread on PHP internals about this issue, https://news-web.php.net/php.internals/112617

cmb69:
Yes, indeed, [this page has been removed](http://svn.php.net/viewvc?view=revision&revision=352184).

PR #282, opened 2020-12-22
Fixed oop return type of DateTimeInterface::getOffset.

mumumu:
It does not return false, but zero.

https://github.com/php/php-src/blob/caa710037e663fd78f67533b29611183090068b2/ext/date/php_date.c#L3099-L3130

Note: if this fix is correct, we need to fix ext/date/php_date.stub.php.

### DIFF
--- a/reference/datetime/datetimeinterface/getoffset.xml
+++ b/reference/datetime/datetimeinterface/getoffset.xml
@@ -13,15 +13,15 @@
   &reftitle.description;
   <para>&style.oop;</para>
   <methodsynopsis role="oop">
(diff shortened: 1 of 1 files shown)


cmb69 (commented):
Ah, good catch! Actually, these methods could have returned false prior to PHP 8.0.0, due to failing `DATE_CHECK_INITIALIZED`. So yes, php_date.stub.php should be fixed analogously.

cmb69 (approved):

mumumu:
closes by commit: https://github.com/php/doc-en/commit/260c0b8f4dd23dcf2336582ca6c2707b770c76dd

PR #284, opened 2020-12-23
Generate FFI methodsynopses based on stubs

kocsismate:


reference/ffi/ffi/alignof.xml
```diff
@@ -10,8 +9,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <modifier>static</modifier> <type>int</type><methodname>FFI::alignof</methodname>
-   <methodparam><type>mixed</type><parameter role="reference">ptr</parameter></methodparam>
+   <modifier>public</modifier> <modifier>static</modifier> <type class="union"><type>int</type><type>null</type></type><methodname>FFI::alignof</methodname>
+   <methodparam><type>object</type><parameter role="reference">ptr</parameter></methodparam>
```
cmb69, line 13: That matches the [stub](https://github.com/php/php-src/blob/959d3c9636fd070c554b281da221899638e2fe99/ext/ffi/ffi.stub.php#L39-L40), but is that stub correct?

Shouldn't the stub rather be `public static function alignOf(FFI\CData|FFI\CType ptr): int`? Also, I don't understand why the parameter is prefer-ref.

kocsismate: Ah, sure, you seem to be right! No idea though why prefer-ref is needed :/

kocsismate: I've created https://github.com/php/php-src/pull/6536

cmb69: Thanks! I left some comments there. Also I wonder why these functions accept `Z_PARAM_ZVAL` instead of `Z_PARAM_OBJ`.

kocsismate: Isn't it because of `ZVAL_DEREF`? nevertheless, we'll know more soon, hopefully :)

kocsismate: I've just applied the recent changes!

cmb69: Great, thanks!

And yes, `Z_PARAM_OBJ` can't work because the prefer-ref is needed.

cmb69:
Thank you! I've commit those changes which are not affected by https://github.com/php/php-src/pull/6543.

PR #288, opened 2020-12-23
Generate sodium methodsynopses based on stubs

kocsismate:
This one looks pretty straightforward despite the large number of changes, since only parameter names changed mostly.


cmb69:
Thank you!

PR #289, opened 2020-12-23
Generate some standard methodsynopses based on stubs

kocsismate:
These are mostly simpler cases (e.g. when only parameter names changed).

### DIFF
--- a/reference/math/functions/srand.xml
+++ b/reference/math/functions/srand.xml
@@ -8,12 +8,13 @@
  <refsect1 role="description">
   &reftitle.description;
    <methodsynopsis>
-    <type>void</type><methodname>srand</methodname>
-    <methodparam choice="opt"><type>int</type><parameter>seed</parameter></methodparam>
-   </methodsynopsis>
+   <type>void</type><methodname>srand</methodname>
+   <methodparam choice="opt"><type>int</type><parameter>seed</parameter><initializer>0</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer><constant>MT_RAND_MT19937</constant></initializer></methodparam>
+  </methodsynopsis>
   <para>
    Seeds the random number generator with <parameter>seed</parameter>
-   or with a random value if no <parameter>seed</parameter> is given.
+   or with a random value if <parameter>seed</parameter> is 0.
   </para>
 
   &note.randomseed;
@@ -99,7 +100,6 @@ $randval = rand();
   </para>
  </refsect1>
 </refentry>
-
 <!-- Keep this comment at the end of the file
 Local variables:
 mode: sgml
--- a/reference/math/functions/srand.xml
+++ b/reference/math/functions/srand.xml
@@ -14,7 +14,7 @@
   </methodsynopsis>
   <para>
    Seeds the random number generator with <parameter>seed</parameter>
-   or with a random value if <parameter>seed</parameter> is 0.
+   or with a random value if <parameter>seed</parameter> is <literal>0</literal>.
   </para>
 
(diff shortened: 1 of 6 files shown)


cmb69:
Thank you! I've [committed most of this](http://svn.php.net/viewvc?view=revision&revision=352191), but left a couple of functions for closer review.

Girgias:
Ping, this seems to have gotten stale

cmb69:
I've merged the changes to abs(), ceil(), floor() and round() now (e2a460ca39fece49eefdd5bd2282e5ba7b488fea).

afilina:
Besides the merge conflicts, what else is preventing this PR from being merged?

reference/info/functions/assert.xml
```diff
@@ -12,13 +12,13 @@
   <methodsynopsis>
    <type>bool</type><methodname>assert</methodname>
    <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
-   <methodparam choice="opt"><type>string</type><parameter>description</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>Throwable</type><type>string</type><type>null</type></type><parameter>description</parameter><initializer>&null;</initializer></methodparam>
```
kamil-tekiela, line 15: Why is this line changed? It looks like this `<methodsynopsis>` should be removed. I think it's better to leave the `assert` page out of this PR and then deal with it separately.

reference/info/functions/assert.xml
```diff
@@ -72,7 +72,7 @@
     argument will contain the file the assertion failed in. The
     second argument will contain the line the assertion failed on and
     the third argument will contain the expression that failed (if
-    any &#x2014; literal values such as 1 or "two" will not be passed via
+    any — literal values such as 1 or "two" will not be passed via
```
kamil-tekiela, line 75: Why this change?

reference/math/functions/srand.xml
```diff
@@ -8,12 +8,13 @@
...
-    <methodparam choice="opt"><type>int</type><parameter>seed</parameter></methodparam>
-   </methodsynopsis>
+   <type>void</type><methodname>srand</methodname>
+   <methodparam choice="opt"><type>int</type><parameter>seed</parameter><initializer>0</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer><constant>MT_RAND_MT19937</constant></initializer></methodparam>
+  </methodsynopsis>
   <para>
    Seeds the random number generator with <parameter>seed</parameter>
-   or with a random value if no <parameter>seed</parameter> is given.
+   or with a random value if <parameter>seed</parameter> is 0.
```
kamil-tekiela, line 17:
```suggestion
   or with a random value if <parameter>seed</parameter> is <literal>0</literal>.
```

kamil-tekiela (changes requested):
I think this can be merged but without `assert` changes. That page needs more work.

reference/info/functions/assert.xml
cmb69: The change itself is fine, but I think there shouldn't be an mdash. Maybe an ndash, or a semicolon, or some rewording is more appropriate. OTOH, this is not really related to this PR.

kocsismate:
I got rid of assert, rebased to master, and implemented the suggested change.

reference/info/functions/assert.xml
kocsismate: > Why this change?

gen_stub.php (ext/dom) changes these by some reason.

kamil-tekiela (approved):

kamil-tekiela:
Thanks. Looks good to me.

PR #291, opened 2020-12-28
ReflectionUnionType::getTypes() returns ReflectionNamedType[]

BenMorel:
AFAIK, `ReflectionUnionType::getTypes()` can only ever return instances of `ReflectionNamedType`.

### DIFF
--- a/reference/reflection/reflectionuniontype/gettypes.xml
+++ b/reference/reflection/reflectionuniontype/gettypes.xml
@@ -26,7 +26,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   An array of <classname>ReflectionType</classname> objects.
+   An array of <classname>ReflectionNamedType</classname> objects.
   </para>
  </refsect1>
 


cmb69:
Ah, right! I'm not sure whether we better leave that as is, though, for forward compatibility reasons.

BenMorel:
Is there any chance this can change in the future? Note that PHPStorm already documents it as [ReflectionNamedType[]](https://github.com/JetBrains/phpstorm-stubs/blob/master/Reflection/ReflectionUnionType.php) for static analysis purposes, and I'm planning to PR the same for Psalm, but I thought it would be the correct thing to do it here first.

cmb69:
Consider we may introduce intersection types in a future version. These would not be named types, but likely could be used with union types. I'm not saying that we shouldn't document the only valid return type exactly now, but that we should be careful about this. :)

BenMorel:
*If* this happens one day, can't we just change the doc then?

cmb69:
Well, any client code relying on ReflectionNamedType[] to be returned, could break then, e.g. https://3v4l.org/0hJrX.

BenMorel:
I understand. That being said, if intersection types were introduced, the BC break here could be documented.

I'll leave it up to you, in the meantime I've PR the changed with Psalm: https://github.com/vimeo/psalm/pull/4903

Let's see what they think about this, too.

nikic:
The return type should stay as-is.

PR #293, opened 2020-12-29
Make SplFixedArray intro say it can be resized

TysonAndre:
SplFixedArray::setSize exists. The name is misleading.

I think this should say `difference ... is` or `differences ... are`


cmb69 (approved):
Thank you!

PR #299, opened 2020-12-30
Enhancement: Switch from Travis CI to GitHub Actions

localheinz:
This PR

* [x] switches from Travis CI to GitHub Actions

💁‍♂️ You can see this in action at https://github.com/localheinz/doc-en/pull/1.


.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          path: "en"
+
+      - name: "Checkout salathe/phpdoc-base"
+        uses: "actions/checkout@v2"
+        with:
+          repository: "salathe/phpdoc-base"
```
cmb69, line 31: Any reason why you use salathe/phpdoc-base instead of php/doc-base?

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.2"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
```
localheinz, line 24: For reference, see https://github.com/actions/checkout.

.github/workflows/integrate.yaml
localheinz: @cmb69 

Thank you for taking a look - that was a leftover from my outdated fork. Updated!

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,38 @@
...
+name: "Integrate"
+
+on:
+  pull_request: null
+  push:
+    branches:
+      - "master"
+
+env:
+  LANGUAGE: "en"
```
localheinz, line 12: I have extracted an environment variable here, so it can be referenced in multiple places

- for the directory where to check out this repository
- for the command that runs to build the documentation

Does that make sense?

cmb69: I don't see any real advantage, but it certainly doesn't hurt either.

Anyhow, thanks a lot for the PR!

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,38 @@
+# https://docs.github.com/en/actions
+
+name: "Integrate"
```
localheinz, line 3: The name of the workflow appears in the Checks section:

CleanShot 2020-12-30 at 23 22 48

as well as in the Checks tab:

CleanShot 2020-12-30 at 23 24 12

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches:
+      - "master"
+
+env:
+  LANGUAGE: "en"
+
+jobs:
+  build:
+    name: "Build"
```
localheinz, line 16: The name of the job appears in the Checks section:

CleanShot 2020-12-30 at 23 22 48

as well as in the Checks tab:

CleanShot 2020-12-30 at 23 24 12

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.2"
```
Girgias, line 23: Any reason for picking 7.2?

localheinz: Yes, it is the same version that is currently used on Travis CI, see:

https://github.com/php/doc-en/blob/06e8e34994d0d5183cf43b72932a2ffa48406dce/.travis.yml#L5

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,44 @@
...
+          path: "${{ env.LANGUAGE }}"
+
+      - name: "Checkout php/doc-base"
+        uses: "actions/checkout@v2"
+        with:
+          repository: "php/doc-base"
+          path: "doc-base"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
```
localheinz, line 38: For reference, see https://github.com/shivammathur/setup-php.

If preferred, I can use the corresponding binary directly, but using this action, setting up PHP is more convenient.

.github/workflows/integrate.yaml
localheinz: Can we use a modern version?

Which one would you suggest?

nikic (dismissed):
Turns out we can't use GH actions right now, as we need someone to verify the @php-pulls email address. Marking as changes requested until we can get that done.

.github/workflows/integrate.yaml
Girgias: PHP 8.0 works as I've built the docs using that. :)

localheinz: Will update in a moment!

.github/workflows/integrate.yaml
localheinz: We could parse the language from the name of the repository, making the workflows identical across all variations!

cmb69: Would it be possible to simplify the individual actions for each translation by moving some common stuff to doc-base?

localheinz: @cmb69 

> Would it be possible to simplify the individual actions for each translation by moving some common stuff to doc-base?

Not that I can think of, but perhaps something will come up!

localheinz:
@nikic 

Thank you for the update!

.github/workflows/integrate.yaml
jhdxr: "ubuntu-latest" includes a pre-installed php already, and it's good enough to build the document already. I prefer to remove this part entirely to make the whole script simpler.

localheinz: @jhdxr

Adjusted!

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
+# https://docs.github.com/en/actions
+
+name: "Integrate"
+
+on:
+  pull_request: null
```
afilina, line 6: Unless I'm mistaken, this will only run when we push to master, but not when we open a PR, meaning that we can't see whether a PR breaks the build. I don't see the checks on this PR:
image

I would expect to see checks on a PR that first introduces GitHub actions. See checkmark next to a commit on this PR:
image

This can be solved by specifying nothing instead of `null`. Example on [roave/dont](https://github.com/Roave/Dont/blob/1.4.x/.github/workflows/phpunit.yml#L4)

afilina (changes requested):

.github/workflows/integrate.yaml
localheinz: @afilina

> Unless I'm mistaken, this will only run when we push to master, but not when we open a PR, meaning that we can't see whether a PR breaks the build. I don't see the checks on this PR:

Yes, you are mistaken. This will run jobs for pull requests.

The reason you are not seeing checks here is because a workflow for GitHub Actions has not been defined in this repository yet, and because I do not have write privileges. I am pushing from a fork.

As pointed out in the body of the pull request, you can see this in action in https://github.com/localheinz/doc-en/pull/1.

cmb69:
> […], as we need someone to verify the @php-pulls email address.

Who could do that?

nikic:
@cmb69 This is already done, so it's fine to merge.

localheinz:
Quick heads-up, the workflow is exactly the same as the one in https://github.com/php/doc-de/pull/2, with one slight difference:

```diff
--- doc-de/.github/workflows/integrate.yaml	2021-01-08 18:06:49.000000000 +0100
+++ doc-en/.github/workflows/integrate.yaml	2021-01-08 18:08:30.000000000 +0100
@@ -17,7 +17,7 @@
     strategy:
       matrix:
         language:
-          - "de"
+          - "en"

     steps:
       - name: "Checkout php/doc-${{ matrix.language }}"
```

The idea is to use workflows with the least differences across all repositories.

Instead of using a matrix, we could also parse the language identifier from the repository name, if that makes sense.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,43 @@
...
```
localheinz, line 6: @afilina 

Quick note on why I use 

```yaml
  pull_request: null
```

here: I use [`yamllint`](https://github.com/adrienverge/yamllint) in my repositories to lint YAML files, and the configuration I use complains when leaving it empty.

For an example, see https://github.com/ergebnis/php-library-template/blob/f6af5cb5c234a146f620061e95b34b432bf2e1c8/.github/workflows/integrate.yaml#L32-L37.

cmb69:
> The idea is to use workflows with the least differences across all repositories.

That makes sense.

> Instead of using a matrix, we could also parse the language identifier from the repository name, if that makes sense.

I think it's good as is.

Is this ready to be merged, @localheinz ?

.github/workflows/integrate.yaml
afilina: Thanks for clarifying :)

localheinz:
@cmb69 

👍

cmb69:
Thank you, @localheinz !

localheinz:
Thank you, @cmb69!

localheinz:
@cmb69 

Is there interest in adding more tooling to the workflows?

cmb69:
Well, it might be nice to actually build the documentation for pull request, so it could be viewed, but I have doubts that that's reasonably doable. Other than that, running ./configure as it's done now, is probably all we need. :)

localheinz:
@cmb69 

That sounds quite cool, actually, and should be possible!

PR #306, opened 2021-01-01
Improved parameter markup in strtok function description.

mumumu:
We use `<parameter>` tag to distinguish parameter "token" with the smaller strings token.
Ditto with `string` parameter.

### DIFF
--- a/reference/strings/functions/strtok.xml
+++ b/reference/strings/functions/strtok.xml
@@ -21,19 +21,19 @@
   <para>
    <function>strtok</function> splits a string (<parameter>string</parameter>)
    into smaller strings (tokens), with each token being delimited by any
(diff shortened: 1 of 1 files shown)


reference/strings/functions/strtok.xml
```diff
@@ -21,19 +21,19 @@
   <para>
    <function>strtok</function> splits a string (<parameter>string</parameter>)
    into smaller strings (tokens), with each token being delimited by any
-   character from <parameter>string</parameter>.
+   character from <parameter>token</parameter>.
```
mumumu, line 24: also, fixed invalid parameter replacement.

cmb69 (approved):
Thank you!

PR #307, opened 2021-01-02
Whitespace fix for datetime.getoffset

dcolt:
Fix for the missing space in for the getOffset function:
https://www.php.net/manual/en/datetime.getoffset
https://www.php.net/manual/en/class.datetime.php
https://www.php.net/manual/en/class.datetimeimmutable.php


cmb69 (approved):
Thank you!

PR #309, opened 2021-01-03
Generate openssl methodsynopses based on stubs

kocsismate:


reference/openssl/functions/openssl-cms-encrypt.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.openssl-cms-encrypt" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>openssl_cms_verify</refname>
+  <refpurpose>Verify signature</refpurpose>
```
kocsismate, line 6: the refpurpose of the new functions is not good yet... I've just realized that the refname here is also wrong.

kocsismate:
It's getting better! I've just added more change log entries, and removed more `resource` references. I and removed the stub documentation for `openssl_cms_*` functions which I added, as I've just noticed that had already been committed to master.

cmb69:
Thank you!

reference/openssl/functions/openssl-csr-get-public-key.xml
```diff
@@ -41,9 +41,32 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       On success, this function returns a <classname>OpenSSLAsymmetricKey</classname> instance now;
```
kocsismate, line 62: I've just noticed that I use the wrong article before `OpenSSLAsymmetricKey`. Thanks for fixing it (among other issues I suppose)

PR #310, opened 2021-01-04
Update openssl-get-cipher-methods.xml

fynntang:
Update PHP7.4 All methods under version


afilina:
LGTM.

- Since OpenSSL 1.1.1, the uppercase variants are no longer returned. 
- It makes sense to include the more recent cipher methods, since we're updating this page anyway.

cmb69:
Thank you! I don't think this PR is particularly useful, since the supported ciphers depend on the OpenSSL version (and likely on configuration options as well), and we also support LibreSSL, which may support different ciphers. OTOH, applying this PR doesn't hurt either.

> Since OpenSSL 1.1.1, the uppercase variants are no longer returned.

I'm going to document that.

PR #313, opened 2021-01-05
fix typo in arithmetic assignment operators

decima:
A small typo in documentation :)


decima:
Was reading the French one and saw it :D So I checked in English

PR #314, opened 2021-01-07
fix return type of array_change_key_case error

drealecs:
Fix return type of `array_change_key_case` in case of parameter wrong type that changed from false to null starting PHP 5.3 ~11.5 years ago
This can be checked here: https://3v4l.org/23Zcf


cmb69 (approved):
Thank you!

PR #315, opened 2021-01-09
Remove references to `mysql` extension and to versions below 7.0

afilina:
- Remove references to `mysql` extension and related examples.
- Unify and cleanup preserved examples.
- Add missing `mysqli` charset example.
- Remove link to `reference.mysql.book` so that we could eventually retire it.
- Remove references to old PHP versions, except where it made sense to keep it (I think it makes sense to keep "PHP version introduced" in case we add new extensions in the future).

Note: this PR only focuses on mysqlinfo pages in order to make this bug obsolete: https://bugs.php.net/bug.php?id=72158

### DIFF
--- a/reference/mysqlinfo/concepts.xml
+++ b/reference/mysqlinfo/concepts.xml
@@ -70,6 +70,7 @@ if ($uresult) {
    }
 }
 $uresult->close();
+?>
 ]]>
    </programlisting>
   </example>
@@ -88,6 +89,7 @@ if ($uresult) {
        echo $row['Name'] . PHP_EOL;
    }
 }
+?>
 ]]>
    </programlisting>
   </example>
@@ -134,6 +136,7 @@ $mysqli->set_charset('utf8mb4');
 
 // But, this will NOT affect it (UTF-8 vs utf8mb4) -- don't use dashes here
 $mysqli->set_charset('UTF-8');
+?>
 ]]>
    </programlisting>
   </example>
@@ -167,6 +170,7 @@ if (!$mysqli->set_charset('utf8mb4')) {
 }
 
 echo 'Your current character set is: ' . $mysqli->character_set_name() . "\n";
+?>
 ]]>
    </programlisting>
   </example>
@@ -177,6 +181,7 @@ echo 'Your current character set is: ' . $mysqli->character_set_name() . "\n";
 <![CDATA[
 <?php
 $pdo = new PDO("mysql:host=localhost;dbname=world;charset=utf8mb4", 'my_user', 'my_pass');
+?>
 ]]>
    </programlisting>
   </example>
--- a/reference/mysqlinfo/concepts.xml
+++ b/reference/mysqlinfo/concepts.xml
@@ -181,24 +181,6 @@ $pdo = new PDO("mysql:host=localhost;dbname=world;charset=utf8mb4", 'my_user', '
    </programlisting>
   </example>
 
-  <example>
-   <title>Setting the character set example: mysqli</title>
-   <programlisting role="php">
-<![CDATA[
-<?php
-$mysqli = mysqli_connect("localhost", "my_user", "my_pass", "world");
-
-echo 'Initial character set: ' . $mysqli->character_set_name() . "\n";
-
-if (!$mysqli->set_charset('utf8mb4')) {
-    echo "Error: Unable to set the character set.\n";
-    exit;
-}
-
-echo 'Your current character set is: ' . $mysqli->character_set_name() . "\n";
-]]>
-   </programlisting>
-  </example>  
  </section>
 
 </chapter>
--- a/reference/mysqlinfo/concepts.xml
+++ b/reference/mysqlinfo/concepts.xml
@@ -61,7 +61,7 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-$mysqli  = new mysqli("localhost", "my_user", "my_pass", "world");
+$mysqli  = new mysqli("localhost", "my_user", "my_password", "world");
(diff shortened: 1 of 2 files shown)


reference/mysqlinfo/concepts.xml
```diff
@@ -61,7 +61,7 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-$mysqli  = new mysqli("localhost", "my_user", "my_password", "world");
+$mysqli  = new mysqli("localhost", "my_user", "my_pass", "world");
```
kamil-tekiela, line 64: The parameter name is `$password` and we use `my_password` throughout the manual, can we keep it as it was?

reference/mysqlinfo/concepts.xml
```diff
@@ -192,43 +167,35 @@ if (!$mysqli->set_charset('utf8mb4')) {
...
 <?php
 $pdo = new PDO("mysql:host=localhost;dbname=world;charset=utf8mb4", 'my_user', 'my_pass');
-?>
 ]]>
    </programlisting>
   </example>
 
   <example>
-   <title>Setting the character set example: mysql</title>
+   <title>Setting the character set example: mysqli</title>
```
kamil-tekiela, line 185: I think that there already is mysqli charset example few lines above. This one should probably be removed.

reference/mysqlinfo/concepts.xml
```diff
@@ -192,43 +167,35 @@ if (!$mysqli->set_charset('utf8mb4')) {
 
 echo "New character set information:\n";
 print_r( $mysqli->get_charset() );
```
kamil-tekiela, line 169: Can this be changed to `character_set_name()`?

reference/mysqlinfo/set.xml
```diff
@@ -25,12 +25,10 @@
    <title>Introduction</title>
    <abstract>
     <para>
-     Depending on the version of PHP, there are either two or three PHP APIs
-     for accessing the MySQL database. PHP 5 users can choose between the
-     deprecated <link linkend="book.mysql">mysql</link> extension,
-     <link linkend="book.mysqli">mysqli</link>, or
-     <link linkend="ref.pdo-mysql">PDO_MySQL</link>. PHP 7 removes the mysql
-     extension, leaving only the latter two options.
+     There are several PHP APIs
```
kamil-tekiela, line 28: There are 2 in the standard. Can we change the sentence to avoid the word several?

kamil-tekiela (changes requested):

reference/mysqlinfo/concepts.xml
cmb69, line 185: This appears to duplicate the [mysqli example above](https://www.php.net/manual/en/mysqlinfo.concepts.charset.php#example-1631).

reference/mysqlinfo/set.xml
afilina: I changed to "several" to avoid chasing down occurrences of this whenever new APIs are added or removed in the future. I have seen this become a big maintenance burden over time in documentation in other projects.

reference/mysqlinfo/concepts.xml
afilina: In the same file, I have seen my_pass on many occasions. In any case, this was not the original intent of this PR, so I'll just undo this change to keep this PR more focused.

kamil-tekiela:
LGTM

reference/mysqlinfo/concepts.xml
```diff
@@ -70,7 +70,6 @@ if ($uresult) {
    }
 }
 $uresult->close();
-?>
```
cmb69, line 73: Well, PEAR CS mandates these, and the examples are still supposed to [adhere to that CS](http://doc.php.net/tutorial/style.php). IMO we should change that, but that would likely require discussion on the docs mailing list.

afilina: Alright, I changed all examples back to use closing tag. First I address the remaining 513 documentation bugs, then I fight for a different CS. Priorities :)

kamil-tekiela: I am not sure that the PEAR CS demands an empty statement or a closing PHP tag at the end of every PHP code block. I think it is a convention or tradition rather than a rule in CS. It certainly is not the best practice. It doesn't make much difference in PHP manual, but I don't think it is something we should enforce. 

IMHO this should be a topic of yet another discussion. If it is something that should be changed then we should do it rather than talk about it. Arguing about such small details leads to bikeshedding and in the meantime nothing gets done. 

However, it makes absolute no sense to remove `?>` and leave `$uresult->close();` there. Either remove both or leave both. Personal recommendation of mine would be to avoid using `mysqli_result::close()` at all times, but as we know the manual recommends using it many places for reasons unknown.

afilina: Closing connections can be critical in long-running processes. I just had a client last week with such a problem. Since it seems that more debate would be required and this change is not explicitly part of the PR, I'd shelf it for the time being.

kamil-tekiela: No, not closing the connection (although I do believe that should also be avoided but it's an entirely different topic). I meant closing `mysqli_result` which IMHO is absolutely useless. 

But, as much as I would like to complain, this is probably the only place in the whole manual where `mysqli_result::close()` is used correctly. Almost... That function is used on unbuffered results to consume internally all remaining records waiting on the wire. The problem in this example is that all rows have been read already, there is nothing else to consume.

afilina: I actually have a better reason to remove it: that's not the point of the example and the other example doesn't close results. Sent a commit.

cmb69:
Thank you!

PR #316, opened 2021-01-09
Make class type declaration context independent

Girgias:
Not exactly sure how to formulate this, but the current formulation is rather bad due to being context dependent (as it comes from a copy and paste)

### DIFF
--- a/language/types/declarations.xml
+++ b/language/types/declarations.xml
@@ -43,7 +43,7 @@
       <entry><type>self</type></entry>
       <entry>
        The value must be an &instanceof; the same class as the one the
-       method is defined on.
+       in which the type declaration is used.
        Can only be used in classes.
       </entry>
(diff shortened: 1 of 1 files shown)


language/types/declarations.xml
```diff
@@ -43,7 +43,7 @@
       <entry><type>self</type></entry>
       <entry>
        The value must be an &instanceof; the same class as the one the
```
afilina, line 45:
```suggestion
       The value must be an &instanceof; the same class as the one
```

afilina (changes requested):
Besides a word that should have been removed in the reformulation, I think that these changes make it clearer.

language/types/declarations.xml
Girgias: Thanks for noticing! :-)

PR #317, opened 2021-01-09
Renamed 7.0 classes migration to new-classes (same as 7.4 new-classes migration)

eusonlito:
`New Features` and `New Functions` sections have the `new-` prefix. Also PHP 7.4 `New Classes and Interfaces` is `new-classes`.

### DIFF
--- a/appendices/migration70.xml
+++ b/appendices/migration70.xml
@@ -12,7 +12,7 @@
  &appendices.migration70.deprecated;
  &appendices.migration70.changed-functions;
  &appendices.migration70.new-functions;
- &appendices.migration70.classes;
+ &appendices.migration70.new-classes;
  &appendices.migration70.constants;
  &appendices.migration70.sapi-changes;
  &appendices.migration70.removed-exts-sapis;
--- a/appendices/migration70/new-classes.xml
+++ b/appendices/migration70/new-classes.xml
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
 
-<sect1 xml:id="migration70.classes" xmlns:xlink="http://www.w3.org/1999/xlink">
+<sect1 xml:id="migration70.new-classes" xmlns:xlink="http://www.w3.org/1999/xlink">
  <title>New Classes and Interfaces</title>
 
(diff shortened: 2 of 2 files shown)


afilina:
Once we rename the file, will that cause problems with switching to the different translations on that page? I think that it makes sense to replicate this change in every translation repo at the same time. Also, what impact does that have on the translation tooling (question for maintainers)?

cmb69:
I don't think there'd be any issues regarding the tooling, but indeed, this change would break all translations. Also it would affect bookmarks. I'd prefer to leave this as is.

Girgias:
Yeah -1 from my side too, this doesn't provide anything really.

eusonlito:
Ok, no problem, but for bookmarks and sync with other languages, yesterday were removed 78 doc pages on PR https://github.com/php/doc-en/commit/027f187b3ed8cdec60db6f83baf8f62b536d1101 without be removed or updated other languages:

https://www.php.net/manual/en/appendices.php vs https://www.php.net/manual/es/appendices.php

Girgias:
> 
> 
> Ok, no problem, but for bookmarks and sync with other languages, yesterday has been removed 78 doc pages on PR [027f187](https://github.com/php/doc-en/commit/027f187b3ed8cdec60db6f83baf8f62b536d1101) without be removed or updated any other language:
> 
> https://www.php.net/manual/en/appendices.php vs https://www.php.net/manual/es/appendices.php

How is this relevant at all to this conversation. We removed PHP 5 docs, a version which is not supported any more by the project. The reason for not applying this to translations is that they need to do it themself to no break builds (and I've done it on the French version).

eusonlito:
Ok, sorry, PR closed.

In my opinion, not supported versions can (and should) remaining without problems into docs for reference like docs of many other languages.

Anyway, no problem 😊

Girgias:
> 
> 
> Ok, sorry, PR closed.
> 
> In my opinion, not supported versions can (and should) remaining without problems into docs for reference like docs of many other languages.
> 
> Anyway, no problem 😊

You can download the archive of PHP 5 (and PHP 4) on the website. Moreover, Zend has decided to maintain online docs for PHP 5 and PHP 4, but this is a Zend issue not a PHP project issue.

PR #318, opened 2021-01-09
Close bug #54863 by clarifying return values

afilina:
https://bugs.php.net/bug.php?id=54863

The issue is really that the docs should be more clear that non-SELECT queries will not let you fetch from the results. That should close the ticket.

The last commenter confirmed this behavior on PostgreSQL 9.2 + PHP 5.4/5.5. I confirmed this on PostgreSQL 9.6/12.5 + PHP 7.4


cmb69:
Thank you!

PR #319, opened 2021-01-09
Link to SplFileInfo from is_file()/file_exists() docs

TysonAndre:
And update the list of values that can be returned by getType based on
php 8.0 ext/standard/filestat.c

SplFileInfo is the only way to do some things, such as

1. Check for the existence of a file that may or may not be readable
   without emitting a notice
   that is caught by whatever error handler the project uses.

   (e.g. file_exists can return false for unix sockets)
2. Check for file types such as unix sockets
   (e.g. `$info->getType() === 'socket'`)

Recently, I was looking for how to check if a given path was a unix
socket and the filesystem functions didn't have any links to that
information.


cmb69:
Thank you!

PR #320, opened 2021-01-10
[WIP] Refactor error pages

Girgias:
This renames traditional errors to diagnostic errors
Attempts to unify Throwable Errors with Exceptions
Move try, catch, and finally control structures to the control structure section

Feedback on this is very welcome and I'm open to suggestion, as some of the text still needs to be updated/improved.
This is partly based on how [Rust](https://doc.rust-lang.org/book/ch09-00-error-handling.html) talks about error handling.

Some ideas I have also been considering:
 - Create a new page in the control structure section for ``throw``
 - Split the two section in ``error-handling.xml`` into two files in a new subfolder ``error-handling/``

A todo list of sort:
 - [ ] Trim exception page
 - [ ] Maybe move extending exception page/trim it?
 - [ ] More examples in each specific section.


Girgias:
I think this is now ready for review, I hope I didn't remove any information by splitting it up.

language/error-handling.xml
```diff
@@ -0,0 +1,304 @@
...
+  </note>
+  <warning>
+   <para>
+    <classname>Throwable</classname> objects cannot be cloned.
+    Attempting to <link linkend="language.oop5.cloning">clone</link> such an
+    object will result in an <classname>Error</classname> being thrown with
+    the following message:
+    <literal>Trying to clone an uncloneable object of class Exception</literal>.
+   </para>
+  </warning>
```
Girgias, line 75: I wonder if this warning is relevant here? As it would be pretty unusual to clone an exception, and this behaviour is described on their respective docs.

language/error-handling.xml
```diff
@@ -0,0 +1,304 @@
...
+    Another typical usage is to elevate diagnostic errors to exceptions with
+    <link linkend="class.errorexception">ErrorException</link>.
+   </para>
+   <example>
+    <title>Elevating diagnostic errors to exceptions</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function exceptions_error_handler($severity, $message, $filename, $lineno) {
+    throw new ErrorException($message, 0, $severity, $filename, $lineno);
```
Firehed, line 294: This implementation forgoes `if (error_reporting() & $severity)) { throw }`, which breaks `@`.

Girgias: Copy pasted this from the previous page, which was also broken then.
Will fix.

language/error-handling.xml
```diff
@@ -0,0 +1,304 @@
...
+     <link linkend="errorfunc.constants">severities</link>.
+    </para>
+   </listitem>
+  </itemizedlist>
+ </para>
+
+ <sect1 xml:id="language.error-handling.throwable">
+  <title><classname>Throwable</classname> errors</title>
+  <para>
+   Most errors in PHP are of this type.
```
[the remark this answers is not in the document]
Girgias: I can add an `As of PHP 8,` before it, but even in PHP 7 many of the warnings generated by the Engine are catchable `Error`s

language/control-structures/try-catch.xml
```diff
@@ -0,0 +1,240 @@
...
+    <programlisting role="php">
+     <![CDATA[
+<?php
+
+class A extends Exception {}
+class B extends Exception {}
+
+try {
+    throw new A;
+} catch(A) {
```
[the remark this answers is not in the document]
Girgias: I think I'll change it to a variable binding as re-ordering the sections seems pretty hard.

language/control-structures/finally.xml
```diff
@@ -0,0 +1,186 @@
...
+First finally.
+Caught error: Division by zero.
+Second finally.
+Hello World
+]]>
+  </screen>
+ </example>
+
+ <para>
+  When the call stack is unwound after a <classname>Throwable</classname> error
```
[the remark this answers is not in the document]
Girgias: I'll keep the technical language (as this is copied from the previous document) but will attempt to add a simpler explanation along side it.

language/error-handling.xml
```diff
@@ -0,0 +1,304 @@
...
+   <para>
+    Another typical usage is to elevate diagnostic errors to exceptions with
+    <link linkend="class.errorexception">ErrorException</link>.
+   </para>
+   <example>
+    <title>Elevating diagnostic errors to exceptions</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function exceptions_error_handler($severity, $message, $filename, $lineno) {
```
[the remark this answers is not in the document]
Girgias: Not sure the documentation should decide what are considered best practices, I'll try to emphasis this more however. :-)

appendices/migration70/incompatible/error-handling.xml
```diff
@@ -19,7 +19,7 @@
 
  <para>
   A fuller description of how errors operate in PHP 7 can be found
```
Crell, line 21: "Fuller" isn't good English here. 

```suggestion
  A more complete description of how errors operate in PHP 7 can be found
```

language/control-structures/finally.xml
```diff
@@ -0,0 +1,390 @@
+<?xml version="1.0" encoding="utf-8"?>
+<sect1 xml:id="control-structures.finally" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <title>finally</title>
+ <?phpdoc print-version-for="finally"?>
+ <para>
+  A &finally; block may also be specified after or instead of &catch; blocks.
+  Code within the &finally; block will always be executed after the &try; and
+  &catch; blocks, regardless of whether a <classname>Throwable</classname> error
+  has been thrown, and before normal execution resumes.
```
Crell, line 9: It also gets executed even if there is a return statement inside the try or catch, which is also worth mentioning explicitly.

language/control-structures/finally.xml
```diff
@@ -0,0 +1,390 @@
+<?xml version="1.0" encoding="utf-8"?>
+<sect1 xml:id="control-structures.finally" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <title>finally</title>
+ <?phpdoc print-version-for="finally"?>
+ <para>
```
Crell, line 5: I feel like it would be better to front-load the logic of how finally behaves vis a vis throws, returns, yields, exit, etc., perhaps with a bulleted list, so it's easier to reference.  Then proceed with a bunch of examples demonstrating it.

language/control-structures/try-catch.xml
```diff
@@ -0,0 +1,240 @@
...
+]]>
+   </screen>
+  </example>
+  <caution>
+   <para>
+    Throwing an object withing a &catch; block will <emphasis>not</emphasis>
+    be caught by an adjacent &catch; block.
+   </para>
+   <example>
+    <title>NEEDS TITLE</title>
```
Crell, line 181:
```suggestion
    <title>Sequential catch blocks</title>
```

language/error-handling.xml
```diff
@@ -0,0 +1,322 @@
...
+ </sect1>
+
+ <sect1 xml:id="language.error-handling.diagnostic">
+  <title>Diagnostic errors</title>
+  <para>
+   Diagnostic errors, also known as traditional errors, are used to signal a
+   number of different conditions, and can be displayed and/or logged as required.
+   By default PHP will report all diagnostic errors (prior to PHP 8.0,
+   <constant>E_NOTICE</constant> and <constant>E_DEPRECATED</constant>
+   diagnostics were not reported by default).
```
Crell, line 215:
```suggestion
   By default PHP will report all diagnostic errors.  (Prior to PHP 8.0,
   <constant>E_NOTICE</constant> and <constant>E_DEPRECATED</constant>
   diagnostics were not reported by default.)
```

Grammar fix.

language/operators.xml
```diff
@@ -1848,16 +1848,41 @@ echo $foo ?? $bar ?? $baz ?? $qux; // outputs 1
...
+    PHP supports one error control operator: the at sign (<literal>@</literal>).
+    When prepended to an expression in PHP, any diagnostic error that might
+    be generated by that expression will be suppressed.
    </simpara>
+   <para>
+    If a custom error handler function is set with
+    <function>set_error_handler</function>, it will still be called even though
+    the diagnostic has been suppressed, as such the custom error handler should
+    call <function>error_reporting</function> and verify that the
+    <literal>@</literal> operator was used in the following way:
```
Crell, line 1860:
```suggestion
    the diagnostic has been suppressed.  As a result, the custom error handler should
    call <function>error_reporting</function> and verify that the
    <literal>@</literal> operator was used in the following way:
```

Crell:
Various small issues noted above.

My biggest concern is that this puts exceptions under flow-of-control... but exceptions, as we are told repeatedly, are not meant to be used for flow of control.  Yet, they do change the flow of control.  My concern is that, without any statement to the contrary, this could tacitly imply that using exceptions the same way you would use if, foreach, or goto is a-OK, when that's very much not the case.

Otherwise, everything else here seems reasonable.

language/error-handling.xml
```diff
@@ -0,0 +1,322 @@
...
+    Another typical usage is to elevate diagnostic errors to exceptions with
+    <link linkend="class.errorexception">ErrorException</link>.
+   </para>
+   <example>
+    <title>Elevating diagnostic errors to exceptions</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function exceptions_error_handler($severity, $message, $filename, $lineno) {
+    if (error_reporting() & $severity)) {
```
Girgias, line 301:
```suggestion
    if (error_reporting() & $severity) {
```

language/error-handling.xml
afilina: I agree that this is a generic behavior that is already described in a central location. I see no reason to mention that in every place where it applies. Should the user see such an error, they will be able to find the correct docs through a search.

afilina:
@Girgias I resolved the outdated comments and provided my feedback on your own questions. I hope this helps to move this PR forward.

Girgias:
Thanks, I'll try to get back to this, but I need to sort out HTTP redirection for php-web and rebase, I should have time again in September. :-)

afilina:
@Girgias Is it worth to keep this PR open? This looks like a slice that is too big to be feasible. It can be referred to in a ticket (todo) instead, so that you don't lose sight of some of the wording that might be reusable.

Girgias:
> @Girgias Is it worth to keep this PR open? This looks like a slice that is too big to be feasible. It can be referred to in a ticket (todo) instead, so that you don't lose sight of some of the wording that might be reusable.

I will try to work on this PR this months as I'm planning to start working on the Migration guide and do more doc works this month (as I also won't have access to my desktop with 24 cores which makes php-src work more annoying)

PR #322, opened 2021-01-10
Improvements to stat() - document other file types

TysonAndre:
Document character device and sockets (this info is also seen in user-submitted notes)

Link to SplFileInfo.
Unlike stat(), SplFileInfo throws instead of emitting a notice,
which is convenient for avoiding the error handler.
(file_exists() returns files that exist but are unreadable)

Related to GH-319

The mode documentation was introduced in 3c8132d8240441503401c9e66ae23b15067ee2f1

### DIFF
--- a/reference/filesystem/functions/stat.xml
+++ b/reference/filesystem/functions/stat.xml
@@ -161,6 +161,10 @@
      </thead>
      <tbody>
       <row>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

PR #323, opened 2021-01-10
Remove unnecessary try-catch statements from PDO manual

kamil-tekiela:
Since now PHP has exception mode set to default, the try-catch on the constructor page is useless. Previously, people have been catching these exceptions to silence them just like the rest of PDO functions. Now, there is no observable difference between the constructor or any other PDO function with default settings in terms of exception throwing. 

The idea is to make the examples simple and to the point without any unnecessary noise. The try-catch statements serve no visible purpose and without proper explanation they are just confusing to new developers. 

The information is not lost, as we have 2 dedicated pages in the manual for [connections](https://www.php.net/manual/en/pdo.connections.php) and [error handling](https://www.php.net/manual/en/pdo.error-handling.php). 

There were 3 other examples where the try-catch was unnecessarily used. These were even worse because there was no code to recover from the exceptional situation, there was only a print statement, and they were encompassing arbitrary code within a function. If code fails then it should be handled by the error handler if the code can't recover from it. A try-catch with no means to recover is just dirty code. 
The examples used 2 spaces for indentation, I changed it to 4. 
Now, the examples are clean and short, and they continue to demonstrate the idea behind them.

### DIFF
--- a/reference/pdo/pdo/construct.xml
+++ b/reference/pdo/pdo/construct.xml
@@ -118,7 +118,7 @@
   &reftitle.errors;
   <para>
    <function>PDO::__construct</function> throws a <classname>PDOException</classname> if the attempt
-   to connect to the requested database fails.
+   to connect to the requested database fails regardless of which <constant>PDO::ATTR_ERRMODE</constant> is currently set.
   </para>
  </refsect1>
 
@@ -134,11 +134,7 @@ $dsn = 'mysql:dbname=testdb;host=127.0.0.1';
 $user = 'dbuser';
 $password = 'dbpass';
 
-try {
-    $dbh = new PDO($dsn, $user, $password);
-} catch (PDOException $e) {
-    echo 'Connection failed: ' . $e->getMessage();
-}
+$dbh = new PDO($dsn, $user, $password);
 
 ?>
 ]]>
@@ -169,11 +165,7 @@ $dsn = 'uri:file:///usr/local/dbconnect';
 $user = '';
 $password = '';
 
-try {
-    $dbh = new PDO($dsn, $user, $password);
-} catch (PDOException $e) {
-    echo 'Connection failed: ' . $e->getMessage();
-}
+$dbh = new PDO($dsn, $user, $password);
 
 ?>
 ]]>
@@ -197,11 +189,7 @@ $dsn = 'mydb';
 $user = '';
 $password = '';
 
-try {
-    $dbh = new PDO($dsn, $user, $password);
-} catch (PDOException $e) {
-    echo 'Connection failed: ' . $e->getMessage();
-}
+$dbh = new PDO($dsn, $user, $password);
 
 ?>
 ]]>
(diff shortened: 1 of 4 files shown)


reference/pdo/pdo/construct.xml
```diff
@@ -118,7 +118,7 @@
   &reftitle.errors;
   <para>
    <function>PDO::__construct</function> throws a <classname>PDOException</classname> if the attempt
-   to connect to the requested database fails.
+   to connect to the requested database fails regardless of which <constant>PDO::ATTR_ERRMODE</constant> is currently set.
```
afilina, line 121:
```suggestion
   to connect to the requested database fails, regardless of which <constant>PDO::ATTR_ERRMODE</constant> is currently set.
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -215,35 +215,23 @@ kiwi
...
+    $sql = 'SELECT hand, won, bet FROM mynumbers ORDER BY BET';
     $stmt = $dbh->prepare($sql, array(PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL));
     $stmt->execute();
     while ($row = $stmt->fetch(PDO::FETCH_NUM, PDO::FETCH_ORI_NEXT)) {
-      $data = $row[0] . "\t" . $row[1] . "\t" . $row[2] . "\n";
-      print $data;
+        $data = $row[0] . "\t" . $row[1] . "\t" . $row[2] . "\n";
+        print $data;
     }
-    $stmt = null;
```
afilina, line 226: Good catch! There's no point in setting a local var to null right before the end of a method.

reference/pdo/pdostatement/setfetchmode.xml
```diff
@@ -102,21 +102,14 @@
...
 <![CDATA[
 <?php
-$sql = 'SELECT name, colour, calories FROM fruit';
-try {
-  $stmt = $dbh->query($sql);
-  $result = $stmt->setFetchMode(PDO::FETCH_NUM);
-  while ($row = $stmt->fetch()) {
+$stmt = $dbh->query('SELECT name, colour, calories FROM fruit');
+$result = $stmt->setFetchMode(PDO::FETCH_NUM);
+foreach ($stmt as $row) {
```
afilina, line 107: Although I prefer teaching `foreach` as well, didn't `$stmt->fetch()` make it a bit easier for people to wrap their head around the concept of "fetch" mode?

afilina (changes requested):
Agreed. I see way too much nonsensical error handling in legacy code, possibly inspired by such examples.

I just have a tiny correction and a couple of questions.

reference/pdo/pdostatement/setfetchmode.xml
kamil-tekiela: Hi Anna, Thanks for pointing it out and sorry for the delayed reply. This was changed on purpose. `fetch()` method already accepts fetch style as the first argument. The idea here was to show how `setFetchMode()` can be used when iterating over the statement with `foreach`. This is something that is not obvious, especially for someone migrating from mysqli. I would like to keep my change if that is ok.

afilina: Yeah, that's fine then.

afilina (approved):

cmb69:
Thank you!

PR #324, opened 2021-01-11
mysqli_query can return a mysqli_result for other types

afilina:
From the mailing list:

> Under 'Return Values,' it says:
> 
> >Returns false on failure. For successful SELECT, SHOW, DESCRIBE or EXPLAIN queries mysqli_query() will return a mysqli_result object. For other successful queries mysqli_query() will return true
> 
> This is not correct. I can provide one counterexample, which is when you run an optimize query:
> 
> OPTIMIZE TABLE my_table;
> 
> This query will return a result set with two records whether the table exists or not.

I have confirmed this behavior using this code:

```php
$mysqli = mysqli_connect("localhost", "my_user", "my_password", "world");
print_r($mysqli->query('optimize table City'));
print_r($mysqli->query('optimize table Invalid'));
```

Both returned a `mysqli_result` object showing two rows. I therefore updated the doc as per the recommendation.

### DIFF
--- a/reference/mysqli/mysqli_stmt/get-result.xml
+++ b/reference/mysqli/mysqli_stmt/get-result.xml
@@ -40,7 +40,8 @@
    Returns &false; on failure. For successful queries which produce a result set, such as <literal>SELECT, SHOW, DESCRIBE</literal> or
    <literal>EXPLAIN</literal>, <function>mysqli_stmt_get_result</function> will return
    a <classname>mysqli_result</classname> object. For other successful queries, <function>mysqli_stmt_get_result</function> will
-   return &true;.
+   return &false;. The <function>mysqli_errno</function> function can be
+   used to distinguish between the two reasons for &false;.
   </para>
  </refsect1>
 
--- a/reference/mysqli/mysqli_stmt/get-result.xml
+++ b/reference/mysqli/mysqli_stmt/get-result.xml
@@ -37,9 +37,10 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns a resultset for successful SELECT queries, or &false; for other DML
-   queries or on failure. The <function>mysqli_errno</function> function can be
-   used to distinguish between the two types of failure.
+   Returns &false; on failure. For successful queries which produce a result set, such as <literal>SELECT, SHOW, DESCRIBE</literal> or
+   <literal>EXPLAIN</literal>, <function>mysqli_stmt_get_result</function> will return
+   a <classname>mysqli_result</classname> object. For other successful queries, <function>mysqli_stmt_get_result</function> will
+   return &true;.
   </para>
  </refsect1>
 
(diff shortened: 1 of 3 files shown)


kamil-tekiela:
I'm probably asking too much, but would it possible for you to align https://www.php.net/manual/en/mysqli.reap-async-query.php and https://www.php.net/manual/en/mysqli-stmt.get-result.php too?

afilina:
> I'm probably asking too much, but would it possible for you to align https://www.php.net/manual/en/mysqli.reap-async-query.php and https://www.php.net/manual/en/mysqli-stmt.get-result.php too?

Aligned.

reference/mysqli/mysqli_stmt/get-result.xml
```diff
@@ -37,9 +37,10 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns a resultset for successful SELECT queries, or &false; for other DML
-   queries or on failure. The <function>mysqli_errno</function> function can be
-   used to distinguish between the two types of failure.
+   Returns &false; on failure. For successful queries which produce a result set, such as <literal>SELECT, SHOW, DESCRIBE</literal> or
+   <literal>EXPLAIN</literal>, <function>mysqli_stmt_get_result</function> will return
+   a <classname>mysqli_result</classname> object. For other successful queries, <function>mysqli_stmt_get_result</function> will
+   return &true;.
```
kamil-tekiela, line 43: I don't think `mysqli_stmt::get_result()` ever returns true. The only possible return values for this method are `mysqli_result` or `false`. It is different than `mysqli_query()` in that matter. 

Side note, the information about `mysqli_errno` was quite relevant, however strange it might sound. Remember that mysqli is extremely messed up and things are not as straight forwards as they seem. Usually, you would use `mysqli_stmt_errno` to check the error message, but in the case of `mysqli_stmt::get_result()` there was a bug causing the exception to never be raised and the statement error fields never populated. The only way to distinguish between failure or success is to check `mysqli_errno`. Yeah, I know... but fortunately, it has been fixed in PHP 8/ 7.4.13.

afilina: You are correct. Not sure how I missed that in my local sandbox. I wrote several additional cases to verify each assertion made by this paragraph. Thanks.

I clarified the return types and added back the `mysqli_errno` mention, but I refer to "two reasons for false" instead of failures, because an INSERT query is not a failure. Does that solve it?

kamil-tekiela: I'm not sure actually what to do about `mysqli_errno`. That bug has been fixed and to check for failure now you should use `mysqli_stmt_errno`. Should we mention both with the PHP version to which it applies to?

cmb69: Usually, we do not document bug fixes, but in this case it might be worthwhile.

cmb69:
Thank you. I'm going to add the info regarding the bug fix in PHP 7.4.13.

PR #325, opened 2021-01-11
Rework setAccessible description and add example

afilina:
Attempt to fix https://bugs.php.net/bug.php?id=55026

- Provide a description that makes it clearer how the property becomes accessible.
- Add example to demonstrate a Fatal Error in case your try to access members _not_ via reflection.


cmb69:
Thank you!

PR #327, opened 2021-01-12
Some cleanup in FAQ section

Girgias:
I wonder if we should remove most of the FAQ entries, as they seem to be grossly outdated (the build FAQ mention mostly issues about building PHP 4)


Girgias:
Maybe went a bit overboard by throwing in the removal of register_globals in this.

This needs: https://github.com/php/doc-base/pull/15 to build.

afilina (approved):
LGTM.
It's refreshing to see how much simpler the docs become when we don't have to mention all those legacy directives. I also agree with the removal of FAQs about outdated OSes and web servers.

cmb69:
Ah, autoexec.bat…

Thanks for the PR!

PR #329, opened 2021-01-13
Refactor OO section part 2

Girgias:
Some misses from the first time and doing some drive by fixes.

### DIFF
--- a/language/oop5/autoload.xml
+++ b/language/oop5/autoload.xml
@@ -14,25 +14,15 @@
    if they are currently not defined. By registering autoloaders, PHP is given
    a last chance to load the class or interface before it fails with an error.
   </para>
-  <tip>
+  <caution>
    <para>
-    Although the <function>__autoload</function> function can also be used for
-    autoloading classes and interfaces, it's preferred to use the
-    <function>spl_autoload_register</function> function. This is because it is
-    a more flexible alternative (enabling for any number of autoloaders to be
-    specified in the application, such as in third party libraries). For this
-    reason, using <function>__autoload</function> is discouraged and deprecated
-    as of PHP 7.2.0.
+    Prior to PHP 8.0.0, it was possible to use <function>__autoload</function>
+    to autoload classes and interfaces. However, it is a less flexible
+    alternative to <function>spl_autoload_register</function> and
+    <function>__autoload</function> is deprecated as of PHP 7.2.0, and removed
+    as of PHP 8.0.0.
    </para>
-  </tip>
-  <note>
-   <para>
-    If the class name is used e.g. in <function>call_user_func</function> then
-    it can contain some dangerous characters such as <literal>../</literal>.
-    It is recommended to not use the user-input in such functions or at least
-    verify the input in <function>__autoload</function>.
-   </para>
-  </note>
+  </caution>
   <para>
    <example>
     <title>Autoload example</title>
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -3,7 +3,19 @@
  <sect1 xml:id="language.oop5.magic" xmlns="http://docbook.org/ns/docbook">
   <title>Magic Methods</title>
   <para>
-   The function names
+   Magic methods are special methods which override PHP's default's action
+   when certain actions are performed on an object.
+  </para>
+  <caution>
+   <simpara>
+    All methods names starting with <literal>__</literal> are reserved by PHP.
+    Therefore, it is not recommended to use such method names unless overriding
+    PHP's behavior.
+   </simpara>
+  </caution>
+  <para>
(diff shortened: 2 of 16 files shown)


language/oop5/autoload.xml
```diff
@@ -14,31 +14,27 @@
...
     Autoloading is not available if using PHP in CLI
     <link linkend="features.commandline">interactive mode</link>.
    </para>
   </note>
-  <note>
+  <warning>
+   <!-- Is this still true? -->
    <para>
     If the class name is used e.g. in <function>call_user_func</function> then
     it can contain some dangerous characters such as <literal>../</literal>.
```
afilina, line 35: I don't see how one can supply `../` as the class name here. I don't see a test for it and I can't reproduce the behavior. Perhaps this refers to some very old bug. Without a code snipper demonstrating it, that warning doesn't really make any sense to me.

language/oop5/magic.xml
```diff
@@ -27,11 +27,21 @@
    with them.
   </para>
 
-  <note>
+  <warning>
    <simpara>
-    All magic methods <emphasis>MUST</emphasis> be declared as <literal>public</literal>
+    All magic methods <emphasis>must</emphasis> be declared as
+    <literal>public</literal>, otherwise an <constant>E_WARNING</constant>
+    is emitted. Prior to PHP 8.0.0, no diagnostic was emitted.
```
afilina, line 34: This doesn't seem to be the case even in 8.0.0: https://3v4l.org/hW7eN

afilina (changes requested):
Partial review. Will check the rest later today.

language/oop5/magic.xml
Girgias: No visibility === public visibility, however it seems a diagnostic depends on the magic methods, amazing... (see https://3v4l.org/PS0lf and https://3v4l.org/q5N43)

language/oop5/static.xml
```diff
@@ -18,9 +18,8 @@
 
   <para>
    Declaring class properties or methods as static makes them accessible
-   without needing an instantiation of the class. A property declared as
-   static cannot be accessed with an instantiated class object (though
-   a static method can).
+   without needing an instantiation of the class.
+   These can also be accessed within an instantiated class object.
```
afilina, line 22:
```suggestion
   These can also be accessed statically within an instantiated class object.
```

Not sure whether the distinction is needed here.

afilina (changes requested):

language/oop5/autoload.xml
salathe: The note was added in 2007 in response to https://bugs.php.net/42744.  PHP added validation to the class name in 2014 (php/php-src@2d31eadbbf147a157cb4a0c89adaf30fee7371f0, PHP 5.4.24 and PHP 5.5.8), which rendered the note obsolete for folks using those or newer versions of PHP.

afilina: Thanks for finding this out, @salathe. I was looking for the error in phpt files starting with `call_user_func*`, heh.

language/oop5/magic.xml
```diff
@@ -27,11 +27,21 @@
...
    </simpara>
-  </note>
+  </warning>
+  <warning>
+   <para>
+    If type declarations are used in the definition of a magic method, they
+    must be identical to the signature described in this document.
+    Otherwise a fatal error is emitted.
```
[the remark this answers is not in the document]
Girgias: There isn't a rule, and I probably just copy pasted it from code output. But yeah not great to be inconsistent.

language/oop5/static.xml
```diff
@@ -29,16 +28,16 @@
...
 
    <warning>
     <para>
      Calling non-static methods statically throws an <classname>Error</classname>.
     </para>
     <para>
-     Prior to PHP 8.0.0, calling non-static methods statically is deprecated, and will
-     generate an <constant>E_DEPRECATED</constant> warning.
+     Prior to PHP 8.0.0, calling non-static methods statically is deprecated, and
+     generated an <constant>E_DEPRECATED</constant> warning.
```
Crell, line 40: The verb tenses here are hurting my brain.  It really feels like "prior to 8.. were deprecated..." is more correct.  I can see the reason for using "is", but I really think "were" is more correct.

```suggestion
     Prior to PHP 8.0.0, calling non-static methods statically were deprecated, and
     generated an <constant>E_DEPRECATED</constant> warning.
```

Girgias: Fixed and I think it makes more sense indeed.

cmb69: Not a native speaker, but the plural looks wrong to me (the past tense looks right, though present perfect may be even more appropriate).

PR #331, opened 2021-01-13
Fixed param name LDAP_OPT_X_TLS_KEYILE

p4tat0r:
The param name is not ok in https://www.php.net/manual/en/function.ldap-get-option.php


cmb69:
Thank you!

PR #332, opened 2021-01-14
Remove PHP4 constructors from ext/soap documentation

MCMic:
Contructors were still documented as aliases to classname::classname
 with the real documentation in classname::classname.


cmb69:
We cannot remove the old style constructors from the manual, because they are still supported in PHP 7, and __construct() is only the alias in those versions. I think it makes sense, though, to swap the old and new style constructors in the docs (i.e. make __construct() the default, and claim that the old style constructors are aliases which are removed as of PHP 8.0.0).

MCMic:
> We cannot remove the old style constructors from the manual, because they are still supported in PHP 7, and __construct() is only the alias in those versions. I think it makes sense, though, to swap the old and new style constructors in the docs (i.e. make __construct() the default, and claim that the old style constructors are aliases which are removed as of PHP 8.0.0).

I do not understand.

This is the only classes I’ve seen documenting PHP4 constructors.

For instance: https://www.php.net/manual/fr/class.datetime.php does not list DateTime::DateTime.
The stub file for ext/soap does not list those functions either.

cmb69:
````
php -a
> echo PHP_VERSION;
7.2.35-dev
> SoapVar::__construct();
Warning: Uncaught Error: Non-static method SoapVar::SoapVar() cannot be called statically in php shell code:1
> DateTime::DateTime();
Warning: Uncaught Error: Call to undefined method DateTime::DateTime() in php shell code:1
````

Soap is special in this regard.

**Comment by MCMic** [comment]
Hmm.

I still feel like this is a corner case which should not appear in the documentation.
What difference does it make in actual code?

The only one I see is that you can call `$object->SoapVar(…` and get the same result as `$object->__contruct(…` but there is no real reason to do this.

**Comment by cmb69** [comment]
If these methods are removed from the manual, users might think they are no longer supported. Or even worse, they might think that they are even supported with PHP 8. We should keep the old style constructors, and explicitly mark them as being removed as of PHP 8.0.0 (basically like https://www.php.net/imap_header).

**Comment by nikic** [comment]
I agree with @MCMic here, in that it shouldn't matter to the end user whether the class uses an old-style or a new-style constructor. That's essentially an implementation detail.

File: reference/soap/functions/is-soap-fault.xml [hunk 1]
```diff
@@ -93,8 +93,8 @@ try {
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member><methodname>SoapClient::SoapClient</methodname></member>
-    <member><methodname>SoapFault::SoapFault</methodname></member>
+    <member><methodname>SoapClient::__contruct</methodname></member>
```
nikic, line 96: __construct :)

cmb69:
Okay, there is a supermajority for merging this. :)

PR #333, opened 2021-01-15
Added PHP 8 into revision version set for var functions.

Tpojka:
`debug_zval_dump` has inconsistencies and docs for it need to be checked again.


cmb69:
Thank you for the PR!

> debug_zval_dump has inconsistencies […]

That page is not even up to date for PHP 7; see https://externals.io/message/87264#87285.

PR #334, opened 2021-01-16
Added PHP 8 into versions.xml for strings functions

mumumu:
I noticed by #333, that `PHP 8` version is not added in most of versions.xml.

Old versions.xml leads to some confusion for PHP newbie, because It seems most of these functions cannot be used in PHP 8.

For fixing this issue, I created PR for ext/standard/strings functions for starting point.

### DIFF
--- a/reference/strings/versions.xml
+++ b/reference/strings/versions.xml
@@ -4,107 +4,107 @@
   Do NOT translate this file
 -->
 <versions>
- <function name="addcslashes" from="PHP 4, PHP 5, PHP 7"/>
- <function name="addslashes" from="PHP 4, PHP 5, PHP 7"/>
- <function name="bin2hex" from="PHP 4, PHP 5, PHP 7"/>
- <function name="chop" from="PHP 4, PHP 5, PHP 7"/>
- <function name="chr" from="PHP 4, PHP 5, PHP 7"/>
- <function name="chunk_split" from="PHP 4, PHP 5, PHP 7"/>
+ <function name="addcslashes" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
+ <function name="addslashes" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
+ <function name="bin2hex" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
+ <function name="chop" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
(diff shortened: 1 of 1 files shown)


mumumu:
We can find deleted functions in PHP 8 by the folloing codes.

```php
<?php

$xml = simplexml_load_file("/path/to/reference/strings/versions.xml");
foreach ($xml as $data) {
    $funcname = $data["name"];
    $evalstr = $funcname . "('param');";
    try {
        eval($evalstr);
    } catch (Error $err) {
        $msg = $err->getMessage();
        if (str_contains($msg, "Call to undefined function")) {
            echo "\nundefined!! " . $funcname . "\n";
        }   
    } catch (Exception) {}
}
```

cmb69:
Thank you for the PR!

Regarding the detection of available functions: I suggest to not check that by availability (neither by calling the functions nor via reflection), because some extension may not be loaded (and may not be loadable at all, e.g. ext/com_dotnet on Linux or ext/sysvsem on Windows), and some functions are defined conditionally, so may not be available on the used build.

Instead it might more reliable to check the respective stub files; in this case https://github.com/php/php-src/blob/PHP-8.0/ext/standard/basic_functions.stub.php.

cmb69:
Applied as 0aeef38a9cb30ab6b7a5fa787e11713db9481829.

mumumu:
> Instead it might more reliable to check the respective stub files; in this case https://github.com/php/php-src/blob/PHP-8.0/ext/standard/basic_functions.stub.php.

Thank you for your suggestion!
I'll generate versions.xml file based on stub, like methodsynopses.

```php
<?php

if (PHP_MAJOR_VERSION !== 8) {
    die("Error: this script must run with PHP 8\n");
}

$versions_file = "/path/to/doc-en/reference/sockets/versions.xml";
$stub_file = "/path/to/php-src/ext/sockets/sockets.stub.php";

$xml = simplexml_load_file($versions_file);
$stub = file_get_contents($stub_file);

foreach ($xml as $data) {
    $funcname = $data["name"];
    $evalstr = $funcname . "('param');";
    $from = $data["from"];
    $already_in_php8 = str_contains($from, "PHP 8");
    $already_in_stub = str_contains($stub, $funcname);
    $can_add = (!$already_in_php8 && $already_in_stub);

    if ($can_add) $data["from"] .= ", PHP 8";
}

file_put_contents($versions_file, $xml->asXML());
```

PR #335, opened 2021-01-16
Added PHP 8 into versions.xml for json based on stubs, php-src.

mumumu:
- function, interfaces
  * https://github.com/php/php-src/blob/PHP-8.0/ext/json/json.stub.php
- JsonException
  * https://github.com/php/php-src/blob/PHP-8.0/ext/json/json.c#L52


cmb69:
Thank you!

PR #336, opened 2021-01-16
Added PHP 8 into versions.xml for sockets based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/sockets/sockets.stub.php


cmb69:
Thank you!

PR #337, opened 2021-01-16
Added version 'PHP 8' to array functions' pages

Tpojka:
Also fixed and improved some of examples on pages:

https://www.php.net/manual/en/function.array-walk.php
  - removed <br> tag from print function

https://www.php.net/manual/en/function.key.php
  - removed '<br>' tag and add "\n" in echo

https://www.php.net/manual/en/function.range.php
  - improved examples' readability

What is left there that I can see:

https://www.php.net/manual/en/function.array-fill.php
  - different behavior in PHP 8

https://www.php.net/manual/en/function.array-udiff-assoc.php
  - different behavior for PHP 5, PHP 7, PHP 8

https://www.php.net/manual/en/function.array-udiff-uassoc.php
  - different behavior for PHP 7, PHP 8

https://www.php.net/manual/en/function.count.php
  - different behavior for PHP 7, PHP 8

Changelog section would be appropriate similarly to 
https://www.php.net/manual/en/function.array-unique.php

### DIFF
--- a/reference/array/functions/range.xml
+++ b/reference/array/functions/range.xml
@@ -72,17 +72,26 @@ foreach (range(0, 12) as $number) {
     echo $number;
 }
 
+echo "\n";
+
 // The step parameter
 // array(0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100)
-foreach (range(0, 100, 10) as $number) {
+foreach (range(0, 100, 10) as $k => $number) {
     echo $number;
+    if ($k === count(range(0, 100, 10)) - 1) {
+        echo "\n";
+    }
 }
 
 // Usage of character sequences
 // array('a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i');
-foreach (range('a', 'i') as $letter) {
+foreach ($arr = range('a', 'i') as $k => $letter) {
     echo $letter;
+    if ($k === array_key_last($arr)) {
+        echo "\n";
+    }
 }
+
 // array('c', 'b', 'a');
 foreach (range('c', 'a') as $letter) {
     echo $letter;
--- a/reference/array/functions/range.xml
+++ b/reference/array/functions/range.xml
@@ -76,21 +76,19 @@ echo "\n";
 
 // The step parameter
 // array(0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100)
-foreach (range(0, 100, 10) as $k => $number) {
+foreach (range(0, 100, 10) as $number) {
     echo $number;
-    if ($k === count(range(0, 100, 10)) - 1) {
-        echo "\n";
-    }
 }
+
+echo "\n";
 
 // Usage of character sequences
 // array('a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i');
-foreach ($arr = range('a', 'i') as $k => $letter) {
+foreach (range('a', 'i') as $letter) {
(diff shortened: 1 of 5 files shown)


reference/array/versions.xml
```diff
@@ -4,87 +4,87 @@
...
+ <function name='array_unshift' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='array_values' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='array_walk' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='array_walk_recursive' from='PHP 5, PHP 7, PHP 8'/>
+ <function name='arsort' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='asort' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='compact' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='count' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
+ <function name='current' from='PHP 4, PHP 5, PHP 7, PHP 8'/>
  <function name='each' from='PHP 4, PHP 5, PHP 7' deprecated='PHP 7.2.0'/>
```
Girgias, line 65: There should be a removed tag for PHP 8.0.0 for each()

reference/array/functions/range.xml
```diff
@@ -72,17 +72,26 @@ foreach (range(0, 12) as $number) {
...
+echo "\n";
+
 // The step parameter
 // array(0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100)
-foreach (range(0, 100, 10) as $number) {
+foreach (range(0, 100, 10) as $k => $number) {
     echo $number;
+    if ($k === count(range(0, 100, 10)) - 1) {
+        echo "\n";
+    }
```
Girgias, line 83: I don't understand the point of these changes, they just clutter the function.
What you want to do seems to be just solvable by doing ``echo $number, "\n"``

Girgias (changes requested):
The foreach changes are completely unnecessary and just clutter the example

reference/array/functions/range.xml
Tpojka: Result is fairly unreadable if there is no new line. 
How it is for now:
https://3v4l.org/B78cK

reference/array/versions.xml
Tpojka: Didn't know such an exists. 
Like

    <function name='each' from='PHP 4, PHP 5, PHP 7' deprecated='PHP 7.2.0' removed='PHP 8'/>

?

cmb69: I don't think that we use removed attributes; instead we don't update the from attribute, and explicitly add a warning to the page that the function has been removed.

reference/array/functions/range.xml
cmb69: Maybe just output a space unconditionally, and a new line after each loop?

reference/array/versions.xml
Tpojka: Added with `fff33ce` following similar code in docs.

reference/array/functions/range.xml
Tpojka: Changed in `0353cf2`.

cmb69:
Thank you!

PR #338, opened 2021-01-16
Updating mysqli: prepared statements

kamil-tekiela:
Another PR aiming to improve mysqli docs by removing lies and deception. In this episode:
- `mysqli_stmt_bind_result()` cannot be used to bind variables to placeholders
- There is no such thing as *"null-terminated string query"*.
- A prepared statement doesn't have to have any parameter markers. The parameter binding can be skipped. This means that the SQL can contain zero or more markers.
- The trailing semicolon is ignored. The Go command (i.e. `\g`) is not allowed in any application layer client API. No idea why that note was there, but it has been throwing a lot of people off.
- The note which was copied from MySQL docs has been altered. Not only the English language was broken, but incorrect information was added. I am not aware of any limitations in regards to `? IS NULL`. The two pages in the manual had a slightly different version of the note. If it was copied from MySQL then let's bring it up to date. 
- `mysqli_stmt::prepare` can return only a boolean as far as I know
- `mysqli_stmt::execute` contained some strange notes, which I removed. Parameter markers are never replaced into the SQL by neither client nor server. They are sent separately and are used in the execution. `mysqli_stmt_fetch` doesn't have to be used. It might be used in situations where a result set is bound to variables and needs to be fetched, but it is in no way a requirement. That sentence made no sense, but I assume it was meant to inform users that `execute` doesn't return the result set. If that was the intention then there are better ways to communicate it. 
- A copy-paste of a snippet made for some strange information in `mysqli_stmt::__construct`. The first parameter is mandatory and there is no procedural version. The object needs to be a valid mysqli connection to MySQL. 
- Improved code examples and some wording.

### DIFF
--- a/reference/mysqli/mysqli/prepare.xml
+++ b/reference/mysqli/mysqli/prepare.xml
@@ -47,22 +47,20 @@
        question mark (<literal>?</literal>) characters at the appropriate positions.
       </para>
       <note>
+<!-- Copied from https://dev.mysql.com/doc/c-api/8.0/en/mysql-stmt-prepare.html -->
        <para>
         The markers are legal only in certain places in SQL statements.
-        For example, they are allowed in the <literal>VALUES()</literal>
+        For example, they are permitted in the <literal>VALUES()</literal>
         list of an <literal>INSERT</literal> statement (to specify column
         values for a row), or in a comparison with a column in a
         <literal>WHERE</literal> clause to specify a comparison value.
        </para>
        <para>
-        However, they are not allowed for identifiers (such as table or
-        column names), in the select list that names the columns to be
-        returned by a <literal>SELECT</literal> statement, or to specify both
+        However, they are not permitted for identifiers (such as table or
+        column names), or to specify both
         operands of a binary operator such as the <literal>=</literal> equal
         sign. The latter restriction is necessary because it would be
-        impossible to determine the parameter type. It's not allowed to
-        compare marker with <literal>NULL</literal> by 
-        <literal>? IS NULL</literal> too. In general, parameters are legal
+        impossible to determine the parameter type. In general, parameters are legal
         only in Data Manipulation Language (DML) statements, and not in Data
         Definition Language (DDL) statements.
        </para>
--- a/reference/mysqli/mysqli/prepare.xml
+++ b/reference/mysqli/mysqli/prepare.xml
@@ -25,7 +25,7 @@
    operations on the statement. The query must consist of a single SQL statement.
   </para>
   <para>
-   The statement template can contain zero or more question mark (?) parameter markers⁠—also called 
+   The statement template can contain zero or more question mark (<literal>?</literal>) parameter markers⁠—also called 
    placeholders. The parameter markers must be bound to application variables using
    <function>mysqli_stmt_bind_param</function> before executing the statement.
   </para>
--- a/reference/mysqli/mysqli/prepare.xml
+++ b/reference/mysqli/mysqli/prepare.xml
@@ -25,8 +25,9 @@
    operations on the statement. The query must consist of a single SQL statement.
   </para>
   <para>
-   The statement template can contain zero or more question mark (<literal>?</literal>) parameter markers⁠—also called 
-   placeholders. The parameter markers must be bound to application variables using
+   The statement template can contain zero or more question mark
+   (<literal>?</literal>) parameter markers⁠—also called placeholders.
+   The parameter markers must be bound to application variables using
    <function>mysqli_stmt_bind_param</function> before executing the statement.
   </para>
  </refsect1>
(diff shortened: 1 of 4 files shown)


reference/mysqli/mysqli/prepare.xml
```diff
@@ -25,10 +25,9 @@
    operations on the statement. The query must consist of a single SQL statement.
   </para>
   <para>
-   The parameter markers must be bound to application variables using
-   <function>mysqli_stmt_bind_param</function> and/or 
-   <function>mysqli_stmt_bind_result</function> before executing the
-   statement or fetching rows.
+   The statement template can contain zero or more question mark (?) parameter markers⁠—also called
```
Girgias, line 28:
```suggestion
   The statement template can contain zero or more question mark (<literal>?</literal>) parameter markers⁠—also called 
```

reference/mysqli/mysqli_stmt/construct.xml
```diff
@@ -23,7 +23,14 @@
   &reftitle.parameters;
   <para>
    <variablelist>
-    &mysqli.link.description;
```
Girgias, line 26: Is this just a misuse of the entity or should the entity be adjusted?

reference/mysqli/mysqli_stmt/execute.xml
```diff
@@ -20,10 +20,9 @@
    <methodparam><type>mysqli_stmt</type><parameter>stmt</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Executes a query that has been previously prepared using the
-   <function>mysqli_prepare</function> function. When executed any
-   parameter markers which exist will automatically be replaced with the
-   appropriate data.
+   Executes previously prepared statement. The statement must be successfully prepared prior to execution,
+   using either the <function>mysqli_prepare</function> or <function>mysqli_stmt_prepare</function>
+   function, or by passing the second argument to <methodname>mysqli_stmt::__construct</methodname>.
```
Girgias, line 25: Could you split this into more lines to make them less than 80 characters/line (helps with reading diffs in a CLI)

reference/mysqli/mysqli_stmt/execute.xml
```diff
@@ -94,47 +83,31 @@ $val3 = 'Aquitaine';
...
-
-/* close connection */
-$mysqli->close();
-?>
 ]]>
    </programlisting>
-  </example>
-  <example>
-   <title>&style.procedural;</title>
+   <para>&style.procedural;</para>
```
Girgias, line 97: Shouldn't this mimic the change made above?
```suggestion
   <title><methodname>mysqli_stmt::execute</methodname> example</title>
   <para>&style.procedural;</para>
```

reference/mysqli/mysqli_stmt/prepare.xml
```diff
@@ -21,13 +21,12 @@
...
   <para>
-   Prepares the SQL query pointed to by the null-terminated string query. 
+   Prepares a statement for execution. The query must consist of a single SQL statement.
   </para>
   <para>
-   The parameter markers must be bound to application variables using
-   <function>mysqli_stmt_bind_param</function> and/or 
-   <function>mysqli_stmt_bind_result</function> before executing the
-   statement or fetching rows.
+   The statement template can contain zero or more question mark (?) parameter markers⁠—also called
```
Girgias, line 27:
```suggestion
   The statement template can contain zero or more question mark (<literal>?</literal>) parameter markers⁠—also called 
```
Also potentially split into more lines

reference/mysqli/mysqli_stmt/prepare.xml
```diff
@@ -118,88 +112,63 @@
...
-/* close connection */
-$mysqli->close();
-?>
+printf("%s is in district %s\n", $city, $district);
 ]]>
    </programlisting>
-  </example>
-  <example>
-   <title>&style.procedural;</title>
+   <para>&style.procedural;</para>
```
Girgias, line 145:
```suggestion
   <title><methodname>mysqli_stmt::prepare</methodname> example</title>
   <para>&style.procedural;</para>
```
As above

Girgias (changes requested):
Minor nits, as I can't comment on the content of the updates as I don't know the mysqli API

reference/mysqli/mysqli_stmt/construct.xml
kamil-tekiela: The entity is ok, but it should not be used on this page. It just doesn't apply here.

reference/mysqli/mysqli_stmt/execute.xml
kamil-tekiela: No. The text should only be `&style.procedural;` and it should not be emphasised like the example title. I can't accept that suggestion as it was done this way on purpose. Sorry. If I were to use a title here then the layout would look worse.

afilina:
Looks like all the comments were addressed here. @Girgias?

cmb69:
Thank you!

Generally, please consider to provide more fine grained PRs. The review process for coarse grained PRs doesn't really work well (well, maybe it doesn't work well for fine grained PRs either :)).

PR #340, opened 2021-01-16
Generate intl methodsynopses - IntlDateFormatter

kocsismate:
As far as I saw, no notable change happened with this class

### DIFF
--- a/reference/intl/dateformatter/create.xml
+++ b/reference/intl/dateformatter/create.xml
@@ -12,40 +12,37 @@
   <para>
    &style.oop;
   </para>
-  <methodsynopsis>
-   <modifier>public</modifier>
-   <modifier>static</modifier>
-   <type>IntlDateFormatter</type><methodname>IntlDateFormatter::create</methodname>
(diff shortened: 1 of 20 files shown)


cmb69:
There have been a few commits regarding the stubs recently; could you please check whether that would change anything?

kocsismate:
> There have been a few commits regarding the stubs recently; could you please check whether that would change anything?

I've tried to search for some updated signatured - possibly I missed quite a few. Unfortunately, there's still more 1000 changed files if I generate the methodsynopses based on the whole `ext/` directory :(

kocsismate:
Oh, I've just realized that you probably meant updates in connection with this PR. Just pushed one change!

cmb69:
Yes, I was referring to this PR in particular. :)

cmb69:
Thank you!

My favorite amendment of this PR (for `IntlDateFormatter::setTimezone()`):
````
Returns &null; on success and &false; on failure.
````
heh

PR #345, opened 2021-01-17
Added PHP 8 into versions.xml for zip based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/zip/php_zip.stub.php
- the following functions were not found in stub file, but undocumented in appendices/migration80*
  * ZipArchive::delete
  * ZipArchive::stat

### DIFF
--- a/reference/zip/versions.xml
+++ b/reference/zip/versions.xml
@@ -5,68 +5,68 @@
 -->
 <versions>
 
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

`ZipArchive::delete()` and `ZipArchive::stat()` apparently were never been part of the bundled ZIP extension (and while the former was definitely part of early PECL alpha releases, I cannot find the latter). Anyway, I'm going to remove these entries altogether.

PR #350, opened 2021-01-17
Added PHP 8 into versions.xml for phar based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/phar/phar_object.stub.php
- the following functions were not found in stub file, but undocumented in appendices/migration80*
  * Phar::convertToTar
  * Phar::convertToZip
  * Phar::convertToPhar
  * Phar::isPhar
  * Phar::isTar
  * Phar::isZip
  * pharfileinfo_decompress
  * pharfileinfo_compress

### DIFF
--- a/reference/phar/versions.xml
+++ b/reference/phar/versions.xml
@@ -5,110 +5,110 @@
 -->
 <versions>
 
- <function name='phar' from='PHP 5 &gt;= 5.3.0, PHP 7, PECL phar &gt;= 1.0.0'/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you for the PR!

The missing functions *might* have been removed before Phar has been bundled.

I wonder if we should remove the PECL versions, since the last PECL version is from 2009, so may not be compatible with PHP 7 (and nobody is supposed to use that anymore).

PR #353, opened 2021-01-18
Added PHP 8 into versions.xml for stream based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/standard/user_filters.stub.php
- Notable changes
  * `stream_notification_callback` function seems to be deleted, but undocumented in appendices/migration80*
  * `streamWrapper` class is not a real class, but exists in manual, so I added "PHP 8" manually.
    - https://www.php.net/manual/en/class.streamwrapper.php

### DIFF
--- a/reference/stream/versions.xml
+++ b/reference/stream/versions.xml
@@ -4,84 +4,84 @@
   Do NOT translate this file
 -->
 <versions>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! `stream_notification_callback` is also not a real function, but documented in the manual and still supported, so I'm going to update the version right away.

PR #354, opened 2021-01-18
Added PHP 8 into versions.xml for info based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/standard/dl.stub.php
- https://github.com/php/php-src/blob/PHP-8.0/sapi/cli/php_cli_process_title.stub.php
- Notable changes
  * deleted functions
    - get_magic_quotes_gpc
    - get_magic_quotes_runtime
    - restore_include_path

### DIFF
--- a/reference/info/versions.xml
+++ b/reference/info/versions.xml
@@ -4,61 +4,61 @@
   Do NOT translate this file
 -->
 <versions>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! I'll document the removal of these 3 functions right away.

PR #355, opened 2021-01-18
Added PHP 8 into versions.xml for session based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/session/session.stub.php
- Notable changes
  * already deleted in ancient times.
    - session_is_registered
    - session_register
    - session_unregister
  * Maybe, the following function is mistake, because these definitions does not exist in manual nor stub file.
    - sessionhandler::updatetimestamp
    - sessionhandler::validateid

### DIFF
--- a/reference/session/versions.xml
+++ b/reference/session/versions.xml
@@ -4,56 +4,56 @@
   Do NOT translate this file
 -->
 <versions>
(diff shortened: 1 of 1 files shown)


cmb69:
I removed the mentioned entries with https://github.com/php/doc-en/commit/b41704883cac98283b6cdae5a0b3773bbbfcbf37.

PR #361, opened 2021-01-18
Use 'ENT_QUOTES|ENT_SUBSTITUTE' for HTML encoding and decoding functions

craigfrancis:
I'm not familiar with the PHP development process, but it looks like [PR 6583](https://github.com/php/php-src/pull/6583) has been accepted in [50eca61](https://github.com/php/php-src/commit/50eca61f68815005f3b0f808578cc1ce3b4297f0), and will be in PHP 8.1.

These should update the documentation to reflect this change.

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -660,7 +660,7 @@ array_fill(value: 50, num: 100, start_index: 0);
 <?php
 htmlspecialchars($string, double_encode: false);
 // Same as
-htmlspecialchars($string, ENT_COMPAT | ENT_HTML401, 'UTF-8', false);
+htmlspecialchars($string, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401, 'UTF-8', false);
 ?>
 ]]>
      </programlisting>
--- a/reference/strings/functions/get-html-translation-table.xml
+++ b/reference/strings/functions/get-html-translation-table.xml
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>array</type><methodname>get_html_translation_table</methodname>
    <methodparam choice="opt"><type>int</type><parameter>table</parameter><initializer><constant>HTML_SPECIALCHARS</constant></initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer><constant>ENT_COMPAT</constant></initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer><constant>ENT_QUOTES</constant> | <constant>ENT_SUBSTITUTE</constant></initializer></methodparam>
    <methodparam choice="opt"><type>string</type><parameter>encoding</parameter><initializer>"UTF-8"</initializer></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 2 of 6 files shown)


Girgias:
Thanks for the PR, this will not be merged until we're closer to the PHP 8.1 release as per the current doc process.

Might need to be rebased in 6-8 months time.

craigfrancis:
Yep, np, and feel free to edit/delete if it's not right.

afilina:
@Girgias Is there a "pending milestone" tag or something along those lines to make it more apparent in the PR list? Sorry to be a bother, it's just that it makes it harder to sort all the open PRs when I go through the list periodically to try and help PRs along.

salathe:
Did these change deliberately exclude `ENT_HTML401` from the default values, or just an oversight (the files "before" also didn't reference it in the prototypes but did in the descriptions, for the most part)?

cmb69:
@salathe, nor do the stubs specify `ENT_HTML401`, likely because its value is `0` so hasn't any effect on the bitmask.

craigfrancis:
I'm fairly sure `ENT_HTML401` is the default ([value 0](https://github.com/php/php-src/blob/01b3fc03c30c6cb85038250bb5640be3a09c6a32/ext/standard/html.h#L39).

We did consider using `ENT_HTML5` due to feedback from [Claude Pache](https://externals.io/message/112617#112794), and I did try adding it to the patch, but removed it to keep the change small, and because I felt that should have had a bit more of a discussion (which I don't think happened)... my main focus was in getting `ENT_QUOTES` included, due to the security issues it can cause:

    $url = "/' onerror='alert(1)";
    echo "";

cmb69:
> I'm fairly sure ENT_HTML401 is the default (value 0).

Yes, and it might be good to add it (also to the stubs) for clarity. However, `ENT_QUOTES|ENT_HTML401 === ENT_QUOTES` (etc.)

salathe:
Thanks @craigfrancis 

Apologies for not looking at this sooner and raising the inconsistency while the PR was still open.

Going forward, I think the docs should have the default value be at least consistent within each page, even if `ENT_QUOTES | ENT_SUBSTITUTE` and `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401` are really one and the same set of bits.  My preference is to align on using `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`. 

Thoughts, violent disagreement, volunteers?

craigfrancis:
I have a slight preference to including `ENT_HTML401` to note that it's the default.

Some developers might see that and then decide to use `ENT_HTML5`, even though `&apos;` still upsets Microsoft Outlook (yep, it's been well over 13 years).

If everyone is happy with that, are you in a better position to make that change? If it's a pain, let me know, and I should be able to create a new pull-request on the weekend.

cmb69:
But please fix the stubs first (otherwise this is likely to be overwritten later): https://github.com/php/php-src/blob/c37b35fa41be7ec4278e51c6b6715a8805ebb725/ext/standard/basic_functions.stub.php#L686

craigfrancis:
I've created a pull-request to update the stubs: https://github.com/php/php-src/pull/7514

craigfrancis:
The stubs have been updated - Thanks to @nikic who finished it off, and merged it it via [2b25ac6](https://github.com/php/php-src/commit/2b25ac6f3ce5b298788fe55ae64bb8907f69f4bb).

PR #365, opened 2021-01-19
Added PHP 8 into versions.xml for ldap based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/ldap/ldap.stub.php
- Deleted functions from PHP8. These functions were already documented in migration guide.
  * ldap_sort
  * ldap_control_paged_result
  * ldap_control_paged_result_response


cmb69:
Thank you!

PR #368, opened 2021-01-20
Add floats without exponent to numeric string grammar

lucaswerkmeister:
The formal grammar for numeric strings included integers, and
floating-point numbers (decimals) with a mandatory exponent, but not
decimal numbers without exponent.


lucaswerkmeister:
`{DNUM} | {EXPONENT_DNUM}` is modeled after `zend_language_scanner.l`:
https://github.com/php/php-src/blob/a48d9ede08fc066ae50dcfbe1f378762ac516e4c/Zend/zend_language_scanner.l#L2134

Girgias (approved):
Thanks for catching this!

PR #369, opened 2021-01-20
Replace with include expression instead of statement

zonuexe:
[Expressions](https://github.com/php/php-langspec/blob/master/spec/10-expressions.md) and [Statements](https://github.com/php/php-langspec/blob/master/spec/11-statements.md) are strictly distinguished by the PHP language specification.

In all PHP versions, "require" and "include" are expressions that returns a value, not statements.

```php
<?php

if (false) {
   $v = include __FILE__;
   $v = require __FILE__;
   $v = include_once __FILE__;
   $v = require_once __FILE__;
}
```

https://3v4l.org/hHYEF

### DIFF
--- a/language/control-structures/include.xml
+++ b/language/control-structures/include.xml
@@ -5,7 +5,7 @@
  <title>include</title>
  <?phpdoc print-version-for="include"?>
  <simpara>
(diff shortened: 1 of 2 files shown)


cmb69 (approved):
Good catch! Thank you.

PR #370, opened 2021-01-22
Small grammar fix for the Attributes docs

grimskin:
- using `is` instead of `are` with a singular noun
- typo fix in word `separated`


cmb69:
Thank you!

PR #373, opened 2021-01-22
Fix documentation about deprecated and removed constants in VALIDATE_URL filter

bohwaz:


reference/filter/filters.xml
```diff
@@ -194,13 +192,25 @@
        </row>
       </thead>
       <tbody>
+       <row>
+        <entry>8.0.0</entry>
+        <entry>
+         <constant>FILTER_FLAG_SCHEME_REQUIRED</constant> and <constant>FILTER_FLAG_HOST_REQUIRED</constant> constants have been <link xlink:href="https://wiki.php.net/rfc/deprecations_php_7_3#filter_flag_scheme_required_and_filter_flag_host_required">removed</link>.
```
afilina, line 198: I'm not sure about the correct form that the manual proposes, but it seems to me that a better link would be to the migration docs instead of the RFC: https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.filter

reference/filter/filters.xml
```diff
@@ -194,13 +192,25 @@
...
         <entry>7.4.0</entry>
         <entry>
          Added <parameter>min_range</parameter> and <parameter>max_range</parameter>
          options for <constant>FILTER_VALIDATE_FLOAT</constant>.
         </entry>
        </row>
+       <row>
+        <entry>7.3.0</entry>
+        <entry>
+         <constant>FILTER_FLAG_SCHEME_REQUIRED</constant> and <constant>FILTER_FLAG_HOST_REQUIRED</constant> constants are now <link xlink:href="https://wiki.php.net/rfc/deprecations_php_7_3#filter_flag_scheme_required_and_filter_flag_host_required">deprecated</link>, they never had any effect.
```
afilina, line 211: Similar to the above comment: https://www.php.net/manual/en/migration73.deprecated.php#migration73.deprecated.filter

afilina (changes requested):

reference/filter/filters.xml
cmb69: The man pages are neither supposed to link to RFCs nor the migration guide. The migration guide is supposed to link to the man pages (as soon as these have been updated). RFCs should only be mentioned in XML comments, if at all.

afilina: Ah, thanks for the clarification. I learn every day :)

afilina, line 198:
```suggestion
         <constant>FILTER_FLAG_SCHEME_REQUIRED</constant> and <constant>FILTER_FLAG_HOST_REQUIRED</constant> constants have been removed.
```

reference/filter/filters.xml
afilina, line 211:
```suggestion
         <constant>FILTER_FLAG_SCHEME_REQUIRED</constant> and <constant>FILTER_FLAG_HOST_REQUIRED</constant> constants have been deprecated; they never had any effect.
```

afilina (changes requested):

reference/filter/filters.xml
```diff
@@ -159,8 +159,6 @@
          <parameter>default</parameter>
         </entry>
         <entry>
-         <constant>FILTER_FLAG_SCHEME_REQUIRED</constant>,
-         <constant>FILTER_FLAG_HOST_REQUIRED</constant>,
```
cmb69, line 163: The manual is currently supposed to document PHP 7.0.0 and up, so these lines should not be removed, but rather it should be clarified that these constants are only there "(prior to PHP 8.0.0)".

reference/filter/filters.xml
cmb69: Instead of "they never had any effect", maybe "they are implied anyway"?

reference/filter/filters.xml
bohwaz: I suggest removal as mentioning them suggests that this is implemented and that the behavior will change but it won't, so it would be confusing to leave them as is IMO.

reference/filter/filters.xml
afilina: I think that you're referring to my other suggestion below. I updated it.

afilina:
@cmb69 I believe that this PR is waiting on a review pass from the maintainers.

bohwaz:
Thank you @afilina I think I missed the notifications. Should be good now?

Girgias:
There are conflicts @bohwaz, could you pull the latest master and rebase the PR on top of it?

bohwaz:
Sorry I have exactly zero time.

lacatoire:
This can be closed. The file it changes, `reference/filter/filters.xml`, was deleted in #4158, and the same information now lives in `reference/filter/constants.xml`, where both constants are already marked deprecated in 7.3.0 and removed in 8.0.0:
- https://www.php.net/manual/en/filter.constants.php#constant.filter-flag-scheme-required
- https://www.php.net/manual/en/filter.constants.php#constant.filter-flag-host-required
There is nothing left to rebase.

PR #374, opened 2021-01-23
Added PHP 8 into versions.xml for oci8 based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/oci8/oci8.stub.php
- Note
  * OCICollection, OCILob related functions are migrated from procedural to oop.
  * non-existent [function|alias] on PHP 8
    - oci_internal_debug
    - ocicloselob
    - ocicollassign
    - ociwritetemporarylob

### DIFF
--- a/reference/oci8/versions.xml
+++ b/reference/oci8/versions.xml
@@ -4,145 +4,145 @@
   Do NOT translate this file
 -->
 <versions>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

PR #376, opened 2021-01-23
Added PHP 8 into versions.xml for simplexml based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/simplexml/simplexml.stub.php
- Note
  * `simplexmliterator::*` methods could not be generated from stub, because inherited from SimpleXMLElement
  *  but works on my PHP 8.0.1 environment.


cmb69:
Thank you!

PR #378, opened 2021-01-23
Added PHP 8 into versions.xml for date based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/date/php_date.stub.php
- Note
  * functions derived from `PECL date_time:0.1` were not found in php-src.
  * the following functions were not found on stub, but works on my environment.
    - gettimeofday
    - microtime
    - strptime

### DIFF
--- a/reference/datetime/versions.xml
+++ b/reference/datetime/versions.xml
@@ -5,103 +5,103 @@
 -->
 <versions>
  <!-- Methods -->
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

`gettimeofday`, `microtime` and `strptime` are actually part of the standard library, and as such the stubs are in ext/standard/basic_functions.stub.php.

The `PECL date_time:0.1` functions never made it into PHP 5.2; I'll remove these entries with a separate commit.

PR #379, opened 2021-01-23
Added PHP 8 into versions.xml for opcache based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/opcache/opcache.stub.php
- Notable changes were not found.


cmb69:
Thank you!

PR #380, opened 2021-01-23
$this is not a **reference** to the called object.

TysonAndre:
It's the value of the called object.

Additionally, in php 7.1, it was no longer possible to reassign $this
indirectly through a reference.
https://wiki.php.net/rfc/this_var#disable_ability_to_re-assign_this_indirectly_through

Even in php 7.0, I'd think it would initially be more like a value
(that references can be created from) than
a reference, but I may be misunderstanding what this means or what $this
did in php 5 or older.

(The manual dropped support for documenting php 5's behavior months ago, anyway)

### DIFF
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -47,7 +47,7 @@ class SimpleClass
    <para>
     The pseudo-variable <varname>$this</varname> is available when a
     method is called from within an object context.
-    <varname>$this</varname> is a reference to the calling object.
+    <varname>$this</varname> is the value of the calling object.
    </para>
    <warning>
     <para>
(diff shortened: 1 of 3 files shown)


Girgias:
Thanks for catching this, it might make sense to move the documentation of ``$this`` to the OOP section or the predefined variables section.

TysonAndre:
It already has documentation in https://www.php.net/manual/en/language.oop5.basic.php - which turns out to also need to be updated

cmb69:
I think this wording stems from PHP 4, where `$this` had to had reference semantics. The term "value" appears to be right as of PHP 5.0.0, but we should check the manual. E.g. https://www.php.net/manual/en/language.operators.assignment.php#language.operators.assignment states:

> An exception to the usual assignment by value behaviour within PHP occurs with objects, which are assigned by reference. Objects may be explicitly copied via the clone keyword.

It seems to me that we should clearly define the "value of objects" meaning the "object handle" (or something like that), opposed to the values of the objects properties (cf. `==`).

Girgias:
This is documented here: https://www.php.net/manual/en/language.oop5.references.php

PR #382, opened 2021-01-24
Added PHP 8 into versions.xml for bzip2 based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/bz2/bz2.stub.php
- Notable changes were not found.


cmb69:
Thank you!

PR #385, opened 2021-01-24
Added PHP 8 into versions.xml for xmlreader based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/xmlreader/php_xmlreader.stub.php
- Note
  * `xmlreader::resetstate` is not Yet Implemented in libxml, but is supposed to be alive.
    - https://github.com/php/php-src/blob/PHP-8.0/ext/xmlreader/php_xmlreader.c#L890-L895
    - This function does not exist in stub, so I do not add `PHP8` entry.

### DIFF
--- a/reference/xmlreader/versions.xml
+++ b/reference/xmlreader/versions.xml
@@ -5,33 +5,33 @@
 -->
 <versions>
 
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

PR #386, opened 2021-01-24
Added PHP 8 into versions.xml for uodbc based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/odbc/odbc.stub.php
- Note
  * `odbc_tableprivilegess` does not exist from ancient times.


cmb69:
Thank you for the PR!

`odbc_tableprivilegess` is almost certainly a typo. I'm going to remove that entry.

PR #387, opened 2021-01-24
Added PHP 8 into versions.xml for libxml based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/libxml/libxml.stub.php
- Note
  * `LibXMLError` class could not be found in stub, but already implemented apparently.
  * added its definition to stub.
    - https://github.com/php/php-src/pull/6637


cmb69:
Thank you!

PR #392, opened 2021-01-24
Added PHP 8 into versions.xml for xsl based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/xsl/php_xsl.stub.php
- Note
  * the following methods did not have `PHP 7` entry, but already implemented apparently.
    - XSLTProcessor::setProfiling
    - XSLTProcessor::setSecurityPrefs
    - XSLTProcessor::getSecurityPrefs
  * https://github.com/php/php-src/blob/php-7.0.0/ext/xsl/xsltprocessor.c#L100-L102

### DIFF
--- a/reference/xsl/versions.xml
+++ b/reference/xsl/versions.xml
@@ -5,20 +5,20 @@
 -->
 <versions>
 
- <function name="XSLTProcessor" from="PHP 5, PHP 7"/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! Nice catch regarding the 3 methods!

PR #393, opened 2021-01-24
Added PHP 8 into versions.xml for com_dotnet based on stubs.

mumumu:
  * https://github.com/php/php-src/blob/PHP-8.0/ext/com_dotnet/com_extension.stub.php
  * https://github.com/php/php-src/blob/PHP-8.0/ext/com_dotnet/com_persist.stub.php
- Note
  * `com_exception` did not exist in stub file, but implemented apparently in PHP 8, too.
    - https://github.com/php/php-src/blob/PHP-8.0/ext/com_dotnet/com_extension.c#L167-L170

### DIFF
--- a/reference/com/versions.xml
+++ b/reference/com/versions.xml
@@ -4,58 +4,58 @@
   Do NOT translate this file
 -->
 <versions> 
- <function name='com' from='PHP 4 &gt;= 4.1.0, PHP 5, PHP 7'/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! I'll fix the stubs wrt. to the missing `com_exception` shortly.

cmb69:
`com_exception` stub added with https://github.com/php/php-src/commit/bff7761653ab44c848e71f9d9727c7801588a50c.

PR #395, opened 2021-01-24
Added PHP 8 into versions.xml for classobj based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_builtin_functions.stub.php
- Note
  * `__autoload` was deleted in PHP 8.0.0, but its entry still alive.


cmb69:
Thank you!

PR #399, opened 2021-01-24
Added PHP 8 into versions.xml for snmp based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/snmp/snmp.stub.php
- Note
  * the following methods did not have `PHP 7` entry, but already implemented apparently.
    - https://github.com/php/php-src/blob/php-7.0.0/ext/snmp/snmp.c#L403-L406
      * snmp2_getnext
      * snmp2_real_walk
      * snmp2_set
      * snmp2_walk
  * `snmpexception` does not exist in stub, but implemented in PHP 8.0.0 apparently.
    - https://github.com/php/php-src/blob/PHP-8.0/ext/snmp/snmp.c#L2037-L2039

### DIFF
--- a/reference/snmp/versions.xml
+++ b/reference/snmp/versions.xml
@@ -4,46 +4,46 @@
   Do NOT translate this file
 -->
 <versions>
- <function name="snmp_get_quick_print" from="PHP 4, PHP 5, PHP 7"/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! Very thorough work, again! I'll add the `snmpexception` stub shortly.

cmb69:
`SNMPException` stub added with https://github.com/php/php-src/commit/a86443f5a35b9bba087fd78500290f5172ce38d5.

PR #400, opened 2021-01-24
Added PHP 8 into versions.xml for enchant based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/enchant/enchant.stub.php
- Note
  * the following functions were not implemented even in PHP 7.0.0, so maybe wrong entries.
    - enchant_free
    - enchant_new
    - enchant_new_dmy

### DIFF
--- a/reference/enchant/versions.xml
+++ b/reference/enchant/versions.xml
@@ -4,29 +4,29 @@
   Do NOT translate this file
 -->
 <versions> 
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! The three function have been renamed to `enchant_broker_free`, `enchant_broker_init` and `enchant_broker_dict_exists`, respectively, in the first released package (0.1). The are still used by their old names in the proto comments. I'm going to remove these entries right away.

PR #403, opened 2021-01-24
Added PHP 8 into versions.xml for funchand based on stubs.

mumumu:
  * https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_builtin_functions.stub.php
- Note
  * `create_function` was deleted as of PHP 8.0.0, but still alive entry.


cmb69:
Thank you!

PR #404, opened 2021-01-25
Added PHP 8 into versions.xml for outcontrol based on stubs.

mumumu:
  * https://github.com/php/php-src/blob/PHP-8.0/ext/zlib/zlib.stub.php
- Note
  * `ob_iconv_handler` does not exist in stub file, but already implemented in PHP8, too
    - https://github.com/php/php-src/blob/PHP-8.0/ext/iconv/iconv.c#L220-L221
    - https://github.com/php/php-src/blob/PHP-8.0/ext/iconv/iconv.c#L292-L345
  * `ob_tidyhandler` does not exist in stub file, but already implemented in PHP8, too
    - https://github.com/php/php-src/blob/PHP-8.0/ext/tidy/tidy.c#L851
    - https://github.com/php/php-src/blob/PHP-8.0/ext/tidy/tidy.c#L961-L1001

### DIFF
--- a/reference/outcontrol/versions.xml
+++ b/reference/outcontrol/versions.xml
@@ -4,25 +4,25 @@
   Do NOT translate this file
 -->
 <versions>
- <function name="flush" from="PHP 4, PHP 5, PHP 7"/>
- <function name="ob_clean" from="PHP 4 &gt;= 4.2.0, PHP 5, PHP 7"/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

Regarding `ob_iconv_handler` and `ob_tidyhandler`: these pseudo-functions cannot be called directly from userland (in fact, they are not defined there), and as such shouldn't have stubs.

PR #410, opened 2021-01-25
Added PHP 8 into versions.xml for misc based on stubs.

mumumu:
  * https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_builtin_functions.stub.php
- Note
  * the followings are language construct. these can be found in `Zend/zend_language_scanner.l`.
    - [__halt_compiler](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_language_scanner.l#L1697-L1699)
    - [die](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_language_scanner.l#L1389-L1391)
    - [eval](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_language_scanner.l#L1653-L1655)
    - [exit](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_language_scanner.l#L1385-L1387)

### DIFF
--- a/reference/misc/versions.xml
+++ b/reference/misc/versions.xml
@@ -4,37 +4,37 @@
   Do NOT translate this file
 -->
 <versions>
- <function name="__halt_compiler" from="PHP 5 &gt;= 5.1.0, PHP 7"/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

PR #411, opened 2021-01-25
Added PHP 8 into versions.xml for tidy based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/tidy/tidy.stub.php
- Note
  * the following methods were not implemented in PHP 7.0.0, even if PHP 5.6, so maybe mistake.
    - `tidynode::isxhtml`
    - `tidynode::isxml`

### DIFF
--- a/reference/tidy/versions.xml
+++ b/reference/tidy/versions.xml
@@ -5,48 +5,48 @@
 -->
 <versions>
 
- <function name="tidy" from="PHP 5, PHP 7, PECL tidy &gt;= 0.5.2"/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! Actually, `tidynode::isxhtml` [exists](https://github.com/php/php-src/blob/php-8.0.0/ext/tidy/tidy.c#L1531) (also in PHP 7), but `tidynode::isxml` indeed does not (that likely has been confused with `tidy::isxml`). I'm going to amend the PR in this regard.

PR #412, opened 2021-01-25
Added PHP 8 into versions.xml for network based on stubs.

mumumu:

- Note
  * `define_syslog_variables` was deleted as of PHP 5.4.0. We can safely delete it.


cmb69:
Thank you!

cmb69:
`define_syslog_variables` docs deleted with https://github.com/php/doc-en/commit/1fb41f1900406a130e59f37d6e66538a64048524.

PR #413, opened 2021-01-25
Added PHP 8 into versions.xml for apache based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/sapi/cgi/cgi_main.stub.php
- https://github.com/php/php-src/blob/PHP-8.0/sapi/apache2handler/php_functions.stub.php
- Note
  * `apache_reset_timeout` is function for Apache1 which reached EOL in ancient times.
    - We cannot find its implementation in even PHP 7.0.0, so can safely delete it.

### DIFF
--- a/reference/apache/versions.xml
+++ b/reference/apache/versions.xml
@@ -4,18 +4,18 @@
   Do NOT translate this file
 -->
 <versions> 
- <function name="apache_child_terminate" from="PHP 4 &gt;= 4.0.5, PHP 5, PHP 7"/>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

cmb69:
`apache_reset_timeout()` docs removed with https://github.com/php/doc-en/commit/89cb55b8608cbe21a6c65c5b06da42fa4f820a94 (thanks for pointing this out!)

PR #417, opened 2021-01-26
Fix example

big-dream:
Boolean to Int


cmb69:
Thank you for the PR! However, that parameter is of type bool as of PHP 8.0.0, what needs to documented as well as that it was of type int previously.

PR #418, opened 2021-01-26
Add missing visibility modifiers for SPL

villfa:
See:
* https://www.php.net/manual/en/class.splqueue.php
* https://www.php.net/manual/en/class.splstack.php


cmb69:
Thank you!

PR #420, opened 2021-01-26
Added PHP 8 into versions.xml for image based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/gd/gd.stub.php
- Note
  * the following functions were deleted as of PHP 8.0.0
    - image2wbmp
    - jpeg2wbmp
    - png2wbmp


cmb69:
Thank you!

PR #425, opened 2021-01-28
Update fgets.xml

tomasfejfar:
As described in the docs "If there is no more data to read in the file pointer, then false is returned."

### DIFF
--- a/reference/filesystem/functions/fgets.xml
+++ b/reference/filesystem/functions/fgets.xml
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>string</type><methodname>fgets</methodname>
+   <type>string|bool</type><methodname>fgets</methodname>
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/fgets.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>string</type><methodname>fgets</methodname>
+   <type>string|bool</type><methodname>fgets</methodname>
```
cmb69, line 12: Note that this should be written as:
```suggestion
   <type class="union"><type>string<type><type>bool</type></type><methodname>fgets</methodname>
```

cmb69 (commented):
Thank you for the PR!

cmb69:
Well, should actually be `string|false` not `string|bool`.

tomasfejfar:
> Well, should actually be string|false not string|bool.

Yes, I wasn't sure if this was a supported notation in PHP manual...

PR #426, opened 2021-01-28
Added PHP 8 into versions.xml for mysqli based on stubs.

mumumu:

- https://github.com/php/php-src/blob/PHP-8.0/ext/mysqli/mysqli.stub.php
- Notable changes
  * deleted functions as of PHP 8
    - `mysqli_embedded_server_start`
    - `mysqli_embedded_server_end`
    - `mysqli_driver::embedded_server_end`
    - `mysqli_driver::embedded_server_start`
  * not exist in manual nor PHP 5, PHP 7, PHP 8 source.
    - `mysqli_result_*` function, for example, `mysqli_result_current_field`
    - `mysqli_host_info`
  * exists in PHP5, PHP 7 source, but not exist in manual.
    - `mysqli_row_seek`
    - `mysqli_read_query_result`

### DIFF
--- a/reference/mysqli/versions.xml
+++ b/reference/mysqli/versions.xml
@@ -4,73 +4,73 @@
   Do NOT translate this file
 -->
 <versions>
- <function name="mysqli_affected_rows" from="PHP 5, PHP 7"/>
- <function name="mysqli_autocommit" from="PHP 5, PHP 7"/>
- <function name="mysqli_begin_transaction" from="PHP 5 &gt;= 5.5.0, PHP 7"/>
- <function name="mysqli_change_user" from="PHP 5, PHP 7"/>
- <function name="mysqli_character_set_name" from="PHP 5, PHP 7"/>
- <function name="mysqli_close" from="PHP 5, PHP 7"/>
- <function name="mysqli_commit" from="PHP 5, PHP 7"/>
- <function name="mysqli_connect" from="PHP 5, PHP 7"/>
- <function name="mysqli_connect_errno" from="PHP 5, PHP 7"/>
- <function name="mysqli_connect_error" from="PHP 5, PHP 7"/>
- <function name="mysqli_data_seek" from="PHP 5, PHP 7"/>
- <function name="mysqli_debug" from="PHP 5, PHP 7"/>
- <function name="mysqli_dump_debug_info" from="PHP 5, PHP 7"/>
+ <function name="mysqli_affected_rows" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_autocommit" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_begin_transaction" from="PHP 5 &gt;= 5.5.0, PHP 7, PHP 8"/>
+ <function name="mysqli_change_user" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_character_set_name" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_close" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_commit" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_connect" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_connect_errno" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_connect_error" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_data_seek" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_debug" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_dump_debug_info" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_embedded_server_end" from="PHP 5 &gt;= 5.1.0, PHP 7 &lt; 7.4.0"/>
  <function name="mysqli_embedded_server_start" from="PHP 5 &gt;= 5.1.0, PHP 7 &lt; 7.4.0"/>
- <function name="mysqli_errno" from="PHP 5, PHP 7"/>
- <function name="mysqli_error" from="PHP 5, PHP 7"/>
- <function name="mysqli_error_list" from="PHP 5 &gt;= 5.4.0, PHP 7"/>
- <function name="mysqli_escape_string" from="PHP 5, PHP 7"/>
- <function name="mysqli_execute" from="PHP 5, PHP 7"/>
- <function name="mysqli_fetch_all" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
- <function name="mysqli_fetch_array" from="PHP 5, PHP 7"/>
- <function name="mysqli_fetch_assoc" from="PHP 5, PHP 7"/>
- <function name="mysqli_fetch_field" from="PHP 5, PHP 7"/>
- <function name="mysqli_fetch_field_direct" from="PHP 5, PHP 7"/>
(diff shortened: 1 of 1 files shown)


reference/mysqli/versions.xml
```diff
@@ -4,73 +4,73 @@
...
+ <function name="mysqli_get_client_stats" from="PHP 5 &gt;= 5.3.0, PHP 7, PHP 8"/>
+ <function name="mysqli_get_client_version" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_get_connection_stats" from="PHP 5 &gt;= 5.3.0, PHP 7, PHP 8"/>
+ <function name="mysqli_get_host_info" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_get_links_stats" from="PHP 5 &gt;= 5.6.0, PHP 7, PHP 8"/>
+ <function name="mysqli_get_proto_info" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_get_server_info" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_get_server_version" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_get_warnings" from="PHP 5 &gt;= 5.1.0, PHP 7, PHP 8"/>
  <function name="mysqli_host_info" from="PHP 5, PHP 7"/>
```
kamil-tekiela, line 51: This function never existed. It must have been an oversight. It looks similar to `mysqli_get_host_info` which is why we probably missed it before.

reference/mysqli/versions.xml
```diff
@@ -4,73 +4,73 @@
...
+ <function name="mysqli_real_connect" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_real_escape_string" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_real_query" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_reap_async_query" from="PHP 5 &gt;= 5.3.0, PHP 7, PHP 8"/>
+ <function name="mysqli_refresh" from="PHP 5 &gt;= 5.3.0, PHP 7, PHP 8"/>
+ <function name="mysqli_release_savepoint" from="PHP 5 &gt;= 5.5.0, PHP 7, PHP 8"/>
+ <function name="mysqli_report" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_result_current_field" from="PHP 5, PHP 7"/>
  <function name="mysqli_result_data_seek" from="PHP 5, PHP 7"/>
  <function name="mysqli_result_fetch_all" from="PHP 5 &gt;= 5.3.0, PHP 7"/>
```
kamil-tekiela, line 76: Down to line 88. These functions don't exist and probably never did. Probably they were meant to be added as aliases

reference/mysqli/versions.xml
```diff
@@ -86,140 +86,140 @@
  <function name="mysqli_result_free" from="PHP 5, PHP 7"/>
  <function name="mysqli_result_lengths" from="PHP 5, PHP 7"/>
  <function name="mysqli_result_num_rows" from="PHP 5, PHP 7"/>
- <function name="mysqli_rollback" from="PHP 5, PHP 7"/>
+ <function name="mysqli_rollback" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_row_seek" from="PHP 5, PHP 7"/>
```
kamil-tekiela, line 90: These functions don't exist and probably never did. Probably they were meant to be added as aliases

reference/mysqli/versions.xml
```diff
@@ -4,73 +4,73 @@
...
+ <function name="mysqli_multi_query" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_next_result" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_num_fields" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_num_rows" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_options" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_ping" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_poll" from="PHP 5 &gt;= 5.3.0, PHP 7, PHP 8"/>
+ <function name="mysqli_prepare" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_query" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_read_query_result" from="PHP 5, PHP 7"/>
```
kamil-tekiela, line 66: This function never made it into the final mysqli release in PHP 5. We can remove it.

reference/mysqli/versions.xml
cmb69: This likely refers to the property `mysqli::$host_info`; maybe add a respective comment?

kamil-tekiela: Where should I add a comment? Shouldn't we just remove this line from this file?

cmb69: It might be worthwhile to actually check whether this refers to the property, i.e. to remove the version of the method, do a doc build and see whether the version information is still there. If that is the case, I think it would be good to keep the property version. An XML comment could be placed after the `<function>` element on the same line.

chack1172:
So after further investigation I found that embedded server functions have been removed with version 5.3.

chack1172:
@mumumu  @kamil-tekiela is there something I can do to help? Many clients don't want to update to PHP 8 because version is missing in documentation for mysqli xD

kamil-tekiela:
@chack1172 Good question. I think primarily someone needs to check the comment from CMB:
> It might be worthwhile to actually check whether this refers to the property, i.e. to remove the version of the method, do a doc build and see whether the version information is still there. If that is the case, I think it would be good to keep the property version. An XML comment could be placed after the `<function>` element on the same line.

We don't want to add wrong information or add too little. If you can validate that these entries are not needed then we can remove them, but as CMB said they might be required in the docs for technical reasons.

kamil-tekiela:
What can we do about this PR? This is a point of a lot of confusion. Can we maybe do another PR that fixes the things we are certain of, and then leave the rest until someone verifies how the manual works?

cmb69:
> Can we maybe do another PR that fixes the things we are certain of, and then leave the rest until someone verifies how the manual works?

Yes, sure.

kamil-tekiela (approved):

kamil-tekiela:
Thanks for the PR. I decided that this is good to merge as is. We can fix other issues with the versions later. There's no reason why they should block this.

reference/mysqli/versions.xml
salathe: @cmb69 some notes, to the best of my understanding... 

**Short version**

The versions from the `mysqli_get_host_info` line in *versions.xml* will get used instead, if the `mysqli_host_info` line gets removed.

**Long version**

So long as there is an item in *versions.xml* corresponding to one (or more) of the `refname` elements (in our case, in *…/mysqli/get-host-info.xml*), then the versions info will be available to add to the rendered page.  

When looking up versions info for the page, the (ref)names are tried in source order and if a match is found, that info is used. (Aside: for the versions lookup, special characters in the name are converted to underscores: `mysqli::$host_info` becomes `mysqli_host_info`).

Currently, the `refname` elements in `get-host-info.xml` have `mysqli::$host_info` first, and `mysqli_get_host_info` second.  So `mysqli_host_info` is the first name used (see "aside", above) in the versions lookup, and finds a match,  so those versions are used.

If `mysqli_host_info` was missing from *versions.xml*, then a match wouldn't be found, `mysqli_get_host_info` would be tried next, and would find a match.

PR #427, opened 2021-01-28
password_hash(): Fix for incorrect return type #77218

AllenJB:
Fix the return type (null instead of false on failure) and add a changelog entry for 8.0 error handling changes.

### DIFF
--- a/reference/password/functions/password-hash.xml
+++ b/reference/password/functions/password-hash.xml
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type class="union"><type>string</type><type>false</type></type><methodname>password_hash</methodname>
+   <type class="union"><type>string</type><type>null</type></type><methodname>password_hash</methodname>
    <methodparam><type>string</type><parameter>password</parameter></methodparam>
    <methodparam><type>mixed</type><parameter>algo</parameter></methodparam>
    <methodparam choice="opt"><type>array</type><parameter>options</parameter></methodparam>
@@ -162,7 +162,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns the hashed password, &return.falseforfailure;.
+   Returns the hashed password, or &null; on failure (with a warning) prior to 8.0.
   </para>
   <para>
    The used algorithm, cost and salt are returned as part of the hash. Therefore,
@@ -184,6 +184,12 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        Throws an Error instead of returning &null; with a warning on errors.
+       </entry>
+      </row>
       <row>
        <entry>7.4.0</entry>
        <entry>


cmb69:
Thanks for the PR! However, the ugly truth is that `password_hash()` could have returned `false` or `null` on failure. From what I can tell from looking at the code, `null` returns where accompanied by `E_WARNING`, but `false` returns where silent. And it *seems* to me that the function returned `null` on unexpected input, what would be covered by the note on https://www.php.net/manual/en/functions.internal.php.

AllenJB:
Can you provide an example of "unexpected input"? I tested an invalid value for each of the parameters (`null` for hash, "invalid" for the algorithm, and cost values of both "foo" and 0) on 3v4l and they all returned null for all versions prior to PHP 8 that supported password_hash.

I don't believe developers should be expected to have ever read that internal functions page in the manual (I don't recall reading it before today, altho I probably have at some point and forgotten) - all possible return values should be documented on the functions page (at least for supported PHP versions).

cmb69:
`false` could have been returned prior to PHP 7.4.0, for instance if a 1 GB long salt was passed, or for some unlikely to occur internal errors.

All PHP 5 and 7 programmers should be aware that almost all internal functions return `null` (or sometimes `false`) for input which can not be juggled to the expected type, e.g. https://3v4l.org/UIYGh. I really don't think it makes sense to document this, because these are programmer *errors* (as of PHP 8.0.0 all functions and methods in php-src throw a `TypeError` instead). Besides that there is practical limitation: there are about 10,000 functions in the manual…

AllenJB:
I don't think it is right to say "all PHP developers" (from "my first PHP script" to veterans of 20 years) should be aware of a note on a page no one reads.

In my opinion the manual page for a function should accurately and completely document all possible return values and error handling behaviors. The PHP developers have been doing a great job of fixing these issues in PHP 8, but many developers are still on PHP 7 and may be for some time to come.

As an alternative, is there some method where a note could be added to every page of the manual? Perhaps as a template that's added to the return type section?

In this case it's far more likely developers are going to pass through invalid parameter values than 1GB salt (I'm sure that does happen), and they are definitely being surprised by the return values, which shouldn't be the case, especially in the case of a security related function like password_hash. (It was a conversation on Freenode ##php that led me to the bug in search results - which is still open and expecting docs to be updated since 2019 - and hence this PR)

Girgias:
> I don't think it is right to say "all PHP developers" (from "my first PHP script" to veterans of 20 years) should be aware of a note on a page no one reads.
> 
> In my opinion the manual page for a function should accurately and completely document all possible return values and error handling behaviors. The PHP developers have been doing a great job of fixing these issues in PHP 8, but many developers are still on PHP 7 and may be for some time to come.
> 

You're basically invoking undefined behaviour and you get a warning, I don't think documenting *every single* function and page makes *any* sense, and that's ignoring the ***massive*** burden on translation this would just throw on them.


> As an alternative, is there some method where a note could be added to every page of the manual? Perhaps as a template that's added to the return type section?
> 

Not every function is subject to this issue so this cannot be done.

afilina:
Although we cannot realistically update all 10,000 functions in the manual, does that fact prevent us from mentioning `false` on this page? I don't believe that the two things are mutually exclusive.

cmb69:
[According to the stubs](https://heap.space/xref/PHP-8.0/ext/standard/basic_functions.stub.php?r=b10bcb4f#1137), `password_hash()` always returns `string` as of PHP 8.0.0.

Generally, I don't think it would be a good idea to actually document what a function returns in case the programmer triggered undefined behavior. Besides that this would be a contradiction, starting with a single function could easily trigger a lot of PRs or doc issues.

Crell:
Closing due to clear lack of interest.

PR #433, opened 2021-02-03
Update for function.getopt parameters description.

markskayff:
It's somewhat confusing on what's the right formatting for the **long_options** parameter .
The current docs clearly describes how the **short_options** should be described but leaves most of the long_options ones to the imagination. There's a brief note below but does not shed much light about it. 

This is an attempt to clearly describe what the format should be for the **long_options** param on the getopt() function.


cmb69:
Thank you!

PR #434, opened 2021-02-05
Change abs parameter to int|float

pableu:
The behavior changed in PHP 8. Passing "foo" always throws a TypeError, passing a number as string throws a TypeError if `strict_types=1`.


Crell:
I think this needs a changelog block to go with it, to indicate that it got stricter in 8.0.  (Even in strict mode, "4" works in PHP 7.x.)

cmb69:
Yes, that would need a changelog entry, plus maybe further clarification. Also note that this is already covered by PR #289.

Girgias:
This has been fixed via #289

PR #435, opened 2021-02-05
[apcu] Clarify the effect of apc.ttl

mszabo-wikia:
The current documentation implies that the apc.ttl setting would mark all cache
entries as liable to be evicted if they have not been accessed in this many
seconds. However, apc_cache_entry_expired() and apc_cache_entry_soft_expired()
reveal that apc.ttl is only taken into account for entries that do not have an
explicit TTL set. Accordingly, clarify that the effect of apc.ttl is limited
only to entries that do not have an explicit TTL set.

### DIFF
--- a/reference/apcu/ini.xml
+++ b/reference/apcu/ini.xml
@@ -18,14 +18,23 @@
   is bundled with the extension should be copied somewhere into the docroot and
   viewed with a browser as it provides a detailed analysis of the internal 
   workings of APCu. If GD is enabled in PHP, it will even display some
-  interesting graphs. If APCu is working, the <literal>Cache full count
+  interesting graphs.</para>
+  <para>
+  If APCu is working, the <literal>Cache full count
   </literal> number (on the left) will display the number of times the cache
-  has reached maximum capacity and has had to forcefully clean any entries that
-  haven't been accessed in the last <literal>apc.ttl</literal> seconds. This
-  number is minimized in a well-configured cache.  If the cache is constantly
-  being filled, and thusly forcefully freed, the resulting churning will have
-  disparaging effects on script performance. The easiest way to minimize this 
-  number is to allocate more memory for APCu.
+  has reached maximum capacity and has had to evict entries to free up memory.
+  During eviction, if <literal>apc.ttl</literal> was specified, APCu will first
+  attempt to remove expired entries, i.e. entries whose TTL has either expired,
+  or entries that have no TTL set and haven't been accessed in the last
+  <literal>apc.ttl</literal> seconds. If <literal>apc.ttl</literal> was not set,
+  or removing expired entries did not free up enough space, APCu will clear the
+  entire cache.
+  </para>
+  <para>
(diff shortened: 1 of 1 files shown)


mszabo-wikia:
CC'ing APCu maintainers for feedback/review: @nikic @remicollet

afilina:
Ping @nikic @remicollet

reference/apcu/ini.xml
```diff
@@ -20,12 +20,13 @@
...
   </literal> number (on the left) will display the number of times the cache
-  has reached maximum capacity and has had to forcefully clean any entries that
-  haven't been accessed in the last <literal>apc.ttl</literal> seconds. This
-  number is minimized in a well-configured cache.  If the cache is constantly
-  being filled, and thusly forcefully freed, the resulting churning will have
-  disparaging effects on script performance. The easiest way to minimize this 
-  number is to allocate more memory for APCu.
+  has reached maximum capacity and has had to forcefully clean expired entries
+  (i.e. entries whose TTL has either expired, or entries that have no TTL set
+  and haven't been accessed in the last <literal>apc.ttl</literal> seconds.
```
nikic, line 25: Missing a closing `)`

nikic: The actual behavior is more complicated than that: If a global TTL is set, then there will be an attempt to clear all hard expired (entry TTL) and soft expired (global TTL) entries. If that does not free enough space, or if no global TTL is set, then the cache is fully cleared.

reference/apcu/ini.xml
```diff
@@ -200,16 +201,12 @@
...
-      available memory, the cache will be completely
-      expunged if ttl is equal to 0.  Otherwise, if the
-      ttl is greater than 0, APC will attempt to remove
-      expired entries.
+      Consider cache entries without an explicit TTL to be
+      expired if they have not been accessed in this many
+      seconds. Effectively, this allows such entries to be
+      removed opportunistically during a cache insert, or
+      prior to a full expunge. This setting has no effect on
+      cache entries with an explicit TTL set.
```
nikic, line 209: It may be worth noting that entries can still be readable after `apc.ttl` expires. As you note, they are removed opportunistically, unlike the hard per-entry TTL.

reference/apcu/ini.xml
mszabo-wikia: Thank you. I've expanded the section to clarify the behavior.

reference/apcu/ini.xml
mszabo-wikia: Thanks. I added a note about that.

mszabo-wikia:
I've rebased this patch and fixed a typo as well

afilina:
Pinging @nikic to approve the changes so we can merge.

tstarling:
I reviewed the change and @nikic's comments. It looks good to me, and I can see that @nikic's suggestions have been incorporated. Ideally we would have maintainer approval, but it's been more than a year now.

PR #437, opened 2021-02-05
Added missing documentation for openssl_pkey_derive

dvaeversted:
Adding documentation for `openssl_pkey_derive`.

Was introduced here: https://github.com/php/php-src/pull/3197
As a result of: https://bugs.php.net/bug.php?id=76026

### DIFF
--- a/reference/openssl/functions/openssl-pkey-derive.xml
+++ b/reference/openssl/functions/openssl-pkey-derive.xml
@@ -1,0 +1,135 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+
+<refentry xml:id="function.openssl-pkey-derive" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>openssl_pkey_derive</refname>
+  <refpurpose>Computes shared secret for public value of remote and local DH or ECDH key</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type class="union"><type>string</type><type>false</type></type><methodname>openssl_pkey_derive</methodname>
+   <methodparam><type class="union"><type>OpenSSLAsymmetricKey</type><type>OpenSSLCertificate</type><type>array</type><type>string</type></type><parameter>public_key</parameter></methodparam>
+   <methodparam><type class="union"><type>OpenSSLAsymmetricKey</type><type>OpenSSLCertificate</type><type>array</type><type>string</type></type><parameter>private_key</parameter></methodparam>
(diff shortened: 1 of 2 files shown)


reference/openssl/functions/openssl-pkey-derive.xml
```diff
@@ -0,0 +1,138 @@
...
+  <refpurpose>Computes shared secret for public value of remote and local DH or ECDH key</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type class="union"><type>string</type><type>false</type></type><methodname>openssl_pkey_derive</methodname>
+   <methodparam><type>mixed</type><parameter>peer_pub_key</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter>priv_key</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>keylen</parameter><initializer>NULL</initializer></methodparam>
```
cmb69, line 16: Please use the signature (parameter names, types and initializers) as defined in the [PHP 8.0 stub](https://github.com/php/php-src/blob/e2c30c6a2ee4968b546e08eb19cb5f483c4c8088/ext/openssl/openssl.stub.php#L209-L213). If there would be relevant difference to PHP 7, these should be noted in a changelog.

Please don't forget to adjust the parameter names below.

reference/openssl/functions/openssl-pkey-derive.xml
```diff
@@ -0,0 +1,138 @@
...
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   The derived secret on success&return.falseforfailure;.
+  </para>
+  <para>
+   
+  </para>
```
cmb69, line 64: Superfluous paragraphp.

reference/openssl/functions/openssl-pkey-derive.xml
```diff
@@ -0,0 +1,138 @@
...
+ </refsect1>
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>openssl_pkey_derive</function> example</title>
+    <programlisting role="php">
+     <![CDATA[
```
cmb69, line 73: The convention is not to indent `<![CDATA[` tags (also below for the closing tag).

cmb69 (changes requested):
Thank you! Looks pretty good, but please address the comments below.

reference/openssl/functions/openssl-pkey-derive.xml
dvaeversted: I am not aware of any differences from PHP 7 and 8 in this regard, the rest of issues has been corrected.

dvaeversted:
Thank you for looking at it, i have corrected the items you commented on.
Please let me know if there is anything else that needs to be done.

cmb69:
Thank you!

PR #438, opened 2021-02-07
Add a note that json is always available in php 8.0

TysonAndre:
See https://wiki.php.net/rfc/always_enable_json


cmb69:
Thank you!

PR #439, opened 2021-02-07
Update APCu settings defaults documentation

TysonAndre:
https://pecl.php.net/package-changelog.php?package=APCu&release=5.1.19


cmb69:
Thank you! I'll reword the changlog entries.

PR #440, opened 2021-02-07
Add the standard PECL installation instructions to APCu

TysonAndre:
https://www.php.net/manual/en/apcu.installation.php would be confusing
to new users. A recent user note coincidentally mentions this.
(noticed while looking for how installation instructions are written in general)

This starts the installation section with the following message:

> Information for installing this PECL extension may be found in the manual chapter titled [Installation of PECL extensions](https://www.php.net/manual/en/install.pecl.php). Additional information such as new releases, downloads, source files, maintainer information, and a CHANGELOG, can be located here: » https://pecl.php.net/package/apcu.

### DIFF
--- a/reference/apcu/setup.xml
+++ b/reference/apcu/setup.xml
@@ -12,6 +12,10 @@
  <section xml:id="apcu.installation">
   &reftitle.install;
   <para>
+   &pecl.info;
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

PR #442, opened 2021-02-09
Allow Throwable for $previous param

Sebiworld:
In current PHP versions the Exception-class seems to allow all Throwable values for its constructor's $previous param instead of only values of type Exception.


cmb69:
Thank you!

PR #444, opened 2021-02-11
Remove incorrect note about autoloading in interactive CLI

IMSoP:
This was apparently added in response to a bug report against PHP
5.2 [https://bugs.php.net/bug.php?id=40775]. Testing suggests the
limitation was removed at least as long ago as 5.6, and probably
earlier.

### DIFF
--- a/language/oop5/autoload.xml
+++ b/language/oop5/autoload.xml
@@ -27,12 +27,6 @@
   </tip>
   <note>
    <para>
-    Autoloading is not available if using PHP in CLI
-    <link linkend="features.commandline">interactive mode</link>.
-   </para>
-  </note>
-  <note>
-   <para>
     If the class name is used e.g. in <function>call_user_func</function> then
     it can contain some dangerous characters such as <literal>../</literal>.
     It is recommended to not use the user-input in such functions or at least


morrisonlevi (approved):
This was fixed at least as far back as PHP 5.4:
```php
$ php --version
PHP 5.4.45 (cli) (built: Nov  2 2020 21:56:05) (DEBUG)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2014 Zend Technologies
$ php -a 
Interactive shell

php > spl_autoload_register(function($c) { echo "Autoloading $c"; eval("class $c{}"); });
php > $a = new Foo;
Autoloading Foo
php > var_dump($a);
object(Foo)#2 (0) {
}
```

cmb69:
Thank you! I'm not sure whether `__autoload()` works the same, but I don't really bother. I'm going to remove the respective info features/commandline right away.

cmb69:
Screwed up GH tag: applied as 99b8716337e2c27b6addf7e184b8ac7a50bb38b0.

IMSoP:
Ah, I didn't spot that the note was added both ways around; good catch. :)

I did check `__autoload` too, and it works fine (until 8.0, when of course it doesn't work anywhere).

PR #446, opened 2021-02-13
Fixed simplexml_import_dom return type

ondrejmirtes:
The PHP 8 stubs say that `simplexml_import_dom` returns `null` on failure. I verified that: https://3v4l.org/9fVRO


cmb69:
Thank you!

PR #447, opened 2021-02-13
In OpenSSL extension, configargs parameter name was renamed to options.

mumumu:
Maybe, should we replace this `configargs` too?


cmb69 (approved):
Good catch, thanks!

PR #448, opened 2021-02-15
Delete the PHP 5 SPL_Types extension.

TysonAndre:
Months ago, PHP removed php 5 documentation and changelogs
of core functionality - I don't see a reason to continue documenting SPL_Types either

- https://pecl.php.net/package/SPL_Types was last updated in 2012
  and is not maintained.
- SPL_Types is easy to confuse with the SPL because the same prefix is used
- SPL_Types can cause confusion because PHP has typed properties(7.4),
  and will probably also have actual enums in 8.1
  https://www.php.net/spl_types


cmb69:
Thank you!

PR #450, opened 2021-02-16
Added full example of overwriting for array

AlexandrErohin:
Added final example includes all variations of type casting and overwriting for array.
This is done to simplify the understanding of working with array keys for those who are starting to study PHP, as you can see all the possibilities of keys in one example.


cmb69:
Thank you!

PR #451, opened 2021-02-16
Added array dereferencing example with [] style

AlexandrErohin:
Added array dereferencing example with [] style since PHP 7.1

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -326,6 +326,9 @@ $secondElement = getArray()[1];
 
 // or
 list(, $secondElement) = getArray();
+
+// or from PHP 7.1
+[, $secondElement] = getArray();
 ?>
 ]]>
     </programlisting>
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -327,7 +327,7 @@ $secondElement = getArray()[1];
 // or
 list(, $secondElement) = getArray();
 
-// or from PHP 7.1
+// or since PHP 7.1
 [, $secondElement] = getArray();
 ?>
 ]]>
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -324,11 +324,14 @@ function getArray() {
 
 $secondElement = getArray()[1];
 
-// or
+// or with list
 list(, $secondElement) = getArray();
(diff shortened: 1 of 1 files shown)


language/types/array.xml
```diff
@@ -324,8 +324,14 @@ function getArray() {
 
 $secondElement = getArray()[1];
 
-// or
+// or using list
 list(, $secondElement) = getArray();
+// or using list since PHP 7.1
+list(2 => $secondElement) = getArray();
```
afilina, line 330: Index 2 refers to the 3rd element. It should be index 1.

language/types/array.xml
```diff
@@ -324,8 +324,14 @@ function getArray() {
...
+
+// or using [] style since PHP 7.1
+[, $secondElement] = getArray();
+[2 => $secondElement] = getArray();
```
afilina, line 334: Index 2 refers to the 3rd element. It should be index 1.

language/types/array.xml
```diff
@@ -324,8 +324,14 @@ function getArray() {
 
 $secondElement = getArray()[1];
 
-// or
+// or using list
 list(, $secondElement) = getArray();
+// or using list since PHP 7.1
+list(2 => $secondElement) = getArray();
+
+// or using [] style since PHP 7.1
```
afilina, line 332: Perhaps it would make sense to split the `list` and `[]` examples, or at the very least add a newline and comment between each of the 4 example.

cmb69:
It looks like this has some overlap with PR #145, and it might be best to keep array destructuring (and list()) out of the already too long [array page](https://www.php.net/manual/en/language.types.array.php).

AlexandrErohin:
My main idea was to add examples of using [] style to main documentation (currently it is not included to array page) and only can be found in the migration page
https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.symmetric-array-destructuring

Maybe it is better to add to the array page - separate example of Array dereferencing since php 7.1 ?

afilina:
@cmb69 @AlexandrErohin Is this PR still relevant?

PR #452, opened 2021-02-16
Update methodsynopsis: getenv() may return false

villfa:
See https://www.php.net/manual/en/function.getenv.php


cmb69 (approved):
Thank you!

PR #453, opened 2021-02-16
Added protected property for Object To Array example

AlexandrErohin:


cmb69:
Thank you for the PR!

I don't think that a protected property should be added to *this* example, because the example is about the potential confusion regarding private and public properties. And indeed, `var_dump()` will usually not show the NUL bytes, so there are seemingly two elements with the same key `"AA"` (https://3v4l.org/mKoY3).  Maybe there should be an additional example first, which uses `var_export()` (something like https://3v4l.org/oDXeG).

AlexandrErohin:
@cmb69 Thank you! Agree. I have updated examples.

cmb69:
Thank you!

PR #454, opened 2021-02-16
Updated actual functions using in examples of get_resource_type

AlexandrErohin:
mysql_connect is used in the example for get_resource_type which is deprecated since PHP 5.5.0. As it is better to use actual functions in examples -  it was replaced by pg_pconnect with similar example output

### DIFF
--- a/reference/var/functions/get-resource-type.xml
+++ b/reference/var/functions/get-resource-type.xml
@@ -58,8 +58,8 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-// prints: mysql link
-$c = mysql_connect();
+// prints: pgsql link persistent
+$c = pg_pconnect();
 echo get_resource_type($c) . "\n";
 
 // prints: stream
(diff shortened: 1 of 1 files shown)


cmb69:
Ah, good catch! However, `pg_pconnect()` has 1 mandatory parameter, so the example won't work as is. Even worse, the following example calls `new_xmldoc()`, which is removed as of PHP 5.0.0. Maybe we should just stick with the `fopen()` example?

AlexandrErohin:
@cmb69 you are right! and I have missed new_xmldoc. ah. I have updated the example. I suggest to add curl instead mysql_connect and pg_pconnect. As two examples are better then one :)

cmb69:
Thank you! I agree that more examples are good, but note that `curl_init()` no longer returns a resource as of PHP 8.0.0. This is also the case for [many other functions](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.resource2object), and more resource to object migrations are likely to happen in future minor releases. I think it is okay to stick with the `curl_init()` example, but there should be a comment pointing out that this only works prior to PHP 8.0.0.

AlexandrErohin:
@cmb69 Thank you! Agree. I have added the comment to curl_init

cmb69 (approved):
Thank you!

PR #455, opened 2021-02-17
Update json-encode.xml

etkaar:
Should be `flags` instead of `options` because argument `$options` was renamed to `$flags`.


cmb69 (approved):
Thank you!

PR #457, opened 2021-02-18
Explain what consistitutes a more specific type

Girgias:
As it's not just relevant to a single object type.

@tiffany-taylor might want to look at this as she wrote the whole document.


tiffany-taylor (approved):
Looks good to me.

PR #458, opened 2021-02-22
Document the MYSQLI_TYPE_JSON constant

mauriciofauth:
https://github.com/php/php-src/blob/php-8.0.2/ext/mysqli/mysqli.c#L760-L762
https://github.com/php/php-src/blob/php-8.0.2/ext/mysqli/tests/mysqli_constants.phpt#L200-L202
https://dev.mysql.com/doc/refman/5.7/en/json.html


cmb69:
Thank you!

PR #459, opened 2021-02-23
Fix operator associativity of null coalescing operator

dennispansegrau:
The operator associativity of the null coalescing operator should be left instead of right.
Here an example from the documentation. It is solved from left to right...

<?php
$foo = null;
$bar = null;
$baz = 1;
$qux = 2;

echo $foo ?? $bar ?? $baz ?? $qux; // outputs 1
?>

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -216,7 +216,7 @@
         </entry>
        </row>
        <row>
-        <entry>right</entry>
+        <entry>left</entry>
         <entry><literal>??</literal></entry>
         <entry>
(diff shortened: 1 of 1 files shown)


nikic:
`??` is right-associative, as currently documented: https://github.com/php/php-src/blob/da011a312a6c6cd7ff12fe1aa0de1e33fba2f167/Zend/zend_language_parser.y#L66

cmb69:
So we can close this PR. :)

dennispansegrau:
Ok, sorry. I thought it's wrong. Maybe someone can explain to me why it is right associative?

PR #460, opened 2021-02-23
Update index_key handling in array_column for php8

Riimu:
As per PR https://github.com/php/php-src/pull/5487, the `array_column()` now uses standard type handling for keys. As a result, if an object is used for the key, it will no longer be cast to string.

Example of the behavior change: https://3v4l.org/1qoLQ


afilina:
LGTM

cmb69:
Thank you!

PR #463, opened 2021-02-25
Clarify that Exception::getPrevious may return a Throwable

M1ke:
Spotted this when using PhpStorm's documentation (which copies the PHP docs) which incorrectly typehints an Exception - use of this as Exception could cause a TypeError if a non-Exception Throwable is returned. The definitions here for return params of this method are correct so its just the top line description that needs to change.

### DIFF
--- a/language/predefined/exception/getprevious.xml
+++ b/language/predefined/exception/getprevious.xml
@@ -4,7 +4,7 @@
 <refentry xml:id="exception.getprevious" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>Exception::getPrevious</refname>
-  <refpurpose>Returns previous Exception</refpurpose>
+  <refpurpose>Returns previous Throwable</refpurpose>
  </refnamediv>
  
  <refsect1 role="description">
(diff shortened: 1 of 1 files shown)


language/predefined/exception/getprevious.xml
```diff
@@ -14,7 +14,7 @@
    <void />
   </methodsynopsis>
   <para>
-   Returns previous exception (the third parameter of <methodname>Exception::__construct</methodname>).
+   Returns previous throwable (which may be an exception or error, passed as the third parameter of <methodname>Exception::__construct</methodname>).
```
Girgias, line 17:
```suggestion
   Returns previous <classname>Throwable</classname> (which may be an <classname>Exception</classname> or <classname>Error</classname, passed as the third parameter of <methodname>Exception::__construct</methodname>).
```

Also can you add a line break in that line so it's less than 80 characters? (it's not a hard rule but helps with looking at diffs in a terminal)

Girgias (changes requested):

language/predefined/exception/getprevious.xml
cmb69: IMO there's no need to explicitly tell that a Throwable is an Exception or Error, here.

cmb69:
Thank you for the PR. I'm going to apply a slightly modified variant.

PR #464, opened 2021-02-26
Fixed Ticks usage example

AlexandrErohin:
There is a mistake in Tick usage example. There is `tick_handler();` after if clause - but it should not. The last tick event in the example is `print($a);`

### DIFF
--- a/language/control-structures/declare.xml
+++ b/language/control-structures/declare.xml
@@ -122,13 +122,13 @@ function tick_handler()
     echo "tick_handler() called\n";
 }
 
-register_tick_function('tick_handler');
+register_tick_function('tick_handler'); // this is a tick event
 
-$a = 1;
+$a = 1; // this is a tick event
 
 if ($a > 0) {
-    $a += 2;
-    print($a);
+    $a += 2; // this is a tick event
+    print($a); // this is a tick event
 }
 
 ?>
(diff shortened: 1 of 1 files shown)


cmb69:
Ah, thanks, you're right. However, isn't the `if` a tick event? See https://3v4l.org/HUWGN.

AlexandrErohin:
@cmb69 I thought, that it is working this way
https://3v4l.org/HUWGN
1) first tick event is - function call of register_tick_function
2) second tick event is print - print 1; Where function tick event occurs after execution of print
3) third tick event is print - print 2;
4) fourth tick event is print - print 3;
And according to https://www.php.net/manual/en/control-structures.declare.php
```
Not all statements are tickable. Typically, condition expressions and argument expressions are not tickable.
```
So I think `if` is not a tick event

afilina:
I'm not sure I understand the purpose of the second example. If it was to highlight the occurrences of the tick event from the first example, then the comments on the first example would be sufficient, and the second example shouldn't be necessary.

cmb69:
Oh, you're right! I think it's indeed best to drop the second example altogether.

AlexandrErohin:
@cmb69 @afilina Good point! Thanks. I have deleted the second example and made the comments more clear

cmb69:
Thank you! Now that part of the documentation finally makes sense.

I'm going to shorten the comments for better readability, though; at least the comment on the register_tick_function() call is likely to wrap, what doesn't look good.

PR #465, opened 2021-02-26
Show info about PHP error reporting

kestutisv:
Reader needs to be informed about PHP errors.

### DIFF
--- a/language/types/type-juggling.xml
+++ b/language/types/type-juggling.xml
@@ -29,8 +29,8 @@
 $foo = "1";  // $foo is string (ASCII 49)
 $foo *= 2;   // $foo is now an integer (2)
 $foo = $foo * 1.3;  // $foo is now a float (2.6)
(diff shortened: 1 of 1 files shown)


cmb69:
Ah, thanks, good catch! We should note, though, that no error was raised [prior to PHP 7.1.0](https://3v4l.org/Q6j5M), and I'm also concerned regarding the line length (examples just wrap around); maybe move the comment about the errors *above* the "10 little piggies" line, and use multiple lines for it, and state that it applies to the following two lines?

cmb69 (changes requested):
After having had a closer look, I think we should split that example (i.e. the last two statements). That could better incorporate the somewhat detached comment regarding these statements, so we don't have to add these comments to the example in the first place.

afilina:
@kestutisv Pinging author.

kamil-tekiela:
This page has been completely redesign in the meantime. This PR is now outdated. If the new page still needs adjusting, please raise a new PR.

PR #466, opened 2021-02-28
Rewrite debug_zval_dump manual based on current (>=7.0) behaviour

IMSoP:
- The recommended use of a call-time pass-by-reference has been
  impossible since PHP 5.4, making the examples unusable.
- The re-design of references in PHP 7.0 means that the refcount
  no longer reflects these.


cmb69:
Thank you very much!

PR #467, opened 2021-03-02
Move iterable variance section to OOP section

Girgias:
Makes more sense to group it together but we lose the example.

### DIFF
--- a/language/oop5/variance.xml
+++ b/language/oop5/variance.xml
@@ -31,6 +31,11 @@
      <type>float</type> is changed to <type>int</type>
     </simpara>
    </listitem>
+   <listitem>
+    <simpara>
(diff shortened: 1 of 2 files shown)


cmb69:
Makes sense to me. Thanks! I wonder, though, whether we need that special casing at all, or if it would be sufficient that `iterable` basically is `array|Traversable` (and accordingly, variance rules apply like for union types in general).

Girgias:
> Makes sense to me. Thanks! I wonder, though, whether we need that special casing at all, or if it would be sufficient that `iterable` basically is `array|Traversable` (and accordingly, variance rules apply like for union types in general).

Probably makes sense tbh, just going to merge this in the mean time. :-)

PR #469, opened 2021-03-03
Clarify the syntax and role of arguments for echo and print

IMSoP:
- Rather than implying that parentheses are optional, make clear that they are never part of the syntax for these keywords.
- Don't spend so many examples on different types of string.
- Do include examples of non-string arguments.
- Add a note with additional examples of how parentheses interact.
- Don't refer to either construct as "function-like", instead use the term "expression" consistently.
- Additional tidying to make the pages more consistent with each other.
- Add "print" to the list of language constructs which are documented as functions.

(Inspired by [this Stack Overflow question](https://stackoverflow.com/q/66452331/157957).)

### DIFF
--- a/reference/strings/functions/echo.xml
+++ b/reference/strings/functions/echo.xml
@@ -10,27 +10,23 @@
   &reftitle.description;
   <methodsynopsis>
    <type>void</type><methodname>echo</methodname>
-   <methodparam><type>string</type><parameter>arg</parameter></methodparam>
-   <methodparam rep="repeat"><type>string</type><parameter>args</parameter></methodparam>
+   <methodparam rep="repeat"><type>string</type><parameter>expressions</parameter></methodparam>
   </methodsynopsis>
   <simpara>
-   Outputs all parameters. No additional newline is appended.
+   Outputs one or more expressions, with no additional newlines or spaces.
   </simpara>
   <para>
-   <literal>echo</literal> is not actually a function (it is a
-   language construct), so you are not required to use parentheses
-   with it. <literal>echo</literal> (unlike some other language
-   constructs) does not behave like a function, so it cannot
-   always be used in the context of a function. Additionally, if you want to
-   pass more than one parameter to <literal>echo</literal>, the parameters
-   must not be enclosed within parentheses.
+   <literal>echo</literal> is not a function but a language language construct.
+   Its arguments are a list of expressions following the <literal>echo</literal>
+   keyword, separated by commas, and not delimited by parentheses.
+   Unlike some other language constructs, <literal>echo</literal> does not have
+   any return value, so it cannot be used in the context of an expression.
   </para>
   <para>
    <literal>echo</literal> also has a shortcut syntax, where you can
-   immediately follow the opening tag with an equals sign. Prior to PHP 5.4.0,
-   this short syntax only works with the
-   <link linkend="ini.short-open-tag">short_open_tag</link> configuration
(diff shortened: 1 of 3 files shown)


IMSoP:
The method synopsis at the top of both pages will still display with parentheses. Ideally, we'd use a different DocBook markup to give synopses for non-function constructs like this.

reference/strings/functions/echo.xml
```diff
@@ -50,17 +47,14 @@ I have <?=$foo?> foo.
...
-    <varlistentry>
-     <term><parameter>args</parameter></term>
+     <term><parameter>expressions</parameter></term>
      <listitem>
       <para>
+       One or more string expressions to output, separated by commas.
+       Non-string values will be coerced to strings, even when
+       <link linkend="language.types.declarations.strict">the
+       <literal>strict_types</literal> directive</link> is set to
+       <literal>1</literal>.
```
Girgias, line 57:
```suggestion
       <literal>strict_types</literal> directive</link> is enabled.
```
Maybe?

reference/strings/functions/echo.xml
```diff
@@ -109,10 +126,51 @@ echo $some_var ? 'true': 'false'; // changing the statement around
...
+    </programlisting>
+   </example>
+  </para>
+  </note>
+
   <tip>
    <para>
     A benefit to passing in multiple arguments over using concatenation in
-    <function>echo</function> regards the precedence of the period operator in
+    <literal>echo</literal> regards the precedence of the period operator in
```
Girgias, line 173: Is this still accurate as of PHP 8? Precedence of `.` was changed

reference/strings/functions/print.xml
```diff
@@ -31,10 +32,12 @@
...
    <variablelist>
     <varlistentry>
-     <term><parameter>arg</parameter></term>
+     <term><parameter>expression</parameter></term>
      <listitem>
       <para>
-       The input data.
+       The expression to be output. Non-string values will be coerced to strings,
+       even when <link linkend="language.types.declarations.strict">the
+       <literal>strict_types</literal> directive</link> is set to <literal>1</literal>.
```
Girgias, line 40: Same possible improvement as with echo.

Girgias (commented):
Looks rather good except that a clarification could be needed about precedence with `.`

reference/strings/functions/echo.xml
IMSoP: Good catch, the `+` example now does the right thing, but the ternary still needs the parentheses: https://3v4l.org/Rgp73

Girgias: Ah yes good point about the ternary.

IMSoP: Updated. I've left the addition example in, because there's still plenty of people running 7.x. I have removed a reference to pre-5.4 behaviour elsewhere, though.

cmb69:
Very good PR! Thank you!

> Ideally, we'd use a different DocBook markup to give synopses for non-function constructs like this.

ACK. Maybe adding a "role" attribute and supporting this in PhD might be a viable solution.

Generally, I am more concerned regarding the constrol structures documentation which is super informal. Given that https://github.com/php/php-langspec is abandoned, it would be good to have a formal grammar specification in the manual. Apparently, [DocBook 4.5 supports EBNF](https://tdg.docbook.org/tdg/4.5/productionset.html), so one may dream. :)

PR #470, opened 2021-03-03
Fix ternary precedence

mvorisek:
it is right associative since PHP 7.4, see https://wiki.php.net/rfc/ternary_associativity

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -223,7 +223,7 @@
         </entry>
        </row>
        <row>
-        <entry>left</entry>
+        <entry>right</entry>
         <entry><literal>? :</literal></entry>
         <entry>
          <link linkend="language.operators.comparison.ternary">ternary</link>


cmb69:
The ternary operator is *non*-associative as of PHP 7.4.0. Anyhow, we also need to document the behavior of previous PHP 7 versions; not quite sure how to do it in this case.

cmb69:
Correction: the ternary operator is still left-associative in PHP 7.4, but that usage is deprecated. It is non-associative as of PHP 8.0.0.

Usually, we add a changelog column to such tables, but that would break the layout. Maybe having a changelog section at the bottom of the operator precendence page is a better solution?

And the [actual description of the ternary operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary) needs to be fixed in this regard. The second note there seems to be misleading.

mvorisek:
Please either close or update this PR.

PR #472, opened 2021-03-04
[67034] Document warning when trying to compare DatePeriod objects

deleugpn:
Add changelog note about DatePeriod comparison as reported in https://bugs.php.net/bug.php?id=67034

### DIFF
--- a/reference/datetime/dateperiod.xml
+++ b/reference/datetime/dateperiod.xml
@@ -181,6 +181,12 @@
       </thead>
       <tbody>
        <row>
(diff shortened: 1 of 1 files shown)


cmb69 (changes requested):
Thanks for the PR, but that bug ticket is about Date*Interval* objects, not Date*Period* objects. Apparently, the latter can be compared: https://3v4l.org/HHjtJ

deleugpn:
My bad, thanks for catching that! I'll reopen as soon as I fix that.

PR #473, opened 2021-03-04
[80766] Add changelog about defining custom assert

deleugpn:
Added changelog notes to `assert` about it no longer being possible to declare a custom function.

### DIFF
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -225,7 +225,7 @@
      </thead>
      <tbody>
       <row>
-       <entry>8.0</entry>
+       <entry>8.0.0</entry>
        <entry>Defining a custom <function>assert</function> function will now raise a fatal error.</entry>
       </row>
       <row>
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -225,6 +225,14 @@
      </thead>
      <tbody>
       <row>
+       <entry>8.0</entry>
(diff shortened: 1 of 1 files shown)


reference/info/functions/assert.xml
```diff
@@ -224,6 +224,14 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0</entry>
```
Girgias, line 228:
```suggestion
       <entry>8.0.0</entry>
```

Girgias (changes requested):

afilina:
I'm not sure what you mean by "Defining a custom assert function". If you mean using a custom handler, then there is no deprecation warning in 7.3.0: https://3v4l.org/a3QB8

There is a deprecation warning for calling `assert` with a string argument though, maybe that's what you meant in the second changelog? https://3v4l.org/maNfU

deleugpn:
By "defining a custom assert function" I mean defining your own assert function.

https://3v4l.org/vlNfn

cmb69:
Defining a custom assert() function in the global namespace never worked, see https://3v4l.org/vlNfn (normal PHP behavior, not to allow redefinition functions). The deprecation/prohibition is actually about custom assert() functions in namespaces, see https://3v4l.org/XTtAh. Could you please clarify that in the changelog entries?

cmb69 (changes requested):
Please address my comment above. :)

cmb69:
I'm closing this in favor of #526.

PR #474, opened 2021-03-04
[80772] Update documentation related to PHP Tokens

deleugpn:
This is an attempt to improve some wording as reported by https://bugs.php.net/bug.php?id=80772


cmb69 (approved):
Thank you! A nice improvement.

Note that I'm going to fix the indentation of the last paragraphp (the one with the example). I'll also fix the wording (the manual uses "prior to PHP x.y.z" and "as of PHP x.y.z", respectively).

PR #476, opened 2021-03-04
Note that all visibilities of trait abstract methods are supported.

Crell:
The docs didn't specify before, so I just clarified that it does now work on all of them.


cmb69:
The word "now" appears to be important here. Please clarify that `private` is supported only as of PHP 8.0.0.

Crell:
Done.

cmb69 (approved):
Thank you!

PR #477, opened 2021-03-04
[80538] Shorthand syntax for array destruction

deleugpn:
This pull request updates the return value documentation for functions to reflect newer syntax provided by PHP as reported in https://bugs.php.net/bug.php?id=80538

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -744,7 +744,7 @@ function small_numbers()
 // Array destruction will collect each member of the array individually
 [$zero, $one, $two] = small_numbers();
 
-// Equivalent to array destruction using list() construct
+// Prior to 7.1.0, the only equivalent alternative is using list() construct
 list($zero, $one, $two) = small_numbers();
 
 ?>
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -741,7 +741,7 @@ function small_numbers()
 {
     return [0, 1, 2];
 }
-// Array destruction will collect each member of the array individually
+// Array destructuring will collect each member of the array individually
 [$zero, $one, $two] = small_numbers();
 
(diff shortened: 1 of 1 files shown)


language/functions.xml
```diff
@@ -739,9 +739,13 @@ echo square(4);   // outputs '16'.
 <?php
 function small_numbers()
 {
-    return array (0, 1, 2);
+    return [0, 1, 2];
 }
-list ($zero, $one, $two) = small_numbers();
+[$zero, $one, $two] = small_numbers();
+
+// For PHP 7.0 or lower
```
Girgias, line 746: This comment doesn't make much sense, we don't document PHP 5 any more, so something more along the way of this being a "functional" (by using a function-style construct) way is probably better.

cmb69: Array destructoring is only available as of PHP 7.1.0, though, so it makes sense to mention that here. :)

deleugpn: That was precisely my intention - to mention an alternative for versions that don't have Array destruction yet.

cmb69: The usual wording in the manual is `prior to PHP 7.1.0` (or `as of PHP 7.1.0`), though.

language/functions.xml
```diff
@@ -739,9 +739,14 @@ echo square(4);   // outputs '16'.
 <?php
 function small_numbers()
 {
-    return array (0, 1, 2);
+    return [0, 1, 2];
 }
-list ($zero, $one, $two) = small_numbers();
+// Array destruction will collect each member of the array individually
```
kamil-tekiela, line 744:
```suggestion
// Array destructuring will collect each member of the array individually
```

kamil-tekiela (changes requested):

kamil-tekiela (approved):
LGTM

PR #478, opened 2021-03-04
Stringable

Crell:
I think that covers it, but please make sure I didn't miss a step...

### DIFF
--- a/language/predefined/stringable.xml
+++ b/language/predefined/stringable.xml
@@ -31,7 +31,7 @@
 
 <!-- {{{ Synopsis -->
    <classsynopsis>
-    <ooclass><interfacename>Stringable</interfacename></ooclass>
+    <ooclass><classname>Stringable</classname></ooclass>
 
 <!-- {{{ Class synopsis -->
     <classsynopsisinfo>
--- a/language/predefined/stringable.xml
+++ b/language/predefined/stringable.xml
@@ -42,14 +42,12 @@
 <!-- }}} -->
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
-<!--
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[not(@role='procedural')])">
-     <xi:fallback />
-    </xi:include>
--->
+<!-- Bring back if we add the toString method as its own page later. For now,
+     just link to the existing description in magic.xml.
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[not(@role='procedural')])">
      <xi:fallback />
     </xi:include>
+-->
    </classsynopsis>
 <!-- }}} -->
 
@@ -104,8 +102,6 @@ showStuff($ip);
   </section>
 
  </partintro>
-
- &language.predefined.stringable.tostring;
 
 </phpdoc:classref>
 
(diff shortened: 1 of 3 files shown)


language/predefined/stringable.xml
```diff
@@ -0,0 +1,122 @@
...
+ <title>The Stringable interface</title>
+ <titleabbrev>Stringable</titleabbrev>
+
+ <partintro>
+
+<!-- {{{ Stringable intro -->
+  <section xml:id="stringable.intro">
+   &reftitle.intro;
+   <para>
+    The <classname>Stringable</classname> interface denotes an object as
```
cmb69, line 15: nit: 
```suggestion
    The <classname>Stringable</classname> interface denotes a class as
```

language/predefined/stringable.xml
```diff
@@ -0,0 +1,122 @@
...
+    </classsynopsisinfo>
+<!-- }}} -->
+    
+    <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[not(@role='procedural')])">
+     <xi:fallback />
+    </xi:include>
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[not(@role='procedural')])">
+     <xi:fallback />
+    </xi:include>
```
cmb69, line 50: I think this will yield warnings (that's a known issue), since there is no constructor-/methodsysnopsis, so better comment these lines out.

language/predefined/stringable.xml
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public function __toString(): string {
+        return "$oct1.$oct2.$oct3.$oct4";
+    }
+}
+
+$ip = new IPv4Address('123', '234', '42', '9');
+print $ip;
+?>
```
cmb69, line 84: Maybe, just maybe, extend the example to show the use of `string|Stringable`?

cmb69 (commented):
Thank you! I think it is okay, but consider my comments.

cmb69 (approved):

language/predefined/stringable.xml
```diff
@@ -0,0 +1,131 @@
...
+123.234.42.9
+]]>
+     </screen>
+    </example>
+   </para>
+  </section>
+
+ </partintro>
+
+ &language.predefined.stringable.tostring;
```
cmb69, line 108: Did you forget to push this file?

cmb69 (changes requested):
Besides the missing stringable/tostring.xml, there are more issues:

- commenting out the xi:includes accidentially commented out the `</classsynopsis>` tag
- the stringable page needs to be added to languages/predefined/interfaces.xml (and then the methodsynopses should be included again (i.e. only comment out the contructorsysnopses)
- there should be an entry in languages/predefined/versions.xml

Could you please take a look, @Crell?

cmb69 (approved):

jmikola:
@cmb69: I just came across https://www.php.net/stringable and looking for a reason that the `__toString()` method was not documented as an interface method led me back to this PR. Was that intentional?

I noted you referred to "commenting out the xi:includes" in your previous comment but it still wasn't clear to me what you meant by that, as I didn't find discussion about commenting that out in any previously resolved  comments.

cmb69:
> […] and looking for a reason that the __toString() method was not documented as an interface method […]

That documentation needs yet to be written (the [PR has a respective comment](https://github.com/php/doc-en/pull/478/files#diff-aca9a13102a64cc8a3a838403b3537cfe717134b6bd449416f396b4ed85e7790R45)); until this file is in place, the `<xi:include>` should be commented out, because otherwise the build system shows an error message (this should be fixed at some point, if possible). The linking to the magic methods doesn't work (I've overlooked that when approving), because `<function>__toString</function>` is not recognized by the build system; that should have been `<link linkend="object.tostring">__toString()</link>` instead.

Anyhow, a PR would be welcome!

PR #480, opened 2021-03-04
[80502] Document 0 < (empty string)

deleugpn:
As mentioned in https://bugs.php.net/bug.php?id=80502 the comparison between `0 < ''` has changed in PHP 8.0.0 due to saner string comparison RFC.

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -1486,7 +1486,7 @@ Expression: 0 = -4 << 62
     <para>
      Prior to PHP 8.0.0, if a <type>string</type> is compared to a number
      or a numeric string then the <type>string</type> was converted to a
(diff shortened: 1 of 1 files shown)


cmb69 (changes requested):
Thanks for the PR! I think having a single example, and highlighting the differences there would be more beneficial than two examples, though.

deleugpn:
I'll try to make a table with `Expression | Result (< 7.4) | Result (> 8.0)`

afilina:
@deleugpn Pinging author.

kamil-tekiela:
This has already been merged. If you would like to improve it, please open a fresh PR.

PR #482, opened 2021-03-06
Update attributes.xml

fullbl:
SimpleXMLElement::attributes return type doesn't match description


cmb69:
Thank you!

The actual patch is somehow messed up; I'm going to fix this right away.

PR #483, opened 2021-03-08
Fix infinite loop possibility in fwrite example

ecartz:
One of the proposed changes from bug #51006 is incorrect.  Because some error conditions cause fwrite to return 0 (bytes written), this can create a condition where on each iteration, it fails to write without updating `$written`.  In those situations, the code as written loops forever.  

Switching back to the loose check causes it to return when that happens, breaking the loop.  

https://bugs.php.net/bug.php?id=51006
https://www.php.net/manual/en/function.fwrite.php#96951

If this is the wrong protocol for reporting/fixing this, please advise on how to go about proposing this change.

### DIFF
--- a/reference/filesystem/functions/fwrite.xml
+++ b/reference/filesystem/functions/fwrite.xml
@@ -65,7 +65,8 @@
   &reftitle.returnvalues;
   <simpara>
    <function>fwrite</function> returns the number of bytes
-   written, or &false; on error.
+   written, or &false; on some errors.  Note that some error conditions
+   may cause 0 to be returned.
   </simpara>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


cmb69:
That looks about right, but maybe https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.fread-fwrite should be incorporated right away?

nikic:
Right, I don't think this is correct as of PHP 7.4, or at least misleading. Some error conditions *can* return 0, namely EWOULDBLOCK and EINTR, but the former only occurs for non-blocking streams, and the latter is exactly the case where you do want to retry.

reference/filesystem/functions/fwrite.xml
```diff
@@ -65,7 +65,8 @@
   &reftitle.returnvalues;
   <simpara>
```
TysonAndre, line 66: Aside: Above, can the `magic_quotes_runtime` paragraph be removed since the documentation is being removed for php 5? https://www.php.net/manual/en/function.fwrite.php#refsect1-function.fwrite-parameters

I still see it in quite a few places

```
reference/zlib/functions/gzwrite.xml
55:        linkend="ini.magic-quotes-runtime">magic_quotes_runtime</link>
```

cmb69: Yes, info about magic_quotes can be removed.

PR #484, opened 2021-03-08
Expand XSL Documentation Example

fwc:
The current XSL documentation has two examples, the [first](https://www.php.net/manual/en/xsl.examples-collection.php) being very brief. It currently only contains an XML and an XSLT file.

This PR expands the example by adding the code posted as a _User Contributed Note_ on the main [Examples](https://www.php.net/manual/en/xsl.examples.php) page and also adds the HTML output.


cmb69 (changes requested):
This example is already shown on https://www.php.net/manual/en/xsltprocessor.transformtoxml.php.

PR #485, opened 2021-03-10
Add union type to streamWrapper::stream_stat

killertux:
Add a union here to keep the same signature of `stat()` and to document a valid behavior.


kalecser (approved):

cmb69:
Thank you!

PR #486, opened 2021-03-11
Update Swoole ini

sy-records:


reference/swoole/ini.xml
```diff
@@ -47,6 +47,30 @@
       <entry>PHP_INI_SYSTEM</entry>
       <entry><!-- leave empty, this will be filled by an automatic script --></entry>
      </row>
+     <row>
+      <entry><link linkend="ini.swoole.enable-coroutine">swoole.enable_coroutine</link></entry>
+      <entry>On</entry>
+      <entry>PHP_INI_ALL</entry>
+      <entry><!-- leave empty, this will be filled by an automatic script --></entry>
```
cmb69, line 54: Unfortunately, that's not true. There is no automatic script which fills the changelog column. Could you please manually add availability information? I mean something like: `Available as of swoole x.y.z` (of course, replacing x.y.z with the version number where the INI setting has been added).

cmb69 (changes requested):
Thank you very much! Please address my comment. :)

reference/swoole/ini.xml
sy-records: Sure, I'll add it later.

sy-records: done.

cmb69 (approved):
Thank you!

PR #487, opened 2021-03-11
Add example of a float string as an array key

Girgias:
The behaviour is different from a float key

Don't know if I should reorder the key values to move it next to the key `"01"`.

(Also should we fix the mixing of double quotes and single quote string keys?)


cmb69:
Ah, yeah, good catch!

IIRC, this example has recently been added or changed. If so, I think it's fine to make the suggested cosmetic changes; otherwise I'd leave it as is for the sake of translators.

Girgias:
As we added this example recently in 529359031bc69c5f635a459ae3891df4b8dc8df9 I'll tweak the order then

PR #488, opened 2021-03-12
Added Static Properties example for Traits

AlexandrErohin:
There are examples for Static Methods and Static Variables, so it is better to add Static Properties example also


saundefined (approved):

saundefined:
@AlexandrErohin, Thank you!

PR #489, opened 2021-03-15
Proposal - more concrete explanation of the fail for __set_state

AlexandrErohin:
It is proposal to explain more concrete the type of the fail after re-importing objects from var_export.
That it is not only would not be re-imported - but also causes fatal error

### DIFF
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -386,7 +386,7 @@ object(A)#2 (2) {
     <simpara>
      When exporting an object, <function>var_export</function> does not check
      whether <link linkend="object.set-state">__set_state()</link> is
-     implemented by the object's class, so re-importing such objects will result in a Fatal Error,
+     implemented by the object's class, so re-importing such objects will result in an <classname>Error</classname> exception,
      as __set_state() is undefined. Particularly, this affects some
      internal classes.
     </simpara>
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -386,8 +386,8 @@ object(A)#2 (2) {
     <simpara>
      When exporting an object, <function>var_export</function> does not check
      whether <link linkend="object.set-state">__set_state()</link> is
-     implemented by the object's class, so re-importing such objects will result in an <classname>Error</classname> exception,
-     as __set_state() is undefined. Particularly, this affects some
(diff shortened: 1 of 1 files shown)


language/oop5/magic.xml
```diff
@@ -386,8 +386,8 @@ object(A)#2 (2) {
     <simpara>
      When exporting an object, <function>var_export</function> does not check
      whether <link linkend="object.set-state">__set_state()</link> is
-     implemented by the object's class, so re-importing such objects will fail,
-     if __set_state() is not implemented. Particularly, this affects some
+     implemented by the object's class, so re-importing such objects will result in a Fatal Error,
```
afilina, line 389: Is this true for all PHP versions, or does this have a different result based on the version?

AlexandrErohin:
@afilina 
Yes - it results with Fatal error since version 5.1.0
https://3v4l.org/C6SrZ

cmb69:
Actually, that is an `Error` exception as of PHP 7.0.0 (older versions are no longer relevant for the manual).

language/oop5/magic.xml
afilina, line 389:
```suggestion
     implemented by the object's class, so re-importing such objects will result in an <classname>Error</classname> exception,
```

afilina (changes requested):

language/oop5/magic.xml
AlexandrErohin: @afilina Thank you!

afilina (approved):

language/oop5/magic.xml
```diff
@@ -386,8 +386,8 @@ object(A)#2 (2) {
     <simpara>
      When exporting an object, <function>var_export</function> does not check
      whether <link linkend="object.set-state">__set_state()</link> is
-     implemented by the object's class, so re-importing such objects will fail,
-     if __set_state() is not implemented. Particularly, this affects some
+     implemented by the object's class, so re-importing such objects will result in an <classname>Error</classname> exception,
+     as __set_state() is undefined. Particularly, this affects some
```
cmb69, line 390:
```suggestion
     implemented by the object's class, so re-importing objects will result in an <classname>Error</classname> exception,
     if __set_state() is not implemented. Particularly, this affects some
```

cmb69 (approved):
Thank you!

PR #491, opened 2021-03-16
Document hard timeout

NattyNarwhal:
This has affected me and others before, but it's sparsely documented (as well as its effects).


cmb69:
Ah, very nice one as well. Thank you!

PR #492, opened 2021-03-18
Correct Closure::bindTo failure return value

Firehed:
`Closure::bindTo()` appears to return `null` on failure, not the documented `false`. This corrects the documentation to match the exhibited behavior.

Context: https://github.com/phpstan/phpstan/issues/4716 and repro script: https://3v4l.org/no5qG

Internals: https://github.com/php/php-src/blob/dcac654fd56a2cecec0c187ba061ecc82bbc9031/Zend/zend_closures.c#L219-L221

Also, reflection shows `?Closure`:

```
$ php --rc Closure
Class [ <internal:Core> final class Closure ] {

  ...

  - Methods [3] {
  ...
    Method [ <internal:Core> public method bindTo ] {

      - Parameters [2] {
        Parameter #0 [ <required> ?object $newThis ]
        Parameter #1 [ <optional> object|string|null $newScope = "static" ]
      }
      - Return [ ?Closure ]
    }
  ...
  }
}
```

### DIFF
--- a/language/predefined/closure/bindto.xml
+++ b/language/predefined/closure/bindto.xml
@@ -87,7 +87,7 @@
   &reftitle.returnvalues;
   <para>
    Returns the newly created <classname>Closure</classname> object
-   &return.nullforfailure;
+   or &.null on failure.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Yes, apparently it returns NULL on failure at least as of PHP 7.0.0. Thank you!

PR #493, opened 2021-03-18
Match fgetcsv parameter names to values in reflection

Firehed:
The documented parameters differ from the names exposed in reflection/source code. Now that named parameters are supported in PHP 8, this discrepancy is confusing and impactful.

```
 $ php --rf fgetcsv
Function [ <internal:standard> function fgetcsv ] {

  - Parameters [5] {
    Parameter #0 [ <required> $stream ]
    Parameter #1 [ <optional> ?int $length = null ]
    Parameter #2 [ <optional> string $separator = "," ]
    Parameter #3 [ <optional> string $enclosure = "\"" ]
    Parameter #4 [ <optional> string $escape = "\\" ]
  }
  - Return [ array|false ]
}
```

### DIFF
--- a/reference/filesystem/functions/fgetcsv.xml
+++ b/reference/filesystem/functions/fgetcsv.xml
@@ -10,9 +10,9 @@
   &reftitle.description;
   <methodsynopsis>
    <type>array</type><methodname>fgetcsv</methodname>
-   <methodparam><type>resource</type><parameter>handle</parameter></methodparam>
+   <methodparam><type>resource</type><parameter>stream</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you; the patch is fine! However, there is already [work in progress](https://github.com/php/doc-en/pulls?q=is%3Apr+is%3Aopen+label%3Amethodsynopses) to semi-automatically update the documentation from the *.stub.php files. I think it's generally preferable to do it this way, since there are soo many changes.

Firehed:
That's great to hear - we'll add benefit from that kind of automation :)

PR #494, opened 2021-03-19
Generate ext/pdo methodsynopses based on stubs

kocsismate:


reference/pdo/pdostatement/fetchall.xml
```diff
@@ -10,10 +10,9 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
```
kocsismate, line 12: As far as I saw this is the only signature change which is worth a mention via a changelog. I added one, although I couldn't come up with a description good enough for the variadic `$args` parameter in the parameter description section.

cmb69: I think it's okay.

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -82,7 +81,7 @@
      <term><parameter>ctor_args</parameter></term>
```
cmb69, line 81: I think this should be
```suggestion
     <term><parameter>args</parameter></term>
```

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -110,6 +109,30 @@
...
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <parameter>Args</parameter> is now a variadic parameter.
```
cmb69, line 127:
```suggestion
        <parameter>args</parameter> is now a variadic parameter.
```

reference/pdo/pdostatement/bindcolumn.xml
```diff
@@ -10,13 +10,13 @@
...
-    <methodparam><type>mixed</type><parameter>column</parameter></methodparam>
-    <methodparam><type>mixed</type><parameter role="reference">param</parameter></methodparam>
-    <methodparam choice="opt"><type>int</type><parameter>type</parameter></methodparam>
-    <methodparam choice="opt"><type>int</type><parameter>maxlen</parameter></methodparam>
-    <methodparam choice="opt"><type>mixed</type><parameter>driverdata</parameter></methodparam>
-   </methodsynopsis>
+   <modifier>public</modifier> <type>bool</type><methodname>PDOStatement::bindColumn</methodname>
+   <methodparam><type class="union"><type>string</type><type>int</type></type><parameter>column</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter role="reference">var</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>type</parameter><initializer>0</initializer></methodparam>
```
kamil-tekiela, line 16: Default should be PDO_PARAM_STR not 0

kocsismate: Nice catch! Thanks! I've fixed it in this PR as well as in php-src itself.

reference/pdo/pdo/query.xml
```diff
@@ -10,30 +10,31 @@
...
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>fetchMode</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam rep="repeat"><type>mixed</type><parameter>fetchModeArgs</parameter></methodparam>
   </methodsynopsis>
 
   <methodsynopsis>
-   <modifier>public</modifier> <type>PDOStatement</type><methodname>PDO::query</methodname>
-   <methodparam><type>string</type><parameter>statement</parameter></methodparam>
-   <methodparam><type>int</type><parameter>fetch_style</parameter><initializer>PDO::FETCH_COLUMN</initializer></methodparam>
-   <methodparam><type>int</type><parameter>colno</parameter></methodparam>
+   <modifier>public</modifier> <type class="union"><type>PDOStatement</type><type>false</type></type><methodname>PDO::query</methodname>
```
kamil-tekiela, line 20: I think you should manually revert the changes here. The first signature should have only 2 arguments. The other 3 should stay as they were, but you can add nullability to the second argument. The point is that there really are 4 signatures for this method. We can't express this in stub files, but we should put it in the manual.

reference/pdo/pdo/setattribute.xml
```diff
@@ -12,7 +12,7 @@
   <methodsynopsis>
    <modifier>public</modifier> <type>bool</type><methodname>PDO::setAttribute</methodname>
    <methodparam><type>int</type><parameter>attribute</parameter></methodparam>
-   <methodparam><type>mixed</type><parameter>value</parameter></methodparam>
+   <methodparam><type class="union"><type>array</type><type>int</type></type><parameter>value</parameter></methodparam>
```
kamil-tekiela, line 15: I think that's a mistake in the stub file. There should be 3: int, array, bool.

reference/pdo/pdostatement/bindcolumn.xml
```diff
@@ -10,13 +10,13 @@
...
-    <methodparam choice="opt"><type>int</type><parameter>type</parameter></methodparam>
-    <methodparam choice="opt"><type>int</type><parameter>maxlen</parameter></methodparam>
-    <methodparam choice="opt"><type>mixed</type><parameter>driverdata</parameter></methodparam>
-   </methodsynopsis>
+   <modifier>public</modifier> <type>bool</type><methodname>PDOStatement::bindColumn</methodname>
+   <methodparam><type class="union"><type>string</type><type>int</type></type><parameter>column</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter role="reference">var</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>type</parameter><initializer>PDO::PARAM_STR</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>maxLength</parameter><initializer>0</initializer></methodparam>
+   <methodparam choice="opt"><type>mixed</type><parameter>driverOptions</parameter><initializer>&null;</initializer></methodparam>
```
kamil-tekiela, line 18: I believe this one should be nullable.

reference/pdo/pdostatement/bindparam.xml
```diff
@@ -11,11 +11,11 @@
...
-   <methodparam><type>mixed</type><parameter>parameter</parameter></methodparam>
-   <methodparam><type>mixed</type><parameter role="reference">variable</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>data_type</parameter><initializer>PDO::PARAM_STR</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>length</parameter></methodparam>
-   <methodparam choice="opt"><type>mixed</type><parameter>driver_options</parameter></methodparam>
+   <methodparam><type class="union"><type>string</type><type>int</type></type><parameter>param</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter role="reference">var</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>type</parameter><initializer>PDO::PARAM_STR</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>maxLength</parameter><initializer>0</initializer></methodparam>
+   <methodparam choice="opt"><type>mixed</type><parameter>driverOptions</parameter><initializer>&null;</initializer></methodparam>
```
kamil-tekiela, line 18: Same here, nullable

reference/pdo/pdostatement/debugdumpparams.xml
```diff
@@ -10,8 +10,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>void</type><methodname>PDOStatement::debugDumpParams</methodname>
-   <void />
+   <modifier>public</modifier> <type class="union"><type>bool</type><type>null</type></type><methodname>PDOStatement::debugDumpParams</methodname>
```
kamil-tekiela, line 13: This is not documented yet. According to the manual this method is void

reference/pdo/pdostatement/errorcode.xml
```diff
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>string</type><methodname>PDOStatement::errorCode</methodname>
+   <modifier>public</modifier> <type class="union"><type>string</type><type>null</type></type><methodname>PDOStatement::errorCode</methodname>
```
kamil-tekiela, line 13: I think this might be a bug. How can this function return null?

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,14 +11,14 @@
   &reftitle.description;
   <methodsynopsis>
    <modifier>public</modifier> <type>mixed</type><methodname>PDOStatement::fetch</methodname>
-   <methodparam choice="opt"><type>int</type><parameter>fetch_style</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>cursor_orientation</parameter><initializer>PDO::FETCH_ORI_NEXT</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>cursor_offset</parameter><initializer>0</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer>PDO::FETCH_BOTH</initializer></methodparam>
```
kamil-tekiela, line 14: Incorrect. The default is NULL. Bug in stubs?

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -10,10 +10,9 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type class="union"><type>array</type><type>false</type></type><methodname>PDOStatement::fetchAll</methodname>
-   <methodparam choice="opt"><type>int</type><parameter>fetch_style</parameter></methodparam>
-   <methodparam choice="opt"><type>mixed</type><parameter>fetch_argument</parameter></methodparam>
-   <methodparam choice="opt"><type>array</type><parameter>ctor_args</parameter><initializer>array()</initializer></methodparam>
+   <modifier>public</modifier> <type>array</type><methodname>PDOStatement::fetchAll</methodname>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer>PDO::FETCH_BOTH</initializer></methodparam>
+   <methodparam rep="repeat"><type>mixed</type><parameter>args</parameter></methodparam>
```
kamil-tekiela, line 15: Where did this change come from? What happened to the third argument? I can't see any change in the code. The old signature is correct.

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -110,6 +109,30 @@
...
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <parameter>args</parameter> is now a variadic parameter.
```
kamil-tekiela, line 127: It's really not. 
> ArgumentCountError: PDOStatement::fetchAll() expects at most 3 arguments, 5 given

reference/pdo/pdostatement/setfetchmode.xml
```diff
@@ -12,25 +12,25 @@
   <methodsynopsis>
    <modifier>public</modifier> <type>bool</type><methodname>PDOStatement::setFetchMode</methodname>
    <methodparam><type>int</type><parameter>mode</parameter></methodparam>
+   <methodparam rep="repeat"><type>mixed</type><parameter>args</parameter></methodparam>
```
kamil-tekiela, line 15: Same as my first comment. This needs to be manually adjusted to what it was before

reference/pdo/pdo/setattribute.xml
kocsismate: I believe `mixed` is correct, since any type is accepted, and there is no notice or warning when providing a different type than int/array/bool.

reference/pdo/pdostatement/bindcolumn.xml
kocsismate: It is `mixed`, so by definition nullable :)

reference/pdo/pdo/setattribute.xml
kamil-tekiela: `mixed` is also fine. Either `mixed` or 3-way union.

reference/pdo/pdostatement/errorcode.xml
kocsismate: I'm not sure what this exactly does:
```
if (stmt->error_code[0] == '\0') {
    RETURN_NULL();
}
```

but it was added in https://github.com/php/php-src/commit/9334bbd650c65f34c2f7cb5e96e12ba0518f119b#diff-cedda5e61bbd718fe650fa21c5c402d39868708b19ad317a8e3c68b001128357R1781

reference/pdo/pdostatement/fetch.xml
kocsismate: are you sure? As far as I see it's defined as `zend_long how = PDO_FETCH_USE_DEFAULT;` Or which parameter do you mean?

reference/pdo/pdostatement/fetchall.xml
kocsismate: Have you updated your PHP-8.0 branch? It looks like yours is out of sync. Nevertheless it came from https://github.com/php/php-src/commit/cfaa270da5b511e2821aa0a5e036ef5879a56bed

reference/pdo/pdostatement/fetch.xml
kamil-tekiela: If you provide `PDO::FETCH_BOTH` you overwrite the default. To have the default you need to pass null.

reference/pdo/pdostatement/fetchall.xml
kamil-tekiela: So the problem here is that we can't represent the signature correctly in the stub file. We have 3 overloads for this function, same as with query and setFetchMode. We can't remove this information from the manual, so we have to leave the manual as it was before.

reference/pdo/pdostatement/fetchall.xml
kamil-tekiela: Yeah, let's not add the changelog here. The arguments didn't change. The stub file is misrepresenting it.

reference/pdo/pdostatement/fetchall.xml
cmb69: While `PDO::query()` is documented with overloaded signatures, `PDOStatement::fetchAll()` is not yet, but it should be done.

reference/pdo/pdostatement/errorcode.xml
kamil-tekiela: NVM, it turns out this is the intended behaviour

reference/pdo/pdostatement/fetch.xml
kocsismate: I've just realized that `PDO_FETCH_USE_DEFAULT != PDO::FETCH_BOTH`. But I still don't understand how `null` is related, since the parameter doesn't accept it (officially).

kocsismate: Although... there is this line: `dbh->default_fetch_type = PDO_FETCH_BOTH;` in `pdo_dhb.c`. So no matter if you provide `0` (which is `PDO_FETCH_USE_DEFAULT `) to the `$mode` param of `PDOStatement::fetch()` or `PDO::FETCH_BOTH`, the `how` param ends up being `PDO_FETCH_BOTH` due to this piece of code in `do_fetch()` in `pdo_stmt.c`:

```
if (how == PDO_FETCH_USE_DEFAULT) {
	how = stmt->default_fetch_type;
}
```

Or am I missing something?

kamil-tekiela: If I understand it correctly if you don't provide a param then the default is used which could have been changed in the constructor arguments or using `setFetchMode()` method. It's never guaranteed that the default is `PDO_FETCH_BOTH`.

kocsismate: OK, I see it now! I missed that this default can be changed. Thanks for the explanation!

kocsismate: I've pushed an update. Hopefully it fixes all the review suggestions

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -10,32 +10,37 @@
...
     <varlistentry>
-     <term><parameter>fetch_style</parameter></term>
+     <term><parameter>mode</parameter></term>
      <listitem>
       <para>
        Controls the contents of the returned array as documented in
        <function>PDOStatement::fetch</function>.
        Defaults to value of <constant>PDO::ATTR_DEFAULT_FETCH_MODE</constant>
-       (which defaults to <constant>PDO::FETCH_BOTH</constant>)
+       (which defaults to <constant>PDO::FETCH_DEFAULT</constant>)
```
kamil-tekiela, line 37: 😄 The default defaults to FETCH_BOTH. It was correct as it was. Please don't change this line

reference/pdo/constants.xml
```diff
@@ -118,6 +118,17 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="pdo.constants.fetch-default">
+   <term>
+    <constant>PDO::FETCH_DEFAULT</constant>
+     (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Specifies that the default fetch method shall be used.
```
kamil-tekiela, line 128: Fetch mode is better than fetch method

reference/pdo/pdo/query.xml
```diff
@@ -10,29 +10,30 @@
...
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>fetchMode</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
 
   <methodsynopsis>
-   <modifier>public</modifier> <type>PDOStatement</type><methodname>PDO::query</methodname>
-   <methodparam><type>string</type><parameter>statement</parameter></methodparam>
-   <methodparam><type>int</type><parameter>fetch_style</parameter><initializer>PDO::FETCH_COLUMN</initializer></methodparam>
+   <modifier>public</modifier> <type class="union"><type>PDOStatement</type><type>false</type></type><methodname>PDO::query</methodname>
+   <methodparam><type>string</type><parameter>query</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>fetchMode</parameter><initializer>&null;</initializer></methodparam>
```
kamil-tekiela, line 21: Correct me if I'm wrong but I think this should be exactly the same as setfetchmodea and fetchall. Instead of Null show the overloaded signature.

reference/pdo/pdostatement/fetch.xml
kamil-tekiela: This signature hasn't been updated. It should say PDO::FETCH_DEFAULT instead of PDO::FETCH_BOTH as you have changed it.

reference/pdo/pdostatement/fetchall.xml
kocsismate: Ah, yes.. :D

reference/pdo/pdo/query.xml
kocsismate: No, I also agree!

reference/pdo/pdostatement/fetchobject.xml
```diff
@@ -9,8 +9,8 @@
   &reftitle.description;
   <methodsynopsis>
    <modifier>public</modifier> <type class="union"><type>object</type><type>false</type></type><methodname>PDOStatement::fetchObject</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>class_name</parameter><initializer>"stdClass"</initializer></methodparam>
-   <methodparam choice="opt"><type>array</type><parameter>ctor_args</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>class</parameter><initializer>"stdClass"</initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>array</type><type>null</type></type><parameter>ctorArgs</parameter><initializer>&null;</initializer></methodparam>
```
kamil-tekiela, line 13: The argument is not nullable. 

>Uncaught TypeError: PDOStatement::fetchObject(): Argument #2 ($ctorArgs) must be of type array, null given

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -10,20 +10,25 @@
...
+   <methodparam choice="opt"><type>mixed</type><parameter>args</parameter></methodparam>
   </methodsynopsis>
 
+  <methodsynopsis>
+   <modifier>public</modifier> <type>array</type><methodname>PDOStatement::fetchAll</methodname>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer>PDO::FETCH_DEFAULT</initializer></methodparam>
+   <methodparam choice="opt"><type>mixed</type><parameter>args</parameter></methodparam>
+   <methodparam choice="opt"><type>array</type><parameter>ctor_args</parameter><initializer>[]</initializer></methodparam>
+  </methodsynopsis>
  </refsect1>
```
kamil-tekiela, line 24: I think fetchAll should be documented exactly the same way as setFetchMode with the 4 overloads. The second PDO::FETCH_DEFAULT isn't helpful here

reference/pdo/pdostatement/setfetchmode.xml
```diff
@@ -47,31 +48,31 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>colno</parameter></term>
+     <term><parameter>args</parameter></term>
```
kamil-tekiela, line 51: The name changes need to be reverted, too.

reference/pdo/pdo/setattribute.xml
kamil-tekiela: What's the status on this one? Will you change it back to mixed?

reference/pdo/pdostatement/fetchobject.xml
kocsismate: Fantastic! Another stub issue :/

kocsismate: (But nice catch!)

kocsismate: Proposed: https://github.com/php/doc-en/pull/494

reference/pdo/pdo/query.xml
```diff
@@ -10,29 +10,30 @@
...
   <methodsynopsis>
-   <modifier>public</modifier> <type>PDOStatement</type><methodname>PDO::query</methodname>
-   <methodparam><type>string</type><parameter>statement</parameter></methodparam>
-   <methodparam><type>int</type><parameter>fetch_style</parameter><initializer>PDO::FETCH_CLASS</initializer></methodparam>
+   <modifier>public</modifier> <type class="union"><type>PDOStatement</type><type>false</type></type><methodname>PDO::query</methodname>
+   <methodparam><type>string</type><parameter>query</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>fetchMode</parameter><initializer>PDO::FETCH_CLASS</initializer></methodparam>
    <methodparam><type>string</type><parameter>classname</parameter></methodparam>
-   <methodparam><type>array</type><parameter>ctorargs</parameter></methodparam>
+   <methodparam><type>array</type><parameter>ctor_args</parameter></methodparam>
```
kamil-tekiela, line 30:
```suggestion
   <methodparam><type>array</type><parameter>constructorArgs</parameter></methodparam>
```

reference/pdo/pdostatement/fetchall.xml
kamil-tekiela: Is this still WIP? If not, I'd like to remind that you copied too much. The name of the function and return parameters changed too. Also, the first signature is missing `= PDO::FETCH_DEFAULT`

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -35,7 +51,7 @@
        To return an array consisting of all values of a single column from
        the result set, specify <constant>PDO::FETCH_COLUMN</constant>. You
        can specify which column you want with the
-       <parameter>fetch_argument</parameter> parameter.
+       <parameter>args</parameter> parameter.
```
kamil-tekiela, line 54:
```suggestion
       <parameter>colno</parameter> parameter.
```
We removed the args variadic signature now

reference/pdo/pdostatement/setfetchmode.xml
```diff
@@ -55,23 +56,23 @@
...
+     <term><parameter>constructorArgs</parameter></term>
      <listitem>
       <para>
        Constructor arguments.
       </para>
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>object</parameter></term>
+     <term><parameter>args</parameter></term>
```
kamil-tekiela, line 75:
```suggestion
     <term><parameter>object</parameter></term>
```
revert back

kamil-tekiela (changes requested):

reference/pdo/pdo/query.xml
kocsismate: Thanks for all these findings. I came to the conclusion that I really hate these overloaded functions/methods :D So I'll definitely propose an RFC next time to deprecate them whenever possible (an example: https://github.com/php/php-src/pull/6754).

reference/pdo/pdostatement/fetchall.xml
kocsismate: Gosh, no, it wasn't supposed to be WIP.. I'm just inattentive :/

kocsismate: I really hope that I managed to got it right this time

kamil-tekiela (approved):
This looks ok to me now. I still have a feeling we missed something. @cmb69  What's your opinion?

cmb69:
From a quick look, this looks okay. Of course, there is the issue that some of these methods may not return false on failure for ERRMODE_EXCEPTION, but I don't think it's worth to explicitly document that (IMHO, letting clients choose how issues are reported is a bad idea in the first place).

kamil-tekiela:
@cmb69 Yeah, I noticed that too, but I think we can make a separate PR for that. The return values will have to be adjusted for the new signatures.

cmb69 (approved):

reference/pdo/constants.xml
```diff
@@ -118,6 +118,17 @@
...
+   <term>
+    <constant>PDO::FETCH_DEFAULT</constant>
+     (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Specifies that the default fetch mode shall be used.
+    </simpara>
+   </listitem>
+  </varlistentry>
```
jonaslinn, line 131: Hello, just stumbled over this constant. As nice as it would be, it does not exists (8.0.6). Or is this planned?
`Fatal error: Uncaught Error: Undefined constant PDO::FETCH_DEFAULT`

kocsismate: We documented it a bit prematurely... since the constant has just been added by https://github.com/php/php-src/pull/6937, it's going to be available in the next patch release

cmb69: Please add that info ("Available as of PHP 8.0.7.")

kocsismate: Submitted: https://github.com/php/doc-en/pull/597 Although I'm wondering if this information should be added to the usages of the constant?

PR #495, opened 2021-03-20
Format

rochdiNassah:


cmb69:
I don't think that we handle capitalization in code comments consistently, and this doesn't look like an improvement. Thank you for the PR, nonetheless.

PR #496, opened 2021-03-20
Updating mysqli select db

kamil-tekiela:
Simplify code examples on https://www.php.net/manual/en/mysqli.select-db.php


cmb69 (approved):
Nice. Thank you!

PR #498, opened 2021-03-20
Updating mysqli: real_escape_string

kamil-tekiela:
Tidied up the page:
- Removed alias name from this page and moved it to where it belongs. 
- Removed old note about returning NULL. This function did do that up to PHP 7.2 but I am not able to find reasons why. Either way we document only the current behaviour when it comes to return parameters. 
- Reverted [Doc Bug #55757](https://bugs.php.net/bug.php?id=55757). This is not a place to keep such notes. Migration from old mysql_* API to mysqli is a much broader topic that should not be described on some arbitrary function's manual page. 
- Provided a new code example. The one we currently have isn't following best practices. This leads to many people believing some common fallacies e.g. escaping is only part of data sanitization and should be done at the beginning of the script, or that escaping should be only performed during data insertion. The new example shows how to property escape and format the query in one go. 
One thing I couldn't figure out is what to do with the overly long output.

### DIFF
--- a/reference/mysqli/functions/mysqli-escape-string.xml
+++ b/reference/mysqli/functions/mysqli-escape-string.xml
@@ -2,6 +2,7 @@
 <!-- $Revision$ -->
 <refentry xml:id="function.mysqli-escape-string" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
+  <refname>mysqli::escape_string</refname>
   <refname>mysqli_escape_string</refname>
   <refpurpose>&Alias; <function>mysqli_real_escape_string</function></refpurpose>
  </refnamediv>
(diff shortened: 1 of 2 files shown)


cmb69:
Hmm, these alias pages likely have been introduced to work around the limited search capabilities (see https://bugs.php.net/bug.php?id=49956). Not sure what to do about that.

kamil-tekiela:
@cmb69 These pages have been in the documentation since the very beginning. I don't really know why the manual was structured this way. We might be able to clean it up sometime in the future but for the moment it is what it is. I doubt this has anything to do with the bug you linked. 
My idea here is that we should document all 4 variants on a single page or keep 2 here and move escape_string to its own alias page. It should be consistent either way.

cmb69 (approved):
Sorry, I should have read what you done; instead I only saw "remove alias" and was alerted about the annoying search bug (just recently marked 4 or 5 other tickets as duplicate).

Regarding removing the note about the NULL return: usually, we want to document such changes in the changelog, but in this case the note doesn't really belong there, because that's a general issue of calling internal functions with invalid parameters.

So everything is fine. Thanks for the PR!

PR #499, opened 2021-03-20
Updating mysqli: clean up all aliases removed in PHP 5.4

kamil-tekiela:


GrahamCampbell:
Shouldn't this remain in the documentation, marked as PHP 5.4 and lower only?

cmb69:
The documentation is supposed to be for PHP 7.0.0 and newer. Old documentation [is archived](https://www.php.net/manual/php5.php).

cmb69 (approved):
Thank you!

PR #500, opened 2021-03-21
fix errors in the comments of example number 6

gabrielpl:


cmb69 (approved):
Good catch, thank you!

PR #501, opened 2021-03-22
Fix runkit7 functions arg type

fan-k-morii:
* [runkit7_constant_redefine()](https://www.php.net/manual/en/function.runkit7-constant-redefine.php): $newVisibility shoud be `int`
* [runkit7_function_redefine()](https://www.php.net/manual/en/function.runkit7-function-redefine.php): $is_strict should be `bool`


cmb69 (approved):
Thank you!

PR #502, opened 2021-03-24
Fix typo: shorted -> shorter

vreeken:


cmb69 (approved):
Good catch, thank you!

PR #505, opened 2021-03-29
Add swoole_error_log for swoole

sy-records:


reference/swoole/functions/swoole-error-log.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+
+<refentry xml:id="function.swoole-version" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
```
cmb69, line 4:
```suggestion
<refentry xml:id="function.swoole-error-log" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
```

sy-records: done

cmb69: Thanks!

reference/swoole/functions/swoole-error-log.xml
```diff
@@ -0,0 +1,66 @@
...
+  &reftitle.description;
+  <methodsynopsis>
+   <type>void</type><methodname>swoole_error_log</methodname>
+   <methodparam><type>int</type><parameter>level</parameter></methodparam>
+   <methodparam><type>string</type><parameter>msg</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+
+  </para>
+
```
cmb69, line 20: If it is not possible to document the function for now, please add:
```suggestion
  &warn.undocumented.func;
```

cmb69 (changes requested):
Thank you for the PR! However, it would be good to actually document the function, i.e. to describe the general purpose, the parameters (especially the `$level`; are there constant defined in swoole for that purpose?). Also, there should be return value section (see https://github.com/php/doc-en/pull/507/files#r603979680). If that is not possible, add the `&warn.undocumented.func;` entity reference, like shown below.

Furthermore, there should be an entry for the function in https://github.com/php/doc-en/blob/master/reference/swoole/versions.xml, which states the availability of this function (i.e. in which version of swoole it has been added).

sy-records:
@cmb69 I modified it, can you review it again?

cmb69:
Thanks you, especially for also documenting the constants! I'm going to add availability info to these, and will apply the patch.

cmb69:
Applied as f03dfae1dcaf64d695189673e6db82aa3b7659a1.

PR #506, opened 2021-03-29
is_a() works with interfaces, too

BenMorel:


reference/classobj/functions/is-a.xml
```diff
@@ -3,7 +3,7 @@
 <refentry xml:id="function.is-a" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>is_a</refname>
-  <refpurpose>Checks if the object is of this class or has this class as one of its parents</refpurpose>
+  <refpurpose>Checks if the object is of this class or has this class as one of its parents, or implements this interface</refpurpose>
```
cmb69, line 6: I think the refpurpose is already too long; we should somehow shorten it. The additional mention of interfaces should go into the description section (first paragraph after the methodsynopsis), though.

Girgias, line 6: How about?
```suggestion
  <refpurpose>Checks if the object is an instance of a class/interface</refpurpose>
```

cmb69: I thought about that as well, but the first parameter may be a class/interface *name*.

Girgias:
```suggestion
  <refpurpose>Checks if the object is a subtype of a class/interface</refpurpose>
```
Maybe?

cmb69: I think "subtype" is worse than "instance of", since the former doesn't cover the case where the object is of that type. If in doubt, we should stick with "object" (for now), and change to "instance of". The description section could still clarify.

afilina, line 6: Let me try something completely different to give us new ideas:
```suggestion
  <refpurpose>Compares the object's type to a given class name</refpurpose>
```

afilina: Idea for description:
```
The function will return true if the object is either of that class, a child of that class, or implement that interface.
```

cmb69: Well, `is_a()` is a big mess. You can pass an object or a class/interface name as first argument (the former case should better be served by using `instanceof`), and there is even a third parameter which allows to prevent passing a class/interface name as first argument. How is one supposed to document that in a concise way?

afilina: Assuming the `refpurpose` is about the purpose of the function, not the behavior, then we merely need to state _why_ devs might want to use it. The exact behavior and all its exceptions can be detailed within the relevant parameters.

cmb69: Right, but talking about "the object" is misleading (especially since the function is mostly useful to compare class/interface names). Maybe something very general, like "Checks type relation", is appropriate as refpurpose?

afilina: Some more ideas:
- Checks type correspondence
- Checks type analogy
- Compare types

By the way, `subclass` is a viable term in the context of interfaces as well. See first sentence of LSP: https://en.wikipedia.org/wiki/Liskov_substitution_principle

cmb69: "Subtype" might be preferable, though, but my actual with this term is that T is not a subtype/subclass of T. Isn't there are word which expresses "same type or subtype"?

afilina: I can't think of such a work. Perhaps two words are still short enough: "Checks whether the object is of a given type or subtype"

cmb69: I'm still not happy with "object", but it's overall better than before, so maybe just go with that. :)

afilina, line 6: @cmb69 Like this?
```suggestion
  <refpurpose>Checks whether the object is of a given type or subtype</refpurpose>
```

PR #508, opened 2021-03-30
Added empty string case to is_numeric code example doc

samuelgfeller:


reference/var/functions/is-numeric.xml
```diff
@@ -67,7 +67,8 @@ $tests = array(
     "not numeric",
     array(),
     9.1,
-    null
+    null,
+    ''
```
cmb69, line 71: Thanks for the PR; makes perfect sense to add this. However, please terminate the line with a comma, to avoid unnecessarily complex diffs in the future:
```suggestion
    '',
```

samuelgfeller:
I was about to comment about that I was unsure about changing the `array()` to the alias `[ ]` and if I should have added a comma after the last array entry right when you commented ^^

cmb69 (approved):
Cool, thanks! I think keeping the `array()` is fine here.

PR #509, opened 2021-04-01
Fix/update operators.precedence

guilliamxavier:
- explain why associativity is not applicable to unary operators
- distinguish between unary and binary `+`/`-`
- add some missing reference links
- update the precedence of string concatenation for PHP 8
- fix "(n/a)" ("not applicable") to "non-associative" for ternary

References:
- https://github.com/php/php-src/blob/master/Zend/zend_ast.c
  (search "Operator Precedence")
- https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y
  (search "precedence")
  (NOTE: in opposite order)

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -54,8 +54,13 @@
     Operators of equal precedence that are non-associative cannot be used
     next to each other, for example <literal>1 &lt; 2 &gt; 1</literal> is
     illegal in PHP. The expression <literal>1 &lt;= 1 == 1</literal> on the
-    other hand is legal, because the <literal>==</literal> operator has lesser
+    other hand is legal, because the <literal>==</literal> operator has a lower
     precedence than the <literal>&lt;=</literal> operator.
+   </para>
+   <para>
+    Associativity is only meaningful for binary (and ternary) operators.
+    Unary operators are either prefix or postfix so this notion is not applicable.
+    For example <literal>!!$a</literal> can only be grouped as <literal>!(!$a)</literal>.
    </para>
    <para>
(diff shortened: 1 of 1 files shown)


guilliamxavier:
Notes:
- Actually I had started with just "update the precedence of string concatenation for PHP 8" but that triggered a cascade of related small changes 😅
- About the table layout: [before](https://user-images.githubusercontent.com/7404452/114378514-11b71200-9b88-11eb-93af-5b91991167bc.png) and [after](https://user-images.githubusercontent.com/7404452/114378621-2e534a00-9b88-11eb-8d74-d8b7053b310a.png) the changes 🤔 Maybe we could add a style `white-space: nowrap` to the first two columns ([result](https://user-images.githubusercontent.com/7404452/114378834-63f83300-9b88-11eb-9118-ef86781bab49.png)), or insert some line breaks ([example](https://user-images.githubusercontent.com/7404452/114378900-74101280-9b88-11eb-83ae-71a79593ef39.png))... WDYT?

cmb69:
Thank you for the very good PR! I'm going to merge it.

> About the table layout: before and after the changes 🤔 Maybe we could add a style white-space: nowrap to the first two columns (result), or insert some line breaks (example)... WDYT?

Many tables in the manual have issues (e.g. https://www.php.net/manual/en/filter.filters.validate.php overflows in small browser windows), and often, in my opinion, the best solution would be to replace the tables with lists. In this case I think it might be best to link the operators in the 2nd column directly and add availability info there, and to drop the third column altogether.

PR #510, opened 2021-04-02
Language Ref/Traits: fix incorrect information about abstract methods

jrfnl:
As of PHP 8.0, `abstract private` methods are allowed in `trait`s. `abstract public` method were previously already supported.

Also see: https://3v4l.org/eQY8v and https://3v4l.org/HKLcI


cmb69 (approved):
Good catch, thank you!

jrfnl:
You're welcome ;-)

PR #511, opened 2021-04-02
escapeshellcmd: add warning about spaces in paths on Windows

jrfnl:
It is a known issue that spaces are not escaped in shell commands, which can be especially problematic on Windows.
This adds a warning about this behaviour to the function, including a way to solve this in userland code.

Ref: https://bugs.php.net/bug.php?id=43261 (last two comments)

Also see: https://github.com/squizlabs/PHP_CodeSniffer/pull/3214

### DIFF
--- a/reference/exec/functions/escapeshellcmd.xml
+++ b/reference/exec/functions/escapeshellcmd.xml
@@ -86,6 +86,20 @@ system($escaped_command);
      <function>escapeshellarg</function> should be used instead.
     </para>
    </warning>
+   <warning xmlns="http://docbook.org/ns/docbook">
+    <para>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! However, that wouldn't work if the complete command line is passed, would it? I think that needs clarification.

jrfnl:
@cmb69 Good point. I just tried with an arbitrary command I had been running anyway and Windows 10 seems to handle it fine:

image

Happy to run additional tests with more complex examples to be sure. Suggestions for commands to test are welcome.

cmb69 (approved):
Ah, right, the escape character substitution happens before the command is going to be executed, so everything is fine with this PR.

jrfnl:
Thanks @cmb69

PR #512, opened 2021-04-05
Updating mysqli: bind_param

kamil-tekiela:
This PR aims to improve examples for parameter binding. The existing examples were simplified to show only the important bits, and a new example was added to show how to bind values from arrays. 

I moved variable assignment before prepare() but I am having second thoughts now. Some people are unaware of how binding by reference works, so creating them after bind_param() in the example makes sense. 

I am not sure about the second-last commit cebb442e040d063cbb8af572a36ca23e91c17159. If the PR is better without it, then we can drop it.


cmb69:
Thank you for the PR! In my opinion, cebb442e040d063cbb8af572a36ca23e91c17159 should better be reverted. It makes the example harder to understand, and isn't a general solution either.

cmb69:
Thank you!

PR #513, opened 2021-04-07
Updating mysqli: fetch_all

kamil-tekiela:
This page was missing an example. I have created a simple example with a single fetch mode. 

While I was at it I read the note about mysqlnd. I have spent more time than I care to admit trying to understand what it said. As a result, I decided to remove it. It might be technically correct; processing the data row by row will use less memory as long as you discard the previous row every iteration. This is a no-brainer. But people looking at fetch_all are interested in fetching all of the data into an array, in which case fetch_all is a perfect choice. I removed the note because it was comparing apples to oranges and I couldn't find any sensible way to rewrite it and still make sense. If we compare apples to apples, then fetch_all consumes the same amount of memory and is faster. 

The procedural parameter snippet was expanded to cover `get_result()`

P.S. After PHP 8.1 release we should remember to add a changelog and then we can remove the mysqlnd section.


cmb69:
Thank you! Good as usual.

PR #479, opened 2021-03-04
Describe WeakMaps.

Crell:
The page was already there, but just auto-generated reference stuff.  Now it explains what it actually does.

PR #514, opened 2021-04-08
Remove old weakref extension

Crell:
1. It doesn't work with current PHP versions.
2. `WeakReference` in PHP 7.4 and `WeakMap` in PHP 8.0 obsolete this extension anyway.
3. The build system can't really handle having both core and an extension defining a `WeakMap` class, so we have to remove this before we can merge #479, which documents the core `WeakMap` class.

If you have a good reason to keep the current pages, speak now or forever hold your documentation.


cmb69:
@colder might want to comment. :)

colder:
Sure, kill it!

Thanks for taking care of this

cmb69:
Fine, so I'm going to merge this. :)

PR #517, opened 2021-04-09
Object iteration / Iterator: Replace duplicated examples with direct links #80942

AllenJB:
https://bugs.php.net/bug.php?id=80942

The example for Iterator on the object iteration page is outdated and raises issues under static analysis because it returns values from methods that are declared to have a void return type.

These examples are already well covered by the interface documentation, and it seems pointless to duplicate them here, so I've replaced them with links to the appropriate pages.

This also cleans up the odd note boxes linking to generators and SPL iterators.

I've additionally added cross-links back to this page so that developers who land on the Iterator and generators pages can see alternative methods.


language/generators.xml
```diff
@@ -3,6 +3,10 @@
 <chapter xml:id="language.generators" xmlns="http://docbook.org/ns/docbook">
  <title>Generators</title>
 
+  <para>
+   See also <link linkend="language.oop5.iterations">object iteration</link>.
+  </para>
```
cmb69, line 8: This would not be rendered as:
Screenshot 2021-04-13 153036

Actually not bad, but may be confusing and undesired. Perhaps it's best to move this paragraph to the `language.generators.comparison` section?

cmb69 (commented):
Thank you! Overall looks very good to me, except maybe for the one issue I commented on.

cmb69:
Thank you!

PR #520, opened 2021-04-10
Fix half-deleted sentence about "new" and reference assignment

IMSoP:
This had a stray ")", and was left with a misleading half-sentence
about "new" returning a reference, which it doesn't.

### DIFF
--- a/language/references.xml
+++ b/language/references.xml
@@ -95,10 +95,13 @@ $foo =& find_var($bar);
 ]]>
       </programlisting>
      </informalexample>
-     <link linkend="language.oop5.basic.new">new</link>
-     returns a reference automatically, thus it is syntactically invalid.
-     For more information, see <link linkend="language.oop5.references">Objects
-     and references</link>.)
+    </para>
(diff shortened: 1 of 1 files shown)


Jaysonspears:
Start```
- [ ] > 
```

IMSoP:
@Jaysonspears Sorry, I'm not sure if github mangled your comment, a cat walked over your keyboard, or what?

cmb69:
Thanks! This is basically good, but please don't use "you" in the manual. Passive voice would be more appropriate. Maybe "It is not possible to assign by reference …"

IMSoP:
@cmb69 I'll try to remember that in future. Reading it back, it didn't flow well anyway, so I've come up with a different wording.

cmb69 (approved):
Ah, even better! Thanks.

>  I'll try to remember that in future.

You may want to check the [style guide](http://doc.php.net/tutorial/style.php). It's poor, but at least we have a few rules. :)

PR #522, opened 2021-04-12
Fix attributes overview example emitting E_WARNING

thinkverse:
The method `targetDirectoryExists` in the attributes overview example emits **E_WARNING** if `$targetDirectory` already exists. This PR fixes this issue by wrapping the `mkdir` call in an `is_dir` check. 

### File exists warning.

Example code ran showing the file exists warning

### Cause of the warning.

https://github.com/php/doc-en/blob/68a416eeb74a6df5b7e31b77cfa5e91a4c368844/language/attributes.xml#L69-L72

### DIFF
--- a/language/attributes.xml
+++ b/language/attributes.xml
@@ -71,7 +71,7 @@ class CopyFile implements ActionHandler
         if (!file_exists($this->targetDirectory)) {
             mkdir($this->targetDirectory);
         } elseif (!is_dir($this->targetDirectory)) {
-            throw new RuntimeException("A file already exists with the name $this->targetDirectory");
+            throw new RuntimeException("Target directory $this->targetDirectory is not a directory");
         }
     }
 
--- a/language/attributes.xml
+++ b/language/attributes.xml
@@ -68,8 +68,10 @@ class CopyFile implements ActionHandler
     #[SetUp]
     public function targetDirectoryExists()
     {
-        if (!is_dir($this->targetDirectory)) {
+        if (!file_exists($this->targetDirectory)) {
             mkdir($this->targetDirectory);
+        } elseif (!is_dir($this->targetDirectory)) {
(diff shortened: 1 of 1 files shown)


cmb69:
See https://github.com/php/doc-en/pull/271#discussion_r536909392 ff.

@kamil-tekiela, what do you think regarding `is_dir()` vs. `file_exists()`?

language/attributes.xml
```diff
@@ -68,7 +68,9 @@ class CopyFile implements ActionHandler
     #[SetUp]
     public function targetDirectoryExists()
     {
-        mkdir($this->targetDirectory);
+        if (!is_dir($this->targetDirectory)) {
+            mkdir($this->targetDirectory);
+        }
```
kamil-tekiela, line 73:
```suggestion
        if (!file_exists($this->targetDirectory)) {
            mkdir($this->targetDirectory);
        } elseif (!is_dir($this->targetDirectory)) {
            throw new RuntimeException("A file already exists with the name $this->targetDirectory");
        }
```

kamil-tekiela (changes requested):

language/attributes.xml
thinkverse: Why also use `file_exists` when `is_dir` already checks if the file exists? It does both, checks for the existence, and also if it's a directory, using `file_exists` and a branching statement seems unnecessary in this case. Just throwing a `RuntimeException` after the initial `is_dir` statement should be enough IMHO.

cmb69: The point is that there may be a *file* with that name (which is not a directory).

thinkverse: Correct yes, my previous comment made really no sense after I double-checked the suggestion, that's why I deleted it. 🙂

cmb69: Ah, so everything is fine now. Thanks for the PR!

PR #523, opened 2021-04-12
Remove or rewrite examples using implode("", file(...))

IMSoP:
This is a weirdly inefficient way of getting a file's content into
a string, and I'm not sure why anyone would think it was a good
idea.


cmb69:
Good catch, thanks! These examples apparently predate PHP 4.3.0.

PR #527, opened 2021-04-14
Expand documentation for mb_detect_encoding

IMSoP:
- People expect this function to work miracles. Add some examples of why that's not possible. Hat tip to this FAQ for the cipher  comparison: https://chardet.readthedocs.io/en/latest/faq.html
- The strict parameter wasn't really explained at all. I *think* this is what it actually does.
- General expansion and grammar fixes (I suspect the original author didn't speak English as a first language, which makes sense since the mbstring extension originated in Japan).

### DIFF
--- a/reference/mbstring/functions/mb-detect-encoding.xml
+++ b/reference/mbstring/functions/mb-detect-encoding.xml
@@ -5,7 +5,7 @@
   <refname>mb_detect_encoding</refname>
   <refpurpose>Detect character encoding</refpurpose>
  </refnamediv>
-   
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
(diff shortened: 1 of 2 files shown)


reference/mbstring/functions/mb-detect-encoding.xml
```diff
@@ -75,23 +95,100 @@
...
+// 'áéóú' encoded in ISO-8859-1
+$str = "\xE1\xE9\xF3\xFA";
+
+// The string is not valid ASCII or UTF-8, but UTF-8 is considered a closer match
+var_dump(mb_detect_encoding($str, ['ASCII', 'UTF-8'], false));
+var_dump(mb_detect_encoding($str, ['ASCII', 'UTF-8'], true));
+
+// If a valid encoding is found, the strict parameter does not change the result
+var_dump(mb_detect_encoding($str, ['ASCII', 'UTF-8', 'ISO-8859-1'], true));
+var_dump(mb_detect_encoding($str, ['ASCII', 'UTF-8', 'ISO-8859-1'], true));
```
cmb69, line 134: I think that is supposed to be:
```suggestion
var_dump(mb_detect_encoding($str, ['ASCII', 'UTF-8', 'ISO-8859-1'], false));
var_dump(mb_detect_encoding($str, ['ASCII', 'UTF-8', 'ISO-8859-1'], true));
```

cmb69 (commented):
Very good patch, again! Thanks.

I'm not sure about the $strict parameter, either.

reference/mbstring/functions/mb-detect-encoding.xml
IMSoP: Good catch, fixed! :)

cmb69: Thanks!

PR #528, opened 2021-04-16
PHP 8.0 migration/deprecated: expand on libxml_disable_entity_loader()

jrfnl:
The current text in the migration guide about the deprecation of `libxml_disable_entity_loader()` is misleading and can easily lead to the introduction of XXE vulnerable code.

In select circumstances, when `LIBXML_NOENT` is used, code can still be vulnerable to XXE attacks, even on PHP 8.0.
So I'm proposing to add an appropriate warning and mention the upgrade path in the migration guide.

Includes fixing a typo on the `libxml_disable_entity_loader()` page.


cmb69 (approved):
Very important info. Thank you!

PR #529, opened 2021-04-16
Fix typo

Yurunsoft:


cmb69:
Agreed, this is a better variable name. Thank you!

PR #530, opened 2021-04-16
Undocumented behavior in FILTER_SANITIZE_STRING

ctcq:
We found the description for the filter `FILTER_SANITIZE_STRING` misleading. It says in https://www.php.net/manual/en/filter.filters.sanitize.php:

> Strip tags, optionally strip or encode special characters.

But apart from removing tags, this filter also encodes quotes.

Example: 
``` php
<?php
echo filter_var("'\"", FILTER_SANITIZE_STRING);
// >> &#39;&#34;
```

This seems to be the result of the following line in the sourcecode: 
https://github.com/php/php-src/blob/0236bbc777e8ef1c72a5bb6b88bfc5f843627eab/ext/filter/sanitizing_filters.c#L194

As this behavior is undocumented and led to some confusion, we propose to modify the documentation to better reflect this behavior.

### DIFF
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -335,7 +335,7 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags and encode "', optionally strip or encode special characters.</entry>
+        <entry>Strip tags and encode <literal>"'</literal>, optionally strip or encode special characters.</entry>
        </row>
        <row>
         <entry><constant>FILTER_SANITIZE_STRIPPED</constant></entry>
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -335,7 +335,7 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags, optionally strip or encode special characters.</entry>
+        <entry>Strip tags and encode "', optionally strip or encode special characters.</entry>
        </row>
        <row>
         <entry><constant>FILTER_SANITIZE_STRIPPED</constant></entry>
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -335,7 +335,7 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags and encode <literal>"'</literal>, optionally strip or encode special characters.</entry>
+        <entry>Strip tags and encode <literal>"'</literal>, optionally strip or encode special characters. Encoding quotes can be disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.</entry>
        </row>
        <row>
         <entry><constant>FILTER_SANITIZE_STRIPPED</constant></entry>
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -335,7 +335,10 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags and encode double and single quotes, optionally strip or encode special characters. Encoding quotes can be disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.</entry>
+        <entry>
+         Strip tags and encode double and single quotes, optionally strip or encode special characters. Encoding quotes can be
+         disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.
+        </entry>
        </row>
        <row>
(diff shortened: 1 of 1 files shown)


reference/filter/filters.xml
```diff
@@ -335,7 +335,7 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags, optionally strip or encode special characters.</entry>
+        <entry>Strip tags and encode <literal>"'</literal>, optionally strip or encode special characters.</entry>
```
cmb69, line 338: Maybe better:
```suggestion
        <entry>Strip tags, optionally strip or encode special characters. Encoding quotes can be disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.</entry>
```
I went with "encode", since it is used elsewhere on this page and is also part of the constant's name, but IMO "escape" is the more appropriate term here.

cmb69 (changes requested):
Thanks for the PR, but that doesn't look right. Quotes are some of the special characters, and whether they are escaped depends on the `FILTER_FLAG_NO_ENCODE_QUOTES` flag: https://3v4l.org/1C4L0

reference/filter/filters.xml
kamil-tekiela: Encode is more correct. Escape is too ambiguous.

kamil-tekiela: The PR is valid. Your suggestion is also valid. They both should be merged. By default, this atrocious filter removes all HTML and encodes quotes. It can be disabled with the flag as you said.

ctcq: I've changed the PR to include both notices

```xml
<entry>Strip tags and encode <literal>"'</literal>, optionally strip or encode special characters. Encoding quotes can be disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.</entry>
```

reference/filter/filters.xml
```diff
@@ -335,7 +335,7 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags, optionally strip or encode special characters.</entry>
+        <entry>Strip tags and encode <literal>"'</literal>, optionally strip or encode special characters. Encoding quotes can be disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.</entry>
```
kamil-tekiela, line 338:
```suggestion
        <entry>Strip tags and encode double and single quotes, optionally strip or encode special characters. Encoding quotes can be disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.</entry>
```

kamil-tekiela (changes requested):

reference/filter/filters.xml
cmb69: And please break the overly long line. :)

cmb69 (approved):

kamil-tekiela:
Upon more research, I came to the conclusion that that whole extension is just rubbish. This filter and probably a few more need to be removed as soon as possible. They are an abomination. 
@ctcq  Out of curiosity, what were you doing that you decided to use FILTER_SANITIZE_STRING? I can't see any sane reason to use this filter ever.

reference/filter/filters.xml
```diff
@@ -335,7 +335,10 @@
          <constant>FILTER_FLAG_ENCODE_HIGH</constant>,
          <constant>FILTER_FLAG_ENCODE_AMP</constant>
         </entry>
-        <entry>Strip tags, optionally strip or encode special characters.</entry>
+        <entry>
+         Strip tags and encode double and single quotes, optionally strip or encode special characters. Encoding quotes can be
+         disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.
```
kamil-tekiela, line 340:
```suggestion
         Strip tags and HTML-encode double and single quotes, optionally strip
         or encode special characters. Encoding quotes can be
         disabled by setting <constant>FILTER_FLAG_NO_ENCODE_QUOTES</constant>.
```

kamil-tekiela: Let's call it HTML-encode. I will change it in other places later.

kamil-tekiela (changes requested):

ctcq:
@kamil-tekiela I'm reviewing some legacy code that uses this filter for GET-Parameter filtering. When I looked up this filter in the docs, I feared that the code was vulnerable for HTML-Attribute injection as php's docs don't explicitly state that this filter encodes quotes.

And yes, those fitlers are an abomination. I'd love to have filters more closely in line with best-practice encoding guidelines like seen here: https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html

kamil-tekiela:
Honestly, I don't know what to do about this page. It needs more love than this simple change. I think we can merge your suggestion and then I will take a deeper look into it once I have few minutes. I will also see what we can do about fixing this extension in the long run.

cmb69:
One improvement would be to change the table to some kind of list, since the former may easily overflow (IIRC, there's a bug ticket regarding this).

PR #534, opened 2021-04-18
Updating mysqli: mysqli_stmt_num_rows

kamil-tekiela:
Improved the page to cause less confusion. I am unsure about the page title. I left it almost unchanged but maybe it should be changed too?

### DIFF
--- a/reference/mysqli/mysqli_stmt/num-rows.xml
+++ b/reference/mysqli/mysqli_stmt/num-rows.xml
@@ -28,7 +28,7 @@
    set in the statement handle.
   </para>
   <para>
-   This function returns 0 unless all rows have been fetched from the server.
+   This function returns <literal>0</literal> unless all rows have been fetched from the server.
   </para>
  </refsect1>
 
@@ -44,7 +44,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   An integer representing the number of buffered rows. Returns 0 in unbuffered
+   An <type>int</type> representing the number of buffered rows. Returns <literal>0</literal> in unbuffered
    mode if all rows have not yet been fetched from the server.
   </para>
  </refsect1>
--- a/reference/mysqli/mysqli_stmt/num-rows.xml
+++ b/reference/mysqli/mysqli_stmt/num-rows.xml
@@ -56,33 +56,18 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-/* Open a connection */
+
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 $mysqli = new mysqli("localhost", "my_user", "my_password", "world");
 
-/* check connection */
-if (mysqli_connect_errno()) {
-    printf("Connect failed: %s\n", mysqli_connect_error());
-    exit();
-}
-
 $query = "SELECT Name, CountryCode FROM City ORDER BY Name LIMIT 20";
-if ($stmt = $mysqli->prepare($query)) {
+$stmt = $mysqli->prepare($query);
+$stmt->execute();
 
-    /* execute query */
-    $stmt->execute();
+/* store the result in an internal buffer */
+$stmt->store_result();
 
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli_stmt/num-rows.xml
```diff
@@ -5,7 +5,7 @@
   <refname>mysqli_stmt::$num_rows</refname>
   <refname>mysqli_stmt::num_rows</refname>
   <refname>mysqli_stmt_num_rows</refname>
-  <refpurpose>Return the number of rows in statements result set</refpurpose>
+  <refpurpose>Return the number of rows in statement&apos;s result set</refpurpose>
```
cmb69, line 8: I think we can drop "statement's" altogether.

reference/mysqli/mysqli_stmt/num-rows.xml
```diff
@@ -22,15 +22,13 @@
    <methodparam><type>mysqli_stmt</type><parameter>statement</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Returns the number of rows in the result set.
-   The use of <function>mysqli_stmt_num_rows</function>
-   depends on whether or not you used
+   Returns the number of buffered rows.
+   The use of this function depends on whether you used
```
cmb69, line 26: Please no "you" on manual pages; passive voice seems suitable here.

reference/mysqli/mysqli_stmt/num-rows.xml
```diff
@@ -22,15 +22,13 @@
...
    <function>mysqli_stmt_store_result</function> to buffer the entire result
    set in the statement handle.
   </para>
   <para>
-   If you use <function>mysqli_stmt_store_result</function>,
-   <function>mysqli_stmt_num_rows</function> may be called immediately.
+   This function returns 0 unless all rows have been fetched from the server.
```
cmb69, line 31: Nit:
```suggestion
   This function returns <literal>0</literal> unless all rows have been fetched from the server.
```

reference/mysqli/mysqli_stmt/num-rows.xml
```diff
@@ -46,7 +44,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   An integer representing the number of rows in result set.
+   An integer representing the number of buffered rows. Returns 0 in unbuffered
```
cmb69, line 47:
```suggestion
   An <type>int</type> representing the number of buffered rows. Returns <literal>0</literal> in unbuffered
```

reference/mysqli/mysqli_stmt/num-rows.xml
```diff
@@ -46,7 +44,8 @@
...
+   mode if all rows have not yet been fetched from the server.
```
cmb69, line 48: This sentence reads strange to me. Maybe better: "if not all rows have yet been fetched from the server." Unsure.

cmb69 (commented):
Thanks you for the PR. Looks generally good to me, but a few improvements might be appropriate.

reference/mysqli/mysqli_stmt/num-rows.xml
kamil-tekiela: I am not sure about this. It seems important to me to distinguish the two `num_rows` in `mysqli_stmt` and `mysqli_result`. I have seen people confuse them a lot. I'll think about this sentence and maybe I will rewrite it.

reference/mysqli/mysqli_stmt/num-rows.xml
kamil-tekiela: I missed that. I'll fix it.

kamil-tekiela:
I redid the PR. I used simpler English than what was used before. I strayed away from the original https://dev.mysql.com/doc/c-api/8.0/en/mysql-stmt-num-rows.html but I believe this explains this function much better than MySQL's original description.

cmb69 (approved):
Great, thank you!

PR #535, opened 2021-04-18
Updating mysqli: mysqli_stmt_store_result

kamil-tekiela:
I have rewritten the description and provided a better title. I also simplified the example by removing all redundant stuff.


cmb69:
Thank you!

PR #537, opened 2021-04-19
Rewrite SoapClient constructor documentation

IMSoP:
This constructor takes a large number of options, most of which
were effectively undocumented, leading to a very high volume of
User Notes. Some examples and caveats from these notes have been
incorporated.

The descriptions here are largely based on reverse-engineering
the source code (which isn't particularly well commented) and
experimenting locally, so may not be 100% accurate, or capture
differences between versions.


reference/soap/soapclient/construct.xml
```diff
@@ -44,116 +45,583 @@
...
+        <varlistentry xml:id="soapclient.construct.options.local-cert">
+         <term>
+          <parameter>local_cert</parameter>
+          <type>string</type>
+         </term>
+         <listitem>
+          <para>
+           Path to a client certificate for use with HTTPS authentication.
+           It must be a PEM encoded file which contains your certificate
+           and private key.
```
cmb69, line 204:
```suggestion
           It must be a PEM encoded file which contains the certificate
           and private key.
```

reference/soap/soapclient/construct.xml
```diff
@@ -44,116 +45,583 @@
...
+          <parameter>classmap</parameter>
+          <type>array</type>
+         </term>
+         <listitem>
+          <para>
+           Used to map types defined in the WSDL to PHP classes.
+           It should be specified as an associative <type>array</type> with
+           type names from the WSDL as keys and names of PHP classes as values.
+           Note that the type names of an element is not necessarily the same as
+           the element (tag) name.
```
cmb69, line 356:
```suggestion
           Note that the type name of an element is not necessarily the same as
           the element (tag) name.
```

reference/soap/soapclient/construct.xml
```diff
@@ -44,116 +45,583 @@
...
+           the constant exists only for backwards compatibility.
+           As of PHP 7.2, specifying <constant>SOAP_SSL_METHOD_TLS</constant>
+           also has no effect; in earlier versions, it forced use of TLS 1.0.
+          </para>
+          <para>
+           Note that SSL versions 2 and 3 are considered insecure, and may not
+           be supported by the installed OpenSSL library.
+          </para>
+          <para>
+           This option is <emphasis>DEPRECATED</emphasis> as of PHP 8.1.0.
```
cmb69, line 601: I detest "shouting":
```suggestion
           This option is <emphasis>deprecated</emphasis> as of PHP 8.1.0.
```
or
```suggestion
           This option is <emphasis role="strong">deprecated</emphasis> as of PHP 8.1.0.
```

reference/soap/soapclient/construct.xml
```diff
@@ -44,116 +45,583 @@
...
+           A more flexible alternative, which allows specifying
+           individual versions of TLS, is to use the
+           <link linkend="soapclient.construct.options.stream-context">
+           <parameter>stream_context</parameter></link> option with
+           the 'crypto_method' context parameter.
+           <example>
```
cmb69, line 607: It might be better to use an `<informalexample>` here (since `<example>`s are numbered). That wouldn't accept a title, though.

cmb69 (commented):
Thank you for the PR! I very welcome improvement.

reference/soap/soapclient/construct.xml
IMSoP: While I agree in general, this particular combination is effectively standard style, because it's what's used in the entities in `language-snippets.ent`, and about a dozen other places.

There are only two instances of `<emphasis>deprecated</emphasis>` in the entire manual, and none at all of `<emphasis role="strong">deprecated</emphasis>` or `DEPRECATED` without any markup (except when mentioning the constant `E_DEPRECATED`).

I think if we want to standardise on a less shouty style, we should pick one and use it consistently.

afilina:
@IMSoP Pinging author

reference/soap/soapclient/construct.xml
cmb69: I'm aware (and not pleased) that we use upper case in the language-snippets, but changing that is additional effort for translators (well, not really much effort). I wouldn't want to add more shouting, though.

cmb69:
> It looks like @cmb69 merged this manually at the time, despite the comments: [6105da1](https://github.com/php/doc-en/commit/6105da168f7d377a7b122b83e25ddac6b5dd2191)

Ugh, that might have been by accident. Anyway, since it has been committed and likely already been translated, leaving as is might be the best option.

afilina:
> I'll leave this PR open as a reminder, and if I get time, will raise a new PR with the suggested improvements.

@IMSoP Might you have the time to raise the new PR?

afilina:
Closing in favor of a new PR which addresses the suggestions: https://github.com/php/doc-en/pull/4930

PR #538, opened 2021-04-20
Fix double "language"

afoster:
Replaced two instances of

> .. not a function but a language language construct.

To

> .. not a function but a language construct.


sy-records (approved):
Thanks

PR #540, opened 2021-04-20
Remove repeated words

kamil-tekiela:
Inspired by #538 I grepped the repo for repeated words.


cmb69:
Very nice! Thank you!

PR #542, opened 2021-04-21
Wouldn't it be more readable if we had a list?

shaedrich:


kamil-tekiela:
Maybe it would be a good idea to make it into a list, but I see that you only added numbers. Could you actually transform it into a proper numbered list?

language/predefined/typeerror.xml
```diff
@@ -16,12 +16,13 @@
...
-    being returned from a function does not match the declared function return
-    type. The third is where an invalid number of arguments are passed to a
-    built-in PHP function (strict mode only).
+    <ooclass><classname>TypeError</classname></ooclass> may be thrown:
+    1. The argument type being passed to a function does not match
+    its corresponding declared parameter type.
+    2. A value being returned from a function does not match the
+    declared function return type.
+    3. The third is where an invalid number of arguments are passed
+    to a built-in PHP function (strict mode only).
```
cmb69, line 25: That could be something like:
```suggestion
    <simplelist>
     <member>
      The argument type being passed to a function does not match
      its corresponding declared parameter type.
     </member>
     <member>  
      A value being returned from a function does not match the
      declared function return type.
     </member>
     <member> 
      The third is where an invalid number of arguments are passed
      to a built-in PHP function (strict mode only).
     </member>
    </simplelist>  
```
Not sure whether that would be an improvement, though.

cmb69 (changes requested):
Thanks for the PR, but as is, that wouldn't look like you expect it to.

language/predefined/typeerror.xml
shaedrich: Ah, thanks for that hint. Why do you think, it might not be an improvement? Of course, it doesn't do anything but increasing the readability.

cmb69:
That now looks like this:
Screenshot 2021-04-22 131322

I think this is a nice improvement, but the third item shouldn't start with "The third is where".

shaedrich:
Oh, right. I missed that.

cmb69 (approved):

cmb69:
Thank you!

shaedrich:
You're welcome!

PR #545, opened 2021-04-24
Removed references to PHP 5, PHP 7.0

saundefined:


reference/ftp/configure.xml
```diff
@@ -8,7 +8,7 @@
   installing PHP.
  </para>
  <para>
-  The Windows version of PHP 5 has built-in support for this extension. You do
+  The Windows version has built-in support for this extension. You do
   not need to load any additional extensions in order to use these functions.
```
cmb69, line 12: Oh, that is no longer the case as of PHP 7.0.0, so should be something like:
````
Windows users must enable <filename>php_ftp.dll</filename> in &php.ini;.
````

cmb69 (approved):
Thank you! There are two issues to resolve; otherwise it's fine to merge.

PR #546, opened 2021-04-25
Updated string & array curly brace access with support notice

ssddanbrown:
The documentation still states that array/string curly brace access is possible but this was deprecated as of PHP 7.4 as per this RFC: https://wiki.php.net/rfc/deprecate_curly_braces_array_access

As of PHP 8.0.0 this syntax is no longer supported. This PR specifies the changes in support for this syntax. I did think about removing these notes altogether but thought they'd still have value for people working with, or migrating, older code.

By the way, this is my first time contributing to any official PHP project so please let me know if there's a different process I should be following or if any actions need to be taken before contributing.

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -358,9 +358,10 @@ string(3) "foo"
    
    <note>
     <para>
-     Both square brackets and curly braces can be used interchangeably
-     for accessing array elements (e.g. <literal>$array[42]</literal> and <literal>$array{42}</literal> will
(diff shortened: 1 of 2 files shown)


cmb69 (approved):
Thanks! Good patch; well done. :)

Note that the PHP manual currently is supposed to document PHP 7.0.0 and up, so yes, it would be too early to remove the mention of this "feature".

PR #547, opened 2021-04-25
Add tip on VACUUM INTO in SQLite3::backup

bohwaz:


cmb69 (changes requested):
Ah, thanks, that's helpful! But please [don't use "you" in the manual](http://doc.php.net/tutorial/style.php). Maybe something like "As of SQLite 3.27.0 (2019-02-07) it is possible to use the statement …"

bohwaz:
> Ah, thanks, that's helpful! But please [don't use "you" in the manual](http://doc.php.net/tutorial/style.php). Maybe something like "As of SQLite 3.27.0 (2019-02-07) it is possible to use the statement …"

Thanks, it's fixed :)

cmb69 (approved):
Cool, thanks!

PR #550, opened 2021-04-27
Add note for passing references to invoke methods.

GSokol:
During running unit tests for a private method I faced the lack of
documentation about how to pass arguments by references through `invoke`
method. (See https://bugs.php.net/bug.php?id=80991) I'd like to add
notes to clarify the workaround to make it easier for others.

### DIFF
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -2160,6 +2160,10 @@ as opposed to emitting it. Setting to &false; (the default) will do the opposite
 <!ENTITY reflection.invoke.reference 'If the function has arguments that need
 to be references, then they must be references in the passed argument list.'>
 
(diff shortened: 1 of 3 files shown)


cmb69 (changes requested):
Thank you for the PR! I missed these notes yesterday.

However, since the PHP manual is not supposed to documented behavior prior to PHP 7.0.0 anymore, there is only the need for two notes: `::invoke()` can't be used at all when reference parameters are expected, and `::invokeArgs()` has to be used instead (passing references in the argument list). Since these two notes need only to be documented twice, I suggest to not introduce new XML entities, but rather to insert the text directly. Still, it makes sense to keep the old XML entity for a while, to not break translations.

cmb69 (commented):
Thank you! I'm going to apply with some changes.

PR #551, opened 2021-04-27
Removed mysql extension

saundefined:
It seems like it's possible to remove the `mysql` extension, 
but I'm not sure 🤔


cmb69 (commented):
I think it's a bit early. Unfortunately, this appears to still be widely used. Okay, there are the [legacy docs](https://php-legacy-docs.zend.com/manual/php5/en/index) now, but there is something [misconfigured on the server](https://marc.info/?l=phpdoc&m=161875795124552&w=2), so that the links to the legacy docs are not shown when searching.

saundefined:
Ah.. got it.
Then I'll convert the PR into a draft, let it wait in the wings

afilina:
@cmb69 Is there a plan to fix the nginx config for the 404s? This PR is accumulating merge conflicts. Alternatively, should we close this PR?

saundefined:
@afilina I guess we can close the PR,
if the nginx problem is solved, it's easier to do that again from scratch

PR #553, opened 2021-04-27
Updating mysqli: mysqli_stmt_affected_rows

kamil-tekiela:
Fixed incorrect information and synchronized the page with https://dev.mysql.com/doc/c-api/8.0/en/mysql-affected-rows.html


cmb69:
Thank you!

PR #554, opened 2021-04-27
Updating mysqli: mysqli_affected_rows

kamil-tekiela:
Similar to https://github.com/php/doc-en/pull/553


cmb69:
Thank you!

Next time, please *consider* to not change formatting; I think translators already have a hard time due the relatively huge amount of changes to doc-en overall.

PR #555, opened 2021-04-28
Cleanup after PHP 5 mention removal

Girgias:
Basic fixes that I saw when updating the FR translation.

There are still many pages which still have references to PHP 5 although said file has been touched.
I might do another pass on the whole documentation when I'm caught up with the updating the French version.


cmb69 (approved):
Thank you!

PR #556, opened 2021-04-28
[80999] Fix count TypeError not documented

thinkverse:
This PR adds the **TypeError** `count()` now throws to the changelog, and also updates the example showing that PHP 8.0 throws a **TypeError** instead of emitting and **E_WARNING**.

First time adding to a changelog, so not 100% sure if the versions are displayed in descending order or ascending. 🤔&nbsp; Will change if requested.

### DIFF
--- a/reference/array/functions/count.xml
+++ b/reference/array/functions/count.xml
@@ -91,6 +91,28 @@ $b[0]  = 7;
 $b[5]  = 9;
 $b[10] = 11;
 var_dump(count($b));
+?>
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen role="php">
+<![CDATA[
+int(3)
+int(3)
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>count</function> non Countable|array example (bad example - don't do this)</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$b[0]  = 7;
+$b[5]  = 9;
+$b[10] = 11;
+var_dump(count($b));
 
 var_dump(count(null));
 
@@ -102,7 +124,6 @@ var_dump(count(false));
     <screen role="php">
 <![CDATA[
 int(3)
-int(3)
 int(0)
 int(1)
 ]]>
@@ -110,7 +131,6 @@ int(1)
     &example.outputs.72;
     <screen role="php">
 <![CDATA[
-int(3)
 int(3)
 
 Warning: count(): Parameter must be an array or an object that implements Countable in … on line 12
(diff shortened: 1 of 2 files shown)


saundefined (approved):
Thank you! Descending order is used

cmb69:
It might make sense to split the error cases from the example to their own example, and then have three `<screen>`s introduced with `&example.outputs;`, `&example.outputs.72;` (this entity needs to be defined in [language-snippets.ent](https://github.com/php/doc-en/blob/master/language-snippets.ent)), and `&example.outputs.8`. Maybe that would be overkill, though.

thinkverse:
@cmb69 is that how you wanted it? Never messes with `<screen>`s and `&example.outputs`, used [foreach.xml](https://github.com/php/doc-en/blob/HEAD/appendices/migration70/incompatible/foreach.xml) as a base.

cmb69:
Yes, that's basically what I meant (you can remove the "as of PHP ..." comments now). But it may also make sense to split the example, so that the good cases (the first two `var_dumps()`) and the bad cases (the last two `var_dumps()`) have their own examples. This would make it possible to have a single `&example.outputs;` for the good case, and maybe to add some words regarding the bad behavior example.

thinkverse:
Should the "bad example" example be after the good example then, or should it be added after the recursive example? Cuz `count` has two examples, the first deals with both a good case and a bad case, the second deals with the recursive aspect of `count`.

cmb69:
I'd put the bad example right after the good example.

thinkverse:
I'm a bit lost on how to "add some words" about the bad example, the title might be enough - I looked around the repo and found a doc using this general idea. Might be OK and get the point across? Otherwise, I'm not sure how to add some words. Can you add notes in or between examples? Those usually go after examples, at least from what I've noticed during my years of looking at the documentation. 🤔

reference/array/functions/count.xml
```diff
@@ -87,6 +87,28 @@ $a[1] = 3;
...
+<![CDATA[
+int(3)
+int(3)
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>count</function> non Countable|array example (bad example - don't do this)</title>
```
cmb69, line 108: If you like, you could add a `<para>` with some text (maybe shortening the title right away). Basically:
```suggestion
    <title>…</title>
    <para>
     …
    </para> 
```

thinkverse: English isn't my first language so writing technical documentation isn't a strong suit for me. Been trying to come up with what to write in `<para>` and given what the example does - showing what outputs different PHP versions give given bad input. Haven't come up with anything that wouldn't turn into redundant text. 🤔

Screen Shot 2021-04-29 at 3 08 56 PM

cmb69: Well, I think we can leave it as is. :)

cmb69:
Thank you very much!

thinkverse:
Glad to help. 👍

PR #558, opened 2021-04-29
[80975] Fix: Doc incomplete - explode

thinkverse:
This PR updates the `<note>` about return values if the start or end of `$string` contains a **separator** value. Also updated the changelog since `explode` now throws a `TypeError` when `$separator` is an empty string.

Personally don't think adding an example regarding the return value would be necessary here. `explode` already has *3* examples, *2* with outputs - and a `<note>` about it being binary-safe, so it would be a bit much IMHO. Therefore I opted to update the return values `<note>` instead.

Also, should the **changelog** be added before **notes** or after? I know it's usually after **examples** and before **seealso**, but not sure where the position should be in regards to **notes**.

### DIFF
--- a/reference/strings/functions/explode.xml
+++ b/reference/strings/functions/explode.xml
@@ -6,7 +6,7 @@
   <refpurpose>Split a string by a string</refpurpose>
  </refnamediv>
  
- <refsect1 role="description"><!-- {{{ -->
+ <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
    <type>array</type><methodname>explode</methodname>
@@ -19,9 +19,9 @@
    <parameter>string</parameter> formed by splitting it on
    boundaries formed by the string <parameter>separator</parameter>.
   </para>
- </refsect1><!-- }}} -->
+ </refsect1>
 
- <refsect1 role="parameters"><!-- {{{ -->
+ <refsect1 role="parameters">
   &reftitle.parameters;
   <para>
    <variablelist>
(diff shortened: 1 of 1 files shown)


cmb69:
> Also, should the changelog be added before notes or after? I know it's usually after examples and before seealso, but not sure where the position should be in regards to notes.

The order is supposed to be changelog, examples, notes. See also https://github.com/php/doc-base/blob/master/RFC/skeletons/function.xml.

Other than that, the patch looks good to me. Thank you!

thinkverse:
> The order is supposed to be changelog, examples, notes. See also https://github.com/php/doc-base/blob/master/RFC/skeletons/function.xml.

Ok, I'll push a change to switch the positions. 👍

thinkverse:
Added in the fold markers while I was there as well.

cmb69:
> Added in the fold markers while I was there as well.

I'm not sure anybody uses these; I usually remove them when copying from the skeletons. :)

thinkverse:
> I'm not sure anybody uses these; I usually remove them when copying from the skeletons. :)

Noticed it was in the skeleton and in most of the other doc pages, so figured wouldn't hurt to add them in. 🤷‍♂️

cmb69:
To prevent a possible edit war: @Girgias what do you think about the folding markers?

Girgias:
> To prevent a possible edit war: @Girgias what do you think about the folding markers?

I don't really care about them, so they can be removed. :-)

thinkverse:
Fold marks reverted then. 👍

cmb69:
I should add: if markers are there, it might be best to keep them (this means less work/confusion for translators).

Anyhow, thank you for the PR!

thinkverse:
> I should add: if markers are there, it might be best to keep them (this means less work/confusion for translators).

Will keep that in mind for future reference. 👍&nbsp; As always, glad to help. 🙂

PR #561, opened 2021-04-29
Add missing sections to zookeeper documentation

thinkverse:
This PR adds missing `parameters` and `returnvalues` sections to [zookeeper](https://github.com/php/doc-en/tree/HEAD/reference/zookeeper) documentation. Default **no parameter** and **void return** sections have been added.

### DIFF
--- a/reference/zookeeper/functions/zookeeper_dispatch.xml
+++ b/reference/zookeeper/functions/zookeeper_dispatch.xml
@@ -5,6 +5,7 @@
   <refname>zookeeper_dispatch</refname>
   <refpurpose>Calls callbacks for pending operations</refpurpose>
  </refnamediv>
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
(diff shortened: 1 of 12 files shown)


thinkverse:
@Girgias sorry to ping you but. I've fixed all the issues for the `zookeeper` documentation. Should I add the rest of the fixes to this draft, or change this draft to a PR only for the `zookeeper` fixes and open a separate PR for each reference fix instead? Since this PR might be a bit big if all fixes are in the same batch. What do you think @cmb69?

Girgias:
Doing it extension by extension might be better IMHO.

Girgias:
Or at least split into different commits even if it's still in the same PR (with meaningful commit messages and all changes to an extension squashed into one commit)

thinkverse:
Guess it would be better to keep them separate, would be easier to resolve potential merge conflicts that way. Plus I rather not deal with squashing via the CLI, it's a pain. I'll update the title, description, and open PRs for separate references in the future.

PR #557, opened 2021-04-29
Remove PHP 5, 7 references

saundefined:
Related to #555.

Some more, but I still keep looking :)

PR #563, opened 2021-04-30
Remove PHP 5, 7 references

saundefined:
Sections:

- array
- datetime
- mbstring
- strings
- var
- xml
- xmlreader
- xmlwriter
- json

ref #557, if such a split is ok, then I'll continue :)


reference/xmlreader/setup.xml
```diff
@@ -15,9 +15,8 @@
  <section xml:id="xmlreader.installation">
   &reftitle.install;
   <para>
-   The XMLReader extension was initially a PECL extension for PHP 5. It was later
-   moved to the PHP source (bundled) as of PHP 5.1.0, and later enabled by default 
-   as of PHP 5.1.2.
+   The XMLReader ships with PHP source.
+   This extension is enabled by default.
```
Girgias, line 19: Remove para tags

reference/xmlwriter/setup.xml
```diff
@@ -15,9 +15,8 @@
  <section xml:id="xmlwriter.installation">
   &reftitle.install;
   <para>
-   The XMLWriter extension was initially a PECL extension for PHP 5. It
-   was later added to the PHP source (bundled) as of PHP 5.1.2. This
-   extension is enabled by default.
+   The XMLWriter ships with PHP source.
+   This extension is enabled by default.
```
Girgias, line 19: Remove para tags

Girgias (changes requested):
Most of the comments are inlined, but the `list()` language construct still has mentions of PHP 7 in the changelog which might need to be ported to another section within the document

saundefined:
@Girgias, thank you!

Made some changes, but I'm not sure about `stripslashes`.

Also, what do you mean about `&warn.deprecated.feature-5-6-0;`, should I add a new generic key (e.g. `warn.deprecated.feature`), or just remove PHP 5.6 from `&warn.deprecated.feature-5-6-0;` message (and in future, we remove it at all)?

Girgias:
> @Girgias, thank you!
> 
> Made some changes, but I'm not sure about `stripslashes`.
> 
> Also, what do you mean about `&warn.deprecated.feature-5-6-0;`, should I add a new generic key (e.g. `warn.deprecated.feature`), or just remove PHP 5.6 from `&warn.deprecated.feature-5-6-0;` message (and in future, we remove it at all)?

You could reuse the `&removed.php.future;` entity.

Revert the stripslashes changes as this was more a comment passing by and that can be resolved at a later point in time.

PR #564, opened 2021-04-30
Remove PHP 5, 7 references

saundefined:
Sections:

- zookeeper
- zlib
- zip
- yaf
- uopz
- uodbc
- openssl
- funchand

ref #557


reference/funchand/functions/func-num-args.xml
```diff
@@ -59,50 +59,6 @@ foo(1, 2, 3);
...
-    </screen>
-    <para>
-     Output in PHP 5.3 and later will be something similar to:
-    </para>
-    <screen>
-<![CDATA[
-Warning: func_num_args():  Called from the global scope - no function
-context in /home/torben/Desktop/code/ml/fna.php on line 3
--1
 ]]>
```
Girgias, line 62: Remove all the `&note.funcnoparam;` in this and other relevant functions (and possibly the whole section) as if you look at the content of it, it refers to behaviour prior to PHP 5.3:

```
<!ENTITY note.funcnoparam '<note xmlns="http://docbook.org/ns/docbook"><para>Because this function depends on the
current scope to determine parameter details, it cannot be used as a
function parameter in versions prior to 5.3.0. If this value must be passed, the results should be assigned
to a variable, and that variable should be passed.</para></note>'>
```

reference/openssl/constants.xml
```diff
@@ -631,7 +631,7 @@
   </variablelist>
   <note>
    <para>
-    This constant was added in 5.3.2 and requires PHP to be 
+    This constant requires PHP to be 
     built with OpenSSL 0.9.8j or greater.
```
Girgias, line 635: Does PHP 7 even accept such a low OpenSSL version?

reference/openssl/functions/openssl-encrypt.xml
```diff
@@ -167,7 +167,7 @@ if (in_array($cipher, openssl_get_cipher_methods()))
    </example>
 
    <example>
-    <title>AES Authenticated Encryption example for PHP 5.6+</title>
+    <title>AES Authenticated Encryption example</title>
```
Girgias, line 170: Maybe this is prior to 7.1

reference/uodbc/configure.xml
```diff
@@ -160,9 +160,8 @@
    </varlistentry>
   </variablelist>
  </para>
- &windows.builtin;
  <simpara>
-  As of PHP 7.0.0, however, Windows users must enable
+  Windows users must enable
   <filename>php_odbc.dll</filename> in order to use this extension.
```
Girgias, line 165: This sentence seems to indicate that there was another way for Windows user to use this extension, @cmb69 can you clarify?

reference/uopz/setup.xml
```diff
@@ -7,7 +7,7 @@
  <section xml:id="uopz.requirements">
   &reftitle.required;
   <para>
-   uopz 2 requires PHP 5.4+. uopz 5.0 requires PHP 7. As of uopz 5.1, PHP 7.1+ is required.
+   As of uopz 5.0 requires PHP 7. As of uopz 5.1, PHP 7.1+ is required.
```
Girgias, line 10:
```suggestion
   As of uopz 5.0 PHP 7.0 is required. As of uopz 5.1, PHP 7.1+ is required.
```

reference/zookeeper/setup.xml
```diff
@@ -10,9 +10,6 @@
   <para>
    This extension requires <link xlink:href="&url.zookeeper;">ZooKeeper</link> C Binding (version 3.4.1 or greater). For more information, see <link xlink:href="&url.zookeeper.doc.cbinding;">ZooKeeper Programmer's Guide</link>.
```
Girgias, line 11: While you're at it, can you break this line into 2...

Girgias (changes requested):
Sorry for taking such a long time to get round to this, looks mostly good.

reference/openssl/constants.xml
saundefined: @Girgias, thank you for the review!

The minimal version for PHP 7.0 is >= 0.9.8, 
and for PHP 7.1 >= 1.0.1

reference/uodbc/configure.xml
cmb69: Prior to PHP 7.0.0, odbc was built statically (i.e. always available).

Girgias: Right, probably would be better to use a more standard sentence, but not sure if we even have one.

reference/openssl/constants.xml
Girgias: ACK

Girgias (approved):
The only thing I'm not sure of, is if we should remove the language-snippet entity, as a translation might fallback to it currently, and there needs to be done some cleanup in it anyway (I'm planning on opening a PR which removes all unused ones in all translations)

reference/uodbc/configure.xml
cmb69: Nope, we don't. :(

cmb69:
Yes, better keep the entity declaration.

PR #569, opened 2021-04-30
Clean up reference/yaf documentation.

thinkverse:
This PR fixes some small issues on **reference/yaf** documentation, some `return values` sections were out-of-order or had completely blank `<para>`s. Fixed to the best of my knowledge after reading up and peeking at the `yaf` source. Also combined a duplicate `seealso` into a single.

Is there a warnings section or something that can be added warning about documentation possibly being out-of-date? Looking at the [repository](https://github.com/laruence/yaf) for `yaf` the current version is `3.3.2`, docs target everything from `1.0.0` to `2.2.0`.

### DIFF
--- a/reference/yaf/yaf_controller_abstract/forward.xml
+++ b/reference/yaf/yaf_controller_abstract/forward.xml
@@ -80,7 +80,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-    return FALSE on failure
+   &return.success;
   </para>
  </refsect1>
 
--- a/reference/yaf/yaf_controller_abstract/forward.xml
+++ b/reference/yaf/yaf_controller_abstract/forward.xml
@@ -10,18 +10,18 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>void</type><methodname>Yaf_Controller_Abstract::forward</methodname>
+   <modifier>public</modifier> <type>bool</type><methodname>Yaf_Controller_Abstract::forward</methodname>
    <methodparam><type>string</type><parameter>action</parameter></methodparam>
    <methodparam choice="opt"><type>array</type><parameter>paramters</parameter></methodparam>
   </methodsynopsis>
   <methodsynopsis>
-   <modifier>public</modifier> <type>void</type><methodname>Yaf_Controller_Abstract::forward</methodname>
+   <modifier>public</modifier> <type>bool</type><methodname>Yaf_Controller_Abstract::forward</methodname>
    <methodparam><type>string</type><parameter>controller</parameter></methodparam>
    <methodparam><type>string</type><parameter>action</parameter></methodparam>
    <methodparam choice="opt"><type>array</type><parameter>paramters</parameter></methodparam>
   </methodsynopsis>
   <methodsynopsis>
-   <modifier>public</modifier> <type>void</type><methodname>Yaf_Controller_Abstract::forward</methodname>
+   <modifier>public</modifier> <type>bool</type><methodname>Yaf_Controller_Abstract::forward</methodname>
    <methodparam><type>string</type><parameter>module</parameter></methodparam>
    <methodparam><type>string</type><parameter>controller</parameter></methodparam>
    <methodparam><type>string</type><parameter>action</parameter></methodparam>
--- a/reference/yaf/yaf_controller_abstract/forward.xml
+++ b/reference/yaf/yaf_controller_abstract/forward.xml
@@ -77,6 +77,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+    return FALSE on failure
+  </para>
+ </refsect1>
+
  <refsect1 role="examples">
   &reftitle.examples;
   <example>
@@ -111,14 +118,6 @@ class IndexController extends Yaf_Controller_Abstract
 ]]>
    </screen>
   </example>
- </refsect1>
-
-
- <refsect1 role="returnvalues">
-  &reftitle.returnvalues;
-  <para>
-    return FALSE on failure
-  </para>
  </refsect1>
 
  <refsect1 role="seealso">
--- a/reference/yaf/yaf_route_map/assemble.xml
+++ b/reference/yaf/yaf_route_map/assemble.xml
@@ -44,7 +44,14 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &string; on success.
+   Returns &string; on success or &null; on failure.
+  </para>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   May throw <classname>Yaf_Exception_TypeError</classname>.
   </para>
  </refsect1>
 
--- a/reference/yaf/yaf_route_map/assemble.xml
+++ b/reference/yaf/yaf_route_map/assemble.xml
@@ -44,7 +44,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns string on success.
+   Returns &string; on success.
   </para>
  </refsect1>
 
--- a/reference/yaf/yaf_route_map/assemble.xml
+++ b/reference/yaf/yaf_route_map/assemble.xml
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
+  </para>
+ </refsect1>
+
  <refsect1 role="examples">
   &reftitle.examples;
   <example>
@@ -89,13 +96,6 @@ string(%d) "/foo/bar/_/tkey1/tval1/tkey2/tval2"
 ]]>
    </screen>
   </example>
- </refsect1>
-
- <refsect1 role="returnvalues">
-  &reftitle.returnvalues;
-  <para>
-  
-  </para>
  </refsect1>
 
 </refentry>
--- a/reference/yaf/yaf_route_regex/assemble.xml
+++ b/reference/yaf/yaf_route_regex/assemble.xml
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
+  </para>
+ </refsect1>
+
  <refsect1 role="examples">
   &reftitle.examples;
   <example>
@@ -85,13 +92,6 @@ string(49) "/module/controller/action?tkey1=tval1&tkey2=tval2"
 ]]>
    </screen>
   </example>
- </refsect1>
-
- <refsect1 role="returnvalues">
-  &reftitle.returnvalues;
-  <para>
-  
-  </para>
  </refsect1>
 
 </refentry>
--- a/reference/yaf/yaf_route_regex/assemble.xml
+++ b/reference/yaf/yaf_route_regex/assemble.xml
@@ -44,7 +44,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns string on success.
+   Returns &string; on success.
   </para>
  </refsect1>
 
--- a/reference/yaf/yaf_route_regex/assemble.xml
+++ b/reference/yaf/yaf_route_regex/assemble.xml
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>string</type><methodname>Yaf_Route_Regex::assemble</methodname>
+   <modifier>public</modifier> <type class="union"><type>string</type><type>null</type></type><methodname>Yaf_Route_Regex::assemble</methodname>
    <methodparam><type>array</type><parameter>info</parameter></methodparam>
    <methodparam choice="opt"><type>array</type><parameter>query</parameter></methodparam>
   </methodsynopsis>
@@ -44,7 +44,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &string; on success.
+   Returns &string; on success or &null; on failure.
   </para>
  </refsect1>
 
--- a/reference/yaf/yaf_route_rewrite/assemble.xml
+++ b/reference/yaf/yaf_route_rewrite/assemble.xml
@@ -44,7 +44,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns string on success.
+   Returns &string; on success.
   </para>
  </refsect1>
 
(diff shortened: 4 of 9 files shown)


reference/yaf/yaf_controller_abstract/forward.xml
```diff
@@ -77,6 +77,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+    return FALSE on failure
```
Girgias, line 83:
```suggestion
    Return &false; on failure.
```
But I think we might have an entity which has this text already in language-snippets.ent.

reference/yaf/yaf_route_map/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
```
Girgias, line 47:
```suggestion
   Returns &string; on success.
```

reference/yaf/yaf_route_regex/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
```
Girgias, line 47:
```suggestion
   Returns &string; on success.
```

reference/yaf/yaf_route_rewrite/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
```
Girgias, line 47:
```suggestion
   Returns &string; on success.
```

reference/yaf/yaf_route_simple/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
```
Girgias, line 47:
```suggestion
   Returns &string; on success.
```

reference/yaf/yaf_route_static/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
```
Girgias, line 47:
```suggestion
   Returns &string; on success.
```

reference/yaf/yaf_route_supervar/assemble.xml
```diff
@@ -41,10 +41,24 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns string on success.
```
Girgias, line 47:
```suggestion
   Returns &string; on success.
```

reference/yaf/yaf_route_supervar/assemble.xml
```diff
@@ -41,10 +41,24 @@
...
+  </para>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Throws <classname>Exception</classname> when <parameter>$info</parameter> key ':c' is not set.
```
Girgias, line 54:
```suggestion
   Throws <classname>Exception</classname> when <parameter>$info</parameter>
   key <literal>':c'</literal> is not set.
```

reference/yaf/yaf_view_simple/construct.xml
```diff
@@ -47,6 +47,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+
+  </para>
```
Girgias, line 54: Void? Or look what other constructors do

Girgias (changes requested):

reference/yaf/yaf_view_simple/construct.xml
cmb69: Constructors don't return; IMHO return value section should be omitted for these (unless aliased like DateTime::__construct()).

reference/yaf/yaf_controller_abstract/forward.xml
thinkverse: That might be the case, `yaf` is written in C, so I'm having a difficult time trying to figure it out, but according to the [source](https://github.com/laruence/yaf/blob/master/yaf_controller.c#L719), at least the master, it returns a `bool`, either `true` or `false`. 🤔

reference/yaf/yaf_route_map/assemble.xml
thinkverse: I'll fix the `&string;` ones in a batch instead of committing the suggestion for each, simpler that way.

reference/yaf/yaf_view_simple/construct.xml
cmb69: And if we'd get bored some time, we might write a script to check existing construct.xml (without aliases) for return value sections. :)

reference/yaf/yaf_controller_abstract/forward.xml
cmb69: Indeed; in that case just use `&return.success;` (and fix the return type in the signature, please).

thinkverse: For the signature is that the `<type>` next to the `<modifier>`? If so, is it just to change `<type>void</type>` to `<type>bool</type>`?

thinkverse: Fixes have been pushed. 👍

reference/yaf/yaf_route_map/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &string; on success.
```
cmb69, line 47: What is returned on failure? The function signature says the method returns string.

reference/yaf/yaf_route_regex/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &string; on success.
```
cmb69, line 47: Same as above.

reference/yaf/yaf_route_rewrite/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &string; on success.
```
cmb69, line 47: ditto

reference/yaf/yaf_route_simple/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &string; on success.
```
cmb69, line 47: ditto

reference/yaf/yaf_route_static/assemble.xml
```diff
@@ -41,6 +41,13 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &string; on success.
```
cmb69, line 47: ditto

reference/yaf/yaf_route_supervar/assemble.xml
```diff
@@ -41,10 +41,25 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &string; on success.
```
cmb69, line 47: If the function either throws, or returns a string, I don't think that we need "on success" here (maybe just "Returns a &string;.")

cmb69 (changes requested):
Thank you! Basically good, but I think the "returns &string; on success" should be clarified/improved.

reference/yaf/yaf_route_map/assemble.xml
thinkverse: Will look further into the source to find out. 👍

thinkverse: It throws `ArgumentCountError` if the first value isn't set, don't think that needs to be documented since it's PHPs built-in behavior. But it gets trickier if the first argument is set to an empty array because the error thrown `Yaf_Exception_TypeError` has different messages depending on whether the class itself has its first argument set to `false` or `true` upon initialization.

#### Senarios

##### Senario one.

```php
$router = new Yaf_Router();
$route  = new Yaf_Route_Map();

$router->addRoute('map', $route);
$router->getRoute('map')->assemble([]);
```

This will throw with the message "Undefined the 'controller' parameter for the 1st parameter" because the array is missing the key `:c`, the route map is initialized to use controllers, thus this key is needed in the array.

##### Senario two.

```php
$router = new Yaf_Router();
$route  = new Yaf_Route_Map(true);

$router->addRoute('map', $route);
$router->getRoute('map')->assemble([]);
```

This will throw with the message "Undefined the 'action' parameter for the 1st parameter" because the route map is now using actions, thus looking for the `:a` key in the array instead.

If an empty string is the value to either key, `null` will be returned, but if the value is set to `null` or `0` the corresponding `Yaf_Exception_TypeError` is thrown.

Not sure how to go about documenting that behavior.

reference/yaf/yaf_route_regex/assemble.xml
thinkverse: This one's also a bit tricky. If the class has the argument `reverse` set, it returns whatever `reverse` is set to, and any transformed values it finds. If it's not set, which is the default behavior, assemble seems to always return `null`.

```php
$route = new Yaf_Route_Regex("#^/product/([^/]+)/([^/])+#", ['controller' => 'product']);

$assemble = $router->getRoute('map')->assemble([]);
var_dump($assemble); // null
```

```php
$route = new Yaf_Route_Regex("#^/product/([^/]+)/([^/])+#", ['controller' => 'product'], [], [], '');

$assemble = $router->getRoute('map')->assemble([]);
var_dump($assemble); // string(0) ""
```

```php
$route = new Yaf_Route_Regex("#^/product/([^/]+)/([^/])+#", ['controller' => 'product'], [], [], 'product');

$assemble = $router->getRoute('map')->assemble([]);
var_dump($assemble); // string(7) "product"
```

```php
$route = new Yaf_Route_Regex("#^/product/([^/]+)/([^/])+#", ['controller' => 'product'], [], [], 'product/:a');

$assemble = $router->getRoute('map')->assemble([':a' => 'action']);
var_dump($assemble); // string(14) "product/action"
```

Creating the router and attaching the route map to the router have been omitted from the examples.

Is it safe to assume this can be a return string on success and null on failure? 🤔&nbsp; Seems the default behavior is to return null unless the `reverse` argument is set.

reference/yaf/yaf_route_rewrite/assemble.xml
thinkverse: This one returns the original URL argument passes to `Yaf_Route_Rewrite` on creation even on failure. Cannot find a case where it doesn't return anything but a string.

reference/yaf/yaf_route_simple/assemble.xml
thinkverse: This one will throw `Yaf_Exception_TypeError` if either `:c` or `:a` key is missing from the info array, otherwise it'll return a string.

```php
$route  = new Yaf_Route_Simple('m', 'c', 'a');

$assemble = $router->getRoute('simple')->assemble([
    ':a' => '',
    ':c' => ''
]);

var_dump($assemble); // string(6) "?c=&a=" 
```

```php
$route  = new Yaf_Route_Simple('m', 'c', 'a');

$assemble = $router->getRoute('simple')->assemble([
    ':a' => '',
]);
```

This will throw a `Yaf_Exception_TypeError` with either of two messages. _You need to specify the controller by ':c'_, or _You need to specify the action by ':a'_ depending on which key is missing from the array.

Seems each route class handles itself a bit differently depending on how the class is created and what is sent to the `assemble` method.

But this one seems to only return either a `string` or it `throws`.

reference/yaf/yaf_route_static/assemble.xml
thinkverse: This one is similar to `Yaf_Route_Simple` and `Yaf_Route_Map` where it returns either a string or throws `Yaf_Exception_TypeError` if `:a` or `:c` key is missing from the argument array, it needs those two keys otherwise it'll throw.

So maybe this as an error section?
```xml
<para>
 Throws <classname>Yaf_Exception_TypeError</classname> if; <literal>':c'</literal>' 
 and <literal>':a'</literal> keys are missing from <parameter>info</parameter>.
</para>
```

reference/yaf/yaf_route_supervar/assemble.xml
thinkverse: Need to update the `errors` on this one, also throws `Yaf_Exception_TypeError`  if key `:a` isn't set.

thinkverse: Could add the same error as for `Yaf_Route_Static` to this one, since both requires `:c` and `:a` keys to be set.

```xml
<para>
 Throws <classname>Yaf_Exception_TypeError</classname> if; <parameter>info</parameter> 
 keys <literal>':c'</literal> and <literal>':a'</literal> are not set.
</para>
```

reference/yaf/yaf_route_map/assemble.xml
cmb69: Thanks for the thorough investigation! The behavior is apparently rather contrieved. Anyhow, maybe document that the method returns string on success, and null on failure, and add a error/exception section that the method may throw `Yaf_Exception_TypeError`.

reference/yaf/yaf_route_regex/assemble.xml
cmb69: > Is it safe to assume this can be a return string on success and null on failure?

Let's say, yes. ;) Please also change the function signature.

reference/yaf/yaf_route_rewrite/assemble.xml
cmb69: Well, then just `Returns &string;`.

reference/yaf/yaf_route_simple/assemble.xml
cmb69: Then please remove the "on success", and add a short error/exception section.

reference/yaf/yaf_route_map/assemble.xml
thinkverse: Sorry for the long delay. I've pushed the necessary changes, which should be good for a re-review. 👍

cmb69:
Thank you!

PR #570, opened 2021-04-30
Fix Yac::__construct documentation.

thinkverse:
This PR adds back the missing `returnvalues` section and adds a new `errors` section indicating what causes the constructor to throw an exception. @laruence could possibly confirm if this PR and the #569 PR are correct in terms of return values and expectations.

Note: Not sure if `<constant>` is a valid way to target constants in a package, see line 48 in PR.

### DIFF
--- a/reference/yac/yac/construct.xml
+++ b/reference/yac/yac/construct.xml
@@ -33,18 +33,21 @@
   </variablelist>
  </refsect1>
 
- <!-- Return values commented out, as constructors generally don't return a
-      value. Uncomment this if you do need a return values section (for
-      example, because there's also a procedural version of the method).
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
    
   </para>
  </refsect1>
- -->
 
-
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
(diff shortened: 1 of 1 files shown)


cmb69:
> Not sure if <constant> is a valid way to target constants in a package, see line 48 in PR.

That's exactly how it is supposed to be. :)

Regarding the return value section of __construct(): that should be removed completey; constructors don't return anything (and as the former comment correctly stated, these return value sections only make sense if the constructor is aliased, like https://www.php.net/manual/en/datetime.construct.php).

thinkverse:
The return values section has been hidden again. 👍

reference/yac/yac/construct.xml
```diff
@@ -44,7 +44,14 @@
  </refsect1>
  -->
 
-
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Throws a <classname>Exception</classname> if Yac is not enabled. Throws
```
Girgias, line 50:
```suggestion
   Throws an <classname>Exception</classname> if Yac is not enabled. Throws 
```

cmb69:
Thank you!

PR #571, opened 2021-04-30
Remove redundant if and use short array syntax

kamil-tekiela:


cmb69 (commented):
Thank you! I wonder whether we should add some info regarding error handling on this page ("the following examples assume ERRMODE_EXCEPTION" or "error handling omitted for brevity").

kamil-tekiela:
I don't believe it's necessary. It's the default anyway. Besides, these examples are not self-contained and they don't show the connection part. Thus the error setting mode is also not shown.

cmb69:
Well, ERRMODE_EXCEPTION is the default as of PHP 8.1.0, which is not even released. But okay, I agree that it's not necessary to mention the omitted error handling for these examples.

kamil-tekiela:
It's the default as of PHP 8.0

PR #574, opened 2021-04-30
Add missing parameter sections to XMLReader methods

thinkverse:
This PR solves the issues of some `XMLReader` methods not having any parameter sections. While they don't accept any arguments, this fact should still be documented correctly, similar to other methods and functions.


cmb69:
Thank you!

PR #576, opened 2021-04-30
Fix multiple `wincache_*` functions having missing parameter sections

thinkverse:
This PR fixes multiple `wincache_*` functions not having parameter sections, most missing non-argument ones. 

`wincache_rplist_fileinfo` was missing its `$summaryonly` parameter section and description. Copied `wincache_fcache_fileinfo` parameter section and description for the latter since its name is the same. I will update the description once I know what the argument does. 🤷‍♂️

### DIFF
--- a/reference/wincache/functions/wincache-fcache-meminfo.xml
+++ b/reference/wincache/functions/wincache-fcache-meminfo.xml
@@ -17,6 +17,12 @@
    Retrieves information about memory usage by file cache.
   </para>
  </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  &no.function.parameters;
+ </refsect1>
+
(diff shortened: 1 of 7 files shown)


thinkverse:
Anyone have any clue what the `wincache_rplist_fileinfo` controls? I have no access to a PHP version with wincache so cannot really check. But my gut tells me it;

"Controls whether the returned array will contain information about individual cache entries along with the cache mappings summary."

But that's based on reading what the other `fileinfo` functions do, not on hard evidence.

thinkverse:
Removed the `wincache_rplist_fileinfo` parameter description, could not find information on what the parameter changes. So felt it was best to just leave the `<para>` blank for the future in case someone else knows what it changes.

cmb69:
Thanks for the PR!

PR #578, opened 2021-04-30
Reflection deprecation

Crell:
Considerable chunks of the ReflectionParameter class are marked as deprecated with no indication of what to do instead.  After some experimentation (while trying to remove the deprecated stuff from TYPO3), I offer the following "do this instead" options.  And some other commentary and improvement along the way.

I'm not 100% sure my XML structuring is right.  Let me know what I should do instead and I'll adjust accordingly.

### DIFF
--- a/reference/reflection/reflectionfunctionabstract/getparameters.xml
+++ b/reference/reflection/reflectionfunctionabstract/getparameters.xml
@@ -14,7 +14,7 @@
    <void />
   </methodsynopsis>
   <para>
-   Get the parameters as an array of <type>ReflectionParameter</type>.
+   Get the parameters as an array of <type>ReflectionParameter</type>, in the order in which they are defined in the source.
   </para>
 
   &warn.undocumented.func;
--- a/reference/reflection/reflectionfunctionabstract/getparameters.xml
+++ b/reference/reflection/reflectionfunctionabstract/getparameters.xml
@@ -14,7 +14,8 @@
    <void />
   </methodsynopsis>
   <para>
-   Get the parameters as an array of <type>ReflectionParameter</type>, in the order in which they are defined in the source.
+  Get the parameters as an array of <type>ReflectionParameter</type>,
+  in the order in which they are defined in the source.
   </para>
 
   &warn.undocumented.func;
--- a/reference/reflection/reflectionfunctionabstract/getparameters.xml
+++ b/reference/reflection/reflectionfunctionabstract/getparameters.xml
@@ -14,8 +14,8 @@
    <void />
   </methodsynopsis>
   <para>
-  Get the parameters as an array of <type>ReflectionParameter</type>,
-  in the order in which they are defined in the source.
+   Get the parameters as an array of <type>ReflectionParameter</type>,
+   in the order in which they are defined in the source.
   </para>
 
   &warn.undocumented.func;
--- a/reference/reflection/reflectionnamedtype/isbuiltin.xml
+++ b/reference/reflection/reflectionnamedtype/isbuiltin.xml
@@ -14,7 +14,7 @@
    <void />
   </methodsynopsis>
   <para>
-   Checks if the type is a built-in type in PHP.
+   Checks if the type is a built-in type in PHP. A built-in type is any type that is not a class.
   </para>
  </refsect1>
 
--- a/reference/reflection/reflectionnamedtype/isbuiltin.xml
+++ b/reference/reflection/reflectionnamedtype/isbuiltin.xml
@@ -14,7 +14,8 @@
    <void />
   </methodsynopsis>
   <para>
-   Checks if the type is a built-in type in PHP. A built-in type is any type that is not a class.
+   Checks if the type is a built-in type in PHP. A built-in type is any type that
+   is not a class, interface, or trait.
   </para>
  </refsect1>
 
--- a/reference/reflection/reflectionparameter/gettype.xml
+++ b/reference/reflection/reflectionparameter/gettype.xml
@@ -86,12 +86,37 @@ var_dump($reflectionType2);
     </programlisting>
     &example.outputs.70;
     <screen>
-<![CDATA[
+     <![CDATA[
 int
 NULL
 ]]>
     </screen>
    </example>
+   <example>
+   <title><methodname>ReflectionParameter::getType</methodname> Usage in PHP 8.0.0 and later</title>
+    <para>
+     As of PHP 8.0.0, this method may return a <classname>ReflectionNamedType</classname> instance or
+     a <classname>ReflectionUnionType</classname> instance.  The latter is a collection of the former.
+     To analyze a type, it is often convenient to normalize it to an array of <classname>ReflectionNamedType</classname>
+     objects.  The following function will return an array of 0 or more <classname>ReflectionNamedType</classname>
+     instances.
+    </para>
+   <programlisting role="php">
+<![CDATA[
+<?php
+function getAllTypes(ReflectionParameter $reflectionParameter): bool {
+    $reflectionType = $reflectionParameter->getType();
+
+    if (!$reflectionType) return [];
+
+    return $reflectionType instanceof ReflectionUnionType
+        ? $reflectionType->getTypes()
+        : [$reflectionType];
+}
+?>
+]]>
+   </programlisting>
+  </example>
   </para>
  </refsect1>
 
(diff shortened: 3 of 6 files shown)


reference/reflection/reflectionparameter/gettype.xml
```diff
@@ -86,12 +86,37 @@ var_dump($reflectionType2);
...
+     As of PHP 8.0.0, this method may return a <classname>ReflectionNamedType</classname> instance or
+     a <classname>ReflectionUnionType</classname> instance.  The latter is a collection of the former.
+     To analyze a type, it is often convenient to normalize it to an array of <classname>ReflectionNamedType</classname>
+     objects.  The following function will return an array of 0 or more <classname>ReflectionNamedType</classname>
+     instances.
+    </para>
+   <programlisting role="php">
+<![CDATA[
+<?php
+function getAllTypes(ReflectionParameter $reflectionParameter): bool {
```
azjezz, line 107:
```suggestion
function getAllTypes(ReflectionParameter $reflectionParameter): array {
```

Crell: Copy pasta FTW.  Fixed.

reference/reflection/reflectionparameter/isarray.xml
```diff
@@ -34,6 +34,32 @@
...
+
+ <example>
+  <title>PHP 8.0.0 equivalent</title>
+  <programlisting role="php">
+<![CDATA[
+<?php
+function allowsArray(ReflectionParameter $reflectionParameter): bool {
+    $reflectionType = $reflectionParameter->getType();
+
+    if (!$reflectionType) return false;
```
azjezz, line 50: If it's untyped, it allows array :) 

```suggestion
    if (!$reflectionType) return true;
```

reference/reflection/reflectionparameter/iscallable.xml
```diff
@@ -38,6 +38,31 @@
...
+
+ <example>
+  <title>PHP 8.0.0 equivalent</title>
+  <programlisting role="php">
+<![CDATA[
+<?php
+function allowsCallable(ReflectionParameter $reflectionParameter): bool {
+    $reflectionType = $reflectionParameter->getType();
+
+    if (!$reflectionType) return false;
```
azjezz, line 54: same
```suggestion
    if (!$reflectionType) return true;
```

reference/reflection/reflectionparameter/isarray.xml
Crell: Hm.  Valid point, but also subtle difference from `isArray()`.  If it's untyped, then `isArray()` would have returned false.

By the same token, should this also check for `mixed`, to make it truely a user-space version of the type check, or should it be strictly "does it explicitly allow for array" and then rename the function?  (Same question applies for the `callable` version.)

reference/reflection/reflectionnamedtype/isbuiltin.xml
```diff
@@ -14,7 +14,7 @@
    <void />
   </methodsynopsis>
   <para>
-   Checks if the type is a built-in type in PHP.
+   Checks if the type is a built-in type in PHP. A built-in type is any type that is not a class.
```
cmb69, line 17: Line might be broken as well, but more importantly, not all interfaces are built-in.

reference/reflection/reflectionparameter/gettype.xml
```diff
@@ -86,12 +86,37 @@ var_dump($reflectionType2);
...
 ]]>
     </screen>
    </example>
+   <example>
+   <title><methodname>ReflectionParameter::getType</methodname> Usage in PHP 8.0.0 and later</title>
+    <para>
+     As of PHP 8.0.0, this method may return a <classname>ReflectionNamedType</classname> instance or
+     a <classname>ReflectionUnionType</classname> instance.  The latter is a collection of the former.
+     To analyze a type, it is often convenient to normalize it to an array of <classname>ReflectionNamedType</classname>
+     objects.  The following function will return an array of 0 or more <classname>ReflectionNamedType</classname>
```
thinkverse, line 101:
```suggestion
     objects. The following function will return an array of 0 or more <classname>ReflectionNamedType</classname>
```

Extra space there that's not needed.

reference/reflection/reflectionparameter/gettype.xml
```diff
@@ -86,12 +86,37 @@ var_dump($reflectionType2);
...
 int
 NULL
 ]]>
     </screen>
    </example>
+   <example>
+   <title><methodname>ReflectionParameter::getType</methodname> Usage in PHP 8.0.0 and later</title>
+    <para>
+     As of PHP 8.0.0, this method may return a <classname>ReflectionNamedType</classname> instance or
+     a <classname>ReflectionUnionType</classname> instance.  The latter is a collection of the former.
```
thinkverse, line 99:
```suggestion
     a <classname>ReflectionUnionType</classname> instance. The latter is a collection of the former.
```

Extra space.

reference/reflection/reflectionparameter/isarray.xml
azjezz: >  should this also check for mixed

untyped is the same as mixed, unless you are talking about `@param` annotations, which we should really bother with here;

reference/reflection/reflectionparameter/gettype.xml
cmb69, line 101:
```suggestion
     objects.  The following function will return an array of <literal>0</literal> or more <classname>ReflectionNamedType</classname>
```

reference/reflection/reflectionparameter/isarray.xml
```diff
@@ -34,6 +34,32 @@
   </para>
  </refsect1>
 
+ <section>
+  As of PHP 8.0.0, the following code will report if a type supports arrays,
+  including as part of a union.
+ </section>
```
cmb69, line 40: Same as above (i.e. don't introduce `<section>`s in `<refentries>`).

reference/reflection/reflectionparameter/iscallable.xml
```diff
@@ -38,6 +38,31 @@
   </para>
  </refsect1>
 
+ <section>
+  As of PHP 8.0.0, the following code will report if a type supports callables,
+  including as part of a union.
+ </section>
```
cmb69, line 44: Again. :)

cmb69 (changes requested):
I left some comments; maybe the build error is related to the `<section>`s, but I'm not sure.

reference/reflection/reflectionparameter/gettype.xml
```diff
@@ -86,12 +86,37 @@ var_dump($reflectionType2);
...
+     To analyze a type, it is often convenient to normalize it to an array of <classname>ReflectionNamedType</classname>
+     objects.  The following function will return an array of 0 or more <classname>ReflectionNamedType</classname>
+     instances.
+    </para>
+   <programlisting role="php">
+<![CDATA[
+<?php
+function getAllTypes(ReflectionParameter $reflectionParameter): array {
```
cmb69, line 107: I like 1tbs, but docs still mandate PEAR CS …

reference/reflection/reflectionparameter/isarray.xml
Crell: Semantically, yes.  Reflection would distinguish.  As written, this code would return false for a type declaration of `mixed` as well as omitted.  Which... is consistent with the old `isArray()` method, but not necessarily what you want.  I'm not sure of the best way to represent that distinction.

reference/reflection/reflectionparameter/gettype.xml
thinkverse: Not sure if @Crell knows or have found the doc style guide, I had a very hard time finding it myself, so I'm just gonna leave this here: http://doc.php.net/tutorial/style.php

Girgias: I mean we don't really enforce the code style, it's all over the place in the manual

reference/reflection/reflectionfunctionabstract/getparameters.xml
```diff
@@ -14,7 +14,8 @@
    <void />
   </methodsynopsis>
   <para>
-   Get the parameters as an array of <type>ReflectionParameter</type>.
+  Get the parameters as an array of <type>ReflectionParameter</type>,
+  in the order in which they are defined in the source.
```
Girgias, line 18: Nit: Missing indent for text.

reference/reflection/reflectionparameter/gettype.xml
cmb69: Another idea for a script: sniff all code examples for CS.

cmb69 (approved):

sy-records (approved):

PR #581, opened 2021-04-30
Fix `fontFamilies` No returnvalues sections

thinkverse:
This PR is frankly a little unnecessary since `description` section already says what's returned, but technically should be in the missing `returnvalues` section. 🤷‍♂️

### DIFF
--- a/reference/ui/functions/ui.draw.text.font.fontfamilies.xml
+++ b/reference/ui/functions/ui.draw.text.font.fontfamilies.xml
@@ -24,6 +24,13 @@
   &no.function.parameters;
  </refsect1>
 
+ <refsect1 role="returnvalues">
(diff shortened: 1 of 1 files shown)


reference/ui/functions/ui.draw.text.font.fontfamilies.xml
```diff
@@ -24,6 +24,13 @@
   &no.function.parameters;
  </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns an array of valid font families for the current system
```
Girgias, line 30:
```suggestion
   Returns an &array; of valid font families for the current system.
```

cmb69:
Thank you!

PR #583, opened 2021-05-03
remove already deleted session function on PHP 5.4.0

mumumu:
The following functions exists even after deleted on PHP 5.4.0.
We can safely delete them going with the removal of PHP 5 related references.


cmb69 (approved):
Thank you!

PR #584, opened 2021-05-04
Document Memcached::OPT_NOREPLY

TysonAndre:
This was added in 2009.
A lot of other constants were also added.
See http://docs.libmemcached.org/memcached_behavior.html

### DIFF
--- a/reference/memcached/constants.xml
+++ b/reference/memcached/constants.xml
@@ -225,7 +225,7 @@
      Storage commands will be sent without spending time waiting for a reply 
      (there would be no reply).
      Retrieval commands such as <function>Memcached::get</function> are unaffected by this setting.</para>
-    <para>Type: <type>bool</type>, default: &false;.</para>
+    <para>Type: &boolean;, default: &false;.</para>
    </listitem>
   </varlistentry>
 
--- a/reference/memcached/constants.xml
+++ b/reference/memcached/constants.xml
@@ -216,6 +216,19 @@
    </listitem>
   </varlistentry>
 
+  <varlistentry xml:id="memcached.constants.opt-noreply">
+   <term><constant>Memcached::OPT_NOREPLY</constant></term>
+   <listitem>
+    <para>
+     Enable this to ignore the result of storage commands 
+     (set, add, replace, append, prepend, delete, increment, decrement, etc.). 
+     Storage commands will be sent without spending time waiting for a reply 
+     (there would be no reply).
+     Retrieval commands such as <function>Memcached::get</function> are unaffected by this setting.</para>
+    <para>Type: <type>bool</type>, default: &false;.</para>
+   </listitem>
+  </varlistentry>
+
   <varlistentry xml:id="memcached.constants.opt-tcp-nodelay">
    <term><constant>Memcached::OPT_TCP_NODELAY</constant></term>
    <listitem>


reference/memcached/constants.xml
```diff
@@ -216,6 +216,17 @@
    </listitem>
   </varlistentry>
 
+  <varlistentry xml:id="memcached.constants.opt-noreply">
+   <term><constant>Memcached::OPT_NOREPLY</constant></term>
+   <listitem>
+    <para>Enable this to specify that you really don’t care about the result from your storage commands
```
cmb69, line 222: Ah, you're using the proper `’`; mostly the manual uses the typographically incorrect ASCII `'`. I wonder whether we want to change it throughout the manual, or maybe to avoid contractions at all (the [style guide](http://doc.php.net/tutorial/style.php) says "Contractions should be used appropriately", whatever that means). cc @Girgias 

Even minor nit: I think it's better to have a line break after `<para>` (at least for multi-line paragraphs).

reference/memcached/constants.xml
```diff
@@ -216,6 +216,17 @@
...
   </varlistentry>
 
+  <varlistentry xml:id="memcached.constants.opt-noreply">
+   <term><constant>Memcached::OPT_NOREPLY</constant></term>
+   <listitem>
+    <para>Enable this to specify that you really don’t care about the result from your storage commands 
+    (set, add, replace, append, prepend, delete, increment, decrement, etc.). 
+    This will not wait for a response from the server (there would be no response).
+    Retrieval commands such as get are unaffected by this setting.</para>
+    <para>Type: <literal>boolean</literal>, default: &false;.</para>
```
cmb69, line 226:
```suggestion
    <para>Type: &bool;, default: &false;.</para>
```
That doesn't add proper markup to bool, but that's a general PhD issue.

cmb69 (changes requested):
Thank you!

reference/memcached/constants.xml
Girgias: Huh I didn't know there was a proper apostrophe, it might make more sense to use `'` as I think the ship for accurate typography has sailed long ago in the internet age, but I don't mind either way other then changing it everywhere in the manual will probably need to be done carefully and wait until rev-check can skip a revision

Girgias: Also agree with the nit, a line break is batter for long paragraphs IMHO

cmb69: Good point regarding revcheck! And on a second look there should be no "you" here anyway. Can you please rephrase using passive voice?

TysonAndre:
A lot of the command documentation elsewhere was based on libmemcached docs, such as http://docs.libmemcached.org/memcached_behavior.html#MEMCACHED_BEHAVIOR_RCV_TIMEOUT - I assume we shouldn't have done this initially, but now that it's already translated, it's too late to change the `<literal>boolean</literal>` and use of `you`

reference/memcached/constants.xml
TysonAndre:
```
ERROR (file:////home/runner/work/doc-en/doc-en/en/reference/memcached/constants.xml:228:23)
    <para>Type: &bool;, default: &false;.</para>
-----------------------^
 Entity 'bool' not defined


Previous errors too severe. Stopping here.
```

reference/memcached/constants.xml
TysonAndre: Done

reference/memcached/constants.xml
```diff
@@ -216,6 +216,30 @@
    </listitem>
```
TysonAndre, line 216:
```
    <para>Enables or disables asynchronous I/O. This is the fastest transport
     available for storage functions.</para>
    <para>Type: <literal>boolean</literal>, default: &false;.</para>
```


i.e. there'll be a mix of `boolean` and `bool` - does the translation system have issues with replacing literals?

reference/memcached/constants.xml
cmb69: Oh, right, that still needs to be `&boolean;` (or `<type>bool</type>`). We should probably define `&bool;` as alias of `&boolean;`.

reference/memcached/constants.xml
cmb69: It does not replace literals at all (it's just about the markup).

cmb69 (commented):
I'm going to slightly change the wording to better match the rest of the page, and I'm going to change all type markup.

Thanks for the PR! And thanks for the reminder!

PR #586, opened 2021-05-05
Improve openssl-pkcs7-read - fix arguments

vjardin:
The input shall be a string data.

Basic example along with the error cases.

Fix: related to, https://github.com/php/php-src/commit/bb0107b63dec8de5b0f3c3f535d3c52c2decda28


vjardin:
thanks

PR #587, opened 2021-05-05
Remove extra full stop

SergeyBiryukov:
Just removing an extra full stop before a constant name.

Follow-up to f3b5475 / #528.


cmb69:
Thank you!

PR #589, opened 2021-05-06
Documenting the changes to syslog.filter

SvenRtbg:
https://github.com/php/php-src/pull/4265 restored the previous handling of multi-line log messages by adding `syslog.filter=raw`.

Relates to https://bugs.php.net/bug.php?id=77913

### DIFF
--- a/reference/errorfunc/ini.xml
+++ b/reference/errorfunc/ini.xml
@@ -526,8 +526,7 @@
      <para>
       Specifies the filter type to filter the logged messages. Allowed
       characters are passed unmodified; all others are written in their
-      hexadecimal representation prefixed with <literal>\x</literal>. There are
-      four supported filter types:
+      hexadecimal representation prefixed with <literal>\x</literal>. 
       <itemizedlist>
        <listitem>
         <simpara><literal>all</literal> – the logged string will be split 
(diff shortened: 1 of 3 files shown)


reference/errorfunc/ini.xml
```diff
@@ -538,6 +538,9 @@
        <listitem>
         <simpara><literal>ascii</literal> – all printable ASCII characters and <literal>NL</literal></simpara>
        </listitem>
+       <listitem>
+        <simpara><literal>raw</literal> – all characters, newlines will not trigger additional log lines (identical behaviour prior to PHP 7.3)</simpara>
```
cmb69, line 542: This should also mention that `raw` is available as of PHP 7.3.8 and PHP 7.4.0, respectively. Also better to split the long line. Maybe:
```suggestion
        <simpara>
         <literal>raw</literal> – all characters, newlines will not trigger additional log lines
         (identical behaviour prior to PHP 7.3.0).
         Available as of PHP 7.3.8 and PHP 7.4.0, respectively.
        </simpara>
```

cmb69 (changes requested):
Thank you!

reference/errorfunc/ini.xml
SvenRtbg: I wanted to highlight the fact that line splitting is happening in basically all other modes, which annoyed me very much. Wasn't expecting it when the filter is only talking about ... well ... filtering characters. It effectively destroys the logging context.

Line length has been shortened.

I do want to add a link from "syslog()" to here to highlight there is a setting affecting the syslog calls.

SvenRtbg:
I hope this does not mess up anything unexpectedly. I was trying to recreate what other places use for the same effect. :)

cmb69 (approved):
´Thanks!

PR #590, opened 2021-05-07
file_get_contents methodsynopsis

saundefined:
Rename `maxlen` parameter to `length`.

Bug [#81018](https://bugs.php.net/bug.php?id=81018)
[basic_functions.stub.php](https://github.com/php/php-src/blob/e8ef923dd6f6377c28c58fb64f632d2d4f3e8857/ext/standard/basic_functions.stub.php#L861)


cmb69 (approved):
Thank you, this is good to be merged.

Generally, though, we shouldn't bother with individual changes like this, but rather should address appropriate sets of functions with all relevant stub changes (in this case some initializers are missing; adding these might require changelog enteries wrt. PHP 7).

PR #594, opened 2021-05-10
Fold comments into interfaces page

Crell:
I deleted nearly all of the comments on this page and replaced them with this PR.  Most changes are based on comments that I felt were worth mentioning in some way, but not as a comment.  There's also one or two that may have been true in PHP 5.1 but are no longer true now, like the constructor mention, or coding style conventions.

### DIFF
--- a/language/oop5/interfaces.xml
+++ b/language/oop5/interfaces.xml
@@ -5,7 +5,8 @@
   <para>
    Object interfaces allow you to create code which specifies which methods a
    class must implement, without having to define how these methods are
-   implemented.
+   implemented.  Interfaces share a namespace with classes and traits, so may
+   not use the same name.
   </para>
   <para>
    Interfaces are defined in the same way as a class, but with the <literal>interface</literal>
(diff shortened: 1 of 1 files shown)


kamil-tekiela (changes requested):

language/oop5/interfaces.xml
```diff
@@ -16,14 +17,34 @@
...
+  </para>
   <note>
    <para>
-    It is possible to declare
-    <link linkend="language.oop5.magic">magic methods</link> such as the
-    <link linkend="language.oop5.decon.constructor">constructor</link>
-    in an interface, which can be useful in some contexts,
-    e.g. for use by factories.
+    Although they are supported, including <link linkend="language.oop5.decon.constructor">constructors</link>
+    in interfaces is strongly discouraged. That greatly reduces the flexibility of an object implementing the
```
kamil-tekiela, line 45:
```suggestion
    in interfaces is strongly discouraged. Doing so significantly reduces the flexibility of the object implementing the
```
Isn't this better?

language/oop5/interfaces.xml
```diff
@@ -16,14 +17,34 @@
...
+    interface.  Additionally, constructors are not enforced by inheritance rules, leading to possibly inconsistent
```
kamil-tekiela, line 46:
```suggestion
    interface.  Additionally, constructors are not enforced by inheritance rules, which can cause inconsistent
```

kamil-tekiela (changes requested):

language/oop5/interfaces.xml
```diff
@@ -5,7 +5,8 @@
   <para>
    Object interfaces allow you to create code which specifies which methods a
    class must implement, without having to define how these methods are
-   implemented.
+   implemented.  Interfaces share a namespace with classes and traits, so may
+   not use the same name.
```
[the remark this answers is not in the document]
Crell: Why?

Crell, line 9:
```suggestion
   implemented.  Interfaces share a namespace with classes and traits, so they may
   not use the same name.
```

kamil-tekiela (approved):

kamil-tekiela:
Did I not approve it correctly? Why did you rerequest review?

language/oop5/interfaces.xml
```diff
@@ -16,14 +17,34 @@
    All methods declared in an interface must be public; this is the nature of an
    interface.
   </para>
-
+  <para>
+   In practice, interfaces serve two complementary purposes:
+  </para>
+  <simplelist>
+   <member>
+    To allow developers to create multiple objects that may be used interchangeably
```
cmb69, line 25: Isn't that true for classes as well? Maybe "create objects of different classes"?

language/oop5/interfaces.xml
```diff
@@ -200,20 +229,84 @@ class d implements c
...
+<?php
+interface A
+{
+    public function foo(string $s): string;
+
+    public function bar(int $i): int;
+}
+
+// An abstract class may implement only a portion of an interface.
+// It must not re-define the interface methods.
```
cmb69, line 264: Wouldn't that mean that implementing any interface method is not allowed?

Crell: No.  If you have an interface with 2 methods, and an abstract class that implements it, that abstract class can implement zero, one or two of those methods.  However, it may not re-specify as abstract one of the methods from the interface.  That's redundant, and apparently the linter complains about it.

language/oop5/interfaces.xml
Crell:
```suggestion
    To allow developers to create objects of different classes that may be used interchangeably
```

language/oop5/interfaces.xml
cmb69: It's about the wording. I've learned that "must" means "has to", but "must not" means "is not allowed to" – that might be wrong, though.

Crell: Hm.  A previous comment and my earlier (perhaps wrong) testing implied it really was a "is not allowed to."  I just tested again, though, and it appears that's not the case, at least not in current versions: https://3v4l.org/M0oAr

I'll remove that part of the comment.

PR #595, opened 2021-05-10
Fold in comments for get_object_vars()

Crell:
These were the only comments I thought worth folding in.  The rest were for PHP 4 and didn't even make sense then...

Another 37-ish comments exterminated.


cmb69:
Thank you!

PR #596, opened 2021-05-10
Redesign parameter list for PDO_Statement::fetchAll

kamil-tekiela:
This was missed in review. The overloaded signature is not identical to `setFetchMode`. 
Additionally, specified that these parameters do not work with named parameters.

### DIFF
--- a/reference/pdo/pdostatement/fetchall.xml
+++ b/reference/pdo/pdostatement/fetchall.xml
@@ -65,32 +65,25 @@
       </para>
      </listitem>
     </varlistentry>
+   </variablelist>
+   The following are dynamic parameters that are depending on the fetch mode.
+   They can't be used with named parameters.
+   <variablelist>
     <varlistentry>
-     <term><parameter>args</parameter></term>
+     <term><parameter>column</parameter></term>
      <listitem>
       <para>
-       This argument has a different meaning depending on the value of 
-       the <parameter>mode</parameter> parameter:
-       <itemizedlist>
-        <listitem>
-         <para>
-          <constant>PDO::FETCH_COLUMN</constant>: Returns the indicated 0-indexed 
-           column.
-         </para>
-        </listitem>
-        <listitem>
-         <para>
-          <constant>PDO::FETCH_CLASS</constant>: Returns instances of the specified
-          class, mapping the columns of each row to named properties in the class.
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
I have no idea how this looks like because I don't know how to compile the documentation. @cmb69 Would you be able to check if it looks correct, please?

cmb69:
Would look like
Screenshot 2021-06-27 131519
That seems to be an improvement, but I'm not sure whether the two variablelists could cause issues with some tooling.

It's hard to document such contrieved APIs.

kamil-tekiela:
Then I am not sure what to do. It looks to be an improvement, but it's still not as clear as I have hoped. And if it conflicts with tools then I don't know if we should do it. 
The synopsis needs to be fixed either way to correct the typo Mate introduced accidentally.

afilina:
@kamil-tekiela It's not perfect, but the improvement is worth it, in my opinion. However, do we _need_ to have separate variablelists? It doesn't improve clarity and potentially causes issues with external tooling.

Does it make sense to document each param in a way that removes ambiguity? We could add a tiny usage example such as `$stmt->fetchAll(PDO::FETCH_COLUMN, 0)` mention that it is used _only_ with the specified mode, and add the warning about named parameters. It's more repetition, but at least it will be more clear.

kamil-tekiela:
@afilina Thanks. I'll think about it. Please feel free to submit an alternative PR. I admit my idea to solve this issue wasn't the best.

afilina:
@kamil-tekiela I don't think that I'll have the time for a new PR in a reasonably near future. I want to finish reviewing what's already open. I also have some packages with long dependency chains that need to get ready for PHP 8.1.

afilina:
@cmb69 I think that although this is not prefect, it's an improvement that is worth merging. Should multiple variablelists cause an issue with some tooling, we'll fix it at that time. So far, we know that phd can generate xhtml, and that could be deemed good enough. Thoughts?

cmb69:
Indeed, should have merged this long ago. If there are any tooling issues, users can complain, and we can see. But otherwise, this is actually a bugfix (parameter names have been wrong).

If we had a requirement that those designing new APIs would have to write the documentation first, this API mess would not have happened.

Anyhow, thanks all!

PR #600, opened 2021-05-13
Update ext/pcntl methodsynopses based on stubs

kocsismate:


reference/pcntl/functions/pcntl-async-signals.xml
```diff
@@ -9,12 +8,12 @@
...
   &reftitle.description;
-  <methodsynopsis role="procedural">
+  <methodsynopsis>
    <type>bool</type><methodname>pcntl_async_signals</methodname>
-   <methodparam choice="opt"><type>bool</type><parameter>on</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>bool</type><type>null</type></type><parameter>enable</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   If the <parameter>on</parameter> parameter is omitted,
+   If the <parameter>enable</parameter> parameter is omitted,
```
Girgias, line 16:
```suggestion
   If the <parameter>enable</parameter> parameter is omitted or &null,
```
That's for what the `null` defaut is for right?
As such this probably needs a PHP 8 changelog entry

reference/pcntl/functions/pcntl-sigwaitinfo.xml
```diff
@@ -9,13 +9,13 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>pcntl_sigwaitinfo</methodname>
-   <methodparam><type>array</type><parameter>set</parameter></methodparam>
-   <methodparam choice="opt"><type>array</type><parameter role="reference">siginfo</parameter></methodparam>
+   <type class="union"><type>int</type><type>false</type></type><methodname>pcntl_sigwaitinfo</methodname>
```
Girgias, line 12: Need to specify the return false on failure entity in the return section

reference/pcntl/functions/pcntl-strerror.xml
```diff
@@ -10,8 +9,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type class="union"><type>string</type><type>false</type></type><methodname>pcntl_strerror</methodname>
-   <methodparam><type>int</type><parameter>errno</parameter></methodparam>
+   <type>string</type><methodname>pcntl_strerror</methodname>
+   <methodparam><type>int</type><parameter>error_code</parameter></methodparam>
```
Girgias, line 13: Did this return false in PHP 7?

reference/pcntl/functions/pcntl-wstopsig.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>pcntl_wstopsig</methodname>
+   <type class="union"><type>int</type><type>false</type></type><methodname>pcntl_wstopsig</methodname>
```
Girgias, line 12: Same false on failure in return section

reference/pcntl/functions/pcntl-wtermsig.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>pcntl_wtermsig</methodname>
+   <type class="union"><type>int</type><type>false</type></type><methodname>pcntl_wtermsig</methodname>
```
Girgias, line 12: Ditto

reference/pcntl/functions/pcntl-async-signals.xml
kocsismate: >That's for what the null defaut is for right?

Yes, that's true.

> As such this probably needs a PHP 8 changelog entry

IIRC we haven't added changelog so far when the default value handling changed. But let's ask @cmb69 what his suggestion is.

kocsismate: Another question:
Should we really mention "omitted" in the sentence above? Personally, I'd prefer if we completely forgot the legacy of wrong default value handling, and only mention what happens in case the argument is assigned its default value, no matter if it was omitted or explicitly provided.

reference/pcntl/functions/pcntl-strerror.xml
kocsismate: Yes, in case of ZPP failure

reference/pcntl/functions/pcntl-async-signals.xml
cmb69: Changelog should state that the param is nullable as of PHP 8.0.0. The "omitted" can IMO be removed.

reference/pcntl/functions/pcntl-wtermsig.xml
kocsismate: Thank you for spotting all these glitches!

reference/pcntl/functions/pcntl-async-signals.xml
kocsismate: Thank you, I'll fix my latest commit!

reference/pcntl/functions/pcntl-exec.xml
```diff
@@ -9,10 +9,10 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>void</type><methodname>pcntl_exec</methodname>
+   <type>bool</type><methodname>pcntl_exec</methodname>
```
cmb69, line 12: Does this function really return `false` on success, as well as on execution failure?

kocsismate: you are right! OMG, that's crazy :D

cmb69: Maybe change it so it returns a random number. ;)

kamil-tekiela: `4` is a good random number...

cmb69 (commented):
Thank you! I've applied with some additions.

I should note that `pcntl_wexitstatus()`, `pcntl_wstopsig()` and `pcntl_wtermsig()` should be revised, so that they are not defined if the feature is not available, instead of returning `false` as it's now. If someone needs to get the `false` value, they can still call `pcntl_exec()`.

I'm also shocked to see that there's `&return.nullorfalse;` (which is still used twice in the manual), which resolves to "Returns &null; on success&return.falseforfailure;."

PR #604, opened 2021-05-16
Update types for IntlChar::charType()

colinodell:
Based on https://github.com/php/php-src/blob/5b01c4863fe9e4bc2702b2bbf66d292d23001a18/ext/intl/uchar/uchar.stub.php#L29-L30


colinodell:
It's been a while since I last updated the docs so I'd appreciate a double-check of my work here :)

cmb69:
Thank you!

PR #606, opened 2021-05-18
Use delimiter instead of seperator for fgetcsv

alexander-schranz:
See https://github.com/php/php-src/blob/ac47162add6fa529836467929e7bcf20025638b7/ext/standard/file.c#L1985

### DIFF
--- a/reference/filesystem/functions/fgetcsv.xml
+++ b/reference/filesystem/functions/fgetcsv.xml
@@ -12,7 +12,7 @@
    <type>array</type><methodname>fgetcsv</methodname>
    <methodparam><type>resource</type><parameter>stream</parameter></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>length</parameter><initializer>0</initializer></methodparam>
-   <methodparam choice="opt"><type>string</type><parameter>separator</parameter><initializer>","</initializer></methodparam>
+   <methodparam choice="opt"><type>string</type><parameter>delimiter</parameter><initializer>","</initializer></methodparam>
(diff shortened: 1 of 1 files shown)


cmb69:
This change would need to go into https://github.com/php/php-src/blob/PHP-8.0/ext/standard/basic_functions.stub.php first, but I doubt a respective PR would be accepted, since it is a BC break (named arguments), and separator appears to be quite apt for CSV (comma separated values).

alexander-schranz:
@cmb69 atleast something wrong in the docs:

 - https://www.php.net/manual/en/function.fgetcsv.php
 - https://www.php.net/manual/en/function.fputcsv.php
 
 Bot are in the following file  `seperator`:
 
 https://github.com/php/php-src/blob/PHP-8.0/ext/standard/basic_functions.stub.php

 but in soure code `delimiter`:
 
 https://github.com/php/php-src/blob/ac47162add6fa529836467929e7bcf20025638b7/ext/standard/file.c#L1985

alexander-schranz:
Also want to mention here that example in german docs they are both `delimiter`:

 - https://www.php.net/manual/de/function.fgetcsv.php
 - https://www.php.net/manual/de/function.fputcsv.php 
 
For fgetcsv its different between languages.

cmb69:
> but in soure code delimiter:

That is an internal name, and as such pretty much irrelevant for userland. As of PHP 8.0.0, what matters is Reflection, see https://3v4l.org/2n243. doc-en for fputcsv is wrong, though. DE docs are likely grossly out-dated.

alexander-schranz:
Okay seems in php 8 the `fputcsv` and `fgetcsv` is really different:

https://3v4l.org/64Kch

PR #607, opened 2021-05-18
Fix fputcsv for php 8 to separator

alexander-schranz:
As pointed out in https://github.com/php/doc-en/pull/606#issuecomment-843098871 by @cmb69 

fputcsv is now seperator not delimiter: https://3v4l.org/BBCBF


cmb69 (approved):
Thanks!

Note that it is generally preferable to accumulate such changes, though; see https://github.com/php/doc-en/pulls?q=is%3Apr+is%3Aopen+label%3Amethodsynopses.

salathe:
The parameter is still referred to as `delimiter` (both as parameter name and noun) elsewhere on the `fputcsv()` page.
@alexander-schranz  would you mind fixing that up?

PR #608, opened 2021-05-19
Clean up Classname::Method refs

saundefined:
Removed unnecessary `<methodname>Classname::Method</methodname>` refs


cmb69 (approved):
D'oh, the manual is full of nonsense. Thanks for improving!

PR #610, opened 2021-05-20
Fix the default value of `mysqli.rollback-on-cached-plink`

ada-u:
This PR fixes the default value of `mysqli.rollback-on-cached-plink`. It's is `"0"` as set here. 
https://github.com/php/php-src/blob/php-8.0.6/ext/mysqli/mysqli.c#L490


cmb69:
Thank you! That is correct even for PHP 7.0.0, cf. https://github.com/php/php-src/blob/php-7.0.0/ext/mysqli/mysqli.c#L515.

PR #611, opened 2021-05-20
Clarify that \Q and \E don't affect delimiters

IMSoP:
Inspired by https://bugs.php.net/bug.php?id=81060


reference/pcre/pattern.syntax.xml
```diff
@@ -600,7 +600,11 @@
    regexp metacharacters in the pattern. For example:
    <literal>\w+\Q.$.\E$</literal> will match one or more word characters,
    followed by literals <literal>.$.</literal> and anchored at the end of
-   the string.
+   the string. Note that this does not change the behaviour of
```
cmb69, line 603: I suggest to use American English:
```suggestion
   the string. Note that this does not change the behavior of 
```

cmb69 (approved):
Thank you! Nice clarification.

PR #612, opened 2021-05-20
Remove outdated docs

Girgias:
Some of the other extension docs that could be removed:

- Gender
- Mhash
- radius
- ssdeep
- zmq


Girgias:
If there are no comments on this I'll proceed with merging this on Wednesday

cmb69:
Thanks for the PR!

* the [trunk of FrontbaseSQL](http://svn.php.net/viewvc/pecl/fbsql/trunk/) claims PHP 7.3 support; there are no releases though; not sure what to do. cc @fmk 
* geopip 1.1.1 claims PHP 7 support; maybe keep it for now?

Rest should be removed.

Regarding the other extensions:
* gender 1.1.0 claims PHP 7 support; maybe keep it for now?
* mhash is bundled with PHP 7 (part of ext/hash); docs need to be kept
*  not sure about MongoDB's legacy driver; @jmikola, can you please clarify the status of that extension?
* radius 1.4.0b1 claims PHP 7 support
* what is ssdeed?
* stats can be removed
* zmq 1.1.3 claims PHP 7 support

jmikola:
The [`mongo`](http://pecl.php.net/package/mongo) extension has been deprecated for a few years now (and was never updated for PHP 7). I think it's reasonable to remove it from the PHP manual. I'll discuss internally with @alcaeus about whether we (MongoDB) want to mirror the final version of its documentation for historical reference.

Thanks for asking!

Girgias:
> Thanks for the PR!
> 
>     * the [trunk of FrontbaseSQL](http://svn.php.net/viewvc/pecl/fbsql/trunk/) claims PHP 7.3 support; there are no releases though; not sure what to do. cc @fmk
> 
>     * geopip 1.1.1 claims PHP 7 support; maybe keep it for now?
> 
> 
> Rest should be removed.
> 
>     * what is ssdeed?

Sorry was meant to be ``ssdeep``

> 
>     * stats can be removed
> 

Will add this then


> The [`mongo`](http://pecl.php.net/package/mongo) extension has been deprecated for a few years now (and was never updated for PHP 7). I think it's reasonable to remove it from the PHP manual. I'll discuss internally with @alcaeus about whether we (MongoDB) want to mirror the final version of its documentation for historical reference.

I'll add it to the PR, let me know if you want to archive it, but it should also be available on Zend's PHP 5 doc mirror

cmb69:
> Sorry was meant to be ssdeep

np! This seems to support PHP 7 as well, so better keep the docs for now.

Girgias:
@jmikola could you review the changes made to `mongodb/architecture.xml`, as this document (and `set.xml`) referred to the old driver.

reference/mongo/manual.xml
```diff
@@ -1,53 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
-
-<part xml:id="mongo.manual" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
- <title>Manual</title>
- <partintro xml:id="mongo.manual.intro">
-  <warning>
-   &mongo.deprecated.note;
```
jmikola, line 8: This reminds me that there are a number of entities in `language-snippets.ent` that can be deleted. Everything specific to the `mongo` extension (and suitable for deletion) _should_ be organized under the [`<!-- Mongo -->` comment](https://github.com/php/doc-en/blob/3c31131ae899b6bea6dc1d46475160afe0995773/language-snippets.ent#L2353). `mongo.deprecated.note` is the last entity before the `<-- mongodb -->` snippets (for the current driver) start. I would suggest just deleting everything between `<!-- Mongo -->` and `<!-- mongodb -->`, rebuilding the docs, and assuming everything is OK provided you don't get errors for missing entities.

There are also a number of [entities in `doc-base/entities/global.ent`](https://github.com/php/doc-base/blob/30b06be29baf37d907b28bb405ecf777a5fcf082/entities/global.ent#L268); however, I'm not entirely certain that the entries under `<!-- pecl/mongo -->` aren't also used by the `mongodb` docs. Feel free to leave `global.ent` as-is and either @alcaeus or I can clean that up after this gets merged.

If there is some PHP docs tooling that generates a report of defined-but-unreferenced entities, that'd be even better. I'm not aware that any such tool exists, though, and I don't have much experience with `phd` beyond basic manual builds.

jmikola: Tracking in [PHP-1545](https://jira.mongodb.org/browse/PHP-1545).

Girgias: We do have some, it wasn't up to scratch but I'm currently going through all our scripts and remove outdated ones, and more the ones still relevant into a folder: https://github.com/php/doc-base/pull/30

But I think removing the entities should be done at a later stage as this breaks translation builds until they remove the docs (e.g. the French one), moreover, a larger cleanup is needed so doing this all in one go is probably better. :-)

jmikola: Ah, I forgot to consider that other translations may still be using these entities. In that case, I'll close the JIRA ticket I just created and leave this for the PHP docs team to clean up at a later date (when we can be sure no translations are still using them). Thanks!

jmikola: Clarification: the concern for translations only applies to `global.ent`, correct? I assume everything under `language-snippets.ent` can still be removed since those are specific to the `doc-en` repo.

Girgias: I'm not exactly sure of how the build system exactly works, but translation might pull the english version of the language-snippers as a back-up/

jmikola (approved):

Girgias:
I've merge what was decided should go away, leaving FrontbaseSQL until further notice.

And I just realized that none of the mysqlnd plugins have had a release since 2013 and are unmaintained, so adding them here.

afilina:
@Girgias Look like besides the merge conflict, this is mergeable. Would you like me to resolve those conflicts for you?

Girgias:
> 
> 
> @Girgias Look like besides the merge conflict, this is mergeable. Would you like me to resolve those conflicts for you?

I haven't heard back from FrontbaseSQL so I'm not sure about this one, but yes rebasing and possibly merging the removal of  MySQLnd plugin docs might be useful

cmb69:
> I haven't heard back from FrontbaseSQL so I'm not sure about this one, […]

AFAIK, there is no PECL package, so we may consider this private code, and as such remove the documentation, in my opinion.

Girgias:
> 
> 
> > I haven't heard back from FrontbaseSQL so I'm not sure about this one, […]
> 
> AFAIK, there is no PECL package, so we may consider this private code, and as such remove the documentation, in my opinion.

I'll rebase and remove it then. :-)

PR #29, opened 2020-01-30
Add documentation for ReflectionProperty::getDefaultValue/hasDefaultValue

beberlei:
Its a bit early for this, please don't merge yet, but this related to https://github.com/php/php-src/commit/7a52bf21d358567980bc20a95cef730bc9e9f00d

PR #613, opened 2021-05-21
Add link to spec

shaedrich:


reference/datetime/dateperiod/construct.xml
```diff
@@ -73,7 +73,8 @@
      <term><parameter>isostr</parameter></term>
      <listitem>
       <para>
-       An ISO 8601 repeating interval specification.
+       An <link
+        xlink:href="&url.iso-8601.Repeating_intervals;">ISO 8601 repeating interval specification</link>.
```
cmb69, line 77: These entities  are not magically defined, but need to be defined in https://github.com/php/doc-base/blob/master/entities/global.ent before they can be used. :)

shaedrich: Thanks for that hint! :)

Added in [#29](https://github.com/php/doc-base/pull/29)

saundefined (approved):

shaedrich:
@cmb69 @saundefined I don't have write access to this repository so one of you guys has to merge it instead, I guess.

saundefined:
@shaedrich, thank you!

shaedrich:
You're welcome!

PR #616, opened 2021-05-22
Fix parameter names for array functions

kalehmann:
After  #614 and #615 I realized, that the documentations uses different parameter names than the actual implementation for many functions. This is a huge problem with the [named arguments][1] from PHP8.

I have hacked together a script, that parses the parses the xml files of the documentation and compares the parameter names to the actual ones using reflection. This script is available in [this gist][2].

The following name differences were detected by the script for function under `reference/array`. As this table is **automatically generated**, some entries may be **wrong** or **missing** and all entries must be manually reviewed.

Function                                |Expected parameter name                 |Actual parameter name                   
----------------------------------------|----------------------------------------|----------------------------------------
array_diff_ukey                         |arrays                                  |rest                                    
array_diff_ukey                         |arrays                                  |rest                                    
array_diff_ukey                         |arrays                                  |rest                                    
array_intersect_uassoc                  |arrays                                  |rest                                    
array_intersect_uassoc                  |arrays                                  |rest                                    
array_intersect_uassoc                  |arrays                                  |rest                                    
array_intersect_ukey                    |arrays                                  |rest                                    
array_intersect_ukey                    |arrays                                  |rest                                    
array_intersect_ukey                    |arrays                                  |rest                                    
array_multisort                         |array1                                  |array                                   
array_multisort                         |array1_sort_order                       |rest                                    
array_multisort                         |array1                                  |array                                   
array_multisort                         |array1_sort_order                       |rest                                    
array_multisort                         |array1                                  |array                                   
array_multisort                         |array1_sort_order                       |rest                                    
array_udiff_assoc                       |arrays                                  |rest                                    
array_udiff_assoc                       |arrays                                  |rest                                    
array_udiff_assoc                       |arrays                                  |rest                                    
array_udiff_uassoc                      |arrays                                  |rest                                    
array_udiff_uassoc                      |arrays                                  |rest                                    
array_udiff_uassoc                      |arrays                                  |rest                                    
array_udiff                             |arrays                                  |rest                                    
array_udiff                             |arrays                                  |rest                                    
array_udiff                             |arrays                                  |rest                                    
array_uintersect_assoc                  |arrays                                  |rest                                    
array_uintersect_assoc                  |arrays                                  |rest                                    
array_uintersect_assoc                  |arrays                                  |rest                                    
array_uintersect_uassoc                 |array1                                  |array                                   
array_uintersect_uassoc                 |arrays                                  |rest                                    
array_uintersect_uassoc                 |array1                                  |array                                   
[...]

PR #617, opened 2021-05-22
Fix parameter names of string comparison functions

kalehmann:
Same methods as in #616 were used to find mismatches in parameter names for string functions.


cmb69:
Oh, I think this is already covered by c44475e1fafcbee203ed4935a6d5d7a01379fcdc. Could you please rebase onto master?

kalehmann:
Yes, you are right. Didn't see this

PR #618, opened 2021-05-22
Fix the examples for character_set_name

kamil-tekiela:
The example had incorrect return value. It does not return current collation, but the charset. Fixed the example, which now set the charset before reading to show the behaviour more clearly.

### DIFF
--- a/reference/mysqli/mysqli/character-set-name.xml
+++ b/reference/mysqli/mysqli/character-set-name.xml
@@ -4,7 +4,7 @@
  <refnamediv>
   <refname>mysqli::character_set_name</refname>
   <refname>mysqli_character_set_name</refname>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! This looks good to me, but maybe we can fix the confusing description right away. What's correct: "Returns the default character set for the database connection" or "Returns the current character set for the database connection." Apparently the latter, in which case the return value section should also be fixed.

kamil-tekiela:
@cmb69 Heh, good question. I stumbled on this for a second too, but I decided to leave it as it was. In hindsight, I should have changed it too. It is confusing. 
Default charset is what is used initially. For some databases it is latin1 and for others it might be utf8mb4. This method does not change the default charset! It only changes the connection charset. So I think we should use the latter. Let me do that now. Thanks!

PR #619, opened 2021-05-22
Updating mysqli: mysqli_close

kamil-tekiela:
This page was missing its own example. I have also replaced the old shared note, with a better description. The link was not particularly useful.

### DIFF
--- a/reference/mysqli/mysqli/close.xml
+++ b/reference/mysqli/mysqli/close.xml
@@ -50,9 +50,45 @@
 
  <refsect1 role="examples">
   &reftitle.examples;
-  <para>
-   See <function>mysqli_connect</function>.
-  </para>
+  <example>
+   <title><methodname>mysqli::close</methodname> example</title>
+   <para>&style.oop;</para>
+   <programlisting role="php">
+<![CDATA[
+<?php
+
+mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
+$mysqli = new mysqli("localhost", "my_user", "my_password", "world");
+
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli/close.xml
```diff
@@ -22,7 +22,14 @@
...
    Closes a previously opened database connection.
   </para>
-  &mysql.close.connections.result.sets;
+  <para>
+   Open non-persistent MySQL connections and result sets are automatically
+   destroyed when a PHP script finishes its execution. Explicitly closing open 
+   connections and freeing result sets is optional. However, it's a good idea
+   to close the connection as soon as the script finishes performing all of its
+   database operations, if it still has a lot of processing to do after
+   getting the results.
```
cmb69, line 31: I think we should not talk about finishing a script, but rather about the objects being destroyed.

kamil-tekiela: Ok, can you suggest how to word it better?

kamil-tekiela: How about this?

>Open non-persistent MySQL connections and result sets are automatically destroyed along with their objects.

cmb69: Yes, that's good. Or maybe explicitly refer to "close", e.g. "Open non-persistent MySQL connections and result sets are automatically closed when their objects are destroyed."

cmb69:
Frankly, this looks like suggesting general optimizations, which apply to a lot of objects (e.g. `GdImage`s), but that appears to be quite fitting to ext/mysqli. :)

Thank you for the PR!

PR #620, opened 2021-05-22
Updating mysqli: connect_error & connect_errno

kamil-tekiela:
- Removed unhelpful notes
 - Replaced `die()` with `error_log()` - if default error reporting is disabled then the errors should probably be logged to a file
 - Removed mention of `mysqli_connect`. The connection error can come from all connection attempts except `change_user`
 - Added explanatory comments in the code examples


cmb69:
Thank you!

PR #621, opened 2021-05-22
Doc Bug #81073 Fixed example titles in PDOStatement::execute

kamil-tekiela:
Similar to what was proposed in the bug report but used the word "positional" for question mark placeholder values. Fixed the third title too.


cmb69:
Thank you!

PR #622, opened 2021-05-22
Update get_headers function

shyim:
This change is happend with https://github.com/php/php-src/commit/76e4bf3068bc7f92a2ece604344a92fd4d6c683d#diff-96c05a0551ccbeca3ff610c111f88b54a0d727888cf353f477ea8dc951a18637R1442
and isn't documented somewhere 🙈


cmb69:
Thank you! Note that I've updated the complete signature right away.

PR #623, opened 2021-05-23
Generate ext/gd methodsynopses based on stubs

kocsismate:
This contains lots of manual changes due to the resource to object conversion and the nullable parameter changes


cmb69:
Thank you!

PR #628, opened 2021-05-24
fpassthru error description (bug 81067)

saundefined:
Related bug [81067](https://bugs.php.net/bug.php?id=81067).

### DIFF
--- a/reference/filesystem/functions/fpassthru.xml
+++ b/reference/filesystem/functions/fpassthru.xml
@@ -70,7 +70,6 @@
      </row>
     </thead>
     <tbody>
-     &gd.changelog.image-param;
      <row>
       <entry>8.0.0</entry>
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR! I'm not sure if we should document TypeErrors for individual functions, though. I mean passing invalid types is supposed to always throw a TypeError as of PHP 8.0.0. Resources are somewhat special, but still. @Girgias, what do you think?

Girgias:
I'm not really keen on this, the docs do mention that this is invoking undefined behaviour (see https://www.php.net/manual/en/functions.internal.php). I think just the return section changes are sufficient.

saundefined:
Thank you! Fixed

cmb69 (approved):
Thanks!

PR #629, opened 2021-05-24
Update `spl_autoload_register` parameter section.

thinkverse:
This PR might not be necessary but noticed `spl_autoload_register` didn't have a signature for its callback parameter, so I decided to add it.

Choose to base it on the signature of `spl_autoload` which is the default method called if the callback is *null*. Hence the `string $class_name` parameter and `void` return type.

Feel this will be quicker for end-users to see the signature here instead of scrolling down to the examples section to see the implementation.


cmb69:
Nice one, thank you!

PR #632, opened 2021-05-27
Remove obsolete opcache.fast_shutdown config recommendation

anaggh:
`opcache.fast_shutdown` was removed in PHP 7.2


cmb69:
Thank you for the PR! However, the docs are currently for PHP 7.0.0 - 8.0.x, and prior to PHP 7.2.0 this setting makes apparently sense. Maybe better to leave the setting, but add a comment that it is unnecessary as of PHP 7.2.0?

anaggh:
Sorry, I thought 7.2.x was EOL - https://www.php.net/supported-versions.php

There is already a User Contributed Comment stating the same - https://www.php.net/manual/en/opcache.installation.php

Closing this.

PR #635, opened 2021-05-28
Fix function description

saundefined:
Seems that `not` is missed.

The previous description was:
```
If only <parameter>res_x</parameter> is given, the horizontal and vertical resolution  are set to this value.
```


cmb69 (approved):
Ah, right, thanks!

PR #636, opened 2021-05-28
Update stream_bucket_make_writeable return types

bytestream:
The current documentation states `stream_bucket_make_writeable` returns `object`. 

This is incorrect, it returns `object|null`: 
* https://github.com/php/php-src/blob/dd83ced6bd600e0dd5b09cf26fe4302344521e97/ext/standard/user_filters.c#L382
* https://github.com/php/php-src/blob/dd83ced6bd600e0dd5b09cf26fe4302344521e97/ext/standard/basic_functions.stub.php#L1456


cmb69 (approved):
Thank you!

Issue #633, opened 2021-05-27
DateInterval: weeks and days can be combined since 8.0


gharlan:
From manual page: https://php.net/dateinterval.construct

---

> W | weeks. These get converted into days, so can not be combined with D.

The "so can not be combined with D" part is no longer true since php 8.0: https://3v4l.org/RNqK8
https://github.com/derickr/timelib/pull/85

kamil-tekiela:
Thanks for the issue. We also encourage making pull requests. There's a lot of things pending to be updated in the manual and we appreciate any help.

PR #637, opened 2021-05-28
DateInterval: weeks and days can be combined since 8.0.0

saundefined:
Related issue #633

### DIFF
--- a/reference/datetime/dateinterval/construct.xml
+++ b/reference/datetime/dateinterval/construct.xml
@@ -128,6 +128,28 @@
   </para>
  </refsect1>
 
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Maybe also add to a changelog?

saundefined:
@kamil-tekiela, sure, added. Thank you!

reference/datetime/dateinterval/construct.xml
```diff
@@ -65,8 +65,8 @@
           <row>
            <entry><literal>W</literal></entry>
            <entry>
-            weeks. These get converted into days,
-            so can not be combined with <literal>D</literal>.
+            weeks. Converted into days.
```
nikic, line 68: Does this "Converted into days" comment still make sense? I don't get what it's trying to say.

kamil-tekiela: Yes, apparently this is how it works. If you use `P1W` then it gets converted into 7 days. If you use `P1W2D` then the days will take precedence and overwrite the value from weeks in PHP < 8

cmb69: See https://3v4l.org/RNqK8; the object doesn't store weeks, but only days internally.

Issue #640, opened 2021-05-29
SplDoublyLinkedList::__construct() does not exist


porthd:
From manual page: https://php.net/class.spldoublylinkedlist

---
The documentation describe a constructor.
https://www.php.net/manual/de/spldoublylinkedlist.construct.php
https://www.php.net/manual/en/spldoublylinkedlist.construct.php

I think, there is no constructor. (or misunderstand I something?)

I tried to extend 'SplDoublyLinkedList' in an own class and got the unexpected fatal error (see comment in code) 

```
<?php
        
        class Liste extends SplDoublyLinkedList 
        {
            public function __construct() {
                parent::__construct(); //<b>Fatal error</b>:  Uncaught Error: Cannot call constructor in [...][...]:6
                $this->push('klaus');
                
            }
            
            public function liste() {
                $this->push('herbert');
            }
        }

$hallo = new Liste();
$hallo->push('padina');
$hallo->push('doof');
$hallo->liste();
foreach($hallo as $key=>$item) {
    echo($key.' => '.$item."\n");
```

cmb69:
Right, [`SplDoublyLinkedList` has no contructor](https://3v4l.org/J3aER). The page should be removed.

PR #641, opened 2021-05-31
Remove spldoublylinkedlist construct

sy-records:
close https://github.com/php/doc-en/issues/640


cmb69:
Thak you! But see https://github.com/php/doc-en/pull/641/checks?check_run_id=2706906842#step:5:30. The inclusion of the constructorsynopsis on the class page should be commented out.

nikic:
@cmb69 Hrm, why does that not cause a build failure?

cmb69:
Oh, on a closer look I am surprised by that. Usually, there should be an [`<xi:fallback>`](https://github.com/php/doc-base/blob/f4eff8d4b5b2afb8220027194a2ba4811903f51a/RFC/skeletons/classname.xml#L65-L67), in which case the error would be bogus (apparently, a glitch in libxml2 or the configure script), but in this case there isn't:
https://github.com/php/doc-en/blob/3263a6218e51bab67de19e1cc4c772aca35a523f/reference/spl/spldoublylinkedlist.xml#L87

sy-records:
> @cmb69 Hrm, why does that not cause a build failure?

Stops only when a fatal error is encountered

https://github.com/php/doc-base/blob/master/configure.php#L223-L227

cmb69:
Ah, thanks!

cmb69 (approved):

PR #643, opened 2021-05-31
Add a note to strtok

colshrapnel:
There are some nuances, explained on the strtok man page, https://man7.org/linux/man-pages/man3/strtok.3.html namely starting, ending and duplicated tokens are ignored.

### DIFF
--- a/reference/strings/functions/strtok.xml
+++ b/reference/strings/functions/strtok.xml
@@ -42,7 +42,7 @@
     First, a sequence of two or more contiguous <parameter>token</parameter> characters in the parsed string 
     is considered to be a single delimiter. 
     Also, a <parameter>token</parameter> situated at the start or end of the string is ignored.
-    That is, if you have a string like ";aaa;;bbb;", successive calls to <function>strtok</function> 
+    That is, if you have a string like <literal>";aaa;;bbb;"</literal>, successive calls to <function>strtok</function> 
     with ";" as a <parameter>token</parameter> would return strings "aaa" and "bbb", and then &false;,
     hence effectively returning only two elements, whereas <function>explode</function> would return an array of 5 elements.
    </para>
@@ -122,10 +122,9 @@ var_dump($first_token, $second_token);
   </para>
   <para>
    <example>
-    <title>the difference between <function>strtok</function> and <function>exolode</function></title>
+    <title>The difference between <function>strtok</function> and <function>explode</function></title>
     <programlisting role="php">
 <![CDATA[
-<?php
 <?php
 $string = ";aaa;;bbb;";
 
--- a/reference/strings/functions/strtok.xml
+++ b/reference/strings/functions/strtok.xml
@@ -36,6 +36,17 @@
    tokenized when any one of the characters in the <parameter>token</parameter> argument is
    found.
   </para>
+  <note>
+   <para>
+    This function behaves slightly different from you may probaly expect being faimiar with <function>explode</function>. 
+    First, a sequence of two or more contiguous <parameter>token</parameter> characters in the parsed string 
+    is considered to be a single delimiter. 
+    Also, a <parameter>token</parameter> situated at the start or end of the string is ignored.
+    That is, if you have a string like ";aaa;;bbb;", successive calls to <function>strtok</function> 
+    with ";" as a <parameter>token</parameter> would return strings "aaa" and "bbb", and then &false;,
+    hence effectively returning only two elements, whereas <function>explode</function> would return an array of 5 elements.
+   </para>
+  </note>
  </refsect1>
 
  <refsect1 role="parameters">
@@ -85,7 +96,6 @@ while ($tok !== false) {
     echo "Word=$tok<br />";
     $tok = strtok(" \n\t");
 }
-?>
 ]]>
     </programlisting>
    </example>
@@ -99,7 +109,6 @@ while ($tok !== false) {
 $first_token  = strtok('/something', '/');
 $second_token = strtok('/');
 var_dump($first_token, $second_token);
-?>
 ]]>
     </programlisting>
     &example.outputs;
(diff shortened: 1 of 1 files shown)


reference/strings/functions/strtok.xml
```diff
@@ -36,6 +36,17 @@
    tokenized when any one of the characters in the <parameter>token</parameter> argument is
    found.
   </para>
+  <note>
+   <para>
+    This function behaves slightly different from you may probaly expect being faimiar with <function>explode</function>.
```
kamil-tekiela, line 41: According to the [contributing guidelines](http://doc.php.net/tutorial/) we don't use "you" in the documentation. The idea is to maintain the manual in a very formal way. So maybe replace this with

> There are two main differences between this function and <function>explode</function>

Then you can list the differences below.

reference/strings/functions/strtok.xml
```diff
@@ -36,6 +36,17 @@
    tokenized when any one of the characters in the <parameter>token</parameter> argument is
    found.
   </para>
+  <note>
+   <para>
+    This function behaves slightly different from you may probaly expect being faimiar with <function>explode</function>. 
+    First, a sequence of two or more contiguous <parameter>token</parameter> characters in the parsed string 
+    is considered to be a single delimiter. 
+    Also, a <parameter>token</parameter> situated at the start or end of the string is ignored.
+    That is, if you have a string like ";aaa;;bbb;", successive calls to <function>strtok</function>
```
kamil-tekiela, line 45:
```suggestion
    That is, if you have a string like <literal>";aaa;;bbb;"</literal>, successive calls to <function>strtok</function> 
```

reference/strings/functions/strtok.xml
```diff
@@ -36,6 +36,17 @@
...
   </para>
+  <note>
+   <para>
+    This function behaves slightly different from you may probaly expect being faimiar with <function>explode</function>. 
+    First, a sequence of two or more contiguous <parameter>token</parameter> characters in the parsed string 
+    is considered to be a single delimiter. 
+    Also, a <parameter>token</parameter> situated at the start or end of the string is ignored.
+    That is, if you have a string like ";aaa;;bbb;", successive calls to <function>strtok</function> 
+    with ";" as a <parameter>token</parameter> would return strings "aaa" and "bbb", and then &false;,
+    hence effectively returning only two elements, whereas <function>explode</function> would return an array of 5 elements.
```
kamil-tekiela, line 47: This sentence is too complex IMO. You have big words like "hence" and "whereas", but the sentence could be split up. 

Other points:

- We have a line limit of 80 characters. 
- Starting sentences with "First,", "Also,", "That is," isn't very formal and I would avoid using such constructs too much. 
- I see why you would want to compare it to `explode` but these functions have more differences than the two you listed. One returns an array and the other is a weird function with memory that returns a different string each time you call it. 
- The note might look better situated near its example.

reference/strings/functions/strtok.xml
```diff
@@ -85,7 +96,6 @@ while ($tok !== false) {
     echo "Word=$tok<br />";
     $tok = strtok(" \n\t");
 }
-?>
```
kamil-tekiela, line 88: We try to not make too many changes to formatting or style. While I agree that the closing statement is redundant, most of the documentation still uses it. If it doesn't harm anyone then let's keep it here.

reference/strings/functions/strtok.xml
```diff
@@ -99,14 +109,43 @@ while ($tok !== false) {
...
 <![CDATA[
     string(9) "something"
     bool(false)
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title>the difference between <function>strtok</function> and <function>exolode</function></title>
```
kamil-tekiela, line 125:
```suggestion
    <title>The difference between <function>strtok</function> and <function>explode</function></title>
```

reference/strings/functions/strtok.xml
```diff
@@ -99,14 +109,43 @@ while ($tok !== false) {
...
+    <programlisting role="php">
+<![CDATA[
+<?php
+<?php
```
kamil-tekiela, line 129:
```suggestion
```

kamil-tekiela (changes requested):
Good idea to add a note like that, but I think it needs some improvement.

colshrapnel:
Thank you.
Tried my best to implement everything you mentioned.

reference/strings/functions/strtok.xml
```diff
@@ -36,6 +36,20 @@
...
+    This function behaves slightly different from one may probaly expect being faimiar with
+    <function>explode</function>. 
+    First, a sequence of two or more contiguous <parameter>token</parameter> characters 
+    in the parsed string is considered to be a single delimiter. 
+    Also, a <parameter>token</parameter> situated at the start or end of the string is ignored.
+    For example, if you have a string like <literal>";aaa;;bbb;"</literal>, successive calls to 
+    <function>strtok</function> with ";" as a <parameter>token</parameter> would return strings
+    "aaa" and "bbb", and then &false;.
+    As a result, it will find only two elements, while <function>explode</function> would return
+    an array of 5 elements.
```
kamil-tekiela, line 50:
```suggestion
    This function behaves slightly different from what one may expect being
    familiar with <function>explode</function>. 
    First, a sequence of two or more contiguous <parameter>token</parameter>
    characters in the parsed string is considered to be a single delimiter. 
    Also, a <parameter>token</parameter> situated at the start or end of the
    string is ignored.
    For example, if a string <literal>";aaa;;bbb;"</literal> is used, successive
    calls to <function>strtok</function> with <literal>";"</literal> as a
    <parameter>token</parameter> would return strings
    "aaa" and "bbb", and then &false;.
    As a result, the string will be split into only two elements, while
    <literal>explode(";", $string)</literal> would return an array of 5 elements.
```

kamil-tekiela (approved):

cmb69:
Thank you!

PR #646, opened 2021-06-01
slight improvement to the language

powellblyth:
Just makes the language feel a little more natural


kamil-tekiela (approved):
Thanks. I was a little bit thorn on `output`/`outputted` since both are ok here. However, I think your suggestion is ok, as the word `output` is more commonly used, it seems.

PR #647, opened 2021-06-02
Fix Extension membership, and rebuild extension list file from automated script

Girgias:
Messed up with git... so this is the same as #644

Content of the previous PR:

> doc-base has a script to generate the extensions.xml file, so it makes to use and check for it in CI.

> To do this I've added and/or adjusted the extension membership of all currently documented extensions, except pht as it is not available on PECL, @tpunt  is there any reason for it not being on PECL?

### DIFF
--- a/appendices/extensions.xml
+++ b/appendices/extensions.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
 <!--
   DO NOT TRANSLATE THIS FILE! All the content that is displayed
   on the extension categorization page in your translated manual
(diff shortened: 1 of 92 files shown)


reference/hrtime/book.xml
```diff
@@ -2,6 +2,7 @@
 <!-- $Revision$ -->
 
 <book xml:id="book.hrtime" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <?phpdoc extension-membership="core" ?>
```
cmb69, line 5: This is not a core extension (although the is a core function named `hrtime()`).

appendices/extensions.xml
```diff
@@ -196,24 +204,27 @@
     <listitem><para><xref linkend="book.dir"/></para></listitem>
     <listitem><para><xref linkend="book.errorfunc"/></para></listitem>
     <listitem><para><xref linkend="book.exec"/></para></listitem>
+    <listitem><para><xref linkend="book.ffi"/></para></listitem>
     <listitem><para><xref linkend="book.filesystem"/></para></listitem>
-    <listitem><para><xref linkend="book.filter"/></para></listitem>
+    <listitem><para><xref linkend="book.fpm"/></para></listitem>
     <listitem><para><xref linkend="book.funchand"/></para></listitem>
     <listitem><para><xref linkend="book.hash"/></para></listitem>
+    <listitem><para><xref linkend="book.hrtime"/></para></listitem>
```
cmb69, line 212: see below

cmb69 (commented):
Very nice! Looks good, except for the hrtime issue.

PR #652, opened 2021-06-03
Ini settings: upload_max_filesize: Add note regarding post_max_size

AllenJB:
Currently the upload_max_filesize entry makes no reference to post_max_size, which creates a gotcha when increasing its value. (There is a similar note under post_max_size, but users who don't know about this aren't likely to be looking there)


cmb69:
Thank you!

PR #661, opened 2021-06-06
Remove legacy Windows docs

Girgias:
Trying to go through the installation sections of the doc and the Windows part is taking a bigger chunk, that's why I'm making it its own PR.

PR #662, opened 2021-06-06
Remove old install docs

Girgias:
This should cover the instal/ folder with #661


install/fpm/configuration.xml
```diff
@@ -882,7 +882,7 @@ php_admin_value[memory_limit] = 32M
      cannot be overridden with <function>ini_set</function>.
     </para>
     <para>
-     As of 5.3.3, PHP settings are also possible to be set in webserver.
+     PHP settings can also possible to be set in the webserver.
```
jhdxr, line 885: hmm, I think it should be `PHP settings can be set in the webserver.` if you look at grammar.

install/pecl.xml
```diff
@@ -332,7 +341,7 @@ $ pecl install extname-0.1
    Sometimes, using the <literal>pecl</literal> installer is not an option.
    This could be because you're behind a firewall, or it could be because the
    extension you want to install is not available as a PECL compatible
-   package, such as unreleased extensions from SVN.  If you need to build such
+   package, such as unreleased extensions from git or SVN.  If you need to build such
```
jhdxr, line 344: Git

jhdxr (changes requested):

install/fpm/configuration.xml
cmb69: Much better, but stills feelsnot quit right. Maybe "PHP settings can be set in the webserver configuration"?

afilina, line 885:
```suggestion
     PHP settings can be set in the webserver configuration.
```

afilina:
@Girgias The module identifier was renamed in PHP 8, so I resolved the merge conflict to `php_module` instead of `php8_module`. I also added the comment as a mergeable suggestion, so after this is done, the PR looks ready to go.

install/pecl.xml
```diff
@@ -331,7 +340,7 @@ $ pecl install extname-0.1
...
```
mumumu, line 343:
```suggestion
   package, such as unreleased extensions from git.  If you need to build such
```

Girgias:
> > I also added the comment as a mergeable suggestion, so after this is done, the PR looks ready to go.
> 
> +1. If no one is opposed to merge this PR for a while, I want to commit suggestions and process this PR.
> It is not good for PHP users, especially for newbies to leave this PR as it is.

Sorry forgot about this PR, but feel free to apply suggestions and commit. :-)

PR #666, opened 2021-06-07
Fix section issues

saundefined:
Some more fixed issues. Not sure about the following return values sections:

`snmp_set_valueretrieval` based on [source](https://github.com/php/php-src/blob/master/ext/snmp/snmp.c#L1450)

`snmp_read_mib` based on [source](https://github.com/php/php-src/blob/master/ext/snmp/snmp.c#L1480)

`snmp_set_enum_print` based on [source](https://github.com/php/php-src/blob/master/ext/snmp/snmp.c#L1480)

### DIFF
--- a/reference/snmp/functions/snmp-set-enum-print.xml
+++ b/reference/snmp/functions/snmp-set-enum-print.xml
@@ -32,13 +32,6 @@
    </varlistentry>
   </variablelist>
  </refsect1>
-
- <refsect1 role="returnvalues">
-  &reftitle.returnvalues;
-  <para>
-   Returns &true;.
-  </para>
- </refsect1>
    
  <refsect1 role="examples">
   &reftitle.examples;
--- a/reference/snmp/functions/snmp-set-enum-print.xml
+++ b/reference/snmp/functions/snmp-set-enum-print.xml
@@ -32,6 +32,13 @@
    </varlistentry>
   </variablelist>
  </refsect1>
+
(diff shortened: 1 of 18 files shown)


cmb69:
I'm not sure about these 3 functions either; the rest looks good to me.

reference/snmp/functions/snmp-set-valueretrieval.xml
```diff
@@ -55,6 +55,12 @@
   </para>
  </refsect1>
  
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &true;.
+  </para>
+ </refsect1>
```
Girgias, line 63: This used to return false with an E_WARNING prior to PHP 8.0.0

reference/snmp/functions/snmp-set-enum-print.xml
```diff
@@ -32,6 +32,13 @@
    </varlistentry>
   </variablelist>
  </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &true;.
```
Girgias, line 39: Ffs, this used to return false on a type error (https://heap.space/xref/PHP-7.4/ext/snmp/snmp.c?r=9690ded2#1572), so leave it as such.

Girgias (approved):
Looks good other then the nit on `snmp_set_valueretrieval()`

PR #668, opened 2021-06-07
Document Exception as implementing, not extending Throwable, + fix me…

bdsl:
…ssage param type.

We were getting an Error trying to pass null to construct an exception, and found that "" works but null doesn't.

Also correcting the inheritance keyword since Throwable is an interface.

This brings these docs more into line with the PHPStorm stubs.

### DIFF
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -15,7 +15,7 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-class Exception extends Throwable
+class Exception implements Throwable
 {
     protected $message = 'Unknown exception';   // exception message
     private   $string;                          // __toString cache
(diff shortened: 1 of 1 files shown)


bdsl:
This also brings this copy of the exception class for the page on extending exceptions more into line with the page on the class itself: https://www.php.net/manual/en/class.exception.php

cmb69:
Thank you!

bdsl:
You're very welcome @cmb69 . This is my first successful contribution to the PHP docs btw - I tried years ago before they were moved to git and then gave up. Out of interest how long does it normally take for the changes to be released and visible at https://www.php.net/manual/en/language.exceptions.extending.php now that they're merged to master?

Girgias:
Docs are rebuilt once a day, so you'll see it tomorrow.

bdsl:
Thanks George

bdsl:
Seen it.

PR #672, opened 2021-06-09
Clarify assert() behavior in PHP 8

wouterj:
As far as I understand, the signature and behavior of `assert()` has changed in PHP 7 and remained the same in PHP 8.

The docs didn't yet mention PHP 8, leaving you guess whether PHP 8 behaves the same as PHP 7. I think it's worth making it clearer.

### DIFF
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -14,7 +14,7 @@
    <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
    <methodparam choice="opt"><type>string</type><parameter>description</parameter></methodparam>
   </methodsynopsis>
-  <para>PHP 7</para>
+  <para>PHP 7 and higher</para>
   <methodsynopsis>
    <type>bool</type><methodname>assert</methodname>
    <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


Girgias:
As we've removed most of the PHP 5 docs, it would make more sense to rewrite the docs such that no versions are mentioned as you can drop PHP 5

wouterj:
Hmm, I would have loved to help you with this, but I'm afraid I won't have the time (and knowledge about php.net practices) for it. The "Traditional assertions" logic seems to be removed from PHP 8, but lots of the related options of `assert_options()` are still working in sometimes unexpected ways. This makes it very difficult for me to judge how this should be documented if PHP 5 should be removed.

For instance: "Traditional assertions" are PHP 5, and supported in PHP 7 for compatibility reasons. PHP 8 only supports the "Expectations" (which were introduced in PHP 7). Most options, like `ASSERT_WARNING` and `assert.exception = 0` are there in PHP 7 and 8 for compatibility reasons with PHP 5 if I'm correct. So I'm unsure what PHP 7+ docs should contain.

Btw, there is also some weird behavior when mixing the assert options with the new expectations, e.g. `ASSERT_BAIL` always results in an "uncaught" exception while being surrounded by a correct `try...catch` statement: https://3v4l.org/8OIsK

PR #679, opened 2021-06-10
Update FILTER_NULL_ON_FAILURE usage

flow-control:
Hey there :vulcan_salute: 

I found the following comment in the docs to be true https://www.php.net/manual/en/filter.filters.flags.php#125956
The `FILTER_NULL_ON_FAILURE` flag can not only be used with `FILTER_VALIDATE_BOOL`/`FILTER_VALIDATE_BOOLEAN` but also with other validating filters.

This PR updates the docs accordingly, see https://3v4l.org/raIiu

Kind regards

### DIFF
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -606,16 +606,7 @@ filter.default_flags = 0
       <row>
        <entry><constant>FILTER_NULL_ON_FAILURE</constant></entry>
        <entry>
-        <constant>FILTER_VALIDATE_BOOLEAN</constant>,
-        <constant>FILTER_VALIDATE_BOOL</constant>,
-        <constant>FILTER_VALIDATE_INT</constant>,
-        <constant>FILTER_VALIDATE_FLOAT</constant>,
-        <constant>FILTER_VALIDATE_DOMAIN</constant>,
-        <constant>FILTER_VALIDATE_EMAIL</constant>,
-        <constant>FILTER_VALIDATE_IP</constant>,
-        <constant>FILTER_VALIDATE_MAC</constant>,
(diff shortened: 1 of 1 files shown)


reference/filter/filters.xml
```diff
@@ -597,10 +597,18 @@ filter.default_flags = 0
...
-        <constant>FILTER_VALIDATE_BOOL</constant>
+        <constant>FILTER_VALIDATE_BOOL</constant>,
+        <constant>FILTER_VALIDATE_INT</constant>,
+        <constant>FILTER_VALIDATE_FLOAT</constant>,
+        <constant>FILTER_VALIDATE_DOMAIN</constant>,
+        <constant>FILTER_VALIDATE_EMAIL</constant>,
+        <constant>FILTER_VALIDATE_IP</constant>,
+        <constant>FILTER_VALIDATE_MAC</constant>,
+        <constant>FILTER_VALIDATE_REGEXP</constant>,
+        <constant>FILTER_VALIDATE_URL</constant>
```
cmb69, line 608: Since FILTER_NULL_ON_FAILURE can be used with any validation failure, maybe we should just say so, instead of listing all validation filters?

flow-control: What do you think about `any <constant>FILTER_VALIDATE_*</constant>` ?

cmb69: I think we're using such pseudo constant elsewhere in the manual as well, so this is fine for me. Maybe add a link to the relevant section, i.e. `<link linkend="filter.filters.validate"><constant>FILTER_VALIDATE_*</constant></link>`?

flow-control: thanks for the hint with the link, I updated the pr accordingly

PR #681, opened 2021-06-11
Clarify the order of sort functions in their description

jhdxr:
we have several sort functions, but it's not clear if they sort in ascending or descending order. I updated them accordingly.

ps. Do you prefer a more layman definition for them? e.g. `from lowest to highest` instead of `ascending order`.


cmb69:
Ah, thank you, very good! I would actually suggest to change the descriptions which state "from lowest to highest" and vice versa to use "ascending order" and "descending order", respectively, as well.

cmb69:
Thank you!

PR #686, opened 2021-06-13
Update array sorting docs

Girgias:
These always return true, even in PHP 7
Clarify sorting is unstable prior to PHP 8.0
Normalize the docs between the functions

### DIFF
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -137,7 +137,8 @@ from one file system to another.</para></note>'>
 
 <!ENTITY note.sort-unstable '<note xmlns="http://docbook.org/ns/docbook">
  <para>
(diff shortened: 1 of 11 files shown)


fulldecent:
This PR touches return value documentation of true and false values. So it collides with https://github.com/php/doc-en/pull/858

Girgias:
> 
> 
> This PR touches return value documentation of true and false values. So it collides with #858

What? You don't modify these files in your PR.

fulldecent:
I modify the same files in my PR. My PR is much wider and shallower.

But just wanted to note upcoming conflict.

salathe:
Thanks! Applied as d1df62f.

PR #689, opened 2021-06-14
Fold in comments from is_callable() page

Crell:
I'm not sure this is structured properly.  I can restructure to some other format if someone tells me what the format should be. :-)

### DIFF
--- a/reference/var/functions/is-callable.xml
+++ b/reference/var/functions/is-callable.xml
@@ -153,18 +153,18 @@ bool(false)
 
  <refsect1 type="notes">
   <simplelist>
-   <member><para>
+   <member>
     An object is always callable if it implements <function>__invoke()</function>,
     and that method is visible in the current scope.
-   </para></member>
-   <member><para>A class name is callable if it implements <function>__callStatic()</function></para></member>
-   <member><para>
+   </member>
+   <member>A class name is callable if it implements <function>__callStatic()</function></member>
+   <member>
     If an object implements <function>__call()</function>, then this function will
     return true for any method on that object, even if the method is not defined.
-   </para></member>
-   <member><para>
+   </member>
+   <member>
     This function may trigger autoloading if called with the name of a class.
-   </para></member>
+   </member>
   </simplelist>
  </refsect1>
 
--- a/reference/var/functions/is-callable.xml
+++ b/reference/var/functions/is-callable.xml
@@ -151,7 +151,7 @@ bool(false)
   </para>   
  </refsect1>
 
- <refsect1 type="notes">
+ <refsect1 role="notes">
   <simplelist>
    <member>
     An object is always callable if it implements <function>__invoke()</function>,
--- a/reference/var/functions/is-callable.xml
+++ b/reference/var/functions/is-callable.xml
@@ -4,7 +4,7 @@
  <refnamediv>
   <refname>is_callable</refname>
   <refpurpose>
-   Verify that the contents of a variable can be called as a function from the current scope.
+   Verify that a value can be called as a function from the current scope.
   </refpurpose>
  </refnamediv>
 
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Maybe add a note on visibility of the `__invoke` method?

reference/var/functions/is-callable.xml
```diff
@@ -69,6 +69,20 @@
   </para>
  </refsect1>
 
+ <refsect1>
+  <para>An object is always callable if it implements <function>__invoke()</function>.</para>
+  <para>A class name is callable if it implements <function>__callStatic()</function></para>
+  <para>
+   If an object implements <function>__call()</function>, then this function will
+   return true for any method on that object, even if the method is not defined.
+  </para>
```
cmb69, line 78: I think this can go into a single para.

reference/var/functions/is-callable.xml
```diff
@@ -4,7 +4,7 @@
  <refnamediv>
   <refname>is_callable</refname>
   <refpurpose>
-   Verify that the contents of a variable can be called as a function
+   Verify that the contents of a variable can be called as a function from the current scope.
```
cmb69, line 7:
```suggestion
   Verify that a value can be called as a function from the current scope.
```

reference/var/functions/is-callable.xml
```diff
@@ -69,6 +69,20 @@
   </para>
  </refsect1>
 
+ <refsect1>
```
cmb69, line 72: The refsect1 should have a role. If role="note", it should be put below the examples. But maybe the info is better suited in the description above?

cmb69 (commented):
Thank you!

Crell:
Revised.

reference/var/functions/is-callable.xml
```diff
@@ -151,6 +151,23 @@ bool(false)
   </para>   
  </refsect1>
 
+ <refsect1 type="notes">
+  <simplelist>
+   <member><para>
+    An object is always callable if it implements <function>__invoke()</function>,
+    and that method is visible in the current scope.
+   </para></member>
```
cmb69, line 159: `<para>` is not allowed as child of `<member>`, so:
```suggestion
   <member>
    An object is always callable if it implements <function>__invoke()</function>,
    and that method is visible in the current scope.
   </member>
```

Crell: I thought it was the other way around, you had to include something inside member?

cmb69: You may confuse that with `<listitem>` which indeed needs a `<para>` or so.

reference/var/functions/is-callable.xml
```diff
@@ -151,6 +151,23 @@ bool(false)
   </para>   
  </refsect1>
 
+ <refsect1 type="notes">
```
cmb69, line 154:
```suggestion
 <refsect1 role="notes">
```

cmb69:
@girgias, what do you think about the layout of the notes section? It's rather different from what we usually use (i.e. multiple `<note>`s and such), but I think I like this better.

Girgias:
The layout might make more sense then what we currently have at times, so it's all green on my end. :-)

reference/var/functions/is-callable.xml
```diff
@@ -151,6 +151,23 @@ bool(false)
   </para>   
  </refsect1>
 
+ <refsect1 role="notes">
+  <simplelist>
+   <member>
+    An object is always callable if it implements <function>__invoke()</function>,
+    and that method is visible in the current scope.
+   </member>
+   <member>A class name is callable if it implements <function>__callStatic()</function></member>
```
salathe, line 160: `<function>__callStatic()</function>` should be `<link linkend="object.callstatic">__callStatic()</link>`

reference/var/functions/is-callable.xml
```diff
@@ -151,6 +151,23 @@ bool(false)
...
+   <member>
+    If an object implements <function>__call()</function>, then this function will
```
salathe, line 162: `<function>__call()</function>` should be `<link linkend="object.call">__call()</link>`

reference/var/functions/is-callable.xml
```diff
@@ -151,6 +151,23 @@ bool(false)
...
+    return true for any method on that object, even if the method is not defined.
```
salathe, line 163: `true` should be `&true;`

salathe (commented):
Some post-merge comments. Poor effort from the code reviewers on this one. 😛

cmb69:
Thank you, @salathe! Fixed as 83a17a7324c2597bd6385148abf19f76127229f5.

PR #690, opened 2021-06-15
Fix sodium_bin2hex description

saundefined:
Seems a little wrong


cmb69 (approved):
Good catch, thank you! Likely a copy & paste issue.

PR #691, opened 2021-06-15
Fix issues in mcrypt and datetime directories

saundefined:
I'm not sure about the `localtime()` function changes 🤔


derickr (approved):
LGTM

PR #700, opened 2021-06-18
Fix issues in mysqli directory

saundefined:
I'm not sure about the `note` tag.

Some functions have this note as plain text (e.g. [`mysqli_get_client_stats`](https://www.php.net/function.mysqli-get-client-stats)).

If it should be a note, I'll wrap this in other functions, 
if not — replace it as plain text in these functions.


kamil-tekiela:
>If it should be a note, I'll wrap this in other functions,

Yeah, it should be a note.

saundefined:
@kamil-tekiela, thank you! updated.

cmb69 (approved):
Thank you!

PR #704, opened 2021-06-19
Update return type for SessionHandlerInterface::gc()

othercorey:
The type is `int|false` but documented as `int|bool`.

https://github.com/php/php-src/blob/PHP-8.0.8/ext/session/session.stub.php#L83


kamil-tekiela (approved):

sy-records (approved):

othercorey:
fyi, I can't merge this.

kamil-tekiela:
Thanks for letting us know. Merged now.

PR #708, opened 2021-06-21
Update installation guide for macOS Monterey

saundefined:
Apple [removed](https://developer.apple.com/forums/thread/681907) build-in PHP in macOS Monterey

### DIFF
--- a/install/macos/bundled.xml
+++ b/install/macos/bundled.xml
@@ -3,7 +3,7 @@
 <sect1 xml:id="install.macosx.bundled" xmlns="http://docbook.org/ns/docbook">
  <title>Using the bundled PHP prior to macOS Monterey</title>
  <simpara>
-  PHP has bundled with macOS since macOS X (10.0.0) prior to macOS Monterey (12.0.0).
+  PHP is bundled with macOS since macOS X (10.0.0) prior to macOS Monterey (12.0.0).
   Enabling PHP with the default web server requires uncommenting a few lines in the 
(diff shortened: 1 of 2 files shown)


install/macos/bundled.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
 <sect1 xml:id="install.macosx.bundled" xmlns="http://docbook.org/ns/docbook">
- <title>Using the bundled PHP</title>
+ <title>Using the bundled PHP prior to macOS Monterey</title>
  <simpara>
-  PHP has come standard with Macs since Mac OS X version 10.0.0. Enabling PHP 
-  with the default web server requires uncommenting a few lines in the 
+  PHP has bundled with macOS since macOS X (10.0.0) prior to macOS Monterey (12.0.0).
```
cmb69, line 6:
```suggestion
  PHP is bundled with macOS since macOS X (10.0.0) prior to macOS Monterey (12.0.0).
```
same below

cmb69 (commented):
I don't know about macOS, but looks good to me (besides the minor grammar issue).

install/macos/bundled.xml
saundefined: Thanks! fixed

saundefined:
ACK, updated macOS to 12.0, there is no built-in PHP.

PR #709, opened 2021-06-21
Use a better example for using header() to set status code

TysonAndre:
This may be the first thing users see when searching how to set the
response code - most will not be familiar with differences between
http/1.0 and 1.1

- HTTP/1.0 does not support connection keepalive
- http_response_code(404) is a shorter way to set a status code with the default
  status message

https://www.php.net/manual/en/function.header.php#92305 also mentions
that the wrong protocol can cause issues with (older?) web servers

https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#History http 1.1 was officially released in 1997

### DIFF
--- a/reference/network/functions/header.xml
+++ b/reference/network/functions/header.xml
@@ -67,7 +67,12 @@ exit;
         <programlisting role="php">
 <![CDATA[
 <?php
-header("HTTP/1.0 404 Not Found");
+// This example illustrates the "HTTP/" special case
+// Better alternatives in typical use cases include:
+// 1. header($_SERVER["SERVER_PROTOCOL"] . " 404 Not Found");
+//    (to override http status messages for clients that are still using HTTP/1.0)
+// 2. http_response_code(404); (to use the default message)
+header("HTTP/1.1 404 Not Found");
 ?>
 ]]>
         </programlisting>


cmb69:
Thank you! Maybe also add `http_response_code()` to the see-also section at the bottom of the page?

> https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#History http 1.1 was officially released in 1997

Prior to PHP 8.0.0, PHP's http wrapper defaulted to HTTP/1.0, though. ;)

TysonAndre:
> Thank you! Maybe also add http_response_code() to the see-also section at the bottom of the page?

It's already there https://www.php.net/manual/en/function.header.php#refsect1-function.header-seealso

cmb69:
I must be blind!

salathe:
👎  on this change.  I'd rather see the code comments that were added to the example, become part of the description above the example.

TysonAndre:
> There are two special-case header calls.  The first is a header
> that starts with the string "<literal>HTTP/</literal>" (case is not
> significant), which will be used to figure out the HTTP status
> code to send.

The point of the example is is illustrating the special case - it's not as self-contained of an example if you have to explain SERVER_PROTOCOL always starts with HTTP/ https://www.php.net/manual/en/reserved.variables.server.php

Additionally, some legacy clients may need exact http protocols or benefit from status messages (`HTTP/1.1 404 SomeApplicationName requires HTTP POST`)

header() illustrates how header() is used, not how http_status_code() is used.

There may be a typical way to work in better alternative functions inside of the documentation - I'm just not familiar with writing style to do that

cmb69:
What about the following?
````.diff
 reference/network/functions/header.xml | 19 +++++++++++--------
 1 file changed, 11 insertions(+), 8 deletions(-)

diff --git a/reference/network/functions/header.xml b/reference/network/functions/header.xml
index 7e15f3d8f3..c2570aeb88 100644
--- a/reference/network/functions/header.xml
+++ b/reference/network/functions/header.xml
@@ -57,21 +57,24 @@ exit;
        There are two special-case header calls.  The first is a header
        that starts with the string "<literal>HTTP/</literal>" (case is not
        significant), which will be used to figure out the HTTP status
-       code to send. For example, if you have configured Apache to
+       code to send. For example, if Apache is configured to
        use a PHP script to handle requests for missing files (using
-       the <literal>ErrorDocument</literal> directive), you may want to
-       make sure that your script generates the proper status code.
+       the <literal>ErrorDocument</literal> directive),
+       the script should generate the proper status code.
+      </para>
+      <para>
+       The following script shows how this can be done.
+       To support old clients which do not implement HTTP/1.1,
+       instead of hard-coding the protocol,
+       <varname>$_SERVER["SERVER_PROTOCOL"]</varname> can be used.
+       And unless there is the need for a non-standard message,
+       using <function>http_response_code</function> is preferable.
       </para>
       <para>
        <informalexample>
         <programlisting role="php">
 <![CDATA[
 <?php
-// This example illustrates the "HTTP/" special case
-// Better alternatives in typical use cases include:
-// 1. header($_SERVER["SERVER_PROTOCOL"] . " 404 Not Found");
-//    (to override http status messages for clients that are still using HTTP/1.0)
-// 2. http_response_code(404); (to use the default message)
 header("HTTP/1.1 404 Not Found");
 ?>
 ]]>
````

TysonAndre:
> `+       instead of hard-coding the protocol,`

Maybe clarify as something along the lines of "Instead of hard-coding the protocol string (`'HTTP/1.1'`)," in case there's confusion and it's misunderstood as SERVER_PROTOCOL being `'1.1'`. Can that link to _SERVER?

PR #715, opened 2021-06-23
LDAP link identifier replaced with LDAP resource

saundefined:
Also, is there a changelog entry needed?


cmb69 (approved):
No changelog entry is needed, because resources are used at least since PHP 5.3. It may make sense, though, to declare and use an XML entity for this text.

PR #718, opened 2021-06-23
Fix example of datefmt_get_timezone_id

kamil-tekiela:
I am not sure what exactly CN was and I can't find out what the purpose of this example was, but I replaced it with a valid timezone name.


cmb69:
Thank you!

PR #719, opened 2021-06-23
Remove Magic quotes runtime

kamil-tekiela:
This PR removes information related to magicquotes which has been removed in PHP 7.0. 
I did _**not**_ change:
https://www.php.net/manual/en/function.parse-ini-file
https://www.php.net/manual/en/function.get-magic-quotes-gpc
https://www.php.net/manual/en/function.get-magic-quotes-runtime.php
https://www.php.net/manual/en/function.mysql-escape-string
https://www.php.net/manual/en/function.mysql-real-escape-string.php
https://www.php.net/security.variables

### DIFF
--- a/appendices/aliases.xml
+++ b/appendices/aliases.xml
@@ -222,11 +222,6 @@
       <entry><link linkend="ref.ldap">LDAP</link></entry>
      </row>
      <row>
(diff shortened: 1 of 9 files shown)


cmb69:
Thank you!

PR #721, opened 2021-06-24
Fix issues in phar directory

saundefined:
`Phar::createDefaultStub` methodsynopsis  based on [stubs](https://github.com/php/php-src/blob/PHP-8.0.0/ext/phar/phar_object.stub.php#L154).

### DIFF
--- a/reference/phar/PharFileInfo/compress.xml
+++ b/reference/phar/PharFileInfo/compress.xml
@@ -27,6 +27,22 @@
   </para>
  </refsect1>
 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>compression</parameter></term>
+     <listitem>
+      <para>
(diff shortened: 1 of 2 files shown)


reference/phar/PharFileInfo/compress.xml
```diff
@@ -27,6 +27,22 @@
...
 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>compression</parameter></term>
+     <listitem>
+      <para>
+       Compression must be <literal>Phar::GZ</literal> or <literal>Phar::BZ2</literal>.
```
cmb69, line 38: Maybe
```suggestion
       Compression must be <constant>Phar::GZ</constant> or <constant>Phar::BZ2</constant>.
```

cmb69 (approved):
Thank you!

reference/phar/PharFileInfo/compress.xml
saundefined: Sure, thank you!

PR #723, opened 2021-06-24
Fix issues in pcre directory

saundefined:
Parameters description get from [`preg_replace`](https://www.php.net/manual/ru/function.preg-replace.php).


cmb69 (commented):
I'm not sure about this. Duplicating this likely results in unsynchronized copies; introducing an entity reference might be overkill.

saundefined:
Hm.. Then maybe just add the parameters section, but instead of the parameter list 
write "see the `preg_replace` parameters for more information"?

cmb69:
I don't like that either. A while ago @Girgias moved the printf param description to language-snippets.ent. Maybe he has an opinion on this PR?

Girgias:
I think it might be better to reference the other page, we still do this to some extend with [sscanf](https://www.php.net/manual/en/function.sscanf.php), and the entity for sprintf formats are used around 6 times.

(Also the description is a bit out of dated since the e/ flag doesn't work as of PHP 8.0.0)

saundefined:
@Girgias, just something like that, or is it better to list each parameter with `Parameter is described in ...`?

<details>
  <summary>parameters</summary>
  
  ```xml
 <refsect1 role="parameters">
  &reftitle.parameters;
  <para>
   Parameters are described in the documentation for <function>preg_replace</function>.
  </para>
 </refsect1>
  ```
</details>

Girgias:
> 
> 
> @Girgias, just something like that, or is it better to list each parameter with `Parameter is described in ...`?
> parameters
> 
> ```
> <refsect1 role="parameters">
> &reftitle.parameters;
> <para>
>  Parameters are described in the documentation for <function>preg_replace</function>.
> </para>
> </refsect1>
> ```

I think like this should be fine.

PR #724, opened 2021-06-24
Remove unneeded note

kamil-tekiela:
`set_socket_blocking` was removed in PHP 7.0. `socket_set_blocking` is still an alias and is not deprecated, but since we don't document aliases in notes, I removed the whole note.


cmb69:
Thank you!

Issue #720, opened 2021-06-24
Clearer wording in second paragraph for ucfirst


shahovkit:
"The definition of a word is any string of characters that is immediately **after** any character listed"
not only after the character, but also the first word in the string

because ucwords('lorem-ipsum-dolor', '-') isn't lorem-Ipsum-Dolor, but Lorem-Ipsum-Dolor

From manual page: https://php.net/function.ucwords

---

tiffany-taylor:
This is explained in the first paragraph:
> Returns a string with the first character of each word in string capitalized

However, the wording in the second paragraph could be a little clearer, probably.

PR #726, opened 2021-06-24
Revise wording for ucwords

tiffany-taylor:
Revised the wording in the second paragraph of `ucwords`. I'm not entirely sure about the use of double quotes around "word", but I think the changes overall provide clearer meaning of intent.

Resolves https://github.com/php/doc-en/issues/720

### DIFF
--- a/reference/strings/functions/ucwords.xml
+++ b/reference/strings/functions/ucwords.xml
@@ -18,9 +18,10 @@
    <parameter>string</parameter> capitalized, if that character is alphabetic.
   </para>
   <para>
-   The definition of a word is any string of characters that is immediately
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you!

salathe:
I think this would be better off turning things around by detailing what characters make up a word (i.e. **not** `separators`).

> For this function, a word is any string of characters that are not listed in the `separators` parameter.

Nit: can we make the list of separators in the text here, be in the same order as the actual default value? 🙈

tiffany-taylor:
I was a little trigger happy. I'll have a look at applying those changes.

Issue #707, opened 2021-06-21
ClassName::class is magic constant


amritpalnagrame:
1. Every magic method starts with `__` and magic constants even ends with `__`, except `ClassName::class `i.e. specified as [magic constant in docs](https://www.php.net/manual/en/language.constants.magic.php).
2. Isn't this `ClassName` referring to any replaceable values followed by [scope resolution operator `::` ](https://www.php.net/manual/en/language.oop5.paamayim-nekudotayim.php) and [class keyword](https://www.php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class).

So, is it magic constant composed of keyword and operator or is it mistake in [Magic constants docs](https://www.php.net/manual/en/language.constants.magic.php)?

kamil-tekiela:
I agree calling `class` a constant is misleading, but it does fit the definition of a magic constant. To make things clearer I would keep it on the page for magic constants, but remove `ClassName`, so just leave it as `::class`, i.e. the operator and the keyword together. 

On the dedicated page for that keyword I would replace 

>As of PHP 8.0.0, the ::class constant may also be used on objects.

With 

>As of PHP 8.0.0, the ::class keyword may also be used on objects.

Crell:
Fixed via #2112

PR #728, opened 2021-06-24
remove replaceable class from magic constant

amritpalnagrame:
#707 Fix as suggested by @kamil-tekiela.

### DIFF
--- a/language/constants.xml
+++ b/language/constants.xml
@@ -328,7 +328,7 @@ echo ANIMALS[1]; // outputs "cat"
         </entry>
        </row>
        <row xml:id="constant.coloncolonclass">
-        <entry><constant>::class</constant></entry>
+        <entry><constant><replaceable>ClassName</replaceable>::class</constant></entry>
         <entry>
          The fully qualified class name.
         </entry>
--- a/language/constants.xml
+++ b/language/constants.xml
@@ -328,7 +328,7 @@ echo ANIMALS[1]; // outputs "cat"
         </entry>
        </row>
        <row xml:id="constant.coloncolonclass">
-        <entry><constant><replaceable>ClassName</replaceable>::class</constant></entry>
+        <entry><constant>::class</constant></entry>
         <entry>
          The fully qualified class name.
         </entry>
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -621,7 +621,7 @@ Does\Not\Exist
     </example>
    </note>
    <para>
-    As of PHP 8.0.0, the <literal>::class</literal> keyword may also be used on
+    As of PHP 8.0.0, <literal>::class</literal> may also be used on
     objects. This resolution happens at runtime, not compile time. Its effect is
     the same as calling <function>get_class</function> on the object.
    </para>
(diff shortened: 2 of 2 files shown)


sy-records (approved):

kamil-tekiela:
I see now that the value was wrapped in `<replaceable>` and I wonder if that tag has any semantic meaning. In the manual it is replaced with `<span>` with a class that has no styles attached to it. I don't know about other formats, or if we should even worry about it. 

Does anyone know if it is safe to get rid of `<replaceable>` here?

language/constants.xml
```diff
@@ -328,7 +328,7 @@ echo ANIMALS[1]; // outputs "cat"
         </entry>
        </row>
        <row xml:id="constant.coloncolonclass">
-        <entry><constant><replaceable>ClassName</replaceable>::class</constant></entry>
+        <entry><constant>::class</constant></entry>
```
cmb69, line 331: [`<replaceable>`](https://tdg.docbook.org/tdg/4.5/replaceable.html) conveys a meaning, and I think it's better to keep that as is (and maybe to improve PhD to somehow mark it up; perhaps as italics).

language/oop5/basic.xml
```diff
@@ -621,7 +621,7 @@ Does\Not\Exist
     </example>
    </note>
    <para>
-    As of PHP 8.0.0, the <literal>::class</literal> constant may also be used on
+    As of PHP 8.0.0, the <literal>::class</literal> keyword may also be used on
```
cmb69, line 624: Maybe:
```suggestion
    As of PHP 8.0.0, the <constant><replaceable>ClassName</replaceable>::class</constant> magic constant may also be used on
```

kamil-tekiela: I'm on the fence. I'd rather use "keyword", which would be aligned with the rest of the page.

cmb69: Sorry, I lost context here, but still, calling `::class` a keyword isn't correct. Maybe either
```suggestion
    As of PHP 8.0.0, <literal>::class</literal> may also be used on
```
or
```suggestion
    As of PHP 8.0.0, <constant><replaceable>$object</replaceable>::class</constant> may also be used
```
and drop the "objects" on the next line.

kamil-tekiela: This is better.... but the `::class` can be used on other stuff too. See https://3v4l.org/6o6YA
Claiming that it only works on classes and objects, would be an understatement. It is a compile-time keyword/constant. Magic constants are generally run-time constants. `::class` is a category on its own with only one member.

cmb69: Hmm, right, although [most of this](https://3v4l.org/WP7Eb) is not particularly related to PHP 8. The first suggestion of my previous comment might be good for this paragraph. The case where the class does not exist is already catered to be the previous paragraph. Not sure what to do about string literals on the left hand side.

afilina:
I addressed the @cmb69's comments, but that caused the `check-whitespace` job to fail on files unrelated to this PR. I attempted to rebase and force-push, which seems to have messed up the commit history instead. I reopened a new PR from the original branch here: https://github.com/php/doc-en/pull/2112

Crell:
Fixed via #2112

PR #732, opened 2021-06-25
Added PDO::FETCH_GROUP to the Description

ErikThiart:
PDO::FETCH_COLUMN is illustrated in the description, but not PDO::FETCH_GROUP, so I added PDO::FETCH_GROUP as it is one of the most underrated / misunderstood functions in PDO, that can make an incredible difference in production speed if more developers discover it.

A net positive for the community.

### DIFF
--- a/reference/pdo/pdostatement/fetchall.xml
+++ b/reference/pdo/pdostatement/fetchall.xml
@@ -32,6 +32,12 @@
    <methodparam><type>int</type><parameter>mode</parameter><initializer>PDO::FETCH_INTO</initializer></methodparam>
    <methodparam><type>object</type><parameter>object</parameter></methodparam>
   </methodsynopsis>
+  
+   <methodsynopsis>
+   <modifier>public</modifier> <type>array</type><methodname>PDOStatement::fetchAll</methodname>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer>PDO::FETCH_GROUP</initializer></methodparam>
+  </methodsynopsis>
+  
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Thanks for the PR, but we don't maintain all possible fetch modes in the method signature. The purpose of 4 lines in the method signature is to show different overloads for the method parameters when the 3 special modes are used. 
With PDO::FETCH_COLUMN, the method accepts one more argument $colno.
With PDO::FETCH_CLASS, the method accepts two more arguments $class and $constructorArgs.
With PDO::FETCH_FUNC, the method accepts one more argument $callback. <-- this one is wrong in the manual. It was accidently changed recently, and a PR is waiting for review still. 

We don't need to maintain more lines in the signature as there are no more overloads for that method. PDO::FETCH_GROUP is mentioned down on the page, but if you need a proper explanation of fetch modes take a look at https://phpdelusions.net/pdo/fetch_modes

ErikThiart:
Fair enough, but should that information not be on the PHP manual itself, or is https://phpdelusions.net/pdo/fetch_modes part of the official PHP manual? 

I think what I would do in the meantime is submit a user note with some examples for users if that is fine.

kamil-tekiela:
We would appreciate it if you want to add some more examples to the PHP manual. This would certainly improve the state of things. I don't think we maintain a good description of all possible mods in the official PHP manual. The site I linked is not part of the official manual, but it's a really good site for PDO enthusiasts.

ErikThiart:
Good enough for me.

Long live PHP, long live PDO.

Issue #674, opened 2020-11-22
PHP 8.0 documentation tracker


Girgias:
Pages which need to be added/modified for complete documentation of PHP 8.0:

Based of the migration guide: https://www.php.net/manual/en/migration80.php (formally php/doc-en#170)

## Core:

- [x] New features:
   - [x] Type declarations (https://github.com/php/doc-en/pull/174)
     - [x] Union types 
     - [x] mixed
     - [x] static return
  - [x] Constructor Promotion (https://github.com/php/doc-en/pull/188)
  - [x] match expressions (https://github.com/php/doc-en/pull/206)
     - [x] UnhandledMatchError
  - [x] Named arguments allow passing arguments to a function based on the parameter name, rather than the parameter position. (https://github.com/php/doc-en/pull/251)
  - [x] Attributes (https://github.com/php/doc-en/pull/254)
  - [x] Null-safe operator (https://github.com/php/doc-en/pull/244)
  - [x] Stringable interface (https://github.com/php/doc-en/pull/478)
  - [x] WeakMap (core) (https://github.com/php/doc-en/pull/479)
  - [x] $object::class (https://github.com/php/doc-en/pull/239)
  - [x] Catch without variable (https://github.com/php/doc-en/pull/238)
  - [x] traits can now have abstract private methods and they are validated against the implementation provided in the using class, according to our usual inheritance rules. (https://github.com/php/doc-en/pull/476)
  - [x] ValueError (https://github.com/php/doc-en/pull/205)
  - [x] DateTime format specifier `p` (https://github.com/php/doc-en/pull/230)
- [x] Modified features:
  - [x] throw expression (https://github.com/php/doc-en/pull/238)
  - [x]  &new; and &instanceof; can now be used with arbitrary expressions
  - [x] trailing commas in parameter lists (https://github.com/php/doc-en/pull/241)
  - [x] private methods do not enforce inheritance rules on child classes any more (except __construct) (https://github.com/php/doc-en/pull/481)
  - [x] Numeric strings (https://github.com/php/doc-en/pull/215)
    - [x] Comparison
    - [x] When a string is considered numeric
  - [x] Float to string conversion (https://github.com/php/doc-en/pull/215)
  - [x] Operator precedence
    - [x] Concatenation (011596bcff82477d121de738e6cf83bd2b13bb40)
- [x] Deprecated
   - [x] Optional args after mandatory ones (https://github.com/php/doc-en/pull/251, https://github.com/php/doc-en/pull/270)
[...]

PR #733, opened 2021-06-25
Fix ldap functions description

saundefined:
Fixed `result` parameter description, based on stubs for:

- [ldap_free_result](https://github.com/php/php-src/blob/master/ext/ldap/ldap.stub.php#L52)
- [ldap_count_entries](https://github.com/php/php-src/blob/master/ext/ldap/ldap.stub.php#L54)
- [ldap_first_entry](https://github.com/php/php-src/blob/master/ext/ldap/ldap.stub.php#L56)
- [ldap_get_entries](https://github.com/php/php-src/blob/master/ext/ldap/ldap.stub.php#L60)
- [ldap_count_references](https://github.com/php/php-src/blob/master/ext/ldap/ldap.stub.php#L125)
- [ldap_parse_result](https://github.com/php/php-src/blob/master/ext/ldap/ldap.stub.php#L144)

Added `ldap_count_references` function, ref #674

### DIFF
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -1021,6 +1021,8 @@ Passing untrusted data to this parameter is <emphasis>insecure</emphasis>, unles
 
 <!ENTITY ldap.parameter.ldap 'An LDAP resource, returned by <function xmlns="http://docbook.org/ns/docbook">ldap_connect</function>.'>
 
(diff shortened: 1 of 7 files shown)


cmb69 (approved):
Thank you!

PR #734, opened 2021-06-26
Capitalize the first letter of a sentence in match.xml

zors1:


kamil-tekiela:
Thanks for the PR. Despite `match` being the first word, it is a PHP keyword and we don't usually capitalize PHP keywords. I understand that this sentence might not be clear enough, so if you want you can propose a different variant, but please don't capitalize keywords, function names, etc.

zors1:
Ok, sorry about that. The word match is capitalized in the bulleted list describing the key differences between switch and match, so I thought it should be capitalized there too.

cmb69:
Ah, right. It should *not* be capitalized in the bulleted list either.

zors1:
Can I submit a pull request that will fix this?

kamil-tekiela:
Good catch! Yes, please submit a PR to change the capitalization in the list.

PR #735, opened 2021-06-26
is_numeric: See also filter_var (specifically with validation_filters)

AllenJB:
filter_var validation filters provide for more specific filtering than is_numeric (such as "is integer").

I've avoided the `<function>` tag here to specifically link to validation filters. If you can think of a better alternative, let me know.

Edit: Fix hidden 'function' tag name

### DIFF
--- a/reference/var/functions/is-numeric.xml
+++ b/reference/var/functions/is-numeric.xml
@@ -119,6 +119,7 @@ NULL is NOT numeric
     <member><function>is_string</function></member>
     <member><function>is_object</function></member>
     <member><function>is_array</function></member>
+    <member><function>filter_var</function> with <link linkend="filter.filters.validate">validation filters</member>
    </simplelist>
   </para>
  </refsect1>
--- a/reference/var/functions/is-numeric.xml
+++ b/reference/var/functions/is-numeric.xml
@@ -119,7 +119,7 @@ NULL is NOT numeric
     <member><function>is_string</function></member>
     <member><function>is_object</function></member>
     <member><function>is_array</function></member>
-    <member><function>filter_var</function> with <link linkend="filter.filters.validate">validation filters</member>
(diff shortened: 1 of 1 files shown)


reference/var/functions/is-numeric.xml
```diff
@@ -119,6 +119,7 @@ NULL is NOT numeric
     <member><function>is_string</function></member>
     <member><function>is_object</function></member>
     <member><function>is_array</function></member>
+    <member><link linkend="function.filter-var">filter_var</link> - Filters a variable with <link linkend="filter.filters.validate">validation filters</link></member>
```
Girgias, line 122:
```suggestion
    <member><function>filter_var</function> - Filters a variable with <link linkend="filter.filters.validate">validation filters</link></member>
```

The function tag takes care of auto linking functions.

Girgias (changes requested):

reference/var/functions/is-numeric.xml
AllenJB: The function tag also generates a description along side the link. If the suggested change were implemented, it would look something like:

* **filter_var** - Filters a variable with a specified filter - Filters a variable with **validation filters**

I thought of using:

`<member><function>filter_var</function> (see specifically the <link linkend="filter.filters.validate">validation filters</link>)</member>`

Which results in:

* **filter_var** - Filters a variable with a specific filter (see specifically the **validation filters**)

But this feels a bit disjointed and long-winded to me.

cmb69: Maybe don't mention filter_var at all?

AllenJB: I believe that would cause confusion because the validation filters page does not link back to filter_var / indicate how to use them. The only link back is by following breadcrumbs, which I don't think is obvious enough.

afilina:
Bumping reviewers: @Girgias @cmb69

cmb69:
`filter_var()` has already been linked by #1744. In my opinion, we can leave it at this.

PR #742, opened 2021-06-29
Fix issues in runkit7 directory

saundefined:


reference/runkit7/functions/runkit7-zval-inspect.xml
```diff
@@ -28,8 +28,73 @@
...
  </refsect1>
-</refentry>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   The array returned by this function contains the following elements:
+   <simplelist>
+    <member><literal>address</literal></member>
+    <member><literal>refcount</literal></member>
```
TysonAndre, line 38: This should mention that refcount and is_ref are not included for non-refcounted/non-references (e.g. integers that aren't scalars). I don't remember how that's usually done

```suggestion
    <member><literal>refcount</literal> (optional)</member>
```


```c
	addr_len = spprintf(&addr, 0, "0x%0lx", (long)value);
	add_assoc_stringl(return_value, "address", addr, addr_len);
	efree(addr);
	addr = NULL;

	if (Z_REFCOUNTED_P(value)) {
		add_assoc_long(return_value, "refcount", Z_REFCOUNT_P(value));
		add_assoc_bool(return_value, "is_ref", Z_ISREF_P(value));
	}

	add_assoc_long(return_value, "type", Z_TYPE_P(value));
```

saundefined: @TysonAndre, thank you!  Added the `optional` note.

reference/runkit7/functions/runkit7-zval-inspect.xml
```diff
@@ -28,8 +28,73 @@
...
+   </example>
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><link linkend="language.references">References Explained</link></member>
+    <member><link xlink:href="&url.derick.references;">References Explained (by Derick Rethans)</link></member>
```
TysonAndre, line 93: Aside, out of scope: This article was published in 2005 and describes the behavior in php 4.3.0+ - I haven't read through it but would have thought the behavior of php changed in some ways since then, e.g. objects stopped being copy by value around/in php 5.0?

https://derickrethans.nl/php-references-article.html

> In PHP 5, this works
all a little bit better. Here you can expect a warning and
it will behave “properly”. Hopefully, a backported fix
for this problem makes it into a new minor version of
PHP 4—PHP 4.4.

Modern alternatives for php 7 that I'm aware of:
https://www.phpinternalsbook.com/php7/zvals/references.html
https://www.npopov.com/2015/05/05/Internal-value-representation-in-PHP-7-part-1.html
(but I'm pretty sure that is targeted more towards people writing in C than end users of php)

TysonAndre (approved):

reference/runkit7/functions/runkit7-zval-inspect.xml
saundefined: Ah, exactly!

@cmb69, there is also in [debug_zval_dump()](https://www.php.net/en/function.debug-zval-dump) links.

Does it seem that we can replace both to phpinternalsbook, or remove it altogether, wdyt?

cmb69: @derickr, what do you think about still linking to https://derickrethans.nl/php-references-article.html? Is the explanation still up to date for PHP 7/8?

derickr: It's not wrong, but the whole interning of strings and the like make it not fully featured any more. If I could only find the time updating it...

saundefined: Hm..

@cmb69, leave it as it is?

cmb69: Yes. :)

PR #743, opened 2021-06-29
Remove unsafe AddHandler example

marcovtwout:
The internet will always take your example code and repost it out of context. Googling shows plenty of installation guides that use `AddHandler application/x-httpd-php .php`, which potentially leads to parsing unsafe files like `evil.php.jpg`

I propose to totally remove the unsafe example and only show the correct version with `<FilesMatch \.php$>`, similar to https://www.php.net/manual/en/install.unix.apache2.php


cmb69:
Thank you!

PR #749, opened 2021-07-01
New example for class_exists

kamil-tekiela:
`__autoload` has been removed now, so the example using it should be changed too.


cmb69:
Thank you!

PR #750, opened 2021-07-01
New examples for classes

kamil-tekiela:
It's still a dumb example, but I didn't write it. Not sure many people even know about this page.

### DIFF
--- a/reference/classobj/examples.xml
+++ b/reference/classobj/examples.xml
@@ -131,8 +131,8 @@ echo "\nleafy: Methods\n";
 printMethods($leafy);
 
 echo "\nParentage:\n";
-objectBelongsTo($leafy, "Spinach");
-objectBelongsTo($leafy, "Vegetable");
+objectBelongsTo($leafy, Spinach::class);
+objectBelongsTo($leafy, Vegetable::class);
 
 ?>
 ]]>
--- a/reference/classobj/examples.xml
+++ b/reference/classobj/examples.xml
@@ -6,62 +6,66 @@
  <para>
   In this example, we first define a base class and an extension
   of the class. The base class describes a general vegetable,
-  whether it is edible or not and what is its color. The subclass
+  whether it is edible, and what is its color. The subclass
   <varname>Spinach</varname> adds a method to cook it and another to
   find out if it is cooked.
  </para>
  <para>
   <example>
-   <title>classes.inc</title>
+   <title>Class definitions</title>
+   <para><varname>Vegetable</varname></para>
    <programlisting role="php">
 <![CDATA[
 <?php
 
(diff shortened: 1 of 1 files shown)


reference/classobj/examples.xml
```diff
@@ -132,25 +133,38 @@ print_methods($leafy);
 echo "\nParentage:\n";
 class_parentage("leafy", "Spinach");
 class_parentage("leafy", "Vegetable");
```
afilina, line 135:
```suggestion
objectBelongsTo($leafy, "Spinach");
objectBelongsTo($leafy, "Vegetable");
```

reference/classobj/examples.xml
```diff
@@ -79,40 +83,37 @@ class Spinach extends Vegetable {
...
-    echo ", a subclass of $class\n";
-} else {
-    echo "Object $obj does not belong to a subclass of $class\n";
-}
+    if (is_subclass_of($GLOBALS[$obj], $class)) {
+        echo "Object $obj belongs to class " . get_class($GLOBALS[$obj]);
+        echo ", a subclass of $class\n";
+    } else {
+        echo "Object $obj does not belong to a subclass of $class\n";
+    }
```
afilina, line 113: Let's avoid globals by passing the object instead of its variable name.
```suggestion
    if (is_subclass_of($obj, $class)) {
        echo "Object $obj belongs to class " . get_class($obj);
        echo ", a subclass of $class\n";
    } else {
        echo "Object $obj does not belong to a subclass of $class\n";
    }
```

afilina (commented):
I mostly edited the example for slightly better code quality (still meh, but better).

reference/classobj/examples.xml
kamil-tekiela: I thought of this but if we change the parameter then the rest of this function will break with an error as `$obj` is used in string context. Do you have a suggestion how to fix the whole function and its output?

reference/classobj/examples.xml
```diff
@@ -79,40 +83,37 @@ class Spinach extends Vegetable {
...
 }
 
 function print_methods($obj)
 {
-$arr = get_class_methods(get_class($obj));
-foreach ($arr as $method) {
-    echo "\tfunction $method()\n";
-}
+    $arr = get_class_methods(get_class($obj));
```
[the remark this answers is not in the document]
kamil-tekiela: This is a built-in method, so I can't change the case

reference/classobj/examples.xml
afilina: Perhaps we don't need to output the $obj at all.

afilina (approved):

PR #752, opened 2021-07-01
Fix remaining usages of __autoload

kamil-tekiela:
PHAR still uses `__autoload` in their documentation but I don't care that much about changing it right now.

### DIFF
--- a/reference/classobj/functions/class-exists.xml
+++ b/reference/classobj/functions/class-exists.xml
@@ -32,7 +32,7 @@
      <term><parameter>autoload</parameter></term>
      <listitem>
       <para>
-       Whether or not to call &link.autoload; by default.
+       Whether to call &link.autoload; by default.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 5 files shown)


appendices/userlandnaming.xml
```diff
@@ -90,7 +90,7 @@
     </para>
     <itemizedlist>
      <listitem><para><link linkend="object.get">__get()</link></para></listitem>
-     <listitem><para><function>__autoload</function></para></listitem>
+     <listitem><para><link linkend="object.sleep">__sleep()</link></para></listitem>
     </itemizedlist>
    </listitem>
   </itemizedlist>
```
cmb69, line 96: Not sure about this; AFAIK, `__autoload()` is the only function with a `__` prefix, so maybe it makes sense to keep it here for now. If we remove it, we should likely drop the list, in favor of the simpler: `For instance, <link linkend="object.get">__get()</link>`.

cmb69 (commented):
Thank you!

appendices/userlandnaming.xml
kamil-tekiela: Thanks, dropped that commit

cmb69 (approved):

PR #753, opened 2021-07-03
The opened_path paramater is only set when STREAM_USE_PATH is set

othercorey:
If you use the `string` typehint, you will get an error about null passed from `fopen()`.

### DIFF
--- a/reference/stream/streamwrapper/stream-open.xml
+++ b/reference/stream/streamwrapper/stream-open.xml
@@ -14,7 +14,7 @@
    <methodparam><type>string</type><parameter>path</parameter></methodparam>
    <methodparam><type>string</type><parameter>mode</parameter></methodparam>
    <methodparam><type>int</type><parameter>options</parameter></methodparam>
-   <methodparam><type>string</type><parameter role="reference">opened_path</parameter></methodparam>
+   <methodparam><type>?string</type><parameter role="reference">opened_path</parameter></methodparam>
   </methodsynopsis>
   <para>
    This method is called immediately after the wrapper is initialized (f.e. by


afilina:
Do you have a code example where such an error occurs?

othercorey:
I do, but it's in a unit test with more code that isn't worth your time. I don't think there's a bug or anything.

I worded the issue poorly working through issues with adding type hints. I'm suggesting the parameter should be nullable because `null` is passed when `STREAM_USE_PATH` option isn't set. That's the default.

https://github.com/php/doc-en/pull/840

afilina:
@othercorey I don't mind reading an overly verbose test :)

cmb69:
It seems to me that that reference parameter is [always passed as NULL](https://github.com/php/php-src/blob/php-7.4.22/main/streams/userspace.c#L373), and [only if the output value is a string, it is used](https://github.com/php/php-src/blob/php-7.4.22/main/streams/userspace.c#L394). Since only the input type is checked for reference parameters, the type could actually be `null`, but that is not allowed, so `?string` appears to be good.

It would all be clearer, if we had output parameters, and streamWrapper would be an interface. One may dream…

othercorey:
Since you asked for an example:

The StreamWrapper is defined here:
https://github.com/cakephp/cakephp/blob/5.0/tests/test_app/TestApp/Http/Client/Adapter/CakeStreamWrapper.php#L21

This has the correct `?string` type hint. If you use the documented `string` type hint, the error occurs.

The wrapper is registered here:
https://github.com/cakephp/cakephp/blob/5.0/tests/TestCase/Http/Client/Adapter/StreamTest.php#L41

And then a stream is created and opened without options here:
https://github.com/cakephp/cakephp/blob/5.0/tests/TestCase/Http/Client/Adapter/StreamTest.php#L62

That causes this failure:
```
TypeError: TestApp\Http\Client\Adapter\CakeStreamWrapper::stream_open(): Argument #4 ($openedPath) must be of type string, null given

/cake/tests/test_app/TestApp/Http/Client/Adapter/CakeStreamWrapper.php:21
```

reference/stream/streamwrapper/stream-open.xml
```diff
@@ -14,7 +14,7 @@
    <methodparam><type>string</type><parameter>path</parameter></methodparam>
    <methodparam><type>string</type><parameter>mode</parameter></methodparam>
    <methodparam><type>int</type><parameter>options</parameter></methodparam>
-   <methodparam><type>string</type><parameter role="reference">opened_path</parameter></methodparam>
+   <methodparam><type>?string</type><parameter role="reference">opened_path</parameter></methodparam>
```
salathe, line 17: The correct DocBook XML markup for the nullable string type is:

```suggestion
   <methodparam><type class="union"><type>string</type><type>null</type></type><parameter role="reference">opened_path</parameter></methodparam>
```

This will be rendered in the manual as `?string`.

cmb69:
Another example can easily be written by adapting the [stream wrapper example in the PHP manual](https://www.php.net/manual/en/stream.streamwrapper.example-1.php). Adding a `string` type declaration causes the error; a `?string` (or `?whatever`) type makes the example pass.

Anyhow, I've just committed ab8bf7a2cd8dca25afea5e74fecdc16f1ae5c56a.

Thanks you all! :)

PR #755, opened 2021-07-03
Fix typos in setup.xml

zors1:
This fixes a couple of typos on https://www.php.net/manual/en/rar.resources.php


saundefined (approved):
Thank you!

PR #760, opened 2021-07-06
Add period and comma on Phar intro page

zors1:
Added a couple of missing punctuation marks on https://www.php.net/manual/en/intro.phar.php


saundefined (approved):
Thank you!

PR #761, opened 2021-07-06
Sync enhancement

saundefined:


reference/sync/syncsemaphore/construct.xml
```diff
@@ -42,7 +42,7 @@
     <term><parameter>initialval</parameter></term>
     <listitem>
      <para>
- The initial value of the semaphore.  This is the number of locks that may be obtained.
+ The initial value of the semaphore. This is the number of locks that may be obtained.
```
cmb69, line 45: Nit: indentation:
```suggestion
      The initial value of the semaphore. This is the number of locks that may be obtained.
```

cmb69 (commented):
Thank you! Looks good, besides the missing indentation.

saundefined:
@cmb69, thanks! Fixed indentation in all files.

cmb69 (commented):
I'm not sure whether it is a good idea to fix indentation in otherwise unmodified files, but I don't have a strong opinion on that.

saundefined:
@cmb69,  hm.. in one file only indents are fixed (`syncsharedmemory/read.xml`), 
the rest - there're still some changes.

I'm not sure how to do it right: 
reset and fix the indentation only in those paragraphs where the changes were made (good for translators, but then some paragraphs will have an indentation, and some — not), 

or leave it as it is now? 🥺

cmb69:
I think it's okay to fix all indentation for files, where an actual change is made, since translators will have to touch that file anyway, and can skip indentation lines without further change, if these are already oproperly indented in the translation.

saundefined:
@cmb69 thank you!

reverted the `syncsharedmemory/read.xml`.

cmb69 (approved):

PR #765, opened 2021-07-07
Reflect new default for `CURLOPT_TIMECONDITION` since cURL 7.46.0

pfofi:
In cURL 7.46.0 the default for `CURLOPT_TIMECONDITION` was changed from `CURL_TIMECOND_IFMODSINCE` to `CURL_TIMECOND_NONE`.

References:
- https://curl.se/libcurl/c/CURLOPT_TIMECONDITION.html
- https://curl.se/changes.html#7_46_0
- https://github.com/curl/curl/commit/cd2b73b3e

### DIFF
--- a/reference/curl/constants.xml
+++ b/reference/curl/constants.xml
@@ -1802,7 +1802,7 @@
    </term>
    <listitem>
     <simpara>
-     Available since PHP 5.5.0 and cURL 7.9.7
+
     </simpara>
    </listitem>
   </varlistentry>
--- a/reference/curl/constants.xml
+++ b/reference/curl/constants.xml
@@ -1805,11 +1805,6 @@
      Available since PHP 5.5.0 and cURL 7.9.7
     </simpara>
    </listitem>
-   <listitem>
-    <simpara>
-
-    </simpara>
-   </listitem>
   </varlistentry>
   <varlistentry xml:id="constant.curl-timecond-ifmodsince">
    <term>
--- a/reference/curl/constants.xml
+++ b/reference/curl/constants.xml
@@ -1795,6 +1795,17 @@
     </simpara>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curl-timecond-none">
+   <term>
+    <constant>CURL_TIMECOND_NONE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+
+    </simpara>
+   </listitem>
+  </varlistentry>
   <varlistentry xml:id="constant.curl-timecond-ifmodsince">
    <term>
     <constant>CURL_TIMECOND_IFMODSINCE</constant>
--- a/reference/curl/constants.xml
+++ b/reference/curl/constants.xml
@@ -1802,6 +1802,11 @@
    </term>
    <listitem>
     <simpara>
+     Available since PHP 5.5.0 and cURL 7.9.7
+    </simpara>
+   </listitem>
+   <listitem>
+    <simpara>
 
     </simpara>
    </listitem>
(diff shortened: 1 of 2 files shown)


reference/curl/constants.xml
```diff
@@ -1795,6 +1795,17 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curl-timecond-none">
+   <term>
+    <constant>CURL_TIMECOND_NONE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+
```
afilina, line 1805: We need to specify the PHP and cURL version when this one introduced. See: https://github.com/php/doc-en/blob/67bd2e1ec09ddcd9989ccd7b209c613794983a0a/reference/curl/constants.xml#L74

reference/curl/functions/curl-setopt.xml
```diff
@@ -1129,10 +1129,12 @@
             a <literal>"304 Not Modified"</literal> header will be returned
             assuming <constant>CURLOPT_HEADER</constant> is &true;.
             Use <constant>CURL_TIMECOND_IFUNMODSINCE</constant> for the reverse
-            effect. <constant>CURL_TIMECOND_IFMODSINCE</constant> is the
-            default.
+            effect. The default is <constant>CURL_TIMECOND_NONE</constant> which
+            ignores the <constant>CURLOPT_TIMEVALUE</constant>.
```
afilina, line 1133: Was that change introduced in a specific version? Also, I couldn't verify that information in the cURL documentation. Perhaps I'm looking at the wrong page? https://curl.se/libcurl/c/CURLOPT_TIMECONDITION.html

reference/curl/functions/curl-setopt.xml
```diff
@@ -1161,8 +1163,7 @@
            <entry valign="top"><constant>CURLOPT_TIMEVALUE</constant></entry>
            <entry valign="top">
             The time in seconds since January 1st, 1970. The time will be used
-            by <constant>CURLOPT_TIMECONDITION</constant>. By default,
-            <constant>CURL_TIMECOND_IFMODSINCE</constant> is used.
+            by <constant>CURLOPT_TIMECONDITION</constant>.
```
afilina, line 1166: Was that change introduced in a specific version?

reference/curl/constants.xml
pfofi: Thank you, I have added the missing version information.

reference/curl/functions/curl-setopt.xml
pfofi: There is a respective entry in [cURL's changelog](https://curl.se/changes.html#7_46_0):
> RTSP: do not add if-modified-since without timecondition

It links to a post on [Stack Overflow](https://stackoverflow.com/questions/33903982/curl-timecond-none-doesnt-work-how-to-remove-if-modified-since-header) that in turn links to <https://github.com/curl/curl/commit/cd2b73b3e>. From that commit it is apparent the change was introduced in 7.46.0.

cURLs docs don't explicity state what `CURL_TIMECOND_NONE` does, just that it is the default. I admit my explanation was a bit vague. I have changed the PR to make it clearer.

reference/curl/functions/curl-setopt.xml
pfofi: This option was not changed. `CURL_TIMECOND_IFMODSINCE` belongs to `CURLOPT_TIMECONDITION`. I think the default should only be stated there.

reference/curl/constants.xml
```diff
@@ -1795,6 +1795,17 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curl-timecond-none">
+   <term>
+    <constant>CURL_TIMECOND_NONE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Available since PHP 5.5.0 and cURL 7.9.7
```
kamil-tekiela, line 1805: We document changes only as of PHP 7.0. Was this available with the curl version used in PHP 7?

cmb69: [PHP 7.0.0 already required cURL ≥ 7.10.5](https://github.com/php/php-src/blob/php-7.0.0/ext/curl/config.m4#L45), so this info can be removed.

cmb69:
Thank you!

PR #766, opened 2021-07-08
Capitalize word on webPhar page

zors1:
Capitalized a word at the beginning of a sentence on https://www.php.net/manual/en/phar.webphar

PR #767, opened 2021-07-08
Add missing period on webPhar page

zors1:
In addition to #766


tiffany-taylor (approved):
Thanks!

salathe:
👎  The convention is to **not** have a period at the end of `refpurpose`s.

cmb69:
I think the punctuation (and maybe more) is generally wrong; at least for me this refpurpose yields `E_PARSE_ERROR`.

tiffany-taylor:
I think `<refpurpose>` should probably be changed to `<para>` if the punctuation breaks the build.

cmb69:
There's no problem with the build (doc is valid), but I think the purpose of this function should be described more understandably (and there definitly shouldn't be a period at all).

PR #768, opened 2021-07-08
Revise wording for synopsis and summary

tiffany-taylor:
Not sure if this is better. I pulled a basic description from the method summary for the synopsis and muted some of the emotion in the method summary.

Changes are related to #766 and #767.


cmb69:
In my opinion, much better. Thank you! I wouldn't have reformatted that papagraph, though.

PR #769, opened 2021-07-09
Fix wrong params names for mysqli::__construct() [81232]

sy-records:


reference/mysqli/mysqli/construct.xml
```diff
@@ -49,7 +49,7 @@
   <para>
    <variablelist>
     <varlistentry>
-     <term><parameter>host</parameter></term>
+     <term><parameter>hostname</parameter></term>
```
kamil-tekiela, line 52: Please change the parameter name in the text below too.

kamil-tekiela (changes requested):

reference/mysqli/mysqli/construct.xml
sy-records: done

kamil-tekiela (approved):

PR #770, opened 2021-07-09
Updating mysqli: bind_result

kamil-tekiela:
I have split up the note, added a helpful tip, and modernized the code example.

### DIFF
--- a/reference/mysqli/mysqli_stmt/bind-result.xml
+++ b/reference/mysqli/mysqli_stmt/bind-result.xml
@@ -30,18 +30,22 @@
    MySQL client/server protocol places the data for the bound columns into
    the specified variables <parameter>var</parameter>/<parameter>vars</parameter>.
   </para>
+  <para>
+   A column can be bound or rebound at any time, even after a result set has
+   been partially retrieved. The new binding takes effect the next time 
+   <function>mysqli_stmt_fetch</function> is called.
+  </para>
   <note>
    <para>
     Note that all columns must be bound after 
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli_stmt/bind-result.xml
```diff
@@ -30,18 +30,22 @@
    MySQL client/server protocol places the data for the bound columns into
    the specified variables <parameter>var</parameter>/<parameter>vars</parameter>.
   </para>
+  <para>
+   A column can be bound or rebound at any time, even after a result set has
+   been partially retrieved. The new binding takes effect the next time 
+   <function>mysqli_stmt_fetch</function> is called.
+  </para>
```
afilina, line 37: Does it make sense to move this note one position down? I feel like information would flow better if mysqli_stmt_execute/mysqli_stmt_fetch note came before this one.

afilina:
Other than one question, LGTM.

afilina (commented):
Other than one question, LGTM.

reference/mysqli/mysqli_stmt/bind-result.xml
kamil-tekiela: I'm not sure. 

The way I am proposing, it looks logical too:
image


The way you are suggesting, it would visually look a little bit distracting. 
image

I don't know which way is better.

afilina: We can keep what's already there, unless we find a truly compelling reason.

afilina (approved):

PR #772, opened 2021-07-10
Fixed remaining issues in mysqli versions.xml

kamil-tekiela:
As salathe said, there don't need to be duplicate entries for methods in `mysqli_result`. However, for properties, I left it with a comment for compliance with the rest of the file. Otherwise, I would have to deduplicate the whole file. (it's something we might have to do in PHP 9). 
I also removed the two entries for functions that never existed in PHP.


cmb69 (approved):
Thank you!

PR #776, opened 2021-07-11
PHP 8.1: Document mysqli_result::fetch_column

kamil-tekiela:
It can wait...


reference/mysqli/mysqli_result/fetch-column.xml
```diff
@@ -0,0 +1,150 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
```
saundefined, line 2:
```suggestion
```
This comment can be removed.

saundefined (approved):
Except for one nit, looks good to me, thanks!

reference/mysqli/mysqli_result/fetch-column.xml
kamil-tekiela: Why can it be removed? It exists on every page currently. Sorry I am not very familiar with how it works.

PR #777, opened 2021-07-12
Move the SQL injection warning up the page

AshleyPinner:
If you visit this page on a 1080p monitor, the SQL Injection warning is below the fold, so if you just skim the function reference, you will miss possibly the most important security notice for this function. This proposed change is to move the warning up to just after the method synopsis so it's visible before the first fold of the page.

### DIFF
--- a/reference/mysqli/mysqli/query.xml
+++ b/reference/mysqli/mysqli/query.xml
@@ -26,6 +26,19 @@
    Performs a <parameter>query</parameter> against the database.
   </para>
   <para>
+   <warning>
+     <title>Security warning: SQL injection</title>
+     <para>
+        If the query contains any variable input then 
+        <link linkend="mysqli.quickstart.prepared-statements">parameterized
+        prepared statements</link> should be used instead. Alternatively, the
+        data must be properly formatted and all strings must be escaped using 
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
This is such a prevalent problem that I would be ok with moving the warning up or even making the whole page blink in red...
However, I would wait with this change until https://github.com/php/doc-en/pull/703 is approved. Then we can move the warning easily on all 3 pages.

AshleyPinner:
@kamil-tekiela Happy to wait for 703, I only spotted it after I proposed this change. I do feel like it should be above the fold though, so even people skimming arguments are exposed to it. Blinking it in red might have [browser support implications](https://caniuse.com/mdn-html_elements_blink) 😉

kamil-tekiela:
Can you rebase this PR and move the warning in all three places, please? The other PR is approved.

Crell:
@AshleyPinner Are you going to continue this?  It still seems like a good idea.

afilina:
Pinging author @AshleyPinner. Do you need help rebasing or making the additional changes? You can send me an instant message via Telegram, and we can jump on a call from there: https://t.me/afilina_dm

PR #779, opened 2021-07-12
fixed Comparison item between prepared and non-prepared statements.

mumumu:
https://www.php.net/manual/en/mysqli.quickstart.prepared-statements.php

The table `Comparison of prepared and non-prepared statements` has duplicate items `Statement string transferred from client to server`. This PR clarify difference between them.

### DIFF
--- a/reference/mysqli/quickstart.xml
+++ b/reference/mysqli/quickstart.xml
@@ -985,7 +985,7 @@ array(2) {
       <entry>n</entry>
      </row>
      <row>
-      <entry>Statement string transferred from client to server</entry>
+      <entry>Statement string transferred from client to server, repeated (n) execution</entry>
       <entry>1 template, n times bound parameter, if any</entry>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
It looks like it says the same as the line before it. Maybe combine them together?

afilina:
@kamil-tekiela It does not say the same as the line before it:

image

My only question is whether it might be better to simply say "1+n" and "n" like the line above, for uniformity?

kamil-tekiela:
> It does not say the same as the line before it

Then I am sorry, but I have no idea what the original author of this documentation piece meant.

kamil-tekiela:
I decided to drop this table altogether in 60a292997b3c6341cf52099d901aa0b5f8673d87. 
The information it contained was confusing. It wasn't clear whether it was meant to convince people to use prepared statements or push them away. The purpose of this table was not well defined.

PR #783, opened 2021-07-14
Fix space and sort types that define mixed union type

othercorey:
There is a weird space after `float` which seems to be caused by the multi-line list of types. Made the list of types a single line to avoid it.

Also, the list was in a random order which made it awkward to read. There are probably many opinions on the correct order, but picked something that grouped types together better.

Original:
`array|bool|callable|int|float |object|resource|string|null`

New:
`object|resource|callable|array|string|int|float|bool|null`

### DIFF
--- a/language/types/declarations.xml
+++ b/language/types/declarations.xml
@@ -165,8 +165,8 @@ Stack trace:
   <para>
    <type>mixed</type> is equivalent to the <link linkend="language.types.declarations.union">union type</link>
    <type class="union">
-    <type>array</type><type>bool</type><type>callable</type><type>int</type><type>float</type>
-    <type>object</type><type>resource</type><type>string</type><type>null</type>
+    <type>object</type><type>resource</type><type>array</type><type>string</type>
+    <type>int</type><type>float</type><type>bool</type><type>null</type>
    </type>.
    Available as of PHP 8.0.0.
   </para>


language/types/declarations.xml
```diff
@@ -165,8 +165,7 @@ Stack trace:
   <para>
    <type>mixed</type> is equivalent to the <link linkend="language.types.declarations.union">union type</link>
    <type class="union">
-    <type>array</type><type>bool</type><type>callable</type><type>int</type><type>float</type>
-    <type>object</type><type>resource</type><type>string</type><type>null</type>
+    <type>object</type><type>resource</type><type>callable</type><type>array</type><type>string</type><type>int</type><type>float</type><type>bool</type><type>null</type>
```
nikic, line 168:
```suggestion
    <type>object</type><type>resource</type><type>array</type><type>string</type><type>int</type><type>float</type><type>bool</type><type>null</type>
```
Inclusion of callable here doesn't make sense.

othercorey: I'm not the right person to decide so feel free to remove it.

cmb69: This has been copied from [the respective RFC](https://wiki.php.net/rfc/mixed_type_v2#proposal). But yes, callable makes no partuicular sense, because that is either a string, an array or an object.

othercorey: Ok. Removed callable and updated commit description. Hope that helps!

cmb69: Thank you!

cmb69:
> There is a weird space after float which seems to be caused by the multi-line list of types.

That should propbably be handled by PhD (i.e. remove all whitespace from union types).

othercorey:
> That should propbably be handled by PhD (i.e. remove all whitespace from union types).

I don't know how to fix that, sorry.

cmb69:
https://github.com/php/phd/pull/46 should do, I think.

othercorey:
Since having 5 types per line was apparently important, updated to just order and remove callable.

cmb69 (approved):
> Since having 5 types per line was apparently important, […]

It's just for readabilty, especially for translators.

PR #784, opened 2021-07-14
Fixed the link to magic constants

OleksiiBulba:
Page https://www.php.net/manual/en/reserved.constants.php contains incorrect link to magic constants (https://www.php.net/manual/en/language.constants.predefined.php) while the correct one is https://www.php.net/manual/en/language.constants.magic.php


kamil-tekiela (approved):

kamil-tekiela:
Thank you.

PR #785, opened 2021-07-14
Fix minor issues in Apache2 config guide

kamil-tekiela:


install/unix/apache2.xml
```diff
@@ -177,38 +177,38 @@ cp php.ini-development /usr/local/lib/php.ini
...
-      <informalexample>
+     <informalexample>
        <para>
-        For PHP 7:
+        For PHP 8:
        </para>
        <programlisting role="apache-conf">
 <![CDATA[
-LoadModule php7_module modules/libphp7.so
+LoadModule php_module modules/libphp8.so
```
cmb69, line 193: Not sure, but I think this should be
```suggestion
LoadModule php_module modules/libphp.so
```

cmb69 (commented):
Thank you! It might be a good idea to drop the personalization right away (i.e. no "you").

kamil-tekiela:
To be honest, I only wanted to fix the typo and I ended up changing more. Fixing the personalization could be done in another PR as this might involve more changes.

cmb69 (approved):

PR #786, opened 2021-07-15
Syslog does not support the %m format specificer

Girgias:
Nor any other printf specifier

Drop version name for Windows note as a drive by fix.


cmb69 (approved):
Thank you!

sy-records (approved):

PR #788, opened 2021-07-16
Fix typehint

larowlan:
From manual page: https://php.net/class.recursivecallbackfilteriterator

---

The callback can be any callable, e.g. anonymous functions are supported


tiffany-taylor:
Appears correct https://github.com/php/php-src/blob/master/ext/spl/spl_iterators.stub.php#L33

cmb69:
Thank you!

PR #795, opened 2021-07-17
The fallback for undefined constants is removed as of PHP 8

kamil-tekiela:


language/types/array.xml
```diff
@@ -617,8 +617,9 @@ echo $foo[bar];
    </para>
    <warning>
     <simpara>
-     The fallback to treat an undefined constant as bare string is deprecated as
-     of PHP 7.2.0, and issues an error of level <constant>E_WARNING</constant>.
+     The fallback to treat an undefined constant as bare string is removed
+     as of PHP 8.0.0 and throws <classname>Error</classname> exception.
+     It has been deprecated as of PHP 7.2.0, and issues an error of level <constant>E_WARNING</constant>.
```
afilina, line 622: - Follow chronological order
- Past tense for both versions

```suggestion
     The fallback to treat an undefined constant as bare string has been deprecated 
     as of PHP 7.2.0, and issues an error of level <constant>E_WARNING</constant>.
     As of PHP 8.0.0, it has been removed and throws an <classname>Error</classname> exception.
```

afilina (changes requested):

language/types/array.xml
kamil-tekiela: Thanks. This is a good catch. I did it in this order because that is how it was written before (see the last line about E_NOTICE). I think I have to rewrite the whole paragraph now, so it makes sense.

kamil-tekiela:
@afilina I followed your suggestion. Does it look better now?

afilina (approved):

PR #797, opened 2021-07-17
Remove XSS and broken example from list() page

kamil-tekiela:
Let's not have XSS in the manual. 

The example was only applicable to PHP 5, but then the context got lost and the example was left in the manual. It was showing the right-to-left assignment of `list()`


afilina (approved):
LGTM. Indeed, the `list` example has no reason to be.

PR #801, opened 2021-07-20
Fix pthreads return values

saundefined:
Not sure about "indication of state" replacements 🤔

### DIFF
--- a/reference/pthreads/worker/isshutdown.xml
+++ b/reference/pthreads/worker/isshutdown.xml
@@ -26,7 +26,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.success;
+   Returns whether the worker has been shutdown or not.
   </para>
  </refsect1>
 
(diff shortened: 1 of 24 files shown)


reference/pthreads/worker/isshutdown.xml
```diff
@@ -26,7 +26,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns whether the worker has been shutdown or not.
+   &return.success;
```
cmb69, line 29: I would leave that as is; `&return.success;` is certainly wrong for *is*shutdown.

reference/pthreads/worker/isworking.xml
```diff
@@ -36,7 +36,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-  A boolean indication of state
+   &return.success;
```
cmb69, line 39: Maybe
```suggestion
   Returns whether the worker is still working or not.
```

cmb69 (commented):
Thank you!

cmb69 (approved):

PR #803, opened 2021-07-21
Fix quotation mark error

sy-records:


cmb69:
This kind of quoting is not uncommon in English ASCII texts. I think we can leave as is, or use `'` for starting and ending quotation mark.

cmb69 (approved):
Thank you!

PR #805, opened 2021-07-21
Changed the incorrect false for null.

### LINKED ISSUE #804: documentation: dom_import_simplexml incorrect return value.
From manual page: https://php.net/function.dom-import-simplexml

--- Return value is mentioned to be `The DOMElement node added or false if any errors occur.`
While the actual return type is `DOMElement|null`

KevinVanSonsbeek:
Fixes #804

### DIFF
--- a/reference/dom/functions/dom-import-simplexml.xml
+++ b/reference/dom/functions/dom-import-simplexml.xml
@@ -18,7 +18,7 @@
    This function takes the node <parameter>node</parameter> of class
    <link linkend="ref.simplexml">SimpleXML</link> and makes it into a
    <classname>DOMElement</classname> node. This new object can then be used
-   as a native <classname>DOMElement</classname> node.
+   as a native <classname>DOMElement</classname> node. 
   </para>
  </refsect1>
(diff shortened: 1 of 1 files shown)


KevinVanSonsbeek:
Closed the PR due to the issue possibly being incorrect.

cmb69:
Well, the fix is correct (the unrelated whitespace changes should be avoided, though).

KevinVanSonsbeek:
> Well, the fix is correct (the unrelated whitespace changes should be avoided, though).

Yeah. It's just that when i was looking at the source code. I noticed the behaviour of the method changed when going to php8. The stub added tells the return type is nullable. But the null returns have been replaced with throw returns.

cmb69:
The stub has just been fixed. Please change the method signature and the return value section; I don't think that we need a changelog entry.

KevinVanSonsbeek:
> The stub has just been fixed. Please change the method signature and the return value section; I don't think that we need a changelog entry.

Might not be needed. But it could be handy to add.

cmb69 (approved):
Thank you!

PR #806, opened 2021-07-21
Fix inaccurate information in pdo query

kamil-tekiela:
When reading the page I noticed that the first sentence says it only executes the SQL, so I decided to add "prepares and " but then I noticed more inaccuracies.


cmb69 (approved):
Thank you!

PR #809, opened 2021-07-21
Fix documented return type for CachingIterator::hasNext()

nickolasburr:
The return type for `CachingIterator::hasNext` is currently documented as `void` when it should be `bool`.


sy-records (approved):
Thank you.

[Stub](https://github.com/php/php-src/blob/7a9a37d909cdb66d9c2dbd1ac80b7f5579d8bddd/ext/spl/spl_iterators.stub.php#L215)

PR #813, opened 2021-07-23
Add missing parameter to pdo query

kamil-tekiela:
I'm not sure how it would be best to document this parameter. See both commits. I'd appreciate tips on how to make it clearer.

### DIFF
--- a/reference/pdo/pdo/query.xml
+++ b/reference/pdo/pdo/query.xml
@@ -39,7 +39,7 @@
 
   <para>
    <function>PDO::query</function> prepares and executes an SQL statement in
-   a single function call, returning the statement as a PDOStatement object.
+   a single function call, returning the statement as a <classname>PDOStatement</classname> object.
   </para>
   <para>
    For a query that you need to issue multiple times, you will realize better
@@ -89,6 +89,20 @@
         Alternatively, the SQL can be prepared manually before calling 
         <function>PDO::query</function>, with the data properly formatted using
         <function>PDO::quote</function> if the driver supports it.
+       </para>
+      </listitem>
+     </varlistentry>
+    <varlistentry>
+     <term><parameter>fetchMode</parameter></term>
+      <listitem>
+       <para>
+        The default fetch mode for the returned <classname>PDOStatement</classname>.
+        It must be one of the <literal>PDO::FETCH_*</literal> constants.
+       </para>
+       <para>
+        This is a shorthand for calling <methodname>PDOStatement::setFetchMode</methodname>
+        right after <methodname>PDO::prepare</methodname>. The subsequent
+        arguments vary depending on the selected fetch mode.
        </para>
       </listitem>
      </varlistentry>
--- a/reference/pdo/pdo/query.xml
+++ b/reference/pdo/pdo/query.xml
@@ -57,15 +57,6 @@
 
   <note>
    <para>
-    If more than one argument is passed to this function, the remaining arguments
-    will be treated as though you called
-    <function>PDOStatement::setFetchMode</function> on the resultant statement
-    object.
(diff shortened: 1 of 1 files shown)


cmb69:
IMO, the total looks (i.e. both commits) good. It's hard to document such abominations, anyway.

kamil-tekiela:
@cmb69 Thanks. Does this mean you approve the PR as is?

cmb69:
Yes, unless you are planning to document further parameters of that method soon, in which case it would be preferable to do it right away. :)

kamil-tekiela:
I don't feel like documenting the rest right now. These signatures are nasty to document properly. Maybe someone else will want to do that.

reference/pdo/pdo/query.xml
```diff
@@ -92,6 +83,21 @@
...
+    <varlistentry>
+     <term><parameter>fetchMode</parameter></term>
+      <listitem>
+       <para>
+        The default fetch mode for the returned <classname>PDOStatement</classname>.
+        It must be one of the <literal>PDO::FETCH_*</literal> constants.
+       </para>
+       <para>
+        If this argument is passed to the function, the remaining arguments
+        will be treated as though <function>PDOStatement::setFetchMode</function>
```
salathe, line 95: Should be `<methodname>` element, not `<function>`.

reference/pdo/pdo/query.xml
```diff
@@ -92,6 +83,21 @@
        </para>
       </listitem>
      </varlistentry>
+    <varlistentry>
+     <term><parameter>fetchMode</parameter></term>
+      <listitem>
+       <para>
+        The default fetch mode for the returned <classname>PDOStatement</classname>.
+        It must be one of the <literal>PDO::FETCH_*</literal> constants.
```
salathe, line 91: Ideally link to the contants: e.g. `<link linkend="pdo.constants"><literal>PDO::FETCH_*</literal></link>`

reference/pdo/pdo/query.xml
kamil-tekiela: Maybe it should be, but I was going for consistency with the rest of the manual and 90% of PDO manual uses `<function>` instead of `<methodname>`

reference/pdo/pdo/query.xml
kamil-tekiela: Sorry, I copied it from https://www.php.net/manual/en/pdostatement.setfetchmode.php

reference/pdo/pdo/query.xml
cmb69: Ideally, the rest should be fixed as well (AFAIK, PhD currently handles it identically, but there is a semantic difference), but whenever touching parts which have it wrong, it makes sense to fix it right away for the sake of translators.

kamil-tekiela: If this is the rule, then I can spend some time tomorrow and go through it all replacing all usages of <function> with <methodname>. Most of the PDO manual uses <function> so that will be a lot of change. The only difference I see between these two is that the CSS on the site makes method names blue instead of black, so I don't think it's very crucial and I'd value consistency rather than semantic correctness. It's either done everywhere or none at all.

cmb69: > I see between these two is that the CSS on the site makes method names blue instead of black, […]

Isn't this rather the difference about a link, and a plain mention (either because the function/method is unknown, or because it is on its man-page)?

kamil-tekiela: You can see the difference here https://www.php.net/manual/en/pdo.quote.php Take a look how `PDO::quote()` is highlighted in the warning message and in the text

Issue #814, opened 2021-07-24
Copy/paste schrapnel in ini list


mschroebel:
From manual page: https://php.net/ini.list

Some session settings in middle of memchache partially modified

---

memcache.hash_strategy | "standard" | PHP_INI_ALL | Available since memcache 2.2.0.
memcache.hash_function | "crc32" | PHP_INI_ALL | Available since memcache 2.2.0.

session.save_handler | "files" | PHP_INI_ALL | Supported since memcache 2.1.2 
session.save_path | "" | PHP_INI_ALL | Supported since memcache 2.1.2

memcache.protocol | ascii | >PHP_INI_ALL | Supported since memcache 3.0.0
memcache.redundancy | 1 | >PHP_INI_ALL | Supported since memcache 3.0.0

PR #815, opened 2021-07-24
Remove memcache session.save_handler and .save_path from ini list

cmb69:
`session.save_handler` and `session.save_path` are INI settings of the
session module.  However, the memcache extension offers the memcache
session handler so it makes sense to document these two INI settings
for the memcache extension, but we don't want to include them in the
general INI list, so we split them into a separate "Additional Memcache
Configuration Options" table.

Fixes GH-814.

### DIFF
--- a/reference/memcache/ini.xml
+++ b/reference/memcache/ini.xml
@@ -86,7 +86,7 @@
    </tgroup>
   </table>
   <table>
-   <title>Additional Memcache Configuration Options</title>
+   <title>Session Configuration Options Affecting Memcache Behavior</title>
    <tgroup cols="4">
     <thead>
      <row>
--- a/reference/memcache/ini.xml
+++ b/reference/memcache/ini.xml
@@ -53,18 +53,6 @@
       <entry>Available since memcache 2.2.0.</entry>
      </row>
      <row>
-      <entry><link linkend="ini.session.save-handler">session.save_handler</link></entry>
-      <entry>"files"</entry>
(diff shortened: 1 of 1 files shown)


reference/memcache/ini.xml
```diff
@@ -188,40 +203,34 @@
    </listitem>
   </varlistentry>
 
-  <varlistentry xml:id="ini.memcache.session-handler">
+  <varlistentry xml:id="ini.memcache.protocol">
```
afilina, line 206: Are all the changes below this line simply reordered varlistentry elements?

reference/memcache/ini.xml
```diff
@@ -97,6 +85,33 @@
     </tbody>
    </tgroup>
   </table>
+  <table>
+   <title>Additional Memcache Configuration Options</title>
```
afilina, line 89: I'm not sure that "Additional" makes much sense from a reader's perspective.
```suggestion
   <title>Session Configuration Options Affecting Memcache Behavior</title>
```

afilina (changes requested):

reference/memcache/ini.xml
cmb69: Ah, that's nice!

reference/memcache/ini.xml
cmb69: Yes.

afilina (approved):

PR #816, opened 2021-07-24
Replace <function> with <methodname> in PDO

kamil-tekiela:


Girgias:
This is mostly a QA issue, and although I appreciate this being addressed, could we maybe wait before merging this?

I don't know about other active translation, but the French one is 400+ files behind because I didn't have time for a month (and not sure how much I have for the following one) and knowing what I should prioritise would be great, especially as there are still a couple of big stub syncing PRs which need to land (which also need help being reviewed). And it seems a lot of QA changes did land?

Maybe also create a QA to check this? I know one in doc-base checks for functions tags, and previously did some normalization for method names in `<function>` tags so I would imagine duplicating to be able in the future to run this on CI to make sure those mistake don't continue to happen would be good?

kamil-tekiela:
There's absolutely no pressure here. This PR can wait for as long as needed. It affects nothing visually in the PHP manual. I'm only changing this because @salathe has asked me to use `<methodname>` in a recent PR and I don't like inconsistency.

cmb69:
> There's absolutely no pressure here. This PR can wait for as long as needed.

Except that such PRs tend to become stale. :) Could you please fix the merge conflicts when you have some time?

Girgias:
Yeah I didn't expect this to hang for a year :/ sorry about this.

Crell:
Merci bien!

PR #818, opened 2021-07-25
Mark a few interfaces as oointerface rather than ooclass

kocsismate:
A few more to come when the class synopsis pages are generated from stubs.

Needs https://github.com/php/phd/pull/50/ first!


Girgias (approved):
LGTM (although I really wish we would move Countable out from SPL and into the Predefined Interfaces section, but that's something separate).

PR #822, opened 2021-07-26
Delete redundant sections on mqseries requirements page

zors1:


reference/mqseries/setup.xml
```diff
@@ -19,7 +19,7 @@
    </simpara>
   </note>
   <section xml:id="mqseries.requirements.nix">
-   <title>Installation requirements on non windows platforms</title>
+   <title>Installation requirements on non-Windows platforms</title>
```
cmb69, line 22: Not sure if the hypen should be omitted here, i.e.
```suggestion
   <title>Installation requirements on non Windows platforms</title>
```
Review by native English speakers highly welcome!

salathe: The hyphen is okay.  

However, I would be inclined to put the Windows section before this one, and then title this section "Installation requirements on other platforms".

cmb69: Thank you! However, on a closer look, it doesn't make much sense to have two sections at all:
Screenshot 2021-07-26 143625
And that info doesn't belong into the requirements section at all; this section should be about getting and installing the mqseries library. Furthermore, the extension is not supported on Windows at all ([no config.w32](https://github.com/php/pecl-networking-mqseries)).

zors1: Thanks for the feedback! I'll delete these sections.

salathe (approved):
Perfect, thanks @zors1

PR #821, opened 2021-07-26
Update shell_exec doc

VincentLanglet:
Related to https://github.com/php/php-src/pull/7306#issuecomment-886489235

`shell_exec` can return `false`.

PR #823, opened 2021-07-26
additional return value fix of shell_exec func.

mumumu:
#821 newly added return `&false;` in case of pipe establish error, so we need to add it to signatures.

### DIFF
--- a/reference/exec/functions/shell-exec.xml
+++ b/reference/exec/functions/shell-exec.xml
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>string</type><methodname>shell_exec</methodname>
+   <type class="union"><type>string</type><type>false</type><type>null</type></type><methodname>shell_exec</methodname>
    <methodparam><type>string</type><parameter>cmd</parameter></methodparam>
   </methodsynopsis>
   <para>


VincentLanglet:
`null` is also a supposed returned value. Should it be added too ?

salathe:
> `null` is also a supposed returned value. Should it be added too ?

Yes, `null` should be added.

Also, have these been the return types "foverver" (i.e. since whatever the minimum version of PHP that is documented in the manual)? If not, a changelog will be required.

VincentLanglet:
> Also, have these been the return types "foverver" (i.e. since whatever the minimum version of PHP that is documented in the manual)? If not, a changelog will be required.

When I checked, `null` and `false` was already returned in the 7.0 branch.

salathe:
@VincentLanglet thanks for checking, in that case no changelog necessary.
@mumumu thanks for the update.  Looks good to me now. 👍

mumumu:
> Also, have these been the return types "foverver" (i.e. since whatever the minimum version of PHP that is documented in the manual)? If not, a changelog will be required.

I also checked first commit, shell_exec returns false and string :).

https://github.com/php/php-src/commit/257de2baded9330ff392f33fd5a7cc0ba271e18d#diff-a65df53d01f06a96c93e82741c21578c745fde2d972e3de3111e80fa8e2278c4R340-R380

> thanks for the update. Looks good to me now.

OK. I'll merge this PR.
Thank you for your comments, @VincentLanglet @salathe.

mumumu:
[memo] stub is already correct.

https://github.com/php/php-src/blob/bed71393755a84b29eed89327f36e3b6f487ba2b/ext/standard/basic_functions.stub.php#L756

cmb69:
See also https://github.com/php/doc-en/pull/289/files#diff-6cba833390b02f0d0d75f87be96a4bb5d80ad97e57e582790288f4abe3215e87.

PR #828, opened 2021-07-29
Fold in comments to array_map()

Crell:
Also added some types to the examples along the way, because types.

### DIFF
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -385,7 +385,7 @@ array(1) {
     <title><function>array_map</function> - associative arrays</title>
     <para>
      While <function>array_map</function> does not directly support
-     using the array key as an input, that may be simulated using <function>array_keys</function>
+     using the array key as an input, that may be simulated using <function>array_keys</function>.
     </para>
     <programlisting role="php">
 <![CDATA[
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -22,7 +22,9 @@
    used as arguments for the callback.
    The number of parameters that the <parameter>callback</parameter>
    function accepts should match the number of arrays
-   passed to <function>array_map</function>.
+   passed to <function>array_map</function>. If it accepts fewer, excess
+   input arrays will be ignored. If it accepts more, an <classname>ArgumentCountError</classname>
+   is thrown.
   </para>
  </refsect1>
 
@@ -122,11 +124,16 @@ Array
     <programlisting role="php">
 <![CDATA[
 <?php
-$func = function($value) {
+$func = function(int $value): int {
     return $value * 2;
 };
 
 print_r(array_map($func, range(1, 5)));
+
+// Or as of PHP 7.4.0:
+
+print_r(array_map(fn($value): int => $value * 2, range(1, 5)));
+
 ?>
 ]]>
     </programlisting>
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-map.xml
```diff
@@ -22,7 +22,9 @@
    used as arguments for the callback.
    The number of parameters that the <parameter>callback</parameter>
    function accepts should match the number of arrays
-   passed to <function>array_map</function>.
+   passed to <function>array_map</function>. If it accepts fewer, excess
+   input arrays will be ignored. If it accepts more, an <classname>ArgumentCountError</classname>
+   is thrown.
```
kamil-tekiela, line 27: I'm dumb and it took me 2 minutes to understand this sentence. Could we simplify this for non-native speakers?

How about this?
```suggestion
   passed to <function>array_map</function>. Excess
   input arrays are ignored. An <class>ArgumentCountError</class>
   is thrown if an insufficient number of arguments is provided.
```

Also, the previous sentence uses the present tense, so let's stick to using the present tense.

reference/array/functions/array-map.xml
```diff
@@ -371,6 +378,43 @@ array(1) {
...
   }
 }
+]]>
+    </screen>
+   </example>
+   <example>
+    <title><function>array_map</function> - associative arrays</title>
+    <para>
+     While <function>array_map</function> does not directly support
+     using the array key as an input, that may be simulated using <function>array_keys</function>
```
kamil-tekiela, line 388: Missing period at the end. Why not use `this` instead of `that`?

Crell: As a native speaker, "this" feels clumsier in context.  Both are grammatically valid, but "this" sounds worse to my American ear.

reference/array/functions/array-map.xml
Crell: I wouldn't call you dumb. :smile:  I'm not sure why it's unclear, though.  It's a series of if-then statements.  That should be self-evident to developers, no?  And it's already in present tense...

cmb69: At least there is a mismatch between "will be ignored" and "is thrown". :)

kamil-tekiela: Oops seems my suggestion clashed with your initial forced push. Class -> classname

reference/array/functions/array-map.xml
```diff
@@ -371,6 +378,43 @@ array(1) {
...
+    <programlisting role="php">
+<![CDATA[
+<?php
+$arr = [
+    'v1' => 'First release',
+    'v2' => 'Second release',
+    'v3' => 'Third release',
+];
+
+$callback = fn(string $k, string $v): string => "$k was the $v";
```
cmb69, line 399: Maybe either use an anonymous function, or note that arrow functions are only available as of PHP 7.4.0.

Crell: Arrow functions are anonymous functions. :smile: A previous example already offers both versions (also updated in this PR), and within a few months all supported language versions will include short lambdas.  Is there a need to repeat it?

cmb69: The manual is still supposed to document PHP 7.0.0 and up, and most of these versions don't understand arrow functions. IMO, a short comment makes sense.

Crell: Comment added.

cmb69 (approved):
Thank you!

PR #831, opened 2021-07-30
Fold the note on PCRE comments

kamil-tekiela:


reference/pcre/pattern.syntax.xml
```diff
@@ -2161,6 +2161,41 @@
...
+$match = preg_match('/te(?# this is a comment)st/', $subject);
+var_dump($match);
+
+/* Whitespace and # is treated as part of the pattern unless PCRE_EXTENDED is enabled */
+$match = preg_match('/te   #~~~~
+st/', $subject);
+var_dump($match);
+
+/* When PCRE_EXTENDED is enabled, all whitespace data characters and anything
+   that follows an unescaped # in the pattern is ignored */
```
cmb69, line 2183: Maybe:
```suggestion
   that follows an unescaped # on the same line is ignored */
```

cmb69 (commented):
Thank you! Generally good, but the meaning of `#` might need clarification.

kamil-tekiela:
>the meaning of # might need clarification.
Is the text in the paragraph leading to the example not enough? What exactly do you have in mind?

cmb69 (approved):
> What exactly do you have in mind?

I was just referring to my suggestion. :)

PR #836, opened 2021-08-03
Update versions.xml for UOPZ 7.1.0

cmb69:
Cf. <https://github.com/krakjoe/uopz/issues/152#issuecomment-891582155>.

---

This version is not yet released, so this PR has to wait.


cmb69:
[uopz 7.1.0](https://pecl.php.net/package-info.php?package=uopz&version=7.1.0) has been released.

PR #837, opened 2021-08-03
Add T_ATTRIBUTE to list of parser tokens

tiffany-taylor:


appendices/tokens.xml
```diff
@@ -72,11 +72,16 @@ defined('T_FN') || define('T_FN', 10001);
...
-     <row>
+    <row>
      <entry><constant>T_AS</constant></entry>
      <entry>as</entry>
      <entry>&foreach;</entry>
     </row>
+    <row>
+     <entry><constant>T_ATTRIBUTE</constant></entry>
+     <entry>#[</entry>
+     <entry><link linkend="language.attributes">attributes</link></entry>
```
cmb69, line 83: Should also specify availability:
```suggestion
     <entry><link linkend="language.attributes">attributes</link> (as of PHP 8.0.0)</entry>
```

appendices/tokens.xml
```diff
@@ -72,11 +72,16 @@ defined('T_FN') || define('T_FN', 10001);
...
-     <row>
+    <row>
      <entry><constant>T_AS</constant></entry>
      <entry>as</entry>
      <entry>&foreach;</entry>
     </row>
+    <row>
+     <entry><constant>T_ATTRIBUTE</constant></entry>
+     <entry>#[</entry>
+     <entry><link linkend="language.attributes">attributes</link> (as of PHP 8.0.0)</entry>
```
kamil-tekiela, line 83:
```suggestion
     <entry><link linkend="language.attributes">attributes</link> (available since PHP 8.0.0)</entry>
```

kamil-tekiela (approved):
LGTM, but the page uses `available since` so to keep it consistent better do that here as well

cmb69:
> LGTM, but the page uses available since so to keep it consistent better do that here as well

Yes, I've seen that, but why add more wording which explicitly goes against the [style guide](http://doc.php.net/tutorial/style.php)? If at all, I'd fix the "since"s.

tiffany-taylor:
I agree with @cmb69, but I will have to make that change when I return to my other desk.

tiffany-taylor:
I've changed instances of "available since" to "as of," but "as of" feels clumsy in parentheses. Should it maybe be changed to a comma? Like `attributes, as of PHP 8.0.0`? Or maybe a separate column listing version availability? (this was done in another table, I vaguely remember...)

kamil-tekiela:
In my opinion, "available as of" sounds better. The style guide only asks to avoid the word "since"

PR #838, opened 2021-08-04
Remove abstract modifier from ArrayAccess methods

othercorey:
https://www.php.net/manual/en/class.arrayaccess.php

Now that ArrayAccess is shown as an interface, the `abstract` modifier doesn't make sense. This suggests that interfaces can/should use the abstract modifier when it isn't allowed.

```
abstract public offsetExists(mixed $offset): bool
abstract public offsetGet(mixed $offset): mixed
abstract public offsetSet(mixed $offset, mixed $value): void
abstract public offsetUnset(mixed $offset): void
```

### DIFF
--- a/language/predefined/arrayaccess/offsetexists.xml
+++ b/language/predefined/arrayaccess/offsetexists.xml
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>abstract</modifier> <modifier>public</modifier> <type>bool</type><methodname>ArrayAccess::offsetExists</methodname>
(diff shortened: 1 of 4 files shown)


kocsismate (approved):

salathe:
Note, this is how (almost) all interface method definitions are marked up in the manual.  If we're going to remove the `abstract` modifier, we might as do it everywhere.

kocsismate:
> Note, this is how (almost) all interface method definitions are marked up in the manual. If we're going to remove the abstract modifier, we might as do it everywhere.

Yes, it's a known gotcha since the method synopsis generation project was started: https://github.com/php/doc-en/pull/193#discussion_r525116336 I always had to manually fixup the abstract modifiers until now. Fortunately, the unnecessary abstract modifiers will go away as soon as I sync the docs with the recent changes next time.

PR #841, opened 2021-08-06
Clarification of return values descriptions for mysqi fetch functions

afilina:
I fixed the merge conflicts between the two PRs #39 and #207 as best I could. @colshrapnel did not enable changes from maintainers, so I had to open a new PR instead of appending to the existing one.

Closes #39

### DIFF
--- a/reference/mysqli/mysqli_result/fetch-object.xml
+++ b/reference/mysqli/mysqli_result/fetch-object.xml
@@ -29,9 +29,11 @@
   </para>
   <para>
    If two or more columns of the result have the same name, the last
-   column will take precedence. To access the other column(s) of the same
-   name, you either need to access the result with numeric indices by using
-   <function>mysqli_fetch_row</function> or add alias names to the SQL query.
+   column will take precedence and overwrite any previous data. To
+   access multiple columns with the same name,
+   <function>mysqli_fetch_row</function> may be used to fetch the numerically
+   indexed array, or aliases may be used in the SQL query select list to give
+   columns different names.
   </para>
   <note xmlns="http://docbook.org/ns/docbook">
    <simpara>
--- a/reference/mysqli/mysqli_result/fetch-object.xml
+++ b/reference/mysqli/mysqli_result/fetch-object.xml
@@ -79,7 +79,7 @@
    represents the name of the result set's column, or &null;
    if there are no more rows in the result set.
   </para>
-  </refsect1>
+ </refsect1>
 
  <refsect1 role="changelog">
   &reftitle.changelog;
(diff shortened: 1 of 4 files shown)


reference/mysqli/mysqli_result/fetch-row.xml
```diff
@@ -40,7 +40,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   <function>mysqli_fetch_row</function> returns an array of values that corresponds to the fetched row
+   <function>mysqli_fetch_row</function> returns an enumerated array representing the fetched row
```
cmb69, line 43: I think the term "indexed array" is more common elsewhere in the manual, but the MySQLi docs appear to prefer "enumerated array", so this is fine.

reference/mysqli/mysqli_result/fetch-object.xml
```diff
@@ -67,10 +67,18 @@
...
-   Returns an object that corresponds to the fetched
-   row or &null; if there are no more rows in result set.
+   Returns an object representing the fetched row, where each property
+   represents the name of the result set's column or &null;
+   if there are no more rows in the result set.
   </para>
- </refsect1>
+  <para>
+   If two or more columns of the result have the same name, the last
+   column will take precedence. To access the other column(s) of the same
```
kamil-tekiela, line 76: I am not convinced about "the other" as a determiner in this sentence here. (see https://dictionary.cambridge.org/grammar/british-grammar/other-others-the-other-or-another) but I am not a native speaker, so maybe this is ok? However, this is how it is currently written on the fetch_assoc page.

kamil-tekiela (commented):
The changes look good, but I am only concerned about the ease of reading. I think the comma should be placed there to separate the thoughts in the sentence, and if someone could confirm whether "the other" is correct as the determiner in this case, it would be great.

reference/mysqli/mysqli_result/fetch-object.xml
afilina: In this case, "the other" refers to a set that excludes the element previously mentioned, so this looks ok to me.

kamil-tekiela (approved):

kamil-tekiela:
Can you please add "Co-authored-by" with the original author name, please?

reference/mysqli/mysqli_result/fetch-object.xml
salathe: This paragraph (when viewed as a whole) is misplaced in the "Return Values" section.  

The "Return Values" section should not contain any direction about what to do if the described method or function does not do what the reader wants: it should stick to describing the return values.  

Directing readers to other, more appropriate for their needs, functions should be in a "Note" (best for short, to-the-point notes)  or a paragraph in the "Description" section.

afilina: @colshrapnel ^

pthreat: "remaining" instead of the other would be better

salathe (approved):

PR #842, opened 2021-08-06
Fix fetchAll synopsis

kamil-tekiela:
Split off from #596.


sy-records (approved):

cmb69:
I think that needs further fixes on the page (look for `<parameter>colno</parameter>` etc.)

kamil-tekiela:
@cmb69 Oh sorry, I missed that one. Do you see just that one or did I miss any more? I will commit it right away.

cmb69:
Ah, right, only one occurrence. Thanks for catching up! :)

PR #844, opened 2021-08-07
Updating mysqli: set_charset

kamil-tekiela:
I am not sure if the first note on this page is still relevant.

### DIFF
--- a/reference/mysqli/mysqli/set-charset.xml
+++ b/reference/mysqli/mysqli/set-charset.xml
@@ -4,7 +4,7 @@
  <refnamediv>
   <refname>mysqli::set_charset</refname>
   <refname>mysqli_set_charset</refname>
-  <refpurpose>Sets the default client character set</refpurpose>
+  <refpurpose>Sets the client character set</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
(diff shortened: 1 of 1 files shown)


zors1:
I know that it's not directly related to this PR, but isn't `&style.oop;` missing a hyphen?

image

kamil-tekiela:
@zors1 Can you explain where do you think the hyphen is missing? What should it look like with a hyphen?

zors1:
>  Do you mean that is should be "Object-oriented style"?

Yes, I'm sorry my previous comment wasn't clear enough.

kamil-tekiela:
Then, yes. It should be "Object-oriented", but that's not part of this PR.

reference/mysqli/mysqli/set-charset.xml
```diff
@@ -35,7 +35,7 @@
      <term><parameter>charset</parameter></term>
      <listitem>
       <para>
-       The charset to be set as default.
+       The character set to be set.
```
afilina, line 38: Maybe a nitpick, but having the word "set" twice makes it a bit harder to read. Perhaps "The desired character set." or "The character set to be used". We already say "Sets the character set to be used" in the description; the second option merely reiterates that in context.

afilina (changes requested):
Other than the one comment, LGTM.

reference/mysqli/mysqli/set-charset.xml
kamil-tekiela: I agree. I used "The desired character set."

afilina (approved):

Issue #849, opened 2021-08-13
array_combine throws ValueError since PHP 8 


olsavmic:
From manual page: https://php.net/function.array-combine

---

Since PHP 8, array_combine function throws ValueError instead of returning `false` and E_WARNING.

This behaviour is not documented anywhere on the PHP 8 migration guide (https://www.php.net/manual/en/migration80.php) nor the array_combine documentation.

Perhaps a change log should be added similarly to https://www.php.net/manual/en/function.round.php.

sy-records:
Thank you for your feedback.

PR #850, opened 2021-08-13
Fix array_combine throws ValueError since PHP 8

sy-records:
close https://github.com/php/doc-en/issues/849


cmb69:
See also PR #163.

Crell:
Thanks!

PR #851, opened 2021-08-13
Changed name of the capital of Ukraine to the correct one

pro100svitlo:
The capital of Ukraine is called **Kyiv**, not Kiev.
Please respect Ukraine as a country and all people that live there 🙏 

Proofs:
- [wiki](https://en.wikipedia.org/wiki/Kyiv)
- [another wiki](https://en.wikipedia.org/wiki/KyivNotKiev)
- [euronews](https://www.euronews.com/2019/06/21/kyiv-or-kiev-why-does-it-matter-so-much-to-ukrainians)
- [atlanticcouncil](https://www.atlanticcouncil.org/blogs/ukrainealert/kyiv-not-kiev-why-spelling-matters-in-ukraines-quest-for-an-independent-identity/)
- [nytimes](https://www.nytimes.com/2019/11/13/us/politics/kiev-pronunciation.html)
- [stanfordpolitics](https://stanfordpolitics.org/2020/01/06/kyiv-vs-kiev-what-two-letters-mean-for-ukrainian-independence/)

### DIFF
--- a/reference/datetime/timezones.xml
+++ b/reference/datetime/timezones.xml
@@ -610,7 +610,7 @@
      </row>
      <row>
       <entry>Europe/Kaliningrad</entry>
-      <entry>Europe/Kiev</entry>
+      <entry>Europe/Kyiv</entry>
       <entry>Europe/Kirov</entry>
       <entry>Europe/Lisbon</entry>
      </row>


nikic:
PHP uses timezone names specified by IANA (see http://www.iana.org/time-zones). Please contact them if you wish to request a timezone name change. PHP will automatically adopt the new name after it has been changed.

cmb69:
This is a well known issue, but certainly the PHP documentation cannot be changed, but it would lie then, and even the PHP timezone DB cannot be changed, because the data are retrieved (and constantly updated) from the [IANA timezone DB](https://www.iana.org/time-zones), and [their stance is that this will not change](https://github.com/eggert/tz/blob/main/europe#L3761-L3772).

pro100svitlo:
@cmb69 thanks for letting me know.
Sure, I will ask them again 😉 

Have a good day!

PR #852, opened 2021-08-15
Fix typo in example function call

azeemhassni:
# Description

Looks like the example function name and the function call do not match, if it's an oversight this commit fixes it.

image


cmb69:
Thank you!

PR #853, opened 2021-08-16
clarify response type for Memcached::deleteMulti

lauripiisang:
- the documentation does not clearly say that response 
  codes will be returned, although the implementation
  does exactly that.
 - see https://github.com/php-memcached-dev/php-memcached/blob/605a8a63a7461983512ce12601e995dad18f5a9b/php_memcached.c#L2278-L2282


tstarling:
I think a few problems still remain after this commit, so I made #1957 as an alternative.

afilina:
@tstarling Did I correctly understand that https://github.com/php/doc-en/pull/1957 makes this PR no longer relevant, and therefore can be closed?

tstarling:
Yes, the issue was addressed in #1957.

PR #863, opened 2021-08-18
Fix apcu_key_info return value

kamil-tekiela:
As per https://github.com/krakjoe/apcu/blob/b68953ccbc1a5234c2caf8ed5800afd222270948/php_apc.stub.php#L12

### DIFF
--- a/reference/apcu/functions/apcu-key-info.xml
+++ b/reference/apcu/functions/apcu-key-info.xml
@@ -11,7 +11,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>array</type><methodname>apcu_key_info</methodname>
+   <type class="union"><type>array</type><type>null</type></type><methodname>apcu_key_info</methodname>
    <methodparam><type>string</type><parameter>key</parameter></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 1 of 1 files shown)


reference/apcu/functions/apcu-key-info.xml
```diff
@@ -38,7 +38,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   null or array info
+   An array containing the detailed information about the cache key, or &null; on failure.
```
nikic, line 41:
```suggestion
   An array containing the detailed information about the cache key, or &null; if the key does not exist.
```
Maybe?

cmb69: Maybe both? While a non existing key is likely the cause for `null` return, `null` is also returned if locking of the cache fails: https://github.com/krakjoe/apcu/blob/b68953ccbc1a5234c2caf8ed5800afd222270948/apc_cache.c#L1145-L1147

nikic: Locking can't really fail, it's just implemented defensively. Possibly it should just throw a fatal error instead.

cmb69: Ah, okay, than I think we're good here. :)

cmb69 (approved):
Thank you!

PR #869, opened 2021-08-22
Point the readme to the doc-base repository

deleugpn:
Looks like the readme assume an old structure that is no longer in place? I just started reading the `doc-base` readme and it seems to be the best place to get started


Girgias:
Right, it might make sense that the git clone command should be ``git clone git@github.com:php/doc-en.git en`` so that it's possible to use the configure script in doc-base without needing to create a symlink from ``doc-en`` to ``en``

README
```diff
@@ -2,8 +2,4 @@
 README
 ----------------------------------------------------------------------
 
-Please refer to the README file within the doc-base/ directory. Your
-directory structure is probably like:
-
-  phpdoc/en            (DocBook files for the English language)
-  phpdoc/doc-base      (Tools and resources used for all languages)
+Please refer to the README file within the [doc-base repository](https://github.com/php/doc-base).
```
afilina, line 5:
```suggestion
Please refer to the README file within the [doc-base repository](https://github.com/php/doc-base/blob/master/README.md).
```

PR #870, opened 2021-08-22
Add information about leading and trailing whitespace

deleugpn:
This is how the changes look like. Closes https://bugs.php.net/bug.php?id=79111

image

### DIFF
--- a/reference/var/functions/is-numeric.xml
+++ b/reference/var/functions/is-numeric.xml
@@ -155,7 +155,7 @@ foreach ($tests as $element) {
      <row>
       <entry>8.0.0</entry>
       <entry>
-       Numeric strings ending with whitespace ("42 ") will now return true.
+       Numeric strings ending with whitespace ("42 ") will now return &true;.
       </entry>
      </row>
     </tbody>
--- a/reference/var/functions/is-numeric.xml
+++ b/reference/var/functions/is-numeric.xml
@@ -104,6 +104,63 @@ NULL is NOT numeric
     </screen>
    </example>
   </para>
+
+  <para>
+   <example>
+    <title><function>is_numeric</function> with whitespace</title>
+    <programlisting role="php">
+     <![CDATA[
+<?php
+$tests = [
+    " 42",
+    "42 ",
+    " 9001", // non-breaking space
+    "9001 ", // non-breaking space
+];
+
+foreach ($tests as $element) {
+    if (is_numeric($element)) {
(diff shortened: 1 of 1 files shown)


reference/var/functions/is-numeric.xml
```diff
@@ -104,6 +104,63 @@ NULL is NOT numeric
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       Numeric strings ending with whitespace ("42 ") will now return true.
```
zors1, line 158: `true` should be replaced with `&true;`

reference/var/functions/is-numeric.xml
```diff
@@ -104,6 +104,63 @@ NULL is NOT numeric
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       Numeric strings ending with whitespace ("42 ") will now return &true;.
```
zors1, line 158: I forgot, `"42 "` should be wrapped in `<literal>`.
```suggestion
       Numeric strings ending with whitespace (e.g. <literal>"42 "</literal>) will now return &true;.
```

reference/var/functions/is-numeric.xml
```diff
@@ -104,6 +104,63 @@ NULL is NOT numeric
...
+    </programlisting>
+    &example.outputs;
+    <screen>
+     <![CDATA[
+' 42' is numeric
+'42 ' is numeric
+' 9001' is NOT numeric
+'9001 ' is NOT numeric
+]]>
+    </screen>
```
cmb69, line 139: This is the result for PHP 8, and we should clarify that. At least use `&example.outputs.8;` instead of `&example.outputs;`, or maybe better, show the results for PHP 7 as well.

reference/var/functions/is-numeric.xml
cmb69: Done.

reference/var/functions/is-numeric.xml
cmb69: Done.

cmb69:
Thanks for the PR! I've addressed the review comments, and merged.

PR #871, opened 2021-08-22
Update construct.xml

WinterSilence:
Updated by `ReflectionMethod`


cmb69:
Actually, `SimpleXMLElement::__construct()` is no longer final as of PHP 8.0.0 (https://3v4l.org/vp7BU), so there should *only* be a changelog entry, but no change to the signature.

afilina:
Pinging author: @WinterSilence

PR #872, opened 2021-08-22
Update simplexmlelement.xml

WinterSilence:
Updated by `ReflectionClass`

### DIFF
--- a/reference/simplexml/simplexmlelement.xml
+++ b/reference/simplexml/simplexmlelement.xml
@@ -27,9 +27,13 @@
      <ooclass>
       <classname>SimpleXMLElement</classname>
      </ooclass>
-     
+
      <oointerface>
       <interfacename>Traversable</interfacename>
+     </oointerface>
+
+     <oointerface>
+      <interfacename>Countable</interfacename>
      </oointerface>
     </classsynopsisinfo>
 <!-- }}} -->


saundefined (commented):
Thanks for PR!

PHP version for documentation is 7.0, `Countable` implemented later.
Before PHP 7.3 `is_countable(new SimpleXMLElement())` throws an exception.

Perhaps a changelog or note should also be added?
@cmb69 what do you think?

cmb69:
Yes, there should be a changelog entry (I think it's Countable as of PHP 7.4.0, only). Also, as of PHP 8.0.0, the signature is actually `class SimpleXMLElement implements Stringable, Countable, RecursiveIterator`; that should be documented in the first place, and respective changelog entries should be added. It might be worthwhile to wait how the semi-automated classsynopsis updates (PR #839) will go.

WinterSilence:
@saundefined 
> Before PHP 7.3 is_countable(new SimpleXMLElement()) throws an exception.

but `Countable` added only in 7.4... :rofl

WinterSilence:
@cmb69 
> It might be worthwhile to wait how the semi-automated classsynopsis updates (PR #839) will go.

as after 7.4? 🤣

cmb69:
To clarify: SimpleXMLElement supported `count()` forever; prior to PHP 7.4.0, this was done due to an internal mechanism; as of PHP 7.4.0, it properly implements `Countable`.

Regarding PR #839: I hope we can resolve this soon. If so, semi-automatic updates can be done for all classes, so that this PR would be obsolete.

PR #877, opened 2021-08-25
CI: Render XHTML chunked manual

cmb69:
The rendered XHTML chunked manual is provided as artifact.

---

I think this is useful for those who don't have a full doc build environment set up, but it slows down CI considerably, and it's still not possible to view the docs online. Maybe we could push to GH pages or something, and the full workflow should only be run when manually triggered?

Thoughts welcome!

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -39,14 +39,32 @@ jobs:
           path: "doc-base"
           repository: "php/doc-base"
 
+      - name: "Checkout php/phd"
+        uses: "actions/checkout@v2"
+        with:
+          path: "phd"
+          repository: "php/phd"
+
       - name: "Quality Assurance scripts"
         run: "php8.0 doc-base/scripts/qa/extensions.xml.php --check"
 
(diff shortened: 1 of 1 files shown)


saundefined:
Deploing preview on GH-pages (or similar) sounds great ([example](https://github.com/facebook/docusaurus/pull/5410#issuecomment-903799908)) 🎉 

But maybe enable it only for PRs,
and for commits add action just to trigger [`build-docs-lang-rsync`](https://github.com/php/systems/blob/master/build-docs-lang-rsync) (or keep it on cron)?

afilina:
Suggestion from PR https://github.com/php/doc-en/pull/1608#issuecomment-1147457689
> Or provide a dockerized build command. Then you have 1 requirement that almost any developer will have installed or can easily install.

If this is a direction that we're willing to take, I'd be happy to do it. I was actually thinking about it this week when I saw how complicated the setup is for a first-time contributor.

cmb69:
> If this is a direction that we're willing to take, I'd be happy to do it.

Feel free to go ahead. :)

PR #881, opened 2021-08-26
Locale::lookup default value for `$default` is empty string

PowerKiKi:


afilina:
@cmb69 Can you clarify your comment? Is there something that the author is expected to do, or is it a suggestion for other changes that can be made in a separate PR?

cmb69:
I was especially referring to https://bugs.php.net/bug.php?id=70585#1628868244, which suggest further changes to the manual, but indeed that could be fixed by another commit. However, this PR is not quite right anyway, and the signature had been fixed in the meantime (thus the merge conflict), so I'm closing this ticket.

PR #885, opened 2021-08-28
Add documentation for var_representation PECL

TysonAndre:
Information about the extension can be found at https://github.com/TysonAndre/var_representation https://pecl.php.net/var_representation - it was proposed at https://marc.info/?l=pecl-dev&m=162400655627665&w=2 in June.

### DIFF
--- a/reference/var_representation/book.xml
+++ b/reference/var_representation/book.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<book xml:id="book.var_representation" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <?phpdoc extension-membership="pecl" ?>
+ <title>var_representation</title>
+ <titleabbrev>var_representation</titleabbrev>
+
+ <preface xml:id="intro.var_representation">
+  &reftitle.intro;
+  <simpara>
+   The var_representation extension provides a compact alternative to <function>var_export</function> that properly escapes control characters.
+  </simpara>
+ </preface>
+
+ &reference.var-representation.setup;
+ &reference.var-representation.constants;
+ &reference.var-representation.reference;
+
+</book>
+
+<!-- Keep this comment at the end of the file
+Local variables:
+mode: sgml
+sgml-omittag:t
+sgml-shorttag:t
+sgml-minimize-attributes:nil
+sgml-always-quote-attributes:t
+sgml-indent-step:1
+sgml-indent-data:t
+indent-tabs-mode:nil
+sgml-parent-document:nil
+sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+sgml-exposed-tags:nil
+sgml-local-catalogs:nil
+sgml-local-ecat-files:nil
+End:
+vim600: syn=xml fen fdm=syntax fdl=2 si
+vim: et tw=78 syn=sgml
+vi: ts=1 sw=1
+-->
--- a/reference/var_representation/constants.xml
+++ b/reference/var_representation/constants.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appendix xml:id="var-representation.constants" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ &reftitle.constants;
+ &extension.constants;
+ <para>
+  <variablelist>
+   <varlistentry xml:id="constant.var-representation-single-line">
+    <term>
+     <constant>VAR_REPRESENTATION_SINGLE_LINE</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      <function>var_representation</function> flag indicating that newlines 
+      should not be used for whitespace in variable representations.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.var-representation-unescaped">
+    <term>
+     <constant>VAR_REPRESENTATION_UNESCAPED</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      <function>var_representation</function> flag indicating that strings should be encoded
+      as single quoted strings with unescaped control characters.
+     </simpara>
+    </listitem>
+   </varlistentry>
+  </variablelist>
+ </para>
+</appendix>
+<!-- Keep this comment at the end of the file
+Local variables:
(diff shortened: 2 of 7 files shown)


reference/var_representation/constants.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appendix xml:id="var-representation.constants" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ &reftitle.constants;
+ &extension.constants;
+ <para>
+  <variablelist>
+   <varlistentry xml:id="constant.var-representation-single-line">
+    <term>
+     <constant>VAR_REPRESENTATION_SINGLE_LINE</constant>
+     (<type>integer</type>)
```
Girgias, line 10: Should be
```suggestion
     (<type>int</type>)
```
here and bellow

reference/var_representation/book.xml
```diff
@@ -0,0 +1,41 @@
...
+
+<book xml:id="book.var_representation" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <?phpdoc extension-membership="pecl" ?>
+ <title>var_representation</title>
+ <titleabbrev>var_representation</titleabbrev>
+
+ <preface xml:id="intro.var_representation">
+  &reftitle.intro;
+  <simpara>
+   The var_representation extension provides a compact alternative to var_export that properly escapes control characters.
```
Girgias, line 12:
```suggestion
   The var_representation extension provides a compact alternative to <function>var_export</function> that properly escapes control characters.
```

reference/var_representation/book.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+
```
Girgias, line 3: This tag is now unnecessary

reference/var_representation/functions/var-representation.xml
```diff
@@ -0,0 +1,222 @@
...
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>value</parameter></term>
+    <listitem>
+     <para>
+       The variable you want to generate a representation of.
```
Girgias, line 28:
```suggestion
       The variable to generate a representation of.
```
Please avoid the usage of "you".

reference/var_representation/functions/var-representation.xml
```diff
@@ -0,0 +1,222 @@
...
+    <screen>
+<![CDATA[
+"Content-Length: 123\r\n"
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title>Exporting stdClass</title>
```
Girgias, line 102:
```suggestion
    <title>Exporting <classname>stdClass</classname></title>
```

Girgias (changes requested):
Couple of nits.

Also wondering if the comments at the end of each file as still necessary.

reference/var_representation/book.xml
TysonAndre: Will fix. This is automatically generated - the scripts in doc-base should be updated to remove $Revision$, $Id$, etc.

reference/var_representation/functions/var-representation.xml
TysonAndre: This was based on https://www.php.net/manual/en/function.var-export.php#refsect1-function.var-export-parameters so I forgot about that writing rule

TysonAndre:
> Also wondering if the comments at the end of each file as still necessary.

It's what doc-base emits in the skeleton so I consider changing that a separate discussion/project needing more input. `.editorconfig` would work on more IDEs (and I see one's already added) but you'd have to install a plugin for it in vim, though I've already done that.

Maybe some of the regular commiters use the settings, but I'm not sure if it's making much of a difference other than indentation.

In some cases the `sw=1` common settings are going to save time preventing bad identation if a new contributor uses vim/a vim-aware IDE

Girgias:
> > Also wondering if the comments at the end of each file as still necessary.
> 
> It's what doc-base emits in the skeleton so I consider changing that a separate discussion/project needing more input. `.editorconfig` would work on more IDEs (and I see one's already added) but you'd have to install a plugin for it in vim, though I've already done that.
> 
> Maybe some of the regular commiters use the settings, but I'm not sure if it's making much of a difference other than indentation.
> 
> In some cases the `sw=1` common settings are going to save time preventing bad identation if a new contributor uses vim/a vim-aware IDE

Oh for sure I wasn't saying that it should be done now, but it's an interesting discussion to have if/when we get a way to skip commits for revcheck.

However can you run the ``doc-base/scripts/qa/extensions.xml.php`` QA script so that the extension gets listed in the appendice (this is what is causing the build failure)

reference/var_representation/functions/var-representation.xml
```diff
@@ -0,0 +1,222 @@
...
+<refentry xml:id="function.var-representation" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>var_representation</refname>
+  <refpurpose>Returns a short, readable, parsable string representation of a variable</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type class="union">string</type><methodname>var_representation</methodname>
```
Girgias, line 11:
```suggestion
   <type>string</type><methodname>var_representation</methodname>
```
Probably? Not exactly sure how those are meant to work any more @cmb69 is it fine to have a single type in a union?

TysonAndre: Removed, misread something else when basing this change on it

TysonAndre:
> However can you run the doc-base/scripts/qa/extensions.xml.php QA script so that the extension gets listed in the appendice (this is what is causing the build failure)

Done. doc-base is the next cause of the new build failure (which I believe would be fixed by https://github.com/php/doc-base/pull/43)

```
> 
ERROR (/home/runner/work/doc-en/doc-en/doc-base/manual.xml:378:0)
  <set xml:id="refs.webservice">
^
 IDREF attribute linkend references an unknown ID "book.var_representation"
```

TysonAndre:
Review comments have been addressed.

Are there any objections or any more remaining tasks before this and the doc-base PR can be merged?

cmb69:
Thank you!

PR #887, opened 2021-08-28
Updating mysqli: data_seek

kamil-tekiela:
I have provided a better example. The secondary example should replace the [user note](https://www.php.net/manual/en/mysqli-result.data-seek.php#85629).


cmb69:
Thank you!

PR #890, opened 2021-08-29
Updated examples of casting to boolean

jorgsowa:
Updated examples of casting to boolean adding one case that is pretty often appearing in code: string `"0"`.


cmb69:
Thank you!

PR #897, opened 2021-09-03
Add missing documentation for value parameter of numfmt_parse

darthsteven:
First contribution, so go easy on me!

Seems like we're missing the documentation for the `$value` parameter from this function, which has confused at least my colleague @anotherjames that pointed this out to me.


cmb69:
Thank you!

PR #898, opened 2021-09-03
Add a separate example section for escaping of characters

dryabov:
See merged PR https://github.com/php/php-src/pull/7420 (https://github.com/php/php-src/commit/d3a6054d4441a6d431e6ca7aa63de62cee5585c2)

### DIFF
--- a/reference/filesystem/functions/parse-ini-file.xml
+++ b/reference/filesystem/functions/parse-ini-file.xml
@@ -250,15 +250,64 @@ negative_two = ~1
 ; () is used for grouping
 seven = (8|7)&(6|5)
 
-; \ is used to escape a value.
-newline_is = "\\n" ; results in the string "\n", not a newline character.
-with quotes = "She said \"Exactly my point\"." ; Results in a string with quote marks in it.
-
 path = ${PATH}
 also_five = ${five}
 
 ]]>
     </programlisting>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title>escaping characters</title>
+    <para>
+     Some characters have special meaning in double-quoted strings and must be escaped by the backslash prefix.
+	 First of all, these are the double quote <code>"</code> as the boundary marker, and the backslash <code>\</code> itself
+	 (if followed by one of the special characters):
+    </para>
+    <programlisting>
+<![CDATA[
+quoted = "She said \"Exactly my point\"." ; Results in a string with quote marks in it.
+hint = "Use \\\" to escape double quote" ; Results in: Use \" to escape double quote
+]]>
+    </programlisting>
+    <para>
+     There is an exception made for Windows-like paths: it's possible to don't escape trailing backslash
+	 if the quoted string is directly followed by a linebreak:
+    </para>
+    <programlisting>
+<![CDATA[
+save_path = "C:\Temp\"
+]]>
+    </programlisting>
+    <para>
+     If one does need to escape double quote followed by linebreak in a multiline value,
+	 it's possible to use value concatenation in the following way
+	 (there is one double-quoted string directly followed by another one):
+    </para>
+    <programlisting>
+<![CDATA[
+long_text = "Lorem \"ipsum\"""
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/parse-ini-file.xml
```diff
@@ -250,17 +250,66 @@ negative_two = ~1
...
 path = ${PATH}
 also_five = ${five}
 
 ]]>
     </programlisting>
    </example>
   </para>
+  <para>
+   <example>
+    <title>escaping characters</title>
```
cmb69, line 262:
```suggestion
    <title>Escaping Characters</title>
```

reference/filesystem/functions/parse-ini-file.xml
```diff
@@ -250,17 +250,66 @@ negative_two = ~1
...
+    <para>
+     Some characters have special meaning in double-quoted strings and must be escaped by the backslash prefix.
+	 First of all, these are the double quote <code>"</code> as the boundary marker, and the backslash <code>\</code> itself
```
cmb69, line 265: Nit: indentation (also several lines further down).

reference/filesystem/functions/parse-ini-file.xml
```diff
@@ -250,17 +250,66 @@ negative_two = ~1
...
+	 (if followed by one of the special characters):
+    </para>
+    <programlisting>
+<![CDATA[
+quoted = "She said \"Exactly my point\"." ; Results in a string with quote marks in it.
+hint = "Use \\\" to escape double quote" ; Results in: Use \" to escape double quote
+]]>
+    </programlisting>
+    <para>
+     There is an exception made for Windows-like paths: it's possible to don't escape trailing backslash
```
cmb69, line 275:
```suggestion
     There is an exception made for Windows-like paths: it's possible to not escape trailing backslash
```

reference/filesystem/functions/parse-ini-file.xml
```diff
@@ -250,17 +250,66 @@ negative_two = ~1
...
+	 (there is one double-quoted string directly followed by another one):
+    </para>
+    <programlisting>
+<![CDATA[
+long_text = "Lorem \"ipsum\"""
+ dolor" ; Results in: Lorem "ipsum"\n dolor
+]]>
+    </programlisting>
+    <para>
+     Another character with special meaning is <code>$</code> (the dollar).
```
cmb69, line 295:
```suggestion
     Another character with special meaning is <code>$</code> (the dollar sign).
```

cmb69 (commented):
Thank you very much! Besides the few nits further down, I assume this documentation is only for PHP 8.1+. If so, that should be clarified, since the manual is supposed to cover both PHP 7 and 8.

reference/filesystem/functions/parse-ini-file.xml
dryabov: Previous example section ("value interpolation") has lowercase title, so let either keep both lowercase, or change pervious title to the "Proper Case" too.

reference/filesystem/functions/parse-ini-file.xml
dryabov: Other suggestions have been merged.

dryabov:
>  I assume this documentation is only for PHP 8.1+

No, it's a general one that works in all PHP versions (since 5.3, if I'm not mistaken). The difference between 8.0 and 8.1 is an edge case of `\"` character sequence that previously was possible to encode both as `\\"` and `\\\"`, but starting from 8.1 the only possible way is to use `\\\"` (similar to PHP's escaping rules).

reference/filesystem/functions/parse-ini-file.xml
cmb69: I suggest to change "value interpolation" to "Value Interpolation", but would leave the other example titles as they are.

dryabov: Done

cmb69:
Ah, fine then. Thank you!

PR #899, opened 2021-09-03
Fix #896: InternalIterator is not documented

### LINKED ISSUE #896: InternalIterator class is reserved as of PHP 8.0.0
We cannot instantiate or extend global scope class `InternalIterator` which is introduced as of PHP 8.0.0 ( https://github.com/php/php-src/commit/ff19ec2df3ef0dca8c29be83eddcde58234f4095 ) and seems to be reserved as internal iterator classname, but this fact is Undocumented.

https://3v4l.org/OKO6a

We should document it in [appendices/migration80/incompatible.xml](https://github.com/php/doc-en/blob/master/appendices/migration80/incompatible.xml)?

cmb69:
This is a skeleton for now. Help to actually document the class is welcome!

### DIFF
--- a/appendices/migration80/new-features.xml
+++ b/appendices/migration80/new-features.xml
@@ -232,6 +232,11 @@ class ChildClass extends ParentClass {
       <code>(int) $resource</code>. It provides the same functionality under a clearer API.
      </para>
     </listitem>
+    <listitem>
+     <para>
+      The <classname>InternalInterface</classname> has been added.
+     </para>
+    </listitem>
    </itemizedlist>
   </sect3>
  </sect2>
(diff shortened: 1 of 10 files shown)


thinkverse:
For anyone wanting a bit of information about `InternalIterator`, it was introduced in PR (https://github.com/php/php-src/pull/5216).

From my limited understanding, it's the generic iterator class some PHP classes use internally to represent Traversable objects? For instance, running a WeakMaps `getIterator` method against an `instanceof \InternalIterator` returns true, as shown here ([3v4l.org/HRpmk](https://3v4l.org/HRpmk#v8.0.0)). Though I might be off so please, feel free to correct me if I am.

cmb69:
@thinkverse, I think you're pretty much right on.

Girgias:
Should we just merge this as is, and enhance later?

cmb69:
I've fleshed that out a bit.

PR #900, opened 2021-09-03
Document the null behaviour for http_build_query function

mauriciofauth:


mauriciofauth:
Closed by 7a7ad98fe2604fcb8c72d6b14ad054a093650847.

cmb69:
Thanks! I wouldn't have expected this behavior (I had rather thought that a NULL was converted to an empty string). [Resources are also skipped](https://github.com/php/php-src/blob/php-7.4.23/ext/standard/http.c#L159-L161), but I this might not need to be documented.

PR #902, opened 2021-09-04
Update curl-version.xml

### LINKED ISSUE #901: curl_version function parameter is not updated.
https://www.php.net/manual/en/function.curl-version.php

`age` parameter of `curl_version()` function is removed as of PHP 8.0.0, but its documentation is not updated.

sy-records:
Fix #901


cmb69:
Thank you! I've added the respective changelog entries right away.

PR #908, opened 2021-09-05
Generate ext/intl methodsynopses based on stubs (NumberFormatter, ResourceBundle)

kocsismate:


cmb69:
Thank you!

BTW: "Parse a string into a double and a currency using the current formatter." does not appear to be the most useful behavior.

kocsismate:
> Thank you!

Thanks for your fixes as well! I've just seen that e.g. I missed the nullability change of `pattern`.

PR #911, opened 2021-09-05
Generate ext/intl methodsynopses based on stubs (IntlTimeZone)

kocsismate:
This one is pretty straightforard, since mostly only parameter names have changed, and return types are not documented.

### DIFF
--- a/reference/intl/intltimezone/getidforwindowsid.xml
+++ b/reference/intl/intltimezone/getidforwindowsid.xml
@@ -77,7 +77,7 @@
      <row>
       <entry>8.0.0</entry>
       <entry>
-       <parameter>region</parameter> is nullable now.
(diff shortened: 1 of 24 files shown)


reference/intl/intltimezone/getidforwindowsid.xml
```diff
@@ -64,6 +63,28 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       <parameter>region</parameter> is nullable now.
```
Girgias, line 80:
```suggestion
       <parameter>region</parameter> is now nullable.
```

Girgias (approved):

PR #913, opened 2021-09-05
Remove some unnecessary methodsynopsis role attributes

kocsismate:
These attributes don't seem necessary for me, and only makes noise when method synopses are generated from stubs.

### DIFF
--- a/reference/curl/curlfile/construct.xml
+++ b/reference/curl/curlfile/construct.xml
@@ -17,7 +17,7 @@
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>posted_filename</parameter><initializer>&null;</initializer></methodparam>
   </constructorsynopsis>
   <para>&style.procedural;</para>
-  <methodsynopsis role="procedural">
+  <methodsynopsis>
    <type>CURLFile</type><methodname>curl_file_create</methodname>
    <methodparam><type>string</type><parameter>filename</parameter></methodparam>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>mime_type</parameter><initializer>&null;</initializer></methodparam>
--- a/reference/com/com/construct.xml
+++ b/reference/com/com/construct.xml
@@ -8,7 +8,7 @@
 
  <refsect1 role="description">
   &reftitle.description;
-  <constructorsynopsis role="oop">
+  <constructorsynopsis>
    <modifier>public</modifier> <methodname>com::__construct</methodname>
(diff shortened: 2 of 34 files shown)


kocsismate:
These changes should be very straightforward, so I'd like to ask for a review.

Girgias:
I don't mind the change, but I'm wondering what the harm is in them? Is it just because of the stub script generation changes?

Also I've realised recently that *pedantically* speaking for the prototype of functions, we shouldn't have been using ``<methodsynopsis>`` but rather ``<funcsynopsis>``, however I don't even know if PhD handles this case correctly, and if it makes any sense to fix this at this point in time.

kamil-tekiela:
I tried it once but I got an error. So it's not allowed

cmb69:
[`<funcsynopsis>`](https://tdg.docbook.org/tdg/4.5/funcsynopsis.html) should be supported by our DocBook versions, but it requires rather different markup than `<methodsynopsis>`, so probably we should not attempt to switch for now (esp. since it is unlikely to be properly supported by PhD).

Regarding this change, I'm in the same boat as @Girgias.

kocsismate:
> I don't mind the change, but I'm wondering what the harm is in them? Is it just because of the stub script generation changes?

Yes, exactly that's the case. These `role` attributes are not necessary, therefore the generated method synopses don't contain them, resulting in unnecessary output changes.

If we can minimize the diff between the output of the script and the actual method synopses, then we'll be able to quasi-automatically synchronize signatures with the docs.

cmb69 (approved):
Okay, than this is fine for me. Thank you!

salathe:
This change has undesired side-effects.

**Before**

curlfile-before

**After**

curlfile-after

reference/curl/curlfile/construct.xml
```diff
@@ -17,7 +17,7 @@
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>posted_filename</parameter><initializer>&null;</initializer></methodparam>
   </constructorsynopsis>
   <para>&style.procedural;</para>
-  <methodsynopsis role="procedural">
+  <methodsynopsis>
```
cmb69, line 20: Ah, this shouldn't have been changed.

kocsismate: I think we should add support for `funcsynopsis`. Currently, the usage of the `role` attribute is not very clear (it's usually used for aliases, but not always - like in this case), so that gen_stub.php doesn't always generate it when it's needed.

PR #914, opened 2021-09-05
Remove some unnecessary abstract modifiers from methodsynopses

kocsismate:
These are unnecessary since the `interface` keyword is displayed on the class synopsis pages.


kocsismate:
These changes should be very straightforward, so I'd like to ask for a review.

PR #918, opened 2021-09-06
Fix #81365: Flipping opcache.file_cache_only is broken

cmb69:
@nikic, is that a correct interpretation of https://bugs.php.net/bug.php?id=81365#1629279195?


nikic (approved):

cmb69:
I'm going to merge despite the fact that this mentions PHP 8.1.0 (which is not yet released).

PR #919, opened 2021-09-06
DomElement::setAttribute does not always return a new DomAttr instance

DennisBirkholz:
See https://3v4l.org/8YrUO


cmb69:
Thank you!

PR #920, opened 2021-09-06
 bugs in md5

Alikareem1996:
I found Bugs in MD5
If we make a comparison with the number 0 and text for example QLTHNDT and we encode the text to MD5 and we will see the comparison, there will be a valid comparison even though the number 0 is not equal to text encryption
0e405967825401955372549139051580 As you can see this is text encoding
When comparing with the 0, the comparison will appear correct, and the reason is that it only takes the first number 0 and compares with it and does not compare with the entire hash, so this error leads to a loophole in the system when it is configured
for example
<?php

  $val1=0;
  $val2="QLTHNDT";
if($val1==md5($val2)){

     echo "good";
}
else
{
echo ":(";
}

?>

### DIFF
--- a/README
+++ b/README
@@ -7,3 +7,23 @@ directory structure is probably like:
 
   phpdoc/en            (DocBook files for the English language)
   phpdoc/doc-base      (Tools and resources used for all languages)
+
+I found Bugs in MD5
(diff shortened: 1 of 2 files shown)


kamil-tekiela:
There is no bug in `md5`. You just used loose comparison which caused type juggling. 
See https://3v4l.org/io4MX
Use strict checking or `hash_equals` and your code will work fine.

Alikareem1996:
You can See https://3v4l.org/MiW09

Issue #916, opened 2021-09-05
Typo; incorrect symbol displayed


SirJimmothy:
From manual page: https://php.net/function.http-build-query

---

`$arg_separator` default is incorrect:

```php
 http_build_query(
    mixed $data,
    string $numeric_prefix = ?,
    string $arg_separator = ?,
    int $encoding_type = PHP_QUERY_RFC1738
): string
```

Girgias:
Should be:
```php
 http_build_query(
    mixed $data,
    string $numeric_prefix = "",
    ?string $arg_separator = null,
    int $encoding_type = PHP_QUERY_RFC1738
): string
```

Based on the stubs

PR #922, opened 2021-09-07
Add missing parameter defaults for `http_build_query`

thinkverse:
This PR adds the missing parameter initializers for the `http_build_query` function And updates the signature to match the one in [/basic_functions.stub.php](https://github.com/php/php-src/blob/f983373e44dd45c7e09036fc67476fc946306b3c/ext/standard/basic_functions.stub.php#L1298). As discussed in #916


kamil-tekiela (approved):

kamil-tekiela:
Thank you.

PR #923, opened 2021-09-07
DOMAttr: setting value triggers XML entity parsing

DennisBirkholz:
See https://3v4l.org/Lhu2c

### DIFF
--- a/reference/dom/domattr.xml
+++ b/reference/dom/domattr.xml
@@ -123,6 +123,14 @@
      <term><varname>value</varname></term>
      <listitem>
       <para>The value of the attribute.</para>
+      <note>
+       <para>
+        Note, XML entities are expanded upon setting a value.
+        Thus the <constant>&amp;</constant> character has a special meaning.
+        Setting <varname>value</varname> to itself will fail when <varname>value</varname> contains an <constant>&amp;</constant>.
+        To avoid entity expansion, use <methodname>DOMElement::setAttribute</methodname> instead.
+       </para>
+      </note>
      </listitem>
     </varlistentry>
    </variablelist>


reference/dom/domattr.xml
```diff
@@ -122,6 +122,13 @@
      <term><varname>value</varname></term>
      <listitem>
       <para>The value of the attribute</para>
+      
+      <para>
+       Note, XML entities are decoded upon setting value.
```
cmb69, line 127: I think they are encoded (not decoded), but the proper term seems to be expanded.

reference/dom/domattr.xml
```diff
@@ -122,6 +122,13 @@
      <term><varname>value</varname></term>
      <listitem>
       <para>The value of the attribute</para>
+      
+      <para>
```
cmb69, line 126: I suggest to turn that into a note (i.e. wrap the new `<para>` in a `<note>` element). Or maybe it should even be a `<caution>`.

reference/dom/domattr.xml
```diff
@@ -122,6 +122,13 @@
...
+       Note, XML entities are decoded upon setting value.
+       Thus the <constant>&amp;</constant> character has a special meaning.
+       Setting <varname>value</varname> to itself will fail when <varname>value</varname> contains an <constant>&amp;</constant>.
+       Use <function>DOMElement::setAttribute</function> to set unescaped strings instead.
```
cmb69, line 130:
```suggestion
       To avoid entity expansion, use <methodname>DOMElement::setAttribute</methodname> instead.
```

cmb69 (commented):
Thank you for the PR (which apparently felt through the cracks)! It is actually missing documentation regarding https://github.com/php/php-src/pull/10132, and given that the current behavior doesn't match the DOM specification, I think we should make it more prominent. And maybe we should add an example.

Girgias:
This needs a rebase.

Also @nielsdos does this description make sense to you?

nielsdos:
If the changes requested by @cmb69 are applied, and this PR is rebased, then it's good.

DennisBirkholz:
I rebased the patch and added the changes proposed by @cmb69.

Girgias:
Thank you!

PR #930, opened 2021-09-08
Fix `password_needs_rehash` method signature

thinkverse:
Noticed that the current method signature for `password_needs_rehash` in the [documentation](https://www.php.net/manual/en/function.password-needs-rehash.php) was wrong.

`$algo` should be of `string|int|null` like `password_hash` and not `mixed`. It was also missing the default initializer for `$options`.

This PR makes sure the signature matches up with the internal method signature found in [standard/basic_functions.stub.php](https://github.com/php/php-src/blob/c19e4b9997cd429c86c20e9210ff3d9b2e94171d/ext/standard/basic_functions.stub.php#L1511).


kamil-tekiela (approved):

kamil-tekiela:
Thanks

PR #931, opened 2021-09-08
Clean up filesystem function signatures

thinkverse:
This PR intends to clean up the function signatures for the filesystem functions. Some are missing default initializers, mostly the nullable parameters, and some have missing return types. `copy` though for instance had the wrong parameter names, in the [documentation](https://www.php.net/manual/en/function.copy.php), they are `$source` and `$dest` respectively but the actual names - found in the [php-src](https://github.com/php/php-src/blob/bf92bddb878dbd86d23db2900bad1cdfcd449eb7/ext/standard/basic_functions.stub.php#L853), are `$from` and `$to`.

This PR will be squashed down to one commit before it is ready to merge.


thinkverse:
I'm pretty sure I got all mismatched signatures, if I missed any let me know. 👍

Girgias:
I'm surprised this isn't caught by #903, @kocsismate any idea why?

kocsismate:
> I'm surprised this isn't caught by #903, kocsismate any idea why?

Yes, I only added a subset of all the changes, and I left most of the functions in ext/standard untouched until the majority of my outstanding PRs are merged.

Girgias:
Okay that explains that then. :-)

reference/filesystem/functions/fgets.xml
```diff
@@ -10,8 +10,8 @@
   &reftitle.description;
   <methodsynopsis>
    <type class="union"><type>string</type><type>false</type></type><methodname>fgets</methodname>
-   <methodparam><type>resource</type><parameter>handle</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>length</parameter></methodparam>
+   <methodparam><type>resource</type><parameter>stream</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>length</parameter><initializer>&null;</initializer></methodparam>
```
Girgias, line 14: The type declaration needs to be amended to be a union of int and null

reference/filesystem/functions/copy.xml
```diff
@@ -10,13 +10,13 @@
   &reftitle.description;
   <methodsynopsis>
    <type>bool</type><methodname>copy</methodname>
-   <methodparam><type>string</type><parameter>source</parameter></methodparam>
-   <methodparam><type>string</type><parameter>dest</parameter></methodparam>
-   <methodparam choice="opt"><type>resource</type><parameter>context</parameter></methodparam>
+   <methodparam><type>string</type><parameter>from</parameter></methodparam>
+   <methodparam><type>string</type><parameter>to</parameter></methodparam>
+   <methodparam choice="opt"><type>resource</type><parameter>context</parameter><initializer>&null;</initializer></methodparam>
```
Girgias, line 15: The type declaration needs to be amended to be a union of resource and null

reference/filesystem/functions/fgets.xml
```diff
@@ -63,15 +63,15 @@
...
+if ($fp) {
+    while (($buffer = fgets($fp, 4096)) !== false) {
         echo $buffer;
     }
-    if (!feof($handle)) {
+    if (!feof($fp)) {
         echo "Error: unexpected fgets() fail\n";
     }
-    fclose($handle);
+    fclose($fp);
```
Girgias, line 74: I'm not sure ``$fp`` is better than ``$handle``

reference/filesystem/functions/file-get-contents.xml
```diff
@@ -12,9 +12,9 @@
    <type class="union"><type>string</type><type>false</type></type><methodname>file_get_contents</methodname>
    <methodparam><type>string</type><parameter>filename</parameter></methodparam>
    <methodparam choice="opt"><type>bool</type><parameter>use_include_path</parameter><initializer>&false;</initializer></methodparam>
-   <methodparam choice="opt"><type>resource</type><parameter>context</parameter></methodparam>
+   <methodparam choice="opt"><type>resource</type><parameter>context</parameter><initializer>&null;</initializer></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>offset</parameter><initializer>0</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>length</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>length</parameter><initializer>&null;</initializer></methodparam>
```
Girgias, line 17: Same here, union types are needed.

reference/filesystem/functions/touch.xml
```diff
@@ -39,11 +39,11 @@
...
     </varlistentry>
     <varlistentry>
-     <term><parameter>time</parameter></term>
+     <term><parameter>mtime</parameter></term>
      <listitem>
       <para>
-       The touch time. If <parameter>time</parameter> is not supplied, 
-       the current system time is used.
+       The touch time. If <parameter>mtime</parameter> is not supplied, 
+       the current system <function>time</function> is used.
```
Girgias, line 46: I don't see the point in referencing the time *function* here?

reference/filesystem/functions/fwrite.xml
```diff
@@ -79,20 +79,20 @@ if (is_writable($filename)) {
...
-    if (fwrite($handle, $somecontent) === FALSE) {
+    if (fwrite($fp, $somecontent) === FALSE) {
         echo "Cannot write to file ($filename)";
         exit;
     }
 
     echo "Success, wrote ($somecontent) to file ($filename)";
 
-    fclose($handle);
+    fclose($fp);
```
Girgias, line 95: Again I don't think fp is a better name here.

reference/filesystem/functions/fwrite.xml
```diff
@@ -113,9 +113,9 @@ if (is_writable($filename)) {
     <programlisting role="php">
      <![CDATA[
 <?php
-function fwrite_stream($fp, $string) {
-    for ($written = 0; $written < strlen($string); $written += $fwrite) {
-        $fwrite = fwrite($fp, substr($string, $written));
+function fwrite_stream($fp, $data) {
+    for ($written = 0; $written < strlen($data); $written += $fwrite) {
+        $fwrite = fwrite($fp, substr($data, $written));
```
Girgias, line 118: In the Para just above we are talking about strings, so I'm not sure changing to ``$data`` is better.

Girgias (changes requested):
I haven't marked all the union type issues as it is a recurrent issue.

The var name changes in the examples are dubious IMHO.

reference/filesystem/functions/fgets.xml
thinkverse: `$fp` and `$fh` are what most other functions use for their examples, hence the change from `handle` to  `fp` to keep it more consistent.

reference/filesystem/functions/fwrite.xml
thinkverse: Because the parameter was named `string` and not `data`, I figured the original author named the variable `$string` because the parameter was named that.

After changing the parameter to its correct name, I figured changing the example variable to match would make it easier for readers to understand the connection.

Can change it back to `$string` if desired.

reference/filesystem/functions/fwrite.xml
thinkverse: Same reason as before, changed it to keep it consistent with other similar examples.

reference/filesystem/functions/touch.xml
thinkverse: Previously `time` now `mtime` had its default set to `time()` - I'm guessing to indicate its default behavior when passed null, referencing the time _function_ here keeps that early reference.

reference/filesystem/functions/fwrite.xml
cmb69: I think it's better not to change variable names in the examples for now (unless that would be actual parameter names).

reference/filesystem/functions/fgets.xml
cmb69: Generally, the supposedly proper function signatures can be found in https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php.

reference/filesystem/functions/file-get-contents.xml
thinkverse: Will update all union types. 👍

reference/filesystem/functions/fgets.xml
thinkverse: Some have a strict union of `int|null` and some are nullable, meaning `?int`, do we have a `<type>` for that, or is `int|null` the correct choice to indicate nullable? 🤔

reference/filesystem/functions/touch.xml
cmb69: That makes some sense; however, `is not supplied` should better be changed to `is &null`, and there should be a changelog entry about these parameters being nullable as of PHP 8.0.0.

reference/filesystem/functions/fgets.xml
cmb69: `<type class="union"><type>int</type><type>null</type></type>` will be displayed as `?int` anyway.

thinkverse: Ooh, gotcha. 👍

reference/filesystem/functions/fwrite.xml
thinkverse: Ok, I'm changing it back to `$string`.

reference/filesystem/functions/touch.xml
thinkverse: `is &null;` and changelog added.

reference/filesystem/functions/touch.xml
```diff
@@ -53,7 +53,7 @@
       <para>
        If present, the access time of the given filename is set to 
        the value of <parameter>atime</parameter>. Otherwise, it is set to
-       the value passed to the <parameter>time</parameter> parameter.
+       the value passed to the <parameter>mtime</parameter> parameter.
        If neither are present, the current system time is used.
```
Girgias, line 57: This needs similar adjustment in wording to remove "is present".

Girgias (changes requested):
Still the minor complaint about using ``$fp`` but that's fine.

reference/filesystem/functions/touch.xml
```diff
@@ -69,6 +69,34 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       <parameter>mtime</parameter> is nullable now.
```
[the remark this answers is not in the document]
cmb69: I've written it several times like this, though. :)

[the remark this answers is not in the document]
Girgias: So you are to blame for the inconsistency. :-p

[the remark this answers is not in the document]
cmb69: Well, I might have been the first to write this, but I didn't "correct" later on, because I wasn't sure about the wording. I guess I would have introduced an entity early on, if these were parameterizable.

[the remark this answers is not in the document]
thinkverse: I grabbed the wording from the [mbstring.changelog.encoding-nullable](https://github.com/php/doc-en/blob/3cdd39bb9505e6735d802da83a04870cfa8f2311/language-snippets.ent#L1055) entity, I have https://github.com/php/doc-en/pull/630 open which adds another changelog entity. I could commit an update to the wording of the `encoding-nullable` entity in that PR since it's related to mb string changelogs already. Might help reduce the inconsistency.

reference/filesystem/functions/touch.xml
thinkverse: So changing it from using `if present` to `if &null;` instead? Would this replacement work OK?

```
If &null;, the access time of the given filename is set to 
the value of <parameter>atime</parameter>. Otherwise, it is set to
the value passed to the <parameter>mtime</parameter> parameter.
If both are &null;, the current system time is used.
```

reference/filesystem/functions/fwrite.xml
```diff
@@ -9,15 +9,15 @@
...
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>fwrite</methodname>
-   <methodparam><type>resource</type><parameter>handle</parameter></methodparam>
-   <methodparam><type>string</type><parameter>string</parameter></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>length</parameter></methodparam>
+   <type class="union"><type>int</type><type>false</type></type><methodname>fwrite</methodname>
+   <methodparam><type>resource</type><parameter>stream</parameter></methodparam>
+   <methodparam><type>string</type><parameter>data</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>length</parameter><initializer>&null;</initializer></methodparam>
```
kocsismate, line 15: needs a change log entry + mention `null` instead of "argument is given"

reference/filesystem/functions/file-get-contents.xml
```diff
@@ -12,9 +12,9 @@
    <type class="union"><type>string</type><type>false</type></type><methodname>file_get_contents</methodname>
    <methodparam><type>string</type><parameter>filename</parameter></methodparam>
    <methodparam choice="opt"><type>bool</type><parameter>use_include_path</parameter><initializer>&false;</initializer></methodparam>
-   <methodparam choice="opt"><type>resource</type><parameter>context</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>resource</type><type>null</type></type><parameter>context</parameter><initializer>&null;</initializer></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>offset</parameter><initializer>0</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>length</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>length</parameter><initializer>&null;</initializer></methodparam>
```
kocsismate, line 17: needs a change log entry

kocsismate (commented):
A few nits. Please fix these, and then finally we can merge this PR :)

reference/filesystem/functions/unlink.xml
```diff
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>bool</type><methodname>unlink</methodname>
    <methodparam><type>string</type><parameter>filename</parameter></methodparam>
-   <methodparam choice="opt"><type>resource</type><parameter>context</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>resource</type><type>null</type></type><parameter>context</parameter><initializer>&null;</initializer></methodparam>
```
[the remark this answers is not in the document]
thinkverse: The `&note.context-support;` entity is used on `dir`, `file`, `fopen`, `mkdir`, `readfile`, `rename`, `rmdir` and `unlink`, looking at the src for 8.0 it seems all except `null` so should I update the entity and add the paragraph to `language-snippets.ent` instead?

```
<!ENTITY note.context-support '<para xmlns="http://docbook.org/ns/docbook">A <link linkend="stream.contexts">context stream</link>
<type>resource</type>. If &null; a default context will be used.</para>'>
```

Not sure how to word that exactly.

reference/filesystem/functions/touch.xml
```diff
@@ -69,6 +69,34 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       <parameter>mtime</parameter> is now nullable.
```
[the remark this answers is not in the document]
thinkverse: Would the following work OK instead?

```suggestion
       <parameter>mtime</parameter> and <parameter>atime</parameter>
       are now nullable.
```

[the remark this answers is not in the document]
kocsismate: Yes! That's the wording what I also use :)

reference/filesystem/functions/unlink.xml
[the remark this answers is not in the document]
kocsismate: Yes, I think the entity should be updated in this case. I'm ok with the suggested wording, but @cmb69 may have a better idea.

reference/filesystem/functions/file-get-contents.xml
thinkverse: Changelog entry added.

reference/filesystem/functions/fwrite.xml
thinkverse: Changelog entry and parameter description clarified.

reference/filesystem/functions/touch.xml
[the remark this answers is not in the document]
thinkverse: Changelog entries combined.

reference/filesystem/functions/umask.xml
```diff
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis>
    <type>int</type><methodname>umask</methodname>
-   <methodparam choice="opt"><type>int</type><parameter>mask</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>mask</parameter><initializer>&null;</initializer></methodparam>
```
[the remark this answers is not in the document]
thinkverse: Changelog entry and parameter description updated.

reference/filesystem/functions/unlink.xml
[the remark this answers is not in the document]
thinkverse: Then I think it might be better for a more senior doc member to update the entry. 👍

reference/filesystem/functions/file-get-contents.xml
```diff
@@ -135,6 +135,12 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <parameter>length</parameter> is nullable now.
```
kocsismate, line 141: The docs uses the `is nullable now` wording elsewhere

reference/filesystem/functions/fwrite.xml
```diff
@@ -43,10 +43,9 @@
      <term><parameter>length</parameter></term>
      <listitem>
       <para>
-       If the <parameter>length</parameter> argument is given, writing will
-       stop after <parameter>length</parameter> bytes have been written or
-       the end of <parameter>string</parameter> is reached, whichever comes
-       first.
+       If <parameter>length</parameter> is not &null;, writing will stop
```
kocsismate, line 46: When suggesting to use null, I didn't take the current wording into account. So with that in mind,
```
If <parameter>length</parameter> is an &integer;, writing will stop
```

would sound much better.

thinkverse:
```suggestion
       If <parameter>length</parameter> is an &integer;, writing will stop
```

reference/filesystem/functions/file-get-contents.xml
thinkverse: > The docs uses the `is nullable now` wording elsewhere

Yes, and we don't want to keep that wording consistent between functions? 🤔

kocsismate: Ah... I wanted to write `is now nullable`, sorry!

kocsismate: Although, I've just checked and they are almost used equally. OK, then I don't mind which one is used now :)

kocsismate (approved):
I'm going to merge this tomorrow, unless someone else spots any new issues.

reference/filesystem/functions/file-get-contents.xml
thinkverse: Yes, they're used interchangeably around the docs. I used `is nullable now` because that's what is used in the MB entity `mbstring.changelog.encoding-nullable` (https://github.com/php/doc-en/commit/ab493e941a14dcc760adbbe08976650b3a105be6) and that's the wording I found first. LDAP also uses `is nullable now`.

Issue #928, opened 2021-09-07
error_reporting default value description is wrong


mumumu:
https://www.php.net/manual/en/ini.list.php

As of PHP 8.0.0, the default `error_reporting` level is now E_ALL, but in `error_reporting` description, its value is `NULL`.

In php-src, `ZEND_INI_ENTRY` sets NULL as default value of it [*1], it seems to be correct.
On the other hand, `ZEND_INI_MH` returns `E_ALL` [*2].

Is this a bug of PHP Manual? 

[*1] https://github.com/php/php-src/blob/php-8.0.0/Zend/zend.c#L183
[*2] https://github.com/php/php-src/blob/php-8.0.0/Zend/zend.c#L98-L106

Girgias:
This was fixed with b9d5dcd7821e4337cd15e83a82ba6d07a39360dc

PR #932, opened 2021-09-08
Updated error_reporting default value description.

mumumu:
As of PHP 8.0.0, `error_reporting` ini derective default value was changed.

closes: #928

### DIFF
--- a/appendices/ini.list.xml
+++ b/appendices/ini.list.xml
@@ -268,9 +268,15 @@
       </row>
       <row>
        <entry><link linkend="ini.error-reporting">error_reporting</link></entry>
-       <entry>NULL</entry>
+       <entry>E_ALL</entry>
        <entry>PHP_INI_ALL</entry>
-       <entry></entry>
+       <entry>
+        Prior to PHP 8.0.0, the default value was
+        <constant>E_ALL</constant> &amp;
+        ~<constant>E_NOTICE</constant> &amp;
+        ~<constant>E_STRICT</constant> &amp;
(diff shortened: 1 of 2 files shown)


cmb69 (commented):
I'm not quite sure about this. Technically, the default is unchanged; without any php.ini: `php -i | grep error_reporting` prints `error_reporting => no value => no value` and `php -r 'var_dump(ini_get("error_reporting"));'` prints `string(0) ""`. However, in practice, `E_DEPRECATED` and `E_NOTICE` are included as of PHP 8.0.0. Since the [migration guide states it this way](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.other), it is probably a good idea to follow suit. I'm not sure about `E_STRICT`, though.

Girgias:
The default value seems to be ``"-1"``?

Also E_STRICT is part of E_ALL since 5.4 IIRC

cmb69:
> The default value seems to be "-1"?

`-1` is `E_ALL`; you likely have a php.ini somewhere. Or is that different on Windows?

Girgias:
> > The default value seems to be "-1"?
> 
> `-1` is `E_ALL`; you likely have a php.ini somewhere. Or is that different on Windows?

I was trying on 3v4l.org (https://3v4l.org/IK12A) to see what the value is from the INI when called from PHP, but that may be also a php.ini

mumumu:
#1307 is now correct update.

Girgias:
Oh, I forgot about this :S

PR #933, opened 2021-09-08
Add note about NOP code

kamil-tekiela:
Based on https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.other


cmb69:
Thank you!

PR #935, opened 2021-09-09
Generate Zend classsynopses based on stubs

kocsismate:


language/predefined/error.xml
```diff
@@ -56,16 +57,33 @@
...
+     <modifier>private</modifier>
+     <type class="union"><type>Throwable</type><type>null</type></type>
+     <varname linkend="error.props.previous">previous</varname>
+     <initializer>null</initializer>
+    </fieldsynopsis>
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.error')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[1])" />
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.error')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[1])" />
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.error')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[not(@role='procedural')])">
```
kocsismate, line 81: Question: it would be more future-proof to always generate `[not(@role='procedural')]` at the end, right? For now, it's only used for constructorsynopses, but regular methods use `[1]`.

language/predefined/error.xml
```diff
@@ -99,6 +117,24 @@
         <para>The line where the error happened</para>
        </listitem>
       </varlistentry>
+      <varlistentry xml:id="error.props.previous">
+       <term><varname>previous</varname></term>
+       <listitem>
+        <para>The previously thrown error</para>
```
kocsismate, line 123: I guess these descriptions could be improved

language/predefined/errorexception.xml
```diff
@@ -25,37 +20,46 @@
...
     <fieldsynopsis>
      <modifier>protected</modifier>
      <type>int</type>
      <varname linkend="errorexception.props.severity">severity</varname>
+     <initializer>E_ERROR</initializer>
     </fieldsynopsis>
- 
+
     <classsynopsisinfo role="comment">&InheritedProperties;</classsynopsisinfo>
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('exception.synopsis')/descendant::db:fieldsynopsis)" />
```
kocsismate, line 51: I noticed that a few classes - mainly exceptions - use a different scheme for accessing properties than the rest: they are referenced via`exception.synopsis` (rather than `class.NAME`) and a shorter pointer.

As far as I noticed, we can use both pointers, at least both worked for me. Am I right?

language/predefined/stringable.xml
```diff
@@ -31,26 +30,21 @@
...
      </oointerface>
     </classsynopsisinfo>
-<!-- }}} -->
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
-<!-- Bring back if we add the toString method as its own page later. For now,
-     just link to the existing description in magic.xml.
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[not(@role='procedural')])">
-     <xi:fallback />
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[1])">
```
kocsismate, line 44: What does it cause if it's included? I couldn't notice any change.

kocsismate (commented):
Apart from a few questions, the changes are very straightforward.

language/predefined/error.xml
cmb69: Yes, should always use `[not(@role='procedural')]`; see also https://github.com/php/doc-base/blob/master/RFC/skeletons/classname.xml#L65-L73.

language/predefined/stringable.xml
cmb69: The documentation for `Stringable::__toString()` is completely missing.

language/predefined/error.xml
kocsismate: Thanks for the link!

language/predefined/stringable.xml
kocsismate: Yes, but apparently it did no harm for me when I actually tried to include it.

cmb69: Indeed, it does no harm because there is a `<xi:fallback/>` element, but of course, if there are no methods documented, there is no difference whether the `<xi:include>` is there or not. Anyhow, the `<xi:include>` should be there, just in case that method ever will be documented. :)

language/predefined/errorexception.xml
Girgias: @salathe do you have an answer to this?

salathe: I'm not aware of any particular *reason* why it is (still) done that way today.  It is just [what docgen spits out](https://github.com/php/doc-base/blob/56a99a2768b59efdf4940cd16a84cceb98dff75f/scripts/docgen/docgen.php#L581-L584) and is basically the same since [inherited properties was added to docgen](https://github.com/php/doc-base/commit/c34b5eb07623b10da98c180be586a39e15bae5b4).

kocsismate:
I'm closing this PR and will create a few smaller PRs which are easier to review.

PR #936, opened 2021-09-09
Add typed properties reference to TypeError page

bdsl:


language/predefined/typeerror.xml
```diff
@@ -15,9 +15,13 @@
   <section xml:id="typeerror.intro">
    &reftitle.intro;
    <para>
-    There are three scenarios where a
+    There are four scenarios where a
     <ooclass><classname>TypeError</classname></ooclass> may be thrown:
```
cmb69, line 19:
```suggestion
    A <ooclass><classname>TypeError</classname></ooclass> may be thrown in the following cases:
```
This gets rid of the number (4), which at some point is likely to get out of date.

language/predefined/typeerror.xml
```diff
@@ -15,9 +15,13 @@
...
     <simplelist>
+    <member>
```
Girgias, line 21: Nit, missing indentation

language/predefined/typeerror.xml
bdsl: I thought this list was short enough and close enough to the number that it probably wouldn't go out of date with regard to the list. This isn't a long page, so I'd expect anyone editing it to read the whole thing. But we can do without it.

Maybe 
```suggestion
    A <ooclass><classname>TypeError</classname></ooclass> is thrown when:
```

language/predefined/typeerror.xml
bdsl: Thanks, fixed.

language/predefined/typeerror.xml
cmb69: That's fine for me. Thanks!

language/predefined/typeerror.xml
```diff
@@ -15,9 +15,13 @@
   <section xml:id="typeerror.intro">
    &reftitle.intro;
    <para>
-    There are three scenarios where a
+    There are four scenarios where a
     <ooclass><classname>TypeError</classname></ooclass> may be thrown:
     <simplelist>
+     <member>
+      The value being set for a class property does not match
+      its corresponding declared type.
```
salathe, line 23:
```suggestion
      The type of the value being set for a class property does not match
      the property's declared type.
```

language/predefined/typeerror.xml
salathe: The `<ooclass>...</ooclass>` tags are not desired here. Suggest removing them, while you're changing this line anyway.  (Aside: this occurs throughout the "Predefined Exceptions" part of the manual.)

```suggestion
    A <classname>TypeError</classname> may be thrown when:
```

bdsl: I don't really know what those tags are - I'd rather leave someone who knows about them to make that change.

language/predefined/typeerror.xml
bdsl: I'm not sure it makes sense to talk about the 'type of the value'. I don't think the value really has a type. I suppose the existing entries are a little inconsistent between the "argument type" and the "value being returned". I was following the example of the 'value being returned'.

bdsl:
```suggestion
      The value being set for a class property does not match
      the property's corresponding declared type.
```

language/predefined/typeerror.xml
cmb69: @salathe [certainly knows that](https://github.com/php/doc-en/commits?author=salathe). :)

salathe: > I'd rather leave someone who knows about them to make that change.

Someone who knows (me) requested the change.  I hope that's okay. 😊

salathe: Aside: circling back to *"This gets rid of the number (4), which at some point is likely to get out of date."*  This is a great suggestion, especially considering that the last case in the list, of an incorrect number of arguments, does not raise a TypeError as of PHP 7.1.0 (ArgumentCountError instead). That particular case can be relegated to a changelog entry on this page.  (Not a suggestion for this PR, but if you want to….)

bdsl: @cmb69 @salathe OK - I've accepted the change suggestion. I didn't really feel like I was well placed to review it as a suggestion.

salathe:
Thanks @bdsl 🥇

PR #937, opened 2021-09-09
[81385] Add `ReflectionProperty::isPromoted` documentation

thinkverse:
This PR adds documentation for the `isPromoted` method on the `ReflectionProperty` class. It was generated using the gendoc script from [doc-base](https://github.com/php/doc-base/) and wording was based on [isPrivate](https://www.php.net/manual/en/reflectionproperty.isprivate.php) and example forked from [isDefault](https://www.php.net/manual/en/reflectionproperty.isdefault.php) and re worked to match `isPromoted`.

Feel free to suggest changes and or additions. 👍


cmb69 (commented):
Thank you! This looks generally good to me, but I think there should be a link to the [Constructor Promotion section](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion) (`<link linkend="language.oop5.decon.constructor.promotion"/>`). Either in the see-also section, or maybe link the word "promoted" in the description.

PR #938, opened 2021-09-10
mixed is a reserved word as of PHP 8.0.0.

mumumu:
mixed is now reserved word as of PHP 8.0.0, so updated migration guide and reserved word list.

* trait
  - https://3v4l.org/kZ0Mc
* class
  - https://3v4l.org/5NCWs
* interface
  - https://3v4l.org/FsR5g

### DIFF
--- a/appendices/migration80/incompatible.xml
+++ b/appendices/migration80/incompatible.xml
@@ -73,6 +73,11 @@
     </listitem>
     <listitem>
      <para>
+      <literal>mixed</literal> is now a reserved word, so it cannot be used to name a class, interface or trait, and is also prohibited from being used in namespaces.
+     </para>
+    </listitem>
+    <listitem>
+     <para>
       Assertion failures now throw by default. If the old behavior is desired,
       <code>assert.exception=0</code> can be set in the INI settings.
      </para>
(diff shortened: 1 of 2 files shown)


appendices/migration80/incompatible.xml
```diff
@@ -71,6 +71,11 @@
       <literal>match</literal> is now a reserved keyword.
```
mumumu, line 71: reserved `keyword` and reserved `word` is a bit confusing. should we update this line about `match` keyword?

kamil-tekiela: They are different. One is just a reserved word and the other is a reserved keyword/token.

mumumu: ah, reserved `token` is easy to understand for me.

cmb69: Right, see https://github.com/php/php-src/blob/4914ead29245be4141cb83d934bd99e23b986d67/Zend/zend_language_parser.y#L102-L180.

jrfnl:
Was just looking at making this change + adding `never` to the list as well. What is this PR waiting for ? (don't really want to create a conflicting PR when someone has already done most of the work

mumumu:
> What is this PR waiting for ?

Nothing. If there is no objection for a while, I want to merge this PR.

@cmb69 How do you think about merging this PR?

cmb69 (approved):
Yes, please go ahead and merge this PR!

PR #939, opened 2021-09-10
Tidy up SVN docs

kamil-tekiela:
An offshoot from #862.

I don't know the extension, but some of the descriptions bothered me as they contained the return values.


Girgias:
Could you maybe set-up the XML for the params, even if you leave their description blank (as this is how undocumented functions should be presented IIRC in the manual).

saundefined (approved):
Looks good

PR #940, opened 2021-09-10
Reword PDOStatement::nextRowset example description

kamil-tekiela:


cmb69:
Thank you! Very nice!

PR #942, opened 2021-09-12
Fix `mixed` return type declarations and parameters of Misc. Functions

alexanderpas:
Noticed that the return types for some functions in the Misc Functions were wrong.

This PR makes sure the return types matches up with the internal return types found in [standard/basic_functions.stub.php.](https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php)

### DIFF
--- a/reference/misc/functions/get-browser.xml
+++ b/reference/misc/functions/get-browser.xml
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis>
    <type class="union"><type>object</type><type>array</type><type>false</type></type><methodname>get_browser</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>user_agent</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>user_agent</parameter><initializer>&null;</initializer></methodparam>
    <methodparam choice="opt"><type>bool</type><parameter>return_array</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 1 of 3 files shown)


reference/misc/functions/get-browser.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>mixed</type><methodname>get_browser</methodname>
+   <type class="union"><type>object</type><type>array</type><type>false</type></type><methodname>get_browser</methodname>
    <methodparam choice="opt"><type>string</type><parameter>user_agent</parameter></methodparam>
```
thinkverse, line 13: Might as well fix the parameter type and the fallback value while you're at it. 🙂

```suggestion
   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>user_agent</parameter><initializer>&null;</initializer></methodparam>
```

thinkverse (changes requested):

reference/misc/functions/get-browser.xml
cmb69: And the fact that `get_browser()` returns `false` on failure should also be mentioned in the return values section.

reference/misc/functions/highlight-string.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>mixed</type><methodname>highlight_string</methodname>
+   <type class="union"><type>string</type><type>bool</type></type><methodname>highlight_string</methodname>
    <methodparam><type>string</type><parameter>str</parameter></methodparam>
```
kamil-tekiela, line 13:
```suggestion
   <methodparam><type>string</type><parameter>string</parameter></methodparam>
```

kamil-tekiela: Same in the parameter section.

kamil-tekiela (changes requested):

kamil-tekiela (approved):

kamil-tekiela:
LGTM. Thanks.

Issue #663, opened 2021-06-06
Add documentation about missing functions introduced in PHP 7 cycle.


Girgias:
Found via the upgrading docs

 - [x]  PHP 7.3
   - [x] `fpm_get_status()` (#1420)
   - [x]  `net_get_interfaces()` (#847)
 - [x] PHP 7.4
   - [x] `get_mangled_object_vars()` (#953)
   - [x] `imagecreatefromtga()` (#950)
   - [x] `pcntl_unshare()` (#958)

PR #950, opened 2021-09-16
Add `imagecreatefromtga` description

saundefined:
Relates to #663


cmb69 (approved):
Thank you! This also needs an entry in https://github.com/php/doc-en/blob/master/reference/image/versions.xml.

saundefined:
> Thank you! This also needs an entry in https://github.com/php/doc-en/blob/master/reference/image/versions.xml.

Thank you! Done

cmb69:
Thank you!

Issue #951, opened 2021-09-16
Typo - Example #1 and #2 titles are misplaced


sivaramkjs:
From manual page: https://php.net/language.attributes.classes

---
I think the **Example `#1` and `#2` titles are misplaced** on this page. They should be as below:

> Example `#1` Simple Attribute Class

> Example `#2` Using target specification to restrict where attributes can be used

PR #952, opened 2021-09-17
Fix attribute example titles

thinkverse:
As pointed out in #951, the titles in the [attribute classes](https://www.php.net/language.attributes.classes) example seems to be mixed up, this PR fixes that.


kamil-tekiela (approved):

kamil-tekiela:
Thanks.

PR #953, opened 2021-09-17
Add `get_mangled_object_vars` description

saundefined:
Relates to #663

### DIFF
--- a/reference/classobj/functions/get-mangled-object-vars.xml
+++ b/reference/classobj/functions/get-mangled-object-vars.xml
@@ -1,0 +1,144 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="function.get-mangled-object-vars" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>get_mangled_object_vars</refname>
+  <refpurpose>Returns an array of mangled object properties</refpurpose>
+ </refnamediv>
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>array</type><methodname>get_mangled_object_vars</methodname>
+   <methodparam><type>object</type><parameter>object</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Returns the mangled properties of the given <parameter>object</parameter>.
+   Does not respect property visibility.
+  </para>
+ </refsect1>
+ 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>object</parameter></term>
+     <listitem>
+      <para>
+       An object instance.
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+ 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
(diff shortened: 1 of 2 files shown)


kamil-tekiela:
Can you please explain in the description what it means "mangled" properties?

saundefined:
> Can you please explain in the description what it means "mangled" properties?

What do you think about the following wording?
>If the same property defines multiple times, PHP "mangles" the property name by including the type of the property and the defining class

--
Also, maybe add a link to [phpinternalsbook](https://www.phpinternalsbook.com/php5/classes_objects/internal_structures_and_implementation.html) in the see also section?

Girgias:
Might be interesting to copy what is written in the converting to array section: https://www.php.net/manual/en/language.types.array.php#language.types.array.casting

reference/classobj/functions/get-mangled-object-vars.xml
```diff
@@ -12,8 +12,12 @@
    <methodparam><type>object</type><parameter>object</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Returns the mangled properties of the given <parameter>object</parameter>.
-   Does not respect property visibility.
+   Returns an &array; whose elements are the <parameter>object</parameter>'s
+   properties. The keys are the member variable names, with a few notable
+   exceptions: integer properties are inaccessible;
```
Girgias, line 17:
```suggestion
   exceptions: integer properties are inaccessible;
```
An object cannot have integer properties since 7.2 (and it was only possible when casting an array to an object).

reference/classobj/functions/get-mangled-object-vars.xml
```diff
@@ -0,0 +1,148 @@
...
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+ 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns an &array; with the same result as <parameter>object</parameter> with the exception
+   that it ignores overloaded array casts, such as used by <classname>ArrayObject</classname>.
```
Girgias, line 44: This sentence is far from clear.

Something like:
> Returns an &array; containing all properties, regardless of visibility, of <parameter>object</parameter>

maybe?

Girgias (changes requested):

reference/classobj/functions/get-mangled-object-vars.xml
saundefined: Looks good, thank you!

reference/classobj/functions/get-mangled-object-vars.xml
```diff
@@ -0,0 +1,146 @@
...
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+ 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns an &array; containing all properties, regardless of visibility, of object.
```
Girgias, line 42:
```suggestion
   Returns an &array; containing all properties, regardless of visibility, of <parameter>object</parameter>.
```
Final nit.

Girgias (approved):

reference/classobj/functions/get-mangled-object-vars.xml
saundefined: Thanks!

PR #954, opened 2021-09-17
ReflectionFunctionAbstract is abstract

midlan:
Abstract keyword was missing in synopsis of class `ReflectionFunctionAbstract`.


cmb69:
Thank you!

PR #955, opened 2021-09-19
DOMDocument::documentElement property can be NULL

eclipxe13:
Property `documentElement` is `NULL` when there is no document element since PHP 5.4.0. https://3v4l.org/vOYYo

```php
var_dump( (new DOMDocument())->documentElement );
NULL
```

Other NULL properties on `DOMDocument` and `DOMNode` has been set before, see https://github.com/php/doc-en/pull/52

### DIFF
--- a/reference/dom/domdocument.xml
+++ b/reference/dom/domdocument.xml
@@ -61,7 +61,7 @@
     <fieldsynopsis>
      <modifier>public</modifier>
      <modifier>readonly</modifier>
-     <type>DOMElement</type>
+     <type class="union"><type>DOMElement</type><type>null</type></type>
      <varname linkend="domdocument.props.documentelement">documentElement</varname>
     </fieldsynopsis>
     <fieldsynopsis>
@@ -200,7 +200,8 @@
      <listitem>
       <para>
        This is a convenience attribute that allows direct access to the
-       child node that is the document element of the document.
+       child node that is the document element of the document. This is
+       &null; when does not exists.
       </para>
      </listitem>
     </varlistentry>


cmb69:
Thank you!

tiffany-taylor:
The wording is a little clumsy.

The simplest improvement would be:
> This is &null; when it does not exist.

But there are other ways it could be further improved:
> This attribute is &null; when the DOM is empty.

> This attribute evaluates to &null; when the DOM is empty.

> If the DOM is empty, this will evaluate to &null;.

Some variation that specifies that the DOM has zero contents (no nodes), which will cause the attribute to evaluate to NULL.

cmb69:
D'oh, poor review from my side. Maybe we should get inspiration [from the specs](https://dom.spec.whatwg.org/#document-element):

> The document element of a document is the element whose parent is that document, if it exists; otherwise null.

eclipxe13:
Thanks for notice the clumsy language. I will make other PR.

I don't like the phrase *"when the DOM is empty"* because the DOM might be not empty and still return NULL. It is only about the `DOMElement`.

```php
$document = new DOMDocument();
$comment = $document->createComment('This is a comment');
$document->appendChild($comment);
var_dump($document->documentElement);
var_dump($document->saveXML());
```
```text
NULL
string(47) "<?xml version="1.0"?>
<!--This is a comment-->
"
```

Based on https://www.php.net/manual/en/class.domnode.php#domnode.props.nextsibling
What about this phrase?

The DOMElement object that is the document element of the document. If there is no such element, this returns null.

Please advice.

tiffany-taylor:
I'll try to give a better review later, but I want to note that "return" isn't the same as "evaluate" in PHP. Something may "evaluate" to NULL and the NULL is printed, but it's not a return. However, I'm not sure if DOMDocument::documentElement is a return or an evaluate.

tiffany-taylor:
I need to improve as far as following up on things. Going to give this a look now.

PR #958, opened 2021-09-21
Add `pcntl_unshare` description

saundefined:
Relates to #663

### DIFF
--- a/reference/pcntl/constants.xml
+++ b/reference/pcntl/constants.xml
@@ -7,450 +7,456 @@
   functions. Please see your systems signal(7) man page for details
   of the default behavior of these signals.
  </para>
- <variablelist>
+
+ <variablelist xml:id="pcntl.constants.common">
+  <title>Process Control constants</title>
   <varlistentry xml:id="constant.wnohang">
    <term>
-    <constant>WNOHANG</constant> 
+    <constant>WNOHANG</constant>
     (<type>int</type>)
    </term>
    <listitem>
     <simpara>
-     
+
     </simpara>
    </listitem>
   </varlistentry>
   <varlistentry xml:id="constant.wuntraced">
    <term>
-    <constant>WUNTRACED</constant> 
+    <constant>WUNTRACED</constant>
     (<type>int</type>)
    </term>
    <listitem>
     <simpara>
-     
+
     </simpara>
    </listitem>
   </varlistentry>
+ </variablelist>
+
+ <variablelist xml:id="pcntl.constants.sig">
+  <title>SIG_* constants</title>
   <varlistentry xml:id="constant.sig-ign">
    <term>
-    <constant>SIG_IGN</constant> 
+    <constant>SIG_IGN</constant>
     (<type>int</type>)
    </term>
    <listitem>
     <simpara>
-     
+
     </simpara>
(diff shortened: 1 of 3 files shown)


reference/pcntl/functions/pcntl-unshare.xml
```diff
@@ -0,0 +1,77 @@
...
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>flags</parameter></term>
+     <listitem>
+      <para>
+       The <parameter>flags</parameter> parameter is a bitmask that specifies which parts of the execution context should be unshared.
+       This parameter is specified by ORing together zero or more of the <literal>CLONE_*</literal> constants:
```
cmb69, line 31: The last character on that line should either be a period, or you may want to list the individual constants.

reference/pcntl/constants.xml
```diff
@@ -965,6 +965,83 @@
     </simpara>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.clone-newns">
```
cmb69, line 968: Maybe add some grouping here (similar to https://www.php.net/manual/en/imagick.constants.php); then later on we could link to that xml:id.

reference/pcntl/functions/pcntl-unshare.xml
```diff
@@ -0,0 +1,77 @@
...
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns <literal>0</literal> on success, <literal>-1</literal> otherwise.
+   On failure it sets an error code, that can be receive with <function>pcntl_get_last_error</function>.
```
cmb69, line 43: Maybe
```suggestion
   On failure it sets an error code, that can be retrieved with <function>pcntl_get_last_error</function>.
```

cmb69 (commented):
Thank you! A few nits, but otherwise looks fine (although I don't know about that function).

reference/pcntl/functions/pcntl-unshare.xml
saundefined: Oops.. fixed it. I listed the constants below. Or is it better to provide a link to the constants group?

reference/pcntl/constants.xml
saundefined: Thanks! Grouped it.

Constants has xml id `constant.clone-newns`, not `pcntl.constant.clone-newns` — is it ok?
Or is it better to add a prefix?

reference/pcntl/functions/pcntl-unshare.xml
cmb69: It's fine this way. :)

reference/pcntl/constants.xml
cmb69: Unless there would be a conflict, `constant.clone-newns` etc. is fine.

Girgias:
I've merged this as one can always improve the documentation afterwards.

Thanks for the contribution!

PR #968, opened 2021-09-23
fixed default value of HTML encoding and decofing functions.

mumumu:
`flags` parameter description is inconsistent with signatures and changelog.


salathe:
See also, today's discussion on this topic in #361

mumumu:
> See also, today's discussion on this topic in #361

OK, it seems that  `ENT_HTML401` is also default value.
I close this PR because stub and signature fix PR will be open in the future.

PR #973, opened 2021-09-25
Reword description on DOMDocument::domElement

eclipxe13:
A better description for documentElement that can be `NULL`.

See #955

### DIFF
--- a/reference/dom/domdocument.xml
+++ b/reference/dom/domdocument.xml
@@ -199,9 +199,8 @@
      <term><varname>documentElement</varname></term>
      <listitem>
       <para>
-       This is a convenience attribute that allows direct access to the
-       child node that is the document element of the document. This is
-       &null; when does not exists.
+       The <classname>DOMElement</classname> object that is the first
+       document element. If not found, this will return &null;.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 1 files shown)


eclipxe13:
@tiffany-taylor could you please review? Thanks in advance

reference/dom/domdocument.xml
```diff
@@ -199,9 +199,8 @@
      <term><varname>documentElement</varname></term>
      <listitem>
       <para>
-       This is a convenience attribute that allows direct access to the
-       child node that is the document element of the document. This is
-       &null; when does not exists.
+       The <classname>DOMElement</classname> object that is the first
+       document element. If not found, this will return &null;.
```
tiffany-taylor, line 203: After testing locally, if the parent isn't found, `DOMDocument::documentElement` evaluates to null. While this doesn't affect most programmers, it can matter for some. I think it's important that the docs specify the distinction.

> If not found, this evaluates to &null;.

tiffany-taylor (changes requested):

reference/dom/domdocument.xml
eclipxe13: Thanks! Rewording to "evaluates to"

tiffany-taylor (approved):

tiffany-taylor:
Thank you!

PR #974, opened 2021-09-26
Fix return value of Imagick::montageImage

LiviaMedeiros:
`Imagick::montageImage` returns a new Imagick, not bool.

Also `thumbnail_geometry` example looks like a typo.
`>` is indeed a valid geometry flag, but it makes sense only for dimensions, not offsets.
Therefore, `120x120+4+3>` would be silently parsed as `120x120>+4+3`
Too misleading for an example value.

### DIFF
--- a/reference/imagick/imagick/montageimage.xml
+++ b/reference/imagick/imagick/montageimage.xml
@@ -77,7 +77,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Creates a composite image and returns it as a new Imagick object.
+   Creates a composite image and returns it as a new <classname>Imagick</classname> object.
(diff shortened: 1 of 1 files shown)


reference/imagick/imagick/montageimage.xml
```diff
@@ -77,7 +77,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &imagick.return.success;
+   Creates a composite image and returns it as a new Imagick object.
```
cmb69, line 80:
```suggestion
   Creates a composite image and returns it as a new <classname>Imagick</classname> object.
```

cmb69:
Thank you!

Issue #972, opened 2021-09-24
[PHP 8.1] Resource to Object Migration


saundefined:
### [Resource to Object Migration](https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.resource2object)

- [x] PgSQL functions (`PgSql\Connection`, `PgSql\Result`, `PgSql\Lob`) #1009
- [x] PSpell functions (`PSpell\Dictionary`, `PSpell\Config`) #976
- [x] LDAP functions (`LDAP\Connection`, `LDAP\Result`, `LDAP\ResultEntry`) #975
- [x] FileInfo functions #985
- [x] IMAP functions #971
- [x] FTP functions #967

mumumu:
@saundefined I want to help this work and am trying to change `PSpell functions`. Do you already work on it?

saundefined:
@mumumu Thanks! Haven't started yet

mumumu:
> @mumumu Thanks! Haven't started yet

OK. I will assign myself to `PSpell` functions.

PR #975, opened 2021-09-26
[PHP 8.1] Use an LDAP instance instead of a resource

saundefined:
Related #972


reference/ldap/functions/ldap-sort.xml
```diff
@@ -34,7 +34,7 @@
      <term><parameter>link</parameter></term>
      <listitem>
       <para>
-       &ldap.parameter.ldap;
+       An LDAP &resource;.
```
Girgias, line 37: This change is incorrect, Also need a changelog entry?

reference/ldap/functions/ldap-parse-exop.xml
```diff
@@ -35,7 +35,7 @@
     <term><parameter>result</parameter></term>
     <listitem>
      <para>
-       An LDAP result resource, returned by <function>ldap_exop</function>.
+      An <classname>LDAP\Result</classname> instance, returned by <function>ldap_exop</function>.
```
Girgias, line 38: Use ldap.parameter.result entity

reference/ldap/functions/ldap-bind-ext.xml
```diff
@@ -30,7 +31,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an LDAP result identifier or &false; on error.
+   Returns an <classname>LDAP\Result</classname> instance,&return.falseforfailure;.
```
Girgias, line 34: Make this an entity maybe? This seems to be used a lot.

Girgias (changes requested):

reference/ldap/functions/ldap-sort.xml
Girgias: As right, is this function removed in 8.0 and has a changelog entry? In that case ignore this.

PR #976, opened 2021-09-27
[PHP 8.1] Use an Pspell\[Dictionary|Config] instance instead of a resource

### LINKED ISSUE #972: [PHP 8.1] Resource to Object Migration

mumumu:
- TODO
  * [x] added Pspell resource description prior to 8.1
    -  Currently, this description does not exist in setup.xml, but exists in [List of Resource Types](https://www.php.net/manual/en/resource.php)
  * [x] added Pspell related object class.
  * [ ] changes pspell function signature and parameters, changelog.

closes #972

### DIFF
--- a/reference/pspell/book.xml
+++ b/reference/pspell/book.xml
@@ -18,6 +18,8 @@
  &reference.pspell.setup;
  &reference.pspell.constants;
  &reference.pspell.reference;
+ &reference.pspell.pspell.dictionary;
+ &reference.pspell.pspell.config;
 
 </book>
 
--- a/reference/pspell/pspell.config.xml
+++ b/reference/pspell/pspell.config.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="utf-8"?>
+<phpdoc:classref xml:id="class.pspell-config" xmlns:phpdoc="http://php.net/ns/phpdoc" xmlns="http://docbook.org/ns/docbook">
+ <title>The PSpell\Config class</title>
+ <titleabbrev>PSpell\Config</titleabbrev>
+
+ <partintro>
+
+  <!-- {{{ PSpell\Config intro -->
+  <section xml:id="pspell-config.intro">
+   &reftitle.intro;
+   <para>
+    A fully opaque class which replaces a <literal>pspell config</literal> resource as of PHP 8.1.0.
+   </para>
+  </section>
+  <!-- }}} -->
+
+  <section xml:id="pspell-config.synopsis">
+   &reftitle.classsynopsis;
+
+   <!-- {{{ Synopsis -->
+   <classsynopsis>
+    <ooclass>
+     <classname>PSpell\Config</classname>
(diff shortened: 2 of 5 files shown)


saundefined (commented):
Thanks! Looks good to me.
Maybe someone else will have any comments

reference/pspell/functions/pspell-new-config.xml
```diff
@@ -44,10 +44,34 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
```
Girgias, line 46: Use ``&pspell.parameter.pspell-config;`` for the config param jsut above?

reference/pspell/functions/pspell-save-wordlist.xml
```diff
@@ -28,7 +28,7 @@
      <term><parameter>dictionary</parameter></term>
      <listitem>
       <para>
-       A dictionary link identifier opened with
+       <classname>PSpell\Dictionary</classname> instance created by
        <function>pspell_new_personal</function>.
```
Girgias, line 32: I'm not sure the detail about the creation is relevant looking at the C source code, but I might be wrong.

Girgias (commented):
LGTM other than the nit and me being somewhat confused if some functions actually care from which function the Directory was created.

reference/pspell/functions/pspell-save-wordlist.xml
Girgias: Yeah, I don't think this is relevant, the example on https://www.php.net/manual/en/function.pspell-save-wordlist.php does not use this function to create the Dict...

reference/pspell/functions/pspell-new-config.xml
mumumu: > Use `&pspell.parameter.pspell-config;` for the config param jsut above?

No, pspell_new_config function returns not PSpell\Config, but PSpell\Dictionary.

https://www.php.net/manual/en/function.pspell-new-config.php

reference/pspell/functions/pspell-save-wordlist.xml
mumumu: > Yeah, I don't think this is relevant,

Right, fixed in https://github.com/php/doc-en/pull/976/commits/513edd168a29ac601332bde4a93acea374413a53

PR #977, opened 2021-09-27
strspn() and strcspn(): Use proper parameter names in docs reference

TobiasBg:
Commit e095023e408c8cb6378ae16bb6870343a3946919 renamed the `$subject`, `$mask`, and `$start` parameters of the `strspn()` and `strcspn()` functions to `$string`, `$characters`, and `$offset`, respectively.

For both of them, three instances in the code example were missed. This PR fixes them.


cmb69:
Thank you!

PR #978, opened 2021-09-28
[PHP 8.1] Update `fputcsv` and `SplFileObject::fputcsv` description

saundefined:
Based on stubs:
- [fputcsv](https://github.com/php/php-src/blob/PHP-8.1/ext/standard/basic_functions.stub.php#L1158)
- [SplFileObject::fputcsv](https://github.com/php/php-src/blob/PHP-8.1/ext/spl/spl_directory.stub.php#L213)

### DIFF
--- a/reference/filesystem/functions/fputcsv.xml
+++ b/reference/filesystem/functions/fputcsv.xml
@@ -10,16 +10,17 @@
   &reftitle.description;
   <methodsynopsis>
    <type class="union"><type>int</type><type>false</type></type><methodname>fputcsv</methodname>
-   <methodparam><type>resource</type><parameter>handle</parameter></methodparam>
+   <methodparam><type>resource</type><parameter>stream</parameter></methodparam>
    <methodparam><type>array</type><parameter>fields</parameter></methodparam>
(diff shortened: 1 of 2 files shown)


reference/filesystem/functions/fputcsv.xml
```diff
@@ -60,22 +61,31 @@
...
        An empty string (<literal>""</literal>) disables the proprietary escape mechanism.
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term><parameter>eol</parameter></term>
+     <listitem>
+      <para>
+       The optional <parameter>eol</parameter> parameter sets
+       a custom End of Line sequence (at most one single-byte character).
```
cmb69, line 78: This parameter is not restricted to a single-byte character; actually, it can be an arbitrary string:
```suggestion
       a custom End of Line sequence.
```
Same for `SplFileObject::fputcsv()` below.

cmb69 (commented):
Thank you!

reference/filesystem/functions/fputcsv.xml
saundefined: Thank you! Fixed

cmb69 (approved):

PR #981, opened 2021-09-28
[PHP 8.1] Deprecated functions and constants

saundefined:


kamil-tekiela:
LGTM

reference/image/functions/imagefilledpolygon.xml
```diff
@@ -82,6 +82,12 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.1.0</entry>
+      <entry>
+       The parameter <parameter>num_points</parameter> has been deprecated.
```
cmb69, line 88: I think that justifies to change the order of the signatures right away, and to mention that the old signature is deprecated.

And the same change needs to be done for imageopenpolygon() and imagepolygon().

reference/imap/constants.xml
```diff
@@ -11,7 +11,7 @@
    </term>
    <listitem>
     <simpara>
-     
+     Constant has been deprecated as of PHP 8.1.0.
```
cmb69, line 14: Maybe just:
```suggestion
     Deprecated as of PHP 8.1.0.
```

reference/mhash/functions/mhash-count.xml
```diff
@@ -30,6 +34,30 @@
...
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.1.0</entry>
+       <entry>
+        This function has been deprecated. Use the <literal>hash_*()</literal> functions instead.
```
cmb69, line 52: Maybe link to the [hash function reference](https://www.php.net/manual/en/ref.hash.php)?

reference/pdo/constants.xml
```diff
@@ -333,6 +333,7 @@
     <simpara>
      As <constant>PDO::FETCH_INTO</constant> but object is provided as a serialized string.
      The class constructor is never called if this flag is set.
+     Constant has been deprecated as of PHP 8.1.0.
```
cmb69, line 336:
```suggestion
     Deprecated as of PHP 8.1.0.
```

cmb69 (commented):
Thank you!

saundefined:
Thanks! Updated, but not sure if I did the signature changes right 🤔

cmb69:
Thank you! I've applied this with a few tweaks.

PR #983, opened 2021-09-29
[PHP 8.1] Add `imageavif` and `imagecreatefromavif` functions

saundefined:
Based on stubs:
- [imageavif](https://github.com/php/php-src/blob/PHP-8.1/ext/gd/gd.stub.php#L142) ([parameters](https://github.com/php/php-src/blob/master/ext/gd/libgd/gd_avif.c#L444))
- [imagecreatefromavif](https://github.com/php/php-src/blob/PHP-8.1/ext/gd/gd.stub.php#L88)

### DIFF
--- a/reference/image/functions/imageavif.xml
+++ b/reference/image/functions/imageavif.xml
@@ -1,0 +1,93 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="function.imageavif" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>imageavif</refname>
+  <refpurpose>&gd.image.output;</refpurpose>
+ </refnamediv>
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>bool</type><methodname>imageavif</methodname>
+   <methodparam><type>GdImage</type><parameter>image</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>resource</type><type>string</type><type>null</type></type><parameter>file</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>quality</parameter><initializer>-1</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>speed</parameter><initializer>-1</initializer></methodparam>
+  </methodsynopsis>
+  
+  <para>
+   Outputs or saves a <acronym>AVIF</acronym> Raster image from the given <parameter>image</parameter>.
+  </para>
+ </refsect1>
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    &gd.image.description;
+    <varlistentry>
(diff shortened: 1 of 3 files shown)


reference/image/functions/imageavif.xml
```diff
@@ -0,0 +1,93 @@
...
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>speed</parameter></term>
+     <listitem>
+      <para>
+       <parameter>speed</parameter> is optional, and ranges from 0 (slower CPU effort, smaller file)
+       to 9 (faster CPU effort, larger file).
+      </para>
+     </listitem>
+    </varlistentry>
```
Girgias, line 52: @cmb69 instead there a constant for explaining the -1 default value?

cmb69: Nope; that matches other functions such as [`imagejpeg()`](https://github.com/php/php-src/pull/7528). It might be a good idea to introduce something like `IMAGE_DEFAULT_QUALITY`, though.

saundefined: @cmb69 As far as I can understand, for example, in `imagejpeg` the quality is about [75](https://github.com/php/doc-en/blob/master/reference/image/functions/imagejpeg.xml#L38), and in `imageavif` - [30](https://github.com/php/php-src/blob/master/ext/gd/libgd/gd_avif.c#L36).

What to do with such differences?
or I'm wrong =)

cmb69: The signature should say `-1`, and the parameter description should explain the details. This looks good here, but you may want to add what the default `-1` actually means.

reference/image/functions/imagecreatefromavif.xml
```diff
@@ -0,0 +1,63 @@
...
+ 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>filename</parameter></term>
+     <listitem>
+      <para>
+        Path to the <acronym>AVIF</acronym> Raster image.
```
cmb69, line 28:
```suggestion
        Path to the <acronym>AVIF</acronym> raster image.
```

reference/image/functions/imageavif.xml
```diff
@@ -0,0 +1,95 @@
...
+       If <literal>-1</literal> is provided, the default value <literal>30</literal> is used.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>speed</parameter></term>
+     <listitem>
+      <para>
+       <parameter>speed</parameter> is optional, and ranges from 0 (slower CPU effort, smaller file)
+       to 10 (faster CPU effort, larger file).
```
cmb69, line 50:
```suggestion
       <parameter>speed</parameter> is optional, and ranges from 0 (slow, smaller file)
       to 10 (fast, larger file).
```

cmb69 (commented):
Thank you. Besides the nits, this looks good to me.

cmb69 (approved):

PR #984, opened 2021-09-29
False isn't allowed for substr()'s $length

dunglas:
When strict types are enabled, passing false currently throws the following error:

> Fatal error: Uncaught TypeError: substr(): Argument #3 ($length) must be of type ?int, bool given

See https://github.com/php/php-src/pull/7533.

### DIFF
--- a/reference/strings/functions/substr.xml
+++ b/reference/strings/functions/substr.xml
@@ -88,7 +88,7 @@ $rest = substr("abcdef", -3, 1); // returns "d"
       </para>
       <para>
        If <parameter>length</parameter> is given and is <literal>0</literal>,
-       &false; or &null;, an empty string will be returned.
+       or &null;, an empty string will be returned.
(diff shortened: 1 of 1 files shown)


reference/strings/functions/substr.xml
```diff
@@ -88,7 +88,7 @@ $rest = substr("abcdef", -3, 1); // returns "d"
       </para>
       <para>
        If <parameter>length</parameter> is given and is <literal>0</literal>,
-       &false; or &null;, an empty string will be returned.
+       or &null;, an empty string will be returned.
```
nikic, line 91: Since PHP 8.0, `null` will return until the end of the string, not an empty string (https://3v4l.org/MbUcs).

kamil-tekiela:
It seems this would require changelog entry.

Crell:
Thanks!

PR #985, opened 2021-09-29
[PHP 8.1] Use an `finfo` instance instead of a resource

saundefined:
Related #972

### DIFF
--- a/reference/fileinfo/functions/finfo-open.xml
+++ b/reference/fileinfo/functions/finfo-open.xml
@@ -63,7 +63,7 @@
   &reftitle.returnvalues;
   <para>
    (Procedural style only)
-   Returns an <classname>finfo</classname> instance on success&return.falseforfailure;.
+   Returns an <classname>finfo</classname> instance on success,&return.falseforfailure;.
   </para>
  </refsect1>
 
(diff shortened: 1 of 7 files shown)


reference/fileinfo/functions/finfo-open.xml
```diff
@@ -63,7 +63,7 @@
   &reftitle.returnvalues;
   <para>
    (Procedural style only)
-   Returns a magic database resource on success&return.falseforfailure;.
+   Returns an <classname>finfo</classname> instance on success&return.falseforfailure;.
```
Girgias, line 66:
```suggestion
   Returns an <classname>finfo</classname> instance on success,&return.falseforfailure;.
```

reference/fileinfo/setup.xml
```diff
@@ -26,7 +26,7 @@
  <section xml:id="fileinfo.resources">
   &reftitle.resources;
   <para>
-   There is one resource used in Fileinfo extension: a magic database
+   Prior to PHP 8.1.0, there was one resource used in Fileinfo extension: a magic database
```
Girgias, line 29:
```suggestion
   Prior to PHP 8.1.0, there was one resource used in the Fileinfo extension: a magic database
```

Girgias (approved):
Two minor nits, otherwise LGTM

PR #986, opened 2021-09-30
[PHP 8.1] Add `array_is_list` description

saundefined:


reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="function.array-is-list" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>array_is_list</refname>
+  <refpurpose>Returns whether a given array is an array with all sequential integers starting from 0</refpurpose>
```
kamil-tekiela, line 5:
```suggestion
  <refpurpose>Checks whether a given &array; is an &array; with keys that are consecutive integers starting from <literal>0</literal></refpurpose>
```
This is important since it doesn't check values in the array, but it checks whether keys are consecutive. Also, the keys are not only sequential but more importantly consecutive.

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+ </refnamediv>
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>bool</type><methodname>array_is_list</methodname>
+   <methodparam><type>array</type><parameter>array</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Returns whether a given <parameter>array</parameter> is an &array; with all sequential integers starting from 0.
```
kamil-tekiela, line 15: Same changes here. You could also add this bit from RFC as part of the description:
>will return true if the array keys are 0 .. count($array)-1 in that order

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &true; if <parameter>array</parameter> is an &array; whose keys are all numeric, sequential, and start at 0.
```
kamil-tekiela, line 38: Again, sequential doesn't explain it properly. The following numbers are sequential: 1,2,3,5,7,11. The keys must consist of consecutive numbers from 0 up to count($array)-1.

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+    <title><function>array_is_list</function> example</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+array_is_list([]); // true
+array_is_list(['apple', 2, 3]); // true
+array_is_list([0 => 'apple', 'orange']); // true
+
+// Key does not start with 0
```
kamil-tekiela, line 55:
```suggestion
// The array does not start at 0
```
or "The array is missing a key at index 0" or "The key with index 0 is missing from the array"

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+array_is_list([1 => 'apple', 'orange']); // false
+
+// Keys are not in order
```
kamil-tekiela, line 58:
```suggestion
// The keys are not consecutive
```

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+array_is_list([1 => 'apple', 0 => 'orange']); // false
+
+// Non-integer keys
+array_is_list([0 => 'apple', 'foo' => 'bar']); // false
+
+// Non-sequential keys
```
kamil-tekiela, line 64:
```suggestion
// Non-consecutive keys
```
However, now it's the same as the other description. Maybe someone has a better idea of how to describe this.

kamil-tekiela (changes requested):

saundefined:
@kamil-tekiela thanks for the review!

What do you think about the following changes:

Refpurpose: so as not to repeat the same thing several times and use a shorter description:
>Checks whether a given <parameter>array</parameter> is a list

Methodsynopsis:
>Checks whether a given <parameter>array</parameter> keys are <literal>0 .. count($array)-1</literal> in that order.

Returnvalues:
>Returns &true;, if the <parameter>array</parameter> keys are <literal>0 .. count($array)-1</literal> in that order.

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+ </refnamediv>
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>bool</type><methodname>array_is_list</methodname>
+   <methodparam><type>array</type><parameter>array</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Checks whether a given <parameter>array</parameter> keys are <literal>0 .. count($array)-1</literal> in that order.
```
kamil-tekiela, line 15: This is good, but I wonder if we can give 140%. If we must duplicate one or the other, I'd prefer the refpurpose and the description to be similar. 
```suggestion
   Determines if the given <parameter>array</parameter> is a list.
   An &array; is considered a list if its keys consist of consecutive numbers
   from <literal>0</literal> to <literal>count($array)-1</literal>.
```

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+<?php
+
+array_is_list([]); // true
+array_is_list(['apple', 2, 3]); // true
+array_is_list([0 => 'apple', 'orange']); // true
+
+// The array does not start at 0
+array_is_list([1 => 'apple', 'orange']); // false
+
+// The keys are not consecutive
```
kamil-tekiela, line 58:
```suggestion
// The keys are not in the correct order
```

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &true;, if the <parameter>array</parameter> keys are <literal>0 .. count($array)-1</literal> in that order.
```
kamil-tekiela, line 38:
```suggestion
   Returns &true; if <parameter>array</parameter> is a list, &false;
   otherwise.
```

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>array</parameter></term>
+     <listitem>
+      <para>
+       An &array;.
```
kamil-tekiela, line 27:
```suggestion
       The &array; being evaluated. 
```

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,99 @@
...
+    </programlisting>
+   </example>
+  </para>
+ </refsect1>
+
+ <refsect1 role="notes">
+  &reftitle.notes;
+  <para>
+   This function returns &true; on empty arrays as well.
+  </para>
```
kamil-tekiela, line 76: If it's ought to be a note, then make it a note. 

```suggestion
  <note>
   <para>
    This function returns &true; on empty arrays as well.
   </para>
  </note>
```

kamil-tekiela (changes requested):
Sorry for changing so much, but I really believe this function needs a clear description. I think that what I am proposing now would be sufficient and it follows the style of the other functions in PHP manual.

saundefined:
Thank you very much for your help!

kamil-tekiela:
It'd be nice if we could get another set of eyes on this before merging. It looks good to me as it is now. 
@cmb69 ?

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,104 @@
...
+
+// The keys are not in the correct order
+array_is_list([1 => 'apple', 0 => 'orange']); // false
+
+// Non-integer keys
+array_is_list([0 => 'apple', 'foo' => 'bar']); // false
+
+// Non-consecutive keys
+array_is_list([0 => 'apple', 2 => 'bar']); // false
+]]>
```
salathe, line 69: There should be a closing PHP tag (manual convention).

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,104 @@
...
+    </programlisting>
+   </example>
+  </para>
+ </refsect1>
+
+ <refsect1 role="notes">
+  &reftitle.notes;
+  <note>
+   <para>
+    This function returns &true; on empty arrays as well.
```
salathe, line 79: Nit: "as well" as what?

```suggestion
    This function returns &true; on empty arrays.
```

reference/array/functions/array-is-list.xml
```diff
@@ -0,0 +1,104 @@
...
+   </para>
+  </note>
+ </refsect1>
```
salathe, line 82: It would be good to have a "see also" section, with at least `array_values()` (a way of making an arbitrary array into a "list").

reference/array/functions/array-is-list.xml
kamil-tekiela: I am not going to argue against adding it, but I don't really see the point. The closing PHP tag is generally considered bad practice if there's no output following the PHP block. Adding it to every example in PHP manual works against that recommendation. I'd only make it a rule that it should be consistent across examples on any given page.

salathe: Regardless of whether you see the point or not, the tag should be there for this example.

saundefined:
Thanks! Committed changes

kamil-tekiela (approved):

PR #991, opened 2021-10-03
Retain original $host parameter name of ldap_connect()

kocsismate:


cmb69:
This PR looks good to me. This function is basically overloaded: either pass the URI, or pass hostname and port. The latter won't work with named arguments, but is soft deprecated anyway. And as it is now, we have two descriptions of the $uri parameter.

kocsismate:
Yeah, I agree with Christoph: while I admit that this PR is very far from perfect, but at least it retains the previous state of the page by fixing an issue with the duplicated param name. As I have no intention for further improving the documentation of `ldap_connect`, I think it's a sensible hotfix for now. Hopefully someone in the future will feel the urge to review the content of this page, and will the document all the missing parameters.

cmb69:
plus actually deprecate the old signature...

afilina:
Pinging reviewer @cmb69: it seems like you gave an informal approval in the comments. Can we formalize the approval or merge?

cmb69:
Oh, indeed, this could be merged. Thank you!

PR #994, opened 2021-10-05
[PHP 8.1] Use GdFont

saundefined:
As of PHP [8.1.0](https://github.com/php/php-src/blob/PHP-8.1/ext/gd/gd.stub.php), a GdFont class was added.

Not sure about `gd font` resource.. because PHP 8.0 already [has](https://github.com/php/php-src/blob/PHP-8.0/ext/gd/gd.stub.php) int values instead of resources 🤔

### DIFF
--- a/appendices/resources.xml
+++ b/appendices/resources.xml
@@ -559,7 +559,7 @@
     <entry>
      <function>imagedestroy</function>
     </entry>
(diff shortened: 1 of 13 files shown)


cmb69 (approved):
Thank you!

PR #995, opened 2021-10-05
Fix the link error on the front page of the manual

sy-records:
https://www.php.net/manual/en/

image

Because `index` is overwritten after a conflict

https://github.com/php/doc-en/blob/8958abfe6c950a41d3bfd0752a88ad6f4a04bb9d/reference/mongodb/book.developer.xml#L38-L41


salathe (approved):
Good catch. 👍

PR #996, opened 2021-10-05
Generate methodsynopses from stubs for ext/intl (IntlCalendar)

kocsismate:


reference/intl/intlcalendar/add.xml
```diff
@@ -44,10 +43,10 @@
   &reftitle.parameters;
   <variablelist>
    <varlistentry>
-    <term><parameter>cal</parameter></term>
+    <term><parameter>calendar</parameter></term>
     <listitem>
      <para>
-      The IntlCalendar resource.
+      The <classname>IntlCalendar</classname> object.
```
Girgias, line 49: Add an entity for this an reuse it as we did for other extensions:
The text should be: ``An <classname>IntlCalendar</classname> instance.`` (IIRC that's how its worded for pspell).

reference/intl/intlcalendar/after.xml
```diff
@@ -56,8 +55,7 @@
   <para>
    Returns &true; if this objectʼs current time is after that of the
    <parameter>calendar</parameter> argumentʼs time. Returns &false; otherwise.
-   Also returns &false; on failure. You can use <link
-   linkend="ini.intl.use-exceptions">exceptions</link> or
+   Also returns &false; on failure. You can use <link linkend="ini.intl.use-exceptions">exceptions</link> or
    <function>intl_get_error_code</function> to detect error conditions.
```
Girgias, line 59: Maybe convert this to an entity in language snippets too, as this seems to be used quite a bit, and reword it as following.

```suggestion
   On failure &false; is also returned. To detect error conditions use  <function>intl_get_error_code</function>,
   or set up Intl to throw <link linkend="ini.intl.use-exceptions">exceptions</link>.
```

reference/intl/intlcalendar/geterrorcode.xml
```diff
@@ -57,7 +56,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   An ICU error code indicating either success, failure or a warning. 
+   An ICU error code indicating either success, failure or a warning. Returns &false; on failure.
```
Girgias, line 59:
```suggestion
   An ICU error code indicating either success, failure or a warning. Returns &false; when there is no error to fetch.
```
Or something similar?

reference/intl/intlcalendar/geterrormessage.xml
```diff
@@ -11,13 +10,13 @@
...
-   <void />
-   </methodsynopsis>
+   <methodsynopsis role="oop">
+   <modifier>public</modifier> <type class="union"><type>string</type><type>false</type></type><methodname>IntlCalendar::getErrorMessage</methodname>
+   <void/>
+  </methodsynopsis>
   <para>&style.procedural;:</para>
   <methodsynopsis role="procedural">
-   <type>string</type><methodname>intlcal_get_error_message</methodname>
+   <type class="union"><type>string</type><type>false</type></type><methodname>intlcal_get_error_message</methodname>
```
Girgias, line 19: Return value section needs to be updated to include false.

reference/intl/intlcalendar/gettime.xml
```diff
@@ -12,16 +11,16 @@
   <para>
    &style.oop;
   </para>
-  <methodsynopsis>
-   <modifier>public</modifier> <type>float</type><methodname>IntlCalendar::getTime</methodname>
-   <void />
+  <methodsynopsis role="oop">
+   <modifier>public</modifier> <type class="union"><type>float</type><type>false</type></type><methodname>IntlCalendar::getTime</methodname>
```
Girgias, line 15: Return value section needs to be updated for false.

reference/intl/intlcalendar/gettimezone.xml
```diff
@@ -12,16 +11,16 @@
   <para>
    &style.oop;
   </para>
-  <methodsynopsis>
-   <modifier>public</modifier> <type>IntlTimeZone</type><methodname>IntlCalendar::getTimeZone</methodname>
-   <void />
+  <methodsynopsis role="oop">
+   <modifier>public</modifier> <type class="union"><type>IntlTimeZone</type><type>false</type></type><methodname>IntlCalendar::getTimeZone</methodname>
```
Girgias, line 15: Return value sections needs to be updated for false.

Girgias (changes requested):

reference/intl/intlcalendar/add.xml
kocsismate: Yes, you were right, although it should have been `A <classname>PSpell\Dictionary</classname> ...` instead of `An <classname>PSpell\Dictionary</classname> ...`.

reference/intl/intlcalendar/geterrorcode.xml
kocsismate: Is it true actually?

```
/* Fetch the object (without resetting its last error code ). */
co = Z_INTL_CALENDAR_P(object);
if (co == NULL)
    RETURN_FALSE;
```

So it returns false when the `Calendar_object` cannot be retrieved, doesn't it?

reference/intl/intlcalendar/equals.xml
```diff
@@ -57,13 +54,11 @@
   <para>
    Returns &true; if the current time of both this and the passed in
    <classname>IntlCalendar</classname> object are the same, or &false;
```
Girgias, line 56:
```suggestion
   <classname>IntlCalendar</classname> object are the same, &return.falseforfailure ;
```
Or maybe just use ``&return.success;`` for the return value section?

reference/intl/intlcalendar/getactualmaximum.xml
```diff
@@ -21,16 +20,14 @@
...
   </methodsynopsis>
   <para>
    Returns a fieldʼs relative maximum value around the current time. The exact
    semantics vary by field, but in the general case this is the value that
-   would be obtained if one would set the field value into the <link
-   linkend="intlcalendar.getleastmaximum">smallest relative maximum</link> for
-   the field and would increment it until reaching the <link
-   linkend="intlcalendar.getmaximum">global maximum</link> or the field value
+   would be obtained if one would set the field value into the <link linkend="intlcalendar.getleastmaximum">smallest relative maximum</link> for
+   the field and would increment it until reaching the <link linkend="intlcalendar.getmaximum">global maximum</link> or the field value
```
Girgias, line 30: Could you put the whole link tag on news lines here? Just to try to reduce the line length.

reference/intl/intlcalendar/getactualminimum.xml
```diff
@@ -21,16 +20,14 @@
...
   </methodsynopsis>
   <para>
    Returns a fieldʼs relative minimum value around the current time. The exact
    semantics vary by field, but in the general case this is the value that
-   would be obtained if one would set the field value into the <link
-   linkend="intlcalendar.getgreatestminimum">greatest relative minimum</link>
-   for the field and would decrement it until reaching the <link
-   linkend="intlcalendar.getminimum">global minimum</link> or the field value
+   would be obtained if one would set the field value into the <link linkend="intlcalendar.getgreatestminimum">greatest relative minimum</link>
+   for the field and would decrement it until reaching the <link linkend="intlcalendar.getminimum">global minimum</link> or the field value
```
Girgias, line 30: Same could you put the link tags on new lines? :)

reference/intl/intlcalendar/indaylighttime.xml
```diff
@@ -47,11 +44,8 @@
   &reftitle.returnvalues;
   <para>
    Returns &true; if the date is in Daylight Savings Time, &false; otherwise.
```
Girgias, line 46: Maybe just ``&return.success;`` ?

Girgias (commented):
There is still one remaining comment from the previous review which does need to be addressed but other than the couple of other nits it looks alright.

reference/intl/intlcalendar/indaylighttime.xml
kocsismate: I'd prefer if we kept the current wording (same for `equals`), I find it more correct.

kocsismate:
Has the last commit addressed your comments, right? Can I merge it?

PR #1000, opened 2021-10-06
Update method return

Luiz8966:
The _item()_ method returns a _DOMElement_ object, but I can't specify which version of PHP it was modified from


cmb69:
No, not necessarily, see https://3v4l.org/dru8d.

Luiz8966:
But what about this case where I use a method of the _DOMElement_ class?

[https://3v4l.org/eLoFT](url)

cmb69:
A DOMElement is a DOMNode, and we need to document the most general type.

PR #1002, opened 2021-10-07
[PHP 8.1] Add `fdatasync` and `fsync` description

saundefined:
Added [`fdatasync`](https://github.com/php/php-src/blob/PHP-8.1/ext/standard/basic_functions.stub.php#L1109) and [`fsync`](https://github.com/php/php-src/blob/PHP-8.1/ext/standard/basic_functions.stub.php#L1106) description.


cmb69 (approved):
Thank you! Looks good to me.

PR #1004, opened 2021-10-07
Reword parameter description of set_exception_handler

kamil-tekiela:
This removes outdated PHP 5 information.

### DIFF
--- a/reference/errorfunc/functions/set-exception-handler.xml
+++ b/reference/errorfunc/functions/set-exception-handler.xml
@@ -30,9 +30,10 @@
      <listitem>
       <para>
        The function to be called when an uncaught exception occurs.
-       Most errors are reported by throwing <classname>Error</classname>
-       exceptions, which will be caught by the handler as well. Both <classname>Error</classname>
-       and <classname>Exception</classname> implements the <classname>Throwable</classname> interface.
+       This handler function needs to accept one parameter,
+       which will be the <classname>Throwable</classname> object that was thrown.
+       Both <classname>Error</classname> and <classname>Exception</classname>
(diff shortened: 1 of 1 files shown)


reference/errorfunc/functions/set-exception-handler.xml
```diff
@@ -29,22 +29,15 @@
...
-      <para>
-       <methodsynopsis>
-        <type>void</type><methodname><replaceable>handler</replaceable></methodname>
-        <methodparam><type>Exception</type><parameter>ex</parameter></methodparam>
-       </methodsynopsis>
-      </para>
-      <para>
-       Since PHP 7, most errors are reported by throwing <classname>Error</classname>
+       was thrown. 
+       Most errors are reported by throwing <classname>Error</classname>
```
Crell, line 36: An `Error` is a throwable, not an `Exception`, so this paragraph should be talking about catching throwables, not exceptions.  (It's a silly design, but such is history.)

cmb69: Good point, although catching `Error`s is discouraged, because these are considered programming errors which should never happen in production.

Crell: Adding a note for that is possibly a good idea.  But the description should be clear that this function applies to call throwables, which includes the Exception and Error sub-categories.

kamil-tekiela: Thanks! I completely forgot about this PR. I reworded the whole thing and made it clearer. Is it better?

cmb69 (approved):
Thank you!

PR #1005, opened 2021-10-08
dirname(): add warning about risk of infinite loop

jrfnl:
🤞🏻 I hope I got the syntax correct ;-)

### DIFF
--- a/reference/filesystem/functions/dirname.xml
+++ b/reference/filesystem/functions/dirname.xml
@@ -82,6 +82,25 @@
    <parameter>path</parameter> with any trailing
    <literal>/component</literal> removed.
   </para>
+
+  <caution>
+   <para>
+    Be careful when using this function in a loop that can reach the
+    top-level directory as this can result in an infinite loop.
+    <informalexample>
+     <programlisting role="php">
+<![CDATA[
+<?php
+dirname('.');    // Will return '.'.
+dirname('/');    // Will return `\` on Windows and '/' on *nix systems.
+dirname('\\');   // Will return `\` on Windows and '.' on *nix systems.
+dirname('C:\\'); // Will return 'C:\' on Windows and '.' on *nix systems.
+?>
+]]>
+     </programlisting>
+    </informalexample>
+   </para>
+  </caution>
  </refsect1>
 
  <refsect1 role="changelog">


cmb69:
That is [not Windows specific](https://3v4l.org/Po1fg), though. I suggest to reword ("top level directory"?)

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,18 @@
...
    <literal>/component</literal> removed.
   </para>
+
+  <warning xmlns="http://docbook.org/ns/docbook">
+   <para>
+    Be careful when using this function in a loop which can reach the
+    top-level directory, as that could result in an infinite loop.
+   </para>
+   <para>
+    <literal>dirname( '.' )</literal> will return <literal>'.'</literal>;
```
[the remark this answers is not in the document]
jrfnl: Done.

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,18 @@
...
+  <warning xmlns="http://docbook.org/ns/docbook">
+   <para>
+    Be careful when using this function in a loop which can reach the
+    top-level directory, as that could result in an infinite loop.
+   </para>
+   <para>
+    <literal>dirname('.')</literal> will return <literal>'.'</literal>;
+    <literal>dirname('/')</literal> will return <literal>'/'</literal>;
+    <literal>dirname('C:\')</literal> will return <literal>'C:\'</literal>.
+   </para>
```
cmb69, line 95: That won't look as intended (everything would be on the same line). You could use [a `<simplelist>`](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.mbstring), but why not use [an `<informalexample>` ](https://github.com/php/doc-en/blob/21129de1745eb016452f0ce8a2c3e47fbb8484de/appendices/migration81/new-features.xml#L16-L25).

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,18 @@
...
+
+  <warning xmlns="http://docbook.org/ns/docbook">
+   <para>
+    Be careful when using this function in a loop which can reach the
+    top-level directory, as that could result in an infinite loop.
+   </para>
+   <para>
+    <literal>dirname('.')</literal> will return <literal>'.'</literal>;
+    <literal>dirname('/')</literal> will return <literal>'/'</literal>;
+    <literal>dirname('C:\')</literal> will return <literal>'C:\'</literal>.
```
cmb69, line 94: `'C:\'` is a syntax error; the backslash needs to be escaped (`'C:\\'`). More importantly, [`dirname('C:\\')` returns `'.' ` on Linux systems](https://3v4l.org/lcjB3), what should be clarified.

reference/filesystem/functions/dirname.xml
kamil-tekiela: Was this not meant to be in one line? It's all one sentence and a single sentence is not usually broken up into multiple lines. I do agree an informalexample might be a better choice here.

jrfnl: Yes, I wrote it to be displayed on one line, though am perfectly happy to change the formatting if you think using `simplelist` would display better.

reference/filesystem/functions/dirname.xml
jrfnl: Thanks for pointing that out. Updated in the commit.

I realize the focus of the original text was very much Windows, that was largely because we'd just been fixing exactly such an infinite loop for Windows ;-)

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,24 @@
    <parameter>path</parameter> with any trailing
    <literal>/component</literal> removed.
   </para>
+
+  <warning xmlns="http://docbook.org/ns/docbook">
+   <para>
+    Be careful when using this function in a loop which can reach the
```
kamil-tekiela, line 88: The word "which" does not suit here. I think you meant to use "that"
```suggestion
    Be careful when using this function in a loop that can reach the
```

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,24 @@
...
+    top-level directory, as that could result in an infinite loop.
```
kamil-tekiela, line 89:
```suggestion
    top-level directory, as this can result in an infinite loop.
```
Not sure about this one, but I believe when we refer to the previous clause in the same sentence we use "this" not "that"

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,24 @@
...
+    <informalexample>
+     <programlisting role="php">
+<![CDATA[
+<?php
+dirname('.');    // Will return '.'.
+dirname('/');    // Will return '/'.
```
kamil-tekiela, line 95: This returns `\` on Windows. I don't really know why though.

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,24 @@
    <parameter>path</parameter> with any trailing
    <literal>/component</literal> removed.
   </para>
+
+  <warning xmlns="http://docbook.org/ns/docbook">
```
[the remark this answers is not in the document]
jrfnl: Fair. Updated.

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,24 @@
    <parameter>path</parameter> with any trailing
    <literal>/component</literal> removed.
   </para>
+
+  <caution>
```
salathe, line 86: The `Return Values` section is not the correct place for notes like this.  Instead, either add a `Notes` section or (since there are a few already) add another item in the `Description`.

salathe (dismissed):

reference/filesystem/functions/dirname.xml
jrfnl: @salathe I did actually consider both when I created the pull request, but as this "gotcha" is very specifically related to potentially "unexpected" return values of this function, it still felt like the "Return Values" section was the most appropriate place for it.

cmb69: I agree; this is similar to [`&return.falseproblem;`](https://github.com/php/doc-en/blob/master/language-snippets.ent#L681-L687).

salathe: Do whatever you want; ignore this old goat.

jrfnl:
> LGTM, but I wonder if we should mention `dirname('\\')` also.

Good point. Added.

reference/filesystem/functions/dirname.xml
```diff
@@ -82,6 +82,25 @@
...
+   <para>
+    Be careful when using this function in a loop that can reach the
+    top-level directory, as this can result in an infinite loop.
```
kamil-tekiela, line 89:
```suggestion
    Be careful when using this function in a loop that can reach the
    top-level directory as this can result in an infinite loop.
```

jrfnl:
Just checking in: was there anything on my end this PR is waiting for ?

cmb69:
I've merged this now. The note(s) could still be moved to another section later.

jrfnl:
Thanks @cmb69 !

PR #1007, opened 2021-10-10
improve headers_list() example

hakre:
Two things with the PHP example code:

1. `X-` prefix header names June 2012 [1], [2].

2.  PHP prepends the `default_charset` [3] on any `text/*` media type in Content-Type header (unless the `charset` is with the `header()` call. The example code output now has it (with the UTF-8 default).

[1]: https://datatracker.ietf.org/doc/html/rfc6648
[2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
[3]: https://www.php.net/manual/en/ini.core.php#ini.default-charset

### DIFF
--- a/reference/network/functions/headers-list.xml
+++ b/reference/network/functions/headers-list.xml
@@ -5,7 +5,7 @@
   <refname>headers_list</refname>
   <refpurpose>Returns a list of response headers sent (or ready to send)</refpurpose>
  </refnamediv>
- 
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
@@ -35,7 +35,7 @@
   &reftitle.examples;
   <para>
    <example>
-    <title>Examples using <function>headers_list</function></title>
+    <title>Example using <function>headers_list</function></title>
     <programlisting role="php">
 <![CDATA[
 <?php
@@ -45,10 +45,10 @@ setcookie('foo', 'bar');
 
 /* Define a custom response header
    This will be ignored by most clients */
-header("X-Sample-Test: foo");
+header("Example-Test: foo");
 
 /* Specify plain text content in our response */
-header('Content-type: text/plain');
+header('Content-Type: text/plain; charset=UTF-8');
 
 /* What headers are going to be sent? */
 var_dump(headers_list());
@@ -56,18 +56,16 @@ var_dump(headers_list());
 ?>
 ]]>
     </programlisting>
-    &example.outputs;
+    &example.outputs.similar;
     <screen>
 <![CDATA[
-array(4) {
+array(3) {
   [0]=>
-  string(23) "X-Powered-By: PHP/5.1.3"
-  [1]=>
   string(19) "Set-Cookie: foo=bar"
+  [1]=>
+  string(17) "Example-Test: foo"
   [2]=>
-  string(18) "X-Sample-Test: foo"
-  [3]=>
-  string(24) "Content-type: text/plain"
+  string(39) "Content-Type: text/plain; charset=UTF-8"
 }
 
 ]]>


tiffany-taylor (commented):
https://github.com/php/doc-en/pull/1007/files#diff-d4e94845aca1cba6c7192a0c37d9e4366d98fdc523ae910a0d6a29c85c833469R63-R64

This could probably be removed and https://github.com/php/doc-en/pull/1007/files#diff-d4e94845aca1cba6c7192a0c37d9e4366d98fdc523ae910a0d6a29c85c833469R65 updated to `[0]`.

salathe (changes requested):
Thanks @hakre !

Added a few details that need to be fixed, as suggested changes.

Other things to consider changing, though not *required*:

- Change `&example.outputs;` to `&examples.outputs.similar;` (i.e. to say `The above examples will output something similar to:` ) since the output may vary by PHP version, configuration, SAPI (?), etc.  These XML entities, and many others, can be found in the `language-snippets.ent` file in the top-level directory of the repo.
- ~~Update, or remove for simplicity, the `X-Powered-By` header. PHP 5.1.3 is pretty ancient.~~ Edit: it looks like you updated the version number while I was typing this. :)

reference/network/functions/headers-list.xml
```diff
@@ -65,9 +65,9 @@ array(4) {
   [1]=>
   string(19) "Set-Cookie: foo=bar"
   [2]=>
-  string(18) "X-Sample-Test: foo"
+  string(18) "Example-Test: foo"
```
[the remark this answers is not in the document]
hakre: edited.

reference/network/functions/headers-list.xml
```diff
@@ -65,9 +65,9 @@ array(4) {
...
   [3]=>
-  string(24) "Content-type: text/plain"
+  string(24) "Content-type: text/plain; charset=UTF-8"
```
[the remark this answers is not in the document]
hakre: edited.

hakre:
There seems to be some non-signigifcant whitespace issue as well, I'll add it.

hakre:
> I think I preferred your first version where you added charset. The header produced by PHP is nonconformant from what I see.

Not sure I get what you mean. The header produced by PHP looks fine, doesn't it?:

~~~
Content-type: text/plain; charset=UTF-8
~~~

HTTP Header name fine, mime media type fine, charset attribute fine. not? @kamil-tekiela

reference/network/functions/headers-list.xml
```diff
@@ -45,29 +45,29 @@ setcookie('foo', 'bar');
...
-  string(23) "X-Powered-By: PHP/5.1.3"
+  string(24) "X-Powered-By: PHP/8.0.11"
   [1]=>
   string(19) "Set-Cookie: foo=bar"
   [2]=>
-  string(18) "X-Sample-Test: foo"
+  string(17) "Example-Test: foo"
   [3]=>
-  string(24) "Content-type: text/plain"
+  string(38) "Content-type: text/plain; charset=UTF-8"
```
salathe, line 70: (When) Does PHP actually produce this exact header?  At least from 5 minutes of limited testing (with the CLI webserver and FPM), it would be: `Content-type: text/plain;charset=UTF-8`

hakre: Yes. I fixed that. The extra space after semicolon was by me in error.

salathe: Thanks.

kamil-tekiela:
Thanks for pointing to the spec. I am aware the spaces are optional and the type, subtype and parameter are case-insensitive. I am not sure about the header name, but as I see user agents treat it in case insensitive way. 

My point is that when we document the output header, it would be safer to use the variant where PHP doesn't change the case or white space. What if we decide to change the `Content-type` to `Content-Type` in a future PHP release? If it's possible to document in such a way that we can be sure of what the output is without relying on the way PHP prepares it internally, or some INI directive, then I think that would be preferable.

hakre:
> My point is that when we document the output header, it would be safer to use the variant where PHP doesn't change the case or white space.  What if we decide to change the `Content-type` to `Content-Type` in a future PHP release?

Well it's docs and the example output is taken from the PHP version as in the X-Powered-By header. No decision for changes in src from my end, the examples should not imply anything regarding that from my perspective.

Similar to the optional white space, case does not matter here (the header names are equivalent, similar to the token and quoted token in the media type. For the header names that is in the HTTP specs, MDN should have it as well as IETF.

* [rfc7230 3.2.  Header Fields](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2)

Therefore I don't see an issue with that. Also `&examples.outputs.similar;` has been edited in.

Additional ref regarding case-sensitivity: [rfc4790 - 9.2.1.  ASCII Casemap Collation Description](https://www.rfc-editor.org/rfc/rfc4790#section-9.2.1) (non-locale-aware, Posix "C" locale like)

But perhaps too technical for a docs discussion maybe @kamil-tekiela

kamil-tekiela:
Sorry, maybe my language skills are at fault here. I am not questioning the RFC or why PHP outputs the header the way it does. All I am saying is that it's a little confusing for readers of the PHP manual to see that you are sending `Content-Type` but the output example says `Content-type`. This could be easily avoided if you send the header in full with the charset. This way, the case in the example's code and the example's output will match. Also, specifying the charset in the header in code, will not leave readers wondering where the charset came from. 

I am not sure how the discussion became technical; I did not mean to make it technical. All I said is that the example looked better when you had:

```
/* Specify plain text content in our response */
header('Content-Type: text/plain; charset=UTF-8');
```
The output will then be exactly what you provided to the `header()` function. 
If I am still not making sense you can ping me in chat in room 11, maybe I will be able to explain it better.

hakre:
Modified to `Content-Type: text/plain; charset=UTF-8`.

reference/network/functions/headers-list.xml
```diff
@@ -45,29 +45,29 @@ setcookie('foo', 'bar');
...
+header('Content-Type: text/plain; charset=UTF-8');
 
 /* What headers are going to be sent? */
 var_dump(headers_list());
 
 ?>
 ]]>
     </programlisting>
-    &example.outputs;
+    &examples.outputs.similar;
```
kamil-tekiela, line 59:
```suggestion
    &example.outputs.similar;
```

kamil-tekiela (changes requested):
Thanks. Other than this one small change it looks good.

reference/network/functions/headers-list.xml
```diff
@@ -45,29 +45,27 @@ setcookie('foo', 'bar');
...
 var_dump(headers_list());
 
 ?>
 ]]>
     </programlisting>
-    &example.outputs;
+    &example.outputs.similar;
     <screen>
 <![CDATA[
 array(4) {
```
kamil-tekiela, line 62:
```suggestion
array(3) {
```

PR #1011, opened 2021-10-11
Remove unnecessary ext/date methodsynopsis roles

kocsismate:
These are necessary to reduce diff with the output of `gen_stub.php`. Method lists on the class synopsis pages are left intact.

`DateTime`:
Screenshot 2021-10-11 at 22 38 20

`DateTimeImmutable`:
Screenshot 2021-10-11 at 22 38 47

`DateTimeInterface`:
Screenshot 2021-10-11 at 22 38 35

### DIFF
--- a/reference/datetime/dateinterval/construct.xml
+++ b/reference/datetime/dateinterval/construct.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-
 <refentry xml:id="dateinterval.construct" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>DateInterval::__construct</refname>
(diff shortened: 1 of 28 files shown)


kocsismate:
> What criteria is used to decide which role="oop" to keep, and which to remove?

I try to match what `gen_stub.php` does: it only adds `role="oop"` and `role="procedural"` for OO-procedural aliases (like `mysqli_stmt::fetch()` and `mysqli_stmt_fetch()`). Probably ext/date is the only exception (but there might be a few other method synopsis pages as well) where there's a need for this role in spite for method-function pairs which are not aliases.

> No <xi:fallback/>s?

No. I manually edited these queries, and I though there's no need for them. Or was I wrong? However, I've just realized that since `gen_stub.php`'s class synopsis generator functionality cannot generate custom include queries like this.. So it would be nice to find a general solution somehow. I just don't have a good idea yet.. other than adding a `role` attribute for every method. Do you have any suggestion by chance?

> The xpointer in datetime.xml is convoluted.

Yes, I agree. It's simply the only query I could come up with which doesn't require manual modification of `gen_stubs`'s method synopsis output.

kocsismate:
I'm closing this in order to create a few smaller PRs

PR #1012, opened 2021-10-11
Fix a few ext/GD parameter names and return types

kocsismate:


reference/image/functions/imagecolorexactalpha.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>imagecolorexactalpha</methodname>
+   <type class="union"><type>int</type><type>false</type></type><methodname>imagecolorexactalpha</methodname>
```
kamil-tekiela, line 11: When does it return false? The return section doesn't mention anything about false.

reference/image/functions/imagefilledpolygon.xml
```diff
@@ -42,20 +42,20 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>num_points</parameter></term>
+     <term><parameter>num_points_or_color</parameter></term>
```
kamil-tekiela, line 45: I'm sorry but I don't understand the motivation behind this. This is one of these hard to document functions. Either the function synopsis has to be updated together with this change or the whole thing should be left alone.

reference/image/functions/imagecolorexactalpha.xml
cmb69: The return section is correct (imagecolorexactalpha() returned false prior to PHP 8.0.0, if the given resource was invalid, but that should not be documented; undefined behavior).

kocsismate:
I'm closing this PR, none of the changes are necessary. I'll fix the superfluous `false` return type in the stubs.

PR #1013, opened 2021-10-11
Generate a few ext/standard method synopses from stubs

kocsismate:


reference/info/functions/phpversion.xml
```diff
@@ -37,13 +37,35 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   If the optional <parameter>extension</parameter> parameter is
-   specified, <function>phpversion</function> returns the version of that
+   If the optional <parameter>extension</parameter> parameter is a
+   <type>string</type>, <function>phpversion</function> returns the version of that
    extension, or &false; if there is no version information associated or
    the extension isn't enabled.
```
kamil-tekiela, line 43:
```suggestion
   Returns the current PHP version as a <type>string</type>.
   If a <type>string</type> argument is provided for <parameter>extension</parameter> parameter, <function>phpversion</function> returns the version of that
   extension, or &false; if there is no version information associated or
   the extension isn't enabled.
```

reference/info/functions/version-compare.xml
```diff
@@ -66,13 +61,13 @@
      <listitem>
       <para>
        If the third optional <parameter>operator</parameter> argument is
-       specified, test for a particular relationship. The possible operators
+       a <type>string</type>, test for a particular relationship. The possible operators
```
kamil-tekiela, line 64:
```suggestion
       An optional operator. The possible operators
```

reference/info/functions/version-compare.xml
```diff
@@ -9,15 +9,10 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>version_compare</methodname>
+   <type class="union"><type>int</type><type>bool</type></type><methodname>version_compare</methodname>
```
kamil-tekiela, line 12: This function does not return `null`. The return value should be adjusted.

reference/network/functions/header-remove.xml
```diff
@@ -46,6 +45,28 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       <parameter>name</parameter> is nullable now.
```
kamil-tekiela, line 62: It would be a good idea to explain what happens when the argument is not provided/null. At the moment only the example explains what happens, but the description doesn't mention that.

kocsismate: Is the new description alright? Or do you have any suggestion?

reference/network/functions/header-remove.xml
```diff
@@ -26,7 +25,7 @@
      <term><parameter>name</parameter></term>
      <listitem>
       <para>
-       The header name to be removed.
+       The header name to be removed. When &null;, all the previously set headers are removed.
```
kamil-tekiela, line 28:
```suggestion
       The header name to be removed. When &null;, all previously set headers are removed.
```

kocsismate: Done :) But I don't understand why the latter is better/more correct. Can you please describe it for me? I'm genuinely interested in it.

kamil-tekiela (approved):
LGTM now. Thank you

reference/network/functions/header-remove.xml
kocsismate: Thanks for the explanation and for the link! :)

PR #1021, opened 2021-10-13
Generate a few ext/standard method synopses based on stubs - part 3

kocsismate:


reference/stream/functions/stream-socket-client.xml
```diff
@@ -9,21 +9,21 @@
...
    Initiates a stream or datagram connection to the destination specified
-   by <parameter>remote_socket</parameter>.  The type of socket created
+   by <parameter>address</parameter>.  The type of socket created
    is determined by the transport specified using standard URL formatting:
    <literal>transport://target</literal>.  For Internet Domain sockets
    (AF_INET) such as TCP and UDP, the <literal>target</literal> portion
-   of the <parameter>remote_socket</parameter> parameter should consist of
+   of the <parameter>address</parameter> parameter should consist of
    a hostname or IP address followed by a colon and a port number.  For Unix
    domain sockets, the <parameter>target</parameter> portion should point
```
saundefined, line 28:
```suggestion
   domain sockets, the <literal>target</literal> portion should point
```
there is no `target` parameter, seems it should be a `literal` tag

saundefined (commented):
Thanks! Looks good to me, except for one old nit.

reference/stream/functions/stream-socket-client.xml
kocsismate: Thanks for the catch!

cmb69:
Thank you!

PR #1024, opened 2021-10-13
[PHP 8.1] Updated mysqli error reporting

kamil-tekiela:


reference/mysqli/mysqli_driver/report-mode.xml
```diff
@@ -86,6 +87,28 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.1.0</entry>
+      <entry>
+       The default value is now <literal>MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT</literal>.
```
saundefined, line 104: Maybe add previous value to changelog too?
```suggestion
       The default value is now <literal>MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT</literal>.
       Previously, it was <constant>MYSQLI_REPORT_OFF</constant>.
```

colshrapnel:
@kamil-tekiela I wanted to propose a minor fix but decided to ask first. Is there any reason to have `__toString()` called manually? In my understanding, there is no use in that, as per the purpose of this magic method.

Also, I feel the urge to make the last example without try-catch, with the explanation like "usually there is no need for try catch as the error must bubble up to the site-wide error handler" but not sure if it will be appropriate.

kamil-tekiela:
@colshrapnel 

> @kamil-tekiela I wanted to propose a minor fix but decided to ask first. Is there any reason to have `__toString()` called manually? In my understanding, there is no use in that, as per the purpose of this magic method.

I have no idea why it is like that. I suppose it must have been like this since the very beginning and I never gave it a second thought why the magic method is called explicitly. Can probably be removed. 


> Also, I feel the urge to make the last example without try-catch, with the explanation like "usually there is no need for try-catch as the error must bubble up to the site-wide error handler" but not sure if it will be appropriate.

Similar situation. http://web.archive.org/web/20200814120300/https://www.php.net/manual/en/mysqli-driver.report-mode.php

It was there before... These examples were very broken. I guess it was meant to show what kind of exception is thrown by enabling mysqli error reporting. That information is mentioned in the table and in the code comment, so I assume we can do without try-catch now. Wrapping whole code in try-catch is a bad practice anyway. 

On the other hand, telling users that there is no reason to catch exceptions should be done in a different place in the manual. This page shouldn't go into that. There are reasons to catch exceptions; just not every time. If one wants to recover from an exception thrown by mysqli, one should be made aware of how to achieve it, no matter how rare that situation is. 

In summary: 
- Definitely remove `__toString()` from these examples. 
- Maybe remove try-catch.
- Don't tell users not to catch exceptions on this page.

PR #1025, opened 2021-10-14
[PHP 8.0] Add new DOM interfaces

saundefined:
Add new `DOMChildNode` and `DOMParentNode` interfaces.

### DIFF
--- a/reference/dom/domchildnode/replaceWith.xml
+++ b/reference/dom/domchildnode/replaceWith.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="domchildnode.replacewith" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>DOMChildNode::replaceWith</refname>
+  <refpurpose>Replace current node with new nodes</refpurpose>
+ </refnamediv>
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <modifier>public</modifier> <type>void</type><methodname>DOMChildNode::replaceWith</methodname>
+   <methodparam rep="repeat"><type class="union"><type>DOMNode</type><type>string</type><type>null</type></type><parameter>nodes</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Replace current node with new <parameter>nodes</parameter>.
+   A combination of <methodname>DOMChildNode::remove</methodname> + <methodname>DOMChildNode::append</methodname>.
+  </para>
+ </refsect1>
+ 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>nodes</parameter></term>
+     <listitem>
+      <para>
+       The replacement nodes.
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+ 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   &return.void;
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><methodname>DOMChildNode::after</methodname></member>
+    <member><methodname>DOMChildNode::before</methodname></member>
+    <member><methodname>DOMChildNode::remove</methodname></member>
+   </simplelist>
+  </para>
+ </refsect1>
+
+</refentry>
+<!-- Keep this comment at the end of the file
+Local variables:
+mode: sgml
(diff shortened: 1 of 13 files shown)


mallardduck (changes requested):
I think updates to `DOMElement` are also necessary. For instance it should be updated so the class reflects that it now `implements DOMParentNode, DOMChildNode`. Going based off changes noted here https://github.com/php/php-src/pull/4709, I've compiled this list of the items needing updates to class synopsis:

- [ ] `DOMElement` needs `implements DOMParentNode, DOMChildNode`
- [ ] `DOMDocument` needs `implements DOMParentNode`
- [ ] `DOMFragment` needs `implements DOMParentNode`
- [ ] `DOMCharacterData` needs `implements DOMChildNode`

saundefined:
@mallardduck thanks!

Yep, as far as I can see, these changes implemented in #839

mallardduck (approved):
Merge along with: https://github.com/php/doc-en/pull/839

reference/dom/domchildnode/replaceWith.xml
```diff
@@ -0,0 +1,74 @@
...
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <modifier>public</modifier> <type>void</type><methodname>DOMChildNode::replaceWith</methodname>
+   <methodparam rep="repeat"><type class="union"><type>DOMNode</type><type>string</type><type>null</type></type><parameter>nodes</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Replace current node with new <parameter>nodes</parameter>.
+   A combination of <methodname>DOMChildNode::remove</methodname> + <methodname>DOMChildNode::append</methodname>.
```
cmb69, line 16:
```suggestion
   Replaces the current node with new <parameter>nodes</parameter>.
   A combination of <methodname>DOMChildNode::remove</methodname> and <methodname>DOMChildNode::append</methodname>.
```

reference/dom/domparentnode/append.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="domparentnode.append" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>DOMParentNode::append</refname>
+  <refpurpose>Appends one or many nodes to the list of children behind the last child node</refpurpose>
```
cmb69, line 5:
```suggestion
  <refpurpose>Appends nodes after the last child node</refpurpose>
```

reference/dom/domparentnode/append.xml
```diff
@@ -0,0 +1,71 @@
...
+ </refnamediv>
+ 
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <modifier>public</modifier> <type>void</type><methodname>DOMParentNode::append</methodname>
+   <methodparam rep="repeat"><type class="union"><type>DOMNode</type><type>string</type><type>null</type></type><parameter>nodes</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Appends one or many <parameter>nodes</parameter> to the list of children behind the last child node.
```
cmb69, line 15:
```suggestion
   Appends one or many <parameter>nodes</parameter> to the list of children after the last child node.
```

reference/dom/domparentnode/prepend.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="domparentnode.prepend" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>DOMParentNode::prepend</refname>
+  <refpurpose>Prepends one or many nodes to the list of children before the first child node</refpurpose>
```
cmb69, line 5:
```suggestion
  <refpurpose>Prepends nodes before the first child node</refpurpose>
```

cmb69 (commented):
Thank you for the PR! It might make sense to add some further methods to the see also sections, e.g. `DOMNode::appendChild` comes to mind. I'm not quite sure about the usage of "current node"; AIUI, this refers to `$this`, so perhaps makes that more explicit.

saundefined:
Sorry for the long implementation =(

@cmb69 Thanks for the review, as always =)

> It might make sense to add some further methods to the see also sections, e.g. `DOMNode::appendChild` comes to mind.

Added new links.

> AIUI, this refers to `$this`, so perhaps makes that more explicit.

What do you think about "object context" instead of "current node"?

cmb69:
> What do you think about "object context" instead of "current node"?

Maybe, not sure. `DOMNode::hasChildNodes()` just states "Checks if node has children", so maybe we should just use "the node" instead of "current node" here? While that is not particularly specific, it might be good enough.

cmb69 (approved):
Thank you!

PR #1028, opened 2021-10-15
Add a very basic match example

BackEndTea:
I was looking up the docs for match, and missed a really basic example, to see the syntax.

### DIFF
--- a/language/control-structures/match.xml
+++ b/language/control-structures/match.xml
@@ -34,15 +34,24 @@ $return_value = match (subject_expression) {
     <programlisting role="php">
 <![CDATA[
 <?php
+$food = 'cake';
+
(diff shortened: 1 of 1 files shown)


language/control-structures/match.xml
```diff
@@ -28,6 +28,22 @@ $return_value = match (subject_expression) {
 ?>
 ]]>
   </programlisting>
+  
+   <example>
+   <title>Basic<literal>match</literal> usage</title>
+    programlisting role="php">
+<![CDATA[
+<?php
+$return_value = match ($food) {
```
kamil-tekiela, line 37: It would be good idea to define `$food` and explain which arm is executed.

PR #1030, opened 2021-10-15
Opcache functions only apply to in-memory cache

frederikbosch:
The functions `opcache_get_status`, `opcache_invalidate`, `opcache_is_script_cached` and `opcache_reset` only apply to in-memory cache, not file cache. Also [fixes bug #75070](https://bugs.php.net/bug.php?id=75070).


cmb69:
Maybe leave the refpurposes alone, but only change the descriptions? Maybe mention file cache there as well?

PR #1033, opened 2021-10-19
[PHP 8.1] Add pcntl_rfork description

saundefined:
Based on [stubs](https://github.com/php/php-src/blob/PHP-8.1/ext/pcntl/pcntl.stub.php#L84) and [`rfork()`](https://www.freebsd.org/cgi/man.cgi?query=rfork&sektion=2&manpath=FreeBSD+4.10-RELEASE) description.

### DIFF
--- a/reference/pcntl/functions/pcntl-rfork.xml
+++ b/reference/pcntl/functions/pcntl-rfork.xml
@@ -49,7 +49,7 @@
         </member>
         <member>
          <constant>RFLINUXTHPN</constant>: If set, the kernel will return SIGUSR1 instead of SIGCHILD upon thread exit for the child.
-         This is intended to mimic certain Linux clone behaviour.
+         This is intended to do Linux clone exit parent notification.
         </member>
        </simplelist>
       </para>
@@ -94,15 +94,15 @@ if ($pid > 0) {
 } else {
   // This is the child process.
   var_dump($pid);
-  sleep(2); // as the child does not wait so we see its "pid"
+  sleep(2); // as the child does not wait, so we see its "pid"
 }
 ?>
 ]]>
     </programlisting>
-    &example.outputs;
+    &example.outputs.similar;
     <screen>
 <![CDATA[
-int(%d)
+int(77093)
 int(0)
 ]]>
     </screen>


reference/pcntl/functions/pcntl-rfork.xml
```diff
@@ -0,0 +1,143 @@
...
+         <constant>RFFDG</constant>: If set, the invoker's file descriptor table is copied;
+         otherwise the two processes share a single table.
+        </member>
+        <member>
+         <constant>RFCFDG</constant>: If set, the new process starts with a clean file descriptor table.
+         Is mutually exclusive with <literal>RFFDG</literal>.
+        </member>
+        <member>
+         <constant>RFLINUXTHPN</constant>: If set, the kernel will return SIGUSR1 instead of SIGCHILD upon thread exit for the child.
+         This is intended to mimic certain Linux clone behaviour.
```
Crell, line 52: Is there a link to more details on this we can provide?  "certain Linux clone behavior" tells me nothing unless I'm a kernel dev, and maybe not even then. :smile:

reference/pcntl/functions/pcntl-rfork.xml
```diff
@@ -0,0 +1,143 @@
...
+  <para>
+   <example>
+    <title><function>pcntl_rfork</function> example</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+$pid = pcntl_rfork(RFNOWAIT|RFTSIGZMB, SIGUSR1);
+if ($pid > 0) {
+  var_dump($pid);
```
Crell, line 92:
```suggestion
  // This is the parent process.
  var_dump($pid);
```

reference/pcntl/functions/pcntl-rfork.xml
```diff
@@ -0,0 +1,143 @@
...
+} else {
+  var_dump($pid);
```
Crell, line 94:
```suggestion
  // This is the child process.
  var_dump($pid);
```

reference/pcntl/functions/pcntl-rfork.xml
```diff
@@ -0,0 +1,143 @@
...
+  sleep(2); // as the child does not wait so we see its "pid"
+}
+?>
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+int(%d)
```
Crell, line 103: It doesn't output `int(%d)`.  It outputs int(some number)`.  Go ahead and invent a number to put there, but also include a comment to that effect.

Crell: Actually, I think the `&example.outputs.similar;` macro with some random number would be sufficient.

reference/pcntl/functions/pcntl-rfork.xml
saundefined: Thanks for the review!

Updated this string, based on description in [freebsd](https://github.com/freebsd/freebsd-src/blob/main/sys/sys/unistd.h#L183).

PR #1036, opened 2021-10-20
Make password_needs_rehash() more prominent

cmb69:
Successful validation should usually be followed by re-hashing the
password, if necessary.  The `password_needs_rehash()` docs already
have a good example, but the `password_verify()` page does not mention
that at all.  Instead of duplicating the example, we refer to it.

We also remove the link to the password compatibility library, since it
is no longer relevant for the manual.


krakjoe:
erm, can we not have one page with example usage and explanations and use "see" on the individual function pages ?

one source of truth is better imo ...

Crell:
I agree.  One page with a complete example, plus lots of links pointing people there.

kamil-tekiela (approved):
I think this can be merged as is. If someone wants to create a common example, it can go into a new PR.

cmb69:
Thank you!

PR #1055, opened 2021-10-29
Fix mysqli return values

kamil-tekiela:


saundefined (approved):
Thanks, looks good to me.

PR #1056, opened 2021-10-29
Add mention on question mark escape feature

julienfastre:
Since PHP 7.4, it is possible to escape question mark in php pdo statement by doubling them: https://wiki.php.net/rfc/pdo_escape_placeholders

I add a simple statement and add an example.

I am sorry the example is only valid using Postgresql database. But some question on multiple places (like [on stackoverflow](https://stackoverflow.com/q/36173440/1572236)) are related to this database (and I use often this RDBMS, I do not know if this is in use elsewhere).

### DIFF
--- a/reference/pdo/pdo/prepare.xml
+++ b/reference/pdo/pdo/prepare.xml
@@ -162,13 +162,12 @@ $yellow = $sth->fetchAll();
     <programlisting role="php">
 <![CDATA[
 <?php
-/* Execute a prepared statement which contains a question mark as sql statement */
 /* note: this is only valid on PostgreSQL databases */
-$sth = $dbh->prepare('SELECT '["a", "b", "c"]'::jsonb ?? ?);
-$sth->execute(array('a'));
-$resultTrue = $sth->fetchAll();
-$sth->execute(array('z'));
-$resultFalse = $sth->fetchAll();
+$sth = $dbh->prepare('SELECT * FROM issue WHERE tag::jsonb ?? ?);
+$sth->execute(array('feature'));
+$featuresIssues = $sth->fetchAll();
+$sth->execute(array('performance'));
+$performanceIssues = $sth->fetchAll();
 ?>
 ]]>
     </programlisting>
--- a/reference/pdo/pdo/prepare.xml
+++ b/reference/pdo/pdo/prepare.xml
@@ -66,6 +66,9 @@
     which is natively supported by the driver.
    </simpara>
   </note>
+  <para>
+    Questions marks can be escaped by doubling them. That means that the “??” string will be translated to “?” when sending the query to the database.
+  </para>
  </refsect1>
  <refsect1 role="parameters">
   &reftitle.parameters;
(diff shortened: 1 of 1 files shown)


reference/pdo/pdo/prepare.xml
```diff
@@ -66,6 +66,9 @@
     which is natively supported by the driver.
    </simpara>
   </note>
+  <para>
+    Questions marks can be escaped by doubling them. That means that the “??” string will be translated to “?” when sending the query to the database.
```
kamil-tekiela, line 70:
```suggestion
    Question marks can be escaped by doubling them. That means that
    the <literal>??</literal> string will be translated to <literal>?</literal>
    when sending the query to the database.
```

reference/pdo/pdo/prepare.xml
```diff
@@ -149,6 +152,22 @@ $red = $sth->fetchAll();
...
 ?>
+]]>
+    </programlisting>
+   </example>
+   <example>
+    <title>SQL statement template with question mark escaped</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+/* Execute a prepared statement which contains a question mark as sql statement */
```
kamil-tekiela, line 163:
```suggestion
/* Execute a prepared statement that contains a question mark */
```
Shouldn't that be the example description? Why is this comment necessary?

reference/pdo/pdo/prepare.xml
```diff
@@ -149,6 +152,22 @@ $red = $sth->fetchAll();
...
+/* note: this is only valid on PostgreSQL databases */
+$sth = $dbh->prepare('SELECT '["a", "b", "c"]'::jsonb ?? ?);
+$sth->execute(array('a'));
+$resultTrue = $sth->fetchAll();
```
kamil-tekiela, line 167: The result is not used in the example. Why fetch the data at all?

reference/pdo/pdo/prepare.xml
julienfastre: You are right, as we have a title, this is not necessary.

reference/pdo/pdo/prepare.xml
julienfastre: I tryed to improve my example.

Thanks for your guidance.

cmb69:
Thank you for the PR! I've applied some minor tweaks on top:
````.diff
 reference/pdo/pdo/prepare.xml | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/reference/pdo/pdo/prepare.xml b/reference/pdo/pdo/prepare.xml
index a23d05f822..ad1aa42d5c 100644
--- a/reference/pdo/pdo/prepare.xml
+++ b/reference/pdo/pdo/prepare.xml
@@ -67,7 +67,7 @@
    </simpara>
   </note>
   <para>
-    Question marks can be escaped by doubling them. That means that
+    As of PHP 7.4.0, question marks can be escaped by doubling them. That means that
     the <literal>??</literal> string will be translated to <literal>?</literal>
     when sending the query to the database.
   </para>
@@ -163,10 +163,10 @@ $yellow = $sth->fetchAll();
 <![CDATA[
 <?php
 /* note: this is only valid on PostgreSQL databases */
-$sth = $dbh->prepare('SELECT * FROM issues WHERE tag::jsonb ?? ?);
-$sth->execute(array('feature'));
+$sth = $dbh->prepare('SELECT * FROM issues WHERE tag::jsonb ?? ?');
+$sth->execute(['feature']);
 $featureIssues = $sth->fetchAll();
-$sth->execute(array('performance'));
+$sth->execute(['performance']);
 $performanceIssues = $sth->fetchAll();
 ?>
 ]]>
````

PR #1058, opened 2021-11-01
fix bad language

sfinktah:
"This iterator allows to unset and modify values and keys while iterating over Arrays and Objects."

**sigh**

### DIFF
--- a/reference/spl/arrayiterator.xml
+++ b/reference/spl/arrayiterator.xml
@@ -10,8 +10,8 @@
   <section xml:id="arrayiterator.intro">
    &reftitle.intro;
    <para>
-    This iterator allows to unset and modify values and keys while iterating
-    over Arrays and Objects.
+     Allows the removal of elements, and the modification of 
+     keys or values while iterating over Arrays or Objects.
    </para>
    <para>
     When you want to iterate over the same array multiple times you need to


Gomesz785 (approved):
LGTM

andrewbrereton (approved):

cmb69:
Thank you for the PR! This seems only like a small improvement; after all, iterating an array allows to unset and [modify elements as well](https://3v4l.org/RCRNl). Furthermore, [nested iteratons over an array is possible as well](https://3v4l.org/J6FvK), so I'm missing the actual purpose of that class in its description.

Maybe you want to improve that description even more?

sfinktah:
I would not assume to be that familiar with the full capability of the function, otherwise I would not have been looking at the documentation in the first place.  :)

tiffany-taylor:
Possibly this:
> This iterator allows unsetting and modifying values and keys while iterating over Arrays and Objects.

Though I am conflicted on whether or not "Arrays" and "Objects" should be capitalized. It makes sense for this particular page, but it lacks consistency with the rest of the manual.

cmb69:
Might just use `&array;s` and `&object;s` then.

sfinktah:
> Possibly this:
> 
> > This iterator allows unsetting and modifying values and keys while iterating over Arrays and Objects.
> 
> Though I am conflicted on whether or not "Arrays" and "Objects" should be capitalized. It makes sense for this particular page, but it lacks consistency with the rest of the manual.

That's the same terrible language the original suffered from.  My quick revision was:

> Allows the removal of elements, and the modification of keys or values while iterating over Arrays or Objects.

With regards to "modify elements elements as well", the verbiage for that can be lifted from `foreach`.  

Unfortunately some actual testing of this `ArrayObject` and `ArrayIterator` have shown that it **does not** allow unsetting keys while iterating, so the entire description is wrong.  The same goes for https://www.php.net/manual/en/arrayiterator.offsetunset.php 

Testing shows https://3v4l.org/T2sXU#v7.4.27 that the following (which is as close as I can get to C++'s approved method of iterator based erasure) results in exactly the same behaviour as attempting deletion during a `foreach` loop.

I think the pythonic solution for that particular problem (`foreach copy($a) as $k => $v { unset($a[$k]; }`)would be more useful than the current documentation. 

```php
// Attempt to delete while iterating, comments in C++
$a = range(0, 10);
$obj = new ArrayObject( $a );
$it = $obj->getIterator();

// for (auto it = a.begin(); it != a.end(); ) 
for ( ; $it->valid() ; ) {
    // if (it->second & 2)
    if ($it->current() & 2)
        // it = it.erase()
        $it->offsetUnset($it->key());
    // else
    else
        // it = it.next()
        $it->next();
}
print_r($obj);
// expected result: [0, 1, 3, 5, 7, 9]
// actual result: [0, 1, 4, 5, 8, 9]
```

afilina:
@cmb69 @tiffany-taylor Bumping for a review.

reference/spl/arrayiterator.xml
```diff
@@ -10,8 +10,8 @@
   <section xml:id="arrayiterator.intro">
    &reftitle.intro;
    <para>
-    This iterator allows to unset and modify values and keys while iterating
-    over Arrays and Objects.
+     Allows the removal of elements, and the modification of 
+     keys or values while iterating over Arrays or Objects.
```
Girgias, line 14: Nit: indent
```suggestion
    Allows the removal of elements, and the modification of 
    keys or values while iterating over Arrays or Objects.
```

sfinktah:
Note: my previous example of deletion while iterating not working was erroneous, this example works as expected (and should probably be included in some neater form in the actual documentation).

```php
// Attempt to delete while iterating, comments in C++
$a = range(0, 10);
$obj = new ArrayObject( $a );
$it = $obj->getIterator();

// for (auto it = a.begin(); it != a.end(); )
for ( ; $it->valid() ; ) {
    // if (it->second % 2 == 0)
    printf("it->key: %s  it->current: %s\n", $it->key(), $it->current());
    # remove even numbers
    if ($it->current() % 2 == 0) {
        // it = it.erase()
        printf("unsetting value: %s\n", $it->current());
        $it->offsetUnset($it->key());
    }
    else
        // it = it.next()
        $it->next();
}
print_r($obj);
// expected result: [1, 3, 5, 7, 9]
// actual result: [1, 3, 5, 7, 9]
```

Girgias:
Could you open a new PR/issue for this? Unlikely this is going to be seen from a merged PR

sfinktah:
@Girgias it's not important, it's mainly just to say I was wrong about something being broken.

PR #1060, opened 2021-11-03
Update ISO 8601 text

betweenbrain:
Use _ISO 8601_, instead of _ISO-8601_, for consistency and alignment with https://www.iso.org/iso-8601-date-and-time-format.html


cmb69:
Thank you!

betweenbrain:
You're welcome!

PR #1062, opened 2021-11-04
Add note when return value is string

hakre:
Note taken from reference/mysqli/mysqli_result/num-rows.xml

### DIFF
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -1869,6 +1869,9 @@ linkend="book.mysqlnd">mysqlnd</link>.'>
 hand-shake/authentication, which mysqlnd will use.</para><para>Libmysqlclient uses the default charset set in the
 <filename>my.cnf</filename> or by an explicit call to <function>mysqli_options</function> prior to
 calling <function>mysqli_real_connect</function>, but after <function>mysqli_init</function>.</para></note>'>
+<!ENTITY mysqli.integer.overflow.as.string.note '<note xmlns="http://docbook.org/ns/docbook">
+<para>If the number of rows is greater than <constant>PHP_INT_MAX</constant>,
+the number will be returned as a string.</para></note>'>
 <!ENTITY mysqli.sqlinjection.warning '<warning xmlns="http://docbook.org/ns/docbook">
 <title>Security warning: SQL injection</title><para>If the query contains any variable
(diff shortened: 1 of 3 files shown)


hakre:
Just a quick fix while stumbling over it in room11. Bear with me, second PR ;)

kamil-tekiela:
Good catch. Yes, the note should be there. And it belongs in this section. However, I would prefer that you add a snippet instead. I am not a fan of duplicating the same content because this will easily get misaligned after some time.

hakre:
@kamil-tekiela: if you can bring me up-speed about snippets, I can do that. some kind of reference or quick hint should suffice. me guesses its an XML entity and can be defined somewhere.

kamil-tekiela:
grep for `mysqli.charset.note` in the EN documentation. You will see an entity defined in language-snippets.ent and then used in a couple places as `&mysqli.charset.note;`

hakre:
found all this, will give it traction now.

doc-building questions:

* are constants automatically referenced?
* add refsect1 similar to:
   ~~~
    <refsect1 role="notes">
     &reftitle.notes;
     &mysqli.charset.note;
    </refsect1>
   ~~~
   ?

further on:

* [?] explain that the relevance is most likely for 32bit systems ?

additionally:

* [?] sort entities us-ascii (caseless)  in `language-snippets.ent` ?

kamil-tekiela:
`integer` not `ingteger`.

kamil-tekiela (approved):
This looks good to me. 

Constants are not automatically referenced. 
I don't think the "notes" section makes sense here. The note is relevant for the return value so it is in the right place. 
Don't mention 32bit systems. It's not necessary and also obvious. You are unlikely to ever have more rows than 64bit max integer. 
The entities aren't sorted. They are grouped, but not sorted. 
I don't think the whitespace causes any issues in the manual. Anyway, that's a separate topic altogether.

hakre:
@kamil-tekiela the reason I mentioned 32bit here is that from memory why this was originally reported in room11 was that it was unclear why the string. it's my understanding that the string is only when integer size is not enough. which is more likely on 32bit than on 64bit.

all these extra notes were informational. fine for separate notes, filed a whitespace check as Github CI workflow in https://github.com/php/doc-en/pull/1093 .

thanks!

PR #1065, opened 2021-11-08
Generate ext/zlib class synopses from stubs

kocsismate:


saundefined (approved):
Thanks! Looks good.
Also it seems that `[skip-revcheck]` can be applied

PR #1068, opened 2021-11-08
Generate ext/xmlreader and ext/xmlwriter class synopses from stubs

kocsismate:


reference/xmlreader/xmlreader.xml
```diff
@@ -170,97 +167,83 @@
       <initializer>4</initializer>
      </fieldsynopsis>
 <!-- }}} -->
+
     <classsynopsisinfo role="comment">&Properties;</classsynopsisinfo>
-<!-- {{{ XMLReader properties -->
-     <fieldsynopsis>
-      <modifier>public</modifier>
-      <modifier>readonly</modifier>
```
mumumu, line 177: Note: `[skip-revcheck]` should not be applied to this PR, because `readonly` modified is deleted and affects manual output.

cmb69:
Thank you!

PR #1074, opened 2021-11-08
Generate ext/openssl class synopses from stubs

kocsismate:


cmb69:
This looks like a whitespace change and comment removal only patch. Is it to make it easier to have follow up patches regarding class synopses? If so, should this be tagged as `[skip-revcheck]`?

kocsismate:
> Is it to make it easier to have follow up patches regarding class synopses?

Yes, that's the case!

> If so, should this be tagged as [skip-revcheck]?

Isn't skip-revcheck only relevant when an English documentation related change (typo fix, grammar improvement etc) is committed to doc-en? The changes in this PR are universal, albeit doesn't relate to any language.. Do you think that skip-revcheck should apply here? If so, then I'm happy to add it.

cmb69:
> Do you think that skip-revcheck should apply here?

I don't know. Maybe some translators could state their opinion? @saundefined, @Girgias, @mumumu, what do you think?

saundefined:
It is in this PR, I think, no difference since there are few modified files.

_In my opinion_, for such changes, the tag does not need to be applied, and leave it to correct grammar/typos/nits in the English version.

Girgias:
As this doesn't change anything in which translators need to act upon, IMHO this should be marked with `[skip-revcheck]` as this is generated by a script and are mostly cosmetic changes it makes more sense to me to skip the revcheck and just run the script on the translations if we want to normalise the markup code style across versions.

mumumu:
IMO, in this case, skip-revcheck can be applied, because this PR changes does not affect manual contents(appearance) and translation output.

But it's no problem even if [skip-revcheck] is not applied.

kocsismate:
So 2:1 for adding `[skip-revcheck]`?

Issue #876, opened 2021-08-25
Undocumented effect of reseting inner pointer of the array to 0 index in group of std methods


Salomon88:
In practice were found a group of functions that after their invocation reset the array pointer to 0 position. Unfortunately, this effect isn't documented and can be unexpected. The example below is a little bit synthetic, but it can be more complex cases:
```
<?php

$arr = array(1, 7, 2);
echo current($arr);
next($arr); //Moves to 1 index
sort($arr); //Rewinds to 0 index
echo current($arr);
```
List of functions to be documented:
rsort, shuffle, uasort, uksort, sort, krsort, ksort, natcasesort, natsort, array_multisort, arsort, asort, array_unshift

Salomon88:
Example - https://3v4l.org/beLfO

cmb69:
Closes via 6a6f43d1c490a57b452656db285de6d136055ed2.

PR #1077, opened 2021-11-09
Documentation of index reset effect of some functions - issue #876

Salomon88:


cmb69:
Thank you! I suggest to move the definition of the `note.reset-index` entity to just below the `note.sort-unstable` entity definition.

Salomon88:
> Thank you! I suggest to move the definition of the `note.reset-index` entity to just below the `note.sort-unstable` entity definition.

Hello! Unfortunately, not all functions have link on `note.sort-unstable` , for example `array-unshift.xml`

cmb69:
I didn't mean to merge both entitiy definition, but rather to move the new definition upwards, so that both are close to each other.

Salomon88:
> I didn't mean to merge both entitiy definition, but rather to move the new definition upwards, so that both are close to each other.

Done

cmb69:
Thank you!

PR #1078, opened 2021-11-09
php memory stream can't be passed to library on Windows

Danack:
apparently.

e.g. fix needed for Imagick on windows: https://github.com/Imagick/imagick/pull/493

### DIFF
--- a/language/wrappers/php.xml
+++ b/language/wrappers/php.xml
@@ -71,12 +71,17 @@
    <simpara>
     <filename>php://memory</filename> and <filename>php://temp</filename> are
     read-write streams that allow temporary data to be stored in a file-like
-    wrapper. The only difference between the two is that
+    wrapper. One difference between the two is that
     <filename>php://memory</filename> will always store its data in memory,
     whereas <filename>php://temp</filename> will use a temporary file once the
     amount of data stored hits a predefined limit (the default is 2 MB). The
     location of this temporary file is determined in the same way as the
     <function>sys_get_temp_dir</function> function.
+   </simpara>
+   <simpara>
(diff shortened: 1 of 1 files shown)


cmb69:
Hmm, would it be possible to attempt such casts from userland? Otherwise I think this should be documented in some internal docs.

Danack:
tbh, I don't really understand the underlying problem (and am fine not knowing) but the note should probably be on the user facing docs, as it seems like the type of thing that a windows user might encounter if an extension is maintained by someone who mostly uses Linux.

cmb69:
> […] as it seems like the type of thing that a windows user might encounter if an extension is maintained by someone who mostly uses Linux.

Especially in this case, wouldn't it be good if the Windows user reported this issue to the maintainers, so they could fix the code (maybe by just reporting the problem explicitily)?

And generally, how would a Windows user even know where such cast to stdio might happen?

And of course, it would be nice if casting memory streams to stdio stream would also be supported on Windows. It *might* be possible to use a memory mapped temporary file for that, but I presume it's not generally possible. Anyway, that's a different story.

afilina:
@cmb69 Is the text in this PR technically correct? Although these things can be documented internally, it may not hurt to have a mention here. Other people might end up on this page in search of information.

cmb69:
Well, I'm still somewhat concerned about the `fopencookie()` mention. At the very least, we should clarify that this is a C function (not a PHP function). Not sure if readers would understand `fopencookie(3)`, or whether we should say `the C function fopencookie()`.

afilina:
@Danack It looks like this is waiting on the author. Gently nudging.

PR #1079, opened 2021-11-09
Add options parameters from comment

shaedrich:
As per https://www.php.net/manual/en/mysqli.debug.php#122219

### DIFF
--- a/reference/mysqli/mysqli/debug.xml
+++ b/reference/mysqli/mysqli/debug.xml
@@ -87,6 +87,7 @@
            <entry>MYSQLND_DEBUG_PROFILE_CALLS</entry>
           </row>
          </tbody>
+        </tgroup>
         </table>
        </para>
      </listitem>
--- a/reference/mysqli/mysqli/debug.xml
+++ b/reference/mysqli/mysqli/debug.xml
@@ -35,6 +35,12 @@
        A string representing the debugging operation to perform
       </para>
       <para>
+        The debug control string is a sequence of colon separated fields as follows:
+        <literal><![CDATA[<field_1>:<field_2>:<field_N>]]></literal>
+        Each field consists of a mandatory flag character followed by an optional "," and comma separated list of modifiers:
+        <literal>flag[,modifier,modifier,...,modifier]</literal>
+      </para>
+      <para>
        <table>
         <title>Recognized Flag Characters</title>
           <tgroup cols="3">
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli/debug.xml
```diff
@@ -34,6 +34,62 @@
       <para>
        A string representing the debugging operation to perform
       </para>
+      <para>
+       <table>
+        <title>The following characters are recognized in the
+        <parameter>options</parameter> parameter string</title>
```
cmb69, line 40: This title appears to be too verbose. Maybe just:
```suggestion
        <title>Recognized Flag Characters</title>
```

cmb69 (commented):
Thank you! Apparently, [some flags are missing](https://github.com/php/php-src/blob/php-7.4.25/ext/mysqlnd/mysqlnd_debug.c#L494-L688), but it's better than having no information here at all. And it might be worthwhile to also explain the [format generally](http://tiebing.blogspot.com/2011/10/cc-dbug-library.html), but again, better something than nothing.

shaedrich:
@cmb69 You are welcome! I added an explanation and a missing flag – as you say, it's still not every information, but we already got closer to the goal

reference/mysqli/mysqli/debug.xml
```diff
@@ -34,6 +34,71 @@
       <para>
        A string representing the debugging operation to perform
       </para>
+      <para>
+        The debug control string is a sequence of colon separated fields as follows:
+        <literal><![CDATA[<field_1>:<field_2>:<field_N>]]></literal>
+        Each field consists of a mandatory flag character followed by an optional "," and comma separated list of modifiers:
```
cmb69, line 40:
```suggestion
        Each field consists of a mandatory flag character followed by an optional <literal>,</literal> and comma separated list of modifiers:
```

reference/mysqli/mysqli/debug.xml
```diff
@@ -34,6 +34,71 @@
...
+        <literal>flag[,modifier,modifier,...,modifier]</literal>
+      </para>
+      <para>
+       <table>
+        <title>Recognized Flag Characters</title>
+          <tgroup cols="3">
```
cmb69, line 46: Shouldn't this be:
```suggestion
          <tgroup cols="2">
```

cmb69 (commented):
Thank you! Looks good to me, but please check-out my comments.

reference/mysqli/mysqli/debug.xml
shaedrich: You're right :+1: I copied and pasted it and changed everything but this to two cols 😅

shaedrich:
You're welcome!

PR #1087, opened 2021-11-12
Match return typehint with documentation

esserj:
When an error occurs, null can be returned

### DIFF
--- a/reference/array/functions/array-replace-recursive.xml
+++ b/reference/array/functions/array-replace-recursive.xml
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>array</type><methodname>array_replace_recursive</methodname>
+   <type>?array</type><methodname>array_replace_recursive</methodname>
    <methodparam><type>array</type><parameter>array</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you for the PR! However, that `null` return is actually [undefined behavior](https://www.php.net/manual/en/functions.internal.php). Relevant are [the stubs](https://github.com/php/php-src/blob/9cd7f41fac87a83766d54d73dd08ccb5ee11176e/ext/standard/basic_functions.stub.php#L187), from which documentation is updated from time to time.

nikic:
> Thank you for the PR! However, that `null` return is actually [undefined behavior](https://www.php.net/manual/en/functions.internal.php). Relevant are [the stubs](https://github.com/php/php-src/blob/9cd7f41fac87a83766d54d73dd08ccb5ee11176e/ext/standard/basic_functions.stub.php#L187), from which documentation is updated from time to time.

Might make sense to clarify that this only applies to PHP < 8.0, and since PHP 8.0 the specified signatures are accurate.

cmb69:
See https://github.com/php/doc-en/commit/14e006b4eea529ce9c3df9e550b27fb507aaa44c.

PR #1089, opened 2021-11-12
strtotime - 'current time' = 'current time zone'

sfinktah:
The function expects to be given a string containing an English date format
   and will try to parse that format into a Unix timestamp (the number of
   seconds since January 1 1970 00:00:00 UTC), relative to the timestamp given
   in <parameter>baseTimestamp</parameter>, or the current time **zone** if
   <parameter>baseTimestamp</parameter> is not supplied.


sfinktah:
My bad, the documentation is extremely confusing, and I mis-interpreted it.

PR #1093, opened 2021-11-13
Add white space checker script to CI

hakre:
Original work is in [git] github mirror.

With ~~local~~ upstream fixes of the comment on top ~~, stale since Jul 2021 changes
upstream~~.

[git]: https://github.com/git/git/blob/master/.github/workflows/check-whitespace.yml

### DIFF
--- a/.github/workflows/check-whitespace.yml
+++ b/.github/workflows/check-whitespace.yml
@@ -1,0 +1,49 @@
+name: check-whitespace
+
+# Get the repository with all commits to ensure that we can analyze
+# all of the commits contributed via the Pull Request.
+# Process `git log --check` output to extract just the check errors.
+# Exit with failure upon white-space issues.
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  check-whitespace:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: git log --check
+        id: check_out
+        run: |
+          log=
+          commit=
+          while read dash etc
+          do
+            case "${dash}" in
+            "---")
+              commit="${etc}"
+              ;;
+            "")
+              ;;
+            *)
+              if test -n "${commit}"
+              then
+                log="${log}\n${commit}"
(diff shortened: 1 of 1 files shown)


cmb69:
Can you please explain what this script does? Is it checking the log messages? What kinds of errors would be reported?

hakre:
> Can you please explain what this script does?

It runs `git-log(1)` with the `--check` option:

> Warn if changes introduce conflict markers or whitespace errors. What are considered whitespace errors is controlled by core.whitespace configuration. By default, trailing whitespaces (including lines that consist solely of whitespaces) and a space character that is immediately followed by a tab character inside the initial indent of the line are considered whitespace errors. Exits with non-zero status if problems are found. Not compatible with `--exit-code`.

> Is it checking the log messages?

Yes, those from `--check`, see above.

> What kinds of errors would be reported?

See as well above, for whitespace those are controlled by gits' _default_ whitespace configuration as there is no other configuration, which are three out of six whitespace checks git ships with and matches the description above.

In this setup it checks every changed line of each commit within the PR. In case of error it requires to re-author each individual line in the existing commits until every commit in the PR is error-free.

This should effectively prevent contributing merge conflict markers and some rudimentary whitespace issues like missing newlines at the end of file or introducing space characters in indents that are immediately followed by a tab character and superfluous trailing whitespace (might require Markdown line-breaks to be protected with a trailing `\` before newline, otherwise should not be any issue).

As the repositories XML files ship with editor configurations (vi etc.) while files contain whitespace errors (e.g. trailing spaces), preparing patches is cumbersome as the editor already corrects these errors, but the changes or interest are not on all those lines.

This check will prevent to keep such errors at least for any new changes per pull requests.

Next to whitespace errors, conflict markers are also part of the existing projects' history so I'd thought to suggest this bae-level check that ships with git for common mistakes we make with git.

Does this answer help you @cmb69?

cmb69:
Thanks, @hakre. I think it is a good idea to have this check.

Crell:
I like this as well.  What else is needed here?  Can we just hit merge and be happy?

cmb69:
Thank you!

PR #1094, opened 2021-11-13
PDOStatement colons for names in array example

hakre:
passing values for named parameters in an array allow array keys being
either the names of the named parameters or (alternatively) the names
prefixed with colons (\<colon> : \<U003A> COLON) mimicking the placeholders
in the SQL query.

therefore the example should show first verbatim names as the array keys
and then highlight the alternative way to have the array keys with a
dedicated comment within the examples code.

### DIFF
--- a/reference/pdo/pdo/prepare.xml
+++ b/reference/pdo/pdo/prepare.xml
@@ -127,8 +127,9 @@ $sql = 'SELECT name, colour, calories
     FROM fruit
     WHERE calories < :calories AND colour = :colour';
 $sth = $dbh->prepare($sql, array(PDO::ATTR_CURSOR => PDO::CURSOR_FWDONLY));
-$sth->execute(array(':calories' => 150, ':colour' => 'red'));
+$sth->execute(array('calories' => 150, 'colour' => 'red'));
 $red = $sth->fetchAll();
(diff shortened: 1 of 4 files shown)


hakre:
similar edit(s) for `PDO:execute()` examples as well, the rationale is identical, we may want to have this on a single commit:

---

reference(pdo): colons for names execute() examples

passing values for named parameters in an array allow array keys being
either the names of the named parameters or (alternatively) the names
prefixed with colons (<colon> : <U003A> COLON) mimicking the placeholders
in the SQL query.

therefore the examples should show first verbatim names as the array keys
and then highlight the alternative way to have the array keys with a
dedicated comment within the examples' code.

additionally removed superfluous curly brackets in the first examples
`$colour` string interpolation.

finally added a space (\<space> " " \<U0020> SPACE) after the comma
(\<comma> "," \<U002C> COMMA) to separate each placeholder from the other
visually in the generated SQL (IN(?, ?, ...) example).

kamil-tekiela (approved):
It looks ok. While "identifiers" might have been a better choice than "names" I think it's understandable as it is.

kamil-tekiela:
Thank you.

PR #1098, opened 2021-11-15
Clarify purpose of ReflectionType

jdreesen:
When calling e.g. `ReflectionParameter::getType()` you get an instance of `ReflectionType`, too, see: https://www.php.net/manual/en/reflectionparameter.gettype.php


cmb69:
Thank you! You're right, but ReflectionProperty::getType() also returns a ReflectionType. Not sure how to word that.

jdreesen:
Oh, you're right. I updated the PR. Let me know if it fits for you, or if you want another wording.

cmb69:
Well, that's okay. Thank you!

PR #1101, opened 2021-11-16
Remove incorrect return comment for ArrayObject::unserialize

DmitryTronin:
The [ArrayObject::unserialize](https://www.php.net/manual/en/arrayobject.unserialize.php) has `void` return type so it looks a bit incorrect describing it as returning an "unserialized ArrayObject"


cmb69 (changes requested):
Good catch, thank you! However, all functions/methods should have a return value section, even if they "return" void.

cmb69:
Thank you!

PR #1103, opened 2021-11-17
Generate ext/sockets class synopses from stubs

kocsismate:
skip-revcheck can be applied


saundefined (approved):
LGTM, thanks!

>skip-revcheck can be applied

yep

PR #1105, opened 2021-11-17
[skip-revcheck] Generate ext/oci8 class synopses from stubs

kocsismate:


reference/oci8/OCI-Collection.xml
```diff
@@ -26,18 +26,20 @@
...
      <ooclass>
       <classname>OCICollection</classname>
      </ooclass>
     </classsynopsisinfo>
-<!-- }}} -->
- 
+
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.OCI-Collection')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[@role='oop'])" />
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.OCI-Collection')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[not(@role='procedural')])">
```
kocsismate, line 40: Is it possible to change a class ID without messing up the link? After the rename of OCI-Collection and OCI-Lob, the correct ID is `class.ocicollection`, so I had to manually fix up this for now.

cmb69: Regarding the Website, we could add error redirections, but that might get messy. Alternatively, we could duplicate these classes and their methods in the docs, just stating that the classes have been renamed. Not sure.

kocsismate: I think we could stay with the current IDs for now, merge the PR (if it's OK) so that the page is 99% compatible with gen_stub's output, and we can fix the rest of the issues later :)

kocsismate: (However, skip-revcheck shouldn't be applied as fallbacks are added)

cmb69: Okay, makes sense.

PR #1106, opened 2021-11-17
[skip-revcheck] Generate ext/shmop class synopses from stubs

kocsismate:


saundefined (approved):
Looks good, thank you!

PR #1107, opened 2021-11-17
Generate ext/session class synopses from stubs

kocsismate:


reference/session/sessionupdatetimestamphandlerinterface.xml
```diff
@@ -29,18 +28,20 @@
...
+     <classname>SessionUpdateTimestampHandlerInterface</classname>
+    </ooclass>
 
-<!-- {{{ Class synopsis -->
     <classsynopsisinfo>
-     <ooclass>
-      <classname>SessionUpdateTimestampHandlerInterface</classname>
-     </ooclass>
+     <oointerface>
+      <interfacename>SessionUpdateTimestampHandlerInterface</interfacename>
```
kocsismate, line 37: Actually, I carved out this change into a 2nd commit so that all but this change is skip-revchecked

cmb69: Thanks for pointing this out!

PR #1113, opened 2021-11-17
Generate ext/gmp class synopses from stubs

kocsismate:


reference/gmp/gmp.xml
```diff
@@ -26,22 +25,20 @@
...
+     <classname>GMP</classname>
+    </ooclass>
 
     <classsynopsisinfo>
      <ooclass>
       <classname>GMP</classname>
      </ooclass>
-     <oointerface>
-      <interfacename>Serializable</interfacename>
-     </oointerface>
```
saundefined, line 37: Thanks! Looks good, but it seems `[skip-revcheck]` can not be applied

kocsismate: Ah, yes! I should have spotted this one.

kocsismate: I've just amended the commit message. I'm not sure though why the doc referenced `Serializable` as implemented interface...

kocsismate: I mean, I see that GMP uses a custom serialization/deserialization handler, but I don't know whether it should also implement the already deprecated `Serializable` interface

cmb69:
````
> var_dump(serialize(gmp_init(1)));
string(29) "C:3:"GMP":14:{s:1:"1";a:0:{}}"
````
That "C" looks like the result of Serializable::serialize(), what might explain that erroneous `implements Serializable`.

kocsismate: OK, so can I really get rid of the interface, right? I double checked that GMP didn't previously implemented the interface (explicitly) in former PHP version (`var_dump((new ReflectionClass(gmp_init(1)))->getInterfaceNames());` returns `array()`)

nikic: Yeah, GMP did something odd (my bad I guess ^^) I've implemented the __serialize/__unserialize methods in https://github.com/php/php-src/commit/17ceed9ae785a560631865af86038150d9785121.

kocsismate: Thanks for the correction!

kocsismate: I've just added an initial documentation for the 2 new methods

cmb69:
Thank you!

PR #1115, opened 2021-11-17
[skip-revcheck] Generate ext/fileinfo class synospes from stubs

kocsismate:


saundefined (commented):
It seems `[skip-revcheck]` can not be applied

kocsismate:
Ah, nice catch! I didn't notice that there's a new `&Methods`. Will fix this a bit later

kocsismate:
I've just amended the commit message

saundefined (approved):

PR #1120, opened 2021-11-17
Generate ext/reflection class synopses from stubs - part 1

kocsismate:


saundefined (approved):
It looks also good to me, thank you!

PR #1121, opened 2021-11-17
Generate ext/reflection class synopses from stubs - part 2

kocsismate:


saundefined (approved):
Thanks! Looks good to me

PR #1124, opened 2021-11-20
Add stream_socket_client usage to stream_get_meta_data

mallardduck:
As I reported to Psalm project in https://github.com/vimeo/psalm/issues/6947

I've discovered that Psalm was missing context for the `crypto` key-value when `stream_get_meta_data` is passed a `stream_socket_client` resource. When I reviewed PHP docs I found that this return data was also absent from the page.

From what I see this functionality was added to this function back in PHP 7.0's change cycle. So I believe adding this to the docs may have simply been overlooked.

Certainly open to feedback on how best to document the dynamic nature of this return value. Given that it only applies to PHP 7+ as well as only is returned for `stream_socket_resources`.

### DIFF
--- a/reference/stream/functions/stream-get-meta-data.xml
+++ b/reference/stream/functions/stream-get-meta-data.xml
@@ -25,7 +25,7 @@
      <listitem>
       <para>
        The stream can be any stream created by <function>fopen</function>,
-       <function>fsockopen</function> and <function>pfsockopen</function>.
+       <function>fsockopen</function> <function>pfsockopen</function> and <function>stream_socket_client</function>.
       </para>
      </listitem>
     </varlistentry>
@@ -121,6 +121,14 @@
      stream.
     </para>
    </listitem>
+   <listitem>
+    <para>
+     <literal>crypto</literal> (array) - the TLS connection metadata for this
+     stream. (Note: Only provided when used with a <link
+     linkend="function.stream-socket-client">stream_socket_client()</link>
+     resources.)
+    </para>
+   </listitem>
   </itemizedlist>
  </refsect1>
 
@@ -128,7 +136,7 @@
   &reftitle.examples;
   <para>
    <example>
-    <title><function>stream_get_meta_data</function> example</title>
+    <title><function>stream_get_meta_data</function> example using fopen</title>
     <programlisting role="php">
 <![CDATA[
 <?php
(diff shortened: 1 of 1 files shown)


reference/stream/functions/stream-get-meta-data.xml
```diff
@@ -121,14 +121,46 @@
...
+       <function>stream_socket_client</function> resources can now be used with this function.
+       Additionally, this will cause the return to include a new <literal>crypto</literal>
+       parameter.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
+ </refsect1>
+
```
Girgias, line 158: This changelog entry is not needed as we only document PHP 7 and above now, you might want to add this to the legacy docs however (which Zend is in charge of)

reference/stream/functions/stream-get-meta-data.xml
```diff
@@ -121,14 +121,46 @@
...
  <refsect1 role="examples">
   &reftitle.examples;
   <para>
    <example>
-    <title><function>stream_get_meta_data</function> example</title>
+    <title><function>stream_get_meta_data</function> example using fopen</title>
```
Girgias, line 163:
```suggestion
    <title><function>stream_get_meta_data</function> example using <function>fopen</funciton></title>
```

reference/stream/functions/stream-get-meta-data.xml
```diff
@@ -176,6 +208,61 @@ Array
...
+    $errorDescription,
+    40,
+    STREAM_CLIENT_CONNECT,
+    $streamContext
+);
+
+
+$meta = stream_get_meta_data($client);
+
+print_r($meta);
```
Girgias, line 241: Maybe use ``var_dump()`` instead?

mallardduck: Mostly used `print_r` as to match the existing example. I'll update both (and their outputs) to match `var_dump` insetad.

reference/stream/functions/stream-get-meta-data.xml
```diff
@@ -121,14 +121,46 @@
      stream.
     </para>
    </listitem>
+   <listitem>
+    <para>
+     <literal>crypto</literal> (array) - the TLS connection metadata for this
+     stream. (Note: Only provided when used with a <link
+     linkend="function.stream-socket-client">stream_socket_client()</link>
```
cmb69, line 128: That's doesn't appear to be true; try `var_dump(stream_get_meta_data(fopen("https://php.net", "r")));`.

mallardduck: Great catch - truthfully I hadn't tested it with other functions yet. So ultimately it's more accuate to say it's only included when using any HTTPS stream. Thank you @cmb69

mallardduck: Updated the note to read as:
> Only provided when the resource's stream uses TLS.

Though still open to suggestions for better phrasing here.

reference/stream/functions/stream-get-meta-data.xml
mallardduck: My added example and the existing one have been updated to use `var_dump()` instead.

reference/stream/functions/stream-get-meta-data.xml
mallardduck: Thank you for the direction on that - I've removed that section.

reference/stream/functions/stream-get-meta-data.xml
```diff
@@ -121,14 +121,20 @@
      stream.
     </para>
    </listitem>
+   <listitem>
+    <para>
+     <literal>crypto</literal> (array) - the TLS connection metadata for this
+     stream. (Note: Only provided when the resource's stream uses TLS.)
```
mallardduck, line 127: @cmb69 - I feel like this is more accurate to the behavior now.
I considered including `SSL` too, but felt it wasn't super necessary.
LMK.

cmb69: Yes, that's fine.

cmb69 (approved):
Thank you!

PR #1126, opened 2021-11-22
Document autovivification from false deprecation.

mumumu:
https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.autovivification-false

Based on php 8.1 migration guide, added autovivification note.

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -437,6 +437,12 @@ $arr[] = <replaceable>value</replaceable>;
      error. Formerly, the string was silently converted to an array.
     </simpara>
    </note>
+   <note>
+    <simpara>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Looks good to me. It's simple and seems to be in the right place. The only thing I am concerned about is that the previous paragraph says:

> If <varname>$arr</varname> doesn't exist yet, it will be created... 

It should specify that if $var doesn't exist or is set to null or false. Or something along these lines.

mumumu:
> It should specify that if $var doesn't exist or is set to null or false

Thank you for your feedback. I fixed the previous paragraph.

PR #1129, opened 2021-11-22
On page for key(), add link to array_key_first()

vlakoff:
`key()` is often used to get the first key of the array. But it brings various issues related to the array internal pointer, hence the introduction of `array_key_first()` (and its sibling `array_key_last()`) in PHP 7.3. Thus, a "see also" link to this function seems really appropriate, especially as people may not be aware of this newer function.


cmb69:
Ah yeah, very nice, thank you!

PR #1133, opened 2021-11-23
[PHP 8.1] Add Intersection types documentation

saundefined:
Based on [RFC](https://wiki.php.net/rfc/pure-intersection-types)


Girgias:
See #982 which should be ready to land.

PR #1134, opened 2021-11-23
[PHP 8.0] Document returning a non-array from __sleep() warning.

mumumu:
https://3v4l.org/m1pdW#v8.1rc3

Based on 8.1.0 [migration guide](https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.magic-sleep), added warning note about __sleep magic method.

### DIFF
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -98,7 +98,7 @@
    </note>
    <note>
     <para>
-     As of PHP 8.1.0, returning a value which is not an array from <link linkend="object.sleep">__sleep()</link> generates a diagnostic.
+     As of PHP 8.0.0, returning a value which is not an array from <link linkend="object.sleep">__sleep()</link> generates a warning. Previously, it generated a notice.
     </para>
    </note>
    <para>
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -96,6 +96,11 @@
      Use <link linkend="object.serialize">__serialize()</link> instead.
     </para>
    </note>
+   <note>
+    <para>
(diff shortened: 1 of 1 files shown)


language/oop5/magic.xml
```diff
@@ -96,6 +96,11 @@
      Use <link linkend="object.serialize">__serialize()</link> instead.
     </para>
    </note>
+   <note>
+    <para>
+     As of PHP 8.1.0, returning a value which is not an array from <link linkend="object.sleep">__sleep()</link> generates a diagnostic.
```
salathe, line 101: It looks like earlier versions also generated "a diagnostic" (assuming this means "a diagnostic error", e.g. E_NOTICE, E_WARNING, and friends).  I don't see any user-facing change that happened for 8.1.0 (but the level did change from E_NOTICE to E_WARNING in 8.0.0) for this situation.

salathe (commented):
The `__sleep()` section would benefit from a more general clean up with regards to (diagnostic) error levels and changes for PHP 8.

language/oop5/magic.xml
kocsismate, line 101: Yes, Salathe is right. I'd write something like this:

```suggestion
     As of PHP 8.0.0, returning a value which is not an array from <link linkend="object.sleep">__sleep()</link> generates a warning. Previously, it generated a notice.
```

afilina: I believe that this suggestion is pending review from @salathe.

afilina: @salathe Gentle nudge (see above). If I read this correctly, the suggestion above could be accepted and the PR merged.

PR #1136, opened 2021-11-24
BackedEnum is also an interface.

### LINKED ISSUE #1135: [Unit|Backed]Enum is an interface
The [manual for UnitEnum](https://www.php.net/manual/en/class.unitenum.php) currently shows:

```php
class UnitEnum {
    /* Methods */
    abstract public static cases(): array
}
```

When it should be:

```php
interface UnitEnum {
    /* Methods */
    public static cases(): array
}
```

Sorry for not sending a PR, I'm still struggling with getting the doc build locally.

mumumu:
Closes #1135
https://github.com/php/doc-en/issues/1135#issuecomment-977262147

### DIFF
--- a/language/predefined/backedenum.xml
+++ b/language/predefined/backedenum.xml
@@ -18,7 +18,7 @@
 <!-- }}} -->
 
   <section xml:id="backedenum.synopsis">
-   &reftitle.classsynopsis;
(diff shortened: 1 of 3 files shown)


cmb69 (approved):
Thank you!

PR #1139, opened 2021-11-24
Generate ext/reflection class synopses from stubs - part 4

kocsismate:


cmb69:
Thank you! I've added a changelog entry regarding `Reflector implements Stringable`.

Issue #980, opened 2021-09-28
PHP 8.1 documentation tracker


Girgias:
Pages which need to be added/modified for complete documentation of PHP 8.1.

Based of the migration guide: https://www.php.net/manual/en/migration81.php

This is currently still heavily a work in progress.

## Core:

- [x] New features:
   - [x] Type declarations
     - [x] Intersection types (#982)
     - [x] never (#889)
  - [x] Integer Octal Literal Prefix (#1091)
  - [x] Array Unpacking with String Keys (#1171)
  - [x] Named Argument After Argument Unpacking (#1329)
    * No RFC. related PR -> https://github.com/php/php-src/pull/7009
  - [x] full-path Key for File Uploads (#1130)
  - [x] Enumerations (#960)
  - [x] Fibers (#1096)
  - [x] First Class Callable Syntax (#1176)
  - [x] new in Initializers (#1327)
  - [x] Readonly properties (#1047)
  - [x] Final class constants (#1041)
- [x] Deprecated
   - [x] Implementing Serializable without __serialize() and __unserialize() (#1141)
   - [x] #1554
   - [x] Implicit incompatible float to int conversions (#1528)
   - [x] Calling a static element on a trait (#1330)
   - [x] Returning by reference from a void function (#1127)
   - [x] Autovivification from false (#1126)
- [x] BC Break
   - [x] $GLOBALS Access Restrictions (#1147)
   - [x] Usage of static Variables in Inherited Methods (#1549)
   - [x] Return Type Compatibility with Internal Classes (#1560)
- [x] Other Changes
   - [x] ReflectionProperty::setAccessible() and ReflectionMethod::setAccessible() no longer have an effect. (https://github.com/php/doc-en/commit/7b6a20ed5d1f1bec158b0c57b69bf021a9ed2799)

## Classes
- [x] New classes
   - [x] IntlDatePatternGenerator (#1026)
   - [x] ReflectionIntersectionType (#1027)
   - [x] CURLStringFile (#1154)
- [x] Resource to Opaque Object conversions (#972)

## Functions/Methods:
- [x] New functions
  - [x] Core
    - [x] `array_is_list()` (#986)
  - [x] Standard
    - [x]  `fsync()` (#1002)
    - [x]  `fdatasync()` (#1002)
  - [x] GD
     - [x] `imagecreatefromavif()` (#983)
     - [x] `imageavif()` (#983)
   - [x] MySQLi
[...]

PR #1141, opened 2021-11-24
[PHP 8.1] Add note about implementing Serializable without __serialize

iggyvolz:
Apologies if this is poorly formed - wasn't able to get the documentation to build properly on my machine but this looks like a fairly straightforward change.

Ref #980 "Implementing Serializable without __serialize() and __unserialize()"

RFC reference https://wiki.php.net/rfc/phase_out_serializable

### DIFF
--- a/language/predefined/serializable.xml
+++ b/language/predefined/serializable.xml
@@ -28,8 +28,7 @@
    
    <warning>
     <para>
-     As of PHP 8.1.0, a class which implements Serializable without also implementing <link linkend="object.serialize">__serialize()</link> and <link linkend="object.unserialize">__unserialize()</link> will generate a deprecation warning.
-     This interface will be removed in PHP 9.0.0.
+     As of PHP 8.1.0, a class which implements <interfacename>Serializable</interfacename> without also implementing <link linkend="object.serialize">__serialize()</link> and <link linkend="object.unserialize">__unserialize()</link> will generate a deprecation warning.
     </para>
    </warning>
   </section>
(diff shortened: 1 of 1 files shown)


saundefined:
Thank you! Maybe something like that?

language/predefined/serializable.xml
```diff
@@ -25,6 +25,13 @@
     a constructor instead of calling __construct(). If you need to execute the standard
     constructor you may do so in the method.
    </para>
+   
+   <warning>
+    <para>
+     As of PHP 8.1.0, a class which implements Serializable without also implementing <link linkend="object.serialize">__serialize()</link> and <link linkend="object.unserialize">__unserialize()</link> will generate a deprecation warning.
```
cmb69, line 31:
```suggestion
     As of PHP 8.1.0, a class which implements <classname>Serializable</classname> without also implementing <link linkend="object.serialize">__serialize()</link> and <link linkend="object.unserialize">__unserialize()</link> will generate a deprecation warning.
```
I also suggest to split that long line into multiple lines (that shouldn't be done for existing XML, but for new XML it's preferable).

language/predefined/serializable.xml
```diff
@@ -25,6 +25,13 @@
...
+     This interface will be removed in PHP 9.0.0.
```
cmb69, line 32: I don't think that we document such prognoses elsewhere, so I'd drop that sentence.

cmb69 (commented):
Thank you for the PR!

language/predefined/serializable.xml
salathe: Should be `<interfacename>`.

language/predefined/serializable.xml
salathe: > I don't think that we document such prognoses …

Not usually, no we don't.  Anything can happen between now and 9.0.0, and also there will need to be a future RFC for the removal.

Crell:
Thanks!

PR #1142, opened 2021-11-24
Correct note to state behavior is possible

agustingomes:
The behavior in the `Example #7 Old way to make arguments nullable` is still possible in PHP today.

For reference, I thought initially this was a bug in the PHP Core, and tried to add the respective test case in the PR:

https://github.com/php/php-src/pull/7627


cmb69:
Thank you!

PR #1147, opened 2021-11-25
[PHP 8.1] Document restriction of $GLOBALS usage.

mumumu:
https://wiki.php.net/rfc/restrict_globals_usage


Crell:
Should this also include a sample of how to work around this change to achieve the previous behavior?  (Basically `$foo = $GLOBALS`, modify $foo, then iterate $foo and assign each property back to $GLOBALS.)

Crell:
Thanks!

PR #1149, opened 2021-11-26
mb_ overloading was removed in 8.0.0

divinity76:
warning: untested, i don't know if `warn.deprecated.function-7-2-0.removed-8-0-0` is a real... thing. but i guess it's a real thing, after studying the code for similar error messages..

### DIFF
--- a/reference/mbstring/overloading.xml
+++ b/reference/mbstring/overloading.xml
@@ -7,7 +7,7 @@
  </title>
 
  <para>
-  &warn.deprecated.feature-7-2-0;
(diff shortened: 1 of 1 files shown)


cmb69:
> i don't know if warn.deprecated.function-7-2-0.removed-8-0-0 is a real... thing

All these entities are defined in https://github.com/php/doc-en/blob/master/language-snippets.ent; if it's already there (and it is), you can use it; otherwise would need to defined the entity in that file first.

cmb69:
Thank you!

Issue #1148, opened 2021-11-26
Add onboarding information to repo


nikic:
As mentioned in https://github.com/php/doc-en/issues/980#issuecomment-979545901, the repo itself currently doesn't contain any info on how to get started with the docs setup. It would be nice to include information either in the README or CONTRIBUTING.md, so people don't have to search for it elsewhere.

ssddanbrown:
Along the same lines of easing contribution access, I've put together a docker container with a preconfigured environment for building the docs:
https://github.com/ssddanbrown/PHP-Docs-Builder

Not sure if/where it would fit in with this repo (or the `doc-base` repo) but just sharing it here in the hopes others would find it useful.

hakre:
@njames: if it helps ([ref](https://github.com/php/doc-en/issues/980#issuecomment-979545901)), I also found http://doc.php.net/tutorial/local-setup.php but relatively late. Until then I kept some notes.

Otherwise I was able to go along starting with checking out the `doc-en` repository with the knowledge that this is docbook (XML) and that I needed to learn about how the php.net manual php pages are generated.

> # My Notes
> 
> > Technically a parenting project with sub-directories pointing to the different git repositories, here a setup to author the English docs, run the docbook pipeline, deploy into the php manual website (which then can be browsed with the PHP development server).
> >
> > Requirements: php-cli (~7.2+) + extensions + git(1) [+ rsync(1)]
> 
> ## Multiple Repositories:
> 
> > **NOTE:** find all repositories in the php org
> 
> * `doc-en` (-> `en`) the english documentation
> * `doc-base` building / docbook base system
> * `phd` PhD - PHP DocBook
> * `web-php`, full php.net source
> 
> ## Generating and Publishing the Manual
> 
> 1. Generate docbook:  \
  `$ php doc-base/configure.php --disable-libxml-check --enable-xml-details --redirect-stderr-to-stdout --with-lang=en`
> 2. Generate PHP manual:  \
   `$ php phd/render.php -d doc-base/.manual.xml -P PHP -f php`
> 3. Rsync the PHP manual:  \
   `$ rsync --info=progress2 -a output/php-web/ web-php/manual/en` 
     > **NOTE:** it is perhaps better to use a symbolic link, `web-php` 
     > suggests this is supported

This setup allowed me to file patches for `doc-en` as well as for `web-php`.

[...]

PR #1150, opened 2021-11-26
Add onboarding information

saundefined:
Added introductory information for new members.

Perhaps something needs to be corrected?

Relates #1148


cmb69:
Thank you!

In my opinion, we should also link to http://doc.php.net/tutorial/, or maybe have multiple links to the relevant sections of that tutorial.

salathe:
This is adding yet another copy of some "quick start" instructions that will not be kept up-to-date. :(

CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+
+```
+├── doc-base/
+└── en/
+└── phd/
+```
+
+Note that language directory must be without the `doc-` prefix.
+
+For cloning, you can use the command `git clone https://github.com/php/doc-en.git en`.
```
nikic, line 29: FWIW, I would use less words, more copy&pastable commands :)
```
# Setup
git clone https://github.com/php/phd.git
git clone https://github.com/php/doc-base.git
git clone https://github.com/php/doc-en.git en

# Test your changes
php doc-base/configure.php
php phd/render.php --docbook doc-base/.manual.xml --package PHP --format xhtml
# Open output/php-chunked-xhtml/ in your browser.
```

cmb69:
> This is adding yet another copy of some "quick start" instructions that will not be kept up-to-date. :(

Maybe we should remove that info from other places?

CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
+# PHP Manual Contribution Guide
```
salathe, line 1: This is missing a section on actually *contributing* changes.  There should probably be instructions on creating a fork (of doc-en, probably also doc-base), and adding the fork as a remote that contributors can push branches to, and creating PRs from their fork.

saundefined: Yes, I think it's definitely not worth pointing out here — there are more other guides on this.

Fixed the title with b201ee9.

salathe: So what's the point of this page if it's *not* going to be guiding folks on CONTRIBUTING(.md) to doc-en? 😕

saundefined: @salathe what do you think about this?
Just place links to the doc website, and "quick start" content?

afilina: I believe that this thread is waiting on a reply from @salathe.

afilina: @salathe Gentle nudge (see above). Does the syggestion by @saundefined make sense?

README.md
```diff
@@ -0,0 +1,36 @@
...
+
+````bash
+# Setup
+git clone https://github.com/php/phd.git
+git clone https://github.com/php/doc-base.git
+git clone https://github.com/php/doc-en.git en
+
+# Test your changes
+php doc-base/configure.php
+php phd/render.php --docbook doc-base/.manual.xml --package PHP --format xhtml
```
Girgias, line 17: I would make the output directory explicit with ``-o``

Girgias (approved):
Minor comment but LGTM

CONTRIBUTING.md
salathe: I've said my piece already, and continue to not approve of the changes here.
That said, my approval isn't (and never was) required. 😊

PR #1151, opened 2021-11-26
fixed typo

mowangjuanzi:
`NUL` is a typo? 
the correct one is  `NULL`?


kamil-tekiela:
No. It doesn't look like a typo. It talks about the NUL byte/character. We use this spelling to differentiate it from the PHP NULL value.

PR #1152, opened 2021-11-27
Add missing final modifier

villfa:
See https://3v4l.org/GheVJ


cmb69:
Thank you!

[ReflectionGenerator is only final as of PHP 8.0.0](https://3v4l.org/QOKvI), and as such needs a changelog entry.

villfa:
I've added the changelog entry.
I hope the phrasing is right.

cmb69:
Yes, it's fine. Thank you!

PR #1155, opened 2021-11-28
[8.0] Document remaining core changes

mallardduck:
This PR is an attempt to solve the remaining core items in: https://github.com/php/doc-en/issues/674

Specifically this will cover the following:

- [x] `new` and `instanceof` can now be used with arbitrary expressions
- [x] The ability to define case-insensitive constants has been removed. The third argument to define() may no longer be true.

### DIFF
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -170,13 +170,12 @@ $instance = new $className(); // new SimpleClass()
     is supported. This allows more complex instantiation if the expression
     produces a <type>string</type>.
    </para>
-   <para>
-    In the given example the arbitrary expression is a simple string concatenation.
-    However, this could also be a call to a function, or class method, that returns a
-    <type>string</type> containing the name of a valid class.
-   </para>
    <example>
     <title>Creating an instance using an arbitrary expression</title>
+    <para>
+     In the given example we show multiple examples of valid arbitrary expressions that produce a class name.
+     This shows a call to a function, string concatenation, and the <constant>::class</constant> constant.
+    </para>
     <programlisting role="php">
      <![CDATA[
 <?php
@@ -188,13 +187,13 @@ class ClassD extends ClassA {}
 
 function getSomeClass(): string
 {
-    return ClassA::class;
+    return 'ClassA';
 }
 
 var_dump(new (getSomeClass()));
 var_dump(new ('Class' . 'B'));
 var_dump(new ('Class' . 'C'));
-var_dump(new ClassD);
+var_dump(new (ClassD::class));
 ?>
 ]]>
     </programlisting>
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -167,8 +167,8 @@ $instance = new $className(); // new SimpleClass()
    </example>
    <para>
     As of PHP 8.0.0, using <literal>new</literal> with arbitrary expressions
-    is supported. This allows more complex instance creation as long as the
-    expression produces a <type>string</type>.
+    is supported. This allows more complex instantiation if the expression
+    produces a <type>string</type>.
    </para>
    <para>
     In the given example the arbitrary expression is a simple string concatenation.
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -166,6 +166,54 @@ $instance = new $className(); // new SimpleClass()
     </programlisting>
    </example>
    <para>
+    Starting in PHP 8.0 and forward using <literal>new</literal> with
+    arbitrary expressions is supported. This allows more complex instance creation
+    as long as the expression produces a <type>string</type>.
+   </para>
+   <para>
+    In the given example the arbitrary expression is a simple string concatenation.
+    However, this could also be a call to a function, or class method, that returns a
+    <type>string</type> containing the name of a valid class.
+   </para>
+   <example>
+    <title>Creating an instance using an arbitrary expression</title>
+    <programlisting role="php">
+     <![CDATA[
+<?php
+
+trait VerboseConstruct {
+    public function __construct() {
+        echo static::class . '::' . __FUNCTION__ . PHP_EOL;
+    }
+}
+
+class ClassA extends \stdClass { use VerboseConstruct; }
+class ClassB extends \stdClass { use VerboseConstruct; }
+class ClassC extends ClassB {}
+class ClassD extends ClassA {}
+
+$className = 'Class';
+foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
+    // Supported as of PHP 8.0+
+    new ($className . $classSuffix);
+    // Before PHP 8.0
+    // $fullClassName = $className . $classSuffix;
+    // new $fullClassName;
+}
+?>
+]]>
+    </programlisting>
+    &example.outputs.similar;
+    <screen>
+     <![CDATA[
+ClassA::__construct
+ClassB::__construct
+ClassC::__construct
+ClassD::__construct
+]]>
+    </screen>
+   </example>
+   <para>
     In the class context, it is possible to create a new object by
     <literal>new self</literal> and <literal>new parent</literal>.
    </para>
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -168,7 +168,7 @@ $instance = new $className(); // new SimpleClass()
    <para>
     As of PHP 8.0.0, using <literal>new</literal> with arbitrary expressions
     is supported. This allows more complex instantiation if the expression
-    produces a <type>string</type>.
+    produces a <type>string</type>. The express must be wrapped in parentheses.
    </para>
    <example>
     <title>Creating an instance using an arbitrary expression</title>
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -181,35 +181,35 @@ $instance = new $className(); // new SimpleClass()
      <![CDATA[
 <?php
 
-trait VerboseConstruct {
-    public function __construct() {
-        echo static::class . '::' . __FUNCTION__ . PHP_EOL;
-    }
-}
-
-class ClassA extends \stdClass { use VerboseConstruct; }
-class ClassB extends \stdClass { use VerboseConstruct; }
+class ClassA extends \stdClass {}
+class ClassB extends \stdClass {}
 class ClassC extends ClassB {}
 class ClassD extends ClassA {}
 
-$className = 'Class';
-foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
-    // Supported as of PHP 8.0.0
-    new ($className . $classSuffix);
-    // Before PHP 8.0
-    // $fullClassName = $className . $classSuffix;
-    // new $fullClassName;
+function getSomeClass(): string
+{
+    return ClassA::class;
 }
+
+var_dump(new (getSomeClass()));
+var_dump(new ('Class' . 'B'));
+var_dump(new ('Class' . 'C'));
+var_dump(new ClassD);
 ?>
 ]]>
     </programlisting>
     &example.outputs.8;
     <screen>
      <![CDATA[
-ClassA::__construct
-ClassB::__construct
-ClassC::__construct
-ClassD::__construct
+object(ClassA)#1 (0) {
+}
+object(ClassB)#1 (0) {
+}
+object(ClassC)#1 (0) {
+}
+object(ClassD)#1 (0) {
+}
+
 ]]>
     </screen>
    </example>
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -194,7 +194,7 @@ class ClassD extends ClassA {}
 
 $className = 'Class';
 foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
-    // Supported as of PHP 8.0+
+    // Supported as of PHP 8.0.0
     new ($className . $classSuffix);
     // Before PHP 8.0
     // $fullClassName = $className . $classSuffix;
@@ -203,7 +203,7 @@ foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
 ?>
 ]]>
     </programlisting>
-    &example.outputs.similar;
+    &example.outputs.8;
     <screen>
      <![CDATA[
 ClassA::__construct
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -168,7 +168,7 @@ $instance = new $className(); // new SimpleClass()
    <para>
     As of PHP 8.0.0, using <literal>new</literal> with arbitrary expressions
     is supported. This allows more complex instantiation if the expression
-    produces a <type>string</type>. The express must be wrapped in parentheses.
+    produces a <type>string</type>. The expressions must be wrapped in parentheses.
    </para>
    <example>
     <title>Creating an instance using an arbitrary expression</title>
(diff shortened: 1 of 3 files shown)


mallardduck:
I've mainly left this as a minor update rather than opting to completely reword and re-contextualize the case-sensitivity changes for constants.
I was considering rewriting it as `7.3` support will be removed in a few days and that's when the deprecation was introduced.
However I think that should wait until `7.4` support is also removed as only the new `8.0` and forward behavior will need to be considered the current behavior.

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
 ]]>
     </programlisting>
    </example>
+   <para>
+    Starting in PHP 8.0 and forward using <literal>new</literal> with
```
tiffany-taylor, line 169: Nit: "Starting in PHP 8.0" should probably be "As of PHP 8.0"

cmb69: Actually, `As of PHP 8.0.0, using …`. :)

mallardduck: Updated per @cmb69's suggestion. Also corrected similar phrase for the `define` change.

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
...
+    arbitrary expressions is supported. This allows more complex instance creation
```
tiffany-taylor: Nit: "instance creation" = instantiation

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
...
+    as long as the expression produces a <type>string</type>.
```
tiffany-taylor: Nit: "as long as" = if

language/operators.xml
```diff
@@ -2717,6 +2717,50 @@ var_dump(FALSE instanceof stdClass);
...
 <![CDATA[
 bool(false)
+]]>
+     </screen>
+    </example>
+   </para>
+   <para>
+    As of PHP 8.0, <literal>instanceof</literal> can now be used with
+    arbitrary expressions, like <code>$obj instanceof (expression)</code>.
+    <!-- RFC: https://wiki.php.net/rfc/variable_syntax_tweaks -->.
```
tiffany-taylor, line 2727: Remove `.` at the end of this line, it's causing two periods to appear in the above line.

reference/misc/functions/define.xml
```diff
@@ -67,13 +67,10 @@
       <warning>
        <simpara>
         Defining case-insensitive constants is deprecated as of PHP 7.3.0.
+        In PHP 8.0 only <literal>false</literal> is an acceptable value, passing
```
tiffany-taylor: Nit: "In PHP 8.0 only" should be "As of PHP 8.0, "

language/operators.xml
```diff
@@ -2717,6 +2717,50 @@ var_dump(FALSE instanceof stdClass);
     <screen>
 <![CDATA[
 bool(false)
+]]>
+     </screen>
+    </example>
+   </para>
+   <para>
+    As of PHP 8.0, <literal>instanceof</literal> can now be used with
+    arbitrary expressions, like <code>$obj instanceof (expression)</code>.
```
tiffany-taylor, line 2726: I'm conflicted on the usage of "like" here, but presently I'm unable to think of better phrasing. Possibly `for example`.

mallardduck: I think that sounds like a good suggestion.

language/operators.xml
```diff
@@ -2717,6 +2717,50 @@ var_dump(FALSE instanceof stdClass);
     <screen>
 <![CDATA[
 bool(false)
+]]>
+     </screen>
+    </example>
+   </para>
+   <para>
+    As of PHP 8.0, <literal>instanceof</literal> can now be used with
+    arbitrary expressions, for example <code>$obj instanceof (expression)</code>.
```
tiffany-taylor, line 2726: My apologies. `(expression)` should probably be something like `stdClass`. `(expression)` isn't horrible, but it could be better. Also, I'm being nitpicky again, but "for example" should have a comma after: `arbitrary expressions, for example, <code>$obj instanceof stdClass</code>.`

tiffany-taylor: Ahhhh, no, I'm wrong. The intent is to show `instanceof` with an expression, so `stdClass` wouldn't work.

mallardduck: I kept it simple (i.e. `(expression)` as the example I include directly after shows the concept better than I felt we could do inline. As IMO using only `stdClass` doesn't quite document the change fully.

I guess to show the concept inline it could be:
```
arbitrary expressions, for example, <code>$obj instanceof ('std' . 'Class')</code>.
```

FWIW I did follow the example set by the change log mention of this behavior update. Though I do understand if being more informal is acceptable there compared to in the canonical docs.

tiffany-taylor: Yeah, I like `('std' . 'Class')` better. It's a working example and it displays how the expression would work.

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
 ]]>
     </programlisting>
    </example>
+   <para>
+    As of PHP 8.0.0, using <literal>new</literal> with arbitrary expressions
+    is supported. This allows more complex instantiation if the expression
+    produces a <type>string</type>.
+   </para>
```
salathe, line 172: A key part of these "arbitrary expressions" is surrounding them with parentheses, which is wholly missing in this description.  (The same goes for the `instanceof` section.)

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
...
+   <para>
+    In the given example the arbitrary expression is a simple string concatenation.
+    However, this could also be a call to a function, or class method, that returns a
+    <type>string</type> containing the name of a valid class.
+   </para>
```
salathe, line 177: Since this is specifically referencing the new example, it should be within that `<example>` section.  Alternatively (my preference), don't reference the example and instead just give the list of some of the things that an "arbitrary expression" might be.

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
...
+
+$className = 'Class';
+foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
+    // Supported as of PHP 8.0+
+    new ($className . $classSuffix);
+    // Before PHP 8.0
+    // $fullClassName = $className . $classSuffix;
+    // new $fullClassName;
+}
+?>
```
salathe, line 203: This is a complicated example that obscures the demonstration of the feature (arbitrary expressions with `new`).  Something about seeing the wood for the trees.

Suggest having just a line or three, clearly demonstrating the feature and ideally nothing more.

language/operators.xml
```diff
@@ -2717,6 +2717,50 @@ var_dump(FALSE instanceof stdClass);
     <screen>
 <![CDATA[
 bool(false)
+]]>
+     </screen>
+    </example>
+   </para>
+   <para>
+    As of PHP 8.0.0, <literal>instanceof</literal> can now be used with
+    arbitrary expressions, for example, <code>$obj instanceof (expression)</code>.
```
salathe, line 2726: Suggest dropping the `for example , <code>$obj instanceof (expression)</code>` because a) it's not an example, b) it's immediately followed by... an `<example>`. 

See also my comment about parentheses, above.

language/operators.xml
```diff
@@ -2717,6 +2717,50 @@ var_dump(FALSE instanceof stdClass);
...
+$className = 'Class';
+foreach ([new ClassA, new ClassB, new ClassC, new ClassD] as $class) {
+    echo "Checking on " . $class::class . PHP_EOL;
+    foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
+        if ($class instanceof ($className . $classSuffix)) {
+            echo $class::class . ' is an instance of ' . ($className . $classSuffix) . PHP_EOL;
+        }
+    }
+}
+?>
```
salathe, line 2748: This is also a very complicated example.  Suggest demonstrating the feature much, much more succinctly (take the lead from other examples in this page).

reference/misc/functions/define.xml
```diff
@@ -99,6 +96,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        passing <literal>true</literal> to <parameter>case_insensitive</parameter> will produce a warning.
+        Note that passing <literal>false</literal> is allowed and will not produce any warning at all.
```
salathe, line 103:
```suggestion
        Passing &true; to <parameter>case_insensitive</parameter> now emits an <constant>E_WARNING</constant>. Passing &false; is still allowed.
```

salathe: Also, the changelog for 7.3.0 needs to be updated, since the `case_insensitive` parameter was *not* removed.
And the example with `case_insensitive` as `true` needs amending (at the very least some comments).

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
...
+}
+
+class ClassA extends \stdClass { use VerboseConstruct; }
+class ClassB extends \stdClass { use VerboseConstruct; }
+class ClassC extends ClassB {}
+class ClassD extends ClassA {}
+
+$className = 'Class';
+foreach (['A', 'B', 'C', 'D'] as $classSuffix) {
+    // Supported as of PHP 8.0+
```
salathe, line 197:
```suggestion
    // Supported as of PHP 8.0.0
```

language/oop5/basic.xml
```diff
@@ -165,6 +165,54 @@ $instance = new $className(); // new SimpleClass()
...
+    new ($className . $classSuffix);
+    // Before PHP 8.0
+    // $fullClassName = $className . $classSuffix;
+    // new $fullClassName;
+}
+?>
+]]>
+    </programlisting>
+    &example.outputs.similar;
```
salathe, line 206:
```suggestion
    &example.outputs.8;
```

language/operators.xml
```diff
@@ -2717,6 +2717,50 @@ var_dump(FALSE instanceof stdClass);
...
+]]>
+     </programlisting>
+     &example.outputs.80;
```
salathe, line 2751:
```suggestion
     &example.outputs.8;
```

salathe (commented):
Added some comments inline after a quick skim over the changes.

language/operators.xml
mallardduck: Here's a more simple example now - still captures what's unique to this change, but in the same format as examples above it.

language/oop5/basic.xml
mallardduck: This is a good point, yet it makes me wonder if the concept of "arbitrary expression" should be captured somewhere as well. From some light research this term is more often used in RFCs and discussion around them compared to in the docs.

The use of this term happens 30+ times in context of RFCs, but only 3 times in the currently published docs. None of the existing 3 instances actually explain needing to wrap in parentheses. Further, there's no mention of the word `parentheses` at all on the [Expressions](php.net/manual/en/language.expressions.php) page - save for mentions of them in comments on that page.

IMHO, with all that in mind it would be a better overall result if we document the concept of "arbitrary expressions" as a subheading on the Expressions page. This way the 3 existing instances -and those I'm adding- can all reference a canonical single location that explains the concept.

language/oop5/basic.xml
mallardduck: Simplified the example and made it similar to the one for `instanceof`

language/operators.xml
mallardduck: done

language/oop5/basic.xml
mallardduck: fixed; moved and rephrased.

cmb69:
> I was considering rewriting it as 7.3 support will be removed in a few days and that's when the deprecation was introduced.

The PHP manual is supposed to be more conservative than the php-src timeline. We dropped PHP 5 related documentation only about one year ago (and there is still quite some PHP 5 which has not been removed), and I consider it unlikely that we drop PHP 7.0 specifics (let alone PHP 7.4 specifics) any time soon.

reference/misc/functions/define.xml
```diff
@@ -67,13 +67,10 @@
...
         Defining case-insensitive constants is deprecated as of PHP 7.3.0.
+        As of PHP 8.0.0, only <literal>false</literal> is an acceptable value, passing
+        <literal>true</literal> will produce a warning.
        </simpara>
       </warning>
-      <note>
-       <para>
-        Case-insensitive constants are stored as lower-case.
-       </para>
-      </note>
```
cmb69, line 76: Please keep that note for PHP 7 users. :)

cmb69 (commented):
Thanks for the PR! This looks overall good to me, but the required parentheses need to be documented.

language/oop5/basic.xml
mallardduck: > no special concept called "arbitrary expression"

Yes, I understand that and it seems my point may not have been expressed well as the point was missed. What I was getting at is there is a concept in the syntax of wrapping `(` and `)` around an expression. JavaScript calls this syntax concept the "grouping operator" and the concept is documented in the MDN manual.

So even if I poorly communicated it, my point still stands that this **is** a concept that PHP uses generally and could document generally. And the community (especially new learners) would benefit from it. As it would allow the mention of required parenthesis here to tie back into a more general concept with a uniform explanation.

reference/misc/functions/define.xml
mallardduck: done

mallardduck:
> I consider it unlikely that we drop PHP 7.0 specifics

This is not what I meant at all. I was not implying that I would remove anything, rather I said:

> reword and re-contextualize the case-sensitivity changes for constants.

language/oop5/basic.xml
mallardduck: Looking over existing PHP docs further to hopefully provide a more clear suggestion here. The closest to having the parentheses grouping documented is on: https://www.php.net/manual/en/language.operators.precedence.php

However this does not IMHO explain the concept as well as the [MDN page I mentioned](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Grouping) does for instance.

mallardduck: Updated the two locations referring to this to mention requirements of using parenthesis.

language/oop5/basic.xml
```diff
@@ -168,7 +168,7 @@ $instance = new $className(); // new SimpleClass()
    <para>
     As of PHP 8.0.0, using <literal>new</literal> with arbitrary expressions
     is supported. This allows more complex instantiation if the expression
-    produces a <type>string</type>.
+    produces a <type>string</type>. The express must be wrapped in parentheses.
```
cmb69, line 171: Typo. Should be "expression" instead of "express".

cmb69:
Generally, the PHP manual is bad regarding the syntax and semantics of the PHP language. The [formal language specification](https://github.com/php/php-langspec/) is way better at that, but is mostly abandoned to due lack of maintainance. It would be good to integrate some of that information into the PHP manual, but we need to be careful to not be *too* diligent, to avoid that getting unmaintained as well.

language/oop5/basic.xml
mallardduck: fixed

cmb69:
Thank you!

PR #1156, opened 2021-11-28
Adding initial documentation to stream bucket make writable

nyamsprod:
Here's an initial documentation of the object returned by `stream_bucket_make_writable`. Let me know in the comment if more informations is needed or if I missed something. Since it's my first contribution to the documentation.

### DIFF
--- a/reference/stream/functions/stream-bucket-make-writeable.xml
+++ b/reference/stream/functions/stream-bucket-make-writeable.xml
@@ -5,15 +5,20 @@
  <refnamediv>
   <refname>stream_bucket_make_writeable</refname>
   <refpurpose>
-   Return a bucket object from the brigade for operating on
+   Returns a bucket object from the brigade to operate on
   </refpurpose>
  </refnamediv>
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
    <type class="union"><type>object</type><type>null</type></type><methodname>stream_bucket_make_writeable</methodname>
    <methodparam><type>resource</type><parameter>brigade</parameter></methodparam>
   </methodsynopsis>
+  <para>
+   This function is called whenever you need to access and operate on the content contains in a brigade. It is
+   typically called from <methodname>php_user_filter::filter</methodname>.
+  </para>
  </refsect1>
 
  <refsect1 role="parameters">
@@ -24,38 +29,45 @@
      <term><parameter>brigade</parameter></term>
      <listitem>
       <para>
-       The resource brigade to return a bucket object from.
+       The brigade to return a bucket object from.
       </para>
      </listitem>
     </varlistentry>
    </variablelist>
   </para>
  </refsect1>
+
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an object with properties listed in the table below or &null;.
-  </para>
-  <para>
-   <table>
-    <title>
-     Returned Object for <function>stream_bucket_make_writeable</function>
-    </title>
-    <tgroup cols="2">
-     <tbody>
-      <row>
-       <entry>data</entry>
-       <entry>the current string in the bucket</entry>
-      </row>
-      <row>
-       <entry>datalen</entry>
-       <entry>the length of the string in the bucket</entry>
-      </row>
-     </tbody>
-    </tgroup>
-   </table>
+   Returns an bucket object with the properties listed below or &null;.
+   <variablelist>
+    <varlistentry>
+     <term>
+      <parameter>data</parameter>
+      <type>string</type>
+     </term>
+     <listitem>
+      <para>
+       <parameter>data</parameter> <literal>bucket</literal> public property containing the current string in the bucket.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term>
+      <parameter>datalen</parameter>
+      <type>integer</type>
+     </term>
+     <listitem>
+      <para>
+       <parameter>datalen</parameter> <literal>bucket</literal> public property containing the length of the string in the bucket.
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
   </para>
  </refsect1>
+
  <refsect1 role="seealso">
   &reftitle.seealso;
   <para>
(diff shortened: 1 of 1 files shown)


reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -14,11 +14,57 @@
    <type class="union"><type>object</type><type>null</type></type><methodname>stream_bucket_make_writeable</methodname>
    <methodparam><type>resource</type><parameter>brigade</parameter></methodparam>
   </methodsynopsis>
-
-  &warn.undocumented.func;
-
```
cmb69, line 19: We should have a description of the function here. Basically:
````
  <para>
   …
  </para>
````
This could just match the `<refpurpose>`, or preferably, could be a little bit more elaborate. Also, the refpurpose is worded strangely, and is a bit long.

File: reference/stream/functions/stream-bucket-make-writeable.xml [hunk 2]
```diff
@@ -14,11 +14,57 @@
...
 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>brigade</parameter></term>
+     <listitem>
+      <para>
+       The resource brigade to return a bucket object from.
```
cmb69, line 27: Might be better to remove the "resource" here (makes it easier when it will be replaced with an object, and the type is already part of the function signature):
```suggestion
       The brigade to return a bucket object from.
```

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -14,11 +14,57 @@
...
+   </variablelist>
+  </para>
+ </refsect1>
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns an object with properties listed in the table below or &null;.
+  </para>
+  <para>
+   <table>
```
cmb69, line 40: I don't think a table is a good solution to document the properties of an object (although we use that elsewhere, e.g. for `imap_fetchstructure()`). How about a `<variablelist>` instead (like for the parameters)? So basically, something like:
````
  <variablelist>
   <varlistentry>
    <term>
     <parameter>data</parameter>
     <type>string</type>
    </term>
    <listitem>
     <para>
      …
     </para>
    </listitem>
   </varlistentry>
   …
  </variablelist>
````

File: reference/stream/functions/stream-bucket-make-writeable.xml [hunk 4]
```diff
@@ -14,11 +14,57 @@
...
+     <listitem>
+      <para>
+       The resource brigade to return a bucket object from.
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+ <refsect1 role="returnvalues">
```
cmb69, line 34: Nit:
```suggestion
 </refsect1>

 <refsect1 role="returnvalues">
```

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -14,11 +14,57 @@
...
+      <row>
+       <entry>datalen</entry>
+       <entry>the length of the string in the bucket</entry>
+      </row>
+     </tbody>
+    </tgroup>
+   </table>
+  </para>
+ </refsect1>
+ <refsect1 role="seealso">
```
cmb69, line 59:
```suggestion
 </refsect1>

 <refsect1 role="seealso">
```

cmb69 (commented):
Thank you for the PR!

reference/stream/functions/stream-bucket-make-writeable.xml
nyamsprod: Indeed I used `imap_fetchstructure` as a base for reference and I did not know if I could use `<variablelist>` will look into using it instead and it enables adding the property type too.

reference/stream/functions/stream-bucket-make-writeable.xml
nyamsprod: I added the missing description

reference/stream/functions/stream-bucket-make-writeable.xml
nyamsprod: `resource` word is removed

reference/stream/functions/stream-bucket-make-writeable.xml
nyamsprod: `<variablelist>` node added.

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -5,20 +5,78 @@
...
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
 
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns an bucket object with the properties listed below or &null;.
```
cmb69, line 43:
```suggestion
   Returns a bucket object with the properties listed below or &null;.
```

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -5,20 +5,78 @@
...
+   <variablelist>
+    <varlistentry>
+     <term>
+      <parameter>data</parameter>
+      <type>string</type>
```
cmb69, line 48: This would basically create a JS link (pointing to the signature, but not working since there is no such parameter). We could use `<property>` instead:
```suggestion
      <property>data</property>
      (<type>string</type>)
```
Since `<property>` is not catered to by PhD yet, I suggest to add parentheses around the type.

Rendered as
Screenshot 2021-11-29 172159
:

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -5,20 +5,78 @@
...
+     </term>
+     <listitem>
+      <para>
+       <parameter>data</parameter> <literal>bucket</literal> public property containing the current string in the bucket.
```
cmb69, line 52: Why not:
```suggestion
       The current string in the bucket.
```

cmb69 (commented):
Thanks for the updates!

cmb69:
Thank you!

nyamsprod:
thanks for the help during the PR review 👍🏾

PR #1163, opened 2021-11-29
Adjusting Changelog versions for getenv() & get_defined_functions() parameters

DmitryTronin:
The inconsistencies were detected during functions reflection inspections:
- for `get_defined_functions()`: https://3v4l.org/T7JFt
- for `getenv()`: https://3v4l.org/IVadX#v540

### DIFF
--- a/reference/funchand/functions/get-defined-functions.xml
+++ b/reference/funchand/functions/get-defined-functions.xml
@@ -61,7 +61,7 @@
       </entry>
      </row>
      <row>
-      <entry>7.0.15, 7.1.1</entry>
+      <entry>7.1.1</entry>
       <entry>
        The <parameter>exclude_disabled</parameter> parameter has been added.
       </entry>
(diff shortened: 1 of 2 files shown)


reference/funchand/functions/get-defined-functions.xml
```diff
@@ -61,7 +61,7 @@
       </entry>
      </row>
      <row>
-      <entry>7.0.15, 7.1.1</entry>
```
cmb69, line 64: Reflection info is wrong; the [parameter is available as of 7.0.15](https://github.com/php/php-src/blob/php-7.0.15/Zend/zend_builtin_functions.c#L1902).

reference/info/functions/getenv.xml
```diff
@@ -82,7 +82,7 @@
       </entry>
      </row>
      <row>
-      <entry>7.0.9</entry>
+      <entry>5.6.24</entry>
```
cmb69, line 85: The manual does no longer document PHP 5 behavior; however, the parameter is available as of 7.0.9 (but not in prior PHP 7.0 versions).

cmb69 (commented):
Thank you for the PR! However, the changes are not correct/irrelevant.

PR #1165, opened 2021-11-29
Drop use of noise word "respectively"

MorganAWS:
That's not what "respectively" means.

http://www.lexico.com/definition/respectively

"The PSpell functions now accept and return, respectively, PSpell\Config objects instead of pspell config resources."
means (a) there are precisely _two_ PSpell functions, where the first one (and you know which one that is) now accepts an object instead of a resource, and the second returns one.


kamil-tekiela (approved):
I approve. That word does not fit here.

PR #1167, opened 2021-12-01
Document PDO::PGSQL_ATTR_DISABLE_PREPARES attribute

therealgaxbo:
The attribute was added in 2014 but never documented.


cmb69:
Thank you!

PR #1168, opened 2021-12-01
Add list of attribute target constants

Firehed:
I wasn't able to find these anywhere in the Attributes documentation, and needed to use reflection to find the valid values.

I think the constant names are self-explanatory enough to not require additional information.

### DIFF
--- a/language/attributes.xml
+++ b/language/attributes.xml
@@ -336,13 +336,13 @@ class MyAttribute
    <para>The following targets can be specified:</para>
    
    <simplelist>
-    <member><literal>Attribute::TARGET_CLASS</literal></member>
-    <member><literal>Attribute::TARGET_FUNCTION</literal></member>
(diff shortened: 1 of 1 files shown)


language/attributes.xml
```diff
@@ -333,6 +333,18 @@ class MyAttribute
     </para>
    </example>
 
+   <para>The following targets can be specified:</para>
+   
+   <simplelist>
+    <member><literal>Attribute::TARGET_CLASS</literal></member>
```
cmb69, line 339:
```suggestion
    <member><constant>Attribute::TARGET_CLASS</constant></member>
```
Same for the others below.

cmb69 (changes requested):
Thank you for the PR! Besides the comment, this looks good to me.

Firehed:
Done! Thank you for the feedback :)

cmb69 (approved):
Thank you!

PR #1169, opened 2021-12-01
get_class does not return false

kamil-tekiela:


reference/classobj/functions/get-class.xml
```diff
@@ -61,7 +60,8 @@
   &reftitle.errors;
   <para>
    If <function>get_class</function> is called with anything other than an
-   object, an <constant>E_WARNING</constant> level error is raised.
+   object, <classname>TypeError</classname> is raised. Prior to PHP 8.0.0,
```
salathe, line 63: If called with no arguments, an `Error` is raised.

salathe (commented):
The change from returning `false` to throwing `Error`, when calling `get_class()` without any arguments from outside of a class, should be a new changelog entry.

The *"The parameter is still optional and calling…"* paragraph in the description of the `object` parameter needs some attention as no `E_WARNING` is emitted in PHP 8.

reference/classobj/functions/get-class.xml
```diff
@@ -61,7 +57,13 @@
...
+   an <constant>E_WARNING</constant> level error was raised.
+  </para>
+  <para>
+   If <function>get_class</function> is called with no arguments,
```
cmb69, line 64:
```suggestion
   If <function>get_class</function> is called with no arguments from outside a class,
```
The (change of) behavior of this function is a mess. Might have been better to just default `$object` to `$this`.

kamil-tekiela: It's not really `$this`. It's `self` but I don't think there exists a mechanism in PHP that would allow expressing this as a parameter default value. The default value must remain `?` for the lack of a better one. https://3v4l.org/F4jID

cmb69 (approved):
Thank you!

PR #1170, opened 2021-12-02
Inherited constants were unintentionally removed from ReflectionEnumUnitCase class synopsis page

kocsismate:


cmb69 (approved):
Ah, right, thanks!

PR #1172, opened 2021-12-03
Moved [overview|example] of Fibers to language/fibers.xml page.

mumumu:
Moved Fibers overview contents to be linked from the top page of manual, because it is core functionality like Generators.

We must merge https://github.com/php/doc-base/pull/53, or build will fail.

### DIFF
--- a/appendices/migration81/new-features.xml
+++ b/appendices/migration81/new-features.xml
@@ -81,7 +81,7 @@ $arr2 = [...$arr1, 'c' => 'd']; //[1, 'a', 'c' => 'd']
    <title>Fibers</title>
 
    <para>
-    Support for <link linkend="class.fiber">Fibers</link> has been added.
+    Support for <link linkend="language.fibers">Fibers</link> has been added.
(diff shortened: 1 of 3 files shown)


cmb69 (commented):
Thank you! Basically, this looks good to me, but I would add a link to `language.fibers` on the Fiber class page. Also this requires a modification to doc-base/manual.in.xml, so that the new language/fibers page is actually added to the manual.

mumumu:
> Also this requires a modification to doc-base/manual.in.xml

added in https://github.com/php/doc-base/commit/d6a6dd152187ac935ee077f3f5ff958e1a9bf920

cmb69:
Thank you! I've merged as is, to avoid broken builds in other PRs etc. We can still improve when we want to. :)

mumumu:
> We can still improve when we want to. :)

Oh, I'm changing language/predefined/fiber.xml now.
Anyway, I'll commit its changes soon.

PR #1173, opened 2021-12-05
Update version information for Generator::__wakeup()

tiffany-taylor:
`Generator::__wakeup()` was removed in PHP 7.4.0.
https://3v4l.org/XlQfU

This change also puts the documented behavior regarding Exception throwing in a warning box. Please advise if the wording should be improved. I considered adding an example, but not sure if it's worth the effort.

### DIFF
--- a/appendices/migration81/new-features.xml
+++ b/appendices/migration81/new-features.xml
@@ -153,7 +153,7 @@ $arr2 = [...$arr1, 'c' => 'd']; //[1, 'a', 'c' => 'd']
    <title>Final class constants</title>
 
    <para>
-    Added support for the <modifier>final</modifier> modifier for class constants.
+    Added support for <link linkend="language.oop5.final.example.php81">the <modifier>final</modifier> modifier for class constants</link>.
     Also, interface constants become overridable by default.
     <!-- RFC: https://wiki.php.net/rfc/final_class_const -->
    </para>
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -60,7 +60,7 @@ class ChildClass extends BaseClass {
    </example>
   </para>
   <para>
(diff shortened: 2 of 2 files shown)


language/predefined/versions.xml
```diff
@@ -88,7 +88,7 @@
  <function name="generator::send" from="PHP 5 &gt;= 5.5.0, PHP 7, PHP 8"/>
  <function name="generator::throw" from="PHP 5 &gt;= 5.5.0, PHP 7, PHP 8"/>
  <function name="generator::valid" from="PHP 5 &gt;= 5.5.0, PHP 7, PHP 8"/>
- <function name="generator::__wakeup" from="PHP 5 &gt;= 5.5.0, PHP 7, PHP 8"/>
+ <function name="generator::__wakeup" from="PHP 5 &gt;= 5.5.0, PHP 7.3"/>
```
cmb69, line 91: This should rather be:
```suggestion
 <function name="generator::__wakeup" from="PHP 5 &gt;= 5.5.0, PHP 7 &lt; 7.4.0"/>
```

cmb69 (commented):
It seems to me that the most important fact is that Generators cannot be (un)serialized. The removal of that method doesn't change that (it's now prevented by other means). I'm not sure how we document the inability to (un)serialize elsewhere. I assume we're missing that for several classes (see https://heap.space/s?refs=ZEND_ACC_NOT_SERIALIZABLE&project=PHP-8.1).

tiffany-taylor:
I think I'll change this PR to reword not-serializable _things_ into something more explicit. I'll need to go through one-by-one based on lxr.

tiffany-taylor:
This was closed by mistake, but I will resolve it later.

PR #1174, opened 2021-12-06
fix FFI non-static method is written to static method

chopins:
fix FFI non-static method is written to static method
`FFI:cast()`, `FFI::new()` and `FFI:type` is mixed non-static and static method,


cmb69:
Thank you!

@kocsismate, these three functions are overloaded (static and non-static). Is there something we can/should do in the stubs to prevent 5770355a66d2949f3132c7c6f4808dc035121a42 from happening again?

kocsismate:
> these three functions are overloaded (static and non-static). Is there something we can/should do in the stubs to prevent 5770355 from happening again?

oh, gosh.. Currently, there's no way to exclude any functions from synchronization unless a function has an `UNKNOWN` param default value. As the FFI methods in question doesn't override the parameter list, we could implement a `@overridden` tag or something like that to make this type of method overriding detectable by `gen_stub.php`.

cmb69:
Yeah, not sure what to do. It would have been better to have separate methods instead of [that kind of overload](https://github.com/php/php-src/blob/fe1f613b7510272f493f3dc4e291e6928347ae95/ext/ffi/ffi.c#L2763-L2780).

kocsismate:
> Yeah, not sure what to do. It would have been better to have separate methods instead of that kind of overload.

I'll try not to forget about these two bastards when I start working on the overridden signature deprecation RFC. :)

cmb69:
Thanks! Note that there are 3 methods: `FFI::new()`, `FFI::cast()` and `FFI::type()`.

kocsismate:
> Thanks! Note that there are 3 methods: FFI::new(), FFI::cast() and FFI::type().

Yeah, I updated my previous comment just after sending it because I also realized there are 3 of them :)

PR #1176, opened 2021-12-06
[PHP 8.1] Document the first class callable syntax.

mumumu:
Added the first class callable syntax docs, based on https://wiki.php.net/rfc/first_class_callable_syntax.

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1509,7 +1509,7 @@ var_export($x);  // Outputs 1
 
    <para>
     The first class callable syntax is introduced as of PHP 8.1.0, as a way of creating <link linkend="functions.anonymous">anonymous functions</link> from <link linkend="language.types.callable">callable</link>.
-    It supersedes existing callable encodings using strings and arrays.
+    It supersedes existing callable syntax using strings and arrays.
     The advantage of this syntax is that it is accessible to static analysis, and uses the scope at the point where the callable is acquired.
    </para>
 
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1504,6 +1504,127 @@ var_export($x);  // Outputs 1
    </sect2>
   </sect1>
 
+  <sect1 xml:id="functions.first_class_callable_syntax">
+   <title>First class callable syntax</title>
+
+   <para>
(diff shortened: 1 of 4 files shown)


language/functions.xml
```diff
@@ -1504,6 +1504,127 @@ var_export($x);  // Outputs 1
    </sect2>
   </sect1>
 
+  <sect1 xml:id="functions.first_class_callable_syntax">
+   <title>First class callable syntax</title>
+
+   <para>
+    The first class callable syntax is introduced as of PHP 8.1.0, as a way to creating <link linkend="functions.anonymous">anonymous functions</link> from <link linkend="language.types.callable">callable</link>.
+    It supersedes existing callable encodings using strings and arrays.
```
kamil-tekiela, line 1512: What are "callable encodings"? Is this phrase defined somewhere? I can only find this phrase used once in RFC and it is a little bit ambiguous. Maybe we should explain what it means in the manual.

language/functions.xml
```diff
@@ -1504,6 +1504,127 @@ var_export($x);  // Outputs 1
...
+    The advantage of this syntax is that it is accessible to static analysis, and respects the scope at the point where the callable is created.
```
kamil-tekiela, line 1513:
```suggestion
    The advantage of this syntax is that it is accessible to static analysis, and uses the scope at the point where the callable is acquired.
```

language/types/callable.xml
```diff
@@ -48,6 +48,12 @@
    passed to a callback parameter.
   </para>
 
+  <note>
+   <para>
+    As of PHP 8.1.0, anonymous functions can also be created by using the <link linkend="functions.first_class_callable_syntax">first class callable syntax</link>.
```
kamil-tekiela, line 53:
```suggestion
    As of PHP 8.1.0, anonymous functions can also be created using the <link linkend="functions.first_class_callable_syntax">first class callable syntax</link>.
```

language/functions.xml
mumumu: > What are "callable encodings"? 

It means traditional callable syntax as following, so I replaced `encodings` with `syntax`.

```php
// traditional callable using string, array
$f7 = 'strlen'(...);
$f8 = [$obj, 'method'](...);
```

PR #1183, opened 2021-12-08
Update DateTime format table for easier to read.

puncoz:
I always have problems while reading PHP DateTime format table for reference. Keeping symbols and their example together will be really helpful for all of us who refer the docs for quick reference.


cmb69:
I'm in favor of the current column order.

tiffany-taylor:
I can see benefit to the change, as it displays the format value and an example of what it looks like next to each other, then description on the far right, which is what feels natural to me. But, I would like to pull these changes locally and verify that that this looks how I'm expecting it to.

tiffany-taylor:
Tested locally, there's an issue in that occasionally we have examples in the `Description` column, and descriptions in the `Example` column, instead of how we have it listed, where examples should be in the `Example` column, and description should be in the `Description` column. First thing that should be done is normalize this: if it's an example, it should be in the `Example` column, if it's a description, it should be in the `Description` column. Once that is complete, I think that having the descriptions on the far right may actually improve readability of this table.

salathe:
I'm :-1: on this change.  The examples are supplementary to the other columns (in this case, format and description), and IMO should not be given higher reading priority than the _required reading_ (to clarify: the description column _is_ required reading).

If still planning to go ahead with this change, I have a few requests:
* Resolve the merge conflicts
* Have a plan (if not in this PR, then a separate Issue/PR) for changing the columns in other similar tables throughout the manual (I haven't checked, but would guesstimate maybe a dozen or so).

tiffany-taylor:
The primary reason I am for this change is that there are times where I need to refer to this table to remember which format character to use, and I will have to mentally juggle the descriptions/examples to determine what I need while also keeping the context of the code I am working on in my head. I usually end up having to use a whiteboard to assist me. However, this is my own use case, and I'd probably be better off creating a cheatsheet for myself rather than altering the manual to fit my needs. I agree that enforcing reading all of the docs is a goal we should strive towards.

afilina:
To help the conversation, here is a screenshot of the page as it was meant by the author. Note that merge conflicts prevented me from rebasing.

Date formats

puncoz:
I still strongly support these changes because I believe that most of us use documentation as a quick reference rather than a comprehensive learning tool. By keeping symbols and their examples together, we can easily find what we need without the added mental effort of trying to decipher lengthy descriptions. Of course, there may be a few exceptions to this, but for the majority of us, this organization will make it much easier to find the information we need quickly. That being said, descriptions can still be useful when we need additional context or clarification on a particular symbol or example.

afilina:
Assuming that @puncoz wishes to continue with this PR:

> If still planning to go ahead with this change, I have a few requests:
> - Resolve the merge conflicts

Merge conflicts were resolved.

> - Have a plan (if not in this PR, then a separate Issue/PR) for changing the columns in other similar tables throughout the manual (I haven't checked, but would guesstimate maybe a dozen or so).

Is there an issue (todo) that lists all the places that need to be changed for consistency in subsequent PRs, and would those changes help readability in these other contexts? I suggest linking to the issue from this PR.

Girgias:
As this is related to datetime I prefer to get @derickr's opinion on this change

derickr:
I would say no to this, and I agree with @salathe

PR #1184, opened 2021-12-08
[PHP 8.1] Document passing null to scalar parameter of built-in functions

mumumu:
https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -979,6 +979,12 @@ $func(); // prints "bar"
      be thrown in this case.
     </simpara>
    </note>
+   <note>
+    <simpara>
+     As of PHP 8.1.0, passing &null; to scalar parameter of built-in functions is deprecated.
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR!

However, that info is misleading (also in the migration guide). In strict_types mode, passing null to internal functions which had non-nullable parameters was [already an error in PHP 7](https://3v4l.org/qIuHI). In non strict_types modes, `null` was coerced to the declared type instead (like with usual type juggling); only this does no longer happen. Especially, stating that "scalar types for built-in functions are nullable by default" is misleading at best.

mumumu:
> only this does no longer happen.

Should we not only close this PR, also delete misleading descrption in migration guide?

cmb69:
> Should we not only close this PR, also delete misleading descrption in migration guide?

Either that, or we should update/fix the description in the migration guide. :)

PR #1191, opened 2021-12-09
Optional parameter changes in 8.1, and description improvements

IMSoP:
* Document change in behaviour of misplaced optional parameters in 8.1
* Expand and update documentation of default/optional parameters

### DIFF
--- a/appendices/migration81/incompatible.xml
+++ b/appendices/migration81/incompatible.xml
@@ -63,8 +63,10 @@ var_dump(B::counter()); // int(4), previously int(2)
     specified before required parameters is now always treated as required,
     even when called using
     <link linkend="functions.named-arguments">named arguments</link>.
-    In PHP 8.0, the below emits a deprecation notice on the definition,
-    but runs successfully when called. As of PHP 8.1, an error is thrown.
+    As of PHP 8.0.0, but prior to PHP 8.1.0, the below emits a deprecation notice
+    on the definition, but runs successfully when called. As of PHP 8.1.0, an error
+    of class <classname>ArgumentCountError</classname> is thrown, as it would be when
+    called with positional arguments.
 
     <informalexample>
      <programlisting role="php">
@@ -87,7 +89,7 @@ catch (Error $e)
      </programlisting>
      &example.outputs.80;
      <screen>
-      <![CDATA[
+<![CDATA[
 Deprecated: Required parameter $flavour follows optional parameter $container
  in example.php on line 3
 Making a bowl of raspberry yogurt.
@@ -102,7 +104,11 @@ ArgumentCountError - makeyogurt(): Argument #1 ($container) not passed
 ]]>
      </screen>
     </informalexample>
-
+   </para>
+   <para>
+    Note that a default value of &null; can be used before required parameters to
+    specify a <link linkend="language.types.declarations.nullable">nullable type</link>,
+    but the parameter will still be required.
    </para>
   </sect3>
 
--- a/appendices/migration81/incompatible.xml
+++ b/appendices/migration81/incompatible.xml
@@ -55,6 +55,57 @@ var_dump(B::counter()); // int(4), previously int(2)
    </para>
   </sect3>
 
+  <sect3 xml:id="migration81.incompatible.core.optional-before-required">
+   <title>Optional parameters specified before required parameters</title>
+
+   <para>
+    An <link linkend="functions.arguments.default">optional parameter</link>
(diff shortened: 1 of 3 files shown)


cronlabspl:
Lgtm

appendices/migration81/incompatible.xml
```diff
@@ -55,6 +55,57 @@ var_dump(B::counter()); // int(4), previously int(2)
...
+catch (Error $e)
+{
+    echo get_class($e), ' - ', $e->getMessage(), "\n";
+}
+?>
+]]>
+     </programlisting>
+     &example.outputs.80;
+     <screen>
+      <![CDATA[
```
saundefined, line 90:
```suggestion
<![CDATA[
```

nit: indentation

cronlabspl: > nit: indentation

why line `90` shouldnt be indented?

cmb69: That's just an (undocumented) convention for the PHP manual sources.

appendices/migration81/incompatible.xml
```diff
@@ -55,6 +55,57 @@ var_dump(B::counter()); // int(4), previously int(2)
...
+  <sect3 xml:id="migration81.incompatible.core.optional-before-required">
+   <title>Optional parameters specified before required parameters</title>
+
+   <para>
+    An <link linkend="functions.arguments.default">optional parameter</link>
+    specified before required parameters is now always treated as required,
+    even when called using
+    <link linkend="functions.named-arguments">named arguments</link>.
+    In PHP 8.0, the below emits a deprecation notice on the definition,
+    but runs successfully when called. As of PHP 8.1, an error is thrown.
```
cmb69, line 67: The convention is to always specify the full version (major, minor, revision), so:
```suggestion
    Prior to PHP 8.1.0, the below emits a deprecation notice on the definition,
    but runs successfully when called. As of PHP 8.1.0, an error is thrown.
```

language/functions.xml
```diff
@@ -384,23 +410,62 @@ Making a bowl of raspberry .
...
+    <note>
+     <simpara>
+      As of PHP 7.1, omitting a parameter which does not specify a default
+      throws an <classname>ArgumentCountError</classname>; in previous versions
+      it raised a Warning.
+     </simpara>
+     <simpara>
+      As of PHP 8.0, <emphasis>defining</emphasis> a function with optional
+      parameters after required parameters (as in the first "yogurt"
+      example above) is DEPRECATED.
```
cmb69, line 466: Maybe better (I don't like screaming):
```suggestion
      example above) is <emphasis>deprecated</emphasis>.
```

cmb69 (commented):
Thank you for the PR!

Maybe it's worthwhile to document that a `null` default still signals that the type is nullable (although the parameter would not be optional).

appendices/migration81/incompatible.xml
IMSoP: The oddity here is that the code *only* runs successfully on 8.0.x, not any earlier versions, so "prior to" feels a bit misleading. I'll try to come up with an alternative wording.

cmb69: Oh, right. Maybe the clumsy "As of PHP 8.0.0, but prior to PHP 8.1.0" will do?

IMSoP: Yeah, I can't think of anything better at the moment.

language/functions.xml
IMSoP: It turns out, this was already mentioned in an oddly placed paragraph right at the top of the page, so I've moved it down into this section.

appendices/migration81/incompatible.xml
IMSoP: I had noticed that convention, but PhpStorm hadn't, so had "helpfully" indented things when I edited them. Should be fixed now.

language/functions.xml
cronlabspl: > Maybe better (I don't like screaming):
> 
> ```suggestion
> 
>       example above) is <emphasis>deprecated</emphasis>.
> 
> ```

Now that's better😀

cmb69:
It might be good to wait on the outcome of https://externals.io/message/116597#116643 before documenting the current behavior.

cronlabspl:
@IMSoP Im perfectly OK with it, but dont have `write-access` here.........

cmb69:
Thank you!

PR #1193, opened 2021-12-10
PHP 8.1: FILTER_SANITIZE_STRING is deprecated

kamil-tekiela:


reference/filter/constants.xml
```diff
@@ -281,6 +281,8 @@
    <listitem>
     <simpara>
      ID of "string" filter.
+     (<emphasis>DEPRECATED</emphasis> as of PHP 8.1.0,
```
cmb69, line 284: I dislike that SCREAMING and would suggest lower-case spelling (even though it's used elsewhere on that page).

kamil-tekiela: I can change it, but that is how we do it in many places in the manual e.g. https://www.php.net/manual/en/function.get-magic-quotes-gpc.php

I agree the screaming is probably not necessary, just following the same style.

cmb69 (approved):
Thank you!

PR #1196, opened 2021-12-11
Fixed info about @ operator

dg:
Corrects some misunderstandings of the UPGRADING document

- It says that the error_reporting (ie. value returned by `error_reporting()` or `ini_get('error_reporting')`) is changed, not the severity value passed to the handler.

- It says that the `error_reporting() == 0` should be replaced with `!(error_reporting() & $err_no)` in code, not the return with return false. So there is currenctly no way how to detect @ operator.

- "Prior to PHP 8.0.0, the value of the severity passed to the custom error handler was always 0" - no true

- "`return false; // Silenced`" - not true, if the function returns false then the normal error handler continues, so it is not "silenced".

(partially reverts "Cleanup and update docs related to @ operator" commit 12b1d4704ea448b55ec155619be1e50c0e67e9f9.)


appendices/migration80/incompatible.xml
```diff
@@ -200,15 +200,15 @@
 // Replace
 function my_error_handler($err_no, $err_msg, $filename, $linenum) {
     if (error_reporting() == 0) {
-        return false; // Silenced
+        return; // Silenced
```
Girgias, line 203: Please use ``return true;`` here.

We document the error handler to be a callable which returns a bool in the docs or ``set_error_handler()``, although the actual semantics are kinda batshit crazy I'm not super keen on providing contradicting information.

language/operators.xml
```diff
@@ -1946,34 +1946,9 @@ echo $foo ?? $bar ?? $baz ?? $qux; // outputs 1
...
-    if (!(error_reporting() & $err_no)) {
-        return false; // Silenced
-    }
-    // ...
-}
-?>
-]]>
-     </programlisting>
-    </informalexample>
+    the diagnostic has been suppressed.
```
Girgias, line 1949: Why are you removing this section?

Just fix the return value in the custom error handler.

Girgias (changes requested):

language/operators.xml
dg: Because it is not possible to know that the @ operator has been used. Since version 8.0, the `error_reporting()` no longer returns 0 when `@` is used. This is pointed out in UPGRADING that a test for zero is no longer possible and must be modified. 

Neither now nor before could the operator be reliably recognized or distinguished from the own error_reporting setting.

appendices/migration80/incompatible.xml
dg: Oh, that's an mistake in the documentation, because the callable must have a return value of ?bool.

See return value explanation, null is important value https://www.php.net/manual/en/function.set-error-handler.php#refsect1-function.set-error-handler-returnvalues

Girgias: No the return value can be literally anything: https://3v4l.org/UeNn4 but only ``false`` will cause PHP to run the built-in error_handler.
Moreover if the function has a return type of ``?bool`` you would still need to use ``return null;``.

And I'm not sure it is really sensible trying to explain the very weird semantics of how the return value is checked. Others may have another opinion on this.

language/operators.xml
Girgias: I'm pretty sure this needs to be kept, as a suppressed/silenced diagnostic if it's ignored in the INI error_reporting will still trigger the error handler even though it should *probably* not do anything. See: https://3v4l.org/0nMbF

This might need rephrasing of the section, but I don't think deleting this make a ton of sense.

appendices/migration80/incompatible.xml
dg: My point is that `return false` is followed by `silenced`, which is confusing. It doesn't really matter what is returned there, but just calling the built-in handler can't be called *silencing*.

language/operators.xml
dg: How can you detect that @ was used? See https://3v4l.org/4WBrj and https://3v4l.org/paLuC

appendices/migration80/incompatible.xml
dg: Actually, I have to say that it's actually better to `return false;` than to `return; // Silenced`, because it populates the values in error_get_last()

dg:
Please, what prevents merging? The documentation doesn't tell the truth and it confuses the user unnecessarily.

language/operators.xml
cmb69: The PHP manual currently is supposed to document PHP 7.0 to 8.1. If there are behavioral differences, this can be indicated with "as of PHP 8.0.0" or "prior to PHP 8.0.0".

Crell:
See @cmb69's comments on 14 January. That still needs to be addressed.

dg:
@Crell done

language/operators.xml
```diff
@@ -1946,31 +1946,14 @@ echo $foo ?? $bar ?? $baz ?? $qux; // outputs 1
...
    </para>
 
    <warning>
     <para>
-     Prior to PHP 8.0.0, the value of the severity passed to the custom error
-     handler was always <literal>0</literal> if the diagnostic was suppressed.
-     This is no longer the case as of PHP 8.0.0.
+     Prior to PHP 8.0.0, the <function>error_report</function> called inside the custom error handler 
+     returned always <literal>0</literal> if the error was suppressed by the <literal>@</literal> operator.
```
Crell, line 1955:
```suggestion
     always returned <literal>0</literal> if the error was suppressed by the <literal>@</literal> operator. 
```

language/operators.xml
```diff
@@ -1946,31 +1946,14 @@ echo $foo ?? $bar ?? $baz ?? $qux; // outputs 1
...
    </para>
 
    <warning>
     <para>
-     Prior to PHP 8.0.0, the value of the severity passed to the custom error
-     handler was always <literal>0</literal> if the diagnostic was suppressed.
-     This is no longer the case as of PHP 8.0.0.
+     Prior to PHP 8.0.0, the <function>error_report</function> called inside the custom error handler 
+     returned always <literal>0</literal> if the error was suppressed by the <literal>@</literal> operator. 
+     Since version 8, it returns the value of <literal>E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE</literal>.
```
Girgias, line 1956:
```suggestion
     As of PHP 8.0., it returns the value <literal>E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE</literal>.
```

language/operators.xml
```diff
@@ -1946,31 +1946,14 @@ echo $foo ?? $bar ?? $baz ?? $qux; // outputs 1
...
    </para>
 
    <warning>
     <para>
-     Prior to PHP 8.0.0, the value of the severity passed to the custom error
-     handler was always <literal>0</literal> if the diagnostic was suppressed.
-     This is no longer the case as of PHP 8.0.0.
+     Prior to PHP 8.0.0, the <function>error_report</function> called inside the custom error handler 
+     always returned <literal>0</literal> if the error was suppressed by the <literal>@</literal> operator. 
+     As of PHP 8.0., it returns the value <literal>E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE</literal>.
```
Girgias, line 1956:
```suggestion
     As of PHP 8.0.0, it returns the value <literal>E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE</literal>.
```

dg:
Thanks for the merge.

PR #1197, opened 2021-12-11
Adding rules for cases

ThomasLandauer:
I basically guessed this - so please doublecheck.
Is there a better specification somewhere that could be linked to?


cmb69:
Thank you for the PR!

The names of cases are labels (aka. identifiers), and in this case are actually syntactically identical to (class) constants. https://www.php.net/manual/en/language.constants.php already documents valid constant names. Maybe just refer to that?

ThomasLandauer:
Perfect, that's exactly what I was looking for! :-)

cmb69:
Thank you!

PR #1198, opened 2021-12-12
Link to spl_object_id from similar functions.

TysonAndre:
Encourage the use of spl_object_id instead.

Using an integer is faster and more memory-efficient than creating a 32-byte
string hash.

spl_object_hash historically obfuscated the object id with xor for security reasons, but it was sufficient for security to not expose raw pointers.

`runkit7_object_id` is one way to polyfill PHP 7.2+ `spl_object_id`
in older PHP 7 versions and continues to exist in the PECL.


cmb69:
Thank you!

PR #1202, opened 2021-12-14
Errors/Exceptions missing in supplementary document.

kayw-geek:


cmb69:
Thank you for the PR!

Given that there are other cases of failure which also raise a warning (missing permissions, and more, like open_basedir violations), I'm not sure if it wouldn't be better to keep the info generic (like for filemtime): "Upon failure, an E_WARNING is emitted."

kayw-geek:
> Thank you for the PR!
> 
> Given that there are other cases of failure which also raise a warning (missing permissions, and more, like open_basedir violations), I'm not sure if it wouldn't be better to keep the info generic (like for filemtime): "Upon failure, an E_WARNING is emitted."

Thank you for your reply. I didn't consider other situations. I have changed it to general description.

kayw-geek:
Sorry, I accidentally closed this PR, new PR https://github.com/php/doc-en/pull/1203

PR #1208, opened 2021-12-16
Fix a few ext/reflection return types

kocsismate:


reference/reflection/reflectionclassconstant/getname.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type class="union"><type>string</type><type>false</type></type><methodname>ReflectionClassConstant::getName</methodname>
+   <modifier>public</modifier> <type>string</type><methodname>ReflectionClassConstant::getName</methodname>
```
cmb69, line 12: According to the stubs, this is only true as of PHP 8.1.0. However, the implementation (`_default_get_name()`) might still return `false` in PHP-8.1, or is this just a case that never can happen in practice (if so, an assertion might be more suitable). Argh, heap.space is not up-to-date. So, yeah, as of PHP 8.1.0 the method throws instead of returning `false` => changelog entry.

cmb69 (commented):
Thank you!

reference/reflection/reflectionclassconstant/getname.xml
kocsismate: Thanks for the finding! I tried to come up with a descriptive change log entry, but feel free to suggest a better one :)

cmb69: Thanks! The changelog entry looks good to me, especially given that this issue likely only [happens with pathological code](https://3v4l.org/nlDV1).

cmb69 (approved):

Issue #1206, opened 2021-12-16
Error handler callback has return type bool but allows others


othercorey:
https://www.php.net/manual/en/function.set-error-handler.php

The `handler` callback passed to `set_error_handler` is documented with return type `bool`. The description for the callback includes:

> If the function returns false then the normal error handler continues.

However, this function can return more than just boolean values. It seems to ignore anything returned except `false`.

I'm not sure this is an actual bug, but I wanted to clarify that it can be confusing when discussing complicated error handler flows where some developers returned boolean values, some returned nothing and so on.

Will the callback ever error on non-boolean values? Should we explain that any other value has no effect?

damianwadley:
The behavior is literally what it says there: if it returns `false`, not 0 or an empty string or anything _loosely_ false, then the normal error handler is invoked. PHP doesn't care whether it returns any other value or even no value at all.

```php
if ($callback($errno, $errstr, $errfile, $errline) === false) {
    normal_error_handler($errno, $errstr, $errfile, $errline);
}
```

othercorey:
You've simply restated my point.

damianwadley:
I'll try again with some quotes:

> I'm not sure this is an actual bug,

It's not.

> it can be confusing when discussing complicated error handler flows where some developers returned boolean values, some returned nothing and so on

What's confusing about "if it returns false then the normal error handler continues"? If you return false then it does that, and if you don't return false then it doesn't do that.

> Will the callback ever error on non-boolean values?

"Ever"? Maybe something will happen in PHP 9 or 10 or 15. But for now, no: all it cares about is false.

> Should we explain that any other value has no effect?

I thought it was implied, but I also haven't closed this issue.

othercorey:
Why are you being so aggressive?

I'm asking about a simple documentation improvement for something that is a little contradictory.

ADmad:
Given that PHP now has stricter typing, one wouldn't be at fault thinking a `bool` means the function can/should only return boolean values. While the description provides future details the `: bool` for the callback is pretty misleading.

ADmad:
> PHP doesn't care whether it returns any other value or even no value at all.

Aren't return types meant to indicate what a function can return? What PHP does with the return value isn't the issue here. Having `bool` return type is misleading since the callback can return a non-boolean value and PHP will act accordingly.

ADmad:
Perhaps in a future minor returning any other value besides `false` from the callback should be deprecated.

cmb69:
[...]

PR #1211, opened 2021-12-17
Add note on return value coercion in error handler

othercorey:
Refs https://github.com/php/doc-en/issues/1206

@cmb69 Maybe this will help.

I don't know the standard wording for type declarations and types in the documentation.

### DIFF
--- a/reference/errorfunc/functions/set-error-handler.xml
+++ b/reference/errorfunc/functions/set-error-handler.xml
@@ -148,7 +148,7 @@
        </variablelist>
       </para>
       <para>
-       If the function returns &false; then the normal error handler continues. The return value is not coerced to &bool;
+       If the function returns &false; then the normal error handler continues. The return value is not coerced to &boolean;
        if the return type declaration is missing.
       </para>
      </listitem>
--- a/reference/errorfunc/functions/set-error-handler.xml
+++ b/reference/errorfunc/functions/set-error-handler.xml
@@ -148,7 +148,8 @@
        </variablelist>
       </para>
       <para>
-       If the function returns &false; then the normal error handler continues.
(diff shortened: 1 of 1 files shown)


reference/errorfunc/functions/set-error-handler.xml
```diff
@@ -148,7 +148,8 @@
        </variablelist>
       </para>
       <para>
-       If the function returns &false; then the normal error handler continues.
+       If the function returns &false; then the normal error handler continues. The return value is not coerced to &bool;
```
saundefined, line 151: to fix the build:

```suggestion
       If the function returns &false; then the normal error handler continues. The return value is not coerced to &boolean;
```

othercorey: Thank you.

othercorey:
Closing due to no interest.

cmb69:
> Closing due to no interest.

There are still 75 open pull requests, many of which are lying around for much longer. This is not because there is no interest, but rather due to lack of time. :)

Anyhow, I think this patch wouldn't really cut it. There is a more general issue, namely that this callback (and others for other functions) do not have to use any type declarations at all, or might even have mismatching type declarations (https://3v4l.org/SochO#v7313; I'm slightly confused regarding the PHP 8.1 behavior), and we should probably document that in a central place, or at least use more general wording.

othercorey:
> > Closing due to no interest.
> 
> There are still 75 open pull requests, many of which are lying around for much longer. This is not because there is no interest, but rather due to lack of time. :)

No worries. That was just meant to comment the closure not criticize it not being merged :). I noticed most disapproved PRs just stay open in php repos so that's all.

PR #1212, opened 2021-12-17
Update fetchAll() return description to match summary wording

othercorey:
It either returns all or remaining rows. These aren't exactly the same operation which can be confusing.


cmb69 (changes requested):
Well, it actually fetches the remaining rows only, see https://3v4l.org/Lq95R.

So better change the refpurpose to "Fetches remaining rows from a result set".

othercorey:
You're right. I was mislead by a comment in the source and didn't test further.

cmb69:
Thank you!

PR #1215, opened 2021-12-17
Update ext/hash documentation

kocsismate:


cmb69:
Thank you! I've added changelog entries for the false returns, and version.xml entries for __(un)serialize().

PR #1217, opened 2021-12-18
Remove &warn.undocumented.func; from Reflection

kamil-tekiela:
Removed from all the places where it wasn't stricly necessary.


cmb69:
Thank you!

PR #1219, opened 2021-12-19
Update documentation for PHP8

comp64:
Length parameter to substr() behaves differently for PHP8+. It recognizes null as if no value was provided at all. Updated the documentation to reflect this.

### DIFF
--- a/reference/strings/functions/substr.xml
+++ b/reference/strings/functions/substr.xml
@@ -87,8 +87,8 @@ $rest = substr("abcdef", -3, 1); // returns "d"
        beyond, an empty string will be returned.
       </para>
       <para>
-       If <parameter>length</parameter> is given and is <literal>0</literal> or
-       &false;, an empty string will be returned.
+       If <parameter>length</parameter> is given and is <literal>0</literal>,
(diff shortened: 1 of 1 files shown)


reference/strings/functions/substr.xml
```diff
@@ -87,8 +87,8 @@ $rest = substr("abcdef", -3, 1); // returns "d"
        beyond, an empty string will be returned.
       </para>
       <para>
-       If <parameter>length</parameter> is given and is <literal>0</literal>,
-       &false; or &null;, an empty string will be returned.
+       If <parameter>length</parameter> is given and is <literal>0</literal> or
+       &false;, an empty string will be returned.
```
cmb69, line 91: Please remove that `&false;` as well, since that is not special to this function, and would [apply for coercive typing mode only](https://3v4l.org/viJ2T):
```suggestion
       If <parameter>length</parameter> is given and is <literal>0</literal>,
       an empty string will be returned.
```

cmb69 (changes requested):
Thank you for the PR!

PR #1221, opened 2021-12-21
Update the min version of curl required by PHP

stof:
PHP 7.3 increased the min version of curl

### DIFF
--- a/reference/curl/setup.xml
+++ b/reference/curl/setup.xml
@@ -11,6 +11,8 @@
    In order to use PHP's cURL functions you need to install the <link
    xlink:href="&url.curl;">libcurl</link> package. PHP requires libcurl
    version 7.10.5 or later.
+   As of PHP 7.3.0, version 7.15.5 or later is required.
+   As of PHP 8.0.0, version 7.29.0 or later is required.
   </para>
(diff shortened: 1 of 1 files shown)


reference/curl/setup.xml
```diff
@@ -10,7 +10,7 @@
   <para>
    In order to use PHP's cURL functions you need to install the <link
    xlink:href="&url.curl;">libcurl</link> package. PHP requires libcurl
-   version 7.10.5 or later.
+   version 7.15.5 or later.
```
cmb69, line 13: The manual is currently supposed to document PHP 7.0 - 8.1, so we need to keep the information for the old versions. Also, as of PHP 8.0.0, curl ≥ 7.29.0 is required so maybe:
```suggestion
   version 7.10.5 or later.
   As of PHP 7.3.0, version 7.15.5 or later is required.
   As of PHP 8.0.0, version 7.29.0 or later is required.
```

cmb69 (changes requested):
Thank you for the PR!

stof:
@cmb69 updated

cmb69:
Thank you!

PR #1224, opened 2021-12-21
Generate ext/spl class synopses from stubs - part 3

kocsismate:


reference/spl/regexiterator.xml
```diff
@@ -165,6 +172,20 @@
...
 
+  <section xml:id="regexiterator.props">
+   &reftitle.properties;
+   <variablelist>
+    <varlistentry xml:id="regexiterator.props.replacement">
+     <term>
+      <varname>replacement</varname>
+     </term>
+     <listitem>
+      <para></para>
```
kocsismate, line 183: what should be the description?

reference/spl/runtimeexception.xml
```diff
@@ -25,31 +20,30 @@
...
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.exception')/db:partintro/db:section/db:classsynopsis/db:fieldsynopsis[preceding-sibling::db:classsynopsisinfo[1][@role='comment' and text()='&Properties;']]))">
+     <xi:fallback/>
+    </xi:include>
 
-<!-- Comment this section out if the exception does not contain any methods --><!--
-    <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.runtimeexception')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[1])" />
--->
-   <classsynopsisinfo role="comment">&InheritedMethods;</classsynopsisinfo>
-    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.exception')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[1])"><xi:fallback /></xi:include>
```
kocsismate, line 51: inherited constructors are usually not listed here as far as I remember. Should we consistently do this, or is it ok to get rid of this one?

reference/spl/regexiterator.xml
cmb69: From what I can tell, the replacement is only relavant for `RegexIterator::REPLACE` for which the documentation states:

> Replace the current entry (see preg_replace(); Not fully implemented yet)

So maybe leave the description empty for now.

reference/spl/runtimeexception.xml
cmb69: It's probably okay to just drop this one for now.

reference/spl/splfixedarray.xml
```diff
@@ -24,36 +24,68 @@
...
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.1.0</entry>
+       <entry>
+        <classname>SplFixedArray</classname> implements
+        <interfacename>JsonSerializable</interfacename> and
+        <interfacename>IteratorAggregate</interfacename> now.
+        <interfacename>Iterator</interfacename> is no longer implemented.
```
cmb69, line 81: Maybe:
```suggestion
        Previously, only <interfacename>Iterator</interfacename> was implemented instead.
```

cmb69 (commented):
Thank you!

PR #1227, opened 2021-12-21
Add undocumented FFI\CType methods

kocsismate:
Currently, the FFI\CType class page warns due to the missing documentation of its methods.

Pages are generated by docgen.php

### DIFF
--- a/reference/ffi/ctype/getalignment.xml
+++ b/reference/ffi/ctype/getalignment.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="ffi-ctype.getalignment" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>FFI\CType::getAlignment</refname>
+  <refpurpose>Description</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <modifier>public</modifier> <type>int</type><methodname>FFI\CType::getAlignment</methodname>
+   <void />
+  </methodsynopsis>
+  <para>
+
+  </para>
+
+  &warn.undocumented.func;
+
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  &no.function.parameters;
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   
+  </para>
+ </refsect1>
+
+
+</refentry>
+<!-- Keep this comment at the end of the file
+Local variables:
+mode: sgml
+sgml-omittag:t
+sgml-shorttag:t
+sgml-minimize-attributes:nil
+sgml-always-quote-attributes:t
+sgml-indent-step:1
+sgml-indent-data:t
+indent-tabs-mode:nil
+sgml-parent-document:nil
+sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+sgml-exposed-tags:nil
+sgml-local-catalogs:nil
+sgml-local-ecat-files:nil
+End:
+vim600: syn=xml fen fdm=syntax fdl=2 si
+vim: et tw=78 syn=sgml
+vi: ts=1 sw=1
+-->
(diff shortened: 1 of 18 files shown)


cmb69:
Oh, indeed, documenting https://github.com/php/php-src/pull/7217 has apparently been forgotten so far. I think we should note that in the migration guide as well, and maybe someone who has already been working with these new APIs might want to fill in some of documentation gaps right away. I'm going to add a help-wanted label, and suggest to wait a while; otherwise committing as is should be okay.

mumumu:
This PR will fix the following build warning.

https://github.com/php/doc-en/commit/45aaa3f3bd6e007e2106af928697de57a98bf991

Fizie92 (commented):
@ #

cmb69:
I just noticed that the FFI\CType reflection API also introduced [a bunch of class constants](https://github.com/php/php-src/blob/master/ext/ffi/ffi.c#L5355-L5396). Apparently, we don't have class constants anywhere in the stub files; is that deliberate? If there is no particular reason to *not* have them there, I suggest to add them for documentation purposes; although, of course, someone would have to implement that. And also, the transition for the docs might be quite some work, since (most?) exisiting documentation puts the class constants in the predefined constants sections, e.g. https://www.php.net/manual/en/zip.constants.php.

kocsismate:
> Apparently, we don't have class constants anywhere in the stub files; is that deliberate?

Yes, unfortunately, this is due to a technical limitation: we can't autogenerate constants which have a C constant value. However, I made a POC in order to add simple constant autogeneration support in stubs: https://github.com/php/php-src/pull/7434

As Nikita wasn't really impressed, I put this PR on hold. However, I think extensions could benefit from this feature, and I agree that it would be advantageous for class synopsis generation as well.

> And also, the transition for the docs might be quite some work, since (most?) exisiting documentation puts the class constants in the predefined constants sections

Oh, I wasn't aware of this. Thanks for bringing this to my attention!

cmb69:
Thanks for pointing out that PR! Actually, I'm not in favor of documenting all constants with their value; while it may make sense for some (e.g. error constants, where one wants to quickly look up what the result of `json_last_error()` means), it easily may defeat the purpose of abstracting over certain hard-coded values. Still, having the defined constants documented without their values is valuable (and it won't take long for a user note to appear which lists the constants with their values).

And note that I'm actually in favor of moving class constants from the predefined constants sections to the classs synopses, although I'm not sure whether this is currently supported by PhD.

kocsismate:
> Still, having the defined constants documented without their values is valuable (and it won't take long for a user note to appear which lists the constants with their values).

I would be more happy if constants in stubs could be used/validated by php-src itself in any way. Otherwise we'll likely end up with outdated information in a while. That's why my PR only adds "simple" constant (~ constants whose value is a literal rather than a C constant) support so that we don't have to refer to C constants in the stubs, but we can still benefit from a more complete class entry and class synopsis page autogeneration. E.g. JSON error codes (`php_json_error_code`) would be supported, while the `LIBXSLT_VERSION` wouldn't.

> And note that I'm actually in favor of moving class constants from the predefined constants sections to the class synopses, although I'm not sure whether this is currently supported by PhD.

`fieldsynopsis` elements can be used to display class constants. For instance: https://www.php.net/manual/en/class.datetimeinterface.php But the migration from the predefined constants sections to the class synopses for sure would be a PITA.

cmb69:
> I would be more happy if constants in stubs could be used/validated by php-src itself in any way. Otherwise we'll likely end up with outdated information in a while. That's why my PR only adds "simple" constant (~ constants whose value is a literal rather than a C constant) support so that we don't have to refer to C constants in the stubs, but we can still benefit from a more complete class entry and class synopsis page autogeneration. E.g. JSON error codes (php_json_error_code) would be supported, while the LIBXSLT_VERSION wouldn't.

One does not necessarily exclude the other. In the stubs, constants that refer to C constants could have an UNKNOWN value, and for the docs we could still decide whether we want to display the values or not.

> fieldsynopsis elements can be used to display class constants.

Ah, right. Thanks!

> But the migration from the predefined constants sections to the class synopses for sure would be a PITA.

Well, in my opinion, that should be done anyway. The only drawback generating this from a stub file would be that [grouping](https://www.php.net/manual/en/imagick.constants.php) would not be supported (at least not out of the box).

kocsismate:
> One does not necessarily exclude the other. In the stubs, constants that refer to C constants could have an UNKNOWN value, and for the docs we could still decide whether we want to display the values or not.

I like this idea! :) Believe it or not, there are only a handful of class synopses left which still needs synchronization :) And the rest of the diff is due to the missing constants. So adding support for constants would improve the situation of class synopsis generation much more.

> Well, in my opinion, that should be done anyway. The only drawback generating this from a stub file would be that grouping would not be supported (at least not out of the box).

Yeah, I think we could add support for grouping by using `@constant-group` in the stubs :)

kocsismate:
How much should we wait with the merge? Can we go ahead now? I doubt there will be any volunteers to help with this PR :S

cmb69 (approved):
Ah, right! Please go ahead.

PR #1238, opened 2021-12-22
date_default_timezone_get: unmention the warning

jtojnar:
It has not been the case since PHP 7.0:
https://github.com/php/php-src/commit/b22caa81ee2797572279fc9aff0918c01f708de5

### DIFF
--- a/reference/datetime/functions/date-default-timezone-get.xml
+++ b/reference/datetime/functions/date-default-timezone-get.xml
@@ -28,18 +28,15 @@
       (if set)
      </para>
     </listitem>
-    <listitem>
-     <para>
-      A warning is shown when this stage is reached. Do not rely on it to be guessed
-      correctly, and set <link linkend="ini.date.timezone">date.timezone</link> to the correct timezone
-      instead.
-     </para>
-    </listitem>
    </itemizedlist>
   </para>
   <para>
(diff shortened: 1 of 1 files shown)


reference/datetime/functions/date-default-timezone-get.xml
```diff
@@ -28,19 +28,16 @@
...
    </itemizedlist>
   </para>
   <para>
    If none of the above succeed, <methodname>date_default_timezone_get</methodname>
    will return a default timezone of <literal>UTC</literal>.
   </para>
+  <para>
+   To make sure the return value matches reality, explicitly set <link linkend="ini.date.timezone">date.timezone</link>
+   to the correct timezone instead of relying on the guessing.
+  </para>
```
Girgias, line 40: I think this can just be dropped altogether, as no guessing is performed anymore

jtojnar: I think it is still valuable to emphasize that the value is useless unless set explicitly. Maybe something like “Note that `date.timezone` does **not** default to the timezone of the operating system so you need to set the option explicitly to have it match reality.”

cmb69: Yeah, but why document that for `date_default_timezone_get()`?

PR #1241, opened 2021-12-23
Generate various class synopses from stubs - part 1

kocsismate:


reference/simplexml/simplexmlelement.xml
```diff
@@ -20,34 +20,68 @@
...
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <classname>SimpleXMLElement</classname> implements
+        <interfacename>Stringable</interfacename> now.
```
cmb69, line 72: It seems to me that `Countable` and `RecursiveIterator` are also only implemented as of PHP 8.0.0. Apparently, these interfaces have been implemented by `SimpleXMLIterator` prior to that version.

cmb69 (commented):
Thank you! Except for the SimpleXMLElement/SimpleXMLIterator interface implementations, this looks good to me.

cmb69:
I think that SimpleXMLIterator needs a changelog entry as well.

kocsismate:
> I think that SimpleXMLIterator needs a changelog entry as well.

SimpleXMLIterator still (indirectly) implements these interfaces, so IMO there's not much change from a child class POV. Or what do you think the change log entry should mention? Although, I'vejustrealized that the newly implemented Stringable interface could be mentioned.

cmb69 (approved):
Ah, right SimpleXMLIterator extends SimpleXMLElement, so everything is fine, in my opinion.

PR #1243, opened 2021-12-23
Generate ext/spl class synopses from stubs - part 7

kocsismate:


cmb69 (approved):
Thank you!

If I may say so, `E_TOO_MANY_ITERATORS`. ;)

kocsismate:
> If I may say so, E_TOO_MANY_ITERATORS. ;)

:D Yeah, I'm also fed up with them... However, good news is that they are almost finished! :) Only 1-2 remained IIRC.

PR #1245, opened 2021-12-23
Generate Zend class synopses from stubs - part 2

kocsismate:
Mostly formatting changes again

### DIFF
--- a/language/predefined/throwable.xml
+++ b/language/predefined/throwable.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-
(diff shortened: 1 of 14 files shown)


language/predefined/throwable.xml
```diff
@@ -31,18 +30,30 @@
...
-<!-- {{{ Class synopsis -->
     <classsynopsisinfo>
      <oointerface>
       <interfacename>Throwable</interfacename>
      </oointerface>
+
+     <ooclass>
+      <modifier>extends</modifier>
+      <classname>Stringable</classname>
+     </ooclass>
```
cmb69, line 45: I think we need a changelog entry for this.

cmb69 (approved):
Thank you!

Looks like Santa brings lots of gifts for translators this year. :)

language/predefined/throwable.xml
kocsismate: Ah yes, I forgot about it

PR #1249, opened 2021-12-24
Generate ext/intl class synopses from stubs - part 2

kocsismate:


reference/intl/intlbreakiterator.xml
```diff
@@ -38,19 +37,20 @@
...
 
-<!-- {{{ Class synopsis -->
     <classsynopsisinfo>
      <ooclass>
       <classname>IntlBreakIterator</classname>
      </ooclass>
 
      <oointerface>
-      <interfacename>Traversable</interfacename>
+      <interfacename>IteratorAggregate</interfacename>
```
cmb69, line 50: This is only as of PHP 8.0.0, so needs a changelog entry. Also, the introduction section above explicitly mentions Traversable, what should probably be changed right away.

cmb69 (commented):
Thank you! Looks good besides the IntlBreakIterator issues mentioned below.

reference/intl/intlbreakiterator.xml
kocsismate: Ah, yes, I didn't notice these ones!

cmb69 (approved):
Thank you!

PR #1250, opened 2021-12-24
Generate ext/dom class synopses from stubs - part 1

kocsismate:


reference/dom/domexception.xml
```diff
@@ -36,46 +32,38 @@ FIXME: Remove me once you perform substitutions
...
-<!-- If the property is documented below (xml:id=domexception.props) use this -->
     <fieldsynopsis>
      <modifier>public</modifier>
-     <modifier>readonly</modifier>
      <type>int</type>
      <varname linkend="domexception.props.code">code</varname>
     </fieldsynopsis>
 
     <classsynopsisinfo role="comment">&InheritedProperties;</classsynopsisinfo>
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.exception')/db:partintro/db:section/db:classsynopsis/db:fieldsynopsis[preceding-sibling::db:classsynopsisinfo[1][@role='comment' and text()='&Properties;']]))">
```
cmb69, line 59: It is unfortunate that we list `$code` as property, and then again as inherited property. Probably, `$code` should not have been made pulic for `DOMException`, even though the DOM standard mandates it (although it should be readonly, what it is not).

cmb69 (approved):
Thank you!

PR #1251, opened 2021-12-25
Generate various class synopses from stubs - part 2

kocsismate:


Fizie92 (commented):
# @

language/predefined/fibererror/construct.xml
```diff
@@ -0,0 +1,44 @@
...
+
+  </para>
+
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  &no.function.parameters;
+ </refsect1>
+</refentry>
```
cmb69, line 24: Maybe
```suggestion
 <refsect1 role="errors">
  &reftitle.errors;
  <para>
   Throws an <classname>Error</classname> exception when called.
  </para>
 </refsect1>
</refentry>
```

reference/dir/directory.xml
```diff
@@ -24,30 +22,34 @@
...
       <classname>Directory</classname>
      </ooclass>
     </classsynopsisinfo>
-<!-- }}} -->
-    
+
     <classsynopsisinfo role="comment">&Properties;</classsynopsisinfo>
     <fieldsynopsis>
      <modifier>public</modifier>
+     <modifier>readonly</modifier>
```
cmb69, line 38: The `path` is [readonly only as of PHP 8.1.0](https://3v4l.org/cC3AN), so there should be a changelog entry. I assume it's the same with the `handle` property.

cmb69 (commented):
Thank you!

cmb69 (approved):
Thank you!

PR #1253, opened 2021-12-25
[PHP 8.1] mysqli::init() is deprecated

kamil-tekiela:
https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.mysqli


cmb69 (approved):
Thank you!

PR #1257, opened 2021-12-26
Remove unnecessary DateTimeImmutable method synopsis roles

kocsismate:
The DateTimeImmutable class page doesn't depend on these roles. Followup to https://github.com/php/doc-en/pull/1011


cmb69 (approved):
Thank you!

(could mark as [skip-revcheck])

PR #1258, opened 2021-12-26
Remove unnecessary DateInterval, DatePeriod, DateTime and DateTimeZone method synopsis roles

kocsismate:
These roles are not necessary, since the respective class pages don't depend on them. Followup to #1011


cmb69 (approved):
Thank you!

(maybe mark as [skip-revcheck])

PR #1261, opened 2021-12-26
Regenerate ext/phar method synopses based on stubs

kocsismate:
Last one for today. 🎄

### DIFF
--- a/reference/phar/Phar/offsetGet.xml
+++ b/reference/phar/Phar/offsetGet.xml
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>PharFileInfo</type><methodname>Phar::offsetGet</methodname>
+   <modifier>public</modifier> <type>SplFileInfo</type><methodname>Phar::offsetGet</methodname>
    <methodparam><type>string</type><parameter>localName</parameter></methodparam>
   </methodsynopsis>
 
--- a/reference/phar/Phar/offsetUnset.xml
+++ b/reference/phar/Phar/offsetUnset.xml
@@ -43,7 +43,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.success;
+   &return.void;
   </para>
  </refsect1>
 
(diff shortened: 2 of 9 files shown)


reference/phar/Phar/offsetGet.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>PharFileInfo</type><methodname>Phar::offsetGet</methodname>
+   <modifier>public</modifier> <type>SplFileInfo</type><methodname>Phar::offsetGet</methodname>
```
cmb69, line 11: Ah, [right](https://github.com/php/php-src/commit/3fc3cfbad01a1046fc13ac864fa82bc83b34efcd)!

reference/phar/Phar/offsetUnset.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>bool</type><methodname>Phar::offsetUnset</methodname>
+   <modifier>public</modifier> <type>void</type><methodname>Phar::offsetUnset</methodname>
```
cmb69, line 11: This is only the case as of PHP 8.1.0. I'm not sure whether we should have a changelog for this, though, since returning bool is an LSP violation (implements ArrayAccess), so nobody should rely on that. We should change the return values section, though, which currently states `&return.success;`.

reference/phar/PharData/offsetUnset.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>bool</type><methodname>PharData::offsetUnset</methodname>
+   <modifier>public</modifier> <type>void</type><methodname>PharData::offsetUnset</methodname>
```
cmb69, line 11: Same as with `Phar::offsetUnset()`.

cmb69 (commented):
Thank you!

cmb69 (approved):
Thank you!

Issue #658, opened 2021-06-04
Fix section order issues for all files


Girgias:
These issues can be found by using the [`section-order.php` QA script](https://github.com/php/doc-base/blob/master/scripts/qa/section-order.php) located in doc-base,
and should be fixed similarly to how it is done in #657

mumumu:
```
$ php doc-base/scripts/qa/section-order.php | awk '{print $2}'| awk -F '/' '{print $8}'|sort | uniq -c | awk '{print "- [ ] " $2 " ("$1 ")"}' | tail -n +2
```
* TODO
  - [ ] apache (6)
  - [ ] apcu (1)
  - [ ] array (1)
  - [ ] bc (2)
  - [ ] classobj (5)
  - [ ] com (1)
  - [ ] cubrid (1)
  - [ ] datetime (34)
  - [ ] dba (1)
  - [ ] dir (8)
  - [ ] dom (9)
  - [ ] eio (1)
  - [ ] errorfunc (5)
  - [ ] ev (1)
  - [ ] event (7)
  - [ ] exec (1)
  - [ ] exif (1)
  - [ ] fdf (3)
  - [ ] filesystem (43)
  - [ ] filter (1)
  - [ ] fpm (1)
  - [ ] funchand (2)
  - [ ] geoip (1)
  - [ ] gmagick (7)
  - [ ] hash (2)
  - [ ] ibase (2)
  - [ ] image (4)
  - [ ] imagick (126)
  - [ ] imap (3)
  - [ ] info (21)
  - [ ] inotify (1)
  - [ ] intl (3)
  - [ ] ldap (2)
  - [ ] libxml (3)
  - [ ] lzf (1)
  - [ ] mailparse (1)
  - [ ] math (4)
  - [ ] mbstring (7)
  - [ ] mcrypt (11)
  - [ ] memcache (8)
  - [ ] memcached (3)
  - [ ] mhash (1)
  - [ ] misc (6)
  - [ ] mongodb (28)
  - [ ] mysql (1)
  - [ ] mysqli (21)
  - [ ] network (5)
  - [ ] oci8 (18)
  - [ ] opcache (1)
  - [ ] openal (2)
  - [ ] openssl (1)
  - [ ] outcontrol (12)
  - [ ] parallel (4)
  - [ ] pcntl (2)
  - [ ] pcre (2)
  - [ ] pdo (16)
  - [ ] pdo_pgsql (1)
  - [ ] pgsql (3)
  - [ ] phar (19)
[...]

PR #1262, opened 2021-12-27
Fix section order issues

saundefined:
Relates #658


Girgias:
I wonder if for destructors we should do the same as constructors and drop the return value section, would need to amend the script for that.

Wdyt @cmb69 ?

cmb69:
I think we should omit the return values section for destructors; if we adapt docgen for that, we should generate `<destructorsynopsis>` instead of `<methodsynopsis>`.

PR #1265, opened 2021-12-27
Drop Windows Notepad usage instructions

cmb69:
We should not encourage use of Notepad for editing PHP files, and those
who use it, should at least know how it works (after all, it's not that
complex).


kamil-tekiela (approved):
Agreed. It's condescending and unnecessary. I can edit a file with any extension using PHP interpreter. I think this note only applies to the context of Apache.

PR #1266, opened 2021-12-27
Explain in "Hello World" e.g. that HTML tags are not needed

ahinkle:
This PR explains that HTML attribute tags are not needed to create a baseline PHP "Hello World" example.

Inspired from[ this tweet](https://twitter.com/swapnakpanda/status/1475139010053894146) which [lead to this ](https://twitter.com/swapnakpanda/status/1475497954559815687) where the user copied the Hello World example from the official PHP docs where it may not be realized to newcomers that PHP is not used to **only** produce HTML code.

However, we include the note that it can be used interchangeably between markup languages.


cmb69:
Yeah, I see your point, and of course that works, because you can basically throw anything at browsers, but it's not a good starting point, in my opinion. The example as is, is correct for usage with a browser (although there has been criticism that mixing PHP and HTML this way is not good practice). It's different for the command line (and most programs in that tweet would use the command line). So maybe we want to extend the tutorial wrt. CLI scripts, and only afterwards tell about Web scripts?

jasonmccreary:
As part of the conversation on Twitter, I do believe a "standalone" example of using a PHP script (not an HTML page) to output something is important for newcomers. Even if it is just to a side note of, "You may also use PHP directly, without embedding it within HTML. Here's an example of a PHP script which outputs the same message:"

cmb69:
@jasonmccreary, yeah, that's what I was suggesting. However, as the patch is now, it says access the page via the browser.

jasonmccreary:
Indeed. Leaving the comment for @ahinkle.

afilina:
@cmb69 I would say that we merge this and create a ticket to add more info about the CLI. If we block this because of added scope, we may never merge it.

cmb69:
Well, we should at least fix the build error. Closing and re-opening to retrigger the build.

afilina:
@ahinkle Triggered the build again.

```
Loading and parsing manual.xml... failed.
ERROR (file:////home/runner/work/doc-en/doc-en/en/chapters/tutorial.xml:118:15)
    </example>
---------------^
 Opening and ending tag mismatch: simpara line 104 and example
 ```
 
You just need to fix this and then we can merge.

ahinkle:
Thanks, @jimwins!

kamil-tekiela (approved):

kamil-tekiela:
Thank you!

PR #1271, opened 2021-12-28
Regenerate ext/mysqli class and method synopses from stubs

kocsismate:
Follow-up to https://github.com/php/doc-en/pull/301

### DIFF
--- a/reference/mysqli/mysqli/init.xml
+++ b/reference/mysqli/mysqli/init.xml
@@ -10,8 +10,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <para>&style.oop;</para>
-  <methodsynopsis role="oop">
-   <modifier>public</modifier> <type class="union"><type>null</type><type>false</type></type><methodname>mysqli::init</methodname>
+  <methodsynopsis>
+   <modifier>public</modifier> <type class="union"><type>bool</type><type>null</type></type><methodname>mysqli::init</methodname>
    <void/>
   </methodsynopsis>
   <para>&style.procedural;</para>
@@ -40,7 +40,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an object.
+   Returns &null; on success, &return.falseforfailure;.
   </para>
  </refsect1>
 
@@ -89,7 +89,6 @@
  </refsect1>
 
 </refentry>
-
 <!-- Keep this comment at the end of the file
 Local variables:
 mode: sgml
--- a/reference/mysqli/mysqli/init.xml
+++ b/reference/mysqli/mysqli/init.xml
@@ -40,7 +40,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &null; on success, &return.falseforfailure;.
+   <methodname>mysqli::init</methodname> returns &null; on success, &return.falseforfailure;.
+   <function>mysqli_init</function> returns an object on success, &return.falseforfailure;.
   </para>
  </refsect1>
 
--- a/reference/mysqli/mysqli_result/construct.xml
+++ b/reference/mysqli/mysqli_result/construct.xml
@@ -19,7 +19,8 @@
   &warn.undocumented.func;
 
   <warning xmlns="http://docbook.org/ns/docbook">
-   <simpara>This method is not usable on its own.  It exists only for internal purposes.
+   <simpara>
+    This method is not usable on its own.  It exists only for internal purposes.
    </simpara>
   </warning>
  </refsect1>
--- a/reference/mysqli/mysqli_result/construct.xml
+++ b/reference/mysqli/mysqli_result/construct.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="mysqli-result.construct" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>mysqli_result::__construct</refname>
+  <refpurpose>Constructs a <classname>mysqli_result</classname> object</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <constructorsynopsis>
+   <modifier>public</modifier> <methodname>mysqli_result::__construct</methodname>
+   <methodparam><type>mysqli</type><parameter>mysql</parameter></methodparam>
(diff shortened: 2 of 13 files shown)


reference/mysqli/mysqli_result/construct.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kocsismate, line 1: This ocumentation is incomplete, so any help is welcome

reference/mysqli/mysqli_stmt/bind-result.xml
```diff
@@ -16,10 +16,10 @@
    <methodparam rep="repeat"><type>mixed</type><parameter role="reference">vars</parameter></methodparam>
   </methodsynopsis>
   <para>&style.procedural;</para>
-  <methodsynopsis>
+  <methodsynopsis role="procedural">
    <type>bool</type><methodname>mysqli_stmt_bind_result</methodname>
-   <methodparam><type>mysqli_stmt</type><parameter>stmt</parameter></methodparam>
    <methodparam><type>mixed</type><parameter role="reference">var</parameter></methodparam>
+   <methodparam><type>mysqli_stmt</type><parameter>statement</parameter></methodparam>
```
kamil-tekiela, line 22: Please fix the parameter order.

reference/mysqli/mysqli_warning/construct.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="mysqli-warning.construct" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>mysqli_warning::__construct</refname>
+  <refpurpose>Private constructor to disallow direct instantiation</refpurpose>
```
kamil-tekiela, line 5: Can you also add the version.xml information?

kamil-tekiela (changes requested):

reference/mysqli/mysqli/init.xml
```diff
@@ -10,8 +10,8 @@
  <refsect1 role="description">
   &reftitle.description;
   <para>&style.oop;</para>
-  <methodsynopsis role="oop">
-   <modifier>public</modifier> <type class="union"><type>null</type><type>false</type></type><methodname>mysqli::init</methodname>
+  <methodsynopsis>
+   <modifier>public</modifier> <type class="union"><type>bool</type><type>null</type></type><methodname>mysqli::init</methodname>
```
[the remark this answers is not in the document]
kocsismate: I'd prefer if `bool|null` was used because `false|null` is an invalid type. The return value section will clarify the actual return value.

reference/mysqli/mysqli_warning/construct.xml
kocsismate: Yes, I definitely forgotten about it.

reference/mysqli/mysqli/init.xml
```diff
@@ -40,7 +40,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an object.
+   Returns &null; on success, &return.falseforfailure;.
```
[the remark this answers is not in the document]
kocsismate: Ahh, I didn't realize that both the procedural and the OOP version is documented on the same page. I'll simply keep both for now, and then a followup fix can do further changes.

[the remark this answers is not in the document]
cmb69: Yep, that's a mess, but splitting into multiple pages might not the best solution, either.

reference/mysqli/mysqli_result/construct.xml
cmb69: In my opinion, it is very import to document that this is not usable on its own (and that it's only meaninful for internal usage).

reference/mysqli/mysqli_stmt/bind-result.xml
cmb69: What's wrong about the parameter order?

reference/mysqli/mysqli_result/construct.xml
kocsismate: Alright, I'll try to come up with a suitable description (I was't aware that the constructor is not usable on its own until now).

reference/mysqli/mysqli_stmt/bind-result.xml
kocsismate: `$statement` used to become the 2nd parameter (because I messed up the `git add`)

reference/mysqli/mysqli_result/construct.xml
```diff
@@ -0,0 +1,61 @@
...
+   <methodparam choice="opt"><type>int</type><parameter>result_mode</parameter><initializer><constant>MYSQLI_STORE_RESULT</constant></initializer></methodparam>
+  </constructorsynopsis>
+  <para>
+
+  </para>
+
+  &warn.undocumented.func;
+
+  <warning xmlns="http://docbook.org/ns/docbook">
+   <simpara>This method is not usable on its own.  It exists only for internal purposes.
```
kocsismate, line 22: Does this warning make any sense?

cmb69, line 22: IMO good, but style nit:
```suggestion
   <simpara>
    This method is not usable on its own.  It exists only for internal purposes.
```

kamil-tekiela: Sorry, it seems I wasn't very accurate in my earlier sentence. One can of course use this constructor from userland. It is usable in that sense. It's just pretty pointless... 

See my Stack Overflow post https://stackoverflow.com/a/62745218/1839439

We should have a manual page describing it, but we can just leave it as "undocumented" method. :) Or we can use the example from my SO post with a simple sentence about what the constructor does.

kamil-tekiela: See also https://github.com/php/php-src/blob/master/ext/mysqli/tests/mysqli_class_mysqli_result_interface.phpt#L126

cmb69: Ah, thanks for the clarification! I don't like undocumented stuff generally, but in this case we have it in the stubs, so users might wonder what it does and how exactly it works, so I suggest to document it.

kamil-tekiela: Let's merge this PR without this extra paragraph. I will think tomorrow about how to provide simple documentation for this constructor.

kocsismate: I like the idea :)

kamil-tekiela (approved):
Thanks. I don't see any other issues.

kocsismate:
Thank you for the review!

PR #1272, opened 2021-12-28
Update some string function parameter default values

kocsismate:
These are probably somewhat controversial changes

### DIFF
--- a/reference/strings/functions/get-html-translation-table.xml
+++ b/reference/strings/functions/get-html-translation-table.xml
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>array</type><methodname>get_html_translation_table</methodname>
    <methodparam choice="opt"><type>int</type><parameter>table</parameter><initializer><constant>HTML_SPECIALCHARS</constant></initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer><constant>ENT_QUOTES</constant> | <constant>ENT_SUBSTITUTE</constant> | <constant>ENT_HTML401</constant></initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer>ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401</initializer></methodparam>
    <methodparam choice="opt"><type>string</type><parameter>encoding</parameter><initializer>"UTF-8"</initializer></methodparam>
   </methodsynopsis>
   <para>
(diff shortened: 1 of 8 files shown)


reference/strings/functions/get-html-translation-table.xml
```diff
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>array</type><methodname>get_html_translation_table</methodname>
    <methodparam choice="opt"><type>int</type><parameter>table</parameter><initializer><constant>HTML_SPECIALCHARS</constant></initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer><constant>ENT_QUOTES</constant> | <constant>ENT_SUBSTITUTE</constant> | <constant>ENT_HTML401</constant></initializer></methodparam>
```
kocsismate, line 14: It's would be a bit tedious to fix gen_stub.php so that it can parse expressions and transform constants to have the `<constant>` tag. If necessary, I could try to implement this change, but I'd also be ok with the currently generated output.

reference/strings/functions/ltrim.xml
```diff
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type>string</type><methodname>ltrim</methodname>
    <methodparam><type>string</type><parameter>string</parameter></methodparam>
-   <methodparam choice="opt"><type>string</type><parameter>characters</parameter><initializer>" \n\r\t\v\0"</initializer></methodparam>
+   <methodparam choice="opt"><type>string</type><parameter>characters</parameter><initializer>" \n\r\t\v\x00"</initializer></methodparam>
```
kocsismate, line 14: Since https://github.com/nikic/PHP-Parser/commit/33d7c8d3d8178b3d2442c85c155b2715d6ed2592#diff-d9d234fe7bea15ea81db29f63acb50bfe9c8e09bbb3ff7267bb319945fb8fc01, PHP Parser UF-8 encodes `\0` (and other invalid characters). Is it OK if the documentation uses this form as well?

reference/strings/functions/get-html-translation-table.xml
cmb69: I think this is acceptable.

reference/strings/functions/ltrim.xml
cmb69: This is fine.

cmb69 (approved):
Thank you!

PR #1273, opened 2021-12-28
Regenerate ext/spl method synopses based on stubs

kocsismate:


reference/spl/multipleiterator/current.xml
```diff
@@ -28,21 +28,45 @@
...
   </para>
  </refsect1>
 
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   A <classname>RuntimeException</classname> if mode <constant>MIT_NEED_ALL</constant>
-   is set and at least one attached iterator is not valid. Or an
-   <classname>IllegalValueException</classname> if a key is &null; and
+   A <classname>RuntimeException</classname> if the iterator is invalid (since PHP 8.1.0),
```
cmb69, line 38:
```suggestion
   A <classname>RuntimeException</classname> if the iterator is invalid (as of PHP 8.1.0),
```

reference/spl/multipleiterator/key.xml
```diff
@@ -28,23 +28,46 @@
...
+   An <type>array</type> of all registered iterator instances.
   </para>
  </refsect1>
 
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   A <classname>LogicException</classname> if mode <constant>MIT_NEED_ALL</constant> is set,
-   and at least one attached iterator is not valid.
+   A <classname>RuntimeException</classname> if the iterator is invalid (since PHP 8.1.0), or
```
cmb69, line 38:
```suggestion
   A <classname>RuntimeException</classname> if the iterator is invalid (as of PHP 8.1.0), or
```

reference/spl/recursiveiteratoriterator/callhaschildren.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type class="union"><type>bool</type><type>null</type></type><methodname>RecursiveIteratorIterator::callHasChildren</methodname>
+   <modifier>public</modifier> <type>bool</type><methodname>RecursiveIteratorIterator::callHasChildren</methodname>
```
cmb69, line 12: The return values section states: "null if the element has children, otherwise false". While that is apparently nonsense, the method may have returned `null` prior to PHP 8.0.0, if the [RecursiveIteratorIterator had no children](https://github.com/php/php-src/blob/PHP-7.4/ext/spl/spl_iterators.c#L777-L779). I believe, though, that this could never happen due to that [check already happening in `spl_recursive_it_get_method`](https://github.com/php/php-src/blob/PHP-7.4/ext/spl/spl_iterators.c#L893-L895) resulting in a fatal error. If I'm right, it would make sense to convert the check in php-src to an assertion. Otherwise, we should have a changelog entry here.

cmb69 (commented):
Thank you!

reference/spl/recursiveiteratoriterator/callhaschildren.xml
kocsismate: It seems that I managed to reproduce the error case: https://3v4l.org/1r6dP

Apparently, the null return is impossible indeed, so I'll fix the return value section.

cmb69: For some reason, I never think of `::newInstanceWithoutConstructor()`. :)

reference/spl/recursiveiteratoriterator/callhaschildren.xml
```diff
@@ -28,11 +28,9 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &null; if the element has children, otherwise &false;
+   Returns whether the element has children;
```
cmb69, line 31:
```suggestion
   Returns whether the element has children.
```

kocsismate: :D Nice catch!

cmb69 (approved):
Thank you!

PR #1274, opened 2021-12-28
Generate ext/intl class synopses from stubs - part 3

kocsismate:
Only formatting changes (+ fallbacks)

### DIFF
--- a/reference/intl/resourcebundle.xml
+++ b/reference/intl/resourcebundle.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<phpdoc:classref xmlns:phpdoc="http://php.net/ns/phpdoc" xml:id="class.resourcebundle" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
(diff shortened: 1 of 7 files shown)


reference/intl/resourcebundle.xml
```diff
@@ -43,20 +43,27 @@
...
-    </classsynopsisinfo>
-    <!-- }}} -->
 
-    <classsynopsisinfo role="comment">Methods</classsynopsisinfo>
-    <xi:include
-     xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.resourcebundle')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[1])" />
-    <xi:include
-     xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.resourcebundle')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[1])" />
+     <oointerface>
+      <interfacename>IteratorAggregate</interfacename>
```
cmb69, line 52: `ResourceBundle` implements `IteratorAggregate` only as of PHP 8.0.0 (previously, it implemented `Traversable` instead), what needs a changelog entry.

reference/intl/resourcebundle.xml
```diff
@@ -43,20 +43,27 @@
...
+     </oointerface>
+
+     <oointerface>
+      <interfacename>Countable</interfacename>
```
cmb69, line 56: `Countable` is only implemented as of PHP 7.4.0, so there should be a changelog entry as well.

cmb69 (commented):
Thank you!

reference/intl/resourcebundle.xml
kocsismate: Oh, right! Added the missing change log entries

cmb69 (approved):
Thanks!

PR #1275, opened 2021-12-28
Generate ext/soap class synopses from stubs

kocsismate:


reference/soap/soapclient.xml
```diff
@@ -24,34 +23,505 @@
...
+    <fieldsynopsis>
+     <modifier>private</modifier>
+     <type class="union"><type>string</type><type>null</type></type>
+     <varname linkend="soapclient.props.-login">_login</varname>
+     <initializer>null</initializer>
+    </fieldsynopsis>
+    <fieldsynopsis>
+     <modifier>private</modifier>
+     <type class="union"><type>string</type><type>null</type></type>
+     <varname linkend="soapclient.props.-password">_password</varname>
```
cmb69, line 106: I wonder whether we should strip leading underscores when generating the IDs (i.e. `soapclient.props.password` instead of `soapclient.props.-password`), or might that lead to duplicates?

kocsismate: Yeah, I'm a bit afraid of duplications, so overall, I'd prefer using the current IDs.

cmb69: Okay, so let's stick with these.

cmb69 (approved):
Thank you for the PR!

Since this is (mostly?) autogenerated info, I guess it's good to be merged.

kocsismate:
Yes, these are autogenerated + I had to manually add descriptions for the newly added properties in an alphabetical order.

PR #1279, opened 2021-12-29
Document mysqli_result::__construct

kamil-tekiela:


reference/mysqli/mysqli_result/construct.xml
```diff
@@ -13,11 +13,12 @@
    <methodparam choice="opt"><type>int</type><parameter>result_mode</parameter><initializer><constant>MYSQLI_STORE_RESULT</constant></initializer></methodparam>
   </constructorsynopsis>
   <para>
-
+   This method constructs a new <classname>mysqli_result</classname> object.
+  </para>
+  <para>
+   It can be used to create the <classname>mysqli_result</classname> object
+   after calling the <function>mysqli_real_query</function> function.
```
cmb69, line 20: Should we explain that this is basically the same as calling mysqli_store_result() or mysqli_use_result()? Or are there relevant differences?

reference/mysqli/mysqli_result/construct.xml
```diff
@@ -26,14 +27,73 @@
...
+       <constant>MYSQLI_USE_RESULT</constant> - creates a
+       <classname>mysqli_result</classname> object with unbuffered result set. 
+       As long as there are pending records waiting to be fetched, the
+       connection line will be busy and all subsequent calls will return error 
+       <literal>Commands out of sync</literal>. To avoid the error all records 
+       must be fetched from the server or the result set must be discarded by
+       calling <function>mysqli_free_result</function>. The connection must
+       remain open for the rows to be fetched.
+      </para>
+     </listitem>
```
cmb69, line 49: Identation seems to be one level too deep:
```suggestion
    <listitem>
     <para>
      The result mode can be one of 2 constants indicating how the result will
      be returned from the MySQL server.
     </para>
     <para>
      <constant>MYSQLI_STORE_RESULT</constant> (default) - creates a
      <classname>mysqli_result</classname> object with buffered result set.
     </para>
     <para>
      <constant>MYSQLI_USE_RESULT</constant> - creates a
      <classname>mysqli_result</classname> object with unbuffered result set. 
      As long as there are pending records waiting to be fetched, the
      connection line will be busy and all subsequent calls will return error 
      <literal>Commands out of sync</literal>. To avoid the error all records 
      must be fetched from the server or the result set must be discarded by
      calling <function>mysqli_free_result</function>. The connection must
      remain open for the rows to be fetched.
     </para>
    </listitem>
```

reference/mysqli/mysqli_result/construct.xml
```diff
@@ -26,14 +27,73 @@
...
   </variablelist>
  </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns a <classname>mysqli_result</classname> object on success.
+  </para>
+ </refsect1>
+
```
cmb69, line 60: We don't document return values of constructors, since __construct() does not actually return the instance:
```suggestion
```

reference/mysqli/mysqli_result/construct.xml
```diff
@@ -26,14 +27,73 @@
...
+]]>
+   </screen>
+  </example>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>mysqli_real_query</function></member>
```
cmb69, line 92: Maybe link to mysqli_store_result() and mysqli_use_result():
```suggestion
    <member><function>mysqli_real_query</function></member>
    <member><function>mysqli_store_result</function></member>
    <member><function>mysqli_use_result</function></member>
```

cmb69 (commented):
Thank you!

reference/mysqli/mysqli_result/construct.xml
kamil-tekiela: No differences. It's just an alternative syntax.

reference/mysqli/mysqli_result/construct.xml
kamil-tekiela: Ahh ok. I copied it over from `PDO::__construct`

cmb69: Thanks for pointing this out! I [fixed it right away](670525ef4ca3b1c716040cdc216c59b095f21670).

cmb69 (approved):
Thank you!

PR #1282, opened 2021-12-30
Drop main dummy manual page

cmb69:
This doesn't appear to be relevant for years, anyway.


cmb69:
@salathe, @Girgias, what do you think?

Girgias:
I mean this shouldn't ever happen. But isn't this generated by PHP when somehow "main" is used as the function in a diagnostic?

But yeah otherwise should be fine to drop this, would be good to amend the section-order QA script if this lands to remove this file from the exclusion list. :-)

cmb69:
> But isn't this generated by PHP when somehow "main" is used as the function in a diagnostic?

Can this still happen, or have all such cases been resolved in the meantime?

My main motivation for dropping this page is actually to update `url.php.bugs`. This is used several times in the manual about php-src bugs, but besides from this page only used once about doc bugs. If we drop this page, we can easily hard-code the single remaining instance referring to docs, and change `url.php.bugs` to point to https://github.com/php/php-src/issues.

Girgias:
Well, that seems like a better reason for dropping it then any potential issue which might arise in a bug in php-src, which can be reported anyway as a bug. And I do think all of the cases should have been resolved if no one has encountered such a case since PHP 5.1

philip:
So long, old timer main() friend, and thanks for all the fish.

PR #1285, opened 2021-12-31
Ambiguous phrasing in generators

colshrapnel:
There is a widespread rumor that generators help to reduce the memory consumption. I see it everywhere. People constantly asking to explain how a generator can reduce the memory consumption, based on the ambiguous wording in the manual.

### DIFF
--- a/language/generators.xml
+++ b/language/generators.xml
@@ -15,8 +15,11 @@
   </para>
 
   <para>
-   A generator allows you to write code that uses &foreach; to iterate over a
-   set of data without needing to build an array in memory, which may cause
+   A generator allows you to write a code that can disguise a regular loop 
+   the way it looks as though it's an array iterated over using &foreach;. 
+   In case your code expects an array, but each item is actually an iteration
+   result of some loop, then you can make it look as an array by creating 
+   a generator, without needing to build an array in memory, which may cause
    you to exceed a memory limit, or require a considerable amount of
    processing time to generate. Instead, you can write a generator function,
    which is the same as a normal
@@ -95,6 +98,17 @@ Single digit odd numbers from xrange(): 1 3 5 7 9
    </screen>
   </example>
 
+  <note>
+   <simpara>
+    It is important to understand that a generator is essentially a syntax sugar
+    that cannot really reduce the memory consumption. It's the undelying loop 
+    that does the job, providing one item at a time. While a generator simply 
+    allows to convert any other loop, such as &while; or &for; into &foreach;,
+    making it possible to use the uniform interface to access either arrays, 
+    streams or sequences generated on the fly.
+   </simpara>
+  </note> 
+
   <sect2 xml:id="language.generators.object">
    <title><classname>Generator</classname> objects</title>
    <para>
@@ -107,7 +121,6 @@ Single digit odd numbers from xrange(): 1 3 5 7 9
    </para>
   </sect2>
  </sect1>
-
  <sect1 xml:id="language.generators.syntax">
   <title>Generator syntax</title>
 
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR!

The effect on memory is a property of iterators, not of generators in particular. So, in my opinon, that info should be moved to the `Iterator` man page, and this page should focus on "generators provide an easy way to implement simple iterators".

Also note that personalization ("you") should be avoided in the manual proper.

colshrapnel:
Thank you for the pointers, I am yet to learn the correct manual language. 

Note that neither iterators can spare you  a bit of memory. It's simply not their job. And *that* is the problem that people think otherwise. Yet it's even harder to implant this idea on the iterators man page, because it uses "all visible properties" as an example. 

Seriously, is there a way to fix at least this page without going down the rabbit hole?  Because really, there *is* the problem. I see this question literally every week. Or recently I came across an article titled "Deal with Memory Gently using “Yield” in PHP" that has quite a lot of views on Medium. People get the wrong idea on generators based on the current manual page. 

I don't insist on my phrasing at all. It could be anything, just without any reference to memory limit.

nikic:
I'm not sure I understand what the issue with the existing wording is, it looks perfectly accurate to me? If you compare a generator to a function that does exactly the same but returns the result as an array, that will reduce memory usage ... right?

colshrapnel:
@nikic there can be different ways different people take the same information. What is obvious for someone, can have a completely different meaning for someone else. Finding correct definitions is sort of my superpower. I spend a lot of time trying to find a definition that's solid all around and cannot (or at least hard to) be misinterpreted. And this case is clear to me, people massively misattribute the reduced memory usage to a generator, not to a loop.

Any description, while being technically correct, can make a wrong impression anyway. Just like with infamous escape_string, which, although states correctly that it escapes *strings*, nevertheless makes a widespread impression that it "protects" any kind of data, making your entire SQL immune to attacks. Same here. You are technically correct with this statement but there is an important part: it is not a generator that reduces the memory usage, but the underlying loop. You can always have the same memory footprint by simply processing every result inside a loop, without a generator; but you cannot have it with a generator but without a loop that can produce one result at a time. A generator just creates a handy interface for the loop. It sort of turns the loop inside out, allowing to process the results one by one not inside, but outside. 

To me, generators (or iterators for that matter) is a fantastic feature, they help to create uniform interfaces for the different data sources, be it an array or a stream or a computed sequence. The epitome of polymorphism. That's the description which I'd like to emphasize. Whereas the memory thing should be diminished and correctly explained. Or so I feel.

colshrapnel:
"If you compare a generator to a function that does exactly the same but returns the result as an array, that will reduce memory usage ... right?" - yes. 

If you compare a loop that delivers a single result at a time to a function that returns the result as an array,  that will reduce memory usage - *also yes*. But it's this loop that does the job. You can always use this loop instead of a generator. It won't look  as nice, but that's what really reduces the memory usage. 

I am in despair. I know I am right, but I cannot express it properly. I know that the current phrasing, although technically correct, deceives many people. But I cannot explain it. It's too subtle. But I also know that every now and then a new article is coming out telling the world that generators help to reduce the memory consumption. But the thing is, *only if you want to use foreach*. This is important. This critical part is somewhat eluding. People tend to overlook the foreach part, and to leap already to the memory consumption. Taking generators for magic wands that reduce the resource consumption in a puff of smoke, for any code they write. 

To reduce the resource consumption, *one should read large sets of data in iterations, one piece at a time*. While a generator can only wrap it nicely, making the code "think" the whole set is delivered at once and can be iterated over using foreach. 

@kamil-tekiela I know you have the power to understand my indistinct blab and put it into proper English. Do you get what I am trying to say here? There must be something that emphasizes on the foreach part.

PR #1293, opened 2022-01-04
Sync DirectoryIterator method return types

kocsismate:


reference/spl/directoryiterator/current.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>DirectoryIterator</type><methodname>DirectoryIterator::current</methodname>
+   <modifier>public</modifier> <type>mixed</type><methodname>DirectoryIterator::current</methodname>
```
kocsismate, line 12: We cannot use DirectoryIterator because `FilesystemIterator` overrides the method and uses a different return type, but the return value section already clarifies exactly what is returned.

reference/spl/directoryiterator/key.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type class="union"><type>int</type><type>false</type></type><methodname>DirectoryIterator::key</methodname>
+   <modifier>public</modifier> <type>mixed</type><methodname>DirectoryIterator::key</methodname>
```
kocsismate, line 11: This is a similar subclassing issue as above: `FilesystemIterator::key()` returns a string instead.

kocsismate:
> PS: to clarify, SPL is the mess, not this PR. :)

Yeah, I understood what you mean :)

> What a mess! See also https://github.com/php/php-src/commit/d0a0518798411a39f8c0cb40567cdb3511e8ae26.

Yep, SPL Is very messy for sure... But the PR is still valid, right? :)

cmb69 (approved):
> But the PR is still valid, right? :)

I think so.

PR #1302, opened 2022-01-05
List all ext/mysqli properties instead of including them - part 1

kocsismate:


kamil-tekiela (approved):
LGTM

P.S. I just noticed that the method names are botched on this page. It seems to be caused by b06b3f19af93c65570a69e1fca4d7cad0e7d43dd Is this something that you can fix?

kocsismate:
Uh, yeah, I'll fix it ASAP

PR #1304, opened 2022-01-05
List all ext/mysqli properties instead of including them - part 2

kocsismate:


reference/mysqli/mysqli_result.xml
```diff
@@ -35,20 +35,61 @@
...
+   &reftitle.properties;
+   <variablelist>
+    <varlistentry xml:id="mysqli-result.props.type">
+     <term>
+      <varname>type</varname>
+     </term>
+     <listitem>
+      <para>
+       Stores whether the result is unbuffered or buffered as an &integer;
+       (<literal>1</literal> or <literal>0</literal>, respectively).
```
kamil-tekiela, line 86:
```suggestion
       Stores whether the result is buffered or unbuffered as an &integer;
       (<literal>0</literal> or <literal>1</literal>, respectively).
```
Small nit. The positive clause usually comes first. 

I wonder whether we should use constants instead of numbers. i.e. `MYSQLI_STORE_RESULT` and `MYSQLI_USE_RESULT`

kocsismate: Thanks for the suggestion, makes sense absolutely. I'd be happy with this as well:
```suggestion
       (<literal>MYSQLI_STORE_RESULT</literal> or <literal>MYSQLI_USE_RESULT</literal>, respectively).
```

cmb69: I think these constants should at least be mentioned there. If we use them instead of the numbers, their values should be added to the mysqli constants page.

kocsismate: Ah, I hoped that they are already documented. OK then, I'll add them :)

kamil-tekiela: Yeah, constants instead of numbers makes sense. The values for these constants aren't defined in the PHP manual (it might be on purpose).

kocsismate: I hoped that the updated description is good enough.

cmb69: Yeah, it's good enough. :)

reference/mysqli/constants.xml
```diff
@@ -143,15 +143,15 @@
...
+      For using buffered result sets. It has a value of <literal>1</literal>.
      </para>
     </listitem>
    </varlistentry>
    <varlistentry xml:id="constantmysqli-use-result">
     <term><constant>MYSQLI_USE_RESULT</constant></term>
     <listitem>
      <para>
-      For using unbuffered resultsets
+      For using unbuffered result sets. It has a value of <literal>0</literal>.
```
kamil-tekiela, line 154: I think it's the other way around. This one is 1 and the other is 0. Can you please double check values?

kocsismate: Grrr, yes, MYSQLI_USE_RESULT is 1. Thanks for catching this, I messed it up. :S

kamil-tekiela (approved):
LGTM Thanks

PR #1305, opened 2022-01-05
List all ext/mysqli properties instead of including them - part 3

kocsismate:


kamil-tekiela:
Can you explain why this change? What do we gain by this? It only feels like the same information is presented twice. It might make more sense for extensions where there is a clear distinction between properties and methods, but in mysqli it is better as it is now IMHO.

kocsismate:
> Can you explain why this change

Yeah, it's because this is the only way we can synchronize the documentation with the stubs which is necessary to guarantee that the documentation can always remain up-to-date (i.e.: the first two PRs added 1-1 missing properties). I agree though that the duplication is not fortunate. But I think we are better with my solution than without it.

cmb69:
Would it be viable to adapt gen_verify_stub and mysqli.stub.php to special case this?

kamil-tekiela:
BTW the synopsis changes look ok in both PRs. My only concern is about the duplication of property information on `mysqli`, `mysqli_stmt` and `mysqli_result` pages. The properties have their own pages in the manual and they are linked to these pages. If you remove this link and instead point to a simple sentence on the same page, it will be confusing. Users will have to click twice to get where they want to go. That simple sentence is redundant with the parameter name. e.g. "error_list" - "Stores a list of errors from the last command executed." That might be enough for a hover tooltip, but if someone clicks the link they'd rather go to the full page for that item. 

Compare it with `mysqli_driver` or `mysqli_warning` where we don't have pages for properties. It makes sense there to have a section for properties and an anchor link to it. 

If these three pages can be exempted I think it would be best.

kocsismate:
> Would it be viable to adapt gen_verify_stub and mysqli.stub.php to special case this?

I've just implemented this: https://github.com/php/php-src/pull/7892/commits/58a2570395bc5e25fbba12880c701296398dcf9c

I'm not 100% happy to hard-code URLs into the stubs... but I think overall, this solution would be ok.

kamil-tekiela:
> > Would it be viable to adapt gen_verify_stub and mysqli.stub.php to special case this?
> 
> I've just implemented this: [php/php-src@58a2570](https://github.com/php/php-src/commit/58a2570395bc5e25fbba12880c701296398dcf9c) And the result looks like this: https://github.com/php/doc-en/pull/1304/files#diff-e87397eeb1dbf3695e5bb2c6641d5d10b8f0eda4cd4dc6e654a1da62f52448ef
> 
> I'm not 100% happy to hard-code URLs into the stubs... but I think overall, this solution would be ok.

I think this is much better. It's not perfect, but that is rather the fault of the extension being messy. In an ideal world, we would have a clear distinction between properties and methods, or we would have getters instead of readonly public properties.

kamil-tekiela (approved):
Thanks!

PR #1306, opened 2022-01-06
Clarify class-types meaning in composite type description.

mumumu:
#982 is a great work, but the word `class-types` is misleading, especially for translators.
I considered it is only included classes, not interfaces.

refs: https://wiki.php.net/rfc/pure-intersection-types

### DIFF
--- a/language/types/declarations.xml
+++ b/language/types/declarations.xml
@@ -395,7 +395,7 @@ NULL
    </listitem>
    <listitem>
     <simpara>
-     Intersection of class-types. As of PHP 8.1.0.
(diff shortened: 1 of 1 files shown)


mumumu:
Thank you!

cmb69:
We should consider chosing (and defining) a word/words (XML entities) for this, because we use that several times in the manual, and the list might further extend in the future (one may argue that enum names are not class names, and what about traits).

Girgias:
It might make sense to add a dedicated page for this like we did for numeric strings?

PR #1307, opened 2022-01-07
Default value for error_reporting is E_ALL as of PHP 8.0.0

Girgias:
I'm not sure if it's valuable to keep the note about the purpose of enabling E_NOTICE, as most of the information currently in this note is incorrect.

Also not sure how to enclose the old value, as ``<constant>`` is not a child of ``<code>`` or ``<literal>`` apparently...

### DIFF
--- a/reference/errorfunc/ini.xml
+++ b/reference/errorfunc/ini.xml
@@ -164,25 +164,18 @@
       <link linkend="ini.display-errors">display_errors</link> directive.
      </para>
      <para>
-      The default value
-      is <constant>E_ALL</constant> &amp;
-      ~<constant>E_NOTICE</constant> &amp;
-      ~<constant>E_STRICT</constant> &amp;
-      ~<constant>E_DEPRECATED</constant>. This setting does not
-      show <constant>E_NOTICE</constant>, <constant>E_STRICT</constant>
-      and <constant>E_DEPRECATED</constant> level errors. You may want
-      to show them during development.
+      The default value is <constant>E_ALL</constant>.
      </para>
      <note>
-      <para>Enabling <constant>E_NOTICE</constant> during development has
-      some benefits. For debugging purposes: NOTICE messages will warn you
-      about possible bugs in your code. For example, use of unassigned values
-      is warned. It is extremely useful to find typos and
-      to save time for debugging. NOTICE messages will warn you about bad style.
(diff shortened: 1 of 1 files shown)


reference/errorfunc/ini.xml
```diff
@@ -164,25 +164,18 @@
...
-      about possible bugs in your code. For example, use of unassigned values
-      is warned. It is extremely useful to find typos and
-      to save time for debugging. NOTICE messages will warn you about bad style.
-      For example, <literal>$arr[item]</literal> is better to be written as
-      <literal>$arr['item']</literal> since PHP tries to treat
-      <literal>"item"</literal> as constant. If it is not a constant, PHP assumes
-      it is a string index for the array.
+      <para>
+       Prior to PHP 8.0.0, the default value was:
+       <userinput><constant>E_ALL</constant> &amp;
```
cmb69, line 172: I'm not happy using `<userinput>` here. I'd use `<code>` instead, but although that [should work for DocBook 4.5](https://tdg.docbook.org/tdg/4.5/code.html), apparently it does not for us. So maybe just stick with `<userinput>`.

reference/errorfunc/ini.xml
```diff
@@ -164,25 +164,18 @@
...
-      is <constant>E_ALL</constant> &amp;
-      ~<constant>E_NOTICE</constant> &amp;
-      ~<constant>E_STRICT</constant> &amp;
-      ~<constant>E_DEPRECATED</constant>. This setting does not
-      show <constant>E_NOTICE</constant>, <constant>E_STRICT</constant>
-      and <constant>E_DEPRECATED</constant> level errors. You may want
-      to show them during development.
+      The default value is <constant>E_ALL</constant>.
      </para>
      <note>
```
cmb69, line 169: I think this shouldn't be a `<note>`.

cmb69 (commented):
Thank you!

reference/errorfunc/ini.xml
Girgias: I'm not super happy either, as that what I used initially but Docbook 5 changed it, which I think we kinda use now?

cmb69: Ah, indeed! I thought we're using a customized 4.5, but might be a customized 5.0 as well (or maybe just some mix).

cmb69 (approved):
Let's go with this. :)

PR #1311, opened 2022-01-08
Properties do not need to be declared with a visibility modifier

Girgias:
The first paragraph is becoming a bit complexe, so maybe it would be better to rewrite it?

### DIFF
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -22,12 +22,12 @@
     <literal>var</literal> keyword instead of a modifier.
    </para>
   </note>
- <note>
-  <simpara>
-   A property declared without a <xref linkend="language.oop5.visibility"/>
-   modifier will be declared as <literal>public</literal>.
-  </simpara>
- </note>
+  <note>
+   <simpara>
+    A property declared without a <xref linkend="language.oop5.visibility"/>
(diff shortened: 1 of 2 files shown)


language/oop5/properties.xml
```diff
@@ -7,34 +7,33 @@
...
-   and <literal>private</literal>.
-  </para>
   <note>
    <para>
-    An alternative and not recommended way of declaring class properties, as it is to maintain backward
-    compatibility with PHP 4, is by using
-    the <literal>var</literal> keyword.
-    It will treat the property identically as it would have been
-    declared as <literal>public</literal>.
+    An outdated way of declaring class properties, is by using the
```
cmb69, line 21: Maybe
```suggestion
    An obsolete way of declaring class properties, is by using the
```

cmb69 (approved):
Thank you! In my opinion, the length of the first paragraph is still okay (on the contrary, I feel that the manual often has too short paragraphs).

Girgias:
> Thank you! In my opinion, the length of the first paragraph is still okay (on the contrary, I feel that the manual often has too short paragraphs).

It was more about how the phrasing felt a bit suboptimal IMHO not about the length :)

PR #1313, opened 2022-01-08
pdo: check for successfull preparation before running execute() et. all

staabm:


kamil-tekiela:
Thank you, but unfortunately this is not an improvement.

staabm:
why is it not required to check the possible false-return beforehand?

kamil-tekiela:
These functions return false only when error reporting is silenced. As this is not the normal/default behavior nor is it recommend, we don't need to put such checks in the documentation. It is assumed that users who intentionally disable error reporting will implement their own checks accordingly.

cmb69:
Maybe we should *explicitly* assume `$pdo->getAttribute(PDO::ATTR_ERRMODE) === PDO::ERRMODE_EXCEPTION` (similar to the mysqli examples).

kamil-tekiela:
> Maybe we should _explicitly_ ~assume~assert `$pdo->getAttribute(PDO::ATTR_ERRMODE) === PDO::ERRMODE_EXCEPTION` (similar to the mysqli examples).

We could, but at this point it would be nothing more than noise. Most PDO users are already using exception error mode and the PDO examples are not self-contained. It makes more sense with mysqli, where the examples contained the connection part anyway and many users were not aware of the fact that error handling even exists. I'd consider adding such assertion to have very little value for PDO examples.

staabm:
a lot of methods in PDO context are defined that they might contain `false`.

IMO it would at least be usefull to have a list of methods which no longer can `return false`, when exception throwing is active

cmb69:
> Most PDO users are already using exception error mode […]

Citation needed. :) ERRMODE_EXCEPTION is the default only as of PHP 8.0.0, so many readers might still assume otherwise. We do not need to add an actual assertion to each example, but it could still make sense to document somewhere that "all PDO examples are written under that assumption (unless stated otherwise)".

PR #1320, opened 2022-01-11
Updates a data type to a valid one in the example

Shakil-Shahadat:
SQLite3 does not have the 'STRING' data type. Please check https://www.sqlite.org/datatype3.html.


cmb69:
Yeah, STRING is a bad choice here (would result in NUMERIC affinity). Thank you for the PR!

PR #1321, opened 2022-01-11
Fix DOMParentNode and DOMChildNode class and method synopses

kocsismate:
Regenerated these pages via gen_stub.php


saundefined (approved):
Thank you!

PR #1322, opened 2022-01-11
Document options for gnupg_init

bukka:
This adds documentation for options parameter added to php-gnupg 1.5.0


bukka:
Will actually add a changelog as well so it's clear that it's for 1.5.0+.

PR #1323, opened 2022-01-12
Rename OCI-Lob and OCI-Collection

kocsismate:
This PR renames the files and references of OCILob and OCICollection according to the conventions.

How should we preserve the original links?


cmb69:
Thanks for the PR! I think we can preserve the original URLs with some additional hackery in [errors.inc](https://github.com/php/web-php/blob/master/include/errors.inc), but I have no idea to avoid the break for the translations (although some search and replace should allow to update most of it).

kocsismate:
Thanks for the pointer! I've just created https://github.com/php/doc-en/pull/1323.

PR #1324, opened 2022-01-13
Clarification for setting 0 to jit_hot_loop/func

stkeke:
I tried to tell JIT to compile every loop and function
by setting both opcache.jit_hot_loop and opcache.jit_hot_func
to 0, but failed to get the expected result.

After having read the source code and done simple experiments, I 
found that setting these two options to 0 will actually disable JIT to
compile any iterations and calls.

So I would like to add clarifications for this behavior in PHP manual.


cmb69:
Thank you!

PR #1327, opened 2022-01-13
[PHP 8.1] New in initializers

saundefined:
Related [RFC](https://wiki.php.net/rfc/new_in_initializers).

### DIFF
--- a/language/oop5/decon.xml
+++ b/language/oop5/decon.xml
@@ -163,6 +163,34 @@ class Point {
      </para>
     </note>
    </sect3>
+
+   <sect3 xml:id="language.oop5.decon.constructor.new">
+    <title>New in initializers</title>
+    <para>
+     As of PHP 8.1.0, objects can be used as default parameter values,
+     static variables, and global constants, as well as in attribute arguments.
+    </para>
+    <example>
+     <title>Using new in initializers</title>
+     <programlisting role="php">
+<![CDATA[
+<?php
+class Service
+{
+    private Logger $logger;
+   
+    public function __construct(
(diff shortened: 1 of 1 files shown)


mumumu:
Also, we should add the following description.

> Objects can also be passed to define() now. 

https://www.php.net/manual/en/migration81.new-features.php#migration81.new-features.core.new-in-initializer

mumumu:
From, [RFC](https://wiki.php.net/rfc/new_in_initializers), we should add some more example code.

> New expressions are allowed in parameter default values, attribute arguments, static variable initializers and global class constant initializers. Parameter default values also include defaults for promoted properties: 

```php
static $x = new Foo;
 
const C = new Foo;
 
function test($param = new Foo) {}
 
#[AnAttribute(new Foo)]
class Test {
    public function __construct(
        public $prop = new Foo,
    ) {}
}
```

mumumu:
Also, we should add the following `limitations`, maybe in `<note><para>...</para></note>` tag.

> The use of a dynamic or non-string class name or an anonymous class is not allowed. The use of argument unpacking is not allowed. The use of unsupported expressions as arguments is not allowed.

```php
// All not allowed (compile-time error):
function test(
    $a = new (CLASS_NAME_CONSTANT)(), // dynamic class name
    $b = new class {}, // anonymous class
    $c = new A(...[]), // argument unpacking
    $d = new B($abc), // unsupported constant expression
) {}
```

saundefined:
@mumumu thanks for the review!

Crell:
Thanks!

PR #1330, opened 2022-01-14
[PHP 8.1] Calling a static element on a trait is deprecated

saundefined:
Deprecated Features [Changelog](https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.static-trait)

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -486,6 +486,11 @@ enum Size
    They may only include methods and static methods.  A trait with properties will
    result in a fatal error.
   </para>
+  
+  <para>
+   As of PHP 8.1.0, calling static trait element directly is deprecated.
+   It should only be accessed on a class using the trait.
+  </para>
 
   <programlisting role="php">
 <![CDATA[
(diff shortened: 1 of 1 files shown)


language/enumerations.xml
```diff
@@ -486,6 +486,11 @@ enum Size
    They may only include methods and static methods.  A trait with properties will
    result in a fatal error.
   </para>
+  
+  <para>
+   As of PHP 8.1.0, calling static trait element directly is deprecated.
+   It should only be accessed on a class using the trait.
```
cmb69, line 492: This is about methods and properties, but you cannot *call* properties (you can *access* them). Unifying methods and properties as element is okay, but would need another verb (maybe "use"), but why not just duplicate the wording from the migration guide here?

saundefined: Thanks! Copied from the migration guide =)

cmb69 (approved):
Thank you!

mumumu:
This deprecated feature is applied `only` to enumerations?
Should we add this to language/oop5/traits.xml ?

saundefined:
@mumumu Oops, I guess I missed it, seems reasonable

cmb69:
I think it should be *moved* to oop5/traits.xml.

PR #1331, opened 2022-01-14
Deleting usage "limitation" to debugging

ThomasLandauer:
I think it's obvious that `->name` might be handy sometimes :-)

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -118,7 +118,7 @@ $a instanceof Suit;  // true
 
    <para>
     All Cases have a read-only property, <literal>name</literal>, that is the case-sensitive name
-    of the case itself.  That may sometimes be useful for debugging purposes.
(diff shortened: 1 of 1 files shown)


language/enumerations.xml
```diff
@@ -118,7 +118,7 @@ $a instanceof Suit;  // true
 
    <para>
     All Cases have a read-only property, <literal>name</literal>, that is the case-sensitive name
-    of the case itself.  That may sometimes be useful for debugging purposes.
+    of the case itself.
```
cmb69, line 121: Yeah, I think it makes sense to remove this sentence. Or is there a particular reason for keeping it @Crell?

Crell: I think that was mainly left over from the spec itself.  I am not wedded to it.  We can kill it.

PR #1334, opened 2022-01-14
Fix openssl cms encoding constants

alexismarquis:
Fix wrong encoding constants being used in openssl-cms-x docs.


cmb69 (approved):
Thank you!

PR #1335, opened 2022-01-14
Reorder DOMDocument properties

kocsismate:
This is the **last class synopsis page which can be normally synced** (and doesn't already have a pending PR). There are some remaining smaller sync issues though (on a total of 15 pages). Some classes don't declare properties (`DateInterval`), or they include methods in a custom way (`DateTimeImmutable`), or explicitly document overridden methods (`SplQueue::__construct()`) etc. The rest of the diff between the docs and the output of gen_stubs comes from missing constant generation support (~ 40-50 pages).

I include a before-after print screen about the properties of DOMDocument because the diff is not easy to overview:
Screenshot 2022-01-14 at 18 16 52
.

### DIFF
--- a/reference/dom/domdocument.xml
+++ b/reference/dom/domdocument.xml
@@ -44,41 +44,8 @@
     <fieldsynopsis>
      <modifier>public</modifier>
      <modifier>readonly</modifier>
-     <type>string</type>
-     <varname linkend="domdocument.props.actualencoding">actualEncoding</varname>
-    </fieldsynopsis>
-    <fieldsynopsis>
-     <modifier>public</modifier>
-     <modifier>readonly</modifier>
-     <type>DOMConfiguration</type>
-     <varname linkend="domdocument.props.config">config</varname>
-    </fieldsynopsis>
-    <fieldsynopsis>
-     <modifier>public</modifier>
-     <modifier>readonly</modifier>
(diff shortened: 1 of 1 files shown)


philip:
Curious, the list appears alphabetical now; what does this new suggested order solve?

cmb69:
It matches the order [in the stub file](https://github.com/php/php-src/blob/master/ext/dom/php_dom.stub.php#L400-L457), so can easily re-generated whenever necessary.

kamil-tekiela:
Why was my suggestion to maintain properties in alphabetical order declined? There was something if I remember correctly.

kocsismate:
> Why was my suggestion to maintain properties in alphabetical order declined? There was something if I remember correctly.

Yes: https://github.com/php/php-src/pull/7531

Just imagine if we sorted the properties of `DateInterval` alphabetically: instead of the totally sensible `$y`, `$m`, `$d`...., you would get a totally random property order.

kocsismate:
Gentle ping in the hope to get reviews :)

cmb69 (approved):
Yeah, I think this is good to be merged (although translators may not like it).

BTW, we should do something about the properties which are documented as deprecated. For instance, `::$actualEncoding` apparently was never reasonably implemented; it always is `null`.

PR #1340, opened 2022-01-17
Jit debug clarification

stkeke:
Clarify jit_debug option and link to zend_jit.h
    
    Clarify how to search for jit_debug options, and add an external link to
    PHP zend_jit.h source file.
    
    Hopefully, the clarification and link can save PHP the developer's and
    user's time/effort with a simple click jumping to where they should go.
    
    This patch has been verified on my local machine and link works well.
    
    This risk is that doc viewer could get a dead link if PHP source code
    file location moved or name changed in future. We just hope this is a
    rare case and will not happen so frequently. We depend on user's
    bug-report to fix any issue.
    
### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -1040,7 +1040,11 @@
     <listitem>
      <para>
       A bit mask specifying which JIT debug output to enable.
-      For possible values, please consult <filename>zend_jit.h</filename>.
+      For possible values, please consult
+      <link xmlns:xlink="http://www.w3.org/1999/xlink"
+            xlink:href="&url.php.git.src.master.view;ext/opcache/jit/zend_jit.h">
+       zend_jit.h
+      </link> (search for macro definitions beginning with <code>ZEND_JIT_DEBUG</code>).
      </para>
     </listitem>
    </varlistentry>


stkeke:
Pend this patch until doc-base PR (https://github.com/php/doc-base/pull/55) is merged.

stkeke:
This final result of this patch looks like
image

cmb69:
Thank you for the PR! I'm not sure if we should go into details regarding debug options for OPcache or even JIT. I think it would be better to have that documentation in https://www.phpinternalsbook.com/ (source hosted at https://github.com/phpinternalsbook/PHP-Internals-Book/).

stkeke:
To tell the truth, I also had a feeling of uncertainty that whether it is appropriate to expose code details or link to source code from the PHP use manual. 

But as zend_jit.h has already been exposed in the manual, why not give users a hyperlink and a little more information? In any way, they have to go to source code file to extract the bits and construct the value which will be given to jit_debug option. This is my logic. 

However, I fully trust and respect maintainer's judgement to any of my PRs because you guys are the most experienced experts in PHP world. 

So, I rely on our maintainer to have the final word! So is to (php/doc-base#55).

stkeke:
> > But as zend_jit.h has already been exposed in the manual, why not give users a hyperlink and a little more information?
> 
> Hmm, I see. Like you already mentioned in the initial commit message, this URL is somewhat brittle, though, and I doubt that we get bug reports from users about that (probably _very_ few actually use that INI setting in the first place). Maybe we should add the link, but leave out the line number?

Thanks @cmb69 for the consideration. I just removed the line number and re-submitted the patch. I verified the link is working.

cmb69:
I've moved the xlink namespace declaration to the top of the file (as customary), and reformatted the link.

Thank you!

PR #1350, opened 2022-01-19
Clarification for usage of JIT tunable options

stkeke:
I once tuned jit_hot_func=300 in my php-fpm and did not catch 
the warning message, which is the main motivation for this patch 
after code reading.

This patch clarifies the usage of the four JIT tunable options -
jit_hot_{loop,func,return,side_exit} - in PHP manual which end users
should be aware of before they start to tune these options.

1. Valid value range
2. Default value will be used in case beyond valid range
3. Special case of zero value

The patch was verified on my machine with phd local website.

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -1114,7 +1114,9 @@
     <listitem>
      <para>
       After how many iterations a loop is considered hot.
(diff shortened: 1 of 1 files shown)


stkeke:
This is the patch result
image

PR #1352, opened 2022-01-19
SimpleXMLElement::children() does not return null

mpdude:
I don't know if the docs are updated from the stubs or the other way round. 

Anyways, here's the PR for the stub:

https://github.com/php/php-src/pull/7969


cmb69:
See https://3v4l.org/a2jSE.

PR #1359, opened 2022-01-27
Fix return type (int → bool)

msamesch:
Mentioned in doc-de/issues/67 by Daijobou (thanks).
Changelog entry?

### DIFF
--- a/reference/imagick/imagick/getimagealphachannel.xml
+++ b/reference/imagick/imagick/getimagealphachannel.xml
@@ -3,7 +3,7 @@
 <refentry xmlns="http://docbook.org/ns/docbook" xml:id="imagick.getimagealphachannel">
  <refnamediv>
   <refname>Imagick::getImageAlphaChannel</refname>
-  <refpurpose>Gets the image alpha channel</refpurpose>
+  <refpurpose>Checks if the image has an alpha channel</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
--- a/reference/imagick/imagick/getimagealphachannel.xml
+++ b/reference/imagick/imagick/getimagealphachannel.xml
@@ -14,7 +14,7 @@
   </methodsynopsis>
 
   <para>
-   Gets the image alpha channel value.
+   Returns whether the image has an alpha channel.
   </para>
  </refsect1>
 
--- a/reference/imagick/imagick/getimagealphachannel.xml
+++ b/reference/imagick/imagick/getimagealphachannel.xml
@@ -33,6 +33,28 @@
   </para>
  </refsect1>
 
+ <refsect1 role="changelog">
+  &reftitle.changelog;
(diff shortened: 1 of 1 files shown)


Daijobou:
Shouldn't it be mentioned as a hint that before 3.6.0 it was an INTEGER? ;)

I found in kohana framework
`if ($watermark->getImageAlphaChannel() !== Imagick::ALPHACHANNEL_ACTIVATE)`

Source https://github.com/koseven/koseven/blob/master/modules/image/classes/Kohana/Image/Imagick.php#L203

So this is now with version Imagick v3.6.0 a _bug_ in this framework (and maybe others).

> imagick::ALPHACHANNEL_ACTIVATE (int)
> This constant is available if Imagick has been compiled against ImageMagick version 6.3.8 or higher.

Source https://www.php.net/manual/en/imagick.constants.php

msamesch:
> Shouldn't it be mentioned as a hint that before 3.6.0 it was an INTEGER? ;) So that developers understand what happened here?

Yes, but I wasn't sure about the exact changelog entry.

I tagged the PR 'help wanted' now.

saundefined:
> Yes, but I wasn't sure about the exact changelog entry.

Not sure, but maybe just something like that?

```xml
<row>
  <entry>imagemagick 3.6.0</entry>
  <entry>
    Returns a &boolean; now; previously, an &integer; was returned.
  </entry>
</row>
```

msamesch:
Hi Sergey,
overlapping comment/edited comment ;)
Thanks for the snippet (added) and the hint. So it's not only about the PHP version.
Which are the conditions for true/false?
Maybe it's better if you or sb else completes this commit.

cmb69:
Maybe @Danack can clarify?

msamesch (commented):
I used the source. ;)

reference/imagick/imagick/getimagealphachannel.xml
```diff
@@ -9,14 +9,12 @@
...
-   <modifier>public</modifier> <type>int</type><methodname>Imagick::getImageAlphaChannel</methodname>
+   <modifier>public</modifier> <type>bool</type><methodname>Imagick::getImageAlphaChannel</methodname>
    <void />
   </methodsynopsis>
 
   <para>
-   Gets the image alpha channel value. The returned value is one of the 
-   <link linkend="imagick.constants.alphachannel">alpha channel constants</link>.
-   &imagick.method.available.0x640;
+   Gets the image alpha channel value.
```
cmb69, line 17: I think this is somewhat misleading; maybe better:
```suggestion
   Returns whether the image has an alpha channel.
```

cmb69 (commented):
Thank you! This looks mostly good to me, to there is one issue that could be addressed.

msamesch:
Thanks for the review and the correction.

PR #1366, opened 2022-01-30
Update dead links for LiteSpeed

istiak101:
Updated old dead links to new links.

### DIFF
--- a/install/unix/litespeed.xml
+++ b/install/unix/litespeed.xml
@@ -38,9 +38,9 @@
    <para>
     To obtain and install either LiteSpeed Web Server or OpenLiteSpeed Web Server, visit the
     LiteSpeed Web Server wiki
-    <link xlink:href="https://www.litespeedtech.com/support/wiki/doku.php/litespeed_wiki:installation">install page</link>
+    <link xlink:href="https://docs.litespeedtech.com/products/lsws/installation/">install page</link>
     or OpenLiteSpeed wiki 
-    <link xlink:href="http://open.litespeedtech.com/mediawiki/index.php/Help:Install">install page</link>.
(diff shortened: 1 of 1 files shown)


install/unix/litespeed.xml
```diff
@@ -38,9 +38,9 @@
    <para>
     To obtain and install either LiteSpeed Web Server or OpenLiteSpeed Web Server, visit the
     LiteSpeed Web Server wiki
```
cmb69, line 40: Apparently, the new documentation is no longer a Wiki, so the wording should be adjusted; maybe just "LiteSpeed Web Server documentation" here.

install/unix/litespeed.xml
```diff
@@ -215,7 +215,7 @@ PHP_LSAPI_MAX_REQUESTS=500 PHP_LSAPI_CHILDREN=35 /path/to/lsphp -b IP_address:po
   OpenLiteSpeed Web Server, and Apache mod_lsapi. For steps on 
   server-side configuration, visit the wiki pages for 
   <link xlink:href="https://www.litespeedtech.com/support/wiki/doku.php/litespeed_wiki:php:configuring-lsws-for-php">LiteSpeed Web Server</link>
```
cmb69, line 217: It seems this link should be updated as well.

cmb69 (changes requested):
Thank you for the PR!

install/unix/litespeed.xml
istiak101: I am updating the phrase.

install/unix/litespeed.xml
istiak101: I couldn't find a better link from current documentation. Let me compare again and see if that's possible.

PR #1371, opened 2022-02-01
PDOStatement::bindParam maxLength is confusing

Karmel0x:
https://www.php.net/manual/en/pdostatement.bindparam.php
___
It has no effect on input parameters, but is even passed in the examples.

### DIFF
--- a/reference/pdo/pdostatement/bindparam.xml
+++ b/reference/pdo/pdostatement/bindparam.xml
@@ -66,7 +66,7 @@
         Explicit data type for the parameter using the <link linkend="pdo.constants"><literal>PDO::PARAM_*</literal>
         constants</link>.
         To return an INOUT parameter from a stored procedure, 
-        use the bitwise OR operator to set the PDO::PARAM_INPUT_OUTPUT bits
+        use the bitwise OR operator to set the <constant>PDO::PARAM_INPUT_OUTPUT</constant> bits
         for the <parameter>type</parameter> parameter.
        </para>
       </listitem>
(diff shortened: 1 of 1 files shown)


reference/pdo/pdostatement/bindparam.xml
```diff
@@ -78,6 +78,7 @@
         Length of the data type. To indicate that a parameter is an OUT
         parameter from a stored procedure, you must explicitly set the
         length.
+        Meaningful only when <parameter>type</parameter> parameter is PDO::PARAM_INPUT_OUTPUT.
```
kamil-tekiela, line 81:
```suggestion
        Meaningful only when <parameter>type</parameter> parameter is <literal>PDO::PARAM_INPUT_OUTPUT</literal>.
```

kamil-tekiela (changes requested):
Thanks. This is a well-spotted mistake. Could you please wrap the constant name in the literal tag?

reference/pdo/pdostatement/bindparam.xml
cmb69: Isn't `<constant>` more appropriate here?

kamil-tekiela: Ahh yes. I have been ill in the past couple of days so forgive my mistakes.

Karmel0x: https://github.com/php/doc-en/blob/be0ed69b711ec94903e802ffa67d80845553c35d/reference/pdo/pdostatement/bindparam.xml#L69
then it should be here too ?

kamil-tekiela (approved):

kamil-tekiela:
Thanks!

PR #1378, opened 2022-02-06
Fix ini directive name for uopz extension

sozpinar:
Directive `uopz.overloads` was misspelled as `uopz.exit` in the explanation section.


cmb69:
Good catch; thank you!

PR #1379, opened 2022-02-06
Add return value "AVIF Support" to gd_info()

Rotzbua:
Available since PHP 8.1.
Other related documentation: #983

### DIFF
--- a/reference/image/functions/gd-info.xml
+++ b/reference/image/functions/gd-info.xml
@@ -93,6 +93,12 @@
        <entry><type>bool</type> value.  &true;
         if <literal>WebP</literal> support is included.</entry>
       </row>
+      <row>
+       <entry>AVIF Support</entry>
+       <entry><type>bool</type> value.  &true;
+        if <literal>AVIF</literal> support is included.
+        Available as of PHP 8.1.0.</entry>
+      </row>
      </tbody>
     </tgroup>
    </table>
@@ -133,6 +139,8 @@ array(10) {
   bool(false)
   ["WebP Support"]=>
   bool(false)
+  ["AVIF Support"]=>
+  bool(false)
 }
 ]]>
     </screen>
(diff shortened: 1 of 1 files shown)


reference/image/functions/gd-info.xml
```diff
@@ -93,6 +93,11 @@
        <entry><type>bool</type> value.  &true;
         if <literal>WebP</literal> support is included.</entry>
       </row>
+      <row>
+       <entry>AVIF Support</entry>
+       <entry><type>bool</type> value.  &true;
+        if <literal>AVIF</literal> support is included.</entry> Available since PHP 8.1.
```
cmb69, line 99: When referring to a specific version of PHP, "since" should not be used. "As of" should be used in this case.
```suggestion
        if <literal>AVIF</literal> support is included.</entry> Available as of PHP 8.1.0.
```

cmb69 (commented):
Thank you!

reference/image/functions/gd-info.xml
```diff
@@ -93,6 +93,11 @@
        <entry><type>bool</type> value.  &true;
         if <literal>WebP</literal> support is included.</entry>
       </row>
+      <row>
+       <entry>AVIF Support</entry>
+       <entry><type>bool</type> value.  &true;
+        if <literal>AVIF</literal> support is included.</entry> Available as of PHP 8.1.0.
+      </row>
```
cmb69, line 100:
```suggestion
       </entry>
      </row>
```

Rotzbua: Thanks for the hint.
I think the line
>        if <literal>AVIF</literal> support is included.</entry> Available as of PHP 8.1.0.
should be replaced by
>        if <literal>AVIF</literal> support is included. Available as of PHP 8.1.0.</entry>
?

cmb69: Ah, indeed!

cmb69:
Thank you!

PR #1384, opened 2022-02-07
Fix argument type in EventBufferEvent::setCallback() method and make it mixed instead of string. Add missed $arg argument to signature of EventBufferEvent::__construct().

tnsoftbear:
$arg type should be mixed, like it is defined at other pages, e.g. in https://www.php.net/manual/en/eventbufferevent.about.callbacks.php
string type doesn't have big sense, because we want to pass there more complex data structures, like array.

EventBufferEvent::__construct() signature should contain the 7th argument "mixed $arg = null".


cmb69:
Thank you!

PR #1386, opened 2022-02-08
Fix parameter type for eio_readlink

kelunik:
See https://github.com/rosmanov/pecl-eio/blob/5291fa10d0ebd8c3454475957c32748920167cf2/php8/php_eio.stub.php#L101


cmb69:
Thank you! That `?mixed` in the stubs doesn't make sense, so I think we should go with `mixed` only in the docs.

PR #1389, opened 2022-02-09
Grammer fix for paamayim-nekudotayim.xml docs

thegreatzeus:
The word "parents" in "When an extending class overrides the parents definition of a method,
  PHP will not call the parent's method." needs to be changed to "parent's".


cmb69:
Thank you!

PR #1390, opened 2022-02-09
ReflectionClass::isEnum method

the-toster:


reference/reflection/reflectionclass/isenum.xml
```diff
@@ -0,0 +1,59 @@
...
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  &no.function.parameters;
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &true; if this is an enum, &false; otherwise.
+   Returns &null; in case of an error.
```
cmb69, line 33: I don't think that this method can ever return `null`, so
```suggestion
```

reference/reflection/reflectionclass/isenum.xml
```diff
@@ -0,0 +1,59 @@
...
+  &reftitle.description;
+  <methodsynopsis>
+   <modifier>public</modifier> <type>bool</type><methodname>ReflectionClass::isEnum</methodname>
+   <void />
+  </methodsynopsis>
+  <para>
+
+  </para>
+
+  &warn.undocumented.func;
```
cmb69, line 20: I suggest to add a short description, so that `&warn.undocumented.func;` can be removed right away.

cmb69 (commented):
Thank you for the PR! There should be a respective entry in [versions.xml](https://github.com/php/doc-en/blob/012af70e473cb3365637d6327114810d47e3de3d/reference/reflection/versions.xml#L181); I think the method is available as of PHP 8.1.0.

@Crell may want to review.

Crell:
I concur with both of @cmb69's points.  @iluuu1994 ?

reference/reflection/reflectionclass/isenum.xml
```diff
@@ -0,0 +1,55 @@
...
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <modifier>public</modifier> <type>bool</type><methodname>ReflectionClass::isEnum</methodname>
+   <void />
+  </methodsynopsis>
+  <para>
+   Checks if a class is an enum.
```
cmb69, line 17: Since enums are a pretty new concept in PHP, maybe it's good to add a link:
```suggestion
   Checks if a class is an <link linkend="language.enumerations">enum</link>.
```

Other than that, the PR looks good to me.

the-toster:
Thanks for feedback, I updated PR accordingly, please check it.

PR #1399, opened 2022-02-11
Should use constant time sodium_bin2hex in examples

carnage:
Examples should use best practice, in this case it should show using the constant time sodium implementation of bin2hex and not the standard PHP version. This helps prevent timing attacks in code.


cmb69:
While this is not necessary here, it makes sense to make readers aware of these functions, so thank you for the PR!

Issue #1403, opened 2022-02-13
PharData::__construct not in PHP 8 ?


marcimat:
From manual page: https://php.net/phardata.construct

Hi there. I’m newbe on PHP documentation works here so please be kind :)

I’m confused to see that PharData::__construct is not written PHP 8 compatible on the documentation (but seems working)

Also why is there  `_construct` and not `__construct`  on line 
`<function name="phardata::_construct" from="PHP 5 &gt;= 5.3.0, PHP 7, PECL phar &gt;= 2.0.0"/>` 
there https://github.com/php/doc-en/blob/master/reference/phar/versions.xml#L80 ?

Thanks for answers ! Have good day.

marcimat:
Ok. so It seems to work on local.

- I see that https://github.com/php/phd/blob/master/tests/php/data/version.xml#L5644 also references `phardata::_construct` but as this file hasn’t change since long time, I suppose this is not an issue. I do not have to worry about that ?
- and just to be sure : if I want that correction to be on french language too, I have nothing more to do, right ? This specific file seems shared between languages. I built both `en` and `fr` documentation, and I see the change on both.
- I don’t understand how the link is done between URL key and the name key on the versions.xml file… I was afraid that changing key  `pharfileinfo_compress` to `pharfileinfo::compress` will impact somehow the doc, but it don’t seems to… I suspect that `&reference.phar.PharFileInfo.compress;` on entities.PharFileInfo.xml is used and can find the ` xml:id="pharfileinfo.compress"`, but how could it found the old `pharfileinfo_compress` ? Does it try multiple name solutions ? and so writing one or other works ?

For (my) memory I used that to test both languages en and fr
```
# en
php doc-base/configure.php
php phd/render.php --docbook doc-base/.manual.xml --package PHP --format php
# fr
php doc-base/configure.php --with-lang=fr --output=doc-base/.manual.fr.xml
php phd/render.php --docbook doc-base/.manual.fr.xml --package PHP --format php --o=output-fr
cd web-php/manual
 ln -s ../../output-fr/php-web fr
```

marcimat:
And so is #1405

cmb69:
> I see that https://github.com/php/phd/blob/master/tests/php/data/version.xml#L5644 also references phardata::_construct but as this file hasn’t change since long time, I suppose this is not an issue.

Right. This is a test file for PhD, but I think all (or at least some of) these tests are broken anyway.

[...]

PR #1405, opened 2022-02-13
Issue #1403 : PharData and PharFileInfo methods missing PHP8 compatibility 

marcimat:
- PharData::__construct and PharData::__destruct exists in PHP8.
- PharFileInfo::compress and PharFileInfo::decompress also.


Girgias:
Thank you!

PR #1406, opened 2022-02-13
Update usort.xml

aleksandr-shevchenko:


reference/array/functions/usort.xml
```diff
@@ -161,9 +161,9 @@ $fruits[2]: lemons
 <![CDATA[
 <?php
 class TestObj {
-    var $name;
+    public string $name;
```
cmb69, line 164: The manual is supposed to document PHP 7.0.0 - current. Since typed properties are only available as of PHP 7.4.0, it might be best to omit the type here:
```suggestion
    public name;
```

cmb69 (commented):
Thank you for the PR! Besides the suggestion below, this looks good to me.

reference/array/functions/usort.xml
aleksandr-shevchenko: I don't think it's necessary to stick with old versions.

saundefined: Thanks for the PR!

The code should be compatible with the minimum supported PHP version of the manual (which is 7.0.0 now, although the version is EOL).
When the minimum supported version will be upgraded, the code can be upgraded too :)

Girgias: The thing is the docs *must* document and provide examples for all documented versions.
So either the type is dropped, or the example needs to be duplicated to cater to versions which do not have typed properties.

aleksandr-shevchenko: Accept all my edits and change the name of the example "Example #3 usort() example using a member function of an object" to "Example #3 usort() example using a member function of an object (for PHP >= 7.4)".
The example that is now in the documentation does not work under PHP 8.1.2 due to the lack of a constructor.

aleksandr-shevchenko: > > Temporary solutions are permanent, […]
> 
> Apparently, your PR contradicts this theory. :)

"Temporary solutions are permanent" I write to comment this suggestion

> The code should be compatible with the minimum supported PHP version of the manual (which is 7.0.0 now, although the version is EOL).
> When the minimum supported version will be upgraded, the code can be upgraded too :)

Issue #1408, opened 2022-02-14
Deprecation of libxml_disable_entity_loader() should note use of LIBXML_NONET constant


andrewnicols:
From manual page: https://php.net/function.libxml-disable-entity-loader

---

libxml_disable_entity_loader() was deprecated in https://github.com/php/php-src/commit/e0fa48f69dd14b52c8f1b2904ac7bd30472849a8 with the note:

> Since PHP now requires libxml >= 2.9.0 external entity loading no longer
> needs to be disabled to prevent these attacks. It is disabled by default.

The [documentation](https://php.net/function.libxml-disable-entity-loader) for `libxml_disable_entity_loader` also notes:

> However, as of libxml 2.9.0 entity substitution is disabled by default, so there is no need to disable the loading of external entities, unless there is the need to resolve internal entity references with LIBXML_NOENT.

However, there is no mention of the many ways it can be enabled, or how to make it safe where it has been enabled.

Previously you could call code such as:
```
libxml_disable_entity_loader(true);
$document = new DOMDocument();
$document->loadXML($xml, LIBXML_DTDLOAD | LIBXML_DTDATTR);
```

The presence of either the `DTDLOAD` or `DTDATTR` constants would normally cause the entity to be fetched, but the presence of `libxml_disable_entity_loader` overrdies and prevents this.

Now, after removing the deprecated call, the entity is fetched.

Whilst the honus is on the developer to understand the change to libxml, the attack vector of an XXE is not sufficiently clear and could be significantly improved in this documentation. The list of [libxml constants](https://www.php.net/manual/en/libxml.constants.php) does not adequately explain the risk of fetching the DTD, or explain that the behaviour of these flags has changed with the deprecation and removal of `libxml_disable_entity_loader`.

I would recommend that:
* the `libxml_disable_entity_loader` docs be updated to note that presence of the relevant `LIBXML_` flags will override the default behaviour to _not_ load the entity with advice on how to mitigate this; and
* the [Predefined Constants](https://www.php.net/manual/en/libxml.constants.php) docs be updated to provide more information on the meaning of the various flags, and their impact.

andrewnicols:
From a quick test, the following constants cause the dtd to be loaded:
* `LIBXML_DTDLOAD`
* `LIBXML_DTDATTR`
* `LIBXML_DTDVALID`

I _believe_ that they are safe from XXE attack _unless_ the `LIBXML_NOENT` is also specified.
[...]

PR #1409, opened 2022-02-14
Update libxml2 docs relating to entity validation

andrewnicols:
This commit adds warning that the `LIBXML_DTD_VALID` constant is also
susceptible to XXE exploit, and that the use of `LIBXML_DTDATTR` or
`LIBXML_DTDLOAD` constants will cause external entity fetching.

Also adds a note to the deprecated libxml_disable_entity_loader function
to note that LIBXML_NONET will prevent loading of external entities.

Prior to deprecation of the `libxml_disable_entity_loader` function, the
use of this function prevented both XXE via DTD parsing, and fetching of
external entities as specified by any of the above constants.

Fixes GH-1408.

### DIFF
--- a/reference/libxml/constants.xml
+++ b/reference/libxml/constants.xml
@@ -46,6 +46,12 @@
     <simpara>
      Default DTD attributes
     </simpara>
+    <caution>
+     <simpara>
+      Enabling loading of DTD attributes will enable fetching of external entities.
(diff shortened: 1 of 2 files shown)


nielsdos (changes requested):
This is a nice improvement, thanks.
It is almost right, right now you say in multiple places that LIBXML_NONET prevents external entity loading, but it only prevents external entity loading over the network, so the current wording is a bit misleading.

PR #1414, opened 2022-02-17
Added passing null to array functions to backward incompatible changes

mmcev106:
Is this already on the list somewhere and I'm just missing it?

### DIFF
--- a/appendices/migration80/incompatible.xml
+++ b/appendices/migration80/incompatible.xml
@@ -301,6 +301,9 @@ function test(int $arg = null) {}
         Attempting to access unqualified constants which are undefined.
         Previously, unqualified constant accesses resulted in a warning and were interpreted as strings.
        </member>
+       <member>
+        Attempting to pass &null; as the array parameter to one of the array functions.
+       </member>
       </simplelist>
      </para>
      <para>


cmb69:
This is not particular to array functions, but rather all internal functions are now more strict regarding unexpected types. This is already [documented elsewhere](https://www.php.net/manual/en/functions.internal.php):

So if you pass `null` where an `array` is expected, the behavior is actually undefined. As such, there is no incompatible change. In my opinion, there is nothing to document in the migration guide.

appendices/migration80/incompatible.xml
```diff
@@ -301,6 +301,9 @@ function test(int $arg = null) {}
         Attempting to access unqualified constants which are undefined.
         Previously, unqualified constant accesses resulted in a warning and were interpreted as strings.
        </member>
+       <member>
+        Attempting to pass &null; as the array parameter to one of the array functions.
```
kamil-tekiela, line 305: I don't understand what this means. Could you give an example? Preferable using https://3v4l.org/ What change are you describing here?

mmcev106: Should the paragraph @cmb69 mentioned should be moved/copied to `incompatible.xml`, since calls crash where they didn't previously?  It also might not hurt to mention `null` explicitly in that paragraph.  As written, I would have assumed `null` was still allowed since it is common practice to pass `null` in place of a missing parameter/variable/object.

I'm happy to make a new PR if you all think the above makes sense.

mmcev106: It looks like `null` behavior is different between string & array functions.  Passing `null` to string functions displays a deprecation warning in 8.1+, while passing `null` to array functions causes a fatal error in 8.0+.  This means only the array functions are backward incompatible.  Does that mean this PR is correct as-is?

kamil-tekiela: No, the PR is not ok as it is now. Please read again our replies carefully. Scalar parameters accepted null values until they were deprecated in PHP 8.1. Everything else was an undefined behaviour and now thrown fatal error in PHP 8.0. It has nothing to do with string or array functions (regardless of what you mean by those terms)

mmcev106: I have re-read your comments, and am still unclear on how to move forward.  Please advise, since you obviously understand the subtleties in this area much better than I do.

kamil-tekiela:
I think this PR should be closed. Bo offense to the author, but at the
current state this change is misleading and unhelpful. This could still be
added to the manual but as I said previously. We could open an issue.

Can someone close it as I am without access to a computer?

cmb69:
> I think this PR should be closed.

Agreed, as per [my previous comment](https://github.com/php/doc-en/pull/1414#issuecomment-1044302038).

PR #1418, opened 2022-02-20
Improve documentation of string encoding conversion functions

IMSoP:
- Move utf8_encode and utf8_decode into the strings chapter, since
  they were moved out of the XML extension in 7.2
- Recommend mb_convert_encoding, iconv, and UConverter::transcode
  when mentioning encoding in passing
- Document UConverter::transcode, based on examination of source
  and upstream ICU docs
- Make the language used more consistent, e.g. "convert" rather
  than "encode"/"decode", "encoding" rather than "charset"

### DIFF
--- a/reference/intl/uconverter/transcode.xml
+++ b/reference/intl/uconverter/transcode.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="uconverter.transcode" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>UConverter::transcode</refname>
-  <refpurpose>Convert string from one charset to another</refpurpose>
+  <refpurpose>Convert a string from one character encoding to another</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
@@ -16,11 +16,8 @@
    <methodparam choice="opt"><type class="union"><type>array</type><type>null</type></type><parameter>options</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-
+   Converts <parameter>str</parameter> from <parameter>fromEncoding</parameter> to <parameter>toEncoding</parameter>.
   </para>
-
-  &warn.undocumented.func;
-
  </refsect1>
 
  <refsect1 role="parameters">
@@ -30,7 +27,7 @@
     <term><parameter>str</parameter></term>
     <listitem>
      <para>
-      
+      The &string; to be converted.
      </para>
     </listitem>
    </varlistentry>
@@ -38,7 +35,7 @@
     <term><parameter>toEncoding</parameter></term>
     <listitem>
      <para>
-      
+      The desired encoding of the result.
      </para>
     </listitem>
    </varlistentry>
@@ -46,7 +43,7 @@
     <term><parameter>fromEncoding</parameter></term>
     <listitem>
      <para>
-      
+      The current encoding used to interpret <parameter>str</parameter>.
      </para>
     </listitem>
    </varlistentry>
@@ -54,7 +51,15 @@
     <term><parameter>options</parameter></term>
     <listitem>
      <para>
-      
+      An optional &array;, which may contain the following keys:
+      <simplelist>
+       <member>
+        <literal>'to_subst'</literal> - the substitution character to use
+        in place of any character of <parameter>str</parameter> which cannot
+        be encoded in <parameter>toEncoding</parameter>. If specified, it must
+        represent a single character in the target encoding.
+       </member>
+      </simplelist>
      </para>
     </listitem>
    </varlistentry>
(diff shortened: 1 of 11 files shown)


reference/intl/uconverter/transcode.xml
```diff
@@ -16,45 +16,58 @@
...
  </refsect1>
 
  <refsect1 role="parameters">
   &reftitle.parameters;
   <variablelist>
    <varlistentry>
-    <term><parameter>str</parameter></term>
+    <term>
+     <parameter>str</parameter>
+    </term>
```
cmb69, line 29: Please stick with the single line formatting; this is also produced by [get_stub.php](https://github.com/php/php-src/blob/master/build/gen_stub.php).

reference/strings/functions/utf8-decode.xml
```diff
@@ -4,8 +4,8 @@
  <refnamediv>
```
cmb69, line 4: Maybe we should add a note that these functions where part of the XML extension prior to PHP 7.2.0, so that users are aware that the functions might not have been available.

cmb69: Ah, right! That's clear enough for me.

reference/strings/functions/utf8-encode.xml
```diff
@@ -55,6 +63,29 @@
   </para>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <example>
+   <title>Basic example</title>
+   <programlisting role="php">
+    <![CDATA[
```
cmb69, line 71:
```suggestion
<![CDATA[
```

reference/strings/functions/utf8-encode.xml
```diff
@@ -86,8 +117,8 @@
    <simplelist>
     <member><function>utf8_decode</function> - Performs the reverse conversion</member>
     <member><function>mb_convert_encoding</function> - Converts between various character encodings, including UTF-8, ISO-8859-1 and Windows-1252</member>
-    <member><function>iconv</function> - Converts between various character encodings</member>
-    <member><function>recode_string</function> - Converts between various character encodings</member>
+    <member><methodname>UConverter::transcode</methodname> - Converts between various character encodings, including UTF-8, ISO-8859-1 and Windows-1252</member>
+    <member><function>iconv</function> - Converts between various character encodings, including UTF-8, ISO-8859-1 and Windows-1252</member>
```
cmb69, line 121: Same as above (duplicate descriptions).

reference/intl/uconverter/transcode.xml
```diff
@@ -30,31 +27,41 @@
...
-      
+      The current encoding of <parameter>str</parameter> to convert from.
      </para>
     </listitem>
    </varlistentry>
    <varlistentry>
-    <term><parameter>options</parameter></term>
+    <term>
+     <parameter>options</parameter>
+    </term>
```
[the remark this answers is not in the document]
IMSoP: Gah, I thought I'd caught all of those...

reference/intl/uconverter/transcode.xml
```diff
@@ -64,10 +71,110 @@
...
+   Returns the converted string&return.falseforfailure;.
   </para>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <example>
+   <title>Converting from UTF-8 to UTF-16 and back</title>
+   <programlisting role="php">
+    <![CDATA[
```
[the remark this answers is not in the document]
IMSoP: I wasn't sure whether to indent those; it does look better without. :)

reference/strings/functions/utf8-decode.xml
```diff
@@ -93,8 +130,8 @@
    <simplelist>
     <member><function>utf8_encode</function> - Performs the reverse conversion</member>
     <member><function>mb_convert_encoding</function> - Converts between various character encodings, including UTF-8, ISO-8859-1 and Windows-1252</member>
-    <member><function>iconv</function> - Converts between various character encodings</member>
-    <member><function>recode_string</function> - Converts between various character encodings</member>
+    <member><methodname>UConverter::transcode</methodname> - Converts between various character encodings, including UTF-8, ISO-8859-1 and Windows-1252</member>
+    <member><function>iconv</function> - Converts between various character encodings, including UTF-8, ISO-8859-1 and Windows-1252</member>
```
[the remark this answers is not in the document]
IMSoP: Ah, makes sense, and saves trying to come up with something to put there.

IMSoP (commented):
Thanks for the thorough review! I think I've caught all those bits now.

IMSoP:
I've re-worded the "from_encoding" and "to_encoding" parameter descriptions after @Crell's comment on #1419 (which should have been here; my fault). They were unnecessarily wordy, and didn't capture the fact that you're telling the function what to *assume* the encoding was.

cmb69:
Thank you very much!

PR #1419, opened 2022-02-20
[RFC] Deprecation and alternatives for utf8_encode and utf8_decode

IMSoP:
Proposed additional wording if https://wiki.php.net/rfc/remove_utf8_decode_and_utf8_encode passes

Builds on top of #1418

### DIFF
--- a/reference/iconv/functions/iconv.xml
+++ b/reference/iconv/functions/iconv.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="function.iconv" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>iconv</refname>
-  <refpurpose>Convert string to requested character encoding</refpurpose>
+  <refpurpose>Convert a string from one character encoding to another</refpurpose>
  </refnamediv>
  
  <refsect1 role="description">
@@ -15,8 +15,7 @@
    <methodparam><type>string</type><parameter>string</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Performs a character set conversion on the string
-   <parameter>string</parameter> from <parameter>from_encoding</parameter>
+   Converts <parameter>string</parameter> from <parameter>from_encoding</parameter>
    to <parameter>to_encoding</parameter>.
   </para>
  </refsect1>
@@ -29,7 +28,7 @@
      <term><parameter>from_encoding</parameter></term>
      <listitem>
       <para>
-       The input charset.
+       The current encoding of <parameter>string</parameter> to convert from.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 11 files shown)


reference/iconv/functions/iconv.xml
```diff
@@ -29,22 +28,22 @@
...
      <term><parameter>to_encoding</parameter></term>
      <listitem>
       <para>
-       The output charset.
+       The encoding which <parameter>string</parameter> should be converted to.
       </para>
       <para>
-       If you append the string <literal>//TRANSLIT</literal> to
+       If the string <literal>//TRANSLIT</literal> is appended to
        <parameter>to_encoding</parameter> transliteration is activated. This
```
Crell, line 43:
```suggestion
       <parameter>to_encoding</parameter>, then transliteration is activated. This
```

reference/intl/uconverter/transcode.xml
```diff
@@ -16,45 +16,58 @@
...
    </varlistentry>
    <varlistentry>
-    <term><parameter>toEncoding</parameter></term>
+    <term>
+     <parameter>toEncoding</parameter>
+    </term>
     <listitem>
      <para>
-      
+      The encoding which <parameter>str</parameter> should be converted to.
```
Crell, line 42:
```suggestion
      The encoding to which <parameter>str</parameter> should be converted.
```

reference/mbstring/functions/mb-convert-encoding.xml
```diff
@@ -31,31 +30,36 @@
...
+       The &string; or &array; to be converted.
       </para>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><parameter>to_encoding</parameter></term>
      <listitem>
       <para>
-       The type of encoding that <parameter>string</parameter> is being converted to.
+       The encoding which <parameter>string</parameter> should be converted to.
```
Crell, line 41:
```suggestion
       The encoding to which <parameter>string</parameter> should be converted.
```

reference/strings/functions/utf8-decode.xml
```diff
@@ -0,0 +1,264 @@
...
+     <![CDATA[
+sring(1) "?"
+]]>
+     </screen>
+    </informalexample>
+   </para>
+  </note>
+ </refsect1>
+
+ <refsect1 role="changelog">
```
Crell, line 207: Should the deprecation be mentioned in the changelog, too?

reference/strings/functions/utf8-encode.xml
```diff
@@ -0,0 +1,230 @@
...
+  </methodsynopsis>
+  <para>
+   This function converts the string <parameter>string</parameter> from the
+   <literal>ISO-8859-1</literal> encoding to <literal>UTF-8</literal>.
+  </para>
+
+  <note>
+   <para>
+    This function does not attempt to guess the current encoding of the provided
+    string, it assumes it is encoded as ISO-8859-1 (also known as "Latin 1")
```
Crell, line 27:
```suggestion
    string. It assumes it is encoded as ISO-8859-1 (also known as "Latin 1")
```

reference/intl/uconverter/transcode.xml
IMSoP: Reading it back, all of these sentences were unnecessarily torturous. I've come up with a new wording across all three functions, which I think is less wordy and more precise.

reference/strings/functions/utf8-decode.xml
IMSoP: Yes, probably. Also, looks like I missed an attribute to make it list as deprecated in indexes.

reference/strings/functions/utf8-encode.xml
IMSoP: I compromised and used a semi-colon 😜

reference/intl/uconverter/transcode.xml
IMSoP: Actually, that's part of https://github.com/php/doc-en/pull/1418 which has the changes which can land even if deprecation doesn't go ahead.

IMSoP:
Sorry, I've made a mess of this; my intention was to compare two branches, but I ended up with two copies of each branch, confusing everything.

Re-opened here for now: https://github.com/IMSoP/doc-en/pull/1

Issue #1422, opened 2022-02-21
Investigate/Fix crypt(), password_hash(), and password_verify() documentation


jonathanbossenger:
From a phpdoc discussion

https://news-web.php.net/php.doc/969388073

https://news-web.php.net/php.doc/969388075

Creating this ticket so I can investigate, and create a PR (if needed)

dregad:
While you're investigating/fixing crypt() documentation, you may want to have a look at #1533.

Girgias:
Fixed by #1424

PR #1424, opened 2022-02-21
`password_verify` is capable of verifying `crypt` hashes

Sc00bz:
> `password_verify` is compatible with `crypt`. Therefore, password hashes created by `crypt` can be used with `password_verify`.

Was added to `password_verify` and `crypt` and removed from `password_hash`. I believe it was added to `password_hash` instead of `password_verify`. Also fixed an error in `crypt` documentation confusing bytes and characters.

### DIFF
--- a/reference/strings/functions/crypt.xml
+++ b/reference/strings/functions/crypt.xml
@@ -20,7 +20,10 @@
   <para>
    <function>crypt</function> will return a hashed string using the
    standard Unix <abbrev>DES</abbrev>-based algorithm or
-   alternative algorithms.
+   alternative algorithms. <function>password_verify</function> is
+   compatible with <function>crypt</function>. Therefore, password
+   hashes created by <function>crypt</function> can be used with
+   <function>password_verify</function>.
   </para>
   <para>
    The <parameter>salt</parameter> parameter is optional. However, <function>crypt</function> creates a weak hash without the <parameter>salt</parameter>, and raises an <constant>E_NOTICE</constant> error without it. Make sure to specify a strong enough salt for better security. 
--- a/reference/strings/functions/crypt.xml
+++ b/reference/strings/functions/crypt.xml
@@ -61,7 +61,7 @@
     <simpara>
      <constant>CRYPT_EXT_DES</constant> - Extended DES-based hash. The "salt" is a
      9-character string consisting of an underscore followed by 4 characters of iteration count
-     and 4 characters of salt. Each of these 4 characters encode 24 bits, least significant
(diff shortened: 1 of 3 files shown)


Sc00bz:
Fixes #1422

reference/strings/functions/crypt.xml
```diff
@@ -57,9 +60,9 @@
    <listitem>
     <simpara>
      <constant>CRYPT_EXT_DES</constant> - Extended DES-based hash. The "salt" is a
-     9-character string consisting of an underscore followed by 4 bytes of iteration count and
-     4 bytes of salt. These are encoded as printable characters, 6 bits per character, least
-     significant character first. The values 0 to 63 are encoded as "./0-9A-Za-z". Using invalid
+     9-character string consisting of an underscore followed by 4 characters of iteration count
+     and 4 characters of salt. Each of these 4 characters encode 24 bits, least significant
+     character first. The values 0 to 63 are encoded as "./0-9A-Za-z". Using invalid
```
cmb69, line 65: nit:
```suggestion
     character first. The values <literal>0</literal> to <literal>63</literal> are encoded as <literal>"./0-9A-Za-z"</literal>. Using invalid
```

cmb69 (commented):
Thank you!

reference/strings/functions/crypt.xml
Sc00bz: That's the sentence after my edit. Also should we get rid of the quotes or put them outside the <literal>?

cmb69: Yes, please drop the quotes.

reference/strings/functions/crypt.xml
```diff
@@ -57,9 +60,9 @@
    <listitem>
     <simpara>
      <constant>CRYPT_EXT_DES</constant> - Extended DES-based hash. The "salt" is a
-     9-character string consisting of an underscore followed by 4 bytes of iteration count and
-     4 bytes of salt. These are encoded as printable characters, 6 bits per character, least
-     significant character first. The values 0 to 63 are encoded as "./0-9A-Za-z". Using invalid
+     9-character string consisting of an underscore followed by 4 characters of iteration count
+     and 4 characters of salt. Each of these 4 characters encode 24 bits, least significant
```
TimWolla, line 64: A single character cannot encode more than 8 bits. Should this read:
```suggestion
     and 4 characters of salt. Together these 4 characters encode 24 bits, least significant
```

cmb69: Or maybe "each of these 4 character strings encode 24 bits"?

cmb69:
Thank you!

PR #1426, opened 2022-02-22
Add documentation for the `pm.status_listen` FPM INI option

ste93cry:
While looking at the content of the `www.conf` file of FPM, I found out a quite interesting option that isn't documented publicy. With this PR I'm adding the missing bits of doc. About the description, I took the one that was in the file itself and rewritten a small part of it to something less repetitive

### DIFF
--- a/install/fpm/configuration.xml
+++ b/install/fpm/configuration.xml
@@ -492,6 +492,22 @@
        </para>
       </listitem>
      </varlistentry>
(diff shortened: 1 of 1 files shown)


ste93cry:
Apparently this has been documented with aa4b1bf3169650c0a89aa68df6d1c8bf15d6ca29

PR #1429, opened 2022-02-24
Disambiguate semantics of `search_value`

arielallon:
The definition of `search_value` in "Parameters" is a bit ambiguous. "these values" _could_ be interpreted to mean that if an array is passed as the `$search_value`, keys of elements of `$search_value` that match _any_ of those `$array` elements will be returned.
Note, the "Description" section does not have this ambiguity as it states "that value" instead of "those values".


cmb69:
Spot on, thank you!

PR #1430, opened 2022-02-25
Improve description of `empty` return value

theodorejb:
Previously it was written from the negative perspective, which was confusing to understand.


cmb69:
Thank you!

PR #1436, opened 2022-03-02
Add default value to setFetchMode's $constructorArgs with FETCH_CLASS

Seldaek:
As per https://github.com/php/php-src/blob/888bec1be33818c91a3359319d6395fe5b700053/ext/pdo/pdo_stmt.c#L1764-L1822 and specifically https://github.com/php/php-src/blob/888bec1be33818c91a3359319d6395fe5b700053/ext/pdo/pdo_stmt.c#L1787


cmb69:
Thank you!

PR #1438, opened 2022-03-03
Improve readability of htmlentities.xml

morganarnel:
This line now reads better when you're in a hurry (imo), and doesn't give the impression that ALL characters will be converted to entities.

### DIFF
--- a/reference/strings/functions/htmlentities.xml
+++ b/reference/strings/functions/htmlentities.xml
@@ -17,7 +17,11 @@
   </methodsynopsis>
   <para>
    This function is identical to <function>htmlspecialchars</function> in all
-   ways, except with <function>htmlentities</function>, all characters which have HTML character entity equivalents are translated into these entities. The <function>get_html_translation_table</function> function can be used to return the translation table used dependent upon the provided <parameter>flags</parameter> constants.
+   ways, except with <function>htmlentities</function>, all characters which
+   have HTML character entity equivalents are translated into these entities.
+   The <function>get_html_translation_table</function> function can be used
+   to return the translation table used dependent upon the provided
+   <parameter>flags</parameter> constants.
   </para>
   <para>
    If you want to decode instead (the reverse) you can use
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR! I wonder whether it would be better to hint at `get_html_translation_table()` instead of referring to the rather vague "have HTML character entity equivalents", and clarifying by mentioning the flags.

cmb69:
Not quite what I had in mind, but looks good to me. Thank you!

reference/strings/functions/htmlentities.xml
```diff
@@ -17,8 +17,7 @@
   </methodsynopsis>
   <para>
    This function is identical to <function>htmlspecialchars</function> in all
-   ways, except with <function>htmlentities</function>, all characters which
-   have HTML character entity equivalents are translated into these entities.
+   ways, except with <function>htmlentities</function>, all characters which have HTML character entity equivalents are translated into these entities. The <function>get_html_translation_table</function> function can be used to return the translation table used dependent upon the provided <parameter>flags</parameter> constants.
```
Crell, line 20:
```suggestion
   ways, except with <function>htmlentities</function>, all characters which have HTML character
   entity equivalents are translated into these entities. The <function>get_html_translation_table</function>
   function can be used to return the translation table used dependent upon the provided
   <parameter>flags</parameter> constants.
```

Text should be wrapped consistently.

Crell, line 20: GitHub isn't letting me do a multi-line suggestion here, but the text should be hard-wrapped consistently.

cmb69, line 20:
```suggestion
   ways, except with <function>htmlentities</function>, all characters which
   have HTML character entity equivalents are translated into these entities.
   The <function>get_html_translation_table</function> function can be used
   to return the translation table used dependent upon the provided
   <parameter>flags</parameter> constants.
```

cmb69:
Thank you!

PR #1451, opened 2022-03-11
PHP 8.1 | Update hash_algos() documentation

jrfnl:
> **MurmurHash3**
>
> Added support for MurmurHash3 with streaming support. The following variants are implemented:
> * murmur3a, 32-bit hash
> * murmur3c, 128-bit hash for x86
> * murmur3f, 128-bit hash for x64
>
> **xxHash**
>
> Added support for xxHash. The following variants are implemented:
> * xxh32, 32-bit hash
> * xxh64, 64-bit hash
> * xxh3, 64-bit hash
> * xxh128, 128-bit hash

Refs:
* https://www.php.net/manual/en/migration81.new-features.php#migration81.new-features.hash.murmurhash3
* https://github.com/php/php-src/pull/6059
* https://github.com/php/php-src/commit/72e91e9fc82b85d0c5a6ff043cd1b2daee6c26d0
* https://www.php.net/manual/en/migration81.new-features.php#migration81.new-features.hash.xxhash
* https://github.com/php/php-src/pull/6524
* https://github.com/php/php-src/commit/23590f7c5348de4f8b69da81439b7c42c7ba084c

### DIFF
--- a/reference/hash/functions/hash-algos.xml
+++ b/reference/hash/functions/hash-algos.xml
@@ -39,6 +39,12 @@
      </thead>
      <tbody>
       <row>
+       <entry>8.1.0</entry>
+       <entry>
+        Support for MurmurHash3 and xxHash algorithms has been added.
+       </entry>
+      </row>
+      <row>
        <entry>7.4.0</entry>
        <entry>
         Support for crc32c has been added.
@@ -56,14 +62,14 @@
    </informaltable>
   </para>
  </refsect1><!-- }}} -->
- 
+
  <refsect1 role="examples">
   &reftitle.examples;
   <para>
    <example>
     <title><function>hash_algos</function> example</title>
     <para>
-     As of PHP 7.4.0, <function>hash_algos</function> will return the
+     As of PHP 8.1.0, <function>hash_algos</function> will return the
      following list of algorithm names.
     </para>
     <programlisting role="php">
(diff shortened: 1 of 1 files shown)


reference/hash/functions/hash-algos.xml
```diff
@@ -116,21 +122,27 @@ Array
...
+    [49] => haval256,3
+    [50] => haval128,4
+    [51] => haval160,4
+    [52] => haval192,4
+    [53] => haval224,4
+    [54] => haval256,4
+    [55] => haval128,5
+    [56] => haval160,5
+    [57] => haval192,5
+    [58] => haval224,5 [59] => haval256,5
```
cmb69, line 145: Missing line break:
```suggestion
    [58] => haval224,5
    [59] => haval256,5
```

cmb69 (commented):
Thank you for the PR!

I'm not 100% convinced that it makes much sense to update the list of algorithms in the example output (it changes "all the time"), but I'm not against doing it.

reference/hash/functions/hash-algos.xml
jrfnl: Oops, missed one...  Thanks for catching that! Fixed now.

jrfnl:
> Thank you for the PR!
> 
> I'm not 100% convinced that it makes much sense to update the list of algorithms in the example output (it changes "all the time"), but I'm not against doing it.

I followed by example as the list was updated for the previous PHP 7.4 update. (though realized just now I forgot to update the "As of PHP ..." bit in the leading paragraph. Fixed that now too.

cmb69:
Thanks! I think we can commit as is, but we should revisit:
Screenshot 2022-03-11 141523

That "as of PHP …" should either go below the example script, or probably better be removed altogether, since the `&example.outputs.similar;` already hints that the exact output may be different.

PR #1454, opened 2022-03-11
Modernize examples on connect_error & connect_errno

kamil-tekiela:
This is an attempt to provide a code example that works both on PHP 8.0 and 8.1. Due to how broken this functionality is, I can't think of a reasonable example that would make sense. 

I removed the output as outputting this information is a really bad idea. 

I thought about putting an example such as 
```
<?php

mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
$mysqli = new mysqli();
try {
    /* @ is used to suppress warnings */
    @$mysqli->real_connect('localhost', 'fake_user', 'my_password', 'does_not_exist');
} catch (mysqli_sql_exception $e) {
    /* Use your preferred error logging method here */
    error_log('Connection error: ' . $mysqli->connect_error);
}
```
but in the end, I decided that the older one is more relevant. 

I also wanted to add a note about warnings and why we use silence operator despite error reporting turned off, but I couldn't think of the suitable wording.

### DIFF
--- a/reference/mysqli/mysqli/connect-errno.xml
+++ b/reference/mysqli/mysqli/connect-errno.xml
@@ -42,9 +42,10 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-/* @ is used to suppress default error messages */
-$mysqli = @new mysqli('localhost', 'fake_user', 'my_password', 'my_db');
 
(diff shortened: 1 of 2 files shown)


cmb69 (approved):
Thank you!

PR #1455, opened 2022-03-12
fixes typo

TENMAJKL:
This pull request fixes typo in chapter `User-defined functions`.


kamil-tekiela:
I feel like `thus` is more correct. I don't think `this` is grammatically correct. Can you provide your reasoning?

TENMAJKL:
Oh i thought its typo, i didn't know its real word.

PR #1459, opened 2022-03-13
Add missing param/return value sections

Girgias:
This, with #1457, fixes the remain (current) issues raised by the section order QA script #658.

After merging both of them we should add the check to CI so that a PR can't introduce section ordering issues.

### DIFF
--- a/reference/pdo/pdo/setattribute.xml
+++ b/reference/pdo/pdo/setattribute.xml
@@ -16,87 +16,171 @@
   </methodsynopsis>
 
   <para>
-   Sets an attribute on the database handle. Some of the available generic
+   Sets an attribute on the database handle. Some available generic
    attributes are listed below; some drivers may make use of
    additional driver specific attributes.
    Note that driver specific attributes <emphasis>must not</emphasis> be used
    with other drivers.
    <itemizedlist>
-    <listitem><para>
-     <literal>PDO::ATTR_CASE</literal>: Force column names to a specific case.
-     <itemizedlist>
-      <listitem><para>
-       <literal>PDO::CASE_LOWER</literal>: Force column names to lower case.
-      </para></listitem>
-      <listitem><para>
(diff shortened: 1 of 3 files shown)


reference/pdo/pdostatement/getattribute.xml
```diff
@@ -28,6 +28,22 @@
 
  </refsect1>
 
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>name</parameter></term>
```
Girgias, line 36: I wonder if this shouldn't be changed in php-src as it's weird it's the only one called ``name`` instead of ``attribute``.

Crell:
Seems OK to me.  Is there anything blocking a merge?

reference/pdo/pdo/setattribute.xml
```diff
@@ -16,97 +16,192 @@
...
+        <para>
+         <constant>PDO::CASE_LOWER</constant>: Force column names to lower case.
+        </para>
+       </listitem>
+       <listitem>
+        <para>
+         <constant>PDO::CASE_NATURAL</constant>: Leave column names as returned
+         by the database driver.
+        </para>
+       </listitem>
```
cmb69, line 39: Wouldn't a `<variablelist>` with nested `<simplelist>`s be more suitable?

Girgias: Probably, I was just fixing the formating of the existing listing, and I'm not sure how the ``<variablelist>`` should look like.

cmb69 (approved):
Thank you! The nested `<variablelist>`s are semantically cleaner, and so it's fine to go with this. My idea was to use `<simplelist>`s for the inner level, which might look nicer on the screen:
Screenshot 2022-04-02 162011
vs.
Screenshot 2022-04-02 162222

But as I said, I'm fine with the current status, too.

PR #1471, opened 2022-03-19
Update assert.xml

braders:
As per the PHP changelog, the functionality to use a string for `assert` has been removed in PHP 8.0: https://github.com/php/php-src/blob/69888c3ff1f2301ead8e37b23ff8481d475e29d2/UPGRADING#L350-L354

### DIFF
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -172,7 +172,7 @@
       <warning>
        <para>
         Using <type>string</type> as the <parameter>assertion</parameter> is
-        <emphasis>DEPRECATED</emphasis> as of PHP 7.2.
+        <emphasis>DEPRECATED</emphasis> as of PHP 7.2 and <emphasis>REMOVED</emphasis> as of PHP 8.0.0.
        </para>
       </warning>
      </listitem>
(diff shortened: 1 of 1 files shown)


reference/info/functions/assert.xml
```diff
@@ -172,7 +172,7 @@
       <warning>
        <para>
         Using <type>string</type> as the <parameter>assertion</parameter> is
-        <emphasis>DEPRECATED</emphasis> as of PHP 7.2.
+        <emphasis>DEPRECATED</emphasis> as of PHP 7.2 and <emphasis>REMOVED</emphasis> as of PHP 8.0.
```
cmb69, line 175: Let's be "precise" here:
```suggestion
        <emphasis>DEPRECATED</emphasis> as of PHP 7.2.0 and <emphasis>REMOVED</emphasis> as of PHP 8.0.0.
```

cmb69 (commented):
Thank you!

Please add a changelog entry as well (like the existing 7.2.0 entry).

braders:
@cmb69 Both comments have been addressed.

cmb69:
Thank you! I've tweaked a bit.

braders:
@cmb69 Thanks for merging. I've just noticed in the changelog `assert()()` is showing with double parenthesis which doesn't seem quite right?

cmb69:
Yeah; that has been fixed with 59af09939ab6e191c9836ea3b33b934e1a99a263.

PR #1473, opened 2022-03-21
Fix "see also" for ReflectionFunction::__toString

WinterSilence:


WinterSilence:
@kamil-tekiela no reason recoment deprecated/removed methods

kamil-tekiela:
It's not a recommendation. It's a see also. If a new function is a replacement for a removed function then it's a good idea to mention what the old function was.

WinterSilence:
@kamil-tekiela ok, roll back

kamil-tekiela (approved):
Thank you.

PR #1479, opened 2022-03-25
Update sqlsrv_rows_affected/db2_num_fields return type as docs say they can return false

Seldaek:


reference/ibm_db2/functions/db2-num-fields.xml
```diff
@@ -11,7 +11,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>db2_num_fields</methodname>
+   <type class="union"><type>int</type><type>false</type><methodname>db2_num_fields</methodname>
```
cmb69, line 14:
```suggestion
   <type class="union"><type>int</type><type>false</type></type><methodname>db2_num_fields</methodname>
```

cmb69 (commented):
Thank you!

PR #1480, opened 2022-03-25
Fix the return value description

colshrapnel:
The comment in the source code [goes](https://github.com/php/php-src/blob/04a4864b655b6ace1150e512179f23ec0d47566f/ext/pdo/pdo_stmt.c#L1152) as "if there are no more rows" and doesn't mention any failure.

### DIFF
--- a/reference/pdo/pdostatement/fetch.xml
+++ b/reference/pdo/pdostatement/fetch.xml
@@ -138,7 +138,7 @@
   &reftitle.returnvalues;
   <para>
    The return value of this function on success depends on the fetch type. In
-   all cases, &false; if there are no more rows.
+   all cases, &false; is returned on failure or if there are no more rows.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


cmb69:
Prior to PHP 8.0.0, the method also returned `false` in case of an invalid fetch mode (this throws `ValueError` as of PHP 8.0.0), and the docs are currently supposed to document PHP 7 and 8. Even as of PHP 8.0.0, there might be other conditions where `false` is returned (would need to carefully examine the code to be sure). Either we list all such failure cases, or we go with "returns &false; on failure (e.g. when there are no more rows)" or something like that.

colshrapnel:
@cmb69 thank you. So it's better to phrase it as "is returned on failure or if there are no more rows"?

cmb69:
That's good for me as well.

cmb69:
Thank you!

Issue #1485, opened 2022-03-28
PASSWORD_DEFAULT is '2y' but document says 'null'


shosho44:
From migration page: https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.password-algorithm-constants

This page includes the message:
> PASSWORD_DEFAULT was int 1; now is null

However, actual value is '2y'
`echo PASSWORD_DEFAULT;`
output: 2y

cmb69:
It [was `null` for a short time](https://3v4l.org/nN4Qi), but as of PHP 7.4.3 it has been [fixed](https://github.com/php/php-src/pull/5104) to be `"2y"`.

ssshotaro44:
Thank you! Can I make PR in order to write about this on migration manual?

cmb69:
> Can I make PR in order to write about this on migration manual?

Yes, sure. :)

ssshotaro44:
Thanks! I could make PR! ;)

vrana:
Fixed by #1488.

PR #1488, opened 2022-03-29
Add detailed explanation about PASSWORD_DEFAULT on manual

ssshotaro44:
related to #1485

### DIFF
--- a/appendices/migration74/incompatible.xml
+++ b/appendices/migration74/incompatible.xml
@@ -83,7 +83,7 @@
    <itemizedlist>
     <listitem>
      <simpara>
-      <constant>PASSWORD_DEFAULT</constant> was int 1; now is &null; as of PHP 7.4.3, it is string '2y'
+      <constant>PASSWORD_DEFAULT</constant> was int 1; now is string '2y' (in PHP 7.4.0, 7.41, and 7.4.2 it was &null;)
      </simpara>
     </listitem>
     <listitem>
--- a/appendices/migration74/incompatible.xml
+++ b/appendices/migration74/incompatible.xml
@@ -83,7 +83,7 @@
    <itemizedlist>
     <listitem>
      <simpara>
-      <constant>PASSWORD_DEFAULT</constant> was int 1; now is string '2y' (in PHP 7.4.0, 7.41, and 7.4.2 it was &null;)
+      <constant>PASSWORD_DEFAULT</constant> was int 1; now is string '2y' (in PHP 7.4.0, 7.4.1, and 7.4.2 it was &null;)
      </simpara>
     </listitem>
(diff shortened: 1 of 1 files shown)


appendices/migration74/incompatible.xml
```diff
@@ -83,7 +83,7 @@
    <itemizedlist>
     <listitem>
      <simpara>
-      <constant>PASSWORD_DEFAULT</constant> was int 1; now is &null;
+      <constant>PASSWORD_DEFAULT</constant> was int 1; now is &null; (version: 7.4.0 - 7.4.2), '2y' (version: 7.4.3 -)
```
cmb69, line 86: I'm super unsure how to express this in the migration guide. Maybe better:
```suggestion
      <constant>PASSWORD_DEFAULT</constant> was int 1; now is &null; as of PHP 7.4.3, it is string '2y'
```
Or maybe just brush the null under the carpet; i.e. don't mention it at all:
```suggestion
      <constant>PASSWORD_DEFAULT</constant> was int 1; now is string '2y'
```

cmb69 (commented):
Thanks for the PR!

appendices/migration74/incompatible.xml
ssshotaro44: Thanks for your review! I think this one is better. This expression maybe help both 7.4.0 - 7.4.2 user and 7.4.3- user 

> <constant>PASSWORD_DEFAULT</constant> was int 1; now is &null; as of PHP 7.4.3, it is string '2y'

tiffany-taylor: "now" should be changed to specify the version that `null` occurs, otherwise there's confusion as to what "now" implies.

tiffany-taylor: To be more specific, does `null` occur in PHP 8.0? 8.1? A specific subversion of 7.4?
The wording of "As of PHP x.x.x, this value is `null`. As of PHP 7.4.3, it is string `'2y'`."

IMSoP: According to https://3v4l.org/nN4Qi the `null` value was in the brief window of 7.4.0 to 7.4.2. Although it reads odd out of context, the "now" is what is used throughout that page. I think it's reasonable for "now" to refer to the *final* value, not the one that only briefly existed, though, so perhaps the anomaly (bug?) could just be mentioned in parentheses:

```
<constant>PASSWORD_DEFAULT</constant> was int 1; now is string '2y' (in PHP 7.4.0, 7.41, and 7.4.2 it was &null;)
```

cmb69:
Thank you! Applied as 68dfe9740823ed1d229e5e79b7018fe71237c665.

PR #1491, opened 2022-03-31
Fix listing of unimplemented DOM methods

mdio:
`DOMNamedNodeMap::removeNamedItem` is listed twice and I suspect the one after `DOMNamedNodeMap::setNamedItemNS` was meant to be `DOMNamedNodeMap::removeNamedItemNS` instead.


cmb69:
Good catch, thank you!

PR #1492, opened 2022-03-31
Usort updates based on comments

Crell:
I have already removed 2/3 of the page's comments. :smile:

### DIFF
--- a/reference/array/functions/usort.xml
+++ b/reference/array/functions/usort.xml
@@ -110,7 +110,7 @@ foreach ($a as $key => $value) {
      simplify the internal comparison even further.
     </para>
     <programlisting role="php">
-     <![CDATA[
+<![CDATA[
 <?php
 function cmp($a, $b)
 {
@@ -253,7 +253,7 @@ z, c
 ]]>
     </screen>
    </example>
-   <example xml:id="function.usort.examples.multipleaxes">
+   <example xml:id="function.usort.examples.multiple-axes">
     <title>
      The spaceship operator allows for straightforward comparison of
      compound values across multiple axes. The following example will sort
--- a/reference/array/functions/usort.xml
+++ b/reference/array/functions/usort.xml
@@ -253,6 +253,41 @@ z, c
 ]]>
     </screen>
    </example>
+   <example xml:id="function.usort.examples.multipleaxes">
+    <title>
+     The spaceship operator allows for straightforward comparison of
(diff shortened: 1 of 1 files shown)


reference/array/functions/usort.xml
```diff
@@ -105,6 +105,28 @@ foreach ($a as $key => $value) {
 4: 6
 ]]>
     </screen>
+    <para>
+     As of PHP 7.0, the spaceship operator may be used to
```
Girgias, line 109:
```suggestion
     The spaceship operator may be used to
```
We don't document PHP 5 anymore, so it's safe to assume the spaceship operator is present

reference/array/functions/usort.xml
```diff
@@ -105,6 +105,28 @@ foreach ($a as $key => $value) {
 4: 6
 ]]>
     </screen>
+    <para>
+     The spaceship operator may be used to
+     simplify the internal comparison even further.
+    </para>
+    <programlisting role="php">
+     <![CDATA[
```
cmb69, line 113:
```suggestion
<![CDATA[
```

reference/array/functions/usort.xml
```diff
@@ -105,6 +105,28 @@ foreach ($a as $key => $value) {
...
+<?php
+function cmp($a, $b)
+{
+    return $a <=> $b;
+}
+
+$a = array(3, 2, 5, 6, 1);
```
cmb69, line 120: Did you use `array()` (instead of `[]`) on purpose? Ah, I see, copy&paste. Then leave as is, or change both occurrences.

reference/array/functions/usort.xml
```diff
@@ -233,6 +250,41 @@ foreach ($array as $item) {
 y, a
 x, b
 z, c
+]]>
+    </screen>
+   </example>
+   <example xml:id="function.usort.examples.multipleaxes">
```
cmb69, line 256: Maybe better:
```suggestion
   <example xml:id="function.usort.examples.multiple-axes">
```
(I was wondering what "multipleaxes" are)

cmb69 (commented):
Thank you!

reference/array/functions/usort.xml
Crell: It's what Vikings use in combat when they want to show off.

cmb69 (approved):

PR #1494, opened 2022-03-31
Comparison operator updates from comments

Crell:
All of these based off of comments, some of which I've deleted and some not (for reference, will delete after this is merged).

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -1733,7 +1733,6 @@ var_dump(min(-100, -10, NULL, 10, 100)); // NULL - (bool)NULL < (bool)-100 is FA
     </example>
    </para>
 
-
    <para>
     <example>
      <title>Transcription of standard array comparison</title>
@@ -1751,6 +1750,41 @@ function standard_array_compare($op1, $op2)
     foreach ($op1 as $key => $val) {
         if (!array_key_exists($key, $op2)) {
             return null; // uncomparable
+        } elseif ($val < $op2[$key]) {
+            return -1;
+        } elseif ($val > $op2[$key]) {
+            return 1;
+        }
+    }
+    return 0; // $op1 == $op2
+}
+?>
+]]>
+     </programlisting>
+    </example>
+   </para>
+
+   <para>
+    <example>
+     <title>Transcription of spaceship operator array comparison</title>
+     <para>
+      Array comparison with the spaceship operator is slightly different. As a result, some array comparisons
+      may return <literal>1</literal> regardless of which value is listed first.
+     </para>
+     <programlisting role="php">
+<![CDATA[
+<?php
+// Arrays are compared like this with the spaceship operator
+function standard_array_compare($op1, $op2)
+{
+    if (count($op1) < count($op2)) {
+        return -1; // $op1 < $op2
+    } elseif (count($op1) > count($op2)) {
+        return 1; // $op1 > $op2
+    }
+    foreach ($op1 as $key => $val) {
+        if (!array_key_exists($key, $op2)) {
+            return 1;
         } elseif ($val < $op2[$key]) {
             return -1;
         } elseif ($val > $op2[$key]) {
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -1877,7 +1877,7 @@ echo ((true ? 'true' : false) ? 't' : 'f');
     </note>
     <note>
      <para>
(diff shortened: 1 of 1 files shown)


language/operators.xml
```diff
@@ -1765,6 +1764,41 @@ function standard_array_compare($op1, $op2)
...
+        } elseif ($val < $op2[$key]) {
+            return -1;
+        } elseif ($val > $op2[$key]) {
+            return 1;
+        }
+    }
+    return 0; // $op1 == $op2
+}
+?>
+]]>
```
cmb69, line 1797: I'm not sure whether we should give the full code here; actually, only a single line is different from the `standard_array_compare()` function definition above (namely the return value in case of `!array_key_exists($key, $op2)`, which is wrong in the function definition above, since [`false` is returned, not `null`](https://3v4l.org/m2p0Y)).

language/operators.xml
```diff
@@ -1778,6 +1812,15 @@ function standard_array_compare($op1, $op2)
     </para>
    </warning>
 
+   <note>
+    <simpara>
+     Be aware that PHP's type juggling is not always obvious when comparing values of different types,
```
cmb69, line 1817: I try to avoid apostrophs in the manual (partly due to typesetting issues), and usually use `of` instead of the Saxon genitive. Feel free to leave it as is, though.

language/operators.xml
```diff
@@ -1778,6 +1812,15 @@ function standard_array_compare($op1, $op2)
...
+     particularly comparing integers to booleans or integers to strings. It is therefore generally
```
cmb69, line 1818: It might make sense to use the respective entity references here:
```suggestion
     particularly comparing &integer;s to &boolean;s or &integer;s to &string;s. It is therefore generally
```

language/operators.xml
```diff
@@ -1778,6 +1812,15 @@ function standard_array_compare($op1, $op2)
...
+     advisable to using <literal>===</literal> and <literal>!==</literal> comparisons rather than
```
cmb69, line 1819: "advisable to using" or "advisable to use"?

language/operators.xml
```diff
@@ -1822,8 +1864,9 @@ if (empty($_POST['action'])) {
     <para>
      It is possible to leave out the middle part of the ternary operator.
      Expression <literal>expr1 ?: expr3</literal> returns
```
cmb69, line 1866: Maybe better "evaluates to" instead of "returns" here.

language/operators.xml
```diff
@@ -1872,7 +1934,7 @@ echo ((true ? 'true' : false) ? 't' : 'f');
    <sect2 xml:id="language.operators.comparison.coalesce">
     <title>Null Coalescing Operator</title>
     <para>
-     Further exists the "??" (or null coalescing) operator.
+     Another useful shorthand operator is the "??" (or null coalescing) operator.
```
cmb69, line 1937: Nice! :)

cmb69 (commented):
Thank you for the PR!

language/operators.xml
Crell: Honestly, no idea.  I was just going by what comments on the page said.  I didn't test it myself. :man_shrugging:

language/operators.xml
Crell: "Saxon genetive"?

language/operators.xml
Crell:
```suggestion
     advisable to use <literal>===</literal> and <literal>!==</literal> comparisons rather than
```

language/operators.xml
Crell:
```suggestion
     Expression <literal>expr1 ?: expr3</literal> evaluates to
```

language/operators.xml
```diff
@@ -1862,6 +1870,25 @@ echo ((true ? 'true' : false) ? 't' : 'f');
...
 // second ternary expression.
 ?>
+]]>
+       </programlisting>
+      </example>
+     </para>
+    </note>
+    <note>
+     <para>
+      Chaining of short-ternaries (<literal>?:</literal>, however, is stable and behaves reasonably.
```
cmb69, line 1880:
```suggestion
      Chaining of short-ternaries (<literal>?:</literal>), however, is stable and behaves reasonably.
```

cmb69 (approved):
LGTM (besides the missing parenthesis).

Issue #1495, opened 2022-04-01
Is Foreign Function Interface still experimental?


BafS:
The documentation mention that FFI is experimental, it has been introduced in 2019 (PHP 7.4), is the documentation still up to date for newer PHP versions? If yes, when do you consider FFI stable?

> This extension is EXPERIMENTAL. The behaviour of this extension including the names of its functions and any other documentation surrounding this extension may change without notice in a future release of PHP. This extension should be used at your own risk.

Thanks in advance!

cmb69:
Hmm, good question! I think we should remove the experimental status from the ffi extension, since it is highly unlikely that we're going to radically change the API now.

BafS:
Thanks for your answer! I did a PR (https://github.com/php/doc-en/pull/1499) to change this, let me know.

PR #1499, opened 2022-04-02
Remove "experimental" warning from Foreign Function Interface (#1495)

BafS:
Foreign Function Interface (FFI) has been introduced in 2019 (PHP 7.4) and the API didn't change since then, it should probably be considered safe to use it.


cmb69:
Thank you!

PR #1500, opened 2022-04-03
Add closure versions of all create_function examples

IMSoP:
Basically the only reason people should be looking at the manual
for create_function right now is to see how *not* to use it. So,
for every example that was already included, I've added the
corresponding version using real anonymous functions.

Also, I've reworded the purpose and description, because these
functions aren't anonymous, as evidenced by the return value being
their name.


reference/funchand/functions/create-function.xml
```diff
@@ -73,36 +73,50 @@
   <para>
    <example>
     <title>
-     Creating an anonymous function with <function>create_function</function>
+     Creating an anonymous function, with and without <function>create_function</function>
```
cmb69, line 76: Since `create_function` doesn't create anonymous functions, this is a bit of an oxymoron. I think we can leave it, but may consider a different wording (also in several cases below).

cmb69 (commented):
Thank you! That may be helpful for those still having to deal with `create_function` (although I hope these cases are rare, now). I'm generally fine with this, but it might make sense to have a second look at the wording (see my comment below).

reference/funchand/functions/create-function.xml
IMSoP: Agreed. I changed the refpurpose and introduction for the same reason. I think "dynamic function" probably captures it more accurately, so I'll go through again and make the distinction consistent.

IMSoP: Done, along with some other improvements.

cmb69 (approved):
Thank you!

PR #1501, opened 2022-04-04
Fix unpacking with string key example

l-vo:
The example seems not relevant to me since it already works and produces the same output in the previous PHP versions.


cmb69:
Good catch, thank you!

PR #1502, opened 2022-04-04
PDO::MYSQL_ATTR_LOCAL_INFILE_DIRECTORY was added in PHP 8.1

zonuexe:
The first change mentioned in the manual for `PDO::MYSQL_ATTR_LOCAL_INFILE_DIRECTORY` was 03f0ebc60c1e2c397ca81997646197d3d0cb4fed.

Then 607ec1f63c524e6951e7d3480377973788d495c5 stated about the version to which the constant was added.
However, it incorrectly appends to `MYSQL_ATTR_LOCAL_INFILE` instead of `MYSQL_ATTR_LOCAL_INFILE_DIRECTORY`.


cmb69:
Good catch, thank you!

PR #1503, opened 2022-04-04
Clean up autoload page

Crell:
Mostly the exceptions, which are bad mojo with `spl_autoload_register()`.

### DIFF
--- a/language/oop5/autoload.xml
+++ b/language/oop5/autoload.xml
@@ -27,15 +27,15 @@
     as of PHP 8.0.0.
    </para>
   </caution>
- <note>
-  <para>
-   <function>spl_autoload_register</function> may be called multiple times in order
-   to register multiple autoloaders. Throwing an exception from an autoload function,
-   however, will interrupt that process and not allow further autoload functions to
-   run. For that reason, throwing exceptions from an autoload function is strongly
-   discouraged. Any return value from the function will be ignored by the system.
-  </para>
- </note>
+  <note>
+   <para>
+    <function>spl_autoload_register</function> may be called multiple times in order
+    to register multiple autoloaders. Throwing an exception from an autoload function,
+    however, will interrupt that process and not allow further autoload functions to
+    run. For that reason, throwing exceptions from an autoload function is strongly
+    discouraged. Any return value from the function will be ignored by the system.
+   </para>
+  </note>
   <para>
    <example>
     <title>Autoload example</title>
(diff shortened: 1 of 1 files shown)


language/oop5/autoload.xml
```diff
@@ -23,6 +27,15 @@
...
   </caution>
+ <note>
+  <para>
+   <function>spl_autoload_register</function> may be called multiple times in order
+   to register multiple autoloaders. Throwing an exception from an autoload function,
+   however, will interrupt that process and not allow further autoload functions to
+   run. For that reason, throwing exceptions from an autoload function is strongly
+   discouraged. Any return value from the function will be ignored by the system.
+  </para>
+ </note>
```
Girgias, line 38: Nit: indent seems off here

language/oop5/autoload.xml
```diff
@@ -23,6 +27,15 @@
     as of PHP 8.0.0.
    </para>
   </caution>
+  <note>
+   <para>
+    <function>spl_autoload_register</function> may be called multiple times in order
+    to register multiple autoloaders. Throwing an exception from an autoload function,
+    however, will interrupt that process and not allow further autoload functions to
+    run. For that reason, throwing exceptions from an autoload function is strongly
+    discouraged. Any return value from the function will be ignored by the system.
```
cmb69, line 36: Is it important to note the return values are ignored? After all, `__autoload()`, `spl_autoload()` and the `$callback` of `spl_autoload_register()` are documented as void functions.

cmb69 (approved):
Thank you for the PR! I'm generally fine with this, but not sure about the return value info (see comment below).

language/oop5/autoload.xml
Crell: Not sure. I was going by the comments, which were full of people talking about exceptions. :-)  I'm trying to recall if I've seen people trying to use the return value for debug purposes in the past.  If they're documented as void, though, then it's probably best to remove that line.

Crell:
```suggestion
    discouraged.
```

Crell: Note: Composer's autoload callback returns true if the class was loaded, null if not.  I don't know if that's good, bad, or silly, but it does. That's why I mentioned it initially.

... Which would actually cause a problem if we ever get around to adding callable signatures and tried to enforce a void return on the callback... :smile:

PR #1504, opened 2022-04-04
Fix syntax errors in examples

Girgias:
Found by running the ``check-examples.php`` QA script from base (https://github.com/php/doc-base/pull/58)


cmb69 (approved):
Wow! Thank you!

PR #1515, opened 2022-04-12
"is-scalar.xml" -> add note for NULL

voku:


reference/var/functions/is-scalar.xml
```diff
@@ -20,8 +20,8 @@
   <para>
    Scalar variables are those containing an <type>int</type>, 
    <type>float</type>, <type>string</type> or <type>bool</type>.  
-   Types <type>array</type>, <type>object</type> and <type>resource</type> 
-   are not scalar.
+   Types <type>array</type>, <type>object</type>, <type>resource</type> 
+   and NULL are not scalar.
```
Girgias, line 24:
```suggestion
   and NULL are not scalars.
```
Might as well fix that at the same time :)

kamil-tekiela: I may be wrong, but I believe "scalar" was meant to be an adjective here, not a noun. Just like in the preceding sentence it says "Scalar variables are", not "Scalars are". So I think the way it is now is correct, but if you want to change it, then it might be a good idea to change it in both places.

cmb69: Shouldn't that better be:
```suggestion
   and <type>null</type> are not scalar.
```

PR #1517, opened 2022-04-13
Improve and fix docs which had incorrect stub info

Girgias:
Not sure what the wording usually is for the changelog and the error sections, so please advise.

### DIFF
--- a/reference/exec/functions/exec.xml
+++ b/reference/exec/functions/exec.xml
@@ -87,7 +87,7 @@
   </para>
   <para>
    Throws a <classname>ValueError</classname> if <parameter>command</parameter>
-   is empty or contain null bytes.
+   is empty or contains null bytes.
   </para>
  </refsect1>
 
--- a/reference/exec/functions/exec.xml
+++ b/reference/exec/functions/exec.xml
@@ -83,7 +83,7 @@
   &reftitle.errors;
   <para>
    Emits an <constant>E_WARNING</constant> if <function>exec</function>
-   is unable to fork the <parameter>command</parameter>.
+   is unable to execute the <parameter>command</parameter>.
   </para>
   <para>
    Throws a <classname>ValueError</classname> if <parameter>command</parameter>
--- a/reference/exec/functions/exec.xml
+++ b/reference/exec/functions/exec.xml
@@ -79,6 +79,42 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Emits an <constant>E_WARNING</constant> if <function>exec</function>
+   is unable to fork the <parameter>command</parameter>.
+  </para>
+  <para>
+   Throws a <classname>ValueError</classname> if <parameter>command</parameter>
+   is empty or contain null bytes.
+  </para>
+ </refsect1>
+
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       If <parameter>command</parameter> is empty or contain null bytes
+       now throws a <classname>ValueError</classname>. Previously it emitted
+       an <constant>E_WARNING</constant> and returned &false;.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
+ </refsect1>
+
  <refsect1 role="examples">
   &reftitle.examples;
   <para>
--- a/reference/exec/functions/exec.xml
+++ b/reference/exec/functions/exec.xml
@@ -105,9 +105,9 @@
      <row>
       <entry>8.0.0</entry>
       <entry>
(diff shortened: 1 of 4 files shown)


reference/exec/functions/exec.xml
```diff
@@ -79,6 +79,42 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       If <parameter>command</parameter> is empty or contain null bytes
```
kocsismate, line 108:
```suggestion
       If <parameter>command</parameter> is empty or contains null bytes
```

kocsismate: I'm not sure about the rest of the sentence, but would this make more sense?
```
If <parameter>command</parameter> is empty or contains null bytes,
it throws a <classname>ValueError</classname> now.
```

cmb69: I think using `<function>exec</function>` (or maybe "this function") instead of "it" would be more appropriate for technical documentation.

reference/gmp/functions/gmp-random-seed.xml
```diff
@@ -37,18 +37,42 @@
...
+   &return.void;
   </para>
  </refsect1>
 
  <refsect1 role="errors"><!-- {{{ -->
   &reftitle.errors;
   <para>
-   Issues an <constant>E_WARNING</constant> and returns &false; if
-   <parameter>seed</parameter> is not valid.
+   Throws a <classname>ValueError</classname> is <parameter>seed</parameter>
```
kocsismate, line 47:
```suggestion
   Throws a <classname>ValueError</classname> if <parameter>seed</parameter>
```

reference/exec/functions/passthru.xml
```diff
@@ -61,6 +61,42 @@
...
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Will emit an <constant>E_WARNING</constant> if <function>passthru</function>
+   is unable to fork the <parameter>command</parameter>.
+  </para>
+  <para>
+   Throws a <classname>ValueError</classname> if <parameter>command</parameter>
+   is empty or contain null bytes.
```
kocsismate, line 72: contain -> contains (there are a bunch of similar issues above)

reference/exec/functions/exec.xml
```diff
@@ -79,6 +79,42 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Emits an <constant>E_WARNING</constant> if <function>exec</function>
+   is unable to fork the <parameter>command</parameter>.
```
cmb69, line 86: It might be better not to use "fork" here. On Windows, processes are never forked (that's just not supported). If we want to be precise, we should clarify that in the description section.
```suggestion
   is unable to execute the <parameter>command</parameter>.
```

reference/exec/functions/passthru.xml
```diff
@@ -61,6 +61,42 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Will emit an <constant>E_WARNING</constant> if <function>passthru</function>
+   is unable to fork the <parameter>command</parameter>.
```
cmb69, line 68: Same as above
```suggestion
   is unable to execute the <parameter>command</parameter>.
```

reference/funchand/functions/register-shutdown-function.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type class="union"><type>bool</type><type>null</type></type><methodname>register_shutdown_function</methodname>
+   <type>void</type><methodname>register_shutdown_function</methodname>
```
cmb69, line 12: This is only [correct as of PHP 8.2.0](https://heap.space/search?project=PHP-8.1&project=php-src&full=&defs=register_shutdown_function&refs=&path=&hist=&type=). I suggest to move this change to another PR (with the proper milestone), and to add a changelog entry there.

cmb69 (commented):
Thank you!

reference/funchand/functions/register-shutdown-function.xml
Girgias: Did you mean 8.1? (or did the patch get backported in between) But will add a changelog

cmb69: Hmm, I likely was confused about the versions. So just a changelog entry.

Girgias: So I just got back to this, and as far as I can see, false was only returned when the callable was invalid, which is since 8.0 a proper ZPP failure. So don't think we need a changelog?

cmb69: Indeed, you are right. The [PHP 8.0 and 8.1 stubs are wrong](https://heap.space/search?project=PHP-8.0&project=PHP-8.1&full=&defs=register_shutdown_function&refs=&path=&hist=&type=), though.

cmb69 (approved):

PR #1524, opened 2022-04-17
Document where wordwrap() wraps

da2x:


cmb69 (changes requested):
The `<refpurpose>` is supposed to be a *brief* summary of the function, and is also shown, for instance, when a function is linked in a see-also section. This refpurpose is definitely too long.

tiffany-taylor:
This change can be moved to [line 23](https://github.com/php/doc-en/blob/d6cf75ebfde169545af475cd5205af33f13a3c37/reference/strings/functions/wordwrap.xml#L23), probably with some slight wording changes.

cmb69 (approved):
Thank you!

Girgias (approved):

PR #1525, opened 2022-04-18
fpm mentions selinux integration

devnexen:


cmb69:
This is only available in the master branch, right?

devnexen:
yes

cmb69:
Then there should also be a changelog entry, like for `--with-fpm-apparmor`.

install/fpm/install.xml
```diff
@@ -63,6 +63,12 @@
            The <literal>--with-fpm-apparmor</literal> option was added.
           </entry>
          </row>
+         <row>
+          <entry>9.0.0</entry>
```
cmb69, line 67: The next PHP version is most likely PHP 8.2.0. :) (there are no concrete plans regarding PHP 9.0 yet)

cmb69: Also, the changelog entries are supposed to be documented in decreasing order regarding the PHP version (i.e. this entry should be above the 8.0.0 entry)

cmb69 (approved):
Thank you! This looks good to me, but has to wait until PHP 8.2.0 is in sight (i.e. somewhere shortly before the GA release). Thus, I've added the PHP-8.2 milestone.

PR #1526, opened 2022-04-18
Update unserialize.xml

00dev:
Missed info about changes in PHP 7.0.0


cmb69:
Thank you for the PR, but we do no longer document PHP 5 specifics, and everything that changed as of PHP 7.0.0 is therefore irrelevant, too.

PR #1528, opened 2022-04-19
[PHP 8.1] Implicit incompatible float to int conversions

saundefined:


language/types/float.xml
```diff
@@ -82,6 +82,31 @@ EXPONENT_DNUM (({LNUM} | {DNUM}) [eE][+-]? {LNUM})
...
+   </simpara>
+   <programlisting role="php">
+<![CDATA[
+<?php
+
+function foo($value): int {
+  return $value; 
+}
+
+var_dump(foo(8.1)); // Deprecated: Implicit conversion from float-string "8.1" to int loses precision as of PHP 8.1.0
```
kamil-tekiela, line 100:
```suggestion
var_dump(foo(8.1)); // "Deprecated: Implicit conversion from float 8.1 to int loses precision" as of PHP 8.1.0
```
It's not a string. Did you want to add both examples?

saundefined: Thanks,

Indeed, I'd add both examples if it doesn't seem redundant

cmb69 (commented):
Thank you for the PR!

language/types/float.xml
```diff
@@ -82,6 +82,34 @@ EXPONENT_DNUM (({LNUM} | {DNUM}) [eE][+-]? {LNUM})
  <sect2 xml:id="language.types.float.casting">
   <title>Converting to float</title>
 
+  <sect3 xml:id="language.types.float.casting.from-integer">
+   <title>From integers</title>
+
+   <simpara>
+    As of PHP 8.1.0, a deprecation notice is emitted when implicitly converting a &float; to &integer;.
```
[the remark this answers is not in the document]
saundefined: Thank you, clarified it, also added an example.

language/types/float.xml
```diff
@@ -82,6 +82,36 @@ EXPONENT_DNUM (({LNUM} | {DNUM}) [eE][+-]? {LNUM})
  <sect2 xml:id="language.types.float.casting">
   <title>Converting to float</title>
 
+  <sect3 xml:id="language.types.float.casting.from-integer">
+   <title>From integers</title>
+
+   <simpara>
+    As of PHP 8.1.0, a deprecation notice is emitted when implicitly converting a not integral &float; to &integer;.
```
Girgias, line 89: This is on the wrong page, it should be in ``language/types/finteger.xml``

Girgias (changes requested):

language/types/float.xml
cmb69: Right (s/finteger/integer, though). And it is not about converting to float from integer, but rather about converting to int from float.

saundefined: Thanks! Moved it

language/types/integer.xml
```diff
@@ -213,8 +213,27 @@ var_dump(round(25/7));  // float(4)
    <simpara>
     When converting from <type>float</type> to <type>int</type>, the number
     will be rounded <emphasis>towards zero</emphasis>.
+    As of PHP 8.1.0, a deprecation notice is emitted when implicitly converting a not integral &float; to &integer;.
```
kocsismate, line 216: What about this?
```suggestion
    As of PHP 8.1.0, a deprecation notice is emitted when implicitly converting a nonintegral &float; to &integer;.
```

kocsismate: Would it be worth to emphasize here in addition to the example below that the notice is only emitted when the conversion loses precision?

language/types/integer.xml
```diff
@@ -213,8 +213,27 @@ var_dump(round(25/7));  // float(4)
    <simpara>
     When converting from <type>float</type> to <type>int</type>, the number
     will be rounded <emphasis>towards zero</emphasis>.
+    As of PHP 8.1.0, a deprecation notice is emitted when implicitly converting and loses precision a non-integral &float; to &integer;.
```
kocsismate, line 216:
```suggestion
    As of PHP 8.1.0, a deprecation notice is emitted when implicitly converting a non-integral &float; to &integer; which loses precision.
```

kocsismate (approved):
This sounds good to me. Hopefully native speakers do agree ^^

language/types/integer.xml
```diff
@@ -213,8 +213,27 @@ var_dump(round(25/7));  // float(4)
...
+   <programlisting role="php">
+<![CDATA[
+<?php
+
+function foo($value): int {
+  return $value; 
+}
+
+var_dump(foo(8.1)); // "Deprecated: Implicit conversion from float 8.1 to int loses precision" as of PHP 8.1.0
+var_dump(foo(8.1)); // 8 before PHP 8.1.0
```
Girgias, line 228:
```suggestion
var_dump(foo(8.1)); // 8 prior to PHP 8.1.0
```

Girgias (approved):

PR #1537, opened 2022-04-22
Add missing constant definition ReflectionAttribute::IS_INSTANCEOF

ollieread:
The constant `ReflectionAttribute::IS_INSTANCEOF` is missing from the [documentation](https://www.php.net/manual/en/class.reflectionattribute.php) entirely.

This PR adds the definition. I was unsure of the exact way to do it, but I made an educated guess based on the language and layout of other constant flags and reflection class constants (Most notably `array_filter` and `ReflectionMethod`).

I hope to create other PRs in the near future adding documentation to the `getAttributes()` methods on the various reflection classes, which will need to reference this constant.

### DIFF
--- a/reference/reflection/reflectionattribute.xml
+++ b/reference/reflection/reflectionattribute.xml
@@ -36,6 +36,14 @@
      </oointerface>
     </classsynopsisinfo>
 
+    <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
+    <fieldsynopsis>
+     <modifier>const</modifier>
+     <type>int</type>
+     <varname linkend="reflectionattribute.constants.is-instanceof">ReflectionAttribute::IS_INSTANCEOF</varname>
+     <initializer>2</initializer>
+    </fieldsynopsis>
+
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.reflectionattribute')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[not(@role='procedural')])">
      <xi:fallback/>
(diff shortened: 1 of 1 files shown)


reference/reflection/reflectionattribute.xml
```diff
@@ -48,6 +56,35 @@
...
+     </varlistentry>
+
+    </variablelist>
+    <note>
+     <para>
+      The values of these constants may change between PHP versions.
+      It is recommended to always use the constants
+      and not rely on the values directly.
+     </para>
+    </note>
```
cmb69, line 83: This could be made a reusable XML entity (see https://github.com/php/doc-en/blob/master/language-snippets.ent). On the other hand, IMO that applies to all constants, so may just be documented once in a central place.

cmb69 (commented):
Thank you for the PR! This looks generally good to me, but I left some comments which might be addressed.

Also note, that there is https://github.com/php/php-src/pull/7434, which would allow to auto-generate the respective documentation (modulo the actual description of the constant).

reference/reflection/reflectionattribute.xml
```diff
@@ -48,6 +56,35 @@
...
+   <section xml:id="reflectionattribute.constants.flags">
+    <title>ReflectionAttribute Flags</title>
+    <variablelist>
+
+     <varlistentry xml:id="reflectionattribute.constants.is-instanceof">
+      <term><constant>ReflectionAttribute::IS_INSTANCEOF</constant></term>
+      <listitem>
+       <para>
+        Retrieve attributes using an
+        <parameter>instanceof</parameter> check.
```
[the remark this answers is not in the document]
ollieread: I was unsure of the best way to describe this, as I am planning to PR docs for all those methods at some point. I'm open to a better description if you can think of one.

reference/reflection/reflectionattribute.xml
ollieread: This particular block was lifted from another reflection classes page, as it's present on all that have them.

If the content is always the same, which it does appear to be, naming it reusable would work well. It would also likely be a larger task, as all of the individual instances would need to be replaced. I'm more than happy to give that a go as a separate PR if you think its worth it.

reference/reflection/reflectionattribute.xml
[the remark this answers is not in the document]
cmb69: After further consideration, I prefer to have the description on the `::getAttributes()` pages, because it is most relevant there. This could be done by another PR.

reference/reflection/reflectionattribute.xml
cmb69: Well, let's leave it as is for now. Can still be improved later.

cmb69:
Thanks again!

PR #1546, opened 2022-04-26
[PHP 8.1] Add new global Sodium constants

saundefined:
Relates [Migration Guide](https://www.php.net/manual/en/migration81.constants.php#migration81.constants.sodium)


cmb69 (approved):
Thank you!

PR #1547, opened 2022-04-26
[PHP 8.0] Add errors section for `version_compare`

saundefined:
Not sure that it should be documented...

https://3v4l.org/kPB2p

### DIFF
--- a/reference/info/functions/version-compare.xml
+++ b/reference/info/functions/version-compare.xml
@@ -92,6 +92,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you for the PR!

Frankly, I'm no longer sure that we should document `ValueError`s for each function individually. @Girgias started some half-automation for this, but the scope was mighty. Furthermore, there would still be the discrepancy between PHP 7 and 8 (and possibly even minor versions). And, maybe most important, it shouldn't matter for userland code (you are not supposed to catch `ValueError`s, but rather to avoid triggering these in the first place; i.e. fix the code which might throw a `ValueError`).

What we definitely should do is to clearly point out which values are permissible or forbidden for certain parameters (what is already the case for `$operator` of `version_compare`).

It might be sensible to discuss this general point on the documentation mailing list (or did we already?)

PR #1548, opened 2022-04-26
All drivers implement getColumnMeta

NattyNarwhal:
Since php/php-src@caa710037e663fd78f67533b29611183090068b2, all bundled drivers (Firebird and ODBC being the omissions before) implement this function. The third-party drivers, even the ones in massive states of disrepair, support it too. It's still optional, so a new PDO driver could take us by surprise, but the list is misleading, so just include the warning.


cmb69 (approved):
Thank you!

PR #1549, opened 2022-04-27
[PHP 8.1] Usage of static Variables in Inherited Methods

saundefined:
Relates [Migration Guide](https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.core.static-variable-inheritance)

### DIFF
--- a/language/oop5/static.xml
+++ b/language/oop5/static.xml
@@ -59,34 +59,6 @@ $classname::aStaticMethod();
 ]]>
      </programlisting>
    </example>
-
-   <para>
-    As of PHP 8.1.0, when a method using static variables is inherited (but not overridden),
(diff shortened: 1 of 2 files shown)


cmb69 (commented):
Thank you! The information is correct, but I think it should be put elsewhere in the manual; see the tip at the beginning of [that page](https://www.php.net/manual/en/language.oop5.static.php). The section about [using static variables](https://www.php.net/manual/en/language.variables.scope.php#language.variables.scope.static) may be a better place, or perhaps the [inheritance](https://www.php.net/manual/en/language.oop5.inheritance.php) section, or the section about [late static binding](https://www.php.net/manual/en/language.oop5.late-static-bindings.php).

PR #1551, opened 2022-04-27
Collator::__construct() and Collator::create() do not accept null

### LINKED ISSUE #1550: Collator::__construct() does not accept null
### Description

The following code:

```php
<?php
new Collator(null);
```

Resulted in this output:
```
Deprecated: Collator::__construct(): Passing null to parameter php/php-src#1 ($locale) of type string is deprecated in /in/AUfEa on line 2
```

However, [the documentation says](https://www.php.net/manual/en/collator.construct.php): “if null is passed for the locale, the default locale's collation rules will be used”.

Therefore, I expect that the `$locale` parameter is nullable, and no deprecation warning is triggered.

(Alternatively, in case `""` has the same effect of what is currently documented for `null`, the documentation may be amended.)

### PHP Version

PHP 8.1.5

### Operating System

_No response_

claudepache:
Fixes #1550

### DIFF
--- a/reference/intl/collator/construct.xml
+++ b/reference/intl/collator/construct.xml
@@ -25,7 +25,8 @@
      <listitem>
       <para>
        The locale whose collation rules should be used. Special values for
-       locales can be passed in - if &null; is passed for the locale, the
+       locales can be passed in - if an empty string (<literal>""</literal>)
+       is passed for the locale, the
        default locale's collation rules will be used. If "root" is passed,
        <link xlink:href="&url.icu.uca;">UCA</link> rules will be used.
       </para>
--- a/reference/intl/collator/create.xml
+++ b/reference/intl/collator/create.xml
@@ -35,7 +35,8 @@
      <listitem>
       <para>
        The locale containing the required collation rules. Special values for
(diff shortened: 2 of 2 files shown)


cmb69:
Thanks for the PR! However, there is already 69e0955f7b1cb0b485afc443da459b9f0ce80077. Do you think that `<literal>""</literal>" is relevant info? If so, please fix the current merge conflicts.

cmb69:
Ah, good catch! Both `::__construct()` and `::create()` internally call [`collator_ctor()`](https://github.com/php/php-src/blob/a4179e4c92b6365d39e09cb9cd63c476848013af/ext/intl/collator/collator_create.c#L25), and that expects a string (but not `null`). There is special handling for empty strings, which causes the default locale's collation rules to be used:
 https://github.com/php/php-src/blob/a4179e4c92b6365d39e09cb9cd63c476848013af/ext/intl/collator/collator_create.c#L49-L51
I'm going to fix the docs right away.

cmb69:
Fixed as da9e1ca20838bdc83aa6809160ef014ebde28837.

PR #1552, opened 2022-04-27
Link to the drivers, since it's not obvious what it's meant

NattyNarwhal:
Sorry for the additional PR so soon; someone asked me what was meant by "drivers in the manual" - this just cross-links to the drivers page.

### DIFF
--- a/reference/pdo/pdostatement/getcolumnmeta.xml
+++ b/reference/pdo/pdostatement/getcolumnmeta.xml
@@ -22,8 +22,9 @@
    <simpara>
     Some drivers may not implement
     <function>PDOStatement::getColumnMeta</function>,
-    as it is optional. However, all extensions documented in the manual
-    implement this function.
+    as it is optional. However,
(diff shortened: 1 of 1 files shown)


reference/pdo/pdostatement/getcolumnmeta.xml
```diff
@@ -22,8 +22,9 @@
    <simpara>
     Some drivers may not implement
     <function>PDOStatement::getColumnMeta</function>,
-    as it is optional. However, all extensions documented in the manual
-    implement this function.
+    as it is optional. However,
+    <link linkend="pdo.drivers">all PDO extensions</link>
```
cmb69, line 26: Ah, yes! But since we explicitly link to the *drivers*, maybe reword
```suggestion
    <link linkend="pdo.drivers">all PDO drivers</link>
```
(although that isn't as eloquent, since we use "drivers" twice in the same sentence)

And maybe move "all" out of the link.

cmb69:
Thank you!

PR #1555, opened 2022-04-28
[PHP 8.1] Passing null to non-nullable parameters of built-in functions

saundefined:
Relates [Migration Guide](https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal)

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1075,7 +1075,9 @@ $func(); // prints "bar"
     <para>
      For example, <function>strlen</function> function expects the parameter <literal>string</literal>
      to be a non-nullable &string;.
-     For historical reasons, PHP allows passing &null; for this parameter in weak mode, and the parameter is implicitly casted to the accepted type. In contrast, a <classname>TypeError</classname> is emitted In strict mode.
+     For historical reasons, PHP allows passing &null; for this parameter in weak mode, and the parameter is
+     implicitly casted to <type>string</type>, resulting in a <literal>""</literal> value.
+     In contrast, a <classname>TypeError</classname> is emitted In strict mode.
     </para>
 
     <example>
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1065,10 +1065,34 @@ $func(); // prints "bar"
     </simpara>
    </note>
    <note>
-    <simpara>
+    <para>
      Scalar types for built-in functions are nullable by default.
-     As of PHP 8.1.0, passing &null; to a function parameter that is not declared nullable is not allowed and emits a deprecation notice.
-    </simpara>
+     As of PHP 8.1.0, passing &null; to a function parameter that is not declared nullable
+     is discouraged and emits a deprecation notice to align with the behavior of user-defined functions,
+     where scalar types need to be marked as nullable explicitly.
+    </para>
+
+    <para>
+     For example, <function>strlen</function> function expects the parameter <literal>string</literal>
+     to be a non-nullable &string;.
+     For historical reasons, PHP allowed passing &null; for this parameter.
+    </para>
+
+    <example>
+     <programlisting role="php">
+<![CDATA[
+<?php
+var_dump(strlen(null));
+// "Deprecated: Passing null to parameter #1 ($string) of type string is deprecated" as of PHP 8.1.0
+// int(0)
+
+var_dump(str_contains("foobar", null));
+// "Deprecated: Passing null to parameter #2 ($needle) of type string is deprecated" as of PHP 8.1.0
+// bool(true)
+?> 
+]]>
+     </programlisting>
+    </example>
    </note>
 
    <sect2 role="seealso">
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1066,18 +1066,18 @@ $func(); // prints "bar"
    </note>
    <note>
     <para>
-     Scalar types for built-in functions are nullable by default in weak mode.
+     Scalar types for built-in functions are nullable by default in coercive mode.
      As of PHP 8.1.0, passing &null; to an internal function parameter that is not declared nullable
-     is discouraged and emits a deprecation notice in weak mode to align with the behavior of user-defined functions,
+     is discouraged and emits a deprecation notice in coercive mode to align with the behavior of user-defined functions,
      where scalar types need to be marked as nullable explicitly.
     </para>
 
     <para>
-     For example, <function>strlen</function> function expects the parameter <literal>string</literal>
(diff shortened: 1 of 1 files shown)


language/functions.xml
```diff
@@ -1064,6 +1064,12 @@ $func(); // prints "bar"
      be thrown in this case.
     </simpara>
    </note>
+   <note>
+    <simpara>
+     Scalar types for built-in functions are nullable by default.
+     As of PHP 8.1.0, passing &null; to a function parameter that is not declared nullable is not allowed and emits a deprecation notice.
```
kocsismate, line 1070: I think "not allowed" is misleading here, since it's allowed but discouraged for now.

In my opinion, the behavior before 8.1 and since 8.1 should both be described with a little bit more detail, as the implicit nullability is a usual source of confusion. Maybe 1 or 2 examples would help.

saundefined: Thank you, reworded a little

language/functions.xml
```diff
@@ -1064,6 +1064,36 @@ $func(); // prints "bar"
...
+     Scalar types for built-in functions are nullable by default.
+     As of PHP 8.1.0, passing &null; to a function parameter that is not declared nullable
+     is discouraged and emits a deprecation notice to align with the behavior of user-defined functions,
+     where scalar types need to be marked as nullable explicitly.
+    </para>
+
+    <para>
+     For example, <function>strlen</function> function expects the parameter <literal>string</literal>
+     to be a non-nullable &string;.
+     For historical reasons, PHP allowed passing &null; for this parameter.
```
kocsismate, line 1078:
```suggestion
     For historical reasons, PHP allows passing &null; for this parameter in weak mode, and the parameter is implicitly casted to the accepted type. In contrast, a <classname>TypeError</classname> is emitted In strict mode.
```

language/functions.xml
```diff
@@ -1064,6 +1064,36 @@ $func(); // prints "bar"
      be thrown in this case.
     </simpara>
    </note>
+   <note>
+    <para>
+     Scalar types for built-in functions are nullable by default.
```
kocsismate, line 1069:
```suggestion
     Scalar types for built-in functions are nullable by default in weak mode.
```

language/functions.xml
```diff
@@ -1064,6 +1064,36 @@ $func(); // prints "bar"
...
+     As of PHP 8.1.0, passing &null; to a function parameter that is not declared nullable
```
kocsismate, line 1070:
```suggestion
     As of PHP 8.1.0, passing &null; to an internal function parameter that is not declared nullable
```

language/functions.xml
```diff
@@ -1064,6 +1064,36 @@ $func(); // prints "bar"
...
+     is discouraged and emits a deprecation notice to align with the behavior of user-defined functions,
```
kocsismate, line 1071:
```suggestion
     is discouraged and emits a deprecation notice in weak mode to align with the behavior of user-defined functions,
```

kocsismate (commented):
It has just come to my mind that the difference between weak and strict mode should be highlighted. I couldn't find an appropriate link describing weak and strict modes (I admit that I wasn't really searching for long), but a link should also be added for reference.

language/functions.xml
```diff
@@ -1064,6 +1064,36 @@ $func(); // prints "bar"
...
+     Scalar types for built-in functions are nullable by default in weak mode.
+     As of PHP 8.1.0, passing &null; to an internal function parameter that is not declared nullable
+     is discouraged and emits a deprecation notice in weak mode to align with the behavior of user-defined functions,
+     where scalar types need to be marked as nullable explicitly.
+    </para>
+
+    <para>
+     For example, <function>strlen</function> function expects the parameter <literal>string</literal>
+     to be a non-nullable &string;.
+     For historical reasons, PHP allows passing &null; for this parameter in weak mode, and the parameter is implicitly casted to the accepted type. In contrast, a <classname>TypeError</classname> is emitted In strict mode.
```
kocsismate, line 1078: Oops, sorry. I made an edit to my suggestion, but I was too late:
```suggestion
     For historical reasons, PHP allows passing &null; for this parameter in weak mode, and the parameter is
     implicitly casted to <type>string</type>, resulting in a <literal>""</literal> value.
     In contrast, a <classname>TypeError</classname> is emitted In strict mode.
```

language/functions.xml
```diff
@@ -1064,6 +1064,38 @@ $func(); // prints "bar"
      be thrown in this case.
     </simpara>
    </note>
+   <note>
+    <para>
+     Scalar types for built-in functions are nullable by default in weak mode.
```
cmb69, line 1069: I don't think we're using the term "weak mode" elsewhere in the manual; instead, the term "coercive mode" is used, if we even mention that mode explicitly.
```suggestion
     Scalar types for built-in functions are nullable by default in coercive mode.
```
More occurrences below.

language/functions.xml
```diff
@@ -1064,6 +1064,38 @@ $func(); // prints "bar"
...
+     As of PHP 8.1.0, passing &null; to an internal function parameter that is not declared nullable
+     is discouraged and emits a deprecation notice in weak mode to align with the behavior of user-defined functions,
+     where scalar types need to be marked as nullable explicitly.
+    </para>
+
+    <para>
+     For example, <function>strlen</function> function expects the parameter <literal>string</literal>
```
cmb69, line 1076: Maybe
```suggestion
     For example, <function>strlen</function> function expects the parameter <literal>$string</literal>
```

language/functions.xml
```diff
@@ -1064,6 +1064,38 @@ $func(); // prints "bar"
...
+     to be a non-nullable &string;.
+     For historical reasons, PHP allows passing &null; for this parameter in weak mode, and the parameter is
+     implicitly casted to <type>string</type>, resulting in a <literal>""</literal> value.
```
cmb69, line 1079:
```suggestion
     implicitly cast to <type>string</type>, resulting in a <literal>""</literal> value.
```

language/functions.xml
```diff
@@ -1064,6 +1064,38 @@ $func(); // prints "bar"
...
+     In contrast, a <classname>TypeError</classname> is emitted In strict mode.
```
cmb69, line 1080:
```suggestion
     In contrast, a <classname>TypeError</classname> is emitted in strict mode.
```

cmb69 (commented):
Thank you for the PR!

cmb69 (approved):
Thank you!

PR #1560, opened 2022-04-29
[PHP 8.1] Return Type Compatibility with Internal Classes

saundefined:
But perhaps there is a more appropriate page 🤔

### DIFF
--- a/language/oop5/inheritance.xml
+++ b/language/oop5/inheritance.xml
@@ -114,8 +114,8 @@ $bar->printPHP();       // Output: 'PHP is great'
    </para>
  
    <para>
-    As of PHP 8.1.0, when extending internal classes, the return value types of non-final methods
-    must be compatible with the class or interface being extended; otherwise, emits a deprecation notice.
+    As of PHP 8.1.0, when extending internal classes, the return type of methods
+    must be compatible with the class or interface being extended; otherwise, a deprecation notice is emitted.
     Note that lack of an explicit return declaration is also considered a signature mismatch,
     and thus results in the deprecation notice.
    </para>
--- a/language/oop5/inheritance.xml
+++ b/language/oop5/inheritance.xml
@@ -122,7 +122,7 @@ $bar->printPHP();       // Output: 'PHP is great'
    </para>
  
    <para>
-    In case the return type cannot be declared for an overriding method due to PHP cross-version compatibility concerns,
+    If the return type cannot be declared for an overriding method due to PHP cross-version compatibility concerns,
     a <code>#[ReturnTypeWillChange]</code> attribute can be added to silence the deprecation notice.
    </para>
  
--- a/language/oop5/inheritance.xml
+++ b/language/oop5/inheritance.xml
@@ -135,8 +135,7 @@ class MyDateTime extends DateTime
     public function modify(string $modifier) { return false; }
 }
  
-// "Deprecated: Return type of MyDateTime::modify(string $modifier) should either be compatible
-// with DateTime::modify(string $modifier): DateTime|false" as of PHP 8.1.0
+// "Deprecated: Return type of MyDateTime::modify(string $modifier) should either be compatible with DateTime::modify(string $modifier): DateTime|false, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice" as of PHP 8.1.0
 ?> 
 ]]>
     </programlisting>
@@ -152,8 +151,7 @@ class MyDateTime extends DateTime
     public function modify(string $modifier): ?DateTime { return null; }
 }
  
-// "Deprecated: Return type of MyDateTime::modify(string $modifier):
-// ?DateTime should either be compatible with DateTime::modify(string $modifier): DateTime|false" as of PHP 8.1.0
+// "Deprecated: Return type of MyDateTime::modify(string $modifier): ?DateTime should either be compatible with DateTime::modify(string $modifier): DateTime|false, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice" as of PHP 8.1.0
 ?> 
 ]]>
     </programlisting>
--- a/language/oop5/inheritance.xml
+++ b/language/oop5/inheritance.xml
@@ -114,8 +114,9 @@ $bar->printPHP();       // Output: 'PHP is great'
    </para>
  
    <para>
-    As of PHP 8.1.0, when extending internal classes, the return type of methods
-    must be compatible with the parent being extended; otherwise, a deprecation notice is emitted.
+    As of PHP 8.1.0, most internal methods started to "tentatively" declare their return type,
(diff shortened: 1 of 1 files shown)


language/oop5/inheritance.xml
```diff
@@ -105,6 +105,82 @@ $bar->printPHP();       // Output: 'PHP is great'
    </example>
   </sect2>
 
+  <sect2 xml:id="language.oop5.inheritance.internal-classes">
+   <title>Return Type Compatibility with Internal Classes</title>
+ 
+   <para>
+    Prior to PHP 8.1, when extending internal classes, there was no need to add a return type,
```
kocsismate, line 112: IIRC this is not exactly true: there have already been a few non-final methods which were recently added and had their return type declarations

language/oop5/inheritance.xml
```diff
@@ -105,6 +105,82 @@ $bar->printPHP();       // Output: 'PHP is great'
...
+    and even a return type mismatch was allowed.
+   </para>
+ 
+   <para>
+    As of PHP 8.1.0, when extending internal classes, the return value types of non-final methods
```
kocsismate, line 117: IMO we don't have to emphasize non-final:
```suggestion
    As of PHP 8.1.0, when extending internal classes, the return type of methods
```

language/oop5/inheritance.xml
```diff
@@ -105,6 +105,82 @@ $bar->printPHP();       // Output: 'PHP is great'
...
+    must be compatible with the class or interface being extended; otherwise, emits a deprecation notice.
```
kocsismate, line 118:
```suggestion
    must be compatible with the class or interface being extended; otherwise, a deprecation notice is emitted.
```

language/oop5/inheritance.xml
```diff
@@ -105,6 +105,82 @@ $bar->printPHP();       // Output: 'PHP is great'
...
+    <programlisting role="php">
+<![CDATA[
+<?php
+class MyDateTime extends DateTime
+{
+    public function modify(string $modifier): ?DateTime { return null; }
+}
+ 
+// "Deprecated: Return type of MyDateTime::modify(string $modifier):
+// ?DateTime should either be compatible with DateTime::modify(string $modifier): DateTime|false" as of PHP 8.1.0
```
kocsismate, line 156: the end of the error message is missing

language/oop5/inheritance.xml
saundefined: What do you think is better: removing the paragraph or rewording to "when extending some/most internal classes..."?

kocsismate: I think the latter one is better

saundefined: Thank you!

language/oop5/inheritance.xml
```diff
@@ -105,6 +105,80 @@ $bar->printPHP();       // Output: 'PHP is great'
    </example>
   </sect2>
 
+  <sect2 xml:id="language.oop5.inheritance.internal-classes">
+   <title>Return Type Compatibility with Internal Classes</title>
+ 
+   <para>
+    Prior to PHP 8.1, when extending most internal classes, there was no need to add a return type,
+    and even a return type mismatch was allowed.
```
kocsismate, line 113: What do you mean here exactly? In methods without return types, there couldn't be a return type mismatch, since any return type was allowed.

language/oop5/inheritance.xml
```diff
@@ -105,6 +105,80 @@ $bar->printPHP();       // Output: 'PHP is great'
...
+   </para>
+ 
+   <para>
+    As of PHP 8.1.0, when extending internal classes, the return type of methods
```
kocsismate, line 117: There are still a few methods whose return type is not declared (usually because the behavior of the method is crazy). On the other hand, interfaces could be mentioned here as well, and the "class or interface" below could be changed to "parent"

language/oop5/inheritance.xml
kocsismate: Maybe the introduction paragraph could mention instead that most internal classes or methods didn't declare their return types prior to php 8.1, and any return type was allowed when extending them.

saundefined: Thank you, rewrote it

language/oop5/inheritance.xml
```diff
@@ -105,6 +105,80 @@ $bar->printPHP();       // Output: 'PHP is great'
...
+  <sect2 xml:id="language.oop5.inheritance.internal-classes">
+   <title>Return Type Compatibility with Internal Classes</title>
+ 
+   <para>
+    Prior to PHP 8.1, most internal classes or methods didn't declare their return types,
+    and any return type was allowed when extending them.
+   </para>
+ 
+   <para>
+    As of PHP 8.1.0, when extending internal classes, the return type of methods
```
kocsismate, line 117: IMO the only missing part is that most internal methods started to "tentatively" declare their return type. In which case, the return type of overriding methods needs to becompatible.

Also, I'd use should here: "the return type of methods *should* be compatible ..."

kocsismate (approved):
Looks good to me. As always, review by native speakers is highly appreciated :)

Crell:
One tiny language note.  Otherwise looks good to this American. :smile:

saundefined:
Thanks everyone!

jmikola:
I came across this PR while reviewing https://github.com/alcaeus/doc-en/pull/3 for @alcaeus, which is also adding `ReturnTypeWillChange` to the MongoDB extension docs. One thing I couldn't figure out is why this PR used a backslash between the opening bracket and "R" in a few places, such as https://github.com/php/doc-en/commit/081cf71ecb97b239c1e07beaadb71af0b0c42764#diff-f9a69bbe508e0ea8560572d071eedc00b00477b19ecba80b44737d95e7e973b5R139

Is that escaping necessary? Searching the PHP documentation, I only see that twice in `en/language/oop5/inheritance.xml` (from this PR), twice in `en/language/predefined/iterator.xml`, and once in `en/language/predefined/arrayaccess/offsetexists.xml`. If it's not actually necessary I can submit a PR to remove it, but I wanted to confirm first.

alcaeus:
The backslash is not to escape the `R`, but rather to provide a fully qualified class name. This may be unnecessary in this case as the snippets don't declare a namespace, but if users were to copy-paste that code to a file declaring a namespace, it would become necessary again.

kocsismate:
It's funny and a very nice catch in the same time! :) I'm sure @saundefined 's intention was to reference `ReturnTypeWillChange`as a FQCN in order to avoid problems with calls from namespaces. However, The `\` is not escaped itself properly, so the manual doesn't display `\`.

saundefined:
Oh, the deprecation notice is copied from output (https://3v4l.org/5WH2c#v8.1.9) 
The backslash is not necessary.

PR #1563, opened 2022-05-02
Fix type for index in ZipArchive::replaceFile

kingkero:
In all of the `ZipArchive` methods, the `index` is always described as `int` (see https://www.php.net/manual/en/class.ziparchive.php). Only for the `replaceFile` it is documented as `string`. However, actually passing a `string` will trigger (in PHP 8.1.5)

> Fatal error: Uncaught TypeError: ZipArchive::replaceFile(): Argument #2 ($index) must be of type int, string given

> TypeError: ZipArchive::replaceFile(): Argument #2 ($index) must be of type int, string given

The error also appears in other languages of the documentation. But _replacefile.xml_ is only present in this repo. So I assume the other language docs will be automatically fixed after merging this?

### DIFF
--- a/reference/zip/ziparchive/replacefile.xml
+++ b/reference/zip/ziparchive/replacefile.xml
@@ -10,7 +10,7 @@
   <methodsynopsis>
    <modifier>public</modifier> <type>bool</type><methodname>ZipArchive::replaceFile</methodname>
    <methodparam><type>string</type><parameter>filepath</parameter></methodparam>
-   <methodparam><type>string</type><parameter>index</parameter></methodparam>
+   <methodparam><type>int</type><parameter>index</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


cmb69:
Oh, right, but we need to fix [the stub](https://github.com/php/php-src/blob/82dae61c04205e2a09df005b00bc7398aa586023/ext/zip/php_zip.stub.php#L94) first. Do you want to provide a PR for php-src? If so, you need to fix the stub, and then run `php build/gen_stub.php ext/zip/php_zip.stub.php` to regenerate php_zip_arginfo.h.

kingkero:
@cmb69 Thanks for the info, created as https://github.com/php/php-src/pull/8481

cmb69:
Since the php-src PR has already been merged, I'm going to merge this as well. Thanks again!

PR #1565, opened 2022-05-02
functions/constant Minor tweak to "errors" section

jblz:
https://www.php.net/manual/en/function.constant.php

This updates the text from:

> An Error exception is thrown, if the constant is not defined.

to:

> If the constant is not defined, an Error exception is thrown.

The existing text is more difficult to read as it puts the condition behind the result. If the existing text is preferred, I'd remove the comma.

### DIFF
--- a/reference/misc/functions/constant.xml
+++ b/reference/misc/functions/constant.xml
@@ -53,9 +53,8 @@
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   An <classname>Error</classname> exception is thrown,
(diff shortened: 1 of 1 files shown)


cmb69:
I'm not strictly against changing this, but we should have some style guideline about those error sections. Otherwise we're risking to change the syntax over and over again, what would be particularly annoying for translators.

Crell:
The current text is just flat out incorrect because of the comma. There should either be no comma, or this change that reverses the clause order.  Either would work.

Since this one is here, I'm good with it for now.

PR #1570, opened 2022-05-05
Added a note about scientific notation strings

machavity:
There's nothing explicit in the manual about scientific notation strings being interpreted as number. A section that people can be pointed to might be helpful

### DIFF
--- a/language/types/numeric-strings.xml
+++ b/language/types/numeric-strings.xml
@@ -30,6 +30,23 @@ NUM_STRING       ({INT_NUM_STRING} | {FLOAT_NUM_STRING})
   This is simply a string which starts like a numeric string followed by
   any characters.
  </para>
+ 
+ <note>
+  <para>
+   Any string that contains the letter <literal>E</literal> (case insensitive)
+   bounded by numbers will be seen as a number expressed in scientific notation.
+   This can produce unexpected results.
+   <programlisting role="php">
+<![CDATA[
+<?php
+var_dump("0D1" == "000"); // false, "0D1" is not scientific notation
+var_dump("0E1" == "000"); // true, "0E1" is 0 * (10 ^ 1), or 0
+var_dump("2E1" == "020"); // true, "2E1" is 2 * (10 ^ 1), or 20
+?>
+]]>
+   </programlisting>
+  </para>
+ </note>
 
  <sect2 xml:id="language.types.numeric-string.conversion">
   <title>Strings used in numeric contexts</title>
--- a/language/types/numeric-strings.xml
+++ b/language/types/numeric-strings.xml
@@ -30,7 +30,6 @@ NUM_STRING       ({INT_NUM_STRING} | {FLOAT_NUM_STRING})
   This is simply a string which starts like a numeric string followed by
   any characters.
  </para>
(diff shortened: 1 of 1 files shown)


language/types/numeric-strings.xml
```diff
@@ -30,6 +30,22 @@ NUM_STRING       ({INT_NUM_STRING} | {FLOAT_NUM_STRING})
   This is simply a string which starts like a numeric string followed by
   any characters.
  </para>
+ 
+   <note>
+    <para>Any string that contains the letter <literal>E</literal> (case insensitive)
```
cmb69, line 35: I suggest to break the line after the `<para>` for better readability:
```suggestion
    <para>
     Any string that contains the letter <literal>E</literal> (case insensitive) 
```

cmb69 (commented):
Thank you for the PR! I'm not sure, though, whether that needs to be explicitly documented; there is already the grammar above which specifies that "e/E" can be part of a numeric string, and there is an example below. I'd like to hear what others think about this.

Girgias:
I think it would be better to have an example section, where one of them is in exponential notation, as the ``e`` specifier is already present just not very visible.

machavity:
I'm fine with either. It's just one of those arcane gotchas of PHP and the manual gives examples (there's some given without explanation at https://www.php.net/manual/en/language.operators.comparison.php but it didn't seem to fit to add any notes there). Was trying to address this question and was surprised the manual has nothing direct https://stackoverflow.com/questions/72126877/illogical-string-comparison-in-php

Crell:
I think I'd favor an example, but either works for me.  @machavity are you going to continue this?

machavity:
What would I need to do to continue? I can make the readability edit cmb mentions if that's all.

cmb69:
> I can make the readability edit cmb mentions if that's all.

Yes, please do that. And also have a look at the indentation; apparently the new lines have 2 spaces too much.

language/types/numeric-strings.xml
```diff
@@ -30,6 +30,23 @@ NUM_STRING       ({INT_NUM_STRING} | {FLOAT_NUM_STRING})
   This is simply a string which starts like a numeric string followed by
   any characters.
  </para>
+ 
+ <note>
+  <para>
+   Any string that contains the letter <literal>E</literal> (case insensitive) 
+   bounded by numbers will be seen as a number expressed in scientific notation.
```
cmb69, line 37: These lines have trailing whitespace:
```suggestion
   Any string that contains the letter <literal>E</literal> (case insensitive)
   bounded by numbers will be seen as a number expressed in scientific notation.
```

language/types/numeric-strings.xml
```diff
@@ -30,6 +30,23 @@ NUM_STRING       ({INT_NUM_STRING} | {FLOAT_NUM_STRING})
   This is simply a string which starts like a numeric string followed by
   any characters.
  </para>
+
```
cmb69, line 33:
```suggestion

```

cmb69:
Oh, there actually was a line with a single space; I've fixed that and committed.

Thanks for the PR again!

Issue #1534, opened 2022-04-21
Ds interfaces are documented as classes


AndrolGenhald:
https://www.php.net/manual/en/class.ds-collection.php
https://www.php.net/manual/en/class.ds-hashable.php
https://www.php.net/manual/en/class.ds-sequence.php

Both the URLs and the Interface synopses incorrectly indicate that these are classes (although it appears [other](https://www.php.net/manual/en/class.jsonserializable.php) interface URLs also say class, so maybe part is intentional?).

cmb69:
The `class.` prefix in the URL is intentional (classes and interface live in the same namespace anyway). 

But, yeah, the markup for the interfaces is broken (and also interfaces are supposed to extend other interfaces, not to implement them).

mallardduck:
@cmb69 - I don't know enough about phd and the few hours I've been playing with it have been fruitless.

However, from the XML side of things I quickly found that this is valid and gets closer to what we'd desire here:

```xml
    <classsynopsisinfo>

     <oointerface>
      <interfacename>Ds\Collection</interfacename>
     </oointerface>

     <oointerface>
      <modifier>extends</modifier>
      <interfacename>Traversable</interfacename>
     </oointerface>

     <oointerface>
      <interfacename>Countable</interfacename>
     </oointerface>

     <oointerface>
      <interfacename>JsonSerializable</interfacename>
     </oointerface>

    </classsynopsisinfo>
```

The issue is that `phd` still renders the initial `implements` rather than only rendering `extends`. So this is the end result:

<img width="783" alt="Screen Shot 2022-04-21 at 10 12 47 PM" src="https://user-images.githubusercontent.com/619938/164582844-1b0588a7-f054-4707-80a7-f2db7affb0d5.png">

cmb69:
@mallardduck, yeah, this isn't properly solveable with current PhD. [gen_stub.php](https://github.com/php/php-src/blob/master/build/gen_stub.php) would generate the following:
````xml
    <classsynopsisinfo>

     <oointerface>
      <interfacename>Ds\Collection</interfacename>
     </oointerface>

     <ooclass>
      <modifier>extends</modifier>
      <classname>Traversable</classname>
     </ooclass>
     <ooclass>
      <classname>Countable</classname>
     </ooclass>
     <ooclass>
      <classname>JsonSerializable</classname>
     </ooclass>

    </classsynopsisinfo>
````
This results in:

[...]

PR #1587, opened 2022-05-18
Fix: `Ds\Collection` extends `IteratorAggregate`.

ZebulanStanphill:
Discovered in https://github.com/phpstan/phpstan/issues/7258.

`Ds\Collection` extends `IteratorAggregate` in both [the `php-ds/php-ds` polyfill](https://github.com/php-ds/polyfill/blob/master/src/Collection.php) and [the `ext-ds` extension](https://github.com/php-ds/ext-ds/blob/master/src/php/classes/php_collection_ce.c), so the documentation should be updated to reflect this.

Note that `Traversable` is removed from the `extends` because `IteratorAggregate` already extends `Traversable`.

### DIFF
--- a/reference/ds/ds.collection.xml
+++ b/reference/ds/ds.collection.xml
@@ -57,7 +57,7 @@
      </thead>
      <tbody>
       <row>
-       <entry>1.4.0</entry>
+       <entry>PECL ds 1.4.0</entry>
        <entry>
         <classname>Collection</classname> implements
         <interfacename>IteratorAggregate</interfacename> now instead of just <interfacename>Traversable</interfacename>. (This change came to the polyfill in 1.4.1.)
--- a/reference/ds/ds.collection.xml
+++ b/reference/ds/ds.collection.xml
@@ -31,8 +31,8 @@
 
     <ooclass><classname>Ds\Collection</classname></ooclass>
 
-    <oointerface><interfacename>Traversable</interfacename></oointerface>
     <oointerface><interfacename>Countable</interfacename></oointerface>
+    <oointerface><interfacename>IteratorAggregate</interfacename></oointerface>
     <oointerface><interfacename>JsonSerializable</interfacename></oointerface>
 
     </classsynopsisinfo>
--- a/reference/ds/ds.collection.xml
+++ b/reference/ds/ds.collection.xml
@@ -45,6 +45,29 @@
 
   </section>
 
+  <section role="changelog">
+   &reftitle.changelog;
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR, but this is a [known issue](https://github.com/php/doc-en/issues/1534). The problem is that PhD isn't yet able to properly render that for the PHP manual.

AndrolGenhald:
@cmb69 I think this is a separate issue isn't it? It would still be rendered wrong, but `Traversable` could be swapped out for `IteratorAggregate` (assuming that change is correct).

cmb69:
Oh, right! Sorry, I was too fast here. Reopening, but indeed the situation needs to be investigated; was that maybe a *change* in some DS version, which would require a changelog entry?

ZebulanStanphill:
@cmb69 The change appears to have been made in these commits:

- for the extension: https://github.com/php-ds/ext-ds/commit/f4fee1adf17c26d9514254f660d70e1016893237
- for the polyfill: https://github.com/php-ds/polyfill/commit/d0ebb780a67c752e7030522102f57baa312d2b85

So it was a change in 1.4.0 for the extension and 1.4.1 for the polyfill.

cmb69:
@ZebulanStanphill, thanks for checking this! Could you please add a changelog entry (similar to [that for CachingIterator](https://github.com/php/doc-en/blob/09f4a3f040083d8832da366139fd3ee1b611cba6/reference/spl/cachingiterator.xml#L164-L185))?

ZebulanStanphill:
@cmb69 Done. 🙂

reference/ds/ds.collection.xml
```diff
@@ -45,6 +45,29 @@
...
+    <tgroup cols="2">
+     <thead>
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>1.4.0</entry>
```
cmb69, line 60: To have that properly presented on https://www.php.net/manual/en/doc.changelog.php:
```suggestion
       <entry>PECL ds 1.4.0</entry>
```

cmb69 (commented):
Besides the minor issue below, this looks good to me. Thank you!

reference/ds/ds.collection.xml
ZebulanStanphill: Thanks. :+1:

cmb69:
Thank you!

PR #1590, opened 2022-05-20
Fix bug 67034: Unable to compare DateInterval instances with == operator

cmb69:


cmb69:
@derickr, I think just adding a respective paragraph to the `DateInterval` page should be sufficient.

@Girgias, please have a look at the "Uncomparable Values" section; is that okay, or should we explicitly specify the results?

Girgias:
> @derickr, I think just adding a respective paragraph to the `DateInterval` page should be sufficient.
> 
> @Girgias, please have a look at the "Uncomparable Values" section; is that okay, or should we explicitly specify the results?

I think the current wording (modulo it being "incomparable" and not "uncomparable" (welcome to the English language)) is best, as we might want to actually change the behaviour when trying to compared incomparable values.

But maybe the comparison operator page already describes the behaviour?

cmb69:
> But maybe the comparison operator page already describes the behaviour?

It only mentions "uncomparable"[sic] twice, but doesn't specify what that means. I'm going to update to use "incomparable".

Girgias (approved):
LGTM

PR #1592, opened 2022-05-21
Added gc_maxlifetime default value to description

php4fan:
For consistency with all other constants, where the default value is also mentioned in the description so you don't have to go back to the table to look it up.


php4fan:
This should fix issue 1591

devnexen:
As mentioned in the ticket, it is more for consistency with a handful of other constants.

PR #1594, opened 2022-05-24
Add sane string to number comparison

iwasherefirst2:
This case was missing in the table: https://www.php.net/releases/8.0/en.php#saner-string-to-number-comparisons

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -1692,7 +1692,9 @@ echo $a <=> $b; // 1
       <row>
        <entry><type>string</type>, <type>resource</type>, <type>int</type> or <type>float</type></entry>
        <entry><type>string</type>, <type>resource</type>, <type>int</type> or <type>float</type></entry>
-       <entry>Translate strings and resources to numbers, usual math</entry>
+       <entry>Translate strings and resources to numbers, usual math. Note that since PHP 8 this has changed for one case: 
+        When comparing a number to a non-numeric string, PHP 8
+        converts the number to a string and uses a string comparison.</entry>
       </row>
       <row>
        <entry><type>array</type></entry>


language/operators.xml
```diff
@@ -1692,7 +1692,9 @@ echo $a <=> $b; // 1
       <row>
        <entry><type>string</type>, <type>resource</type>, <type>int</type> or <type>float</type></entry>
        <entry><type>string</type>, <type>resource</type>, <type>int</type> or <type>float</type></entry>
-       <entry>Translate strings and resources to numbers, usual math</entry>
+       <entry>Translate strings and resources to numbers, usual math. Note that since PHP 8 this has changed for one case: 
+        When comparing a number to a non-numeric string, PHP 8
+        converts the number to a string and uses a string comparison.</entry>
       </row>
```
Girgias, line 1698: I think this deserves a whole new row instead of a comment.

probably something along the lines, non numeric string and numeric string for the wording.

iwasherefirst2: Actually that is what I did first,with "<string><int><float>" in first and second column and then explaining, that only if its string in one and a number in the other, then.. Then I found this too complicated, so I created two rows,
for ""<string> | <int><float>" and ""<int><float> | <string>" which duplicate description, which I found also too complicated, thats why I just added a comment in this row. Also, if we create another row, then it would contradict with this row, therefore an exception handling for the current row has to be added anyway, right?

cmb69: This table has to be read in order, so adding a new row isn't a problem per se. I'm not really sure that we need a new row, or even some lengthy explanation. Why not link to https://www.php.net/manual/en/language.types.numeric-strings.php#language.types.numeric-strings instead?

afilina:
@iwasherefirst2 Gentle nudge to author to look at the review comments.

PR #1597, opened 2022-05-26
Sync ext/date class synopsis with recent stub changes

kocsismate:


reference/datetime/dateperiod.xml
```diff
@@ -41,10 +41,11 @@
 
     <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
     <fieldsynopsis>
+     <modifier>public</modifier>
      <modifier>const</modifier>
      <type>int</type>
      <varname linkend="dateperiod.constants.exclude-start-date">DatePeriod::EXCLUDE_START_DATE</varname>
-     <initializer>1</initializer>
+     <initializer>UNKNOWN</initializer>
```
derickr, line 48: This looks wrong.

reference/datetime/datetimeinterface.xml
```diff
@@ -37,82 +37,95 @@
...
      <initializer>"D, d M Y H:i:s O"</initializer>
     </fieldsynopsis>
     <fieldsynopsis>
+     <modifier>public</modifier>
      <modifier>const</modifier>
      <type>string</type>
-     <varname linkend="datetime.constants.rfc7231">DateTimeInterface::RFC7231</varname>
-     <initializer>"D, d M Y H:i:s \G\M\T"</initializer>
+     <varname linkend="datetimeinterface.constants.rfc7231">DateTimeInterface::RFC7231</varname>
+     <initializer>"D, d M Y H:i:s \\G\\M\\T"</initializer>
```
derickr, line 93: This escape looks wrong.

reference/datetime/datetimeinterface.xml
```diff
@@ -37,82 +37,95 @@
...
      <initializer>"D, d M Y H:i:s O"</initializer>
     </fieldsynopsis>
     <fieldsynopsis>
+     <modifier>public</modifier>
      <modifier>const</modifier>
      <type>string</type>
-     <varname linkend="datetime.constants.rfc3339">DateTimeInterface::RFC3339</varname>
-     <initializer>"Y-m-d\TH:i:sP"</initializer>
+     <varname linkend="datetimeinterface.constants.rfc3339">DateTimeInterface::RFC3339</varname>
+     <initializer>"Y-m-d\\TH:i:sP"</initializer>
```
derickr, line 107: This escape looks wrong.

reference/datetime/datetimeinterface.xml
```diff
@@ -37,82 +37,95 @@
...
     </fieldsynopsis>
     <fieldsynopsis>
+     <modifier>public</modifier>
      <modifier>const</modifier>
      <type>string</type>
-     <varname linkend="datetime.constants.rfc3339_extended">DateTimeInterface::RFC3339_EXTENDED</varname>
-     <initializer>"Y-m-d\TH:i:s.vP"</initializer>
+     <varname linkend="datetimeinterface.constants.rfc3339-extended">DateTimeInterface::RFC3339_EXTENDED</varname>
+     <initializer>"Y-m-d\\TH:i:s.vP"</initializer>
```
derickr, line 114: This escape looks wrong.

reference/datetime/datetimeinterface.xml
```diff
@@ -37,82 +37,95 @@
...
      <initializer>"D, d M Y H:i:s O"</initializer>
     </fieldsynopsis>
     <fieldsynopsis>
+     <modifier>public</modifier>
      <modifier>const</modifier>
      <type>string</type>
-     <varname linkend="datetime.constants.w3c">DateTimeInterface::W3C</varname>
-     <initializer>"Y-m-d\TH:i:sP"</initializer>
+     <varname linkend="datetimeinterface.constants.w3c">DateTimeInterface::W3C</varname>
+     <initializer>"Y-m-d\\TH:i:sP"</initializer>
```
derickr, line 128: This escape looks wrong.

reference/datetime/dateperiod.xml
derickr: It is also missing `DatePeriod::INCLUDE_END_DATE`.

reference/datetime/datetimezone.xml
```diff
@@ -33,88 +33,102 @@
 
     <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
     <fieldsynopsis>
+     <modifier>public</modifier>
      <modifier>const</modifier>
      <type>int</type>
      <varname linkend="datetimezone.constants.africa">DateTimeZone::AFRICA</varname>
-     <initializer>1</initializer>
+     <initializer>UNKNOWN</initializer>
```
derickr, line 40: Why are all these constant `DateTimeZone::*` values `UNKNOWN` ? And if we don't want to show the value, they shouldn't show up at all (instead of saying `UNKNOWN`).

reference/datetime/datetimeinterface.xml
Girgias: If the format is ``Y-m-d\TH:i:sP``, than the escaping is correct as the string is in double quotes. Or am I missing somthing?

kocsismate: Yes, you are right, this is the correct escaping for double quotes.

cmb69: It's also correct for single-quotes (but not necessary there; actually, it's not necessary for double-quoted strings either, since there is no `\T` escape sequence, but for maximum forward compatibility and minimal confusion, it appears to be prudent to escape the backslash).

reference/datetime/datetimeinterface.xml
cmb69: Same as above; escaping is correct.

cmb69 (commented):
Thanks for the PR. This looks good to me, but I think we should do something about the `UNKNOWN` values. Maybe we should link `UNKNOWN` to some section of the manual, where that is explained; or maybe we should use some acronym, so we could mark it up as such, so there'll be a hint on hover (not really accessible, though). Regardless of how we solve it, the meaning should be absolutely clear to readers, and not cause them to file bug reports.

kocsismate:
> This looks good to me, but I think we should do something about the UNKNOWN values. Maybe we should link UNKNOWN to some section of the manual, where that is explained;

Should we use `?` like we do with unknown default parameter values?

cmb69:
> Should we use `?` like we do with unknown default parameter values?

It might be even better to link that to some explanation, maybe somewhere in the [about chapter](https://www.php.net/manual/en/about.php). Unfortunately, that whole chapter needs a thorough overhaul.

kocsismate:
I've just got rid of the UNKNOWN values, and made a few smaller changes: most notably, the properties of DatePeriod were synchronized based on the stubs

reference/datetime/datetimeinterface.xml
```diff
@@ -318,8 +332,7 @@
       <row>
        <entry>8.2.0</entry>
        <entry>
-        The constant <varname
-        linkend="datetime.constants.iso8601_expanded">DateTimeInterface::ISO8601_EXPANDED</varname>
+        The constant <varname linkend="datetimeinterface.constants.iso8601-expanded">DateTimeInterface::ISO8601_EXPANDED</varname>
```
Girgias, line 335:
```suggestion
        The constant <constant linkend="datetimeinterface.constants.iso8601-expanded">DateTimeInterface::ISO8601_EXPANDED</constant>
```
Seems more appropriate?

Girgias (approved):
LGTM as the changes are rather basic.

reference/datetime/dateperiod.xml
```diff
@@ -193,15 +200,15 @@ echo $period->recurrences, "\n";
...
+    <varlistentry xml:id="dateperiod.props.include-end-date">
      <term><varname>include_end_date</varname></term>
      <listitem>
       <para>
        Whether to include the end date in the set of recurring dates or not.
       </para>
      </listitem>
     </varlistentry>
-    <varlistentry xml:id="dateperiod.props.include_start_date">
+    <varlistentry xml:id="dateperiod.props.include-start-date">
```
derickr, line 211: Is this needed? AFAIK this changes HTML anchor targets, which means that already made links to this location (by third parties) will break.

cmb69: > AFAIK this changes HTML anchor targets, which means that already made links to this location (by third parties) will break.

Yeah, that will break, but it is customary for IDs in the docs to use hyphens instead of underscores, and gen_stub.php behaves accordingly. I don't think we do ourselves a favor to introduce special cases for this. We may, however, work around (i.e. redirect) in web-php.

kocsismate: As this is rather a new addition to the manual, I am not very concerned about the breaking change. Especially because visitors can still easily find what they need as the page itself is still shown to them.

derickr:
On 7 November 2022 12:03:14 GMT, "Máté Kocsis" ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>>       <term><varname>include_end_date</varname></term>
>      <listitem>
>       <para>
>        Whether to include the end date in the set of recurring dates or not.
>       </para>
>      </listitem>
>     </varlistentry>
>-    <varlistentry xml:id="dateperiod.props.include_start_date">
>+    <varlistentry xml:id="dateperiod.props.include-start-date">
>
>As this is rather a new addition to the manual, I am not very concerned about the breaking change. Especially because visitors can still easily find what they need as the page itself is still shown to them.
>

FWIW, I don't care enough. So this PR looks good.

But I don't think we can do as Christopher suggests as this isn't part of the actual URL, as the server will not see the anchor?

cmb69:
> But I don't think we can do as Christopher suggests as this isn't part of the actual URL, as the server will not see the anchor?

Oh, right, that's part of the fragment.

cmb69 (approved):
Thank you!

PR #1599, opened 2022-05-26
Sync ext/tidy properties with stubs

kocsismate:


reference/tidy/tidynode.xml
```diff
@@ -21,36 +21,79 @@
 <!-- {{{ Synopsis -->
    <classsynopsis>
     <ooclass>
-     <modifier>final</modifier>
```
kocsismate, line 24: this was in the wrong place and blocked automatic sync until now

cmb69 (approved):
Thank you!

PR #1603, opened 2022-05-28
Clarify `aad` parameter in openssl-encrypt

Krisell:
The `$aad` parameter in `openssl_encrypt` and `openssl_decrypt` is used to pass additional plaintext data that is `authenticated` (i.e. has integrity guaranteed, but no confidentiality). The current documentation uses the word `authentication` for this data, which is misleading and probably a typo of the intended `authenticated`.

The OpenSSL wiki uses `authenticated` consistently for the `$aad` parameter, and never `authentication`: https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption

### DIFF
--- a/reference/openssl/functions/openssl-decrypt.xml
+++ b/reference/openssl/functions/openssl-decrypt.xml
@@ -92,7 +92,7 @@
      <term><parameter>aad</parameter></term>
      <listitem>
       <para>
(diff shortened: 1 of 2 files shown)


cmb69 (approved):
Good catch, thank you!

PR #1609, opened 2022-06-03
socket extension update related to last entries.

devnexen:


cmb69:
Thank you! This looks already good to me, so please switch to "ready for review", unless you're planning to work on this. Having the 8.2 milestone should be sufficient to postpone the merge.

reference/sockets/functions/socket-get-option.xml
```diff
@@ -433,6 +433,16 @@
            <type>int</type>
           </entry>
          </row>
+         <row>
+          <entry><constant>SO_RTABLE</constant></entry>
```
cmb69, line 437: I think these constants should also be listed in constants.xml (just like the other new constants).

cmb69 (commented):
Thank you!

cmb69 (approved):
Thank you!

PR #1612, opened 2022-06-05
SO_REUSEPORT constant, statement no longer really exact.

devnexen:


cmb69 (approved):
Thank you!

I agree that availability as of Linux 3.9 doesn't require additional clarification.

PR #1613, opened 2022-06-05
mb-internal-encoding : ValueError Thrown on invalid encoding in PHP8

### LINKED ISSUE #1611: PHP 8 / mb-internal-encoding / Thrown ValueError when unknown encoding are given
From manual page: https://php.net/function.mb-internal-encoding

Manual said : return true or false

But when giving a bad arg this cause an ValueError 

Documentation should be contain that it thrown ValueError when an unknown encoding are given.

Reproduct : https://3v4l.org/vbrUB

---

Zeuh:
mb-internal-encoding() Thrown ValueError Thrown on invalid encoding since PHP8

Close #1611

### DIFF
--- a/reference/mbstring/functions/mb-internal-encoding.xml
+++ b/reference/mbstring/functions/mb-internal-encoding.xml
@@ -69,6 +69,14 @@
     </thead>
     <tbody>
      &mbstring.changelog.encoding-nullable;
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       Now throws a <classname>ValueError</classname> if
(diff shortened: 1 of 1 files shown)


cmb69 (commented):
Thank you! I think this should also be reflected in a changelog entry, but I'm not sure if we should just duplicate the information there. Maybe just move this info to the changelog, and just state that the function throws a `ValueError` in the errors section (without stating what happened prior to PHP 8.0.0).

Zeuh:
I have added a block into changelog section.

cmb69 (approved):
Okay, let's go with this. Thank you!

PR #1620, opened 2022-06-09
Update fibers.xml

bartvanhoutte:
Added a notice to document the fact that it's currently impossible to switch fibers in the destructor of an object cfr https://github.com/php/php-src/issues/8723.

### DIFF
--- a/language/fibers.xml
+++ b/language/fibers.xml
@@ -34,9 +34,11 @@
    or by throwing an exception into the fiber using <methodname>Fiber::throw</methodname>. The value is returned
    (or exception thrown) from <methodname>Fiber::suspend</methodname>.
   </para>
-  <para>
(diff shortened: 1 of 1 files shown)


language/fibers.xml
```diff
@@ -34,6 +34,9 @@
    or by throwing an exception into the fiber using <methodname>Fiber::throw</methodname>. The value is returned
    (or exception thrown) from <methodname>Fiber::suspend</methodname>.
   </para>
+  <para>
+   Notice: due to current limitations it is not possible to switch fibers in the destructor of an object.
+  </para>
```
cmb69, line 39: Maybe:
```suggestion
  <note>
   <simpara>
    Due to current limitations it is not possible to switch fibers in the destructor of an object.
   </simpara>
  </note> 
```

PR #1621, opened 2022-06-09
Add note that ReflectionProperty::getDefaultValue() does not copy from promoted property

iggyvolz:
Noticed this in https://3v4l.org/vZrFf - surprising behavior but clearly documented in https://wiki.php.net/rfc/constructor_promotion#desugaring.  Writing a similar example and adding a note to `ReflectionProperty::getDefaultValue()`.

### DIFF
--- a/language/oop5/decon.xml
+++ b/language/oop5/decon.xml
@@ -159,7 +159,7 @@ class Point {
      <para>
       <link linkend="language.attributes">Attributes</link> placed on a
       promoted constructor argument will be replicated to both the property
-      and argument.
+      and argument.  Default values on a promoted constructor argument will be replicated only to the argument and not the property.
      </para>
     </note>
    </sect3>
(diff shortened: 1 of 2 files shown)


cmb69:
Well, there is already a [section about "Constructor Promotion"](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion), and I think that we should explicitly spell out the desugaring there. After that is done, I don't see much point in adding a prominent warning to the `ReflectionProperty` methods; instead I'd go with a regular example on that page.

afilina:
@iggyvolz Gentle nudge to author about addressing the review comment.

iggyvolz:
> Well, there is already a [section about "Constructor Promotion"](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion), and I think that we should explicitly spell out the desugaring there. After that is done, I don't see much point in adding a prominent warning to the `ReflectionProperty` methods; instead I'd go with a regular example on that page.

Makes sense - I'll make those changes.  Apologies for dropping the ball on this.

iggyvolz:
Should be all set - I combined the note with another related note about similar behavior with attributes.

Should I squash my commits together before merge?

cmb69:
Thank you!

PR #1623, opened 2022-06-12
Document curl extension `CURLOPT_SAFE_UPLOAD` constant.

devnexen:


reference/curl/functions/curl-setopt.xml
```diff
@@ -518,6 +518,19 @@
...
+           <entry valign="top"><constant>CURLOPT_SAFE_UPLOAD</constant></entry>
+           <entry valign="top">
+            Always &true;, what disables support for the <literal>@</literal> prefix for
+            uploading files in <constant>CURLOPT_POSTFIELDS</constant>, which
+            means that values starting with <literal>@</literal> can be safely
+            passed as fields. <classname>CURLFile</classname> may be used for
+            uploads instead.
+           </entry>
+           <entry valign="top">
+            Can no longer be set to &false; since PHP 7.
```
cmb69, line 531: There is no need to document changes that happened in or prior to PHP 7.0.0, since the manual no longer covers PHP 5.

devnexen: fair enough

reference/curl/functions/curl-setopt.xml
```diff
@@ -518,6 +518,16 @@
...
+          <row>
+           <entry valign="top"><constant>CURLOPT_SAFE_UPLOAD</constant></entry>
+           <entry valign="top">
+            Always &true;, what disables support for the <literal>@</literal> prefix for
+            uploading files in <constant>CURLOPT_POSTFIELDS</constant>, which
+            means that values starting with <literal>@</literal> can be safely
+            passed as fields. <classname>CURLFile</classname> may be used for
+            uploads instead.
+           </entry>
+          </row>
```
cmb69, line 530: Please keep the empty column, i.e.
```suggestion
           <entry valign="top">
           </entry>
          </row>
```
Otherwise an `<td></td>` would not be generated, which breaks the layout. Actually, this should be automatically checked somehow.

cmb69 (approved):
Thank you!

PR #1624, opened 2022-06-12
more concise example code

max-schumacher-shippeo:
remove superfluous variable

### DIFF
--- a/reference/funchand/functions/func-num-args.xml
+++ b/reference/funchand/functions/func-num-args.xml
@@ -49,11 +49,11 @@
     <title><function>func_num_args</function> example</title>
     <programlisting role="php">
 <![CDATA[
+
 <?php
 function foo()
 {
-    $numargs = func_num_args();
-    echo "Number of arguments: $numargs\n";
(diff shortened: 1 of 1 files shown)


reference/funchand/functions/func-num-args.xml
```diff
@@ -49,11 +49,11 @@
     <title><function>func_num_args</function> example</title>
     <programlisting role="php">
 <![CDATA[
+
 <?php
 function foo()
 {
-    $numargs = func_num_args();
-    echo "Number of arguments: $numargs\n";
+    echo "Number of arguments:" . func_num_args() . PHP_EOL;;
```
cmb69, line 56: There is a superfluous semicolon at the end.

I wouldn't use string concatenation here; I'd use either just commas, or `printf()` here.

max-schumacher-shippeo: Hello @cmb69 - I have removed the semicolon and switched to just using commas instead of string concatenation

cmb69:
Thank you!

PR #1626, opened 2022-06-14
Deleted cast to binary string example.

mumumu:
deleted cast to binary string example by the following reason.

- `(binary)` and `(string)` are identical
- b prefix seems to be nothing processed specially[*1]
- all the following code is identical to (string) cast, not cast to binary string.
  * Currently there is no binary string type, isn't it?

```php
$binary = (binary) $string;
$binary = b"binary string";
```

[*1] https://github.com/php/php-src/blob/php-8.1.0/Zend/zend_language_scanner.l#L2445-L2730

### DIFF
--- a/language/types/type-juggling.xml
+++ b/language/types/type-juggling.xml
@@ -262,22 +262,6 @@ $foo = ( int ) $bar;
    </para>
   </note>
 
-  <informalexample>
-   <simpara>
-    Casting literal <type>string</type>s and variables to binary
-    <type>string</type>s:
-   </simpara>
-
-   <programlisting role="php">
-<![CDATA[
-<?php
-$binary = (binary) $string;
-$binary = b"binary string";
-?>
-]]>
-   </programlisting>
-  </informalexample>
-
   <!-- TODO Remove or move into string context section? -->
   <note>
    <simpara>


cmb69:
Thanks for the PR! However, the current documentation is contradictory; in a note it says

> (binary) is an alias of the (string) cast.

But a few paragraphs below there is a caution:

> The (binary) cast and b prefix exists for forward support. Currently (binary) and (string) are identical, however this may change and should not be relied upon.

I think we need to sort that out (should probably be discussed on the internals mailing list), and if we go with the note (`(binary)` is an alias of `(string)`), we should also drop the example. Otherwise we better keep it.

mumumu:
> current documentation is contradictory

Yes, this note is currently contradictory and meaningless, I think.

But I also found the word `for forward support`. This word says these example code will be meaningful in the future.

> I think we need to sort that out

Yeah, we need to sort or change the manual example description, for example:

```
Casting literal strings and variables to binary strings (currently this code is meaningless, but can be changed in the future):
```

cmb69:
> This word says these example code will be meaningful in the future.

Historically, `(binary)` had been introduced to support binary strings in PHP 6, which was supposed to have regular strings as Unicode strings. However, that never has been realized. And, frankly, I don't think this is going to happen. While the [Binary String Deprecation RFC](https://wiki.php.net/rfc/binary_string_deprecation) was declined some years ago, it is [highly unlikely that regular strings will be something different from binary strings ever](https://externals.io/message/96743#96749).

mumumu:
> But I also found the word `for forward support`.

It's true that examples in this PR is meaningless and easy to be get confused.

Also I found that `for forward support` is completely **untranslated** in Japanese manual. By fixing it, Japanese PHP community can avoid some confusion.

@cmb69 Many thanks for many comments.

PR #1636, opened 2022-06-20
Add getimagesize to seealso of filesize

Universal-Omega:


devnexen:
The change does not make sense to me, these are unrelated features, size does not have the same meaning for 'getimagesize' too.

Universal-Omega:
> The change does not make sense to me, these are unrelated features, size does not have the same meaning for 'getimagesize' too.

That's fair enough, and I'm aware they don't have the same usage, but when I was searching for documentation of getimagesize, I ended up at filesize so it seemed like a good idea to have it listed there for when it does happen. Regardless, it's fine not to be, so this was just a suggestion.

PR #1637, opened 2022-06-20
Add filesize to seealso of getimagesize

Universal-Omega:


devnexen:
Looks out of place but I ll let @cmb69 decide for those changes.

cmb69:
I'd agree that this linking is out of place, if `getimagesize()` was a function of ext/gd, but it actually is a function of ext/standard. Still, I'm not convinced that cross-linking the functions makes much sense.

devnexen:
Same remark as your other PR.

PR #1642, opened 2022-06-23
Add reference to get_defined_constants() and constant() each other

zonuexe:
These functions are somewhat similar in use and are useful to be linked. Also, I couldn't reach it from the sidebar because the function categories are different.


mumumu (approved):
Thank you!

PR #1644, opened 2022-06-23
Add note and example to array_fill() to clarify the effect of a negative start_index in PHP 8

overflowerror:


reference/array/functions/array-fill.xml
```diff
@@ -33,12 +33,13 @@
       <para>
        The first index of the returned array.
       </para>
-      <para>
-       If <parameter>start_index</parameter> is negative, 
-       the first index of the returned array will be 
-       <parameter>start_index</parameter> and the following 
+      <para> 
+       Up to (excluding) PHP 8, if
```
cmb69, line 37:
```suggestion
       Prior to PHP 8.0.0, if 
```

reference/array/functions/array-fill.xml
```diff
@@ -108,13 +107,53 @@ Array
...
+    <title>negative <parameter>start_index</parameter> (PHP &lt; 8)</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$b = array_fill(-2, 4, 'pear');
+print_r($b);
+?>
+]]>
+    </programlisting>
+    &example.outputs;
```
cmb69, line 123:
```suggestion
    &example.outputs.8;
```

reference/array/functions/array-fill.xml
```diff
@@ -108,13 +107,53 @@ Array
...
+    <title>negative <parameter>start_index</parameter> (PHP &gt;= 8)</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$c = array_fill(-2, 5, 'apple');
+print_r($c);
+?>
+]]>
+    </programlisting>
+    &example.outputs;
```
cmb69, line 146:
```suggestion
    &example.outputs.8;
```

cmb69 (commented):
Thank you for the PR!

reference/array/functions/array-fill.xml
```diff
@@ -108,13 +107,53 @@ Array
...
+    <title>negative <parameter>start_index</parameter> (PHP &lt; 8)</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$b = array_fill(-2, 4, 'pear');
+print_r($b);
+?>
+]]>
+    </programlisting>
+    &example.outputs.8;
```
Crell, line 123: This is for pre-8.0, so I think we need a new `&example.outputs.74` to use here?

cmb69: Documentation is still for PHP 7.0+, so `&example.outputs.7;` seems appropriate.

cmb69: Anyhow, I don't think we should have two different examples; instead I prefer a single example, which shows the output for PHP 7 and 8.

overflowerror:
#1767 already fixed this.

PR #1645, opened 2022-06-23
Use updated parameter names for ldap_exop

othercorey:
I don't know if these ever get re-generated.

### DIFF
--- a/reference/ldap/functions/ldap-exop.xml
+++ b/reference/ldap/functions/ldap-exop.xml
@@ -12,16 +12,16 @@
   <methodsynopsis>
    <type>mixed</type><methodname>ldap_exop</methodname>
    <methodparam><type>LDAP\Connection</type><parameter>ldap</parameter></methodparam>
-   <methodparam><type>string</type><parameter>reqoid</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


cmb69:
It would be great if the changed parameter names are used throughout that manual page.

cmb69:
And note that there is https://github.com/php/php-src/blob/master/build/gen_stub.php which supports which allows semi-automatic updates.

othercorey:
> It would be great if the changed parameter names are used throughout that manual page.

Whoops, that was silly.

othercorey:
@cmb69 Some of these ldap functions have a reference parameter like `&$controls` that defaults to null but isn't explicitly marked nullable. The same parameter is marked nullable most of the time when it's not a reference. Are references meant to only default to null but not use the nullable type declaration?

cmb69:
> Are references meant to only default to null but not use the nullable type declaration?

Defaulting to `null` is more like a convention for reference parameters to show that they are optional. Actually, in this case they are out parameters, and only for that the type is relevant. Consider the following script which is correct:
````php
$output = "foo";
$status = [];
exec("dir", $output, $status);
````

PR #1648, opened 2022-06-23
Fix Throwable note to mention that Error can be extended (fixes #1647).

### LINKED ISSUE #1647: Class can be throwable by extending Error
https://www.php.net/manual/en/class.throwable.php

> PHP classes cannot implement the Throwable interface directly, and must instead extend [Exception](https://www.php.net/manual/en/class.exception.php).

[`Error` can be extended](https://3v4l.org/QQHoN) now to create a throwable class as well.

AndrolGenhald:


cmb69:
But see https://github.com/php/doc-en/issues/1647#issuecomment-1165439900 f. I'm really unsure whether userland code should extend `Error`.

Girgias:
It might be a good idea to add a note that says although it is possible to extend `Error` this is in general a bad idea as those should only refer to programming/engine errors.

AndrolGenhald:
How sure are you that userland code shouldn't extend `Error`? It doesn't seem that common, but some popular projects are doing it: https://grep.app/search?q=extends%20%5CError&words=true&filter[lang][0]=PHP (a lot of that is stubs and polyfills, but about half of them seem to be actual userland extensions of `Error`).

If `Error` really shouldn't be extended I'll add a note to the `Error` documentation as well, as I mentioned [here](https://github.com/php/doc-en/issues/1647#issuecomment-1165684033) I didn't think the documentation was clear on that. In retrospect, allowing `Error` to be extended in userland makes sense for polyfills, but it should be more clear that it's not intended for general use.

PR #1649, opened 2022-06-23
Fix return type explanation for array_recursive

othercorey:


cmb69:
I don't think this should be documented here. It is [generally undefined behavior to pass values of unexpected types to an internal function](https://www.php.net/manual/en/functions.internal.php).

othercorey:
So just drop the null entirely?

cmb69:
Oh, right!

othercorey:
Should I say previously returned null on error without explaining? I guess it was just a mess when this stuff happened before throwing errors.

I'm simply updating missing docs after updating psalms function signatures.

cmb69:
No, I wouldn't even mention the `NULL` on failure.

cmb69 (approved):
Thank you!

PR #1650, opened 2022-06-23
Add changelog for array_diff and array_intersect functions

othercorey:


cmb69:
Thank you! I think it makes sense to define and use an XML entity, similar to [`array.changelog.by-ref`](https://github.com/php/doc-en/blob/0e8fa6aa168fe2cb9bc5cae74ed22a52d1df891b/language-snippets.ent#L598-L604).

othercorey:
> Thank you! I think it makes sense to define and use an XML entity, similar to [`array.changelog.by-ref`](https://github.com/php/doc-en/blob/0e8fa6aa168fe2cb9bc5cae74ed22a52d1df891b/language-snippets.ent#L598-L604).

Sounds good. I made it include entire row although it looks odd on its own in the tbody section.

cmb69:
Yeah, it takes a while to get used to it, but we use that pattern already elsewhere, so I think it's reasonable.

Could you please remove the unrelated changes (removal of trailing whitespace) from language-snippets.ent? It is generally not a good idea to do that, because the diffs are harder to understand for translators, but especially for that file, it would be unfortunate to mix it with this commit.

othercorey:
I can edit manually. No trailing whitespace is common in project configs. though so wouldn't call it bad. Could adding `trim_trailing_whitespace  = false` to .editorconfig help this?

cmb69:
> Could adding `trim_trailing_whitespace = false` to .editorconfig help this?

Yeah, that seems to be a good idea.

And fixing those trailing whitespace isn't a bad idea, but that should be done in a separate PR, so we can apply `[skip-revcheck]` so that translators don't have to bother with it.

PR #1652, opened 2022-06-24
Make the expression more clear

runbing:
I think the expression about return statement on page "[return](https://www.php.net/manual/en/function.return.php)" is not as clear as that on page "[Migrating from PHP 7.0.x to PHP 7.1.x](https://www.php.net/manual/en/migration71.incompatible.php#migration71.incompatible.typed-returns-compile-time)".

### DIFF
--- a/language/control-structures/return.xml
+++ b/language/control-structures/return.xml
@@ -53,7 +53,7 @@
  </para>
 
  <para>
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR! I tend to agree that the existing sentence is a bit unclear. @Girgias, thoughts?

Girgias (approved):

Girgias:
Thanks!

PR #1654, opened 2022-06-25
Update array-walk.xml

cautnew:
Added an axemple using a callback function.

### DIFF
--- a/reference/array/functions/array-walk.xml
+++ b/reference/array/functions/array-walk.xml
@@ -172,6 +172,33 @@ c. fruit: apple
     </screen>
    </example>
   </para>
+  <para>
+   <example>
+    <title><function>array_walk</function> example using callback</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$elemnts = ['a', 'b', 'c'];
+
+//Remember that the first element of the callable function
+//is the value and the second is the key
+array_walk($elemnts, function ($value, $key){
+  echo "{$key} => {$value}\n";
+});
+
+?>
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen role="php">
+<![CDATA[
+0 => a
+1 => b
+2 => c
+]]>
+    </screen>
+   </example>
+  </para>
  </refsect1>
  <refsect1 role="seealso">
   &reftitle.seealso;


reference/array/functions/array-walk.xml
```diff
@@ -168,6 +168,33 @@ d. fruit: lemon
...
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>array_walk</function> example using callback</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$elemnts = ['a', 'b', 'c'];
```
cmb69, line 181: I don't think this abbreviation makes much sense; for clarity, I'd prefer:
```suggestion
$elements = ['a', 'b', 'c'];
```

reference/array/functions/array-walk.xml
```diff
@@ -168,6 +168,33 @@ d. fruit: lemon
...
+
+//Remember that the first element of the callable function
+//is the value and the second is the key
```
cmb69, line 184: Given that the parameters of the callback are already named accordingly, I'd omit this comment:
```suggestion
```

reference/array/functions/array-walk.xml
```diff
@@ -168,6 +168,33 @@ d. fruit: lemon
...
+array_walk($elemnts, function ($value, $key){
```
cmb69, line 185: Nit: there should be a space between `)` and `{`:
```suggestion
array_walk($elemnts, function ($value, $key) {
```

cmb69 (commented):
Thanks for the PR! I'm not sure, though, whether we should *add* this example; an alternative could be to rewrite the existing example to use an anonymous function.

cautnew:
In my opnion, the first example using string as function name is good for old codes, so i'd like to keep there.

reference/array/functions/array-walk.xml
cautnew: Ok.

reference/array/functions/array-walk.xml
cautnew: Perfect!

reference/array/functions/array-walk.xml
cautnew: Nice!

cautnew (commented):
Greate changes. As you change the array name in the first line, in the function you should change too. Could you do that?

reference/array/functions/array-walk.xml
cautnew: But you have to update the array name for $elements.

reference/array/functions/array-walk.xml
```diff
@@ -168,6 +168,31 @@ d. fruit: lemon
 a. fruit: orange
 b. fruit: banana
 c. fruit: apple
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>array_walk</function> example using callback</title>
```
cmb69, line 177: It is always a callback; I think we should clarify that this is about using an anonymous function.

cautnew: I agree with you, but it's not for all cases. If you pass a string with the function name it works too (this documentation in other languages says that this parameter type is string). I'll change because using string as function name looks weirdo for me.

cmb69: A string which contains a function name is a `callable` aka. callback.

PR #1657, opened 2022-06-27
Add return null return value to in_array

bdsl:
See https://3v4l.org/7nIVl

I'm not entirely sure if this should be a docs fix or a fix to the engine behaviour, but I think we would want to keep the engine the same and adjust the docs.


cmb69:
From the [docs](https://www.php.net/manual/en/functions.internal.php):

So neither the docs nor the implementation are wrong.

PR #1658, opened 2022-06-27
Expand switch description and examples to deprecate comments.

Crell:
I cleaned out most of the comments on the `switch` page.  These are the bits worth including.

### DIFF
--- a/language/control-structures/switch.xml
+++ b/language/control-structures/switch.xml
@@ -29,22 +29,16 @@
  </note>
 
  <para>
-  The following two examples are two different ways to write the
-  same thing, one using a series of <literal>if</literal> and
-  <literal>elseif</literal> statements, and the other using the
-  <literal>switch</literal> statement:
+  In the following example, each code block is equivalent.
+  One uses a series of <literal>if</literal> and
+  <literal>elseif</literal> statements, and the other a
+  <literal>switch</literal> statement.  In each case, the output is the same.
   <example>
    <title><literal>switch</literal> structure</title>
    <programlisting role="php">
 <![CDATA[
 <?php
-if ($i == 0) {
-    echo "i equals 0";
-} elseif ($i == 1) {
-    echo "i equals 1";
-} elseif ($i == 2) {
-    echo "i equals 2";
-}
+// This switch statement:
 
 switch ($i) {
     case 0:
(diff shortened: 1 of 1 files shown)


language/control-structures/switch.xml
```diff
@@ -181,6 +189,77 @@ switch ($i) {
...
+    case $start - 3:
+        print "C";
+        break;
+    case $start - 4:
+        print "C";
+        break;
+}
+?>
+]]>
+   </programlisting>
```
cmb69, line 232: I think we should mention the output of the example above. Not necessarily by adding an `&example.outputs;` section, but rather mentioning that it is "B".

language/control-structures/switch.xml
```diff
@@ -181,6 +189,77 @@ switch ($i) {
...
+        print "A";
+        break;
+    case $start - 2:
+        print "B";
+        break;
+    case $start - 3:
+        print "C";
+        break;
+    case $start - 4:
+        print "C";
```
cmb69, line 227: Just for the sake of clarity:
```suggestion
        print "D";
```

language/control-structures/switch.xml
```diff
@@ -181,6 +189,77 @@ switch ($i) {
...
+        break;
+}
+?>
+]]>
+   </programlisting>
+  </informalexample>
+ </para>
+ <para>
+  For more complex comparisons, use the value <literal>true</literal> as the switch value.
```
cmb69, line 236: I'd go with passive voice here ("the value … can be used"). Analogous for the next line.

language/control-structures/switch.xml
```diff
@@ -181,6 +189,77 @@ switch ($i) {
...
+        print "A";
+        break;
+    case $start - $offset === 2:
+        print "B";
+        break;
+    case $start - $offset === 3:
+        print "C";
+        break;
+    case $start - $offset === 4:
+        print "C";
```
cmb69, line 256:
```suggestion
        print "D";
```

language/control-structures/switch.xml
```diff
@@ -181,6 +189,77 @@ switch ($i) {
...
+        break;
+}
+?>
+]]>
+   </programlisting>
```
cmb69, line 261: Again, what is the output?

language/control-structures/switch.xml
```diff
@@ -66,6 +66,14 @@ switch ($i) {
...
 <![CDATA[
 <?php
+if ($i == 'apple') {
+    echo "i is apple";
+} elseif ($i == 'bar') {
+    echo "i is bar";
+} elseif ($i == 'cake') {
+    echo "i is cake";
+}
+
```
cmb69, line 76: That should have an explanation; maybe we should revert both parts to show the switch first, then have a comment "is the same as:", and then the if-elseif construct. We should probably do the same with the first example, although that has at least some introduction paragraph explaining what's going on there.

cmb69 (commented):
Thank you!

language/control-structures/switch.xml
Crell: The intro paragraph specifically says it applies to both code samples.  The second one was just missing the equivalent code.  (It's all kinda weird and silly, frankly.)  I'll fiddle with it a bit.

Crell: Not clearly, as I read it as being both examples.:-)  But in the end I concluded that having 2 examples offers exactly zero benefit over 1, at least for such similar examples, so I removed one and reworded it accordingly.

cmb69 (approved):
Thank you! Looks good to me.

There is the already existing "It's possible to use a semicolon instead of a colon after a case like:". And while it is indeed possible, I tend to discourage its use, since I'm pretty sure that this was only done to support the alternative syntax `<?php case "whatever"?>` where the trailing semicolon could be omitted.

Crell:
Er, not sure what the semicolon business is about?  That's already there in the docs, and while I agree no one should use it, it's worth noting that it exists, as is already done.  That's unrelated to this PR.

PR #1662, opened 2022-06-29
Fix XMLParser example for more recent PHP versions

cmb69:
The example uses the `$parser` as array index, but that implicit
conversion to integer already raises a notice as of PHP 7.0.0, and
outright fails as of PHP 8.0.0, since the `$parser` is an object now.

Since there doesn't appear a good reason for `$depth` to be an array
which supports different parsers, we change it to an integer, what also
simplifies the example.

This basically integrates user note 127199.

### DIFF
--- a/reference/xml/examples.xml
+++ b/reference/xml/examples.xml
@@ -13,27 +13,23 @@
 <![CDATA[
 <?php
 $file = "data.xml";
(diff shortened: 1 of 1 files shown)


Crell:
Merci bien!

PR #1664, opened 2022-06-29
Move Countable to predefined interfaces out from SPL

Girgias:
This might want a web change to redirect users to the new page.


cmb69:
> This might want a web change to redirect users to the new page.

Either that, or we keep the old page as alias (similar to [`sizeof()`](https://www.php.net/manual/en/function.sizeof.php)).

Crell:
The name is the same from userland, so I would probably favor a redirect.  That said, I don't know how we do redirects around here so maybe an alias page is easier. :upside_down_face:

Girgias:
> The name is the same from userland, so I would probably favor a redirect. That said, I don't know how we do redirects around here so maybe an alias page is easier. upside_down_face

Redirects need to be done via web-php

cmb69:
I don't think there is a respective PR for web-php, but the redirect would be as simple as adding to https://github.com/php/web-php/blob/8b59e9ce4274dbcebbbcdc16a9d55c6555a3e2aa/error.php#L263.

Anyhow, there are merge-conflicts in the PR which need to be resolved.

Girgias:
No redirection is needed as the URL would not change (it would still be `https://www.php.net/manual/en/countable.count.php` regardless of where it is placed as the ID of the root XML tag from the page does not change.

cmb69 (approved):
Thank you! In my opinion we can do without a changelog entry, albeit https://3v4l.org/Dr7I4.

Girgias:
Yeah, it's been in Core for a while, not sure a changelog is _that_ useful.

afilina:
As someone who upgrades a ton of really old legacy, I'd say that we should _always_ have a changelog entry, as I already have to go to core commits to answer my WTFs :)

I remember we were dealing with some unexpected BC break in the tokenizer for the PHPCompatibility project, and the lack of a changelog entry caught us off-guard and slowed us down.

Girgias:
It being in SPL or in Core is barely a distinction, we don't have changelogs for when Iterator got moved out of SPL into Core in 7.4

PR #1666, opened 2022-06-29
Add caveats to mkdir based on comments

Crell:
The useful bits of the comments that have now bitten the dust.

### DIFF
--- a/reference/filesystem/functions/mkdir.xml
+++ b/reference/filesystem/functions/mkdir.xml
@@ -28,9 +28,8 @@
      <term><parameter>directory</parameter></term>
      <listitem>
       <para>
-       The directory path.  It may be a local path string, or a
-       <link linkend="class.streamwrapper">stream wrapper</link> string if the
-       stream wrapper supports directory creation.
+       The directory path.
+       &tip.fopen-wrapper;
       </para>
      </listitem>
     </varlistentry>
--- a/reference/filesystem/functions/mkdir.xml
+++ b/reference/filesystem/functions/mkdir.xml
@@ -28,7 +28,9 @@
      <term><parameter>directory</parameter></term>
      <listitem>
       <para>
-       The directory path.
+       The directory path.  It may be a local path string, or a
+       <link linkend="class.streamwrapper">stream wrapper</link> string if the
+       stream wrapper supports directory creation.
       </para>
      </listitem>
     </varlistentry>
@@ -57,8 +59,8 @@
      <term><parameter>recursive</parameter></term>
      <listitem>
       <para>
-       Allows the creation of nested directories specified in the 
-       <parameter>directory</parameter>.
+       If &true;, then any parent directories to the <parameter>directory</parameter> specified will
+       also be created, with the same permissions.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/mkdir.xml
```diff
@@ -28,7 +28,9 @@
      <term><parameter>directory</parameter></term>
      <listitem>
       <para>
-       The directory path.
+       The directory path.  It may be a local path string, or a
+       <link linkend="class.streamwrapper">stream wrapper</link> string if the
+       stream wrapper supports directory creation.
```
cmb69, line 33: Not quite sure what to do with this. Most functions which accept a path are actually accepting any stream wrapper URL. This is, however, barely documented. Maybe we should have an XML entity, which we can use whereever appropriate.

reference/filesystem/functions/mkdir.xml
```diff
@@ -29,6 +29,7 @@
      <listitem>
       <para>
        The directory path.
+       &tip.fopen-wrapper;
```
cmb69, line 32: Actually, I think this entity is too verbose and too prominent. It should be sufficient to tell users that the parameter also accepts a stream wrapper URL, and link to a single location in the manual which explains all the details (i.e. do'nt link to `fopen()` *and* the wrappers). I'd probably go with a *sentence*, not even a `<simpara>`.

If you don't want to address this (now), feel free to leave as is.

reference/filesystem/functions/mkdir.xml
```diff
@@ -77,6 +78,14 @@
   <para>
    &return.success;
   </para>
+  <note>
+   <para>
+    If the directory to be created already exists, that is considered an error
+    and &false; will still be returned.  Use <function>is_dir</function> or
+    <function>file_exists</function> to check if the directory already exists
+    before trying to create it.
```
cmb69, line 86: Maybe reword in passive voice. Something like:
```suggestion
    and &false; will still be returned.  <function>is_dir</function> or
    <function>file_exists</function> can be used to check if the directory already exists
    before trying to create it.
```

cmb69 (commented):
Thanks for the PR!

reference/filesystem/functions/mkdir.xml
Crell: The passive voice is generally regarded by most writing scholars to be an inferior style, and that it should be avoided.  It keeps being recommended by you, but that is not understood by me.  The imperative tense is favored when directing someone to take a particular action.

cmb69: I'm not sure about this usage, but I'm merely referring to the [personalization section](http://doc.php.net/tutorial/style.php) of the style guide.

Crell: Avoiding "you" is fine (although I disagree that it's not professional to use), but the example there is but one way of doing so.  The imperative tense is sufficiently professional for technical writing when appropriate in context.

reference/filesystem/functions/mkdir.xml
Crell: So... go back to what I had previously, which was:

```
The directory path.  It may be a local path string, or a
       <link linkend="class.streamwrapper">stream wrapper</link> string if the
       stream wrapper supports directory creation.
```

reference/filesystem/functions/mkdir.xml
cmb69: Fair enough. :)

reference/filesystem/functions/mkdir.xml
cmb69: Well, but using an XML entity; also I think we need to merge the relevant info from `fopen()` and the wrappers. But like I said, it's okay to go with you have now.

PR #1668, opened 2022-06-30
fixed SOCKET_EBADF describe

mowangjuanzi:
> https://man7.org/linux/man-pages/man7/unix.7.html

> EBADF  This error can occur for [sendmsg(2)](https://man7.org/linux/man-pages/man2/sendmsg.2.html) when sending a file
>               descriptor as ancillary data over a UNIX domain socket
>               (see the description of SCM_RIGHTS, above), and indicates
>               that the file descriptor number that is being sent is not
 >              valid (e.g., it is not an open file descriptor).

I think it is more appropriate to replace  `Bad file number` with `Bad file descriptor`.

### DIFF
--- a/reference/sockets/constants.xml
+++ b/reference/sockets/constants.xml
@@ -494,7 +494,7 @@
    </term>
    <listitem>
     <simpara>
-     Bad file descriptor.
+     Bad file descriptor number.
     </simpara>
    </listitem>
   </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/sockets/constants.xml
```diff
@@ -494,7 +494,7 @@
    </term>
    <listitem>
     <simpara>
-     Bad file number.
+     Bad file descriptor.
```
cmb69, line 497: Maybe we should be very explicit:
```suggestion
     Bad file descriptor number.
```

cmb69 (commented):
Thank you! I think it's better, but might still be improved.

cc @devnexen

PR #1669, opened 2022-06-30
strtotime() fixes from comments

Crell:
I deleted another bunch of comments.  These are the changes worth keeping.

In particular, the current discussion of separator parsing is incorrect, as the new code sample shows.  It's quite bonkers.

### DIFF
--- a/reference/datetime/functions/strtotime.xml
+++ b/reference/datetime/functions/strtotime.xml
@@ -18,7 +18,10 @@
    and will try to parse that format into a Unix timestamp (the number of
    seconds since January 1 1970 00:00:00 UTC), relative to the timestamp given
    in <parameter>baseTimestamp</parameter>, or the current time if
-   <parameter>baseTimestamp</parameter> is not supplied.
+   <parameter>baseTimestamp</parameter> is not supplied.  The date string parsing
+   is defined in <link linkend="datetime.formats">Date and Time Formats</link>, and
+   has several subtle considerations.  Reviewing the full details there is strongly
+   recommended.
   </simpara>
   <warning>
    <para>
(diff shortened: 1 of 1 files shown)


reference/datetime/functions/strtotime.xml
```diff
@@ -97,6 +97,82 @@
...
+format('5/20/21');
+format('5.20.21');
+format('5-20-21');
+?>
+]]>
+    </programlisting>
+    &example.outputs.similar;
+    <screen>
+     <![CDATA[
+5/6/21 is interpreted as Thu, 06 May 2021 00:00:00 +0200
```
cmb69, line 164: The timezone offset [depends on the configured default timezone](https://3v4l.org/37l7k); either explicitly set the timezone, or point that out (a comment may suffice), or use some other format than `r` which doesn't show the timezone (offset) at all.

cmb69 (commented):
Thank you! Looks good to me, but maybe @derickr likes to review.

reference/datetime/functions/strtotime.xml
```diff
@@ -97,6 +97,82 @@
   </para>
  </refsect1>
 
+ <refsect1>
```
cmb69, line 100: The `<refsect1>` needs a `role` attribute. I think this section should either be integrated with the existing example section, or the existing notes section.

reference/datetime/functions/strtotime.xml
```diff
@@ -97,6 +97,82 @@
...
+   based on the separator used, as well as whether the date is specified using 2 digits or 4 digits.
+  </para>
+  <para>If the year component is 2 digits, then the algorithm is as follows:</para>
+  <simplelist>
+   <member>
+    Strings with a <literal>/</literal> separator are interpreted in US order.
+    That is, <literal>m/d/y</literal>.
+   </member>
+   <member>
+    Strings with a <literal>.</literal> separator are interpreted as a timestamp, not a date.
```
derickr, line 113: That's not called a `timestamp`, which is usually used for the Unix timestamp (Seconds since midnight 1970). I think you should just say `time`.

reference/datetime/functions/strtotime.xml
```diff
@@ -97,6 +97,82 @@
...
+   </member>
+  </simplelist>
+  <para>If the year component is 4 digits, then the algorithm is as follows:</para>
+  <simplelist>
+   <member>
+    Strings with a <literal>/</literal> separator are interpreted in US order.
+    That is, <literal>m/d/y</literal>.
+   </member>
+   <member>
+    Strings with a <literal>.</literal> separator are interpreted in European order.
```
derickr, line 128: `European` vs `US` order is quite western-worldy worded. I am not sure whether you can improve on that?

reference/datetime/functions/strtotime.xml
```diff
@@ -97,6 +97,82 @@
...
+    That is, <literal>d.m.y</literal>.
+   </member>
+   <member>
+    Strings with a <literal>-</literal> separator are interpreted in European order.
+    That is, <literal>d-m-y</literal>.
+   </member>
+  </simplelist>
+  <para>If the selected interpretation results in an illogical date (such as a month out of range), &false; will be returned.</para>
```
derickr, line 136: You might want to add saying that a day of of range for a specific month, does get parsed, but overflows happen.

reference/datetime/functions/strtotime.xml
```diff
@@ -139,6 +215,16 @@ if (($timestamp = strtotime($str)) === false) {
 
  <refsect1 role="notes">
   &reftitle.notes;
+  <note>
+   <para>
+    "Relative" date in this case also means that if a particular component
+    of the date/time stamp is not provided, it will be taken verbatim from
+    the <parameter>baseTimestamp</parameter>.  That is, <code>strtotime('February')</code>,
+    if run on the 31st of May 2022, will be interpreted as <literal>31 February 2022</literal>,
+    which will get mangled into a timestamp on <literal>3 March</literal>.
```
derickr, line 224: `mangled` → `overflowed` or something similar.

derickr: And maybe add that `first day of February` would make this work.

derickr (changes requested):
Left some comments, but I'm wondering whether we should continue to promote `strtotime` over the DateTime classes, where all of these parsing rules also apply. I have a vague idea to deprecate the functions that operate on just timestamps at some point vaguely soon.

reference/datetime/functions/strtotime.xml
Crell: I'm happy to use whatever better terminology you suggest here.

reference/datetime/functions/strtotime.xml
```diff
@@ -97,6 +97,82 @@
...
+  </para>
+  <para>If the year component is 2 digits, then the algorithm is as follows:</para>
+  <simplelist>
+   <member>
+    Strings with a <literal>/</literal> separator are interpreted in US order.
+    That is, <literal>m/d/y</literal>.
+   </member>
+   <member>
+    Strings with a <literal>.</literal> separator are interpreted as a timestamp, not a date.
+    That is, <literal>h.i.s</literal>.
```
guilliamxavier, line 114: The reality seems more complicated: https://github.com/php/doc-en/blob/d45ff7a3e78243470d1b7e5141d7ded022286d7f/reference/datetime/formats.xml#L509-L514 (example: https://3v4l.org/t7b56)

Crell:
Updates based on feedback above.  Most notably, removed the description of the different date separator rules (as they're even more bonkers than I thought), and just made the link to the existing format docs much more prominent.

@derickr I'm all for directing people away from this function toward something else.  If there's a specific "please use this instead" link I should include I can toss that into the description.  Otherwise we can punt on it until there's more internals movement on the subject.  Your call.

derickr (approved):

PR #1671, opened 2022-07-01
preg_replace additional notes

Crell:
The only things worth adding from the now-purged comments.

### DIFF
--- a/reference/pcre/functions/preg-replace.xml
+++ b/reference/pcre/functions/preg-replace.xml
@@ -67,6 +67,10 @@
        Note that backslashes in &string; literals may require to be escaped.
       </para>
       <para>
+       If the <parameter>replacement</parameter> array is associative, keys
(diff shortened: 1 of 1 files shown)


reference/pcre/functions/preg-replace.xml
```diff
@@ -296,6 +300,28 @@ xp***to
     <function>preg_replace</function>.
    </para>
   </note>
+  <note>
+   <para>
+    When both <parameter>pattern</parameter> and <parameter>replacement</parameter> are
```
cmb69, line 305: "When" or "if"? You know better than me. :)

cmb69 (approved):
Thank you!

reference/pcre/functions/preg-replace.xml
Crell: In context, either one is equally valid, I think.  At least in American English.

PR #1686, opened 2022-07-09
PHP 8.1: Update sanitize filters changelog with deprecations

thinkverse:
`FILTER_SANITIZE_STRING` and `FILTER_SANITIZE_STRIPPED` has been deprecated but were not added to the changelog in https://github.com/php/doc-en/pull/1193, only a note was added.

### DIFF
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -425,6 +425,13 @@ filter.default_flags = 0
       </thead>
       <tbody>
        <row>
+        <entry>8.1.0</entry>
+        <entry>
+         <constant>FILTER_SANITIZE_STRING</constant> and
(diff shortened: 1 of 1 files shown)


reference/filter/filters.xml
```diff
@@ -424,6 +424,13 @@ filter.default_flags = 0
        </row>
       </thead>
       <tbody>
+       <row>
+        <entry>8.1.0</entry>
+        <entry>
+         <constant>FILTER_SANITIZE_STRING</constant> and
+         <constant>FILTER_SANITIZE_STRIPPED</constant> has been deprecated.
```
cmb69, line 431: Shouldn't that be
```suggestion
         <constant>FILTER_SANITIZE_STRIPPED</constant> have been deprecated.
```

cmb69 (commented):
Thank you!

PR #1687, opened 2022-07-12
Update Upgrade history of the bundled PCRE library

cmb69:
Cf. <https://github.com/php/php-src/commit/32cceb75bf5e42bebbf4d12bab4369924f1d6885>.

### DIFF
--- a/reference/pcre/configure.xml
+++ b/reference/pcre/configure.xml
@@ -38,6 +38,11 @@
     </thead>
     <tbody>
      <row>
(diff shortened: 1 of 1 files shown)


devnexen (approved):
I m still newbie :) should we wait september before merging 8.2 PR ?

cmb69:
We should wait until the migration guide has been written (should happen during RC). After that, all PHP 8.2 PRs can be merged.

PR #1691, opened 2022-07-13
fread, references fseek. Closes #1690

### LINKED ISSUE #1690: fread needs to mention fseek
The PHP Manual page for fread doesn't list fseek in the See Also section. It should, since fseek is very valuable: it is the way to change the position in the file for the next fread. Currently, fseek is only mentioned in code in the Notes at the end of the entry. A similar issue applies to the page for fwrite.

devnexen:


cmb69 (approved):
Thank you!

PR #1693, opened 2022-07-14
added quotes around literal dot to diferntiate it from the full stop following imediatly after.

winstonDeGreef:


language/types/string.xml
```diff
@@ -1189,7 +1189,7 @@ bool(false)
   <note>
    <para>
     As of PHP 8.0.0, the decimal point character is always
-    <literal>.</literal>. Prior to PHP 8.0.0,
+    "<literal>.</literal>". Prior to PHP 8.0.0,
```
cmb69, line 1192: That would be an option, but why not spell it out:
```suggestion
    a full stop (<literal>.</literal>). Prior to PHP 8.0.0,
```
Not a native speaker, so I'm not sure whether it should be "full stop", or "dot", or "point" or whatever. Maybe @Crell can help.

cmb69 (commented):
Thank you for the PR! I agree that there is room for improvement here.

language/types/string.xml
winstonDeGreef: A full stop is the point character that comes after most sentences.
The character referred to is not a full stop but a point. I prefer putting the point character in quotes instead of spelling it out as it is visually a bit clearer.

Crell: Both spelling it out and including the character with quotes and a literal is what I'd favor.  Do all the things. :smile: 

As for the name, "period" is the name of the character itself.  As @winstonDeGreef notes, "full stop" is only applicable when it's ending a sentence.

```suggestion
    a period ("<literal>.</literal>"). Prior to PHP 8.0.0,
```

winstonDeGreef: I agree with @Crell

Crell: Oh dear god no, don't get into "smart quotes" and stuff.  That exists mainly to screw up character encoding in ways you won't be able to detect.  I've never had good experiences with that mess.

cmb69: Are you suggesting to stick with ASCII? What about the Japanese docs, for instance? Are these an encoding mess? I don't think so.

afilina: My two cents: we should probably accept @Crell's suggestion for now, and have a debate about quotes another day, as to not delay the PR :)

cmb69:
I went ahead and committed @Crell's suggestion (although Unicode calls this character `FULL STOP`).

cmb69:
Thanks all!

Issue #1682, opened 2022-07-05
Wrong result in complex comparisons switch's example's result


gilbertoalbino:
From manual page: https://php.net/control-structures.switch

Where it reads:
_For more complex comparisons, the value true may be used as the switch value. Or, alternatively, if-else blocks instead of switch._

The example's result is:
`// Prints "C"`

Shouldn't it be?:
`// Prints "B"`

damianwadley:
[Yes, it should.](https://3v4l.org/MEaTh)

But if we're in there, that code would make much more sense as `switch ($start - $offset) {`. Maybe a different example is in order?

cmb69:
I agree that we should use a different example. @Crell, an idea?

Crell:
Eh, I came up with this example, and I agree it's not a great one. :-)  I'm plenty to suggestions for a better one to show the "expression case" ability.

PR #1697, opened 2022-07-15
Update switch.xml

asiwak:
Corrected print output for more complex switch example


cmb69:
Thank you! It makes perfect sense to fix the example, even if we may replace it (see #1682).

Issue #1564, opened 2022-05-02
Enums don't support class hoisting


jpheldson:
### Description

Unlike class definitions, it seems that enum definitions currently need to come dynamically before they are used.  So far as I could find this is not in the documentation, and leads to very un-intuitive behavior.

The following code:

```php
<?php

$f = new Foo();
var_export($f->returnEnum());

enum FooEnum {
    case Bar;
    case Qux;
}

// If I had put the call to Foo::returnEnum() here, it would have worked!

class Foo {
    public function returnEnum(): FooEnum
    {
        return FooEnum::Bar;
    }
}
```

Resulted in this output:
```
Fatal error: Uncaught Error: Class "FooEnum" not found in /home/tmp/u.php:14
Stack trace:
#0 /home/tmp/u.php(4): Foo->returnEnum()
php/php-src#1 {main}
  thrown in /home/tmp/u.php on line 16
```

But I expected this output instead:
```
FooEnum::Bar
```

### PHP Version

PHP 8.1.2

### Operating System

FreeBSD 12.3-RELEASE r371270 NFSN64  amd64

cmb69:
> Maybe we can convert this to a documentation issue, […]

But how should the docs explain the complex actual behavior, except to point to [nikic's blog post](https://www.npopov.com/2021/10/20/Early-binding-in-PHP.html)?

If a simplified explanation is sufficient, we [already have](https://www.php.net/manual/en/language.oop5.inheritance.php):
> Unless autoloading is used, the classes must be defined before they are used. If a class extends another, then the parent class must be declared before the child class structure. This rule applies to classes that inherit other classes and interfaces.

This is, however, not sufficient for recent PHP versions.

jpheldson:
I know in my case, something like @iluuu1994's original reply as a notice in the documentation for Enums would have saved me a lot of time.

iluuu1994:
@cmb69 Maybe we just add that enums and classes with __toString() aren't supported without going in much more detail, and potentially adding a link to Nikitas blog (I'm not sure if we do links to other sources).

tiffany-taylor:
> (I'm not sure if we do links to other sources).

There are links to external sources, but it is limited. Hard to say if this is a valid case for that.

cmb69:
I think it is okay to link to that blog post; in the long run  it would be nice to have respective info on phpinternalsbook.com.

PR #1699, opened 2022-07-17
Added enum "Differences from objects" item regarding hoisting

ssddanbrown:
This is a simple addition aimed to address #1564.
Looking at the documentation, I thought this would be the natural place for someone check such a difference as per the original poster of #1564.

This doesn't add the blog post link, nor go into technical detail, like discussed as options in the thread since I thought that goes too far into implementation detail, which is not stated for any other (arguably more significant) differences detailed for enums.

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -585,8 +585,9 @@ $x = Direction::Up['short'];
    <member>Constructors and Destructors are forbidden.</member>
    <member>Inheritance is not supported. Enums may not extend or be extended.</member>
    <member>Static or object properties are not allowed.</member>
-   <member>Cloning an Enum case is not supported, as cases must be singleton instances</member>
+   <member>Cloning an Enum case is not supported, as cases must be singleton instances.</member>
    <member><link linkend="language.oop5.magic">Magic methods</link>, except for those listed below, are disallowed.</member>
+   <member>Enums must always be delcared before they are used.</member>
   </simplelist>
 
   <para>The following object functionality is available, and behaves just as it does on any other object:</para>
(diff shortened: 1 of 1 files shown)


language/enumerations.xml
```diff
@@ -585,8 +585,9 @@ $x = Direction::Up['short'];
    <member>Constructors and Destructors are forbidden.</member>
    <member>Inheritance is not supported. Enums may not extend or be extended.</member>
    <member>Static or object properties are not allowed.</member>
-   <member>Cloning an Enum case is not supported, as cases must be singleton instances</member>
+   <member>Cloning an Enum case is not supported, as cases must be singleton instances.</member>
    <member><link linkend="language.oop5.magic">Magic methods</link>, except for those listed below, are disallowed.</member>
+   <member>Enums must always be delcared before they are used.</member>
```
cmb69, line 590: typo:
```suggestion
   <member>Enums must always be declared before they are used.</member>
```

cmb69 (commented):
Thank you! Looks good (except for the typo).

language/enumerations.xml
ssddanbrown: @cmb69 Apologies, now fixed!

cmb69 (approved):
No problem! :)

Do you have commit karma?

ssddanbrown:
@cmb69 I'm assuming not, as I'm not familiar with that term.
This is only my second, rather small, contribution to the docs so not sure if that awards any kind of karma.

cmb69:
Sorry, I confused you with http://people.php.net/danbrown.

Regarding commit karma: I suggest that you do some more pull request first. :)

PR #1704, opened 2022-07-18
Attempt to add ODBC connection string functions

NattyNarwhal:
Unsure if the language here is clear enough. Also having difficulty with the output scripts, so unsure if these even are in the right place.

### DIFF
--- a/reference/uodbc/functions/odbc-connection-string-quote.xml
+++ b/reference/uodbc/functions/odbc-connection-string-quote.xml
@@ -14,7 +14,10 @@
   <para>
    Quotes a value for a connection string, according to ODBC rules.
    That is, it will be surrounded by quotes, and any ending curly
-   braces will be escaped.
+   braces will be escaped. This should be done for any connection
+   string values that come from user input. Not doing so can lead
+   to issues with parsing the connection string, or values being
+   injected into the connection string.
   </para>
   <para>
    Note that this function does not check if the string is already
--- a/reference/uodbc/functions/odbc-connection-string-quote.xml
+++ b/reference/uodbc/functions/odbc-connection-string-quote.xml
@@ -18,7 +18,7 @@
   </para>
   <para>
    Note that this function does not check if the string is already
-   quoted, nor if the string needs quoting. For that, you should call
+   quoted, nor if the string needs quoting. For that, call
    <function>odbc_connection_string_is_quoted</function> and
    <function>odbc_connection_string_should_quote</function>.
   </para>
@@ -49,8 +49,8 @@
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member>For seeing if a string is quoted: <function>odbc_connection_string_is_quoted</function></member>
-    <member>For seeing if a string should be quoted: <function>odbc_connection_string_should_quote</function></member>
+    <member><function>odbc_connection_string_is_quoted</function></member>
+    <member><function>odbc_connection_string_should_quote</function></member>
    </simplelist>
   </para>
  </refsect1>
--- a/reference/uodbc/functions/odbc-connection-string-quote.xml
+++ b/reference/uodbc/functions/odbc-connection-string-quote.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="function.odbc-connection-string-quote" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>odbc_connection_string_quote</refname>
+  <refpurpose>Quotes an ODBC connection string value</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>string</type><methodname>odbc_connection_string_quote</methodname>
+   <methodparam><type>string</type><parameter>str</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Quotes a value for a connection string, according to ODBC rules.
+   That is, it will be surrounded by quotes, and any ending curly
+   braces will be escaped.
+  </para>
+  <para>
+   Note that this function does not check if the string is already
+   quoted, nor if the string needs quoting. For that, you should call
+   <function>odbc_connection_string_is_quoted</function> and
+   <function>odbc_connection_string_should_quote</function>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>str</parameter></term>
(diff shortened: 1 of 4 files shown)


reference/uodbc/functions/odbc-connection-string-quote.xml
```diff
@@ -0,0 +1,79 @@
...
+   A quoted string, surrounded by curly braces, and properly escaped.
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member>For seeing if a string is quoted: <function>odbc_connection_string_is_quoted</function></member>
+    <member>For seeing if a string should be quoted: <function>odbc_connection_string_should_quote</function></member>
```
cmb69, line 53: There is no need to add some descriptive text here, since the `<refpurpose>` of the linked functions is automatically added. Currently, it looks like:

Screenshot 2022-07-19 133629

So:
```suggestion
    <member><function>odbc_connection_string_is_quoted</function></member>
    <member><function>odbc_connection_string_should_quote</function></member>
```

reference/uodbc/functions/odbc-connection-string-should-quote.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="function.odbc-connection-string-should-quote" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>odbc_connection_string_should_quote</refname>
+  <refpurpose>Description</refpurpose>
```
cmb69, line 5: There should be a meaningful description. :)

reference/uodbc/functions/odbc-connection-string-should-quote.xml
```diff
@@ -0,0 +1,78 @@
...
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>bool</type><methodname>odbc_connection_string_should_quote</methodname>
+   <methodparam><type>string</type><parameter>str</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+    Determines if a string needs to be quoted for an ODBC connection
+    string value; that is, if it contains special characters
```
cmb69, line 16:
```suggestion
    string value; that is, if it contains special characters.
```

reference/uodbc/functions/odbc-connection-string-should-quote.xml
```diff
@@ -0,0 +1,78 @@
...
+     </para>
+    </listitem>
+   </varlistentry>
+  </variablelist>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   True if the string should be quoted.
```
cmb69, line 43:
```suggestion
   &true; if the string should be quoted; &false; otherwise.
```

reference/uodbc/versions.xml
```diff
@@ -8,6 +8,9 @@
  <function name="odbc_binmode" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
  <function name="odbc_close" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
  <function name="odbc_close_all" from="PHP 4, PHP 5, PHP 7, PHP 8"/>
+ <function name="odbc_connection_string_is_quoted" from="PHP 8 &gt;= 8.2"/>
+ <function name="odbc_connection_string_should_quote" from="PHP 8 &gt;= 8.2"/>
+ <function name="odbc_connection_string_quote" from="PHP 8 &gt;= 8.2"/>
```
cmb69, line 13: We prefer the full version numbers:
```suggestion
 <function name="odbc_connection_string_is_quoted" from="PHP 8 &gt;= 8.2.0"/>
 <function name="odbc_connection_string_should_quote" from="PHP 8 &gt;= 8.2.0"/>
 <function name="odbc_connection_string_quote" from="PHP 8 &gt;= 8.2.0"/>
```

reference/uodbc/functions/odbc-connection-string-quote.xml
```diff
@@ -0,0 +1,79 @@
...
+   <methodparam><type>string</type><parameter>str</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Quotes a value for a connection string, according to ODBC rules.
+   That is, it will be surrounded by quotes, and any ending curly
+   braces will be escaped.
+  </para>
+  <para>
+   Note that this function does not check if the string is already
+   quoted, nor if the string needs quoting. For that, you should call
```
cmb69, line 21: We prefer to avoid personalization ("you"). In this case maybe just drop the "you should":
```suggestion
   quoted, nor if the string needs quoting. For that, call
```
Although I'm not sure whether we should mention these functions here, as there are already in the see-also section.

reference/uodbc/functions/odbc-connection-string-quote.xml
```diff
@@ -0,0 +1,79 @@
...
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   A quoted string, surrounded by curly braces, and properly escaped.
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
```
cmb69, line 48: I think adding an example would be helpful for users. You could use https://github.com/php/doc-base/blob/96422cd00d50a923b3f32929c7b15172cf896816/RFC/skeletons/function.xml#L84-L112 as template.

cmb69 (commented):
Thank you for the PR! A really good start! :)

NattyNarwhal:
Made the changes except for a sample, which I'll add. I did figure out my local issues with the toolchain too.

NattyNarwhal:
Hmm, is the whitespace check having a good one? It seems to be failing on a file not touched by these changes.

cmb69:
Oh, the whitespace checker is [only triggered for PRs](https://github.com/php/doc-en/blob/56e9ece746b317a078d0eadae19931b8c96fac64/.github/workflows/check-whitespace.yml#L8-L10), but the files in question have been committed directly. I'm going to fix these issues. It's probably a good idea to also trigger the whitespace checker for commits.

NattyNarwhal:
Just checking since it's been a month here, and the whitespace message was a false positive. Would like to get this in for 8.2. If there's any other changes I should make, let me know.

cmb69 (commented):
Besides the erroneous function names, this looks good to me. Thank you!

Note that we usually merge PHP 8.2 documentation only after the migration guide has been written (what is supposed to happen during the RC phase).

cmb69 (approved):
Thank you!

cmb69:
> Can I merge this PR?

Yes, please! :)

PR #1705, opened 2022-07-20
QuickHash improvement

saundefined:

PR #1706, opened 2022-07-20
Fix foreach links

saundefined:
We'll fix that too :)

Relates https://github.com/php/doc-en/pull/1705#discussion_r925385065


cmb69 (approved):
Nice. Thank you!

Issue #1475, opened 2022-03-23
Examples for filter_var


gesior:
From manual page: https://php.net/function.filter-var

---
There is no example with `if` and I saw a lot of implementations like:
```
if (!filter_var($value, $type)) {
    echo 'error';
}
```
and it should be:
```
if (filter_var($value, $type) !== $value) {
    echo 'error';
}
```
Especially, when you validated with `FILTER_VALIDATE_BOOLEAN` filter.

cmb69:
Comparing to the passed value won't necessarily work for sanitizing filters. You usually want to compare against `=== false`, unless you're using `FILTER_VALIDATE_BOOL`, in which case it might be best to specify a `default`. And of course, if you set `FILTER_NULL_ON_FAILURE`, comparison against `=== null` is needed.

PR #1713, opened 2022-07-22
FILTER_VALIDATE_BOOL: clarify non-string values

meldafert:
It is not clear from the documentation that this filter accepts more than the listed values.
Eg. when using `FILTER_NULL_ON_FAILURE`, `null` is accepted (returns `false`, not `null`), which can be surprising.
Similarly, `'     1      '` (leading/trailing whitespace), `1.0` (float), and `new class { function __toString() { return '1'; } }` are also accepted.

It might also make sense to specifically add these values to the example values - cc #1475 for that.

### DIFF
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -43,7 +43,6 @@
           &null; is returned for all non-boolean values.
          </para>
          <para>
-          Non-string values are first converted to a string using <function>strval</function>.
           String values are trimmed using <function>trim</function> before comparison.
          </para>
         </entry>
@@ -107,7 +106,14 @@
          <constant>FILTER_FLAG_ALLOW_THOUSAND</constant>,
          <constant>FILTER_NULL_ON_FAILURE</constant>
         </entry>
-        <entry>Validates value as float, optionally from the specified range, and converts to float on success.</entry>
+        <entry>
+          <para>
+            Validates value as float, optionally from the specified range, and converts to float on success.
+          </para>
+          <para>
+            String values are trimmed using <function>trim</function> before comparison.
+          </para>
+        </entry>
        </row>
        <row>
         <entry><constant>FILTER_VALIDATE_INT</constant></entry>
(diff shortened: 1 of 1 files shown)


meldafert:
The relevant PHP source code is [here](https://github.com/php/php-src/blob/b1575f996cb0a1c11e4771d01b6b6c93af644d9c/ext/filter/logical_filters.c#L273-279), for reference:

```c
void php_filter_boolean(PHP_INPUT_FILTER_PARAM_DECL) /* {{{ */
{
	char *str = Z_STRVAL_P(value);
	size_t len = Z_STRLEN_P(value);
	int ret;

	PHP_FILTER_TRIM_DEFAULT_EX(str, len, 0);

	/* returns true for "1", "true", "on" and "yes"
	 * returns false for "0", "false", "off", "no", and ""
	 * null otherwise. */
	switch (len) {
```

cmb69:
Thank you for the PR!

Note that the string conversion is not special for `FILTER_VALIDATE_BOOL`, but rather applies to all filters. `filter_var()` documentation already mentions that, so we shouldn't add that info to that particular filter. The info regarding the trimming makes sense, though.

meldafert:
@cmb69 Thank you, I also just found that. I will remove the mention of the string conversion for `FILTER_VALIDATE_BOOL`, and only leave the trimming note.
It appears that `FILTER_VALIDATE_INT` and `FILTER_VALIDATE_FLOAT` also trim the value before checking it, I will add a note to these filters as well.
It might make sense to mention the string conversion again at the head of the "Validate Filters" page, since I assume that many people will (like me) only check this page to find out accepted values?

cmb69:
> It might make sense to mention the string conversion again at the head of the "Validate Filters" page, since I assume that many people will (like me) only check this page to find out accepted values?

Not sure about that. Usually, you call the `filter_*()` functions with strings (or arrays).

cmb69:
Thank you!

PR #1714, opened 2022-07-22
Add call-user-func-array Changelog

bburnichon:
Indicates change in keys handling

### DIFF
--- a/reference/funchand/functions/call-user-func-array.xml
+++ b/reference/funchand/functions/call-user-func-array.xml
@@ -50,7 +50,7 @@
   </para>
  </refsect1>
 
-  <refsect1 role="changelog">
+ <refsect1 role="changelog">
   &reftitle.changelog;
   <para>
    <informaltable>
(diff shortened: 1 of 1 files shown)


reference/funchand/functions/call-user-func-array.xml
```diff
@@ -50,6 +50,30 @@
...
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <parametername>args</parametername> keys will now be interpreted as parameter names, instead of being silently ignored..
```
cmb69, line 68: Double full-stop:
```suggestion
        <parametername>args</parametername> keys will now be interpreted as parameter names, instead of being silently ignored.
```

cmb69 (changes requested):
Thank you!

bburnichon:
Looks like I did not follow the DTD for the XML. I am not used to this one and do not know what I should do. I took an example somewhere else in the doc. Any hints?

cmb69:
There are skeletons for most of the documentation pages in https://github.com/php/doc-base/tree/master/RFC/skeletons/.

Anyway, this looks good now. Thank you!

PR #1715, opened 2022-07-23
Update document for jit_max_root_traces

wxue1:
Add more explanation for root trace and side trace in PHP manual.
Includes the influence when reach the maximum value of
opcache.jit_max_root_traces or opcache.jit_max_side_traces.

Signed-off-by:Wang, Xue <xue1.wang@intel.com>


cmb69:
Thank you!

PR #1725, opened 2022-07-28
Parameter offset explanation refers to field when it should be row.

arius28:
From manual page: https://php.net/mysqli-result.data-seek

---

In the parameters section the part explaining the offset has the following line. 

> The field offset. Must be between zero and the total number of rows minus one (0..[mysqli_num_rows()](https://www.php.net/manual/en/mysqli-result.num-rows.php) - 1).

Shouldn't the **field offset** be **row offset** ?


cmb69:
Indeed. Thank you!

PR #1726, opened 2022-07-28
Rewrite type section

Girgias:
To incorporate typing features added in PHP 8.2.0 and also (hopefully) future proof for additional additions.

This is still a bit WIP as I haven't added DNF types but I already wanted to get some eyes on it to gather some opinions.
I've taken a bit of inspiration from the Rust manual: https://doc.rust-lang.org/reference/types.html

### DIFF
--- a/language/types.xml
+++ b/language/types.xml
@@ -7,8 +7,8 @@
   <title>Introduction</title>
 
   <para>
-   Every single expression and variable in PHP resolves to one of the following
-   built-in type primitive types depending on its value:
+   Every single expression in PHP has one of the following
+   built-in types depending on its value:
    <itemizedlist>
     <listitem><simpara><type>null</type></simpara></listitem>
     <listitem><simpara><type>bool</type></simpara></listitem>
--- a/language/types.xml
+++ b/language/types.xml
@@ -14,6 +14,7 @@
     <listitem><simpara><type>bool</type></simpara></listitem>
     <listitem><simpara><type>int</type></simpara></listitem>
     <listitem><simpara><type>float</type> (floating-point number)</simpara></listitem>
+    <listitem><simpara><type>string</type></simpara></listitem>
     <listitem><simpara><type>array</type></simpara></listitem>
     <listitem><simpara><type>object</type></simpara></listitem>
     <listitem><simpara><type>callable</type></simpara></listitem>
@@ -40,13 +41,13 @@
    <link linkend="language.types.type-juggling">Type Juggling</link>.
   </para>
 
-  <caution>
+  <tip>
    <simpara>
     <link linkend="types.comparisons">The type comparison tables</link>
     may also be useful, as various examples of comparison between values of
     different types are present.
    </simpara>
-  </caution>
+  </tip>
   
   <note>
    <simpara>
@@ -65,6 +66,7 @@
    <link linkend="language.expressions">expression</link>,
    use the <function>get_debug_type</function> function.
    However, to check if an expression is of a certain type use the
+   <!-- TODO When PhD support is there: <function>is_<replaceable>type</replaceable></function> -->
    <literal>is_<replaceable>type</replaceable></literal> functions instead.
 
    <informalexample>
--- a/language/types.xml
+++ b/language/types.xml
@@ -109,37 +109,45 @@
 
    <para>
     To get a human-readable representation of a type for debugging, use the
-    <function>gettype</function> function. To check for a certain type, do
-    <emphasis>not</emphasis> use <function>gettype</function>, but rather the
-    <literal>is_<replaceable>type</replaceable></literal> functions. Some
-    examples:
+    <function>get_debug_type</function> function. To check for a certain type,
+    do <emphasis>not</emphasis> use <function>get_debug_type</function>, but
+    rather the <literal>is_<replaceable>type</replaceable></literal> functions.
+    Some examples:
    </para>
    
    <informalexample>
     <programlisting role="php">
 <![CDATA[
 <?php
-$a_bool = TRUE;   // a boolean
+$a_bool = true;   // a bool
 $a_str  = "foo";  // a string
 $a_str2 = 'foo';  // a string
-$an_int = 12;     // an integer
+$an_int = 12;     // an int
 
-echo gettype($a_bool); // prints out:  boolean
-echo gettype($a_str);  // prints out:  string
+echo get_debug_type($a_bool), "\n";
+echo get_debug_type($a_str), "\n";
 
 // If this is an integer, increment it by four
 if (is_int($an_int)) {
     $an_int += 4;
 }
+var_dump($an_int);
 
 // If $a_bool is a string, print it out
-// (does not print out anything)
 if (is_string($a_bool)) {
     echo "String: $a_bool";
 }
 ?>
 ]]>
     </programlisting>
+    &example.outputs.8;
+    <screen>
+<![CDATA[
+bool
+string
+int(16)
+]]>
+    </screen>
    </informalexample>
   </note>
 
--- a/language/types.xml
+++ b/language/types.xml
@@ -8,7 +8,7 @@
 
   <para>
    Every single expression and variable in PHP resolves to one of the following
-   nine built-in type primitive types depending on its value:
+   built-in type primitive types depending on its value:
    <itemizedlist>
     <listitem><simpara><type>null</type></simpara></listitem>
     <listitem><simpara><type>bool</type></simpara></listitem>
--- a/language/types.xml
+++ b/language/types.xml
@@ -5,116 +5,67 @@
 
  <sect1 xml:id="language.types.intro">
   <title>Introduction</title>
-  
-  <simpara>
-   PHP supports ten primitive types.
-  </simpara>
-  
-  <para>
-   Four scalar types:
-  </para>
-
-  <itemizedlist>
-
-   <listitem>
-    <simpara>
-     <type>bool</type>
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>int</type>
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>float</type> (floating-point number)
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>string</type>
-    </simpara>
-   </listitem>
-
-  </itemizedlist>
 
   <para>
-   Four compound types:
+   Every single expression in PHP resolves to a type, which will be one
+   of these nine primitive types:
+   <itemizedlist>
+    <listitem><simpara><type>null</type></simpara></listitem>
+    <listitem>
+     <simpara>Scalar types:</simpara>
+     <itemizedlist>
+      <listitem><simpara><type>bool</type></simpara></listitem>
+      <listitem><simpara><type>int</type></simpara></listitem>
+      <listitem><simpara><type>float</type> (floating-point number)</simpara></listitem>
+      <listitem><simpara><type>string</type></simpara></listitem>
+     </itemizedlist>
+    </listitem>
+    <listitem><simpara><type>array</type></simpara></listitem>
+    <listitem><simpara><type>object</type></simpara></listitem>
+    <listitem><simpara><type>callable</type></simpara></listitem>
+    <listitem><simpara><type>resource</type></simpara></listitem>
+   </itemizedlist>
   </para>
-
-  <itemizedlist>
-
-   <listitem>
-    <simpara>
-     <type>array</type>
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>object</type>
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>callable</type>
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>iterable</type>
-    </simpara>
-   </listitem>
-
-  </itemizedlist>
 
   <para>
-   And finally two special types:
+   In general, the type of an expression or variable is determined at runtime
+   by PHP depending on the value. However, it is sometimes possible to add
+   <link linkend="language.types.declarations">type declarations</link>
+   to ensure the value is of a specific type.
   </para>
 
-  <itemizedlist>
-
-   <listitem>
-    <simpara>
-     <type>resource</type>
-    </simpara>
-   </listitem>
-
-   <listitem>
-    <simpara>
-     <type>NULL</type>
-    </simpara>
-   </listitem>
-
-  </itemizedlist>
+  <caution>
+   <simpara>
+    An expression/variable might be evaluated as a different type depending
+    on the context it is used. For more information, see the section on
+    <link linkend="language.types.type-juggling">Type Juggling</link>.
+   </simpara>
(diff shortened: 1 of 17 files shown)


language/types/declarations.xml
```diff
@@ -10,141 +10,131 @@
...
+   <title>Scalar types</title>
    <para>
-    Aliases for the above scalar types are not supported.
-    Instead, they are treated as class or interface names.
-    For example, using <literal>boolean</literal> as a type declaration
-    will require the value to be an &instanceof; the class or interface
-    <literal>boolean</literal>, rather than of type <type>bool</type>:
+    The value of a scalar type declaration must be a value of the specified
+    type.
+    An exception exists that the <type>float</type> will always accept values
```
iluuu1994, line 110:
```suggestion
    An exception exists that the type <type>float</type> will always accept values
```

language/types/declarations.xml
```diff
@@ -154,23 +144,191 @@ Stack trace:
...
    </para>
-  </warning>
+  </sect3>
 
- <sect3 xml:id="language.types.declarations.mixed">
-  <title>mixed</title>
+  <sect3 xml:id="language.types.declarations.base.object">
+   <title><type>object</type> type</title>
+   <para>
+    The value must be an &instanceof; any class or enumeration case.
```
iluuu1994, line 164:
```suggestion
    The value must be an &instanceof; any class or enumeration.
```

Enumeration case would be the equivalent of an instance, so one of them is redundant.

language/types/declarations.xml
```diff
@@ -154,23 +144,191 @@ Stack trace:
...
+   <sect4 xml:id="language.types.declarations.base.class-specific.self">
+    <title><type>self</type></title>
+    <para>
+     The value must be an &instanceof; the same class as the one
+     in which the type declaration is used.
+    </para>
+   </sect4>
+
+   <sect4 xml:id="language.types.declarations.base.class-specific.parent">
+    <title><type>self</type></title>
```
iluuu1994, line 183:
```suggestion
    <title><type>parent</type></title>
```

language/types/declarations.xml
```diff
@@ -154,23 +144,191 @@ Stack trace:
...
+     in which the type declaration is used.
+    </para>
+   </sect4>
+
+   <sect4 xml:id="language.types.declarations.base.class-specific.static">
+    <title>static</title>
+    <para>
+     <type>static</type> is a return-only type which requires that the
+     value returned must be an &instanceof; the same class as the one
+     the method is called in.
```
iluuu1994, line 195:
```suggestion
     value returned must be an &instanceof; the same class as the class of the object itself.
```

Maybe the wording can be improved. But it has nothing to do with the caller, you can call said method outside of the given class hierarchy.

language/types/declarations.xml
```diff
@@ -154,23 +144,191 @@ Stack trace:
...
+  <sect3 xml:id="language.types.declarations.void">
+   <title>void</title>
+   <para>
+    <literal>void</literal> is a return-only type declaration indicating the
+    function does not return a value.
+    Therefore, it cannot be part of a
+    <link linkend="language.types.declarations.composite.union">union type</link>
+    declaration. Available as of PHP 7.1.0.
+   </para>
+   <note>
```
iluuu1994, line 253: Maybe a note could be added that `void` functions actually still return `null`.

language/types/declarations.xml
```diff
@@ -154,23 +144,191 @@ Stack trace:
...
+     Consider using an enumeration instead.
+    </simpara>
+   </note>
+  </sect3>
+
+  <sect3 xml:id="language.types.declarations.base.user-defined">
+   <title>User-defined types</title>
+   <para>
+    <!-- todo links -->
+    A user-defined is any interface, class, or enumeration.
```
iluuu1994, line 302:
```suggestion
    A user-defined type is any interface, class, or enumeration.
```

language/types/declarations.xml
```diff
@@ -154,23 +144,191 @@ Stack trace:
...
+
+   <note>
+    <simpara>
+     These types are generally referred to as class-types.
+    </simpara>
+   </note>
+  </sect3>
+
+  <sect3 xml:id="language.types.declarations.base.function">
+   <title>Function types</title>
```
iluuu1994, line 320: Maybe `\Closure` should be mentioned, and the fact that it is an object. It's also not completely clear we're talking about the `callable` type here.

iluuu1994 (commented):
I'm not super familiar with the documentation format so some of these might be wrong (e.g. I'm not sure if `<type>` does something special).

cmb69:
> I'm not sure if `<type>` does something special

Actually, that is supposed to create a link, but IIRC this is broken for quite a while now. Anyhow, could use [XML entities](https://github.com/php/doc-base/blob/ba9dbf211b61ea6e18fde8e01f52d9e8f4d8b59a/entities/global.ent#L556-L564) instead.

language/types/declarations.xml
Girgias: Right, I wasn't totally sure about that so I've made the distinction, but good to know

language/types/declarations.xml
Girgias: This is the current wording that already exist. I'll think about improving it

language/types/declarations.xml
Girgias: We have a dedicated page on the callable type https://www.php.net/manual/en/language.types.callable.php and I thought about mentioning ``\Closure`` but couldn't totally figure out how to write this section.

language/types/declarations.xml
iluuu1994: Maybe we can explain with an example. "For example, an object `Cat` inheriting `Animal::getClone(): static` must return a `Cat`, not just any `Animal`, which would be the behavior of `self`."

language/types/declarations.xml
iluuu1994: Ah ok, maybe just linking it should be fine then.

language/types/declarations.xml
```diff
@@ -10,141 +10,131 @@
...
+      <listitem>
+       <simpara>Classes</simpara>
+      </listitem>
+      <listitem>
+       <simpara>Enumerations</simpara>
+      </listitem>
+     </itemizedlist>
+    </listitem>
+    <listitem>
+     <simpara>Function types</simpara>
```
Crell, line 100: We have function types?  I didn't think we had that yet.  I don't know what you mean otherwise.

Should `callable` also be in this list?

language/types/declarations.xml
```diff
@@ -10,141 +10,131 @@
...
-   </tgroup>
-  </informaltable>
+  <title>Base types</title>
+  <para>
+   Some base types are built-in types which are tightly integrated with the
+   language and cannot be reproduced with user defined types.
+  </para>
 
-  <warning>
+  <para>
```
Crell, line 35: I'm not sure about moving this to a list instead of a table.  The table seems more readable to me for this, no?

language/types/declarations.xml
```diff
@@ -154,23 +144,197 @@ Stack trace:
...
+    Meaning it is the subtype of every other type and can replace any other
+    return type during inheritance.
+   </para>
+  </sect3>
+
+  <sect3 xml:id="language.types.declarations.void">
+   <title>void</title>
+   <para>
+    <literal>void</literal> is a return-only type declaration indicating the
+    function does not return a value.
```
Crell, line 248:
```suggestion
    function does not return a value, but the function may still terminate.
```

language/types/declarations.xml
```diff
@@ -154,23 +144,197 @@ Stack trace:
...
+     <link linkend="language.types.declarations.composite.union">union type</link>.
+    </simpara>
+   </warning>
+  </sect3>
+
+  <sect3 xml:id="language.types.declarations.never">
+   <title>Bottom type - never</title>
+   <para>
+    <literal>never</literal> is a return-only type indicating the function
+    does not return. This means that it either calls <function>exit</function>,
```
Crell, line 231:
```suggestion
    does not terminate. This means that it either calls <function>exit</function>,
```

language/types/declarations.xml
```diff
@@ -154,23 +144,197 @@ Stack trace:
...
+    <literal>void</literal> is a return-only type declaration indicating the
+    function does not return a value.
+    Therefore, it cannot be part of a
+    <link linkend="language.types.declarations.composite.union">union type</link>
+    declaration. Available as of PHP 7.1.0.
+   </para>
+   <note>
+    <simpara>
+     Even if a function has a return type of <literal>void</literal> it will
+     still return a value, this value is always &null;.
```
Crell, line 256:
```suggestion
     still evaluate to a value in the calling code, specifically &null;.
```

language/types/declarations.xml
```diff
@@ -154,23 +144,197 @@ Stack trace:
...
+    <simpara>
+     Prior to PHP 8.2.0 the <literal>false</literal> type could only be used as part of a
+     <link linkend="language.types.declarations.composite.union">union type</link>.
+    </simpara>
+   </warning>
+
+   <note>
+    <simpara>
+     It is not possible to define custom literal types.
+     Consider using an enumeration instead.
```
Crell, line 299: Link to the enum page?

language/types/declarations.xml
```diff
@@ -10,141 +10,131 @@
...
+  <title>Base types</title>
+  <para>
+   Some base types are built-in types which are tightly integrated with the
+   language and cannot be reproduced with user defined types.
+  </para>
 
-  <warning>
+  <para>
+   The list of base types is:
+   <itemizedlist>
```
Crell, line 37: Whether it's a list or stays as a table, each item should deep-link to the spot in the page where it's described.

Crell:
Glancing over the section... I barely understand why this page is what it is and where it is.  Shouldn't a description of different types come way earlier in the section?  It's basically at the end now.  Is the whole Types section due for an overhaul?

language/types/declarations.xml
iluuu1994: It actually does just return the value, just implicitly. So IMO this is accurate.

language/types/declarations.xml
Girgias: Well a ``callable`` is an extremely generic function type. And as I know @iluuu1994 is working on them I thought I could anticipate some doc writing. But I'm seeing this is being very confusing so I'll revert to using ``callable`` types for now.

language/types/declarations.xml
Girgias: I prefer the list as it allows nesting, also the table currently has no real grouping (and I don't know how to make it sensible) + a version release column which acts a bit weirdly live (might be some CSS issue)

Girgias:
> Glancing over the section... I barely understand why this page is what it is and where it is. Shouldn't a description of different types come way earlier in the section? It's basically at the end now. Is the whole Types section due for an overhaul?

Yes, probably, I wrote this page back when PHP 8.0 got released as the information about type declarations where scattered in loads of different places. So this was done to centralise it.

Girgias:
> > I'm not sure if `<type>` does something special
> 
> Actually, that is supposed to create a link, but IIRC this is broken for quite a while now. Anyhow, could use [XML entities](https://github.com/php/doc-base/blob/ba9dbf211b61ea6e18fde8e01f52d9e8f4d8b59a/entities/global.ent#L556-L564) instead.

https://github.com/php/phd/pull/33 is meant to fix it, might need to go back to it

language/types/declarations.xml
Crell: Does it *return* a value, or *evaluate to* a value?  I've always understood it as "a function call that doesn't return evaluates to null".  Because a function with return void isn't allowed to have a return statement in it with any value at all, so it cannot return null, but it can evaluate to null by default.

(I don't know what the engine thinks of that, but that's how I've always conceptualized it from user space.)

Crell: Huh.  TIL.  Also, PHP is screwy.

cmb69 (commented):
It seems to me we're mixing "types" and "type declarations", but we should better not. I think it's better to move much of what we have on the type declaration page to the [introduction page](https://www.php.net/manual/en/language.types.intro.php), and to have only info about type declarations on the former (general info about type declarations, and the special declaration types like `self`).

Girgias:
I don't really think I've addressed your comment @cmb69, but I'm still really blanking on how to break the relevant bits away, as I wrote it all from writing a type declaration perspective :-/

Girgias:
Okay, I've spent some time again rewritting this, and would love to get your comments and opinions on this.

language/types/type-juggling.xml
```diff
@@ -169,18 +188,93 @@
...
+     targets for implicit coercion. In particular no implicit coercions to
+     the <literal>null</literal> and <literal>false</literal> types occur.
+    </para>
+   </note>
+
+   <example>
+    <title>Example of types being coerced into a type part of the union</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
```
cmb69, line 251: This doesn't look like PHP code:
```suggestion
    <programlisting>
<![CDATA[
```
(accordingly, the `?>` needs to be removed as well)

cmb69 (commented):
Thank you! This is quite something.

language/types.xml
```diff
@@ -5,177 +5,134 @@
...
-   </listitem>
-
-  </itemizedlist>
+  <caution>
+   <simpara>
+    <link linkend="types.comparisons">The type comparison tables</link>
+    may also be useful, as various examples of comparison between values of
+    different types are present.
+   </simpara>
+  </caution>
```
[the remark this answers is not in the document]
Girgias: Oh I forgot this existed, that is indeed better.

language/types/type-system.xml
```diff
@@ -0,0 +1,187 @@
...
+        Scalar types:
+        <type>bool</type>, <type>int</type>, <type>float</type>, and <type>string</type>
+       </simpara>
+      </listitem>
+      <listitem>
+       <simpara><type>array</type> type</simpara>
+      </listitem>
+      <listitem>
+       <simpara><type>object</type> type</simpara>
+      </listitem>
```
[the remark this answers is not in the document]
Girgias: For me, "primitive" is built-into the engine, and you cannot recreate them in userland. But I can change it.

Girgias:
Okay, I've only performed limited changes to try to address all the review comments. Hopefully this is now good to land and be stabbed by the pitchforks of translators.

@saundefined @mumumu what are your opinions of this as the translators of the 2 other main up to date languages?

saundefined:
> @saundefined @mumumu what are your opinions of this as the translators of the 2 other main up to date languages?

LGTM, thanks!

mumumu:
@Girgias Personally, I don't have a problem with it. This PR change is not mahoosive than I expected.

cmb69 (commented):
Thank you!

kitrio:
I found some type 
File  language/types/type-system.xml 
line 48 <simpara><type>float</type> string</simpara>
it seems like typo
because line 45 already write float   <simpara><type>float</type> type</simpara>



line 48 <simpara><type>string</type> type</simpara>
Is this right?

Thank you.

cmb69:
Thanks for reporting, @kitrio! Should be fixed with https://github.com/php/doc-en/commit/23cda460978ca25b9c0be003862095cd0eb20b00.

PR #1728, opened 2022-07-29
sleep() additional notes

rxall:
add note about usleep vs sleep usage mirroring the top comment to help myself from yesterday

### DIFF
--- a/reference/misc/functions/sleep.xml
+++ b/reference/misc/functions/sleep.xml
@@ -12,10 +12,10 @@
    <type>int</type><methodname>sleep</methodname>
    <methodparam><type>int</type><parameter>seconds</parameter></methodparam>
   </methodsynopsis>
-  <para>
+  <simpara>
    Delays the program execution for the given number of
    <parameter>seconds</parameter>.
-  </para>
+  </simpara>
   <note>
    <para>
     In order to delay program execution for a fraction of a second, use <function>usleep</function>
--- a/reference/misc/functions/sleep.xml
+++ b/reference/misc/functions/sleep.xml
@@ -12,10 +12,17 @@
    <type>int</type><methodname>sleep</methodname>
    <methodparam><type>int</type><parameter>seconds</parameter></methodparam>
   </methodsynopsis>
-  <simpara>
+  <para>
    Delays the program execution for the given number of
    <parameter>seconds</parameter>.
-  </simpara>
+  </para>
+  <note>
+   <para>
+    In order to delay program execution for a fraction of a second, use <function>usleep</function>
+    as the <function>sleep</function> will cast input to an integer. For example, sleep(0.25) will 
+    pause program execution for 0 seconds.
+   </para>
+  </note>
  </refsect1>
 
  <refsect1 role="parameters">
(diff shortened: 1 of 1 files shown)


reference/misc/functions/sleep.xml
```diff
@@ -12,10 +12,17 @@
    <type>int</type><methodname>sleep</methodname>
    <methodparam><type>int</type><parameter>seconds</parameter></methodparam>
   </methodsynopsis>
-  <simpara>
+  <para>
```
cmb69, line 15: Any particular reason for this change? I prefer `<simpara>` over `<para>` where possible, and if mostly matches the rest of the file.

reference/misc/functions/sleep.xml
```diff
@@ -12,10 +12,17 @@
...
    Delays the program execution for the given number of
    <parameter>seconds</parameter>.
-  </simpara>
+  </para>
+  <note>
+   <para>
```
cmb69, line 20: I would use a `<simpara>` here.

reference/misc/functions/sleep.xml
```diff
@@ -12,10 +12,17 @@
...
+    In order to delay program execution for a fraction of a second, use <function>usleep</function>
+    as the <function>sleep</function> will cast input to an integer. For example, sleep(0.25) will
```
cmb69, line 22: The argument will not really be case to an integer. In coercive mode (the default), it will be coerced, but in strict type mode, it will be rejected. As such I'd reword:
```suggestion
    as the <function>sleep</function> function expects an &integer;. For example, <code>sleep(0.25)</code> will 
```

reference/misc/functions/sleep.xml
```diff
@@ -12,10 +12,17 @@
...
+  <para>
    Delays the program execution for the given number of
    <parameter>seconds</parameter>.
-  </simpara>
+  </para>
+  <note>
+   <para>
+    In order to delay program execution for a fraction of a second, use <function>usleep</function>
+    as the <function>sleep</function> will cast input to an integer. For example, sleep(0.25) will 
+    pause program execution for 0 seconds.
```
cmb69, line 23:
```suggestion
    pause program execution for <literal>0</literal> seconds.
```

cmb69 (commented):
Thank you for the PR!

That should be obvious from the function signature, but given the huge amount of upvotes on user note 54587, it makes sense to add it to the manual.

reference/misc/functions/sleep.xml
rxall: I was using the [defined](https://github.com/php/doc-en/blob/master/reference/misc/functions/defined.xml) manual page as a guide for formatting with <note> tags, have reverted back to <simpara> now

reference/misc/functions/sleep.xml
rxall: <simpara> within <note> too? feel free to edit to preference

reference/misc/functions/sleep.xml
cmb69: It makes perfect sense to check other manual pages, although it might make more sense to check the [skeletons](https://github.com/php/doc-base/tree/master/RFC/skeletons). However, these also use `<para>` inside `<note>`.

For me, the relevant difference between `<para>` and `<simpara>` is that `<simpara>` is mapped to HTML `<p>`, while `<para>` is mapped to `<div>`. These mappings could be changed, but actually they are quite suitable.

cmb69:
Thank you!

PR #1732, opened 2022-08-03
mention is_a() is also working on interface class-strings

staabm:
see https://3v4l.org/HmGbd

### DIFF
--- a/reference/classobj/functions/is-a.xml
+++ b/reference/classobj/functions/is-a.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="function.is-a" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>is_a</refname>
-  <refpurpose>Checks if the object is of this class/interface or has this class/interface as one of its parents</refpurpose>
+  <refpurpose>Checks if the object is of this type or has this type as one of its parents</refpurpose>
  </refnamediv>
  <refsect1 role="description">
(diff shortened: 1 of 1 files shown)


staabm:
thx for the feedback. what do you think about `class-like`?

cmb69:
> phpparser considers also a Trait as a ClassLike, which would be confusing

Yeah. Additionally, "ClassLike" is not really a noun, but rather an adjective.

Further suggestions welcome! Maybe there are some native speakers around?

staabm:
`class-string` just came to my mind

Crell:
In context, the correct term would be "type", or maybe "object type".  That non-basic type is synonymous with object in PHP is incidental.

staabm:
adjusted to 'type' per suggestion. thanks.

cmb69:
I don't think "type" cuts it – we have a lengthy [section about types in the manual](https://www.php.net/types), and most of these types can't be used with `is_a()`. And I've just seen that the respective parameter of `is_a()` is called `$class`, so maybe we should leave the `<refpurpose>` alone, but (verbosly) clarify the description of the function and its parameters?

Crell:
How about "object type", which limits it to typeable object thingies (classes, interfaces, enums, and if Records are ever added in the future probably them, too).

cmb69:
I went with "object type". Thank you all!

Issue #1733, opened 2022-08-03
DomDocument - same object id relates to different class


jifer:
### Description

The following code:

```php
<?php

final class DD extends \DOMDocument {
	public static function create() {
		$dom = new self();
		$root = $dom->createElement('rootelement');
		$dom->appendChild($root);

		$firstChild = $dom->firstChild;

		var_dump(
			spl_object_id($firstChild->ownerDocument), // int(1)
			get_class($firstChild->ownerDocument), // string(2) "DD" - this is correct
		);

		return $firstChild;
	}
	
	public function myMethod(){}
}


$dd = DD::create();
var_dump(
	spl_object_id($dd->ownerDocument), // int(1)
	get_class($dd->ownerDocument), // string(11) "DOMDocument" - this is incorrect
	// $dd->ownerDocument->myMethod() // This will rise exception as myMethod doesn not exist on \DomDocument
);
```

Resulted in this output:
```
int(1)
string(2) "DD"
int(1)
string(11) "DOMDocument"
```

But I expected this output instead:
```
int(1)
string(2) "DD"
int(1)
string(2) "DD"
```

Both var_dumps operates on same object (`dom->firstChild->ownerDocument`) but when out of class scope it returns main class (`DomDocument`) instead of extension class (`DD`).
It seems I can reproduce it only when creating object via static method.
https://onlinephp.io/c/5d173

### PHP Version

8.2.0beta1, 8.1.9, 7.4.30

jifer:
I create `DomDocument (X)`, next I'm adding `DomElement (A)` created using `DomDocument (X)` to that same `DomDocument (X)`. So it is obvieus that `DomElement (A)` is owned by `DomDocument (X)`. But due to some fancy internal mechanism `DomElement (A)` has completly new owner and it is some new, not known to me `DomDocument (Y)`.
Yeap... this sounds exactly like something expected.

KapitanOczywisty:
@jifer All `DOM*` classes are just a representation for the internal XML structure, if all references to any `DOM*` object are removed, then that object is simply destroyed and freed from memory. XML structure is always the source of truth.

Simpler example:
```php
<?php

class DOMElement2 extends DOMElement {
    public $test = 0;
}

$dom = new DomDocument();
$dom->registerNodeClass('DOMElement', 'DOMElement2');
$root = $dom->createElement('rootelement');
$dom->appendChild($root);
$root->test = 100;
[...]

PR #1735, opened 2022-08-04
Fix GH-1733: registerNodeClass can be used with DOMDocument

KapitanOczywisty:
`DOMDocument::registerNodeClass` with `DOMDocument` is possible and can be useful.

Fixes GH-1733

/cc @cmb69


cmb69:
Thank you!

PR #1736, opened 2022-08-04
Fix minor typo (the word "object" should be singular)

caendesilva:
In the current manual, the following is seen: "...if a very large objects...", where it should use the singular form. This PR fixes that so it becomes "...if a very large object...".


cmb69:
Thank you!

PR #1743, opened 2022-08-11
[PHP 8.2] Deprecate `utf8_*` functions

saundefined:


cmb69 (approved):
Thank you!

salathe:
It would be good to point readers at the alternatives / replacements.

cmb69:
> It would be good to point readers at the alternatives / replacements.

There are already respective links in the see-also sections, but I agree that it makes sense to make that more explicit, since these functions are likely used a lot. It seems to be sensible, to also document the alternatives in the migration guide; [UPGRADING currently doesn't mention these](https://github.com/php/php-src/blob/04d5faed23672fb1ba09320cedd0540938ab19a9/UPGRADING#L188).

PR #1744, opened 2022-08-12
Add filter_var() to is_numeric() see also section

8ctopus:
`filter_var` is a cool way to solve some of the [limitations](https://github.com/php/php-src/issues/9311) of `is_numeric` and therefore I suggest adding it in the see also section.


cmb69:
Makes sense to link it. Thank you!

PR #1745, opened 2022-08-12
Fix build

tiffany-taylor:
This should resolve the build. Tested locally.

### DIFF
--- a/reference/datetime/datetime/settime.xml
+++ b/reference/datetime/datetime/settime.xml
@@ -102,7 +102,7 @@
       <entry>The behaviour with double existing hours (during the fall-back
       DST transition) changed. Previously PHP would pick the second occurrence
       (after the DST transition), instead of the first occurrence (before DST
-      transition).
+       transition).</entry>
      </row>
      <row>
       <entry>7.1.0</entry>
--- a/reference/datetime/datetimeimmutable/construct.xml
+++ b/reference/datetime/datetimeimmutable/construct.xml
@@ -211,7 +211,6 @@ echo $time->format('Y/m/d H:i:s'), "\n";
 ]]>
    </screen>
   </example>
- </refsect1>
 
   <example>
    <title>Using a relative date/time string</title>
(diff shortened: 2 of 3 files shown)


reference/datetime/datetime/settime.xml
```diff
@@ -102,7 +102,7 @@
       <entry>The behaviour with double existing hours (during the fall-back
       DST transition) changed. Previously PHP would pick the second occurrence
       (after the DST transition), instead of the first occurrence (before DST
-      transition).
+       transition).</entry>
```
salathe, line 105: Nit: added an unnecessary space character before `transition`

reference/datetime/datetimeimmutable/settime.xml
```diff
@@ -81,7 +81,7 @@
       <entry>The behaviour with double existing hours (during the fall-back
       DST transition) changed. Previously PHP would pick the second occurrence
       (after the DST transition), instead of the first occurrence (before DST
-      transition).
+       transition).</entry>
```
salathe, line 84: Nit: again, added an unnecessary space character before `transition`

reference/datetime/datetimeimmutable/construct.xml
```diff
@@ -211,7 +211,6 @@ echo $time->format('Y/m/d H:i:s'), "\n";
 ]]>
    </screen>
   </example>
- </refsect1>
```
salathe, line 214: Good. :)

salathe (commented):
Thanks for fixing the build! 👍

reference/datetime/datetimeimmutable/settime.xml
tiffany-taylor: Should I push a new commit with these changes?

cmb69: Yes, please. If the changes are irrelevant for the translations (as in this case), please start the commit message with `[skip-revcheck]`.

PR #1746, opened 2022-08-12
Using time() in its own example

jawira:
`time()` function was not used in the example of its own page, so I fixed that.

### DIFF
--- a/reference/datetime/functions/time.xml
+++ b/reference/datetime/functions/time.xml
@@ -47,16 +47,19 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-echo 'Now:       '. date('Y-m-d') ."\n";
-echo 'Next Week: '. date('Y-m-d', strtotime('+1 week')) ."\n";
+$now = time();
+echo 'Timestamp: ' . $now . "\n";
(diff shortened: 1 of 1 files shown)


salathe:
See also #1747

cmb69:
Thank you for the PR! However, I don't see the point in using `date()` and `strtotime()` in this example at all (which is supposed to be about `time()`). Thus, I'm closing in favor of #1747.

jawira:
> Thank you for the PR! However, I don't see the point in using `date()` and `strtotime()` in this example at all (which is supposed to be about `time()`). Thus, I'm closing in favor of #1747.

FYI I didn't add those lines, just adapted it to the use with time().

cmb69:
> FYI I didn't add those lines, just adapted it to the use with time().

Yes, I'm aware of this. Still, the example didn't make sense in the first place. :)

jawira:
Got it! :) And what about adding `strtotime()` to the "See also" section? It might be useful because `strtotime()` can optionally take a timestamp as second argument. I will create a new PR if you agree.

cmb69:
> And what about adding `strtotime()` to the "See also" section?

Yeah, that makes some sense (although one might argue that a lot of other functions could be linked there as well).

PR #1747, opened 2022-08-12
Update document for time method

moeinpaki:
the example was related to `date` method and then I tried to fix it

### DIFF
--- a/reference/datetime/functions/time.xml
+++ b/reference/datetime/functions/time.xml
@@ -54,7 +54,7 @@ echo 'Now: '. time();
     &example.outputs.similar;
     <screen>
 <![CDATA[
-Now:       1660338149
+Now: 1660338149
 ]]>
     </screen>
    </example>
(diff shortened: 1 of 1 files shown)


reference/datetime/functions/time.xml
```diff
@@ -47,16 +47,14 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-echo 'Now:       '. date('Y-m-d') ."\n";
-echo 'Next Week: '. date('Y-m-d', strtotime('+1 week')) ."\n";
+echo 'Now:       '. time();
```
cmb69, line 50: I don't think the indentation makes sense now:
```suggestion
echo 'Now: '. time();
```
(would need to adapt the example output accordingly)

cmb69 (changes requested):
Thank you for the PR!

cmb69 (approved):
Thank you!

PR #1748, opened 2022-08-14
Namespaces/importing: update the list of structures which can be imported

jrfnl:
... for completeness.

Page link: https://www.php.net/manual/en/language.namespaces.importing.php

### DIFF
--- a/language/namespaces.xml
+++ b/language/namespaces.xml
@@ -608,7 +608,7 @@ $b = namespace\CONSTANT; // assigns value of constant CONSTANT to $b
    ability of unix-based filesystems to create symbolic links to a file or to a directory.
   </para>
   <para>
-   PHP can alias(/import) constants, functions, classes, interfaces, and namespaces.
+   PHP can alias(/import) constants, functions, classes, interfaces, traits, enums and namespaces.
   </para>
   <para>
    Aliasing is accomplished with the <literal>use</literal> operator.


cmb69:
Ah, thank you! That raises the point about a [general naming for these "entities"](https://github.com/php/doc-en/pull/1732#issuecomment-1203771553). Would you have a suggestion?

jrfnl:
> Ah, thank you! That raises the point about a [general naming for these "entities"](https://github.com/php/doc-en/pull/1732#issuecomment-1203771553). Would you have a suggestion?

Not really, I've been struggling with finding a generic name for these myself in other contexts. My current go-tos are "OO structures" or "classes and class-like structures".

Having said that, for the manual to be precise and spell things out is not necessarily a bad thing IMO>

cmb69:
> Having said that, for the manual to be precise and spell things out is not necessarily a bad thing IMO.

I agree that in this case it is fine. There are other cases where that is "too" long, though.

jrfnl:
> > Having said that, for the manual to be precise and spell things out is not necessarily a bad thing IMO.
> 
> I agree that in this case it is fine. There are other cases where that is "too" long, though.

Thanks and agreed. Will continue searching for an all-encompassing term for these structures, would be nice and useful terminology-wise to have such a term. Happy to brainstorm this out with others if more people are interested in solving this terminology conundrum.

PR #1749, opened 2022-08-15
formatObject accepts DateTimeInterface

BackEndTea:


cmb69:
Thanks for the PR! Although this is correct, it should first be fixed in the [stub file](https://github.com/php/php-src/blob/3b48a2044d5f578e2674c9f2e0b10f839fb43413/ext/intl/dateformat/dateformat.stub.php#L179); and we want to fix the [error message](https://github.com/php/php-src/blob/3b48a2044d5f578e2674c9f2e0b10f839fb43413/ext/intl/dateformat/dateformat_format_object.cpp#L184), too.

Do you want to file a respective PR against php-src?

BackEndTea:
👍 Will open a MR

PR #1755, opened 2022-08-18
DateTime: change suggestion to DateTimeImmutable

rigrig:
Using "the DateTimeInterface" instead of a DateTime object makes no sense, using a DateTimeImmutable instead of a DateTime does.


cmb69:
Thank you!

PR #1767, opened 2022-08-24
array_fill(examples): php8

duvall:
the result in php8 is different from the result in php7 for the negative keys

### DIFF
--- a/reference/array/functions/array-fill.xml
+++ b/reference/array/functions/array-fill.xml
@@ -114,9 +114,7 @@
 <![CDATA[
 <?php
 $a = array_fill(5, 6, 'banana');
-$b = array_fill(-2, 4, 'pear');
 print_r($a);
-print_r($b);
 ?>
 ]]>
     </programlisting>
@@ -132,6 +130,24 @@ Array
     [9]  => banana
     [10] => banana
 )
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example xml:id="function.array-fill.example.negative-start-index">
+    <title><function>array_fill</function> example with a negative start index</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$a = array_fill(-2, 4, 'pear');
+print_r($a);
+?>
+]]>
+    </programlisting>
+    &example.outputs.7;
+    <screen>
+<![CDATA[
 Array
 (
     [-2] => pear
@@ -141,18 +157,9 @@ Array
 )
 ]]>
     </screen>
-    In PHP 8
+    &example.outputs.8;
     <screen>
 <![CDATA[
-Array
-(
-    [5]  => banana
-    [6]  => banana
-    [7]  => banana
-    [8]  => banana
-    [9]  => banana
-    [10] => banana
-)
 Array
 (
     [-2] => pear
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-fill.xml
```diff
@@ -139,6 +139,27 @@ Array
     [1] => pear
     [2] => pear
 )
+]]>
+    </screen>
+    In PHP 8
```
cmb69, line 144: This should use the readily available XML entity:
```suggestion
    &example.outputs.8;
```
Accordingly, the `&example.outputs;` should be replaced with `&example.outputs.7`.

cmb69 (commented):
Ah, good catch. Thank you for the PR!

Besides the comment below, I suggest to split the example in two, since `array_fill(5, 6, 'banana')` has the same result in all versions. This also allows to briefly mention the difference between PHP 7 and 8 in the introducing paragraph of the second example.

reference/array/functions/array-fill.xml
```diff
@@ -132,17 +130,50 @@ Array
...
+    [-1] => pear
+    [0] => pear
+    [1] => pear
+)
 ]]>
     </screen>
    </example>
   </para>
+  <para>
+   Note that index '-1' is not present in php7 but is in php8.
```
cmb69, line 175: We prefer to be precise about the PHP version, so major.minor.revision should always be specified. We also favor the use of "as of PHP x.y.z" and "prior to PHP x.y.z", respectively. Also, literals should be marked up as such. However, I've just noticed that the description of the `$start_index` parameter only explains the PHP 7 behavior, and that should be fixed right away. And the link to the example needs to be adapted. And since that change to `array_fill` is rather considerable, there also should be a changelog entry.

reference/array/functions/array-fill.xml
```diff
@@ -37,8 +37,10 @@
        If <parameter>start_index</parameter> is negative, 
        the first index of the returned array will be 
        <parameter>start_index</parameter> and the following 
-       indices will start from zero 
-       (see <link linkend="function.array-fill.example.basic">example</link>).
+       indices will start from zero prior to PHP 8.0.0
+       (see <link linkend="function.array-fill.example.negative-start-index">example</link>).
+       As of PHP 8.0.0, negative keys are incremented normally
```
cmb69, line 42: It seems to me that a single link to the example is enough:
```suggestion
       indices will start from zero prior to PHP 8.0.0;
       as of PHP 8.0.0, negative keys are incremented normally
```

reference/array/functions/array-fill.xml
```diff
@@ -132,17 +132,50 @@ Array
...
+    [-1] => pear
+    [0] => pear
+    [1] => pear
+)
 ]]>
     </screen>
    </example>
   </para>
+  <para>
+   Note that index '-1' is not present prior to PHP 8.0.0.
```
cmb69, line 177: To avoid readers to think this is a string, I suggest:
```suggestion
   Note that index <literal>-1</literal> is not present prior to PHP 8.0.0.
```

cmb69 (commented):
Thank you! This looks good to me, except for the two minor issues commented below.

cmb69:
Thank you!

PR #1768, opened 2022-08-24
Add note about how bad weak mode is for in_array.

Crell:
More than 3/4 of the comments on in_array are/were (I deleted them) variations on "beware weak mode, it does weird things."

So let's kill them all and add a note that says the same thing so people stop posting it over and over again.


cmb69 (commented):
Thank you! Makes perfect sense to clarify that; as for me, this note could be moved to the respective parameter description.

Crell:
Makes sense to me.  Gives it higher billing.

cmb69:
Thank you!

PR #1769, opened 2022-08-24
Note the recursion checking for magic methods.

Crell:
A whole bunch of comments are just calling out this fact, so let's call it out ourselves.  (Comments have been deleted.)

### DIFF
--- a/language/oop5/overloading.xml
+++ b/language/oop5/overloading.xml
@@ -112,6 +112,18 @@
     </para>
    </note>
 
+   <note>
+    <para>
+     PHP will not call an overloaded method from within the same overloaded method.
+     That means, for example, writing <code>return $this->foo</code> inside of
+     <link linkend="object.get">__get()</link> will return <literal>null</literal>
+     and raise a Warning if there is no <literal>foo</literal> property defined,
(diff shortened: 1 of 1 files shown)


language/oop5/overloading.xml
```diff
@@ -112,6 +112,18 @@
     </para>
    </note>
 
+   <note>
+    <para>
+     PHP will not call an overloaded method from within the same overloaded method.
+     That means, for example, writing <code>return $this->foo</code> inside of
+     <link linkend="object.get">__get()</link> will return <literal>null</literal>
+     and raise a Warning if there is no <literal>foo</literal> property defined,
```
cmb69, line 120: "Warning" should either be lower-cased, or `<constant>E_WARNING</constant>` should be used (or the verbose `raises an error of level <constant>E_WARNING</constant>`). I'm not sure whether `<literal>foo</literal>` or `<literal>$foo</literal>` is more appropriate.

Crell: Warning fixed.  No idea on foo, so left it for now.  I have no strong preference other than consistency.

cmb69:
Thank you!

PR #1771, opened 2022-08-25
Add "die()" as an option to exit a function with return typ "never"

ShadyMedic:
There was only the `exit()` function mentioned.

### DIFF
--- a/language/types/declarations.xml
+++ b/language/types/declarations.xml
@@ -604,8 +604,8 @@ function &test(): void {}
    <title>never</title>
    <para>
     <literal>never</literal> is a return type indicating the function does not
-    return. This means that it either calls <function>exit</function>, throws
-    an exception, or is an infinite loop.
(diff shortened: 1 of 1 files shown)


ShadyMedic:
Yeah, I know it's a function/construct with the same output, but at the same time, it's also another way to stop the execution of the script. I think it should be listed to not raise doubts about whether are these two functions really the same.

Girgias:
I'm not really a fan of this, ``die`` is akin to an alias of ``exit``.

kamil-tekiela:
That's unnecessary. If you really insist then something like `exit/die` would look better, but since it's the exact same language construct, I don't think it's needed at all.

afilina:
Since there is no interest in listing the alias, should this PR be closed?

cmb69:
Yeah, this PR should be closed. Nonetheless, thanks for working on this, @ShadyMedic!

PR #1772, opened 2022-08-25
Update isrepeated.xml

bobmagicii:
fix the wording on this method. the method name is past tense, the description is future tense. after testing this method, only the past tense is valid. it does not return true if the attribute is flagged repeatable, only if it really has been repeated.


cmb69:
Thank you!

PR #1774, opened 2022-08-26
SoapClient::__construct - Fix method names referenced in trace option

j-applese3d:
`SoapClient::getLastRequest` does not exist. 
The correct name is `SoapClient::__getLastRequest`

Same issue with `getLastRequestHeaders`, `getLastResponse`, and `getLastResponseHeaders`


cmb69:
Thank you!

PR #1776, opened 2022-08-29
magic methods: improve __invoke documentation

houssemz:
I find that the documentation of magic methods is poor of practicals examples. In this PR I try do improve the documentation of `__invoke` method. Hope that will be accepted.

### DIFF
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -364,6 +364,96 @@ bool(true)
 ]]>
     </screen>
    </example>
+   <example>
+    <title>Using <link linkend="object.invoke">__invoke()</link></title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+class Sort
+{
+    private $key;
+
+    public function __construct(string $key)
+    {
+        $this->key = $key;
+    }
+
+    public function __invoke(array $a, array $b): int
+    {
+        return $a[$this->key] <=> $b[$this->key];
+    }
+}
+
+$customers = [
+    ['id' => 1, 'first_name' => 'John', 'last_name' => 'Do'],
+    ['id' => 3, 'first_name' => 'Alice', 'last_name' => 'Gustav'],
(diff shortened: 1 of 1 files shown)


houssemz:
@Girgias need a review, please :) thank you 🙏

cmb69:
Thank you for the PR! I wonder whether this example would make more sense on the `usort()` page; there could be a link from the `__invoke()` documentation.

houssemz:
@cmb69 yes ^^! it is related to both `__invoke ` and `usort `. But, I think it should be here (in the `__invoke` method documentation) because I find this one really poor and could cause confusion. When you arrive for the first time on PHP and you try to understand why we use `__invoke` you do not found a "real-world" example. Hope that is was convincing and clear :)

houssemz:
Ping @cmb69 :)

Crell:
I'm torn here.

1. Yes, we should have examples for the magic methods.
2. However, the current magic method page is stupidly huge.  It may be beneficial to split it up into several pages.
3. Adding long examples to the current page kinda just makes it worse.
4. But at the same time, this particular example does kinda make sense on the `usort` page instead... but that's not going to work for everything.  So we probably should *not* go down that route.  Whether that example is useful on the `usort` page on its own merits is a separate question.
5. I'm not sure if this is the best example for `__invoke()`, since I've never seen it used this way.  It certainly could be, though.

So... maybe we can commit this, and i a follow up see about spitting up the page?

cmb69:
> 4\. Whether that example is useful on the `usort` page on its own merits is a separate question.

I think it is fitting there, because it shows how to use a configurable callable to do the comparisons.

> So... maybe we can commit this, and i a follow up see about spitting up the page?

This is also fine for me.

houssemz:
IMHO, I am okay also with that we should separate this page because it is really huge (I will make some initiatives). 

I think also, that documentation should avoid us to search for examples somewhere else that's why I added what I called a real "word example".

But still, are you ok with this modification? Should I do any modification?

PR #1781, opened 2022-08-30
DateTimeZone::listIdentifiers: fix incorrect param reference

jrfnl:
Refs:
* https://www.php.net/manual/en/datetimezone.listidentifiers.php
* https://3v4l.org/iBYXH


cmb69:
Thank you!

jrfnl:
You're welcome. Ran into trouble trying to use it based on the docs, so figured I better fix it ;-)

Issue #1780, opened 2022-08-30
inconsistency in enum documentation


dlehmann:
there is an inconsistency in the enum examples here

https://wiki.php.net/rfc/enumerations

This enum:

```php
enum SortOrder {
 case ASC;
 case DESC;
}
```
should also be PascalCase:

```php
enum SortOrder {
 case Asc;
 case Desc;
}
```

cmb69:
Yeah, but I don't think it makes sense to "fix" this in the RFC. Or do you think otherwise, @Crell?

salathe:
The same `enum` is used in the manual as well: https://www.php.net/manual/en/language.enumerations.examples.php

Crell:
I'll fix the docs page.  I don't know if it's appropriate to update the RFC at this point, but if someone else did so I wouldn't raise a fuss.

Girgias:
Fixed by #1785

PR #1785, opened 2022-08-31
Standardize casing for Enum examples.

Crell:
Per #1780, if we're telling people to use CamelCase for enum cases, the docs should do so consistently.


cmb69:
Thank you!

PR #1789, opened 2022-09-02
Fix missing object type

sy-records:
Is this missing?  2761e65ea261c33f65a7854b567a5821cf06aaa7

### DIFF
--- a/reference/classobj/functions/is-a.xml
+++ b/reference/classobj/functions/is-a.xml
@@ -34,7 +34,7 @@
      <term><parameter>class</parameter></term>
      <listitem>
       <para>
-       The class name
+       The class or interface name
       </para>
      </listitem>
     </varlistentry>
@@ -53,7 +53,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true; if the object is of this class or has this class as one of
+   Returns &true; if the object is of this object type or has this object type as one of
    its parents, &false; otherwise.
   </para>
  </refsect1>
--- a/reference/classobj/functions/is-a.xml
+++ b/reference/classobj/functions/is-a.xml
@@ -15,7 +15,7 @@
   </methodsynopsis>
   <para>
    Checks if the given <parameter>object_or_class</parameter> is of this object type or has
-   this class as one of its parents.
+   this object type as one of its parents.
   </para>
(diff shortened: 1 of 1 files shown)


reference/classobj/functions/is-a.xml
```diff
@@ -15,7 +15,7 @@
   </methodsynopsis>
   <para>
    Checks if the given <parameter>object_or_class</parameter> is of this object type or has
-   this class as one of its parents.
+   this object type as one of its parents.
```
cmb69, line 18: I think when we speak of types, referring to a "parent" is uncommon; instead we probably want to call it "supertype".

Note that we should fix the return-value section as well. And actually, some other manpages, too, e.g. `is_subclass_of()`, although in that case the [name of the function is misleading](https://3v4l.org/dupND). :(

mumumu:
If we accept this PR, should also fix return type description. See also #1796.

sy-records:
Thank you! @mumumu

reference/classobj/functions/is-a.xml
sy-records: Do I need to change it to `supertype`?

Elimuhub-coder (commented):
change location to Nairobi/Kenya[location](https://maps.google.com/maps/contrib/117273694586709139832) 
@sy-records

reference/classobj/functions/is-a.xml
afilina: @cmb69 Could you suggest a change that you had in mind so that it's clearer?

cmb69, line 18: I'd change
````diff
- this object type as one of its parents.
+ this object type as one of its supertypes.
````

File: reference/classobj/functions/is-a.xml [hunk 2]
```diff
@@ -53,7 +53,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true; if the object is of this class or has this class as one of
+   Returns &true; if the object is of this object type or has this object type as one of
    its parents, &false; otherwise.
```
cmb69, line 57:
````diff
- its parents, &false; otherwise.
+ its supertypes, &false; otherwise.
````

sy-records:
Resubmitted a PR #2238

Issue #1793, opened 2022-09-02
Documentation does not apply to PHP 5


afilina:
From manual page: https://php.net/exception.construct

---

`__construct(string $message = "", int $code = 0, ?Throwable $previous = null)`
The docs say that it applies to PHP 5, 7, 8, but `Throwable` only exists in PHP >=7.0

According to [legacy docs](https://php-legacy-docs.zend.com/manual/php5/en/language.exceptions.extending), PHP 5's constructor would look like this:
`__construct($message = null, $code = 0, Exception $previous = null)`

Since we're no longer maintaining PHP 5 docs, should we _not_ mention PHP 5 on this page, or should we add a note that this docs page doesn't apply to PHP 5, with perhaps a link to legacy docs?

cmb69:
The info `(PHP 5, PHP 7, PHP 8)` only refers to the PHP versions where that method is available. It does not refer to any of the "details", such as the function signature, or any behavior documented.

That said, I don't think that the availability info for old versions is particularly useful, and as such might be removed. Or we could possibly clarify, that it is only about the documented "entity" per se. Or, maybe, it is sufficient to overhaul https://www.php.net/manual/en/about.prototypes.php, which is in desperate need of an update anyway; although that page might be removed after an update (not much to say there since we have scalar types etc.)

afilina:
> I don't think that the availability info for old versions is particularly useful
If you mean that we can _not_ mention PHP 5, that could be a solution, but might be confusing. It can be understood as "this feature did not exist until PHP 7."

> Or, maybe, it is sufficient to overhaul https://www.php.net/manual/en/about.prototypes.php
After 20 years with PHP, I never knew that this about page existed, so it doesn't quite feel like a solution.

Let me open a PR with suggested clarification, and we can debate it with something more tangible :)

PR #1794, opened 2022-09-02
Clarify Exception constructor for PHP 5

afilina:
See: https://github.com/php/doc-en/issues/1793

Draft for clarifying why the constructor signature doesn't work for PHP 5, and why we're not documenting older versions.

### DIFF
--- a/language/predefined/exception/construct.xml
+++ b/language/predefined/exception/construct.xml
@@ -17,6 +17,13 @@
   <para>
    Constructs the Exception.
   </para>
+  
+  <note>
+   <simpara>
+    This constructor signature applies only to PHP 7.0 and higher. 
+    The documentation for unsupported PHP versions has been discontinued.
+   </simpara>
+  </note>
  </refsect1>
  
  <refsect1 role="parameters">
--- a/language/predefined/exception/construct.xml
+++ b/language/predefined/exception/construct.xml
@@ -20,7 +20,7 @@
   
   <note>
    <simpara>
-    This constructor signature applies only to PHP 7.0 and higher. 
+    This constructor signature applies only to PHP 7.0 and higher.
     The documentation for unsupported PHP versions has been discontinued.
    </simpara>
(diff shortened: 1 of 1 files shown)


language/predefined/exception/construct.xml
```diff
@@ -17,6 +17,13 @@
   <para>
    Constructs the Exception.
   </para>
+  
+  <note>
+   <simpara>
+    This constructor signature applies only to PHP 7.0 and higher.
```
afilina, line 23:
```suggestion
    This constructor signature applies only to PHP 7.0 and higher.
```

afilina:
I give up on the whitespace failure. There is no trailing whitespace.

cmb69:
> There is no trailing whitespace.

Well, there is in the "empty" line before the new `<note>`. Anyway, that is not important for now.

Generally, I'm not sure whether we want to add such a note for all affected functions/methods (there are certainly more than only this constructor; adding optional parameters is not uncommon). But if we do, we likely want to have an XML entity for that (or maybe a set of entities, to cater to different versions).

salathe:
> This constructor signature applies only to PHP 7.0 and higher.

Redundant. The PHP Manual, particularly any "prototypes" and associated descriptions, only applies (at the time of writing this) to PHP 7.0.0 and higher.  Any vestiges in the descriptions, examples, etc. are just yet to be tidied up: the last remaining widespread reference to undocumented versions is the "versions list" at the top of many pages where "PHP 5" is still present.

> The documentation for unsupported PHP versions has been discontinued.

How, in any way, would this be considered reasonable to add, particularly in the `Exception` constructor description.  (Note: rhetorical question.)  Even if it were being added in a reasonable place, it's factually incorrect. The PHP Manual does document unsupported PHP versions, just not all of them, and not forever.

Rejecting this Pull Request as unsuitable.  Discussion can, and probably should, continue in Issue #1793.

PR #1800, opened 2022-09-06
Use the alias entity everywhere

kocsismate:


salathe:
> Use the alias entity everywhere

Missed a few in mysqli and xdiff.

PR #1801, opened 2022-09-07
Use even more &alias; entities

kocsismate:
So that https://github.com/php/php-src/pull/9491 can find all the aliases.

### DIFF
--- a/reference/mysqli/functions/mysqli-execute.xml
+++ b/reference/mysqli/functions/mysqli-execute.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="function.mysqli-execute" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>mysqli_execute</refname>
-  <refpurpose>Alias for <function>mysqli_stmt_execute</function></refpurpose>
+  <refpurpose>&Alias; <function>mysqli_stmt_execute</function></refpurpose>
  </refnamediv>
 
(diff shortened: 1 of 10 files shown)


salathe:
A few more to do:
* reference/solr/solrdocument/reset.xml:  `<refpurpose>This is an alias to SolrDocument::clear()</refpurpose>`
* reference/solr/solrparams/add.xml:  `<refpurpose>This is an alias for SolrParams::addParam</refpurpose>`
* reference/solr/solrparams/get.xml:  `<refpurpose>This is an alias for SolrParams::getParam</refpurpose>`

salathe (approved):

kocsismate:
> Note there are also &info.function.alias; and &info.method.alias; which are used for basically the same purpose. I guess we need to cater to these as well; maybe just replace them with &Alias;?

Yeah, I saw them, but I didn't touch them because they were luckily not used in the `<refpurpose>` tag. As these entities use a slightly different wording than `&Alias;`, I decided not to complicate my PR.

cmb69:
> I didn't touch them because they were luckily not used in the `<refpurpose>` tag

Indeed. However, there are some case where they are used in the description, but there is no `&Alias;` in the refpurpuse, e.g. https://github.com/php/doc-en/blob/master/reference/curl/functions/curl-file-create.xml. I assume that needs to fixed.

PR #1806, opened 2022-09-08
Sync alias description with the refpurpose

kocsismate:
These were the only issues I could find.


reference/curl/functions/curl-file-create.xml
```diff
@@ -10,12 +9,11 @@
  <refsect1 role="description">
   &reftitle.description;
   <para>
-   &info.function.alias; <methodname>CURLFile::__construct</methodname>
```
kocsismate, line 13: `curl_file_create()` is not an alias of `CURLFile::__construct()` (even though their implementation basically matches), that's why I tried to find a different description, otherwise https://github.com/php/php-src/pull/9510 will display this as a false positive warning.

kocsismate: I'd appreciate help about the wording as I'm not satisfied with the one I came up with.

cmb69: Ah, right, not a proper alias. I think it would make sense to merge the pages, though, like it is done with https://www.php.net/manual/en/numberformatter.create.php and some others. To not break existing links to the `curl_file_create()` page, we could update https://github.com/php/web-php/blob/master/include/errors.inc.

kocsismate: Filed https://github.com/php/web-php/pull/696

Issue #1803, opened 2022-09-07
PHP 8.2 documentation tracker


Girgias:
Pages which need to be added/modified for complete documentation of PHP 8.2:

Based of the migration guide: https://www.php.net/manual/en/migration82.php (https://github.com/php/doc-en/pull/1799)

> **Note**
> This issue is still a Work In Progress

## Core

- [x] New features:
   - [x] Type system (#1726) 
     - [x] false null standalone 
     - [x] true
     - [x] DNF types
   - [x] Sensitive Parameter attribute
   - [x] Readonly classes (#1996)
   - [x] error_log_mode INI directive (Part of #1799)
   - [x] Declaring constants in traits (#1995)
- [x] Deprecated
  - [x] Dynamic properties (#1942)
  - [x] "${var}" and "${expr}" style string interpolations (#2000)

## Classes:

- [x] New classes
  - [x] Random Extension (based on [random.stub.php](https://github.com/php/php-src/blob/master/ext/random/random.stub.php) https://github.com/php/doc-en/pull/1916 )
    - [x] Random\Engine
      - [x] Random\Engine\Mt19937
      - [x] Random\Engine\PcgOneseq128XslRr64
      - [x] Random\Engine\Xoshiro256StarStar
      - [x] Random\Engine\Secure
    - [x] Random\CryptoSafeEngine
    - [x] Random\Randomizer
    - [x] Random\RandomError
    - [x] Random\BrokenRandomEngineError
    - [x] Random\RandomException

## Functions/Methods:
- [x] New functions:
  - [x] cURL
    - [x] `curl_upkeep()` (#1807)
  - [x] IMAP
    - [x] `imap_is_open()` (#2583) 
  - [x] MySQLi
    - [x] `mysqli_execute_query()` (#1855)
  - [x] OpenSSL
    - [x] `openssl_cipher_key_length()` (#1808)
  - [x] Reflection
    - [x] `ReflectionFunction::isAnonymous()` (#1742)
    - [x] `ReflectionMethod::hasPrototype()` (#1742)
  - [x] Sodium
    - [x] `sodium_crypto_stream_xchacha20_xor_ic()` (#1817)
  - [x] Standard
    - [x] `memory_reset_peak_usage()` (#1809)
    - [x] `ini_parse_quantity()` (#1810)
  - [x] XML
    - [x] `libxml_get_external_entity_loader()` (#1933)
  - [x] ZIP
     - [x] `ZipArchive::clearError()` (f57925e)
[...]

PR #1811, opened 2022-09-09
add `FILTER_FLAG_GLOBAL_RANGE` documentation

realFlowControl:
Hey there :vulcan_salute: 

this will add documentation for `FILTER_FLAG_GLOBAL_RANGE` as found in #1803 and added to PHP 8.2 with https://github.com/php/php-src/commit/d8fc05c05eb6fe189a663c599421a8457bc6c767
I was thinking to write down all global IP ranges (as you can see in the docs for `FILTER_FLAG_NO_PRIV_RANGE`), but it turns out that these are quite a few. So I thought it might be sufficient to link to the RFC where those global ranges are defined.

WDYT?

Kind regards
Florian

### DIFF
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -148,6 +148,7 @@
          <constant>FILTER_FLAG_IPV6</constant>,
          <constant>FILTER_FLAG_NO_PRIV_RANGE</constant>,
          <constant>FILTER_FLAG_NO_RES_RANGE</constant>,
+         <constant>FILTER_FLAG_GLOBAL_RANGE</constant>,
          <constant>FILTER_NULL_ON_FAILURE</constant>
         </entry>
         <entry>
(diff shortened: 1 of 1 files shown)


saundefined:
Thanks! Perhaps, the changelog entry should be added 🤔

realFlowControl:
Changelog entry for `FILTER_FLAG_GLOBAL_RANGE`  :heavy_check_mark:

reference/filter/filters.xml
```diff
@@ -739,6 +740,19 @@ filter.default_flags = 0
...
+      <row>
+       <entry><constant>FILTER_FLAG_GLOBAL_RANGE</constant></entry>
+       <entry>
+        <constant>FILTER_VALIDATE_IP</constant>
+       </entry>
+       <entry>
+        <para>
+         Fails validation for non global IPv4/IPv6 ranges as found in
+         <link xlink:href="&url.rfc;6890">RFC 6890</link> with the
+         Global attribute being False.
```
cmb69, line 752:
```suggestion
         <literal>Global</literal> attribute being <literal>False</literal>.
```

cmb69 (commented):
Thank you! Besides the markup (see comment below), this looks good to me.

cmb69 (approved):

reference/filter/filters.xml
```diff
@@ -739,6 +740,19 @@ filter.default_flags = 0
...
+      <row>
+       <entry><constant>FILTER_FLAG_GLOBAL_RANGE</constant></entry>
+       <entry>
+        <constant>FILTER_VALIDATE_IP</constant>
+       </entry>
+       <entry>
+        <para>
+         Fails validation for non global IPv4/IPv6 ranges as found in
+         <link xlink:href="&url.rfc;6890">RFC 6890</link> with the
+         <literal>Global</literal> attribute being <literal>False</literal>.
```
Girgias, line 752:
```suggestion
         <literal>Global</literal> attribute being &false;.
```
Am I missing some context or shouldn't this use the usual entity @cmb69 ?

cmb69: This refers to RFC 6890, where there are `Global` attributes of type boolean, but these are not really PHP's booleans. So I'd leave this as is.

PR #1816, opened 2022-09-10
Simplify example in `Stringable` docs.

ZebulanStanphill:
Props to Gormack for the idea: https://www.php.net/manual/en/class.stringable.php#126627

The interface was introduced in PHP 8, and you know what else was introduced in that version? Promoted constructor properties. By reducing the boilerplate, the example becomes a lot more focused.

### DIFF
--- a/language/predefined/stringable.xml
+++ b/language/predefined/stringable.xml
@@ -90,6 +90,9 @@ showStuff($ip);
      </screen>
     </example>
    </para>
+   <note>
+    <simpara>The above example uses <link linkend="language.oop5.decon.constructor.promotion">constructor property promotion</link>.</simpara>
+   </note>
   </section>
 
  </partintro>
--- a/language/predefined/stringable.xml
+++ b/language/predefined/stringable.xml
@@ -58,17 +58,12 @@
 <![CDATA[
 <?php
 class IPv4Address implements Stringable {
-    private string $oct1;
-    private string $oct2;
-    private string $oct3;
-    private string $oct4;
-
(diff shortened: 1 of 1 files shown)


devnexen (approved):

cmb69:
I'm not convinced that this change is a good idea. People just upgrading to/learning about PHP 8.0 might be confused by the constructor property promotion when reading about `Stringable`. At the very least, we should clarify that this is new syntax in PHP 8.0, and link to the respective section in the manual.

ZebulanStanphill:
@cmb69 I added a note below the example referencing the docs for constructor property promotion. I agree that it's good to link to the documentation about the feature. One of the main goals of documentation is to teach, after all. And constructor property promotion is (at least in my opinion, and surely that of the majority who voted for it to be added) a best practice that we should be encouraging the usage of.

cmb69:
Thank you! This is okay, but I would have placed the description above the example, to [conform to the template](https://github.com/php/doc-base/blob/231cc1199578e1a06673ed09420e1dc49c5c3235/RFC/skeletons/function.xml#L88-L91), and because it is even more helpful to PHP 8 novices. I wouldn't mark it up as `<note>` in that case. What do you think?

afilina:
@ZebulanStanphill There is one last change requested by @cmb69 in the previous comment.

cmb69 (approved):
Thank you! (To explicitly mention property promotion might not be that important anymore, but still a good idea.)

Girgias (approved):

PR #1821, opened 2022-09-12
[PHP 8.2] Update `getimagesize` description

saundefined:
Ref:

https://github.com/php/php-src/blob/PHP-8.2/UPGRADING#L303

### DIFF
--- a/reference/image/functions/getimagesize.xml
+++ b/reference/image/functions/getimagesize.xml
@@ -192,6 +192,14 @@ if ($size && $fp) {
      </thead>
      <tbody>
       <row>
+       <entry>8.2.0</entry>
+       <entry>
+        Now returns the actual image dimensions, bits and channels of AVIF images;
+        previously, the dimensions were reported as <literal>0x0</literal>,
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you!

reference/image/functions/getimagesize.xml
```diff
@@ -191,6 +191,14 @@ if ($size && $fp) {
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.2.0</entry>
+       <entry>
+        Now returns the actual image dimensions, bits and channels of AVIF images;
+        previously, the dimensions have been reported as <literal>0x0</literal>,
+        and bits and channels have not been reported at all.
```
Girgias, line 199:
```suggestion
        previously, the dimensions were reported as <literal>0x0</literal>,
        and bits and channels were not reported at all.
```
Reads a lot better IMHO

saundefined: Thanks!

PR #1828, opened 2022-09-16
fix the describe of converting float to int

mowangjuanzi:


language/types/integer.xml
```diff
@@ -212,7 +212,7 @@ var_dump(round(25/7));  // float(4)
 
    <simpara>
     When converting from <type>float</type> to <type>int</type>, the number
-    will be rounded <emphasis>towards zero</emphasis>.
+    will be rounded <emphasis>fractions down</emphasis>.
```
cmb69, line 215: "Towards zero" is more correct, see e.g. https://3v4l.org/DYGFl.

mowangjuanzi: no. This sentence, what I understand is `(int)0.9 = 1`, but it is actually `(int)0.9 = 0`

cmb69: Yeah, right, and that is what "round towards zero" means; it rounds down positive numbers, and it rounds up negative numbers, see https://3v4l.org/4Lnjk. We could also say "truncate the decimal places", I think.

mowangjuanzi: I get it, but it does mislead me, probably because of my unfamiliarity with English.

cmb69: I'm not a native English speaker either; maybe somebody else has suggestions on how to improve this.

afilina: This is basically `floor()` so I looked at how it is described there: https://www.php.net/manual/en/function.floor.php
> floor — Round fractions down

This means that @mowangjuanzi is correct, or at least consistent with the rest of the manual. I'd say we can merge.

cmb69: Not really; `floor()` behaves differenly for negative numbers, see https://3v4l.org/eYFWv. So `floor()` indeed rounds down, but `(int)` rounds towards zero. I suggest to keep the documentation as is.

afilina: Should this PR be closed then?

cmb69: Yeah, makes not much sense to keep it open. :)

PR #1835, opened 2022-09-19
the default argument is nullable

divinity76:
prototype is `?string $persistent_id=null`   - UNTESTED, i don't know if `<type>?string</type>` is actually valid or not.. someone should double-check that before merging

### DIFF
--- a/reference/memcached/memcached/construct.xml
+++ b/reference/memcached/memcached/construct.xml
@@ -11,7 +11,7 @@
   &reftitle.description;
   <constructorsynopsis>
    <modifier>public</modifier> <methodname>Memcached::__construct</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>persistent_id</parameter></methodparam>
+   <methodparam choice="opt"><type>?string</type><parameter>persistent_id</parameter><initializer>null</initializer></methodparam>
   </constructorsynopsis>
   <para>
    Creates a Memcached instance representing the connection to the memcache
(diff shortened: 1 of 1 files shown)


reference/memcached/memcached/construct.xml
```diff
@@ -11,7 +11,7 @@
   &reftitle.description;
   <constructorsynopsis>
    <modifier>public</modifier> <methodname>Memcached::__construct</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>persistent_id</parameter></methodparam>
+   <methodparam choice="opt"><type>?string</type><parameter>persistent_id</parameter><initializer>null</initializer></methodparam>
```
Girgias, line 14:
```suggestion
   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>persistent_id</parameter><initializer>null</initializer></methodparam>
```
Is how a union type is marked.

Girgias (changes requested):

reference/memcached/memcached/construct.xml
divinity76: thanks, committed to PR

Girgias (approved):
According to https://github.com/php-memcached-dev/php-memcached/blob/master/php_memcached.stub.php this seems to be correct, although the rest of the parameters are undocumented.

Might make sense to run the methodsynopsis generator on this stub file.

cmb69:
> Might make sense to run the methodsynopsis generator on this stub file.

Yes, that should be done for all PECL extensions which are documented in the PHP manual, and which have stub files. Quite some work, though.

PR #1837, opened 2022-09-20
Update filters.xml

FuZuK:
Fix the name of filter from FILTER_SANITIZE_RAW to FILTER_UNSAFE_RAW
No such filter with name FILTER_SANITIZE_RAW


cmb69:
Thank you!

PR #1839, opened 2022-09-21
[PHP 8.1]  Calling mb_check_encoding() without argument is deprecated

DuckThom:
Passing null or omitting the arguments is deprecated as of this RFC https://wiki.php.net/rfc/deprecations_php_8_1

### DIFF
--- a/reference/mbstring/functions/mb-check-encoding.xml
+++ b/reference/mbstring/functions/mb-check-encoding.xml
@@ -37,8 +37,7 @@
 
       <warning>
         <para>
-          As of PHP 8.1.0, omitting this parameter or passing NULL is deprecated and
-          emits an <constant>E_DEPRECATED</constant> warning.
+          As of PHP 8.1.0, omitting this parameter or passing &null; is deprecated.
         </para>
       </warning>
      </listitem>
@@ -76,7 +75,7 @@
      <row>
       <entry>8.1.0</entry>
       <entry>
-       Calling this function with null as <parameter>value</parameter> or without argument is deprecated.
+       Calling this function with &null; as <parameter>value</parameter> or without argument is deprecated.
       </entry>
      </row>
      <row>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you for the PR! I suggest to *also* add that info to the `$value` parameter description.

DuckThom:
@cmb69 Good suggestion, added!

image

reference/mbstring/functions/mb-check-encoding.xml
```diff
@@ -34,6 +34,13 @@
         The byte stream or &array; to check. If it is omitted, this function checks
         all the input from the beginning of the request.
       </para>
+
+      <warning>
+        <para>
+          As of PHP 8.1.0, omitting this parameter or passing NULL is deprecated and
```
cmb69, line 40: Please use the predefined entity here:
```suggestion
          As of PHP 8.1.0, omitting this parameter or passing &null; is deprecated and
```

reference/mbstring/functions/mb-check-encoding.xml
```diff
@@ -66,6 +73,12 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.1.0</entry>
+      <entry>
+       Calling this function with null as <parameter>value</parameter> or without argument is deprecated.
```
cmb69, line 79:
```suggestion
       Calling this function with &null; as <parameter>value</parameter> or without argument is deprecated.
```

reference/mbstring/functions/mb-check-encoding.xml
```diff
@@ -34,6 +34,13 @@
         The byte stream or &array; to check. If it is omitted, this function checks
         all the input from the beginning of the request.
       </para>
+
+      <warning>
+        <para>
+          As of PHP 8.1.0, omitting this parameter or passing NULL is deprecated and
+          emits an <constant>E_DEPRECATED</constant> warning.
```
cmb69, line 41: In my opinion, this is superfluous; the fact that it's deprecated should imply `E_DEPRECATED`; otherwise, we'd use the word "discouraged" (or maybe "soft-deprecated").

cmb69:
Thank you!

PR #1840, opened 2022-09-21
Fix one 5+ digit year example (sign is required)

guilliamxavier:
But actually neither seems to work currently? https://3v4l.org/9fKt3/rfc#vgit.master :confused:


cmb69:
If the sign is required, the cell before would need to be fixed as well (`[]` means that something is optional). I don't think the sign is required, though.

Anyhow, I also wonder why https://3v4l.org/d0tmu/rfc#vgit.master fails. @derickr, can you please clarify?

derickr:
This sign is required.

derickr:
And FWIW, `[]` does not mean "optional". It means a choice. Adding `?` makes it optional.

PR #1843, opened 2022-09-23
Improve rowCount explanation

kamil-tekiela:
This puts emphasis on the important part. Some have claimed that rowCount will return inconsistent, untrue or broken results. That's not true.


cmb69 (approved):
Thank you!

PR #1845, opened 2022-09-26
Improve `is_writable` doc

PululuK:


cmb69:
Thanks for the PR! However, I don't think this is a real improvement. After all, `$filename` is the name of the parameter, and further sections below clarify that. Also, `is_writable()` works on stream wrappers, which may not even refer to files or directories. And finally, several other (all?) refpurpose sections of related functions also mention "file" only.

PR #1848, opened 2022-09-27
Remove references to "double" and use float instead

Girgias:


cmb69 (approved):
Nice, thank you!

cmb69:
I think the whitespace check still fails because the commits are not squashed yet.

PR #1851, opened 2022-09-28
Fix error on ldap_get_option page

Deni42:
LDAP_OPT_DIAGNOSTIC_MESSAGE is a string


cmb69:
[Indeed](https://linux.die.net/man/3/ldap_get_option). Thank you!

PR #1852, opened 2022-09-28
Update the example in Stringable interface

miniyarov:
The example in Stringable interface is updated to use the constructor promotion and readonly properties.

### DIFF
--- a/language/predefined/stringable.xml
+++ b/language/predefined/stringable.xml
@@ -58,16 +58,12 @@
 <![CDATA[
 <?php
 class IPv4Address implements Stringable {
-    private string $oct1;
-    private string $oct2;
-    private string $oct3;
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you for the PR! However, I'm not sure that this is a good idea. `Stringable` and `constructor property promotion` are only available as of PHP 8.0.0, so anybody new to that version and looking up `Stringable` might be confused about that other new syntax. At the very least, there should be a brief hint about that (including a link to https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion).

miniyarov:
Closing this because this is a duplicate of https://github.com/php/doc-en/pull/1816

cmb69:
Somehow that rang a bell. :)

PR #1853, opened 2022-09-29
Ease of reading

kamil-tekiela:


devnexen (approved):
looks ok even tough the original form is comprehensible enough (albeit the second therefore can be ditched).

PR #1854, opened 2022-09-29
Better wording for generators

kamil-tekiela:
Follow up from #1285. 

I welcome any input. I know this can be worded better, but I also agree with @nikic that the current wording isn't ambiguous. 

I hope it also appeases @colshrapnel who wrote this article https://phpdelusions.net/articles/generators

### DIFF
--- a/language/generators.xml
+++ b/language/generators.xml
@@ -15,16 +15,17 @@
   </para>
 
   <para>
-   A generator allows you to write code that uses &foreach; to iterate over a
-   set of data without needing to build an array in memory, which may cause
-   you to exceed a memory limit, or require a considerable amount of
-   processing time to generate. Instead, you can write a generator function,
+   A generator offers a convenient way to provide data to &foreach; loops without
+   having to build an array in memory ahead of time, which may cause the program
+   to exceed a memory limit or require a considerable amount of
+   processing time to generate. Instead, a generator function can be used,
    which is the same as a normal
    <link linkend="functions.user-defined">function</link>, except that instead
    of
    <link linkend="functions.returning-values">return</link>ing once, a
    generator can &yield; as many times as it needs to in order to provide the
    values to be iterated over.
+   Like with iterators, random data access is not possible.
   </para>
 
   <para>


language/generators.xml
```diff
@@ -15,16 +15,21 @@
...
    <link linkend="functions.user-defined">function</link>, except that instead
    of
    <link linkend="functions.returning-values">return</link>ing once, a
    generator can &yield; as many times as it needs to in order to provide the
    values to be iterated over.
+   This allows data to be created on demand, rather than delivered all at once.
+   A generator can be iterated over using &foreach;, but random data access
+   is not possible.
+   In a sense, a generator allows the delegation of loop processing to another
+   part of the code.
```
afilina, line 32:
```suggestion
```

This is very confusing to me. What is loop processing? What other part of the code and why? It provides no additional value.

language/generators.xml
```diff
@@ -15,16 +15,21 @@
...
+   processing time to generate. Instead, a generator function can be used,
    which is the same as a normal
    <link linkend="functions.user-defined">function</link>, except that instead
    of
    <link linkend="functions.returning-values">return</link>ing once, a
    generator can &yield; as many times as it needs to in order to provide the
    values to be iterated over.
+   This allows data to be created on demand, rather than delivered all at once.
+   A generator can be iterated over using &foreach;, but random data access
+   is not possible.
```
afilina, line 30:
```suggestion
  Like with iterators, random data access is not possible.
```

We already mention providing generators to foreach earlier in the text. I would like to rephrase it instead.

language/generators.xml
```diff
@@ -15,16 +15,21 @@
...
+   having to build an array in memory ahead of time, which may cause the program
+   to exceed a memory limit or require a considerable amount of
+   processing time to generate. Instead, a generator function can be used,
    which is the same as a normal
    <link linkend="functions.user-defined">function</link>, except that instead
    of
    <link linkend="functions.returning-values">return</link>ing once, a
    generator can &yield; as many times as it needs to in order to provide the
    values to be iterated over.
+   This allows data to be created on demand, rather than delivered all at once.
```
afilina, line 28:
```suggestion
```

"All at once" is already mentioned above. We actually don't need that sentence at all, because we already say that it's a convenient way to implement iterators, so we don't need to repeat what an iterator does.

jimwins (approved):
LGTM.

PR #1857, opened 2022-09-30
update string comparisons possible returns.

devnexen:


reference/strings/functions/strcasecmp.xml
```diff
@@ -47,8 +47,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &lt; 0 if <parameter>string1</parameter> is less than
-   <parameter>string2</parameter>; &gt; 0 if <parameter>string1</parameter>
+   Returns -1 if <parameter>string1</parameter> is less than
+   <parameter>string2</parameter>; 1 if <parameter>string1</parameter>
```
cmb69, line 51: I suggest to mark up all numbers here and elsewhere:
```suggestion
   Returns <literal>-1</literal> if <parameter>string1</parameter> is less than
   <parameter>string2</parameter>; <literal>1</literal> if <parameter>string1</parameter>
```

cmb69 (commented):
Thanks for the PR!

All affected functions would need changelog entries; it probably makes sense to define an XML entity for that.

cmb69:
I pushed a commit with the missing changelog entries; I guess the wording can be improved. What do you think, @devnexen?

devnexen:
thanks @cmb69 looks ok to me

cmb69:
So I'm going to commit. :)

PR #1861, opened 2022-10-01
Fix grammar in code example comment

caendesilva:
Updates a comment in the example as "do something here" is a more complete sentence than "do here something".


devnexen (approved):
nice

caendesilva:
> nice

Thanks!

PR #1866, opened 2022-10-02
Document unserialize max_depth option

### LINKED ISSUE #1822: unserialize_max_depth INI setting is undocumented
While looking to add the link in #1799 for the `unserialize_max_depth` INI setting it looks like it is undocumented.

kevincupp:
Resolves #1822 

I initially set out to resolve just #1822 but noticed `max_depth` also wasn't documented as an option, so I took a swing at that, as well.

### DIFF
--- a/reference/var/functions/unserialize.xml
+++ b/reference/var/functions/unserialize.xml
@@ -104,6 +104,18 @@
            </simpara>
           </entry>
          </row>
+         <row>
+          <entry><literal>max_depth</literal></entry>
+          <entry><type>int</type></entry>
+          <entry>
+           <simpara>
+            The maximum depth of structures permitted during unserialization,
+            and is intended to prevent stack overflows. The default depth limit
+            is <literal>4096</literal> and can be disabled by setting
+            <literal>max_depth</literal> to <literal>0</literal>.
+           </simpara>
+          </entry>
+         </row>
         </tbody>
        </tgroup>
       </table>
@@ -145,6 +157,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>7.4.0</entry>
+       <entry>
+        Added the <literal>max_depth</literal> element of
+        <parameter>options</parameter> to set the maximum depth of structures permitted during unserialization.
+       </entry>
+      </row>
       <row>
        <entry>7.1.0</entry>
        <entry>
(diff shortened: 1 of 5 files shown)


reference/var/functions/unserialize.xml
```diff
@@ -104,6 +104,17 @@
...
           </entry>
          </row>
+         <row>
+          <entry><literal>max_depth</literal></entry>
+          <entry><type>int</type></entry>
+          <entry>
+           <simpara>
+            The maximum depth of structures permitted during unserialization,
+            and is intended to prevent stack overflows. The default depth limit
+            is 4096 and can be disabled by setting max_depth to 0.
```
cmb69, line 114:
```suggestion
            is <literal>4096</literal> and can be disabled by setting <literal>max_depth</literal> to <literal>0</literal>.
```

reference/var/ini.xml
```diff
@@ -49,6 +55,22 @@
...
+   <varlistentry xml:id="ini.unserialize-max-depth">
+    <term xml:id="unserialize-max-depth">
+     <parameter>unserialize_max_depth</parameter>
+     <type>int</type>
+    </term>
+    <listitem>
+     <para>
+      The maximum depth of structures permitted during unserialization when
+      using <function>unserialize</function>, and is intended to prevent stack
+      overflows. This can be disabled by setting unserialize_max_depth=0.
```
cmb69, line 67:
```suggestion
      overflows. This can be disabled by setting <literal>unserialize_max_depth=0</literal>.
```

reference/var/ini.xml
```diff
@@ -49,6 +55,22 @@
...
+     </para>
+     <para>
+      See also <function>unserialize</function> and <link linkend="language.oop5.autoload">Autoloading Objects</link>.
```
cmb69, line 70:
```suggestion
      See also <function>unserialize</function> and <link linkend="language.oop5.autoload">Autoloading Classes</link>.
```

cmb69 (commented):
Thank you! This looks good except for the minor issues commented below.

One improvement that could be done right away would be avoid duplicating the info in ini.list.xml by including the `<tbody>`'s contents from var/ini.xml. If you feel up to that, great! Otherwise that can also be done later.

kevincupp:
> One improvement that could be done right away would be avoid duplicating the info in ini.list.xml by including the <tbody>'s contents from var/ini.xml. If you feel up to that, great! Otherwise that can also be done later.

Great call! Best not to duplicate knowledge. Still figuring out PhD. Applied all other suggestions, too.

cmb69 (approved):
Well done, thank you!

PR #1869, opened 2022-10-03
add HTTP Last-Modified header example

divinity76:
ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified


cmb69:
In my opinion, an example for "Last-Modified" would be more suitable for the `header()` page.

divinity76:
dang, it already is on the header() page! ok nvm. wish i knew that yesterday ^^

PR #1874, opened 2022-10-08
Clarify opcache.preload_user behavior

arnaud-lb:


reference/opcache/preload.xml
```diff
@@ -42,10 +42,11 @@ opcache.preload=preload.php
...
-  (PHP-FPM, mod_php, etc.) and load code into persistent memory.  If PHP will be run as
-  root (not recommended), the <link linkend="ini.opcache.preload-user">opcache.preload_user</link>
-  value can specify an alternate system user to run the preloading.  Running preloading as
-  root is not allowed.
+  (PHP-FPM, mod_php, etc.) and load code into persistent memory.  In servers that startup as
+  root before switching to an unprivileged system user, or if PHP will be run as root
+  (not recommended), the <link linkend="ini.opcache.preload-user">opcache.preload_user</link>
+  value can specify the system user to run the preloading.  Running preloading as root is not allowed
+  by default. Set <literal>opcache.preload_user=root</literal> to explicitly allow it.
  </para>
```
arnaud-lb, line 50: With the exception that preloading as root is allowed by default in the CLI SAPI in master, but I'm not sure how to mention it, nor if it's worth it to mention it.

cmb69:
Thank you for the PR! We also should update the respective part of the runtime configuration:
https://github.com/php/doc-en/blob/4c4b82965384d55f5c3efb1ffa80615acd98a737/reference/opcache/ini.xml#L938-L939

reference/opcache/preload.xml
cmb69: Well, if it's "master" only, it shouldn't be documented yet, anyway. If there is some info about that in UPGRADING, it should be catched automatically for the PHP 8.3 migration guide.

cmb69 (approved):
Thank you!

PR #1875, opened 2022-10-09
Update parse-ini-string.xml

mowangjuanzi:
Unified with parse_ini_file

### DIFF
--- a/reference/filesystem/functions/parse-ini-string.xml
+++ b/reference/filesystem/functions/parse-ini-string.xml
@@ -81,7 +81,7 @@
     <literal>false</literal> result in <literal>""</literal>, and values
     <literal>on</literal>, <literal>yes</literal> and <literal>true</literal> result
     in <literal>"1"</literal>, unless <constant>INI_SCANNER_TYPED</constant> mode is used.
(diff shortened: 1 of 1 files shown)


cmb69:
While `[` may be [used in values](https://3v4l.org/FY1X1), I don't think it can be [used in keys](https://3v4l.org/TmnSp). We should clarify that.

mowangjuanzi:
@cmb69 thanks for the reply.

Line [341](https://github.com/php/doc-en/blob/master/reference/filesystem/functions/parse-ini-file.xml#L341) in parse_ini_file.xml  

```xml
    Characters <literal>?{}|&amp;~!()^"</literal> must not be used anywhere in
    the key and have a special meaning in the value.
```

I understand that the processing mechanism of the parse_ini_file function and the parse_ini_string function should be the same.

So the description of parse_ini_file should be modified?

cmb69:
> I understand that the processing mechanism of the parse_ini_file function and the parse_ini_string function should be the same.

Right. And so should the documentation. With regard to this issue (the `[`), neither man page is correct. :(

mowangjuanzi:
OK, then I'll close the current PR. waiting for update.

PR #1876, opened 2022-10-10
Added example for specifying max decimal places in sprintf

stollr:
Formatting numbers with a maximum number of decimal places (and not a fixed number of decimals) is a general need. But the description of the `g` specifier (and its variants `G`, `h` and `H`) is hard to understand.

The example helps to understand how to define the precision correctly.

If you like this PR I'd be happy if you would merge it or label it with [`hacktoberfest-accepted`](https://hacktoberfest.com/participation/). Thanks in advance.

### DIFF
--- a/reference/strings/functions/sprintf.xml
+++ b/reference/strings/functions/sprintf.xml
@@ -206,6 +206,33 @@ echo $formatted;
    </screen>
   </example>
   <example>
+   <title><function>sprintf</function>: specifying the maximum number of decimal places</title>
+   <programlisting role="php">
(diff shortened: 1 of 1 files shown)


jimwins:
I don't think this example shows enough to really make it clear how `%g` and `%h` work. It should show examples of when using the `g` and `h` specifiers flip into scientific notation, and also how `g` is locale-dependent and `h` is not. Also, how the precision controls the number of digits for the whole output, not just the number of digits after the decimal like it does for `%f`.

PR #1878, opened 2022-10-10
Fix consistency for is_a()

victor-prdh:
Hi there, 
During my work for the translation on doc-fr, i saw `reference/classobj/functions/is-a.xml` seems not very consistent. It's speak about "object type" and "class" for the same things, it's maybe better to keep only one of those ?

Related PR on doc-fr [here](https://github.com/php/doc-fr/pull/293) 

Thanks !

### DIFF
--- a/reference/classobj/functions/is-a.xml
+++ b/reference/classobj/functions/is-a.xml
@@ -15,7 +15,7 @@
   </methodsynopsis>
   <para>
    Checks if the given <parameter>object_or_class</parameter> is of this object type or has
-   this class as one of its parents.
+   this object type as one of its parents.
   </para>
  </refsect1>
  <refsect1 role="parameters">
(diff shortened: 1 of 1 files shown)


jorgsowa:
Hi @victor-prdh,
we can pass both class name or object to this function. What do you think that we should keep both names, i.e. in manner `class/object type`. This way we could refer to both types of parameter.

sy-records:
I also have a PR. #1789 
I think we need to determine how to properly describe in the documentation.
cc @cmb69

victor-prdh:
Oh, didn't see your PR @sy-records, sorry ! I think the `class/object type` notation is a good compromise and give all the info with a correct readability.

You wanna take it on your PR and i close this one or we do it on this one ?

mumumu:
relates to #506

victor-prdh:
> relates to #506

Thanks, i close this one !

PR #1879, opened 2022-10-10
Outdated information about warning of array_change_key_case

jorgsowa:
Function throws `Fatal error: Uncaught TypeError: array_change_key_case(): Argument #1 ($array) must be of type array` which is default value of generic incorrect type passed to the function. It's not anymore internal behavior of the function.

### DIFF
--- a/reference/array/functions/array-change-key-case.xml
+++ b/reference/array/functions/array-change-key-case.xml
@@ -55,14 +55,6 @@
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


cmb69:
While the documentation currently is supposed to document PHP 7 and PHP 8, this warning can indeed be removed, since it is not special for that function but rather [general behavior of internal functions](https://www.php.net/manual/en/functions.internal.php).

So thank you!

PR #1880, opened 2022-10-10
Sync SplFixedArray methods with new implementation

felixdorn:
Since PHP 8.0, SplFixedArray does not implement `IteratorAggregate` and thus, the following methods do not exist anymore: `rewind`, `key`, `current`, `valid`, `next`.


Girgias:
This change is not correct, as the manual still documents PHP 7, this should instead have a changelog entry explaining that those methods were removed.
Also ``version.xml`` probably should be looked at

felixdorn:
It seems like `versions.xml` correctly references the removal of the new methods.

https://github.com/php/doc-en/blob/master/reference/spl/versions.xml#L965

felixdorn:
It looks like `versions.xml` correctly references the removal of the new methods.

https://github.com/php/doc-en/blob/master/reference/spl/versions.xml#L965

There's also an  entry in the changelog in `splfixedarray.xml`, 
```
 <row>
       <entry>8.0.0</entry>
       <entry>
        <classname>SplFixedArray</classname> implements
        <interfacename>IteratorAggregate</interfacename> now.
        Previously, <interfacename>Iterator</interfacename> was implemented instead.
       </entry>
      </row>
```

Does that mean that the documentation is accurate and should show the `Iterator`-related methods even though they do not exist anymore (because the docs are for 7 and 8)?

cmb69:
> Does that mean that the documentation is accurate and should show the `Iterator`-related methods even though they do not exist anymore (because the docs are for 7 and 8)?

Ugh, tricky. I think we have two options:

* we can keep these methods, but we should clarify that they have been removed as of PHP 8.0.0 (something like [`&warn.removed.function-7-0-0;`](https://github.com/php/doc-en/blob/4c4b82965384d55f5c3efb1ffa80615acd98a737/language-snippets.ent#L378-L380), but for PHP 8.0.0)
* we can remove these methods, since their documentation doesn't add value to the general `Iterator` method documentation

Either is likely somewhat confusing for users, but I think neither option poses difficulties wrt. gen_stub.php (@kocsismate, please correct me if I'm wrong).

In any way, we need to document `::getIterator()`.

felixdorn:
There are two things:
* The [class synopsis](https://www.php.net/manual/en/class.splfixedarray.php) is wrong, as `SplFixedArray` implements `IteratorAggregate` but shows the methods related to `Iterator` and not the `;;getIterator` method
* The individual methods `::rewind`, `::next`, `::current`, `::key`, `::valid`, (for example, [rewind](https://www.php.net/manual/en/splfixedarray.rewind.php), are definitely missing a warning about their removal in PHP 8.0.0.

My take (don't know if it's possible):
* Remove the `IteratorAggregate` methods from the class synopsis
* Keep the  `IteratorAggregate` methods (in the context of `SplFixedArray`) individual pages, add a warning about their removal in PHP 8.0.0
* In the table of contents, keep the `::rewind`, etc methods with a warning saying that these exist prior to PHP 8.0 (https://3v4l.org/Y69Al)
* Document `::getIterator` in the class synopsis and maybe add a warning about it being PHP8+ in the table of contents.

PR #1882, opened 2022-10-12
Fix/gh 1833 socket addrinfo lookup inaccurate descriptions

### LINKED ISSUE #1833: Documentation socket_addrinfo_lookup some inaccurate description
Manual page: https://www.php.net/manual/en/function.socket-addrinfo-lookup.php#refsect1-function.socket-addrinfo-lookup-parameters

1. The explain for `service` parameter:

> The service to connect to. If service is a name, it is translated to the corresponding port number.

`If service is a name` is not precise to explain the parameter. I think it's better interpreted is: `If service is a service name in services`.
`services(5)` linux man page: https://man7.org/linux/man-pages/man5/services.5.html

2. The explain for `hints` parmaeter:

> Hints provide criteria for selecting addresses returned. You may specify the hints as defined by getadrinfo.

The last word: `getadrinfo`, should be replaced with "getaddrinfo"

AtlanticF:
Fix #1833

### DIFF
--- a/reference/sockets/functions/socket-addrinfo-lookup.xml
+++ b/reference/sockets/functions/socket-addrinfo-lookup.xml
@@ -36,8 +36,8 @@
     <term><parameter>service</parameter></term>
     <listitem>
      <para>
-      The service to connect to.  If service is a name, it is translated to the corresponding
-      port number.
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you very much!

PR #1890, opened 2022-10-15
Removed obsolete mysqlnd install description.

mumumu:
We don't have to use MySQL Client Library(libmysqlclient) for installing MySQL database extension, because we already have mysqlnd.


cmb69 (approved):
Thank you!

PR #1892, opened 2022-10-16
Update return value for levenshtein

zing-james:
The character limit on the 2 string arguments for the levenshtein function was removed in PHP 8. The documentation should be updated to reflection this difference. 

https://github.com/php/php-src/commit/6a8c094e2d86a8d9294e4d7fddabd9486bc92ed8

### DIFF
--- a/reference/strings/functions/levenshtein.xml
+++ b/reference/strings/functions/levenshtein.xml
@@ -113,7 +113,7 @@
      <row>
       <entry>8.0.0</entry>
       <entry>
-       Prior to this version, <function>levenshtein</function> would return -1
+       Prior to this version, <function>levenshtein</function> would return <literal>-1</literal>
        if one of the argument strings is longer than 255 characters.
       </entry>
      </row>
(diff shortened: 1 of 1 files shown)


devnexen:
it might deserve a changelog entry [somewhere here](https://github.com/php/doc-en/blob/master/language-snippets.ent).

reference/strings/functions/levenshtein.xml
```diff
@@ -111,6 +110,13 @@
        with either two or five arguments.
       </entry>
      </row>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       Prior to this version, <function>levenshtein</function> would return -1
```
cmb69, line 116:
```suggestion
       Prior to this version, <function>levenshtein</function> would return <literal>-1</literal>
```

cmb69 (commented):
Thank you!

> it might deserve a changelog entry [somewhere here](https://github.com/php/doc-en/blob/master/language-snippets.ent).

There is no need to define a new entity if that entity isn't reused (at least once, but better twice or more often).

cmb69:
Thank you!

PR #1894, opened 2022-10-16
JSON DECODE: Change `PHP variable` into `PHP value`

### LINKED ISSUE #1431: json_decode documentation could be clearer
re; https://www.php.net/manual/en/function.json-decode.php

It may seem pedantic but I think it would be really helpful to change your code example from 

`$json = '{"foo-bar": 12345}';`

to

`$jsonString = '{"foo-bar": 12345}';`

The whole point of the function is to convert a JSON String into JSON. In which case calling the name of the variable which contains the JSON String $json is misleading.

The description given for the function is 

> Takes a JSON encoded string and converts it into a PHP variable. 

I do not really understand what is meant by "converts it into a PHP variable". Could the function not be used like this:

`echo (json_decode($jsonString));`

If so, would it be converting it into a PHP variable?

I realize I am somewhat confused by this whole issue so apologies if I am barking up the wrong tree.

sjokkateer:
Closes #1431

### DIFF
--- a/reference/json/functions/json-decode.xml
+++ b/reference/json/functions/json-decode.xml
@@ -16,7 +16,7 @@
    <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer>0</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Takes a JSON encoded string and converts it into a PHP variable.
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you!

PR #1895, opened 2022-10-16
Replace arg to num

mowangjuanzi:


reference/math/functions/cosh.xml
```diff
@@ -13,7 +13,7 @@
   </methodsynopsis>
   <para>
    Returns the hyperbolic cosine of <parameter>num</parameter>,
-   defined as <literal>(exp(arg) + exp(-arg))/2</literal>.
+   defined as <literal>(exp(num) + exp(-num))/2</literal>.
```
cmb69, line 16: Maybe we should use `$num` instead:
```suggestion
   defined as <literal>(exp($num) + exp(-$num))/2</literal>.
```

cmb69:
Thank you!

PR #1903, opened 2022-10-19
Reduce overhead of SQL placeholder creation

TinoDidriksen:
The incredibly wasteful way this example creates SQL placeholders has long annoyed me, so here's a fix.

### DIFF
--- a/reference/pdo/pdostatement/execute.xml
+++ b/reference/pdo/pdostatement/execute.xml
@@ -145,7 +145,7 @@ $sth->execute();
 /* Execute a prepared statement using an array of values for an IN clause */
 $params = array(1, 21, 63, 171);
 /* Create a string for the parameter placeholders filled to the number of params */
-$place_holders = implode(',', array_fill(0, count($params), '?'));
+$place_holders = '?'.str_repeat(', ?', count($params)-1);
 
 /*
     This prepares the statement with enough unnamed placeholders for every value
--- a/reference/pdo/pdostatement/execute.xml
+++ b/reference/pdo/pdostatement/execute.xml
@@ -145,7 +145,7 @@ $sth->execute();
 /* Execute a prepared statement using an array of values for an IN clause */
 $params = array(1, 21, 63, 171);
 /* Create a string for the parameter placeholders filled to the number of params */
-$place_holders = '?'.str_repeat(', ?', count($params)-1);
+$place_holders = '?' . str_repeat(', ?', count($params) - 1);
 
 /*
     This prepares the statement with enough unnamed placeholders for every value


cmb69:
In my opinion, code (and especially the code snippets in the manual) should be optimized for readability, and only optimized for performance, if necessary. In this case, the performance doesn't matter (the SQL overhead is way bigger than the array construction and joining of its elements). That said, I don't think what we currently have is particularly readable. Why don't we just use `"?, ?, ?, ?"` instead? We might even inline that variable.

TinoDidriksen:
`str_repeat()` definitely also wins on readability. I do think the manual should show how to handle a variable number of `IN` params, because that is a very common question - and I'd like to be able to point people to the manual without cringing or provide a caveat.

kamil-tekiela:
I also think the example should show how to dynamically create a list of placeholders. Whether it's `str_repeats` or `array_fill` is a matter of choice. Technically calling `implode` + `array_fill` is 2x slower than `str_repeats`, but it doesn't matter. These operations are so fast that their cost is negligible in the code. 

My code style prefers `array_fill` because it can handle array with 0 elements. In this case, it doesn't matter because you cannot have `IN()` with no placeholders. So, technically, str_repeat might be more suitable, but not because of performance. 

To sum it up: I would keep it as it is but if it's an eyesore, we can replace with `str_repeat`.

cmb69:
> I also think the example should show how to dynamically create a list of placeholders.

I agree, but I wouldn't do that for a hard-coded list of parameters. So maybe instead of `$params = array(1, 21, 63, 171);` we should rather fetch a list of IDs from the query string and split it, e.g. `$params = array_map("trim", explode(",", $_GET["ids"]));`). Or just expect `$_GET["ids"]` to be an array; or whatever.

kamil-tekiela:
We could use `$_GET['ids']` but I expect that some people will then complain about lack of validation. What if it's not a one-dimensional array?

cmb69:
We could add a comment, or even actually employ `filter_input_array`.

reference/pdo/pdostatement/execute.xml
```diff
@@ -145,7 +145,7 @@ $sth->execute();
 /* Execute a prepared statement using an array of values for an IN clause */
 $params = array(1, 21, 63, 171);
 /* Create a string for the parameter placeholders filled to the number of params */
-$place_holders = implode(',', array_fill(0, count($params), '?'));
+$place_holders = '?'.str_repeat(', ?', count($params)-1);
```
jimwins, line 148:
```suggestion
$place_holders = '?' . str_repeat(', ?', count($params) - 1);
```

jimwins (approved):
Just a code-style nit, and then I think this is fine to merge.

A fancier example with a dynamic number of entries would be great, but let's not hold up this simple improvement any longer waiting for that.

PR #1905, opened 2022-10-19
Fix markup

mowangjuanzi:
Is this modification correct?


cmb69:
Indeed, there is something wrong with the rendering. Possibly only something that would need to be fixed in PhD, but maybe it's something we should fix in the docs. Reopening, to not forget about it.

PR #1909, opened 2022-10-21
Fix #1897: Document limits of NumberFormatter and possible workaround

### LINKED ISSUE #1897: Narrow currency symbols
### Description

Current PHP versions support currency formatting showing "standard" currency symbols for various locales, such as `Kč`, `$` or `CA$`, however in some context it is bit unexpected for people to see the `CA$` as they expect what they see everyday `$`...

In terminology of the ICU library this is called a `narrow` unit symbol (see [docs](https://unicode-org.github.io/icu/userguide/format_parse/numbers/skeletons.html)).

Could the support for this be added?

repli2dev:
Document a way how to get currency formatted with narrow currency symbol as discussed in #1897.

### DIFF
--- a/reference/intl/messageformatter/format-message.xml
+++ b/reference/intl/messageformatter/format-message.xml
@@ -30,8 +30,8 @@
   <para>
    Quick formatting function that formats the string without having to
    explicitly create the formatter object. Use this function when the format
-   operation is done only once and does not need and parameters or state to be
-   kept.
+   operation is done only once and does not need and parameters, state to be
+   kept or when wanting to customize the output by providing additional context to ICU directly.
   </para>
  </refsect1>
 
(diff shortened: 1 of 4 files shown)


reference/intl/messageformatter/format-message.xml
```diff
@@ -30,8 +30,8 @@
   <para>
    Quick formatting function that formats the string without having to
    explicitly create the formatter object. Use this function when the format
-   operation is done only once and does not need and parameters or state to be
-   kept.
+   operation is done only once and does not need and parameters, state to be
```
cmb69, line 33: It seems to me that there was a typo which renders the sentence confusing. I think it should be:
```suggestion
   operation is done only once and does not need any parameters or state to be
```

reference/intl/messageformatter/format-message.xml
```diff
@@ -108,6 +108,23 @@ echo MessageFormatter::formatMessage("de", "{0,number,integer} Affen auf {1,numb
 4.560 Affen auf 123 Bäumen sind 37,073 Affen pro Baum
 ]]>
   </screen>
+  <example>
+   <title>Instructing ICU to format currency with common and with narrow currency symbol</title>
+   <programlisting role="php">
```
cmb69, line 113: We should mention that this is only supported as of ICU 67 (some users likely use older versions). So maybe:
```suggestion
   <simpara>Requires ICU ≥ 67.</simpara>
   <programlisting role="php">
```

cmb69 (commented):
Thank you! Besides the issues I've mentioned, this looks good to me.

reference/intl/messageformatter/format-message.xml
repli2dev: This seems to be and error in the copy paste, as the same error is in `parse-message.xml`

reference/intl/messageformatter/parse-message.xml
```diff
@@ -30,7 +30,7 @@
   <para>
    Parses input string without explicitly creating the formatter object. Use
    this function when the format operation is done only once and does not need
-   and parameters or state to be kept.
+   any parameters or state to be kept.
```
repli2dev, line 33: @cmb69 ... so I fixed it even here.

cmb69:
Thank you!

PR #1918, opened 2022-10-26
Missing false return val in synopsis of datetime sub/add

webmaster777:


cmb69:
Thank you for the PR! However, these functions only returned `false` prior to PHP 8.0.0 in case invalid argument types had been supplied. This is already [documented generally](https://www.php.net/manual/en/functions.internal.php):

> If the parameters given to a function are not what it expects, such as passing an array where a string is expected, the return value of the function is undefined.

webmaster777:
Okay, but it is inconsistent with the return value specification which still says false is a possible return value. I don't know whether that still holds up then? Compare with e.g. [modify](https://www.php.net/manual/en/datetime.modify.php)

cmb69:
Ah, right! Should be fixed now: https://github.com/php/doc-en/commit/769fdeb145fdc4a022f6607994df5d13ac71d1d8.

PR #1922, opened 2022-10-30
Extend ext/random documentation (2)

TimWolla:


cmb69 (commented):
Thank you! This is fine, except that the (3?) new URLs should be defined as XML entities in https://github.com/php/doc-base/blob/master/entities/global.ent (so that when they change, translations won't have to be updated).

TimWolla:
@zeriyoshi Friendly ping, have you seen the review request? :smiley:

Girgias (approved):
Can't comment on the accuracy of the ``seed`` parameter description, but LGTM

zeriyoshi (approved):

zeriyoshi:
@TimWolla 
I did not have a very deep understanding of the documentation system and was waiting for the noted items to be resolved.
I think we're good to go, thanks!

PR #1923, opened 2022-10-31
DOMNode::$prefix is not nullable

tstarling:
The fieldsynopsis above documents it as non-nullable, so don't say it can be null. Checking the source of dom_node_prefix_read(), it has always returned a string since the initial commit, it can't return null.


cmb69:
Thank you!

PR #1932, opened 2022-11-03
Updates for hash functions that now throw errors

tstarling:
Update the manual for the changes to the hash functions by Mark Randall in August 2019 and subsequent tweaks by Máté Kocsis in March 2020.

Build tested.

### DIFF
--- a/reference/hash/functions/hash-equals.xml
+++ b/reference/hash/functions/hash-equals.xml
@@ -52,14 +52,6 @@
   </para>
  </refsect1>
 
- <refsect1 role="errors">
-  &reftitle.errors;
-  <para>
-   Emits an <constant>E_WARNING</constant> message when either of the
-   supplied parameters is not a string.
-  </para>
- </refsect1>
-
  <refsect1 role="examples">
   &reftitle.examples;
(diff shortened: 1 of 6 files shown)


reference/hash/functions/hash-equals.xml
```diff
@@ -55,11 +55,37 @@
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   Emits an <constant>E_WARNING</constant> message when either of the
+   Throws a <classname>TypeError</classname> exception when either of the
```
cmb69, line 58: We don't document `TypeError`s for individual functions. Instead we rely on the [general documentation](https://www.php.net/manual/en/functions.internal.php):

So the whole error section and the changelog should be removed.

cmb69 (commented):
Thank you! Except for the `TypeError` documentation, this looks good to me.

reference/hash/functions/hash-equals.xml
tstarling: Done.

cmb69:
Thank you!

PR #1933, opened 2022-11-03
Add documentation for libxml_get_external_entity_loader()

tstarling:


cmb69 (approved):
Thank you!

I prefer to omit the folding markers (`<!-- {{{ -->` and `<!-- }}} -->`), since they don't appear to make much sense anymore. It's okay to leave them as is, though.

tstarling:
The folding markers came from the skeleton (doc-base/RFC/skeletons/function.xml). I can make a PR deleting them from there if you like.

PR #1934, opened 2022-11-03
Updates for RFC: Locale-independent case conversion

tstarling:
* Add a detailed description of what is meant by ASCII case conversion to strtolower() and strtoupper().
* Replace locale-related text on ucfirst(), lcfirst() and ucwords() with text about ASCII case conversion.
* Remove entity note.locale-single-byte since it was only used on ucwords().
* Add changelog entries to all the functions mentioned in the RFC.

### DIFF
--- a/reference/array/constants.xml
+++ b/reference/array/constants.xml
@@ -15,7 +15,8 @@
      <constant>CASE_LOWER</constant> is used with
      <function>array_change_key_case</function> and is used to convert array
      keys to lower case. This is also the default case for
-     <function>array_change_key_case</function>.
+     <function>array_change_key_case</function>. Since PHP 8.2.0, only ASCII
+     characters will be converted.
     </simpara>
    </listitem>
   </varlistentry>
@@ -28,7 +29,8 @@
     <simpara>
      <constant>CASE_UPPER</constant> is used with
      <function>array_change_key_case</function> and is used to convert array
-     keys to upper case.
+     keys to upper case. Since PHP 8.2.0, only ASCII characters will be
+     converted.
     </simpara>
    </listitem>
   </varlistentry>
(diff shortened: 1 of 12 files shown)


reference/array/constants.xml
```diff
@@ -15,7 +15,8 @@
      <constant>CASE_LOWER</constant> is used with
      <function>array_change_key_case</function> and is used to convert array
      keys to lower case. This is also the default case for
-     <function>array_change_key_case</function>.
+     <function>array_change_key_case</function>. Since PHP 8.2.0, only ASCII
+     characters will be converted.
```
Girgias, line 19:
```suggestion
     <function>array_change_key_case</function>. As of PHP 8.2.0, only ASCII
     characters will be converted.
```
For versions we use ``As of`` instead of ``since``

reference/array/constants.xml
```diff
@@ -28,7 +29,8 @@
     <simpara>
      <constant>CASE_UPPER</constant> is used with
      <function>array_change_key_case</function> and is used to convert array
-     keys to upper case.
+     keys to upper case. Since PHP 8.2.0, only ASCII characters will be
+     converted.
```
Girgias, line 33:
```suggestion
     keys to upper case. As of PHP 8.2.0, only ASCII characters will be
     converted.
```

reference/array/constants.xml
```diff
@@ -130,10 +132,10 @@
    </term>
    <listitem>
     <simpara>
-    <constant>SORT_FLAG_CASE</constant> can be combined
-         (bitwise OR) with
-         <constant>SORT_STRING</constant> or
-         <constant>SORT_NATURAL</constant> to sort strings case-insensitively.
+     <constant>SORT_FLAG_CASE</constant> can be combined (bitwise OR) with
+     <constant>SORT_STRING</constant> or <constant>SORT_NATURAL</constant> to
+     sort strings case-insensitively. Since PHP 8.2.0, only ASCII case folding
```
Girgias, line 137:
```suggestion
     sort strings case-insensitively. As of PHP 8.2.0, only ASCII case folding
```

reference/strings/functions/lcfirst.xml
```diff
@@ -47,6 +43,29 @@
...
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.2.0</entry>
+      <entry>
+       Case conversion no longer depends on the locale set with
+       <function>setlocale</function>. Only ASCII characters will be converted.
```
Girgias, line 61: Add an XML entity for this and the case folding changelog entries?

Girgias (changes requested):

reference/strings/functions/lcfirst.xml
tstarling: Done.

Girgias (approved):
LGTM :)

PR #1938, opened 2022-11-04
iterator_to_array() and iterator_count() now accepts arrays

Girgias:


reference/spl/functions/iterator-count.xml
```diff
@@ -42,6 +42,30 @@
...
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.2.0</entry>
+      <entry>
+       The type of <parameter>iterator</parameter> has been widened from
+       <type>Traversable</type> to
+       <type class="union"><type>Traversable</type><type>array</type></type>
```
TimWolla, line 61:
```suggestion
       <type class="union"><type>Traversable</type><type>array</type></type>.
```
Make this a complete sentence.

reference/spl/functions/iterator-to-array.xml
```diff
@@ -66,6 +66,30 @@
...
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.2.0</entry>
+      <entry>
+       The type of <parameter>iterator</parameter> has been widened from
+       <type>Traversable</type> to
+       <type class="union"><type>Traversable</type><type>array</type></type>
```
TimWolla, line 85:
```suggestion
       <type class="union"><type>Traversable</type><type>array</type></type>.
```

TimWolla (approved):
Technically the stub uses `iterable`, but I guess it makes sense to use the explicit union here, especially when combined with the internal change.

Girgias:
> Technically the stub uses `iterable`, but I guess it makes sense to use the explicit union here, especially when combined with the internal change.

I've generated the change via gen_stubs, so maybe that's why. But in either case it's equivalent :)

vudaltsov:
Today I suddenly learned about this feature and was very happy to remove `if ($myIterable instanceof Traversable)`. Nevertheless, it was illogical to see `array|Traversable` instead of `iterable` in the documentation. I started researching whether this has any special meaning or not, and checked the implementation PR first, and finally this one.

Do you mind if I change `array|Traversable` to `iterable` in a separate PR?

TimWolla:
> Do you mind if I change `array|Traversable` to `iterable` in a separate PR?

According to Girgias, the explicit union is how it falls out of the automated generator consuming the stubs. Thus a manual change would be undesirable. This would need to be handled in the generator.

Girgias:
`iterable` is converted to `array|Traversable` at compile time and this is the information that Reflection sees, I don't really see a benefit in forcing it to be `iterable`

PR #1939, opened 2022-11-04
dba_fetch() signature has changed to follow userland semantics

Girgias:
The overloaded signature is still supported but discouraged.


cmb69 (approved):
Thank you!

PR #1940, opened 2022-11-04
Rewrite outdated note for session_unset()

### LINKED ISSUE #1906: A note on the session_unset() page regarding older deprecated code.
On the [session_unset() page](https://www.php.net/manual/en/function.session-unset.php) there is a note, saying

> **Note:**
> Only use **session_unset()** for older deprecated code that does not use `$_SESSION`

I suppose it was a standard note that was added to all deprecated functions such as `session_register()`. Eventually such functions got removed, but `session_unset()` remains. Given there is no "older deprecated code" in the language anymore, I suppose this note is no longer relevant and could be removed? I mean, given the function stays, it means it's OK to use it?

Girgias:
This was meant to be used when  was still around.

Fixes #1906

### DIFF
--- a/reference/session/functions/session-unset.xml
+++ b/reference/session/functions/session-unset.xml
@@ -59,20 +59,19 @@
    <para>
     If <varname>$_SESSION</varname> is used, use <function>unset</function> to
     unregister a session variable, i.e.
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you!

It might make sense to deprecate `session_unset()`.

PR #1942, opened 2022-11-04
Document dynamic properties

Girgias:
Because obviously this is not documented *sigh*

### DIFF
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -326,6 +326,28 @@ $test->obj = new stdClass;
    </para>
   </sect2>
 
+  <sect2 xml:id="language.oop5.properties.dynamic-properties">
+   <title>Dynamic properties</title>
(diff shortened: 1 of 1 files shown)


language/oop5/properties.xml
```diff
@@ -326,6 +326,28 @@ $test->obj = new stdClass;
    </para>
   </sect2>
 
+  <sect2 xml:id="language.oop5.properties.dynamic-properties">
+   <title>Dynamic properties</title>
+   <para>
+    If trying to access a non existant property on an &object;,
```
cmb69, line 332: "access" is confusing in this context (could mean read access). Maybe better "assign to":
```suggestion
    If trying to assign to a non existant property on an &object;,
```

cmb69 (commented):
Thank you! Looks mostly good, besides the "access" issue.

PR #1943, opened 2022-11-04
Sync ext/spl class constants with stubs - part 1

kocsismate:
These are usually simple sync changes: visibility modifier is added, value is removed. Only ~2 new constants are added.

### DIFF
--- a/reference/spl/regexiterator.xml
+++ b/reference/spl/regexiterator.xml
@@ -37,40 +37,46 @@
 
     <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
     <fieldsynopsis>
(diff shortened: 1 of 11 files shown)


reference/spl/regexiterator.xml
```diff
@@ -167,6 +173,15 @@
        </para>
       </listitem>
      </varlistentry>
+
```
Girgias, line 176:
```suggestion

```
Whitespaces

kocsismate: Sorry, I force pushed: I made a few whitespace changes on the regexiterator and filesystemiterator pages

cmb69 (approved):
Thank you!

PR #1945, opened 2022-11-04
Sync ext/intl class constants with stubs - part 1

kocsismate:


reference/intl/spoofchecker.xml
```diff
@@ -39,86 +39,86 @@
      </ooclass>
     </classsynopsisinfo>
 
-    <classsynopsisinfo role="comment">Constants</classsynopsisinfo>
```
kocsismate, line 42: Only the order of the constants changed, except that a few of them were declared as `int|float`, but now all of them are `int`.

Girgias (approved):
What I've learned is that the INTL constants are barely documented.

cmb69:
> What I've learned is that the INTL constants are barely documented.

The whole extension's documentation is insufficient. For instance, there are still 86 `&warn.undocumented.func;`. :(

PR #1951, opened 2022-11-05
Extend ext/random documentation (3)

TimWolla:
With this PR, the parameters and return values should be described for every method, except:

- Randomizer::nextInt(): This method isn't great and I'm not sure what can be usefully written there.
- The various __serialize(), __unserialize(), __debugInfo() methods. Is there some existing documentation where I can steal from?

What's missing and is intentionally omitted so far:

- Description for Random\Engine.
- Long-form descriptions for the methods.
- Examples.
- Further Cleanup

### DIFF
--- a/reference/random/functions/random-bytes.xml
+++ b/reference/random/functions/random-bytes.xml
@@ -38,7 +38,7 @@
  <refsect1 role="returnvalues"><!-- {{{ -->
   &reftitle.returnvalues;
   <para>
-   Returns a string containing the requested number of cryptographically
+   A string containing the requested number of cryptographically
    secure random bytes.
   </para>
  </refsect1><!-- }}} -->
--- a/reference/random/random/engine/mt19937/construct.xml
+++ b/reference/random/random/engine/mt19937/construct.xml
@@ -27,7 +27,14 @@
     <term><parameter>seed</parameter></term>
     <listitem>
      <para>
+      Fills the state with values generated with a linear congruential generator
+      that was seeded with <parameter>seed</parameter> interpreted as an unsigned
+      32 bit integer.
+     </para>
 
+     <para>
+      If <parameter>seed</parameter> is omitted or &null;, a random unsigned
+      32 bit integer will be used.
      </para>
     </listitem>
    </varlistentry>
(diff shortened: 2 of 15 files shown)


zeriyoshi:
OK like this?

> Randomizer::nextInt(): This method isn't great and I'm not sure what can be usefully written there.

```
This method returns raw platform size-dependent values by means of a random number generator. In general, if you do not understand the meaning of this description, or if the range of numbers you are looking for is fixed, use getInt() instead.
```

> The various __serialize(), __unserialize(), __debugInfo() methods. Is there some existing documentation where I can steal from?

As with `\DateTime`, it seems unnecessary to describe the details of these magic methods.

reference/random/random/engine/mt19937/construct.xml
```diff
@@ -27,15 +27,47 @@
...
     <term><parameter>mode</parameter></term>
     <listitem>
      <para>
-
+      Use one of the following constants to specify the implementation of the algorithm to use.
+      <informaltable>
+       <tgroup cols="2">
+        <thead>
+         <row>
+          <entry>Constant</entry>
```
Girgias, line 50:
```suggestion
          <entry>&Constants;</entry>
```

reference/random/random/engine/mt19937/construct.xml
```diff
@@ -27,15 +27,47 @@
...
+          <entry><constant>MT_RAND_MT19937</constant></entry>
+          <entry>
+           Uses the standardized Mt19937 implementation.
+          </entry>
+         </row>
+         <row>
+          <entry><constant>MT_RAND_PHP</constant></entry>
+          <entry>
+           Uses an incorrect implementation for backwards compatibility with <function>mt_srand</function> before
+           PHP 7.1.
```
Girgias, line 65:
```suggestion
           Uses an incorrect implementation for backwards compatibility with <function>mt_srand</function> prior to
           PHP 7.1.0.
```
Full version

Girgias (approved):
Other than the nits LGTM

reference/random/functions/random-bytes.xml
```diff
@@ -38,7 +38,7 @@
  <refsect1 role="returnvalues"><!-- {{{ -->
   &reftitle.returnvalues;
   <para>
-   Returns a string containing the requested number of cryptographically
+   A string containing the requested number of cryptographically
```
cmb69, line 41:
```suggestion
   A string containing the requested number of cryptographically.
```

salathe: This would result in the following:

> A string containing the requested number of cryptographically. secure random bytes.

Surely, that's not intended.

cmb69: Oops, thank you!

zeriyoshi (approved):

reference/random/random/engine/mt19937/construct.xml
TimWolla: @Girgias I don't think the plural is appropriate here. In any case, the headlines are stolen from:

https://github.com/php/doc-en/blob/b412bbd26214f7281ac7dd858710e09952a275f2/reference/random/functions/mt-srand.xml#L45

Shall I:
- Create a new entity for the singular and change both.
- Change neither?

PR #1954, opened 2022-11-05
array_unshift page, add an example with two associative arrays

tiffany-taylor:
Added an example that shows how two associative arrays are combined using `array_unshift()`

https://www.php.net/manual/en/function.array-unshift.php

### DIFF
--- a/reference/array/functions/array-unshift.xml
+++ b/reference/array/functions/array-unshift.xml
@@ -104,6 +104,53 @@ Array
     </screen>
    </example>
   </para>
+  <para>
+   <example>
+    <title>Usage with multi-dimensional array</title>
+    <para>
+     If one multi-dimensional array is prepended to another multi-dimensional
+     array, the prepended array is numerically indexed into the former array.
+    </para>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$fruits = [
+     "apple" => "red",
+     "orange" => "orange",
+     "banana" => "yellow",
+     "raspberry" => "red",
+];
+$vegetables = [
+     "lettuce" => "green",
+     "cucumber" => "green",
+     "carrot" => "orange",
+];
+
+array_unshift($fruits, $vegetables);
+print_r($fruits);
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen role="php">
+<![CDATA[
+Array
+(
+    [0] => Array
+        (
+            [lettuce] => green
+            [cucumber] => green
+            [carrot] => orange
+        )
+
+    [apple] => red
+    [orange] => orange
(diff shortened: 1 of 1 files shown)


tiffany-taylor:
Gah, need to fix the spacing on `<![CDATA[`, didn't mean to alter that.

reference/array/functions/array-unshift.xml
```diff
@@ -100,6 +100,53 @@ Array
...
+     "raspberry" => "red",
+];
+$vegetables = [
+     "lettuce" => "green",
+     "cucumber" => "green",
+     "carrot" => "orange",
+];
+
+array_unshift($fruits, $vegetables);
+print_r($fruits);
```
Girgias, line 130:
```suggestion
var_dump($fruits);
```

And maybe also amend the existing example to use this. ``print_r()`` is kinda weird IMHO and to display a variable ``var_dump()`` is the better function.

tiffany-taylor:
Converting this to draft so that I can provide a better example.

cmb69:
I suggest not to think of "multi-dimensional" arrays in PHP; that eases the understanding how these are handled by language constructs and functions.

That said, I'm not against adding such an example for `array_unshift()`, since almost everybody talks about "multi-dimensional" PHP arrays.

tiffany-taylor:
TFW I realize I meant to call it "associative" and not "multi-dimensional" 😆

tiffany-taylor:
`check-whitespace` fails on a line that I fixed in the latest commit. It appears the job is looking at an older commit where the trailing whitespace was present. I'm not sure if the current failed job is a false-positive, or something I need to resolve.

cmb69:
@tiffany-taylor, to make the white-space-checker happy, you need to squash the commits, and force-push.

reference/array/functions/array-unshift.xml
```diff
@@ -84,22 +84,115 @@
...
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title>Usage with associative arrays</title>
+    <para>
+     If one associative array is prepended to another associative array,
+     the prepended array is numerically indexed into the former array.
+    </para>
```
Girgias, line 120: Add ``array_merge()`` to the See Also, and maybe a note somehow that maybe the function wanted is that one?

tiffany-taylor: I'm not sure I agree with that `array_merge` being the wanted one. I think there will be cases that people will want `array_unshift`, but this will likely be situations where people have a stronger grasp of the array_* functions.

Girgias (approved):

PR #1955, opened 2022-11-06
Fix ssh2_shell termtype parameter name

othercorey:
This is reflected as `termtype` - https://github.com/php/pecl-networking-ssh2/blob/master/ssh2.c#L1427.

Found while updating psalm function maps.


Girgias:
Thank you!

PR #1958, opened 2022-11-07
Remove unnecessary role attributes

kocsismate:


reference/datetime/functions/date-create.xml
```diff
@@ -8,7 +8,7 @@
 
  <refsect1 role="description">
   &reftitle.description;
-  <methodsynopsis role="procedural">
+  <methodsynopsis>
```
kocsismate, line 11: procedural roles are only needed when the method synopsis is in the same file with a class method synopsis.

PR #1959, opened 2022-11-07
Specify return type of DateTime::createFromImmutable() and DateTimeImmutable::createFromMutable()

kocsismate:


reference/datetime/datetime/createfromimmutable.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <modifier>static</modifier> <type>DateTime</type><methodname>DateTime::createFromImmutable</methodname>
+   <modifier>public</modifier> <modifier>static</modifier> <type>static</type><methodname>DateTime::createFromImmutable</methodname>
```
kocsismate, line 12: I'm not sure if/how the return type description below needs any clarification.

derickr:
I don't understand why you want to change this. It's more accurate I suppose, but it's much clearer for readers to know the exact type instead of having to then re-parse which class they're reading the docs for.

reference/datetime/datetime/createfromimmutable.xml
cmb69: In my opinion, the return type section doesn't need to be changed; after all, what is returned is an instance of `DateTime(Immutable)` (possibly a subtype, but still an instance of the class).

However, this needs a changelog entry, doesn't it?

cmb69:
> It's more accurate I suppose,

That. And accuracy looks like a good idea for technical documentation.

kocsismate:
> I don't understand why you want to change this. It's more accurate I suppose, but it's much clearer for readers to know the exact type instead of having to then re-parse which class they're reading the docs for.

I'm changing this because `static` is the actual return type of these two methods, so as I'm currently syncing the manual with the stubs, this change came up.

In my opinion, showing the correct return type is vital, it's sometimes very important to know if a new DateTime(immutable), or the actually invoked class is going to be returned. Also, I think we should keep the docs in sync with the reality so that future (maybe more important) changes are much-much easier to document.

Girgias (approved):

reference/datetime/datetime/createfromimmutable.xml
kocsismate: Oh yes! I missed that the behavior changed in PHP 8.0 via https://bugs.php.net/bug.php?id=60302

cmb69 (approved):
Thank you! The changelog entries are fine, as is the rest.

PR #1963, opened 2022-11-07
Mark up unspecific constants as replaceable

cmb69:
Great idea of @Girgias, in my opinion.

### DIFF
--- a/appendices/migration70/changed-functions.xml
+++ b/appendices/migration70/changed-functions.xml
@@ -46,7 +46,7 @@
    <listitem>
     <simpara>
      <function>setlocale</function> function no longer accepts <parameter>category</parameter>
-     passed as string. <constant>LC_*</constant> constants must be used instead.
+     passed as string. <constant><replaceable>LC_*</replaceable></constant> constants must be used instead.
     </simpara>
    </listitem>
    <listitem>
--- a/appendices/migration70/changed-functions.xml
+++ b/appendices/migration70/changed-functions.xml
@@ -46,7 +46,7 @@
    <listitem>
     <simpara>
      <function>setlocale</function> function no longer accepts <parameter>category</parameter>
-     passed as string. <constant><replaceable>LC_*</replaceable></constant> constants must be used instead.
+     passed as string. <constant>LC_<replaceable>*</replaceable></constant> constants must be used instead.
     </simpara>
    </listitem>
    <listitem>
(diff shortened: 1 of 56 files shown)


Girgias:
Does our DTD really not allow ``<constant>LC_<replaceable>*</replaceable></constant>``?

Otherwise, LGTM, but I would wait a bit before merging to not put too much strain on translation before the release of 8.2.

Girgias:
> Does our DTD really not allow `<constant>LC_<replaceable>*</replaceable></constant>`?
> 
> Otherwise, LGTM, but I would wait a bit before merging to not put too much strain on translation before the release of 8.2.

Now that I think about it, having the replaceable tag be just after the constant one will probably help if (one can dream) we get links to constants automatically generated as those will need to be ignored

cmb69:
> Does our DTD really not allow `<constant>LC_<replaceable>*</replaceable></constant>`?

Yeah, I got a failure when running `configure`, although the [constant spec](https://tdg.docbook.org/tdg/4.5/constant.html) says:

> Mixed Content Model
constant ::=
(#PCDATA|[replaceable](https://tdg.docbook.org/tdg/4.5/replaceable.html)|[inlinegraphic](https://tdg.docbook.org/tdg/4.5/inlinegraphic.html)|[inlinemediaobject](https://tdg.docbook.org/tdg/4.5/inlinemediaobject.html)|[indexterm](https://tdg.docbook.org/tdg/4.5/indexterm.html)|
 [beginpage](https://tdg.docbook.org/tdg/4.5/beginpage.html))*

> Now that I think about it, having the replaceable tag be just after the constant one will probably help if (one can dream) we get links to constants automatically generated as those will need to be ignored

Indeed, avoinding mixed content in `<constant>`s should make that easier.

cmb69:
> What was the failure? A quick search/replace to the `<constant>FOO_<replaceable>*</replaceable></constant>` format didn't result in a failure for me locally.

I've tried again, and indeed it works. I must have messed something up in my first attempt.

> I don't buy the argument that `<constant><replaceable>FOO_*</replaceable></constant>` should be (any meaningful amount) easier to work with than `<constant>FOO_<replaceable>*</replaceable></constant>` (e.g. for adding links, generating lists of constants, etc.).

Well, there isn't probably much of a difference, so I'm pushing a commit where only the `*`s are `<replaceable>`.

cmb69:
This has at least partially been done already, so I'm closing this PR.

PR #1964, opened 2022-11-07
Update ext/curl role attributes

kocsismate:
Based on https://github.com/php/php-src/pull/9901


kocsismate:
I've checked that the affected method synopsis elements are not referenced from anywhere else, and there are neither `oop`, nor `procedural` roles remaining in this extension.

PR #1967, opened 2022-11-08
Stop referring to `openssl_random_pseudo_bytes()` outside of ext/openssl

TimWolla:
The `random_bytes()` and `random_int()` alternatives are available since since PHP 7.0, are available by default and directly map to the OS' CSPRNG and thus are likely more secure.

Thus this PR stops mentioning `openssl_random_pseudo_bytes()` as a possible option any more to keep things simple for the reader.


cmb69 (approved):
Makes perfect sense. Thank you!

Girgias (approved):

PR #1968, opened 2022-11-08
random: Remove TypeError listitem from `&csprng.errors;`

TimWolla:
There are two reasons for this:

- It's standard functionality that a TypeError will be thrown for improper parameter types. This is nothing special for the CSPRNG functionality.
- The snippet is now also used for `Secure::generate()` which does not take any parameters.


cmb69 (approved):
Thank you!

PR #1969, opened 2022-11-08
Document the Random\Engine

TimWolla:
This is a small one, as properly explaining of how the `Engine::generate()` method is supposed to be used when implementing your own engine is complicated enough by itself.


cmb69 (commented):
Thank you! Besides the minor grammar issue, this looks good to me.

reference/random/random.engine.xml
```diff
@@ -9,7 +9,21 @@
...
+    whereas other algorithms are slower, but generate generate better randomness, up to
+    cryptographically secure randomness as provided by the <classname>Random\Engine\Secure</classname>
+    engine.
+   </para>
 
+   <para>
+    PHP provides several <classname>Random\Engine</classname>s out of the box to accomodate
+    different use-cases. The <classname>Random\Engine\Secure</classname> engine that is
+    backed by a <acronym>CSPRNG</acronym> is the recommended safe default choice, unless
+    the application requires either reproducible sequences or a very high performance.
```
Girgias, line 26:
```suggestion
    the application requires either reproducible sequences or very high performance.
```
Sounds weird otherwise

reference/random/random/engine/generate.xml
```diff
@@ -29,6 +45,40 @@
...
+    <simpara>
+     If generating randomness fails, a <classname>Random\RandomException</classname> should
+     be thrown. Any other <classname>Exception</classname> thrown during generation should
+     be caught and wrapped into a <classname>Random\RandomException</classname>.
+    </simpara>
+   </listitem>
+   <listitem>
+    <simpara>
+     If the returned string is empty, a <classname>Random\BrokenRandomEngineError</classname>
+     will be thrown by the <classname>Random\Randomizer</classname>.
```
Girgias, line 71:
```suggestion
     will be thrown by <classname>Random\Randomizer</classname>.
```
or
```suggestion
     will be thrown by the <classname>Random\Randomizer</classname> class.
```

Girgias (approved):
Other than nits, LGTM

reference/random/random.engine.xml
```diff
@@ -9,7 +9,21 @@
   <section xml:id="random-engine.intro">
    &reftitle.intro;
    <para>
+    An <classname>Random\Engine</classname> provides a low-level source of randomness by
```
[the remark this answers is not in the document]
TimWolla: When writing this I've implicitly dropped the `Random\` prefix in my head, thus this read to me as “An Engine provides …” where the previous version would've been correct :grin:

reference/random/random/engine/generate.xml
TimWolla: @Girgias I don't really like either of the suggestions. “The Randomizer” as the main interface to ext/random to me reads as a proper noun here, similarly to “the Internet”. Or “the Earth”.

I'm not a native speaker, though and I'm new to the documentation as well: Is there a strong preference for either suggestion? In that case it likely needs to be adjusted further up the file as well, whenever I've references the Randomizer.

cmb69: I've noticed this as well, and thought "it's okay". But if we want to change the wording, maybe use object or instance instead of class. Anyhow, I don't have a strong opinion on this.

zeriyoshi (approved):

reference/random/random/engine/generate.xml
Girgias: Ah yes, that's indeed OK, my brain didn't read the whole thing as a sentence just that specific bit and was "that doesn't look right"

Girgias (approved):

PR #1978, opened 2022-11-10
Update ext/oci8 role attributes

kocsismate:
Based on php/php-src#9901


cmb69 (approved):
Thank you! This is fine, but given that @Girgias recently mentioned [the current strain of translators](https://github.com/php/doc-en/pull/1963#issuecomment-1308861099), it might be best to postpone the merge for a while.

kocsismate:
Sure, I'll wait until PHP 8.2 is released :)

PR #1987, opened 2022-11-10
Update ext/session role attributes

kocsismate:
Based on https://github.com/php/php-src/pull/9901

### DIFF
--- a/reference/session/sessionhandler.xml
+++ b/reference/session/sessionhandler.xml
@@ -81,7 +81,7 @@
     </classsynopsisinfo>
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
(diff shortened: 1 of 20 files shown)


Girgias:
I'm fine with all the similar changes, but could they wait until after PHP 8.2 is released? Don't want to overwhelm translations with QA changes when more important stuff should be worked on.

kocsismate:
> I'm fine with all the similar changes, but could they wait until after PHP 8.2 is released? Don't want to overwhelm translations with QA changes when more important stuff should be worked on.

Yeah, Cristoph also suggested to wait (https://github.com/php/doc-en/pull/1978#pullrequestreview-1175451015) so I'll postpone the merge of these PRs!

cmb69 (commented):
> Can I start merging these PRs today, or should I wait a little bit more?

@Girgias, @saundefined, @mumumu, what do you think?

Girgias:
Eh, go for it, this shouldn't break the build. :)

PR #1988, opened 2022-11-10
Change simplesect to sect1 in exception chapter

Girgias:
So this conflict a tiny bit with my PR #320.

But this will fix the fact that the examples start their number at 3.
The reason for that is the numbering starts within the previous ``<sect1>`` but this is not rendered on the page, but in a follow-up page.
However, we cannot put this sections after the ``<simplesect>`` as that doesn't respect the DTD: https://tdg.docbook.org/tdg/5.0/chapter.html

Thus, we convert them to ``<sect1>``, this however makes them render into different pages (which might actually be better as it stops having so much on the same page).

@TimWolla this should be a helpful explanation :)

### DIFF
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<chapter xml:id="language.exceptions" xmlns="http://docbook.org/ns/docbook">
+<chapter xml:id="language.exceptions" xmlns:phd="http://www.php.net/ns/phd">
+ <!-- we use our own DTD instead of xmlns="http://docbook.org/ns/docbook" so that we can use
+ <sect1 phd:chunk="false"> -->
  <title>Exceptions</title>
  <para>
   PHP has an exception model similar to that of other programming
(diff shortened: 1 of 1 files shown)


language/exceptions.xml
```diff
@@ -1,215 +1,34 @@
...
+  function until it finds a matching &catch; block.  All &finally; blocks it encounters
+  along the way will be executed. If the call stack is unwound all the way to the
+  global scope without encountering a matching &catch; block, the program will
+  terminate with a fatal error unless a global exception handler has been set.
+ </para>
+ <para>
+  The thrown object must be an instance of the
+  <classname>Exception</classname> class or a subclass of
+  <classname>Exception</classname>. Trying to throw an object that is not
+  will result in a PHP Fatal Error.
```
TimWolla, line 24: I know you just moved this, but this should mention Throwable interface, instead of Exception class.

TimWolla (approved):
The result is not pretty and will require further cleaning, but I believe it is an improvement over the current status.

cmb69:
Ah, good find, @Girgias!

> Thus, we convert them to `<sect1>`, this however makes them render into different pages (which might actually be better as it stops having so much on the same page).

If we don't want to split the pages (and I tend to prefer that), we could use `<sect1 phd:chunk="false">` instead of `<sect1>` (that requires the `xmlns:phd="http://www.php.net/ns/phd"` namespace declaration).

PR #1990, opened 2022-11-10
Add predefined attribute section and classes

Girgias:
Rough work but this *should* work.

However I somehow now always hit some memory exhaustion problem when I try to render it although I managed to render it the first time round...


TimWolla (commented):
Something other that comes to my mind when seeing this: The documentation does not appear to support showing applied attributes yet. For the attribute documentation it would be useful to have an indication of the legal attribute targets within class definition.

cmb69:
Thank you for this! I didn't have any rendering issues, even after hooking the new section up in doc-base/manual.xml.in. :)

However, I don't think that language/predefined/attributes/version.xml can work. Instead, this info should be put in language/predefined/versions.xml, and the names of the classes need to be retained (the file names don't matter here).

Girgias:
> Thank you for this! I didn't have any rendering issues, even after hooking the new section up in doc-base/manual.xml.in. :)
> 
> However, I don't think that language/predefined/attributes/version.xml can work. Instead, this info should be put in language/predefined/versions.xml, and the names of the classes need to be retained (the file names don't matter here).

So that's indeed accurate, and I could start rendering again (might have been my RAM being full who knows) but even if PhD stops complaining it still doesn't want to render the version for me >_>

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -0,0 +1,62 @@
...
+ <titleabbrev>SensitiveParameter</titleabbrev>
+
+ <partintro>
+
+  <section xml:id="sensitive-parameter.intro">
+   &reftitle.intro;
+   <para>
+    <!-- TODO Link -->
+    This attribute is used to mark a parameter that is sensitive and should
+    have its value censored if present in a stack trace.
```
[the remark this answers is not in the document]
Girgias: Changed

language/predefined/attributes/allow-dynamic-properties.xml
```diff
@@ -0,0 +1,62 @@
...
+<phpdoc:classref xml:id="class.allow-dynamic-properties" xmlns:phpdoc="http://php.net/ns/phpdoc" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
+ <title>The AllowDynamicProperties class</title>
+ <titleabbrev>AllowDynamicProperties</titleabbrev>
+
+ <partintro>
+
+  <section xml:id="allow-dynamic-properties.intro">
+   &reftitle.intro;
+   <para>
+    <!-- TODO Link -->
```
cmb69, line 11: If there is no PR regarding a general description of dynamic properties, I think it's okay to leave this TODO comment in.

language/predefined/attributes/allow-dynamic-properties.xml
```diff
@@ -0,0 +1,62 @@
...
+    This attribute is used to mark classes that allow Dynamic properties.
```
cmb69, line 12: nit
```suggestion
    This attribute is used to mark classes that allow dynamic properties.
```

language/predefined/attributes/allow-dynamic-properties.xml
```diff
@@ -0,0 +1,62 @@
...
+    <classsynopsisinfo>
+     <ooclass>
+      <modifier>final</modifier>
+      <classname>AllowDynamicProperties</classname>
+     </ooclass>
+    </classsynopsisinfo>
+   </classsynopsis>
+
+  </section>
+
```
cmb69, line 34: Maybe add a small example of the behavior without and with the `AllowDynamicProperties` attribute.

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -0,0 +1,62 @@
...
+<phpdoc:classref xml:id="class.sensitive-parameter" xmlns:phpdoc="http://php.net/ns/phpdoc" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
+ <title>The SensitiveParameter class</title>
+ <titleabbrev>SensitiveParameter</titleabbrev>
+
+ <partintro>
+
+  <section xml:id="sensitive-parameter.intro">
+   &reftitle.intro;
+   <para>
+    <!-- TODO Link -->
```
cmb69, line 11: I don't think there is a better place to explain this attribute than on this page. Maybe use some word from the [RFC](https://wiki.php.net/rfc/redact_parameters_in_back_traces) (or maybe @TimWolla wants to write something). And we should also add documentation for the `SensitiveParameterValue` class here (separate page).

cmb69 (commented):
Could you please also add a page for `ReturnTypeWillChange`?

language/predefined/attributes/sensitive-parameter.xml
Girgias: I'd prefer letting @TimWolla flesh it out properly and have the current page be like a "stub".

cmb69 (approved):
Thank you!

Don't forget to hook it in doc-base.

language/predefined/attributes/sensitive-parameter.xml
TimWolla: @cmb69 The attribute should likely get a mention + example somewhere on the Exception page, for discoverability. That's why I complained about the “Example ordering” in the first place :smiley: Other than that, I can write the documentation for the attribute in a follow-up PR as time permits. For now I'll focus on ext/random as that one is likely less obvious than the SensitiveParameter attribute which was showcased sufficiently in various channels.

PR #1991, opened 2022-11-10
random: Add description to the jumping methods of Xoshiro256StarStar

TimWolla:


reference/random/random/engine/xoshiro256starstar/jump.xml
```diff
@@ -12,11 +12,16 @@
    <void />
   </methodsynopsis>
   <para>
-
+   Moves the algorithm’s state ahead by 2^128 steps, as if <function>Random\Engine\Xoshiro256StarStar::generate</function>
```
cmb69, line 15: I'm not quite sure about the `2^128` notation; are we using that elsewhere in the manual? The PHP equivalent would be `2**128` which *might* be more suitable.

reference/random/random/engine/xoshiro256starstar/jump.xml
```diff
@@ -12,11 +12,16 @@
...
+   was called 2^128 times.
   </para>
 
-  &warn.undocumented.func;
-
+  <para>
+   Jumping allows for splitting a single seeded <classname>Random\Engine\Xoshiro256StarStar</classname>
+   engine instance into 2^128 non-overlapping instances with 2^128 values each, by repeatedly
+   <link linkend="language.oop5.cloning">cloning</link> and then jumping an engine acting as
+   a blueprint.
```
cmb69, line 23: That "and then jumping an engine acting as a blueprint" sounds strange.

reference/random/random/engine/xoshiro256starstar/jumplong.xml
```diff
@@ -12,11 +12,22 @@
...
   <para>
-
+   Moves the algorithm’s state ahead by 2^192 steps, as if <function>Random\Engine\Xoshiro256StarStar::generate</function>
+   was called 2^192 times.
   </para>
 
-  &warn.undocumented.func;
+  <para>
+   Long jumping allows for splitting a single seeded <classname>Random\Engine\Xoshiro256StarStar</classname>
+   engine instance into 2^64 non-overlapping instances with 2^192 values each, by repeatedly
```
cmb69, line 21: Is that "2^64" correct here?

cmb69 (commented):
Thank you!

reference/random/random/engine/xoshiro256starstar/jump.xml
TimWolla: > are we using that elsewhere in the manual?

We are indeed:

https://www.php.net/manual/en/language.types.integer.php:

> If the float is beyond the boundaries of int (usually +/- 2.15e+9 = 2^31 on 32-bit platforms and +/- 9.22e+18 = 2^63 on 64-bit platforms)

https://www.php.net/manual/en/function.lcg-value

> The function combines two CGs with periods of 2^31 - 85 and 2^31 - 249.

--------

The `**` notation (`[0-9]\s*\*\*\s*[0-9]`) only appears in the luasandbox docs. The best solution would be a proper supertext (`<sup>`) feature. Is that available (I didn't test :smile: )?

reference/random/random/engine/xoshiro256starstar/jump.xml
TimWolla: It does indeed. I've rephrased this sentence multiple times and this is the best I came up with. Any suggestions?

reference/random/random/engine/xoshiro256starstar/jumplong.xml
TimWolla: Yes. Xoshiro256** has 256 bits of state and a period of `2^256 - 1`. By jumping ahead 2^192 steps, this period is split into 64 bits of "prefix" and 192 bits of "unique sequence for the prefix".

reference/random/random/engine/xoshiro256starstar/jump.xml
cmb69: Oh, indeed [DocBook 4.5 supports `<superscript>`](https://tdg.docbook.org/tdg/4.5/superscript.html), but I don't know if that is properly supported by PhD. I'll check later, if nobody beats me to it. :)

TimWolla: I've just tested it: It works. Perfect, I'll use that one.

cmb69: Actually quite funny that `<superscript>` is supported by PhD, but never used in the manual so far.

Girgias: Well, I suppose we have some follow up PRs to do (I can imagine this also be useful for the ``**`` operator and ``pow()`` functions pages)

reference/random/random/engine/xoshiro256starstar/jump.xml
Girgias, line 23:
```suggestion
   <link linkend="language.oop5.cloning">cloning</link> and then jumping the engine which acts as
   a blueprint.
```

Maybe? ``the`` because we are referring to the original blueprint instance

reference/random/random/engine/xoshiro256starstar/jump.xml
TimWolla: This one is now resolved.

reference/random/random/engine/xoshiro256starstar/jump.xml
```diff
@@ -2,7 +2,7 @@
 <refentry xml:id="random-engine-xoshiro256starstar.jump" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>Random\Engine\Xoshiro256StarStar::jump</refname>
-  <refpurpose>Efficiently moves the engine ahead by 2^128 steps</refpurpose>
+  <refpurpose>Efficiently move the engine ahead by 2^128 steps</refpurpose>
```
Girgias, line 5:
```suggestion
  <refpurpose>Efficiently move the engine ahead by 2<superscript>128</superscript> steps</refpurpose>
```

reference/random/random/engine/xoshiro256starstar/jumplong.xml
```diff
@@ -2,7 +2,7 @@
 <refentry xml:id="random-engine-xoshiro256starstar.jumplong" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>Random\Engine\Xoshiro256StarStar::jumpLong</refname>
-  <refpurpose>Efficiently moves the engine ahead by 2^192 steps</refpurpose>
+  <refpurpose>Efficiently move the engine ahead by 2^192 steps</refpurpose>
```
Girgias, line 5:
```suggestion
  <refpurpose>Efficiently move the engine ahead by 2<superscript>192</superscript>steps</refpurpose>
```

reference/random/random/engine/xoshiro256starstar/jump.xml
TimWolla: No, the comma is definitely not correct. What I intent to explain here is the “pattern” that is also used for the ext/random example on the PHP 8.2 release page: https://github.com/php/web-php/pull/675#pullrequestreview-1160537517

Basically the following:

```
$blueprint = new Xoshiro256StarStar();

while (($request = $server->accept())) {
  $requestEngine = clone $blueprint; // Clone the blueprint.
  $blueprint->jump(); // Jump the blueprint to ensure $requestEngine has 2^128 non-overlapping values.
  // Now the request may use the $requestEngine without worrying about reused numbers
}
```

Girgias (changes requested):
Refpurpose also need to use the superscript tag

reference/random/random/engine/xoshiro256starstar/jump.xml
TimWolla: This does not render properly. It will show up as `2128` in the method lists.

Girgias: AAAAAAAAAAAAAAAAAAAA gosh do I hate PhD at times... okay leave it then

cmb69: Indeed, basically PhD strips all tags here (see https://github.com/php/phd/blob/master/phpdotnet/phd/Reader.php). So I agree that we should stick with what we have here.

zeriyoshi (approved):

reference/random/random/engine/xoshiro256starstar/jump.xml
```diff
@@ -12,11 +12,17 @@
...
+   2<superscript>128</superscript> times.
   </para>
 
-  &warn.undocumented.func;
-
+  <para>
+   Jumping allows for splitting a single seeded <classname>Random\Engine\Xoshiro256StarStar</classname>
+   engine instance into 2<superscript>128</superscript> non-overlapping instances with 2<superscript>128</superscript>
+   values each, by repeatedly <link linkend="language.oop5.cloning">cloning</link> and
+   then jumping an engine acting as a blueprint.
```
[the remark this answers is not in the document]
TimWolla: @Girgias Thank you. I've taken your suggestion with some rephrasing (“values” → “sequences” and adding some extra information (2^128 possible sequences)). PTAL :smiley:

cmb69 (approved):
Thank you!

Girgias (approved):

zeriyoshi (approved):

PR #1992, opened 2022-11-10
random: Add descriptions to Randomizer::shuffleArray() and Randomizer::shuffleBytes()

TimWolla:


reference/random/random/randomizer/shufflearray.xml
```diff
@@ -12,11 +12,17 @@
...
   </para>
 
-  &warn.undocumented.func;
+  <para>
+   Each possible permutation of the input <parameter>array</parameter> is equally likely to be returned.
+  </para>
 
+  <para>
+   The input <parameter>array</parameter> will not be modified. Array keys of the input <parameter>array</parameter>
+   will not be preserved; the returned &array; will be a list (<function>array_is_list</function>).
```
cmb69, line 24: Shouldn't we move the second sentence to the return-values section?

cmb69 (approved):
Thank you!

reference/random/random/randomizer/shufflearray.xml
Girgias: I agree that it makes more sense to have the second sentence in the return value section, and maybe the first one in the parameter description

TimWolla: Okay, updated, please have another look. I've also fixed the texts properly refer to `shuffleString()`'s parameter as `$bytes`, not `$string`.

cmb69 (approved):
Thank you!

Girgias (approved):

zeriyoshi (approved):

PR #1995, opened 2022-11-11
Update trait docs for PHP-8.2

Girgias:
The docs for traits probably need a good overhaul.

@sj-i could you have a look as you wrote the RFC.

### DIFF
--- a/language/oop5/traits.xml
+++ b/language/oop5/traits.xml
@@ -466,8 +466,8 @@ $example->x;
    </example>
    <para>
     If a trait defines a property then a class can not define a property with
-    the same name unless it is compatible (same visibility and initial value),
-    otherwise a fatal error is issued.
+    the same name unless it is compatible (same visibility and type,
+    readonly marker, and initial value), otherwise a fatal error is issued.
    </para>
    <example xml:id="language.oop5.traits.properties.conflicts">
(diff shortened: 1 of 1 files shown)


language/oop5/traits.xml
```diff
@@ -476,20 +476,74 @@ $example->x;
...
+    readonly public bool $different3; // Fatal error
 }
 ?>
 ]]>
     </programlisting>
    </example>
   </sect2>
 
+ <sect2 xml:id="language.oop5.traits.constants">
+  <title>Constants</title>
```
Girgias, line 498:
```suggestion
  <title>&Constants;</title>
```
Just realized I think we have an entity for this.

language/oop5/traits.xml
```diff
@@ -466,8 +466,8 @@ $example->x;
    </example>
    <para>
     If a trait defines a property then a class can not define a property with
-    the same name unless it is compatible (same visibility and initial value),
-    otherwise a fatal error is issued.
+    the same name unless it is compatible (same visibility and type,
+    readonly marker, and initial value), otherwise a fatal error is issued.
```
Girgias, line 470:
```suggestion
    readonly modifier, and initial value), otherwise a fatal error is issued.
```

Issue #1907, opened 2022-10-21
Trailing comma in function calls not documented


wb027:
Upd.
The bug appeared for a different reason.

_From manual page: https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list
---
I did tests on versions:
7.2 -> PHP Parse error:  syntax error, unexpected ')'
7.4 -> it works

damianwadley:
I don't know what code you tried but [examples 11](https://3v4l.org/j6hjm) and [12](https://3v4l.org/fcWdK) work since at least PHP 7.0, which is the oldest version of PHP that the manual is intended to cover.

wb027:
Oops! The bug appeared for a different reason. Now I will create the correct bug report.

cmb69:
> 7.2 -> PHP Parse error: syntax error, unexpected ')'

This is due to https://wiki.php.net/rfc/trailing-comma-function-calls; only available as of PHP 7.3.0, and apparently not documented. I could only find the [trailing comma in function signatures](https://www.php.net/manual/en/functions.arguments.php), but not in calls.

PR #1999, opened 2022-11-12
Rewrite part of function syntax docs

Girgias:
This includes documenting that function calls allow a trailing comma as of PHP 7.3.0, which fixes GH-1907


language/functions.xml
```diff
@@ -7,12 +7,14 @@
...
+    a name, a list of parameters (which might be empty) seperated by commas
+    (<literal>,</literal>) enclosed in parentheses, followed by the body of
+    the function enclosed in curly braces, such as the following:
    </para>
-   <para>
-    <example>
-     <title>Pseudo code to demonstrate function uses</title>
-     <programlisting role="php">
+   <example>
+    <title>Pseudo code to demonstrate function uses</title>
```
cmb69, line 16: Why "pseudo code"? [This](https://3v4l.org/BkAXO) looks like valid PHP code.

language/functions.xml
```diff
@@ -188,21 +204,24 @@ function recursion($a)
...
-    </example>
-   </para>
+      </programlisting>
+     </informalexample>
+    </para>
+   </note>
+
    <para>
     As of PHP 8.0.0, the list of function arguments may include a trailing comma, which
     will be ignored.  That is particularly useful in cases where the list of arguments is
```
cmb69, line 227: This is slightly confusing. Maybe mark up "function call" above as `<emphasis>`; or maybe add "function definition" to this paragraph. Or maybe leave as is, because that whole chapter is pretty messy anyway.

cmb69 (commented):
Thank you!

language/functions.xml
Girgias: Indeed, don't know why it was written that way in the first place

language/functions.xml
Girgias: Yeah this whole chapter needs a rewrite, I'll put the emphasis and will start working on that.

PR #2000, opened 2022-11-12
Rewrite and flesh out string interpolation docs

Girgias:
This covers the PHP 8.2.0 deprecation of interpolation It also renames simple to basic and complex to advanced as those don't make a judgement on complexity.

### DIFF
--- a/language/types/string.xml
+++ b/language/types/string.xml
@@ -199,7 +199,7 @@ echo 'Variables do not $expand $either';
    <para>
     The most important feature of double-quoted <type>string</type>s is the fact
     that variable names will be expanded. See
-    <link linkend="language.types.string.parsing">string parsing</link> for
+    <link linkend="language.types.string.parsing">string interpolation</link> for
     details.
    </para>
   </sect3>
@@ -592,7 +592,7 @@ FOOBAR;
    <para>
     Nowdocs are to single-quoted strings what heredocs are to double-quoted
     strings. A nowdoc is specified similarly to a heredoc, but <emphasis>no
-    parsing is done</emphasis> inside a nowdoc. The construct is ideal for
+    String interpolation is done</emphasis> inside a nowdoc. The construct is ideal for
     embedding PHP code or other large blocks of text without the need for
     escaping. It shares some features in common with the SGML
     <literal>&lt;![CDATA[ ]]&gt;</literal> construct, in that it declares a
@@ -684,7 +684,7 @@ EOT;
   </sect3>
 
   <sect3 xml:id="language.types.string.parsing">
-   <title>Variable substitution</title>
+   <title>String interpolation</title>
 
    <simpara>
     When a <type>string</type> is specified in double quotes or with heredoc,
@@ -907,7 +907,7 @@ Changing the character at index -3 to o gives strong.
 
     <simpara>
      The advanced syntax permits the interpolation of
-     <emphasis>variables</emphasis> with arbitrary access expressions.
+     <emphasis>variables</emphasis> with arbitrary accessors.
     </simpara>
 
     <simpara>
@@ -945,8 +945,16 @@ echo "This is { $great}";
 // Works, outputs: This is fantastic
 echo "This is {$great}";
 
+class Square {
+    public $width;
+
+    public function __construct(int $width) { $this->width = $width; }
+}
+
+$square = new Square(5);
+
 // Works
-echo "This square is {$square->width}00 centimeters broad.";
+echo "This square is {$square->width}00 centimeters wide.";
 
 
 // Works, quoted keys only work using the curly brace syntax
--- a/language/types/string.xml
+++ b/language/types/string.xml
@@ -693,28 +693,25 @@ EOT;
 
    <simpara>
     There are two types of syntax: a
-    <link linkend="language.types.string.parsing.simple">simple</link> one and a
-    <link linkend="language.types.string.parsing.complex">complex</link> one.
-    The simple syntax is the most common and convenient. It provides a way to
+    <link linkend="language.types.string.parsing.simple">basic</link> one and an
+    <link linkend="language.types.string.parsing.complex">advanced</link> one.
+    The basic syntax is the most common and convenient. It provides a way to
     embed a variable, an <type>array</type> value, or an <type>object</type>
     property in a <type>string</type> with a minimum of effort.
    </simpara>
 
    <simpara>
-    The complex syntax can be recognised by the
+    The advanced syntax can be recognised by the
     curly braces surrounding the expression.
    </simpara>
 
    <sect4 xml:id="language.types.string.parsing.simple">
-    <title>Simple syntax</title>
+    <title>Basic syntax</title>
 
     <simpara>
      If a dollar sign (<literal>$</literal>) is encountered, the parser will
      greedily take as many tokens as possible to form a valid variable name.
-     Enclose the variable name in curly braces to explicitly specify the end of
-     the name.
     </simpara>
-
     <informalexample>
      <programlisting role="php">
 <![CDATA[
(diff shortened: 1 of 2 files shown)


language/types/string.xml
```diff
@@ -692,46 +692,140 @@ EOT;
 
    <simpara>
     There are two types of syntax: a
-    <link linkend="language.types.string.parsing.simple">simple</link> one and a
-    <link linkend="language.types.string.parsing.complex">complex</link> one.
-    The simple syntax is the most common and convenient. It provides a way to
+    <link linkend="language.types.string.parsing.simple">basic</link> one and an
```
iluuu1994, line 695: The pages are still named simple and complex, should those be renamed? I'm also not sure this needs to be renamed at all.

language/types/string.xml
```diff
@@ -692,46 +692,140 @@ EOT;
...
+       </screen>
+       &example.outputs;
+       <screen>
+<![CDATA[
+string(3) "foo"
+string(3) "bar"
+]]>
+       </screen>
+      </informalexample>
+      The <link linkend="language.types.string.parsing.complex">advanced</link>
```
iluuu1994, line 797: The basic one would be fine too for the first example.

language/types/string.xml
```diff
@@ -692,46 +692,140 @@ EOT;
...
    </simpara>
 
    <sect4 xml:id="language.types.string.parsing.simple">
-    <title>Simple syntax</title>
+    <title>Basic syntax</title>
 
     <simpara>
      If a dollar sign (<literal>$</literal>) is encountered, the parser will
      greedily take as many tokens as possible to form a valid variable name.
-     Enclose the variable name in curly braces to explicitly specify the end of
```
iluuu1994, line 713: This comment seems kind of useful. Maybe move it to the advanced section instead?

language/types/string.xml
```diff
@@ -692,46 +692,140 @@ EOT;
...
     curly braces surrounding the expression.
    </simpara>
 
    <sect4 xml:id="language.types.string.parsing.simple">
-    <title>Simple syntax</title>
+    <title>Basic syntax</title>
 
     <simpara>
      If a dollar sign (<literal>$</literal>) is encountered, the parser will
      greedily take as many tokens as possible to form a valid variable name.
```
iluuu1994, line 712: I find that statement a little misleading. The syntax isn't greedy, e.g. only one array dimension is allowed. https://3v4l.org/pJF8E

iluuu1994 (approved):
Only minor things, looks good to me overall.

language/types/string.xml
Girgias: I didn't want to change them in case other pages link to here. But I could check and change them

language/types/string.xml
Girgias: Well brilliant, this is already the existing docs but might as well fix them now. Do you have better wording? I can try to come up with one otherwise.

language/types/string.xml
Girgias: How would code using the advanced syntax use this? And isn't that part of the deprecation?

language/types/string.xml
Girgias: Well I was trying to have an example to demonstrate the reason why it is deprecated, so I copied the RFC example, but if you have a better example I'll take it :)

language/types/string.xml
iluuu1994: Maybe something like:

> The basic syntax only allows for single-dimensional array or property accesses.
>
> ```php
> $foods = ['apples', 'bananas', 'lettuce', 'spinach'];
> echo "$foods[0]\n";
> // apples
> 
> $foods = ['fruits' => ['apples', 'bananas'], 'vegetables' => ['lettuce', 'spinach']];
> echo "$foods[fruits][0]\n";
> // Warning: Array to string conversion in /in/hcu76 on line 7
> // Array[0]
> ```
> To use multi-dimensional array or property accesses, use the advanced syntax.

language/types/string.xml
iluuu1994: The error message only mentions `{$var}` because that will work for all versions of `${var}`, like `${var['foo']}`. That's not valid in the basic syntax, where the `'` need to be omitted: `$var[foo]`. When only simple variables are used, `$var` should be fine as well.

language/types/string.xml
iluuu1994: What this comment is saying that by enclosing the whole expression with `{}`, the parser doesn't need to guess when the expression is done (or rather stop after the first dimension) but will look for the ending `}` instead.

Girgias: > What this comment is saying that by enclosing the whole expression with `{$...}`, the parser doesn't need to guess when the expression is done (or rather stop after the first dimension) but will look for the ending `}` instead.

No, this comment is referring to the ``${...}`` syntax. If you read the current docs that's how it's used.

language/types/string.xml
Girgias: > I find that statement a little misleading. The syntax isn't greedy, e.g. only one array dimension is allowed. https://3v4l.org/pJF8E

Actually this statement is only about the name of a variable, not any following dimension. So I think it's correct.

cmb69: I'm not sure what that statement is really about (there are still places in the docs where it talks about "variables" but means "values", what I consider grossly wrong). However, if it is only about the identifier, the word "token" doesn't make sense.

iluuu1994: Also, braces don't change how the variable name itself is tokenized, so it's not even correct in that context.

Girgias:
I spent some more time again, can you please have another look @iluuu1994 @cmb69 ?

language/types/string.xml
```diff
@@ -861,112 +1057,28 @@ echo "This works: {$arr['key']}";
...
+echo "This works: {$obj->values[3]->name}";
 
-// Won't work, outputs: This is the return value of getName(): {getName()}
-echo "This is the return value of getName(): {getName()}";
+echo "This works: {$obj->$staticProp}";
 
-// Won't work, outputs: C:\folder\{fantastic}.txt
-echo "C:\folder\{$great}.txt"
-// Works, outputs: C:\folder\fantastic.txt
-echo "C:\\folder\\{$great}.txt"
```
cmb69, line 891: This was a fix for https://bugs.php.net/81635. Not sure if we should just drop that.

cmb69 (commented):
Thank you for the PR!

I'm not convinced that this is a major improvement (but certainly causes quite some work for translators). Wouldn't it be better to have a precise description (like https://github.com/php/php-langspec/blob/master/spec/09-lexical-structure.md#string-literals)?

language/types/string.xml
Girgias: Ah, okay I'll readd it then

language/types/string.xml
```diff
@@ -778,13 +859,44 @@ echo "$people->robert greeted the two $people->smiths."; // Won't work
...
-John Smith's wife greeted Robert Paulsen.
-Robert Paulsen greeted the two .
 ]]>
      </screen>
     </example>
+    
+    <warning>
+     <para>
+      The basic syntax does not permit access to an array entry which has a key
+      containing a minus (<literal>-</literal>) sign.
```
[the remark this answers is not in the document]
Girgias: There are proably others yes, but when I was creating my example I used the array key ``string-key`` which when I tried to interpolate didn't work. So I changed the ``-`` to an ``_``, however refering to that key is possible via the normal semantics.

Girgias:
> Thank you for the PR!
> 
> I'm not convinced that this is a major improvement (but certainly causes quite some work for translators). Wouldn't it be better to have a precise description (like https://github.com/php/php-langspec/blob/master/spec/09-lexical-structure.md#string-literals)?

Having a more precise description would be best, but even that page has a dead link to what it refers to as "simple expression", and we would need to define what a name is somewhere.

Which might be a good idea to have a section on basic language components, because the amount of pages where we redeclare what a label is in PHP is too many.

But this is indeed probably the best course, I'll have another go at it.

language/types/string.xml
```diff
@@ -683,92 +683,167 @@ EOT;
   </sect3>
 
   <sect3 xml:id="language.types.string.parsing">
-   <title>Variable parsing</title>
+   <title>Variable substitution</title>
```
cmb69, line 686: I think using "substition" instead of "parsing" is an improvement (but there are 3 occurrences of "parsing" left on that page), but I think "variable substitution" is still not correct; at least with the advanced syntax, it is possible to interpolate the results of evaluating expressions. Speaking of "expression substitution/interpolation" would be too general, though. Not sure what to do here.

language/types/string.xml
```diff
@@ -812,20 +893,23 @@ Changing the character at index -3 to o gives strong.
...
-     This isn't called complex because the syntax is complex, but because it
-     allows for the use of complex expressions.
+     The advanced syntax permits the interpolation of
+     <emphasis>variables</emphasis> with arbitrary access expressions.
     </simpara>
 
     <simpara>
-     Any scalar variable, array element or object property with a
+     Any scalar variable, array element or object property
+     (<modifier>static</modifier> or not) with a
```
cmb69, line 912: TIL

language/types/string.xml
```diff
@@ -861,112 +953,32 @@ echo "This works: {$arr['key']}";
 // Works
 echo "This works: {$arr[4][3]}";
 
-// This is wrong for the same reason as $foo[bar] is wrong  outside a string.
```
cmb69, line 864: There is an example above:
````php
// Works
echo "This square is {$square->width}00 centimeters broad.";
````

`$square` is not defined, though. Might be worth fixing that.

File: language/types/string.xml [hunk 10]
```diff
@@ -812,20 +893,23 @@ Changing the character at index -3 to o gives strong.
...
-   <sect4 xml:id="language.types.string.parsing.complex">
-    <title>Complex (curly) syntax</title>
+   <sect4 xml:id="language.types.string.parsing.advanced">
+    <title>Advanced (curly) syntax</title>
 
     <simpara>
-     This isn't called complex because the syntax is complex, but because it
-     allows for the use of complex expressions.
+     The advanced syntax permits the interpolation of
+     <emphasis>variables</emphasis> with arbitrary access expressions.
```
cmb69, line 907: Not sure about "access expressions". Do we define that term somewhere? If not, it might be sensible to do that. Or maybe use "accessors"?

language/types/string.xml
Girgias: I'm using [string interpolation](https://en.wikipedia.org/wiki/String_interpolation) now

Girgias:
Note to self: related to #2078

language/types/string.xml
jimwins: This section seems clear to me as @Girgias has changed it, enclosing just the variable name in brackets (`${....}`) was deprecated, but still explained in the warning that comes a bit later.

jimwins:
The conflict is just in an example that should be easy to resolve. Otherwise, this is the last PHP 8.2 documentation issue still open on the tracker.

jimwins (approved):
LGTM.

PR #2003, opened 2022-11-13
Add precision to same method name in many interfaces

l-vo:


language/oop5/interfaces.xml
```diff
@@ -186,7 +187,34 @@ class D implements B
 ]]>
      </programlisting>
    </example>
-   <example xml:id="language.oop5.interfaces.examples.ex3">
+      <example xml:id="language.oop5.interfaces.examples.ex3">
+    <title>Multiple interfaces with covariance and contravariance</title>
+     <programlisting role="php">
```
Girgias, line 192:
```suggestion
    <title>Variance compatibility with multiple interfaces</title>
     <programlisting role="php">
```

language/oop5/interfaces.xml
```diff
@@ -186,7 +187,34 @@ class D implements B
 ]]>
      </programlisting>
    </example>
-   <example xml:id="language.oop5.interfaces.examples.ex3">
+      <example xml:id="language.oop5.interfaces.examples.ex3">
```
Girgias, line 190: Nit: indentation

language/oop5/interfaces.xml
```diff
@@ -59,7 +59,8 @@
    <warning>
     <para>
      A class can implement two interfaces which define a method with the
-     same name, only if the method declaration in both interfaces is identical.
+     same name. In this case, the implementations must be compatible with all
+     the interfaces. <link linkend="language.oop5.variance">Covariance and contravariance</link> can be applied.
```
Girgias, line 63: Rewriting the whole paragraph to include multiple instead of 2 makes sense as well.

```suggestion
     same name. In this case, the implementations and signature must be
     <link linkend="language.oop.lsp">compatible</link> with every interface.
```
We've got a better (IMHO) paragraph about signature compatibility here: https://www.php.net/manual/en/language.oop5.basic.php#language.oop.lsp

language/oop5/interfaces.xml
```diff
@@ -224,7 +252,7 @@ class D implements C
 ]]>
      </programlisting>
    </example>
-   <example xml:id="language.oop5.interfaces.examples.ex4">
+   <example xml:id="language.oop5.interfaces.examples.ex5">
```
Girgias, line 255: Not sure changing the ID is a good idea, I don't know if other pages reference these would need to check.

Girgias (commented):
Thanks for fixing this!

language/oop5/interfaces.xml
l-vo: Fixed, thanks

language/oop5/interfaces.xml
l-vo: Reverted

l-vo:
@Girgias Thank you for your review. I fixed some thinks and answered some others.

language/oop5/interfaces.xml
```diff
@@ -59,7 +59,8 @@
    <warning>
     <para>
      A class can implement two interfaces which define a method with the
-     same name, only if the method declaration in both interfaces is identical.
+     same name. In this case, the implementations must be compatible with all
+     the interfaces. <link linkend="language.oop.lsp">Covariance and contravariance</link> can be applied.
```
salathe, line 63: This link (`language.oop.lsp`) points to a section titled "Signature compatibility rules", of which variance is only a part.  Suggest either changing the link text to match, or changing the link to point to the variance page (`language.oop5.variance`), whichever you think most appropriate.

Also, there is a separate "note" saying:

> **Note:** The class implementing the interface must declare all methods in the interface with a [compatible signature](https://www.php.net/manual/en/language.oop5.basic.php#language.oop.lsp).

Which basically reiterates what you've added.  Maybe, as @cmb also suggested, merge these two?

cmb69, line 63: A bit later, there is a note: "The class implementing the interface must declare all methods in the interface with a [compatible signature](https://www.php.net/manual/en/language.oop5.basic.php#language.oop.lsp)." Maybe we should merge these descriptions (frankly, I don't understand why we currently have that order, and why one is a warning and the other a note).

language/oop5/interfaces.xml
```diff
@@ -183,6 +184,33 @@ class D implements B
     }
 }
 ?>
+]]>
+     </programlisting>
+   </example>
+   <example xml:id="language.oop5.interfaces.examples.ex7">
```
salathe, line 190: I know you're just following along with the `id`s from other examples in this page, but it would be nice to use a more descriptive `id` (e.g. ending with `variance`, or some other word or short phrase, instead of `ex7`).

salathe (commented):
Added a few notes.

Also, not directly related to the changes here but, this page would really benefit from a basic example of a class implementing multiple interfaces without any interface inheritance (existing example) or compatibility concerns (the example added in this PR).

language/oop5/interfaces.xml
l-vo: Updated accorded to https://github.com/php/doc-en/pull/2003#discussion_r1030438597

language/oop5/interfaces.xml
l-vo: I merged the new paragraph in the note section, thank you :)

l-vo: Updated, I added the two links 🙂. Thank you.

language/oop5/interfaces.xml
l-vo: Done :)

language/oop5/interfaces.xml
```diff
@@ -79,7 +73,10 @@
    <note>
     <para>
      The class implementing the interface must declare all methods in the interface
-     with a <link linkend="language.oop.lsp">compatible signature</link>.
+     with a <link linkend="language.oop.lsp">compatible signature</link>.A class can implement two interfaces
```
cmb69, line 76:
```suggestion
     with a <link linkend="language.oop.lsp">compatible signature</link>. A class can implement multiple interfaces
```

language/oop5/interfaces.xml
```diff
@@ -79,7 +73,10 @@
...
+     which define a method with the same name. In this case, the implementations must follow the
```
cmb69, line 77:
```suggestion
     which declare a method with the same name. In this case, the implementation must follow the
```

language/oop5/interfaces.xml
```diff
@@ -79,7 +73,10 @@
...
+     <link linkend="language.oop.lsp">signature compatibility rules</link> with all the interfaces. So
```
cmb69, line 78: Not sure if the preposition should be "with" or "for":
```suggestion
     <link linkend="language.oop.lsp">signature compatibility rules</link> for all the interfaces. So
```

l-vo:
@cmb69 suggestions applied, thank you :)

cmb69:
Thank you!

PR #2005, opened 2022-11-14
added SensitiveParameter usage example.

mumumu:
based on RFC: https://wiki.php.net/rfc/redact_parameters_in_back_traces

### DIFF
--- a/language/predefined/attributes/sensitive-parameter.xml
+++ b/language/predefined/attributes/sensitive-parameter.xml
@@ -44,7 +44,7 @@ function defaultBehavior(
     throw new Exception('Error!');
 }
 
-function sensitiveParameterTest(
+function sensitiveParametersWithAttribute(
     #[\SensitiveParameter]
     string $secret
 ) {
@@ -54,13 +54,13 @@ function sensitiveParameterTest(
 try {
     defaultBehavior('password');
 } catch (Exception $e) {
-    var_dump($e->getTrace());
+    echo $e->getTraceAsString(), PHP_EOL;
 }
 
 try {
-    sensitiveParameterTest('password');
+    sensitiveParametersWithAttribute('password');
 } catch (Exception $e) {
-    var_dump($e->getTrace());
+    echo $e->getTraceAsString(), PHP_EOL;
 }
 
 ?>
@@ -69,39 +69,10 @@ try {
     &example.outputs.82.similar;
     <screen>
 <![CDATA[
-array(1) {
-  [0]=>
-  array(4) {
-    ["file"]=>
-    string(9) "example.php"
-    ["line"]=>
-    int(16)
-    ["function"]=>
-    string(15) "defaultBehavior"
-    ["args"]=>
-    array(1) {
-      [0]=>
-      string(8) "password"
-    }
-  }
-}
-array(1) {
-  [0]=>
-  array(4) {
-    ["file"]=>
-    string(9) "example.php"
-    ["line"]=>
-    int(22)
-    ["function"]=>
-    string(22) "sensitiveParameterTest"
-    ["args"]=>
-    array(1) {
-      [0]=>
-      object(SensitiveParameterValue)#3 (0) {
-      }
-    }
-  }
-}
+#0 example.php(16): defaultBehavior('password')
+#1 {main}
+#0 example.php(22): sensitiveParametersWithAttribute(Object(SensitiveParameterValue))
+#1 {main}
 ]]>
     </screen>
    </example>
--- a/language/predefined/attributes/sensitive-parameter.xml
+++ b/language/predefined/attributes/sensitive-parameter.xml
@@ -38,31 +38,70 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-function defaultBehavior(string $secret) {
-    debug_print_backtrace();
-}
-
-function sensitiveParameterTest(
-    #[\SensitiveParameter] string $secret
+function defaultBehavior(
+    string $secret
 ) {
     throw new Exception('Error!');
 }
 
-defaultBehavior('password');
-sensitiveParameterTest('password');
+function sensitiveParameterTest(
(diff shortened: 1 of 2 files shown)


language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,42 @@
 
   </section>
 
+  <section>
+   &reftitle.examples;
+   <example>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function defaultBehaviour(string $secret) {
```
cmb69, line 41: I think we prefer American English spelling, so:
```suggestion
function defaultBehavior(string $secret) {
```
The same change needs to be applied for the function call below.

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,42 @@
...
+) {
+    throw new Exception('Error!');
+}
+
+defaultBehaviour('password');
+sensitiveParameterTest('password');
+?>
+]]>
+    </programlisting>
+    &example.outputs.82;
```
cmb69, line 56: Since we need to display the backtrace, but the filename and line numbers are not universal (and there is no particular need to state the PHP version here), I'd use:
```suggestion
    &example.outputs.similar;
```

cmb69 (commented):
Thank you!

cc @TimWolla

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,42 @@
...
+<![CDATA[
+<?php
+function defaultBehaviour(string $secret) {
+    debug_print_backtrace();
+}
+
+function sensitiveParameterTest(
+    #[\SensitiveParameter] string $secret
+) {
+    throw new Exception('Error!');
```
TimWolla, line 48: Unify both to Exception or both to `debug_print_backtrace()`. I prefer Exception (because that is the main use case). Thus a try-catch should be added around each function call.

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,42 @@
...
+
+defaultBehaviour('password');
+sensitiveParameterTest('password');
+?>
+]]>
+    </programlisting>
+    &example.outputs.82;
+    <screen>
+<![CDATA[
+#0 /in/g7BEP(12): defaultBehaviour('password')
```
TimWolla: The 3v4l paths are pretty ugly. I suggest to rerun this as "example.php", it looks a little nicer.

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,42 @@
...
+   &reftitle.examples;
+   <example>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function defaultBehavior(string $secret) {
+    debug_print_backtrace();
+}
+
+function sensitiveParameterTest(
```
TimWolla, line 45:
```suggestion
function sensitiveParametersWithAttribute(
```

sounds a little nicer and highlights the difference using the function name.

mumumu:
@cmb69 thank you for your suggestions! Changes applied.

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,42 @@
 
   </section>
 
+  <section>
+   &reftitle.examples;
+   <example>
+    <programlisting role="php">
+<![CDATA[
+<?php
+function defaultBehavior(string $secret) {
```
TimWolla, line 41: Maybe also split this one across multiple lines so that the functions look as similar as possible (highlighting the difference better):
```suggestion
function defaultBehavior(
    string $secret
) {
```

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,81 @@
...
+    #[\SensitiveParameter]
+    string $secret
+) {
+    throw new Exception('Error!');
+}
+
+try {
+    defaultBehavior('password');
+} catch (Exception $e) {
+    var_dump($e->getTrace());
```
TimWolla, line 57:
```suggestion
    echo $e->getTrace(), PHP_EOL;
```

This matches the output of an unhandled Exception.

language/predefined/attributes/sensitive-parameter.xml
```diff
@@ -32,6 +32,81 @@
...
+}
+
+try {
+    sensitiveParameterTest('password');
+} catch (Exception $e) {
+    var_dump($e->getTrace());
```
TimWolla, line 63:
```suggestion
    echo $e->getTrace(), PHP_EOL;
```

language/predefined/attributes/sensitive-parameter.xml
TimWolla: I've unresolved this: There is no change to the function name. If you are unhappy with my suggestion, please leave a comment.

TimWolla (commented):
I've also wondering if a non-sensitive parameter should be added to both functions to showcase that the attribute applies for *each* parameter.

mumumu:
> I've also wondering if a non-sensitive parameter should be added to both functions to showcase that the attribute applies for _each_ parameter.

I think that it's easy to understand to add non-sensitive parameter, because we already split sensitive parameter into two lines.

@TimWolla, do you like the following code and output?

https://3v4l.org/aCY0b#v8.2rc3

TimWolla:
> do you like the following code and output?

@mumumu Please use this variant: https://3v4l.org/e5P7O. The difference is that I use `echo $e;`, instead of `$e->getTraceAsString()`. This also shows the Exception message and thus is a little more complete. Apart from that, it was looking good and nicely showcases that the attribute needs to be applied to each parameter.

Apologies for my confusing comments earlier! I probably should have directly sent a complete example, instead of just pointing out the necessary changes.

PR #2007, opened 2022-11-14
random: Improve short descriptions for methods

TimWolla:
I intentionally didn't touch the function [we plan to deprecate with 8.3](https://wiki.php.net/rfc/deprecations_php_8_3#global_mersenne_twister) to not add unnecessary churn for the translators.

### DIFF
--- a/reference/random/functions/random-int.xml
+++ b/reference/random/functions/random-int.xml
@@ -4,7 +4,7 @@
 <refentry xml:id="function.random-int" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>random_int</refname>
-  <refpurpose>Get a cryptographically secure, uniformly selected integer from a closed interval</refpurpose>
+  <refpurpose>Get a cryptographically secure, uniformly selected integer</refpurpose>
  </refnamediv>
 
  <refsect1 role="description"><!-- {{{ -->
--- a/reference/random/functions/random-int.xml
+++ b/reference/random/functions/random-int.xml
@@ -4,7 +4,7 @@
 <refentry xml:id="function.random-int" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>random_int</refname>
-  <refpurpose>Generates cryptographically secure pseudo-random integers</refpurpose>
+  <refpurpose>Get a cryptographically secure, uniformly selected integer from a closed interval</refpurpose>
  </refnamediv>
 
  <refsect1 role="description"><!-- {{{ -->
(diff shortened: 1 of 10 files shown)


reference/random/functions/random-int.xml
```diff
@@ -4,7 +4,7 @@
 <refentry xml:id="function.random-int" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>random_int</refname>
-  <refpurpose>Generates cryptographically secure pseudo-random integers</refpurpose>
+  <refpurpose>Get a cryptographically secure, uniformly selected integer from a closed interval</refpurpose>
```
cmb69, line 7: This refpurpose appears to be "too" long; I don't think that this level of detail is important here, and would stick with:
```suggestion
  <refpurpose>Get a cryptographically secure pseudo-random integer</refpurpose>
```

cmb69 (commented):
Thank you!

reference/random/functions/random-int.xml
TimWolla: The wording was unified with `Randomizer::getInt()` (just adding the cryptographically secure):

https://github.com/php/doc-en/blob/c993be5c5d2c8625ee3958ec63ae2565f146b18a/reference/random/random/randomizer/getint.xml#L5

I consider consistency to be very important here, but I concede that the new phrasing is a little on the long end. How about:

Randomizer::getInt(): Get a uniformly selected integer
random_bytes(): Get a cryptographically secure, uniformly selected integer

Notes:
- I consider the “uniformly” to be important information, because there might be other distributions, such as the [Normal Distribution](https://en.wikipedia.org/wiki/Normal_distribution).
- I dislike the “pseudo” in “pseudo-random”, because it provides no value-add for the user. If the randomness is cryptographically secure, it is sufficiently secure for all use-cases your app might have.

cmb69: > Randomizer::getInt(): Get a uniformly selected integer
> random_bytes(): Get a cryptographically secure, uniformly selected integer

That's fine!

cmb69 (approved):
Thank you!

reference/random/random/randomizer/getint.xml
```diff
@@ -2,7 +2,7 @@
 <refentry xml:id="random-randomizer.getint" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>Random\Randomizer::getInt</refname>
-  <refpurpose>Get a uniformly selected integer from a closed interval</refpurpose>
+  <refpurpose>Get a uniformly selected integer</refpurpose>
```
Girgias, line 5: This info seems important, so maybe document it in the description? (and also remove the undocumented func thing)

Girgias (approved):

reference/random/random/randomizer/getint.xml
TimWolla: Yes, I will add that information into the description when I write that one. Once the actual description is there, then I'll also remove the "undocumented function" warning.

The “uniformly” is already mentioned in the return values, though.

PR #2008, opened 2022-11-14
random: Clean up cross-references for randomness related functions

TimWolla:
Basically:

- Point to Randomizer where useful.
- Do not point from Randomizer to elsewhere (except random_int, random_bytes), developers should use the new API.


Girgias (approved):

cmb69 (approved):
Thank you!

PR #2015, opened 2022-11-16
Add <xi:fallback> to all <xi:includes>

cmb69:
Just a regex search and replace (`<xi:include([^<]*)/>` → `<xi:include$1><xi:fallback/></xi:include>`).


Girgias:
Superseded by https://github.com/php/doc-base/pull/99

PR #2018, opened 2022-11-16
random: Add description to extension constants

TimWolla:


reference/random/constants.xml
```diff
@@ -11,6 +11,8 @@
     </term>
     <listitem>
      <simpara>
+      Indicates that standardized Mt19937 algorithm shall be used when used with
```
Girgias, line 14:
```suggestion
      Indicates the standard Mt19937 algorithm shall be used when used with
```

reference/random/constants.xml
```diff
@@ -21,6 +23,13 @@
     </term>
     <listitem>
      <simpara>
+      Indicates that an incorrect implementation of the Mt19937 algorithm shall be
```
Girgias, line 26:
```suggestion
      Indicates that an incorrect implementation of the Mt19937 algorithm will be
```

Girgias (commented):
Wording which sounds clunky (also doing this on mobile for the first time)

reference/random/constants.xml
cmb69: I don't mind whether "shall" or "will" is used, but it should be consistent on this page (above we have "shall").

"An incorrect implementation" sounds strange; maybe make it clearer that it is "the old incorrect implementation", or something like that.

reference/random/constants.xml
cmb69: Can we reword the "be used when used with"?

TimWolla: @cmb69 Any specific suggestions what you’re unhappy with here? 

@Girgias I've taken "standardized" from https://www.php.net/manual/en/random-engine-mt19937.construct.php and https://www.php.net/manual/en/function.mt-srand.php.

reference/random/constants.xml
TimWolla: @Girgias As cmb noted: I've used “shall” for both and as a non-native speaker it sounds more correct, because I'm *instructing* the Engine use a non-standard implementation.

@cmb69 Likewise I've taken that phrasing “an incorrect implementation” from https://www.php.net/manual/en/random-engine-mt19937.construct.php.

reference/random/constants.xml
cmb69: > Any specific suggestions what you’re unhappy with here?

"used when used" appears to be an unfortunate repetition of "used". Maybe reword to "be used when passed to", or so.

reference/random/constants.xml
cmb69: > Likewise I've taken that phrasing “an incorrect implementation” from https://www.php.net/manual/en/random-engine-mt19937.construct.php.

But there it also says "for backwards compatibility with mt_srand() prior to PHP 7.1.0", so it clarifies which incorrect implementation this is about. Maybe something like:
```suggestion
      Indicates that an incorrect implementation (as used prior to PHP 7.1.0) of the Mt19937 algorithm shall be
```

TimWolla: @cmb69 I believe this information is also available here. Did you miss the second `<simpara>`?

reference/random/constants.xml
Girgias: My complaint about "standardized" is that, to me, it makes it sound like it went through some formal process to become a "standard", whereas just using "standard" means the "usual" implementation.

So maybe "correct" implementation is better?

Might also make sense to link to the actual paper: https://dl.acm.org/doi/10.1145/272991.272995

As to the whole wording how about:

```suggestion
      Indicates the correct Mt19937 algorithm will be used when calling
      <function>Random\Engine\Mt19937::__construct</function> and <function>mt_srand</function>.
```
?

TimWolla:
@cmb69 @Girgias Thanks for your reviews, I've rephrased that sentence entirely and also fixed the standardized -> correct in the Mt19937::__construct() and mt_srand() pages. Please take another look.

cmb69 (approved):
Thank you!

Girgias (approved):

PR #2019, opened 2022-11-18
random: Add description to PcgOneseq128XslRr64::jump()

TimWolla:
Contrary to xoshiro256**’s `jump()` methods (https://www.php.net/manual/en/random-engine-xoshiro256starstar.jump.php) I was not able to think of a “prime use case”. While “splitting the engine” also works with PCG, it works less great due to the smaller overall state and the fact that jump distance is limited by `PHP_INT_MAX`, which is in the realm of collision for 32 Bit PHP at least.

So a simple description of the functionality it is, unless @zeriyoshi has some ideas of what could be added?

### DIFF
--- a/reference/random/random/engine/pcgoneseq128xslrr64/jump.xml
+++ b/reference/random/random/engine/pcgoneseq128xslrr64/jump.xml
@@ -12,11 +12,9 @@
    <methodparam><type>int</type><parameter>advance</parameter></methodparam>
   </methodsynopsis>
   <para>
-
+   Moves the algorithm’s state ahead by the number of steps given by <parameter>advance</parameter>, as if
+   <function>Random\Engine\PcgOneseq128XslRr64::generate</function> was called that many times.
   </para>
-
-  &warn.undocumented.func;
-
  </refsect1>
 
  <refsect1 role="parameters">


TimWolla:
Semi-related: During my research I found that the PCG algorithm also supports efficient *backwards* jumps, which is interesting: https://www.pcg-random.org/using-pcg-cpp.html#void-backstep-state-type-delta.

@zeriyoshi Perhaps it would be a nice addition for PHP 8.3, if the `->jump()` method also supports negative parameters to indicate a “backstep”? Would you be interested in working on this?

zeriyoshi:
@TimWolla 
Yes. It would be nice to have all the methods available upstream. This will not cause a general BC Break either.

cmb69 (approved):
Thank you!

TimWolla:
@zeriyoshi Just to make sure, as you only replied to the second message about backwards jumps: Are you happy with what I've written here or do you have anything to add?

zeriyoshi (approved):
@TimWolla 
Sorry!
I haven't come up with too many practical use cases for this function. State jumping may be useful to improve randomness, but Secure should be used in most such applications. The reason I implemented it is simply because it was included in the original implementation.

PR #2021, opened 2022-11-20
Adapt to current OpenBSD version

msamesch:
This is only a draft. First of all, I'm not sure about the ports section. Do the two subdirectories core and extensions still exist? I could not find them in the ports tree.

### DIFF
--- a/install/unix/openbsd.xml
+++ b/install/unix/openbsd.xml
@@ -85,7 +85,7 @@ Follow the instructions shown with each package!
      requires Xorg to be installed. Unless already installed at base install by
      adding the <filename>xbase.tgz</filename> file set, this can be added at
      post-installation (see
-     <link linkend="https://www.openbsd.org/faq/faq4.html#FilesNeeded">OpenBSD FAQ#4</link>).
+     <link xlink:href="https://www.openbsd.org/faq/faq4.html#FilesNeeded">OpenBSD FAQ#4</link>).
     </simpara>
    </listitem>
   </itemizedlist>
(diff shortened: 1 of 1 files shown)


cmb69:
> First of all, I'm not sure about the ports section. Do the two subdirectories core and extensions still exist? I could not find them in the ports tree.

Maybe @devnexen knows?

devnexen:
extensions metdata are located [here (for example for php 8.1)](https://github.com/openbsd/ports/tree/master/lang/php/8.1/pkg).

Otherwise, since there is two releases per year, it makes sense to remove the info here indeed.

msamesch:
Thanks a lot for your help. As for now I only removed the 'directories' from the wording. I'm completely new to *BSD (installed it just for this update), but this ports section made me curious how this works.
BTW, can these instructions be used for all *BSDs?

devnexen:
Unfortunately, you can t really translate to FreeBSD nor NetBSD, despite the latter being the "father" of OpenBSD however its package system really differs, as for FreeBSD, for example, php extensions are a distinct package independent from the main one.

devnexen:
@cmb69 do you think it s ready to merge ? looks ok to me

cmb69:
I can't comment on (Open)BSD, but indeed, this looks good. Unless @msamesch is planning to do more right away, I suggest to merge.

msamesch:
> Unfortunately, you can t really translate to FreeBSD nor NetBSD, despite the latter being the "father" of OpenBSD however its package system really differs, as for FreeBSD, for example, php extensions are a distinct package independent from the main one.

Thanks for the info. Then this will be a seperate thing to do for the other BSDs. OTOH, maybe it doesn't make that much sense, because BSD users probably know their way around well enough anyway. ;)

msamesch:
> I can't comment on (Open)BSD, but indeed, this looks good. Unless @msamesch is planning to do more right away, I suggest to merge.

@cmb69 Yes, I was, but now it has already taken longer than originally intended. One more typo...

PR #2030, opened 2022-11-28
Fix inconsistent formatting for anonymous function

theodorejb:
The formatting in this example didn't match the other examples on the page.


cmb69:
Well, we really should decide on a CS for the examples in the docs, but this is likely uncontroversial, so thank you!

PR #2032, opened 2022-11-28
scandir - use constant name instead of its value

fajmanjir:
Improved readability. Furthermore, value of SCANDIR_SORT_DESCENDING can possibly change in the future :)


cmb69:
Thank you!

PR #2035, opened 2022-12-01
ZipArchive::open(): bool|int

voku:
- https://github.com/JetBrains/phpstorm-stubs/pull/1499
- https://github.com/php/php-src/blob/ca8247654c1e324d412bb6f40ef14fb506487983/ext/zip/php_zip.c#L1454-L1471

### DIFF
--- a/reference/zip/ziparchive/open.xml
+++ b/reference/zip/ziparchive/open.xml
@@ -72,7 +72,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true; on success, &false; or one of the following the error codes on error:
+   Returns &true; on success, &false; or one of the following error codes on error:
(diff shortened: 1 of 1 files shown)


reference/zip/ziparchive/open.xml
```diff
@@ -72,7 +72,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true; on success, or one of the following error codes:
+   Returns &true; on success, &false; or one of the following the error codes on error:
```
cmb69, line 75: There is a "the" too much:
```suggestion
   Returns &true; on success, &false; or one of the following error codes on error:
```

cmb69 (commented):
Thank you!

cmb69:
Thank you!

PR #2043, opened 2022-12-07
Fix example of interpolation in parse_ini_file

stembrain:
The `also_five` example does not work. The documentation was fixed in https://github.com/php/doc-en/commit/f8b0bb410016270b696b28d3d416672dd2f7e66b from a bug report at https://bugs.php.net/bug.php?id=81620 . However, the example value of `also_five` still existed. Now the example code demonstrates how to interpolate a configuration variable, which is what is supported.


cmb69:
Thank you!

PR #2044, opened 2022-12-07
array_keys(): fix incorrect parameter name

jrfnl:
... which would break function calls using named parameters. Demo: https://3v4l.org/7BNbJ

Ref: https://github.com/php/php-src/blob/11b612af6dbe9fdd60edb58d37e441c97bff79e0/ext/standard/basic_functions.stub.php#L1691

As a side-note: I was surprised to find this bug as I thought the parameter info in the docs was generated from the stub file ? (or in some other automated way kept updated)
If there is an automated way to update this, I'd highly recommend running it as if I found one, there's bound to be more parameter names being out of sync with the stubs...

### DIFF
--- a/reference/array/functions/array-keys.xml
+++ b/reference/array/functions/array-keys.xml
@@ -15,7 +15,7 @@
   <methodsynopsis role="procedural">
    <type>array</type><methodname>array_keys</methodname>
    <methodparam><type>array</type><parameter>array</parameter></methodparam>
-   <methodparam><type>mixed</type><parameter>search_value</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter>filter_value</parameter></methodparam>
    <methodparam choice="opt"><type>bool</type><parameter>strict</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
@@ -23,7 +23,7 @@
    string, from the <parameter>array</parameter>.
   </para>
   <para>
-   If a <parameter>search_value</parameter> is specified,
+   If a <parameter>filter_value</parameter> is specified,
    then only the keys for that value are returned. Otherwise, all
    the keys from the <parameter>array</parameter> are returned.
   </para>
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR!

> If there is an automated way to update this, […]

There is a semi-automatic way using [gen_stub.php](https://github.com/php/php-src/blob/master/build/gen_stub.php), but in this case, the problem is almost certainly the overloaded signature which is an issue. (Unfortunately, we can't change that overloaded signature without BC. It might make sense to introduce a new function `array_key_filter()` or so as a first step, but that's a different story.)

jrfnl:
> in this case, the problem is almost certainly the overloaded signature which is an issue. (Unfortunately, we can't change that overloaded signature without BC. It might make sense to introduce a new function `array_key_filter()` or so as a first step, but that's a different story.)

Ah, that explains why it hadn't been updated. Thanks @cmb69 !

For what it's worth: the overloaded signature for `array_keys()` doesn't conflict with the alternative signature, so named parameters are supported. IMO, that means no new function is needed.

This is in contrast to, for instance, [`imagepolygon`](https://www.php.net/manual/en/function.imagepolygon.php) where the overloaded signature conflicts with the original signature and named parameters are _**not**_ supported. In those kind of situations I would think a new function would make sense.

cmb69:
> the overloaded signature for `array_keys()` doesn't conflict with the alternative signature, so named parameters are supported. IMO, that means no new function is needed.

The problem are not the parameter names, but rather that you can't express both signatures in a single function declaration, since `$filter_value` has no meaningful default. And while `array_filter($array, ...$rest)` is possible, it is certainly more confusing than what we have now. Exposing `undefined` for the docs only might be an option, but might likely be confusing as well. cc @kocsismate 

For `image*polygon()` the situation is different, since the old signature already has been deprecated (`E_DEPRECATED`), and will be removed in the future. After that removal, named parameters will be supported for the 3 parameter signature.

PR #2048, opened 2022-12-08
Fix spelling

danmooney2:
"Existant" is a common misspelling of "Existent". Also, 'non' should act as a prefix to the word, either being hyphenated or treated as a compound word ("nonexistent").  I'm leaving it alone for now.

### DIFF
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -329,7 +329,7 @@ $test->obj = new stdClass;
   <sect2 xml:id="language.oop5.properties.dynamic-properties">
    <title>Dynamic properties</title>
    <para>
-    If trying to assign to a non existent property on an &object;,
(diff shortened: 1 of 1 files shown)


language/oop5/properties.xml
```diff
@@ -329,7 +329,7 @@ $test->obj = new stdClass;
   <sect2 xml:id="language.oop5.properties.dynamic-properties">
    <title>Dynamic properties</title>
    <para>
-    If trying to assign to a non existant property on an &object;,
+    If trying to assign to a non existent property on an &object;,
```
cmb69, line 332: Please "fix" that right away:
```suggestion
    If trying to assign to a non-existent property on an &object;,
```

cmb69 (commented):
Thank you!

cmb69:
Thank you!

PR #2051, opened 2022-12-09
Added option `SNI_server_certs` in  `context.ssl`

tianyiw2013:
My English is poor. This is my first time editing a php document.
The original document lack the option. After studying a lot of materials, this option `SNI_server_certs` was found.
The description of option may not be accurate, please correct it directly.

### Enable SNI (Server Name Indication):
**PEM must be contains certificate and private key.**
```php
$context = stream_context_create([
    'ssl' => [
        'SNI_enabled' => true,
        'SNI_server_certs' => [
            'host1.com' => '/path/host1.com.pem',
            'host2.com' => '/path/host2.com.pem',
        ],
    ]
]);
```

https://github.com/php/php-src/commit/fad14e3180b17127496c95fe7f0a4b1ca4f36196


philip (changes requested):
Thanks for adding the missing context option, here are a few suggestions:

- Remove `-enabled` from the `xml:id` 
- The text was understandably copied from the `local_cert` entry, seems the `SNI_server_certs` description should differ from it? The beginning of the php-src commit message includes "maps host names to appropriate certs should client handshakes advertise the SNI extension" which could help differentiate 
- The php-src commit message includes additional details, like the optional `*.` prefix to also match subdomains

bukka:
There's actually additional format supported now which allows separation of the key and cert as in https://github.com/php/php-src/blob/7b8a61a18c150b992dc9d392046c20739cad0c4d/ext/openssl/tests/sni_server_key_cert.phpt . It might be good to document as well.

Apology for completely missing the review request.

lacatoire:
Thanks for spotting this gap, and sorry it sat for so long.

`SNI_server_certs` is now documented on master: `language/context/ssl.xml` gained a `context.ssl.sni-server-certs` entry in [#5413](https://github.com/php/doc-en/pull/5413), merged, with the id suggested in the earlier review and a description written for the option rather than copied from `local_cert`. That change also rewrote the whole option list, which is why this branch now conflicts across the entire `variablelist`.

Closing as resolved, as requested above.

PR #2069, opened 2022-12-11
Update Error, Exception, and Throwable role attributes

kocsismate:


language/predefined/exception.xml
TateGunning, line 1: The exception code is 30

TateGunning (commented):
The exception code is 30

PR #2075, opened 2022-12-13
Add missing root element attributes

kocsismate:
This fixes a few build warnings


kocsismate:
I'm not sure what happened in the meanwhile, but I don't get the related warnings anymore.. Maybe it was due to the already fixed syntax errors in the includes... 🤔

PR #2081, opened 2022-12-14
Update FilesystemIterator::__construct doc for php 8.2

jawira:
Hello, I'm updating documentation for [FilesystemIterator::__construct](https://www.php.net/manual/en/filesystemiterator.construct.php).

Before PHP 8.2 the `FilesystemIterator::SKIP_DOTS` was always set and couldn't be removed, this changed in PHP 8.2.
If you are using the default value for `flags` parameter then everything work as expected, but if you are using any other value you have to add `FilesystemIterator::SKIP_DOTS` to maintain old behavior. 

You can see an example in <https://3v4l.org/HrmZ2>.

This already provoked a bug (see https://github.com/phingofficial/phing/pull/1685), that's why I also updated documentation in "breaking changes" section.

This change was added in https://github.com/php/php-src/commit/088e5478029f7b3482a9c5ae2b06b77772b92480 and as far as I know it's not part of any pull request, I suppose that's the reason it hasn't been documented yet.

### DIFF
--- a/appendices/migration82/incompatible.xml
+++ b/appendices/migration82/incompatible.xml
@@ -51,7 +51,7 @@
   </para>
 
   <para>
-   <methodname>FilesystemIterator::__construct</methodname>, prior to PHP 8.2
+   <methodname>FilesystemIterator::__construct</methodname>: prior to PHP 8.2.0,
    <constant>FilesystemIterator::SKIP_DOTS</constant> constant was always set
    and couldn't be disabled. In order to maintain the previous behaviour the constant
    must be explicitly set when using the <parameter>flags</parameter> parameter.
--- a/appendices/migration82/incompatible.xml
+++ b/appendices/migration82/incompatible.xml
@@ -53,9 +53,9 @@
   <para>
    <methodname>FilesystemIterator::__construct</methodname>, prior to PHP 8.2
    <constant>FilesystemIterator::SKIP_DOTS</constant> constant was always set
-   and couldn't be disabled. In order to maintain previous behaviour you have to
-   explicitly add this constant if you are using <parameter>flags</parameter>
-   parameter. The default value from <parameter>flags</parameter> parameter has
+   and couldn't be disabled. In order to maintain the previous behaviour the constant
+   must be explicitly set when using the <parameter>flags</parameter> parameter.
+   The default value from <parameter>flags</parameter> parameter has
    not been modified.
   </para>
 
--- a/appendices/migration82/incompatible.xml
+++ b/appendices/migration82/incompatible.xml
@@ -51,6 +51,15 @@
   </para>
 
   <para>
+   <methodname>FilesystemIterator::__construct</methodname>, prior to PHP 8.2
+   <constant>FilesystemIterator::SKIP_DOTS</constant> constant was always set
+   and couldn't be disabled. In order to maintain previous behaviour you have to
+   explicitly add this constant if you are using <parameter>flags</parameter>
+   parameter. The default value from <parameter>flags</parameter> parameter has
+   not been modified.
+  </para>
+
+  <para>
    <function>strtolower</function>,
    <function>strtoupper</function>,
    <function>stristr</function>,
--- a/reference/spl/filesystemiterator/construct.xml
+++ b/reference/spl/filesystemiterator/construct.xml
@@ -98,14 +98,24 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-$it = new FilesystemIterator(dirname(__FILE__));
+$it = new FilesystemIterator(dirname(__FILE__), FilesystemIterator::CURRENT_AS_FILEINFO);
 foreach ($it as $fileinfo) {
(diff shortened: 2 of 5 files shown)


appendices/migration82/incompatible.xml
```diff
@@ -50,6 +50,15 @@
    <link linkend="ini.open-basedir">open_basedir</link>.
   </para>
 
+  <para>
+   <methodname>FilesystemIterator::__construct</methodname>, prior to PHP 8.2
+   <constant>FilesystemIterator::SKIP_DOTS</constant> constant was always set
+   and couldn't be disabled. In order to maintain previous behaviour you have to
+   explicitly add this constant if you are using <parameter>flags</parameter>
+   parameter. The default value from <parameter>flags</parameter> parameter has
+   not been modified.
```
Girgias, line 59:
```suggestion
   and couldn't be disabled. In order to maintain the previous behaviour the constant
   must be explicitly when using the <parameter>flags</parameter> parameter.
   The default value from <parameter>flags</parameter> parameter has
   not been modified.
```

Personalization (i.e. usage of "you") should be avoided

cmb69: > "must be explicitly when using the"

"Must be explicitly set when using the" or "must be explicit when using the", or something like that.

jawira: Hi, I updated the doc. Let me know if I have to update something else.

reference/spl/filesystemiterator/construct.xml
```diff
@@ -75,6 +69,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.2.0</entry>
+       <entry>
+        Prior to PHP 8.2 <constant>FilesystemIterator::SKIP_DOTS</constant> was
```
cmb69, line 75: We always prefer to use the full PHP version (and there should be a comma afterwards), so this should be:
```suggestion
        Prior to PHP 8.2.0, <constant>FilesystemIterator::SKIP_DOTS</constant> was
```

appendices/migration82/incompatible.xml
```diff
@@ -50,6 +50,15 @@
    <link linkend="ini.open-basedir">open_basedir</link>.
   </para>
 
+  <para>
+   <methodname>FilesystemIterator::__construct</methodname>, prior to PHP 8.2
```
cmb69, line 54: This reads a bit strange (I'm not a native English speaker, though). Maybe better:
```suggestion
   <methodname>FilesystemIterator::__construct</methodname>: prior to PHP 8.2.0,
```

reference/spl/filesystemiterator/current.xml
```diff
@@ -52,6 +52,8 @@ foreach ($iterator as $fileinfo) {
     &example.outputs.similar;
     <screen>
 <![CDATA[
+/www/examples/.
```
cmb69, line 55: I'm not sure what we should do with these examples. We could use `&example.outputs.82.similar;` instead of `&examle.outputs.similar;`, but readers might wonder what the output was prior to PHP 8.2.0. Maybe explicitly passing `FilesystemIterator::SKIP_DOTS` in the constructor is the better alternative?

cmb69 (commented):
Thank you!

reference/spl/filesystemiterator/current.xml
jawira: IMHO documentation should reflect current state of PHP, so I would not change example's code, rather I would use `&example.outputs.82.similar;` only for outputs I have changed.

However, for the only example in [FilesystemIterator::__construct](https://www.php.net/manual/en/filesystemiterator.construct.php) -and only for this example- I would display two outputs, one with `&example.outputs.82.similar;` and another with `&example.outputs.81` only to make clear the behavior has changed between versions. I would only display two outputs in this page because I'm sure this is by far the most read page about FilesystemIterator.

I'm against updating the examples with `FilesystemIterator::SKIP_DOTS` because this will leave no example about using `FilesystemIterator` with a single parameter.

Anyway, let me know what do you thing and I will make the changes you request.

cmb69: > However, for the single example in [FilesystemIterator::__construct](https://www.php.net/manual/en/filesystemiterator.construct.php) -and only for this example- I would display two outputs, one with `&example.outputs.82.similar;` and another with `&example.outputs.81` only to make clear the behavior has changed between versions. I would only display two outputs in this page because I'm sure this is by far the most read page about `FilesystemIterator`.

Okay, that sounds like a plan. Maybe instead of `&example.outputs.81;` just use plain XML instead, something like
````xml
<para>Output of the above example prior to PHP 8.2.0 is similar to:</para>
````

jawira: I made the changes, I had to update de code in order to be coherent with outputs.
Do not hesitate to tell me if I have to change something else :)

cmb69: All good now. Thank you!

PR #2086, opened 2022-12-15
ldap: Document removal of ber_identifier param

zerocrates:
The docs for ldap_first_attribute and ldap_next_attribute note that PHP accepts a third, deprecated argument, `$ber_identifier` without raising an error.

In fact, passing 3 parameters to these functions has been an ArgumentCountError since PHP 8.0.0 (see php/php-src@eeec37d31d6aa4452d83e2640aee7c8c25349617).

For both functions, this change removes the argument from the list, removes the note indicating that a third argument can be passed, and adds the change at version 8.0.0 to the changelog.

### DIFF
--- a/reference/ldap/functions/ldap-first-attribute.xml
+++ b/reference/ldap/functions/ldap-first-attribute.xml
@@ -43,25 +43,6 @@
       </para>
      </listitem>
     </varlistentry>
-    <varlistentry>
-     <term><parameter>ber_identifier</parameter></term>
-     <listitem>
-      <para>
-       <parameter>ber_identifier</parameter> is the identifier to internal
-       memory location pointer. It is passed by reference. The same
-       <parameter>ber_identifier</parameter> is passed to
-       <function>ldap_next_attribute</function> , which modifies that
-       pointer.
-      </para>
-      <note>
-       <para>
-        This parameter is no longer used as this is now handled automatically
-        by PHP. For backwards compatibility PHP will not throw an error if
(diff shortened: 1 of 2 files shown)


reference/ldap/functions/ldap-first-attribute.xml
```diff
@@ -87,6 +68,12 @@
     <tbody>
      &ldap.changelog.ldap-object;
      &ldap.changelog.entry-object;
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       The previous third parameter <parameter>ber_identifier</parameter> is no longer accepted; it was deprecated and unncessary since PHP 5.2.4.
```
cmb69, line 74: I don't think we should mention PHP 5.2.4 (or any other PHP version prior to PHP 7.0.0). Maybe just:
```suggestion
       The unused third parameter <parameter>ber_identifier</parameter> is no longer accepted.
```
or something like that.

cmb69 (commented):
Thank you!

reference/ldap/functions/ldap-first-attribute.xml
zerocrates: Happy to take it out if it's an issue. The only reason I mentioned how far back its soft-removal goes was to clarify that there's no realistic backwards-compatibility issue in simply removing it in code where it's still present.

zerocrates:
Okay, that's done.

cmb69:
Thank you!

PR #2090, opened 2022-12-18
List inherited constructors on the class synopsis pages

kocsismate:


kocsismate:
@cmb69 Can this be merged instead of my "default" constructor related additions?

reference/intl/intlcodepointbreakiterator.xml
```diff
@@ -48,6 +48,9 @@
     </xi:include>
 
     <classsynopsisinfo role="comment">&InheritedMethods;</classsynopsisinfo>
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.intlbreakiterator')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[@role='IntlBreakIterator'])">
+     <xi:fallback/>
+    </xi:include>
```
cmb69, line 53: The "inherited" constructor is private, and as such is not really inherited. It is documented to convey that the class may not be instantiated, but I think we should document the class as being abstract instead (and document that it is not actually abstract), and, of course,  actually make it abstract.

cmb69 (commented):
This is generally fine. Thank you!

We should figure out what to do regarding private constructors, though (see comment below).

reference/intl/intlcodepointbreakiterator.xml
kocsismate: Nice catch! In my opinion, simply skipping these private constructors from the inherited method section is the way to go. Would you be ok with this approach?

cmb69: Yeah, sure. Would we need to remove/change the roles for that, or do you have another idea?

kocsismate: No, we don't need to change the markup, gen_stub can handle it :) See https://github.com/php/php-src/pull/10127/commits/b122959b5b0bf0a9d96a475b7692422dcdb0edae

cmb69: Great!

cmb69 (approved):
Thank you!

PR #2091, opened 2022-12-19
Remove fast_shutdown as deprecated since 7.2

Julian-Louis:
Hello !
What do you think about removing this setting from the recommended settings, as it no longer exists? 

Have a good day !


cmb69:
Thank you for the PR! However, the manual is currently supposed to provide documentation for PHP 7.0+, so we cannot yet remove that info yet.

PR #2095, opened 2022-12-19
Better crosslinks between str_replace and preg_replace

IMSoP:
The str_replace page had a helpful note that if you'd somehow got here when you were looking for preg_replace, you didn't need to go there. But if you were actually on the page for preg_replace, there was no link to str_replace at all, not even in the See Also section.

### DIFF
--- a/reference/pcre/functions/preg-replace.xml
+++ b/reference/pcre/functions/preg-replace.xml
@@ -21,6 +21,11 @@
    <parameter>pattern</parameter> and replaces them with
    <parameter>replacement</parameter>.
   </para>
+  <para>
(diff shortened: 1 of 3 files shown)


cmb69 (commented):
Thank you! Overall that makes sense to me, but when we're changing these sentences, we should get rid of the personalization ("you") right away.

IMSoP:
@cmb69 Good point; I was struggling to word it, and I think it's because I was trying to keep too much of the original. What do you think of this version?

cmb69 (approved):
I think I had used passive voice, but this is good as well (and might be better). Thank you!

PR #2098, opened 2022-12-20
Nits to the type juggling page

Girgias:
The big one is the need to clarify the behaviour around the coercive typing mode, as the conversion of objects to strings is relatively straight forward but in theory internal objects that overload the cast handler can convert to other types too.

### DIFF
--- a/language/types/type-juggling.xml
+++ b/language/types/type-juggling.xml
@@ -150,7 +150,7 @@
    <!-- e.g. An object that implements __toString will pass a string type -->
    type, the value is convertable to a scalar type,
    and the coercive typing mode is active
-   (the default), the value may be converted to an accepted scalar valued.
+   (the default), the value may be converted to an accepted scalar value.
    See below for a description of this behaviour.
   </para>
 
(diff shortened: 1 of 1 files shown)


language/types/type-juggling.xml
```diff
@@ -146,8 +146,10 @@
    Two exceptions exist, the first one is: if the value is of type
    <type>int</type> and the declared type is <type>float</type>, then the
    integer is converted to a floating point number.
-   The second one is: if the value and the declared type are
-   <emphasis>scalar</emphasis> types, and the coercive typing mode is active
+   The second one is: if the declared type is a <emphasis>scalar</emphasis>
+   <!-- e.g. An object that implements __toString will pass a string type -->
+   type, the value is convertable to a scalar type,
+   and the coercive typing mode is active
    (the default), the value may be converted to an accepted scalar valued.
```
cmb69, line 153:
```suggestion
   (the default), the value may be converted to an accepted scalar value.
```

cmb69 (approved):
Thank you! Looks good, but I suggest to fix the existing typo right away.

PR #2107, opened 2022-12-22
Deprecation and alternatives for utf8_encode and utf8_decode

IMSoP:


reference/strings/functions/utf8-decode.xml
```diff
@@ -134,6 +134,107 @@ string(1) "?"
...
+5a6feb
+5a6feb
+5a6feb
+]]>
+     </screen>
+    </informalexample>
+    Specifying <literal>'?'</literal> as the <literal>'to_subst'</literal> option
+    to <methodname>UConverter::transcode</methodname> gives the same result as
+    <function>utf8_decode</function> for strings which are invalid or which can not
+    be represented in ISO 8859-1
```
cmb69, line 215: Missing period:
```suggestion
    be represented in ISO 8859-1.
```

reference/strings/functions/utf8-decode.xml
```diff
@@ -134,6 +134,107 @@ string(1) "?"
...
+    <informalexample>
+     <programlisting role="php">
+<![CDATA[
+$utf8_string = "\xE2\x82\xAC"; // € (Euro Sign) does not exist in ISO 8859-1
```
cmb69, line 219: Missing `<?php` (which prevents syntax highlighting):
```suggestion
<?php
$utf8_string = "\xE2\x82\xAC"; // € (Euro Sign) does not exist in ISO 8859-1
```

reference/strings/functions/utf8-decode.xml
```diff
@@ -134,6 +134,107 @@ string(1) "?"
   </example>
  </refsect1>
 
+ <refsect1 role="notes">
+  &reftitle.notes;
+  <note>
+   <title>Deprecation and alternatives</title>
+   <para>
+    This function has been <emphasis>deprecated</emphasis> as of PHP 8.2.0,
```
cmb69, line 142: For a long time I wonder whether that should be "has been" or "is"; I think the latter is correct, but I may be wrong.

cmb69 (commented):
Very nice, thank you!

reference/strings/functions/utf8-decode.xml
IMSoP: I agree, that reads better in this context. I've left it as "has been" in the changelog, though, to match "has been moved".

Girgias (approved):

PR #2110, opened 2022-12-23
Fix typo in PHP 7.0 migration guide

afilina:
The title refers to "break and switch" when it means "break and continue".


cmb69:
Thank you!

PR #2111, opened 2022-12-23
Constant disallowed as break and continue argument as of PHP 7.0

afilina:
This is no longer allowed ([3v4l.org](https://3v4l.org/MMuci)):
```php
while (true) {
    break E_ERROR;
}
```

It seems like this is a continuation of the change introduced in PHP 5.4, and was overlooked in the documentation.
> The break and continue statements no longer accept variable arguments (e.g., break 1 + foo() * $bar;).

Source: https://php-legacy-docs.zend.com/manual/php5/en/migration54.incompatible

### DIFF
--- a/appendices/migration70/incompatible/other.xml
+++ b/appendices/migration70/incompatible/other.xml
@@ -371,6 +371,15 @@ switch (1) {
   </para>
  </sect3>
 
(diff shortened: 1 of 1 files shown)


jimwins (approved):
LGTM

cmb69:
Thank you!

PR #2112, opened 2022-12-23
Remove replaceable class from magic constant

afilina:
Attempting to fix PR #728 which in turn fixes issue https://github.com/php/doc-en/issues/707

### DIFF
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -739,7 +739,7 @@ Does\Not\Exist
     </example>
    </note>
    <para>
-    As of PHP 8.0.0, the <literal>::class</literal> constant may also be used on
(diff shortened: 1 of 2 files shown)


language/oop5/basic.xml
```diff
@@ -739,7 +739,7 @@ Does\Not\Exist
     </example>
    </note>
    <para>
-    As of PHP 8.0.0, the <literal>::class</literal> constant may also be used on
+    As of PHP 8.0.0, <literal>::class</literal> may also be used on
     objects. This resolution happens at runtime, not compile time. Its effect is
     the same as calling <function>get_class</function> on the object.
```
Girgias, line 744: This just reads really clunky. But don't know how else to phrase it :|

afilina: Might just go with it and refine it later.

PR #2113, opened 2022-12-24
Better example for usleep()

zonuexe:
It is unnatural that the only example exceeds that number even though <q>may not be supported by the operating system</q> is mentioned. Also, it makes sense to show that you can sleep with sub-second accuracy.

> Note: Values larger than 1000000 (i.e. sleeping for more than a second) may not be supported by the operating system. Use [sleep()](https://www.php.net/manual/en/function.sleep.php) instead.

Linux timers seem to work with this precision: https://3v4l.org/1GI6a

### DIFF
--- a/reference/misc/functions/usleep.xml
+++ b/reference/misc/functions/usleep.xml
@@ -63,13 +63,19 @@
 <?php
 
 // Current time
-echo date('h:i:s') . "\n";
+echo (new DateTime('now'))->format('H:i:s.v'), "\n";
 
-// wait for 2 seconds
-usleep(2000000);
+// wait for 2 milliseconds
+usleep(2000);
 
 // back!
-echo date('h:i:s') . "\n";
+echo (new DateTime('now'))->format('H:i:s.v'), "\n";
+
+// wait for 30 milliseconds
+usleep(30000);
+
+// back again!
+echo (new DateTime('now'))->format('H:i:s.v'), "\n";
 
 ?>
 ]]>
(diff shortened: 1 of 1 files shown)


reference/datetime/functions/date.xml
```diff
@@ -38,6 +38,15 @@
...
      </para>
+      <note>
+       <simpara>
+        <function>date</function> will always generate
+        <literal>000000</literal> as microseconds since it takes an <type>int</type>
+        parameter, whereas <methodname>DateTime::format</methodname> does
+        support microseconds if <classname>DateTime</classname> was
+        created with microseconds.
+       </simpara>
+      </note>
```
zonuexe, line 49: duplicate

https://github.com/php/doc-en/blob/31b22987dd66a0251de914029b9f1b4c377bf9af/reference/datetime/datetimeinterface/format.xml#L255-L262

cmb69 (approved):
I think this clarification makes sense, but how is it related to `usleep()`?

zonuexe:
@cmb69 
I just tried to write `date('H:i:s.v')` but realized I had to replace it with `DateTime` due to int constraints.
You're right, it's a separate issue, so I could split it into another PR.

zonuexe:
This PR now only has a commit about `usleep()`.

mumumu (approved):
> it makes sense to show that you can sleep with sub-second accuracy.
> ...
> Linux timers seem to work with this precision

Windows precision does not work exactly than linux, but this fact is already mentioned in note. This change seems makes sense for me, too.

PR #2114, opened 2022-12-24
Duplicate note on microsecond precision from datetime.formats

zonuexe:
Duplicate the note because users who tried to use `date()` get confused about the format for microseconds.

refs https://github.com/php/doc-en/pull/2113


mumumu (approved):
The only caveat about `date` in `DateTimeInterface::format` is about microseconds, so this PR looks good to me.

PR #2118, opened 2022-12-25
OOP5/Traits/Constants: fix example code

jrfnl:
The example code contained three errors:
* Reference to non-existent `ConstantTrait` trait (x2).
* Instantiation of non-existent `ConstantExample` class.
* Reference to an undefined constant.

Fixed now.


cmb69:
Thank you!

PR #2123, opened 2022-12-27
Reword excessively hardcore requirement

RobinHood70:
While hilarious as is, this should probably be reworded.

### DIFF
--- a/reference/errorfunc/functions/set-error-handler.xml
+++ b/reference/errorfunc/functions/set-error-handler.xml
@@ -18,26 +18,26 @@
    errors in a script.
   </para>
   <para>
-   This function can be used for defining your own way of handling errors
-   during runtime, for example in applications in which you need to do
-   cleanup of data/files when a critical error happens, or when you need
-   to trigger an error under certain conditions (using
-   <function>trigger_error</function>).
+   This function can be used to define custom error handlers during runtime,
+   for example in applications which need to do file/data cleanup when a critical
+   error happens, or when triggering an error in response to certain conditions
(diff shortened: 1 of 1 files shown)


cmb69:
> While hilarious as is, this should probably be reworded.

Yeah, but while we're at it, we should remove the personalization ("your") right away, at least for that paragraph.

RobinHood70:
> Yeah, but while we're at it, we should remove the personalization ("your") right away, at least for that paragraph.

Done for the whole file. How does it look?

reference/errorfunc/functions/set-error-handler.xml
```diff
@@ -18,26 +18,26 @@
    errors in a script.
   </para>
   <para>
-   This function can be used for defining your own way of handling errors
-   during runtime, for example in applications in which you need to do
-   cleanup of data/files when a critical error happens, or when you need
-   to trigger an error under certain conditions (using
-   <function>trigger_error</function>).
+   This function can be used to define custom error handlers during runtime,
+   for example in applications which need to file/data cleanup when a critical
```
cmb69, line 22:
```suggestion
   for example in applications which need to do file/data cleanup when a critical
```
or something like that, I think.

cmb69 (commented):
Pretty good. Thank you!

cmb69:
Oh, there has been trailing whitespace which made the whitespace-checker scream. I've removed that, squashed and force-pushed (otherwise the whitespace-checker would still complain).

Thank you for the PR!

PR #2126, opened 2022-12-28
Add documentation for missing attribute constructors

kocsismate:
After these changes, there are only:
- 25 file diffs between stubs and class synopsis pages (and validation reports only 4-5 missing classes)
- 43 file diffs between stubs and method synopsis pages (but validation reports tons of missing functions and methods)

### DIFF
--- a/language/predefined/attribute/construct.xml
+++ b/language/predefined/attribute/construct.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="attribute.construct" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>Attribute::__construct</refname>
+  <refpurpose>Construct a new Attribute instance</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <constructorsynopsis role="Attribute">
+   <modifier>public</modifier> <methodname>Attribute::__construct</methodname>
+   <methodparam choice="opt"><type>int</type><parameter>flags</parameter><initializer>Attribute::TARGET_ALL</initializer></methodparam>
+  </constructorsynopsis>
+  <para>
+   Constructs a new <classname>Attribute</classname> instance.
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry><term><parameter>flags</parameter></term>
+     <listitem>
+      <para>
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+</refentry>
+<!-- Keep this comment at the end of the file
+ Local variables:
+ mode: sgml
+ sgml-omittag:t
+ sgml-shorttag:t
+ sgml-minimize-attributes:nil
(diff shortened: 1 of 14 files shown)


language/predefined/attribute/construct.xml
```diff
@@ -0,0 +1,53 @@
...
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry><term><parameter>flags</parameter></term>
+     <listitem>
+      <para>
```
kocsismate, line 26: I didn't yet have time to think about a proper description. I'm open for suggestions though :)

cmb69: Hmm, we would also need to document the `Attribute` class constants (which serve as possible values for this parameter). Maybe just leave the description empty for now; we can still flesh out this documentation later.

cmb69 (commented):
Thank you!

I've added a commit to include the new construct pages into the manual (unfortunately necessary for class documentation in languages/predefined), but I suggest to fix the mix of hyphenated and not hyphenated word (e.g. `allowdynamicproperties` vs. `allow-dynamic-properties`). For consistency with most of what we already had, it's probably best to remove the hyphens.

Girgias:
> Thank you!
> 
> I've added a commit to include the new construct pages into the manual (unfortunately necessary for class documentation in languages/predefined), but I suggest to fix the mix of hyphenated and not hyphenated word (e.g. `allowdynamicproperties` vs. `allow-dynamic-properties`). For consistency with most of what we already had, it's probably best to remove the hyphens.

I decide to put hyphens because I find it easier to read and also AFAIK is better for SEO (not that really should be that big of a concern for us) and would rather prefer we'd use them more then less, but just my opinion :)

kocsismate:
> For consistency with most of what we already had, it's probably best to remove the hyphens.

👍  yes, I agree.

> I decide to put hyphens because I find it easier to read and also AFAIK is better for SEO (not that really should be that big of a concern for us) and would rather prefer we'd use them more then less, but just my opinion :)

The problem is that gen_stub.php cannot really generate these hyphenated references: currently, it simply has to lowercase the class name, but if we wanted to have these SEO friendly names, then it would need to find the parts of the names based on the upper-cased letters, and put a hyphen between them. However, we would need to handle the special cases as well, like upper-cases abbreviations (like CURLFile). That's very convoluted and error-prone.

So I strongly ask for letting these hyphens go :)

language/predefined/attribute/construct.xml
kocsismate: OK, I'm fine with skipping the description for now :) As most of the class synopsis diffs are due to missing class constants, I'm going to continue my project with adding them (albeit with mostly empty descriptions). If we manage to list all the class constants, only a handful of class synopsis diffs remain.

cmb69:
I agree that the hyphenated names are better readable, but Máté's arguments are more relevant for now.

kocsismate:
@Girgias Are you ok with removing the hyphens?

Girgias:
> @Girgias Are you ok with removing the hyphens?

Yeah it's fine, one can always work on the SEO part at another moment

TimWolla:
Will redirects need to be set up for the renamed files? They are linked in https://www.php.net/releases/8.2/en.php at the very least.

cmb69:
> Will redirects need to be set up for the renamed files?

Ah, good catch! [Done](https://github.com/php/web-php/commit/8e8bed3a4d538b63aebbde083d17e75cc5be131f).

PR #2128, opened 2022-12-29
Fix AppendIterator::getInnerIterator() return type

othercorey:


cmb69:
It seems to me `::getInnerIterator()` is only defined on the superclass `InnerIterator`, and as such that file should be removed altogether.

cmb69:
Should be [fixed now](https://github.com/php/doc-en/commit/0189a49610b2e53b9e9be9f0457df9277bffaaee). :)

PR #2129, opened 2022-12-29
Fix CachingIterator::getInnerIterator() return type

othercorey:


cmb69:
It seems to me `::getInnerIterator()` is only defined on the superclass `InnerIterator`, and as such that file should be removed altogether.

PR #2130, opened 2022-12-30
Update return value description for date()

Loyen:
The documentation hints that `timestamp` can be non-numeric, but that will cause a `TypeError` since PHP 8.0.


cmb69:
Ah, good catch! And the behavior prior to PHP 8.0.0 is not specific to this function, it is good to remove that info.

Thank you!

PR #2131, opened 2022-12-31
ArgumentCountError is not limited to user-defined stuff.

divinity76:
For example strlen(); generates ArgumentCountError and is not a user-defined function.

### DIFF
--- a/language/predefined/argumentcounterror.xml
+++ b/language/predefined/argumentcounterror.xml
@@ -11,7 +11,7 @@
    &reftitle.intro;
    <para>
     <ooclass><classname>ArgumentCountError</classname></ooclass> is thrown
-    when too few arguments are passed to a user-defined function or method.
+    when too few arguments are passed to a function or method.
    </para>
   </section>
 <!-- }}} -->


cmb69:
Thank you! Unfortunately, it is not as simple as that. `ArgumentCountError`s are thrown for internal function [only as of PHP 8.0.0](https://3v4l.org/thJr0), and not only when too few arguments were passed, but also [when too many arguments were given](https://3v4l.org/Hh3bN). Prior to that version, both cases raised `E_WARNING` instead.

For userland functions, `ArgumentCountError` is thrown for [too few arguments as of PHP 7.1.0](https://3v4l.org/VZpnB) (previously, `E_WARNING` was raised instead). Excess arguments to userland functions are still ignored (to cater to `func_get_args()` and friends).

I think we should be precise here. :)

divinity76:
@cmb69 fixed, please re-read it

cmb69:
> well PHP7 is EOL so we shouldn't waste effort on PHP7

The manual is still supposed to document PHP 7.0.0 - 8.2.x. There is no need to describe the PHP 7 behavior in the description text, but there should be changelog entries added (see how that's done for [`Throwable`](https://github.com/php/doc-en/blob/bd38d9d20ab0186531a9e923494a30ac23b1ff8c/language/predefined/throwable.xml#L62-L85)).

And it is important to note that

> Excess arguments to userland functions are still ignored

Girgias:
Indeed, didn't see that. GitHub is very annoying with how it sometimes doesn't  link issues...

PR #2132, opened 2022-12-31
Fix bcsqrt return type description

othercorey:
These functions thrown an error if the value is negative not return null. Changed to match bcpow.


cmb69:
Unfortunately, it is not as simple as that. I tried my best with https://github.com/php/doc-en/commit/04c37bc8ca6b2faff2b14b85b4576529444bd6a2, but maybe I've missed some details.

othercorey:
Thanks! I was having no luck with the online editor.

PR #2133, opened 2023-01-01
Add void ternary note

RobinHood70:
Unlike most languages, PHP supports both void and mixed void/value-returning ternaries. I've noted this in the ternary docs. I'm a little unclear if I'm using the correct terminology, though. Suggestions are welcome!

### DIFF
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -1864,7 +1864,7 @@ if (empty($_POST['action'])) {
        <programlisting role="php">
         <?php
         // Both functions are void, so no value assignment is needed.
-        isset($seed) ? mt_srand($seed) : mt_srand();        
+        isset($seed) ? mt_srand($seed) : mt_srand();
 
         // One side returns a value, so an assignment is needed unless the
         // the return value is to be ignored.
--- a/language/operators.xml
+++ b/language/operators.xml
@@ -98,17 +98,17 @@
        <row>
         <entry>(n/a)</entry>
         <entry>
-         <literal>+</literal> 
-         <literal>-</literal> 
-         <literal>++</literal> 
-         <literal>--</literal> 
-         <literal>~</literal> 
-         <literal>(int)</literal> 
-         <literal>(float)</literal> 
-         <literal>(string)</literal> 
+         <literal>+</literal>
+         <literal>-</literal>
+         <literal>++</literal>
+         <literal>--</literal>
+         <literal>~</literal>
+         <literal>(int)</literal>
+         <literal>(float)</literal>
+         <literal>(string)</literal>
          <literal>(array)</literal>
-         <literal>(object)</literal> 
-         <literal>(bool)</literal> 
+         <literal>(object)</literal>
+         <literal>(bool)</literal>
          <literal>@</literal>
         </entry>
         <entry>
@@ -147,8 +147,8 @@
        <row>
         <entry>left</entry>
         <entry>
-         <literal>+</literal> 
-         <literal>-</literal> 
+         <literal>+</literal>
+         <literal>-</literal>
          <literal>.</literal>
         </entry>
         <entry>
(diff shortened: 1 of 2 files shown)


RobinHood70:
I've tried removing all the trailing whitespace I could find with a regex search, but the automated check is still failing.

cmb69:
Thank you for the PR!

> I've tried removing all the trailing whitespace I could find with a regex search, but the automated check is still failing.

The whitespace-checker only checks whitespace in changes you've made, so you don't have to remove unrelated trailing whitespace. And actually, you should not, as that makes the diffs very noisy especially for translators. However, the whitespace-checker looks at each commit, and if there is trailing whitespace in actually changed lines, it reports that. It doesn't notice, though, if a later commit removes that whitespace. So you would have to squash the commits, and then force-push.

Regarding the actual patch: I think it is important to clarify the behavior of void functions in expressions generally (not only for the ternary operator). From what I can tell, currently there is only a note on the [void type page](https://www.php.net/manual/en/language.types.void.php), but that should probably also mentioned on the [expression page](https://www.php.net/manual/en/language.expressions.php). Although when I look at that page, I think that it needs a thorough overhaul, and especially I do not agree with

> PHP provides a full and powerful implementation of expressions, and documenting it entirely goes beyond the scope of this manual.

Where else should that "powerful implementation of expressions" be documented instead?

language/operators.xml
```diff
@@ -1853,6 +1853,30 @@ if (empty($_POST['action'])) {
...
+       <title>Void Ternaries</title>
+       <programlisting role="php">
+        <?php
+        // Both functions are void, so no value assignment is needed.
+        isset($seed) ? mt_srand($seed) : mt_srand();
+
+        // One side returns a value, so an assignment is needed unless the
+        // return value is to be ignored.
+        $initSrand = isset($seed) ? mt_srand($seed) : 'Seed was not specified!';
+        if (isset($initSrand)) {
```
kamil-tekiela, line 1872: I will be honest I have no idea what this says. Why do you use `isset($initSrand)` on a variable that is clearly set and has a non-null value? Is `isset($seed)` significant or just an example of what a ternary looks like? What is "one side"? What return value is to be ignored if in the lines above you said `mt_srand` does not return anything? Why is the assignment needed?

Regarding all other changes, I don't know if they are part of this PR or if you made them accidentally. 

Generally, I think you are right that the behaviour of expression with void functions needs to be better documented, but as cmb69 said, this is neither the place for it nor the right explanation. Thank you for this PR, but I am going to close it. You can make a new PR and document it as cmb69 said on the expression page in a more generic way.

RobinHood70: The four lines above were the only "real" changes; the rest were me trying to figure out why the whitespace checker kept complaining, which cmb69 explained above.

The reason you're checking something that's "clearly set" is precisely the reason void-returning functions need more documentation. :) `isset($initSrand)` will only succeed if `$seed` wasn't set, so `$initSrand` gets the error message, which the `echo()` can then display. If it was assigned the results of `mt_srand()`, which is void, then `isset($initSrand)` will fail and the `echo()` will be skipped. This is clearly bad coding style, but it *is* valid code that demonstrates what happens when a ternary is evaluating a void function.

So, parsing the comment with that in mind: "One side returns a value" is when the ternary returns a string (the right-hand/false side of the ternary). So, you need to assign that value to something unless you plan to ignore it. This is in comparison to the example above it where you don't need any kind of a `$variable =` statement, since both sides of the ternary are void.

RobinHood70: Also, I'm not in a position right now to contribute to the documentation any further than I have, nor am I really set up to. The first PR was actually created just using the GitHub web UI. So, if anyone else wants to mote this information to the Expressions page, don't hesitate.

PR #2137, opened 2023-01-02
Use FQCN for native attributes

TimWolla:
For attributes it's non-obvious that they need to be imported with `use` or referenced with their FQCN, because applying a non-existant attribute is not an error.

Make the examples copy-and-paste safe by adding the leading backslash, as already done for `#[\SensitiveParameterValue]`.

--------------------

See also this user note: https://www.php.net/manual/en/class.allow-dynamic-properties.php#128019

### DIFF
--- a/appendices/migration81/incompatible.xml
+++ b/appendices/migration81/incompatible.xml
@@ -121,7 +121,7 @@ ArgumentCountError - makeyogurt(): Argument #1 ($container) not passed
     inheritance validation.
     In case the return type cannot be declared for an overriding method due to
     PHP cross-version compatibility concerns,
-    a <code>#[ReturnTypeWillChange]</code> attribute can be added to silence
+    a <code>#[\ReturnTypeWillChange]</code> attribute can be added to silence
     the deprecation notice.
    </para>
   </sect3>
--- a/language/oop5/inheritance.xml
+++ b/language/oop5/inheritance.xml
@@ -123,7 +123,7 @@ $bar->printPHP();       // Output: 'PHP is great'
  
    <para>
     If the return type cannot be declared for an overriding method due to PHP cross-version compatibility concerns,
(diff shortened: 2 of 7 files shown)


appendices/migration81/incompatible.xml
```diff
@@ -121,7 +121,7 @@ ArgumentCountError - makeyogurt(): Argument #1 ($container) not passed
     inheritance validation.
     In case the return type cannot be declared for an overriding method due to
     PHP cross-version compatibility concerns,
-    a <code>#[ReturnTypeWillChange]</code> attribute can be added to silence
+    a <code>#[\ReturnTypeWillChange]</code> attribute can be added to silence
```
cmb69, line 124: What about:
```suggestion
    a <classname>ReturnTypeWillChange</classname> attribute can be added to silence
```

language/oop5/inheritance.xml
```diff
@@ -123,7 +123,7 @@ $bar->printPHP();       // Output: 'PHP is great'
  
    <para>
     If the return type cannot be declared for an overriding method due to PHP cross-version compatibility concerns,
-    a <code>#[ReturnTypeWillChange]</code> attribute can be added to silence the deprecation notice.
+    a <code>#[\ReturnTypeWillChange]</code> attribute can be added to silence the deprecation notice.
```
cmb69, line 126: ditto

cmb69 (commented):
Thank you! In my opinion, changing the examples and the other occurrence on the `ReturnTypeWillChange` page makes perfect sense, but I wonder about the two other cases.

appendices/migration81/incompatible.xml
TimWolla: This certainly makes sense, updated.

Girgias (approved):

PR #2141, opened 2023-01-04
Add missing class constants with empty description - part 1

kocsismate:


reference/ffi/ffi.xml
```diff
@@ -59,6 +67,17 @@
 
   </section>
 
+  <section xml:id="ffi-ffi.constants">
```
kocsismate, line 70: `ffi.constants` was already declared for the predefined constants page. Unfortunately, `ffi-ffi.constants` will result in a false positive diff between stubs, but that's the best I could came up with for now :/

kocsismate: Nevermind, I can use the `@link` phpdoc tag to customize the generated link :)

reference/ffi/ffi.xml
```diff
@@ -50,6 +50,14 @@
      </ooclass>
     </classsynopsisinfo>
 
+    <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
+    <fieldsynopsis>
+     <modifier>public</modifier>
+     <modifier>const</modifier>
+     <type>int</type>
+     <varname linkend="ffi-ffi.constants.--biggest-alignment--">FFI::__BIGGEST_ALIGNMENT__</varname>
```
cmb69, line 58: That ID looks awful; can we just trim leading and trailing underscores in genstub? (interestingly, docgen strips leading underscores, but keeps trailing ones)

cmb69 (commented):
Thank you! Looks generally good, but I'd like to improve the handling of leading/trailing underscores (see below).

reference/ffi/ffi.xml
kocsismate: I agree that it looks awful (just like the constant :D), but I'm worried to transform it because doing so might lead to name clashes :S E.g. what if we add a `FFI::BIGGEST_ALIGNMENT` as well?

cmb69: > E.g. what if we add a `FFI::BIGGEST_ALIGNMENT` as well?

I consider that unlikely, but I see your point. @Girgias, thoughts on this?

Girgias: I would rather trim the underscores, and if we are going to "fix" this constant it should be called ``LARGEST_ALIGNMENT`` anyway, so I don't think we'll get a conflict.

kocsismate: The question is not only about `FFI::__BIGGEST_ALIGNMENT__`, but if we generally want to trim the leading/trailing underscores from every property and constant?

kocsismate: Although I can still simply use the `@link` for this one. :) If we'll have more similar names then I'll adapt gen_stub as well

kocsismate, line 58:
```suggestion
     <varname linkend="ffi-ffi.constants.biggest-alignment">FFI::__BIGGEST_ALIGNMENT__</varname>
```

reference/ffi/ffi.xml
```diff
@@ -59,6 +67,17 @@
 
   </section>
 
+  <section xml:id="ffi-ffi.constants">
+   &reftitle.constants;
+   <variablelist>
+    <varlistentry xml:id="ffi-ffi.constants.--biggest-alignment--">
```
kocsismate, line 73:
```suggestion
    <varlistentry xml:id="ffi-ffi.constants.biggest-alignment">
```

cmb69 (approved):
Then I think we're good to go. Thank you!

The whitespace error is my bad; just fixed that via dbe9aea5415708d6e012f3d2297f204351209aea.

PR #2143, opened 2023-01-05
Change the Undefined variable error level

colshrapnel:


language/variables.xml
```diff
@@ -169,7 +169,7 @@ var_dump($unset_obj);
     Relying on the default value of an uninitialized variable is problematic
     in the case of including one file into another which uses the same
     variable name.
-    <link linkend="errorfunc.constants.errorlevels.e-notice">E_NOTICE</link> level error is issued in case of
+    <link linkend="errorfunc.constants.errorlevels.e-warning">E_WARNING</link> level error is issued in case of
```
cmb69, line 172: Ah, good catch. Thank you! However, [`E_WARNING` is only emitted as of PHP 8.0.0](https://3v4l.org/WhATD), and the manual is still supposed to document PHP 7 as well, so:
```suggestion
    <link linkend="errorfunc.constants.errorlevels.e-warning">E_WARNING</link>
    (prior to PHP 8.0.0, <link linkend="errorfunc.constants.errorlevels.e-notice">E_NOTICE</link>)
    level error is issued in case of
```

However, I'm not sure that the `<link>`s are needed; marking up as `<constant>` should be sufficient, i.e.
```suggestion
    <constant>E_WARNING</constant> (prior to PHP 8.0.0, <constant>E_NOTICE</constant>)
    level error is issued in case of
```

colshrapnel:
Thanks, indeed it would be better

colshrapnel:
Looks like I committed the wrong patch. Can you fix ut?

cmb69:
Fixed and merged. Thank you!

PR #2147, opened 2023-01-05
Add missing class constants with empty description - part 3

kocsismate:


reference/phar/Phar.xml
```diff
@@ -77,7 +159,124 @@
 <!-- }}} -->
  
   </section>
- 
+
+  <section xml:id="phar.constants">
```
cmb69, line 163: For `Phar`, the class constants are already documented at https://www.php.net/manual/en/phar.constants.php. Can we just remove this section, or would there be issues when we regenerate from the stubs?

reference/sqlite3/sqlite3.xml
```diff
@@ -42,6 +266,271 @@
 
   </section>
 
+
+  <section xml:id="sqlite3.constants">
```
cmb69, line 270: We already use `sqlite.constants` for [global constants defined by ext/sqlite3](https://www.php.net/manual/en/sqlite3.constants.php). Maybe we should change all these IDs to `sqlite3.class-constants` or so.

reference/zip/ziparchive.xml
```diff
@@ -72,61 +674,766 @@
...
+     <term><varname>comment</varname></term>
+     <listitem>
+      <para>Comment for the archive</para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </section>
+<!-- }}} -->
+
+  <section xml:id="ziparchive.constants">
```
cmb69, line 733: The `ZipArchive` class constants are already documented on the [constants page](https://www.php.net/manual/en/zip.constants.php). We could just drop this section (see above for potential issues), or we could move the documentation from that constants page. Having them on a separate page might be better, because that class documentation is already pretty large.

cmb69 (commented):
Thank you! Unfortunately, this needs quite some work; see my comments below.

reference/phar/Phar.xml
kocsismate: Good idea! Thankfully, doing so wouldn't be an issue when regenerating from stubs, because only the class/method synopsis is generated and then replaced.

kocsismate:
How does it look now? :)

reference/pdo/pdo.xml
```diff
@@ -30,6 +30,457 @@
      </ooclass>
     </classsynopsisinfo>
 
+    <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
+    <fieldsynopsis>
+     <modifier>public</modifier>
+     <modifier>const</modifier>
+     <type>int</type>
+     <varname linkend="pdo.constants.param-null">PDO::PARAM_NULL</varname>
```
Girgias, line 38: Quick question, why isn't it using the same pattern as for the SQLite3 where it has a ``class`` part in the anchor? Or why is SQLite3 using that pattern when the other classes don't seem to do that?
```suggestion
     <varname linkend="pdo.class.constants.param-null">PDO::PARAM_NULL</varname>
```

Girgias (commented):
Could you just remind me of the reason of changing the links? Is it to be able to render the page from the stubs?

Girgias (approved):

PR #2151, opened 2023-01-07
Remove unnecessary documentation

kocsismate:


reference/soap/soapserver.xml
```diff
@@ -34,11 +34,6 @@
     </classsynopsisinfo>
 
     <classsynopsisinfo role="comment">&Properties;</classsynopsisinfo>
-    <fieldsynopsis>
```
kocsismate, line 37: As far as I saw, there is no such property (not even a dynamic one). Even if there is one, it's private, so it's not a big loss not to document it.

reference/spl/splstack.xml
```diff
@@ -41,11 +41,6 @@
      <xi:fallback/>
     </xi:include>
 
-    <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
```
kocsismate, line 44: Only `SplStack::setIteratorMode()` is listed in this section, which is an inherited method from `SplDoublyLinkedList`. The latter file has all the necessary information for SplStacks (with a little bit of addition).

cmb69 (approved):
Thank you! Looks good to me (besides the minor wording issue).

reference/spl/spldoublylinkedlist/setiteratormode.xml
```diff
@@ -44,6 +44,13 @@
       <para>
        The default mode is: <constant>SplDoublyLinkedList::IT_MODE_FIFO</constant> | <constant>SplDoublyLinkedList::IT_MODE_KEEP</constant>
       </para>
+
+      <warning>
+       <para>
+        The direction of iteration can no longer be changed for <classname>SplStack</classname>s.
```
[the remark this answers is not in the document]
kocsismate: Nice catch! I've recently pushed a new commit to fix that as well.

PR #2152, opened 2023-01-08
Remove outdated ReflectionClass::getTraitAliases() null return description

othercorey:
This was dropped with the RETURN_THROWS conversions.


cmb69:
Thank you! This was not really outdated, though; the method only returned `null` in case of invalid parameters, in which case the behavior was undefined previously.

Issue #2156, opened 2023-01-09
`stdClass` is spelled in different cases: `StdClass` / `stdclass`


nsfisis:
Because class names in PHP are case-insensitive, `StdClass` and `stdclass` are exactly the same as `stdClass`. In the official manual, however, it is better to use consistent case.

```
$ git grep -hio stdclass | sort | uniq -c
  46 StdClass
 207 stdClass
  21 stdclass
```

cmb69:
Indeed. Thanks for reporting!

> 21 stdclass

Most of these are parts of IDs, and these are supposed to be lowercased.

PR #2157, opened 2023-01-09
Fix GH-2156: stdClass is spelled in different cases: StdClass / stdclass

cmb69:
Not sure if this commit should be marked with `[skip-revcheck]`. @mumumu, @Girgias, @saundefined, what do you think?

### DIFF
--- a/appendices/migration72/incompatible.xml
+++ b/appendices/migration72/incompatible.xml
@@ -135,7 +135,7 @@ var_dump(
     count(null), // NULL is not countable
     count(1), // integers are not countable
     count('abc'), // strings are not countable
-    count(new stdclass), // objects not implementing the Countable interface are not countable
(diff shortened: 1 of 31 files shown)


cmb69:
Yeah, that's my point. It might be simpler for translators to do a searh & replace, instead of walking through this commit.

Girgias:
Ah right, yeah using skip-revcheck then makes more sense.

mumumu:
Since example is the target in this case, I think it's better to match EN with other locales. I will follow this commit even if it is marked `[skip-revcheck]`.

saundefined:
> Since example is the target in this case, I think it's better to match EN with other locales. I will follow this commit even if it is marked as `[skip-revcheck]`.

Agree

cmb69:
Thanks all! So I've committed without `[skip-revcheck]`.

PR #2158, opened 2023-01-09
Make `check-whitespace.yml`check the final result only

TimWolla:


cmb69:
Thank you!

Would that new approach automatically squash and commit? If so, I'm not sure whether that is a good idea (personally, I prefer squashing only right before merging).

Also, would this only check modified lines? That was the intention of the original whitespace-checker (to avoid spurious whitespace changes, which are not nice for translators).

cc @hakre

TimWolla:
> Would that new approach automatically squash and commit? If so, I'm not sure whether that is a good idea (personally, I prefer squashing only right before merging).

It squash-merges in the temporary CI clone whenever the PR contents are modified …

> Also, would this only check modified lines? That was the intention of the original whitespace-checker (to avoid spurious whitespace changes, which are not nice for translators).

… and then checks the single resulting commit with the previous check (`--check` flag of `git log`).

So it behaves 100% identical, with the difference that it effectively evaluates the result after “Squash and merge”, instead of each commit.

cmb69 (approved):
Thanks for the explanation! Looks good to me then.

hakre:
looks good to me, `git log --check ...` does the standard git whitespace checks for the lines in the files of those revisions `HEAD^..HEAD`  would be on the squash commit then as I understand it, so fine.

Additionally `--check`  also looks for merge conflict markers which is sane, as you never want them in (and it once happened even here), just FYI, this one is for free.

And being late as review passed, some bits from looking at the changes:

- `git diff --check` -- does the same just on the changes. as the squash merge is already in the staging area (the cache), this would be `git diff --check --cached`.
- `git -P diff --check --cached` with no pager, no need to commit then.
- however if you need to do a commit but the user name/email is not configured, for a single commit, just prefix with the `USER` and `EMAIL` environment parameter, e.g. `USER=hk EMAIL="h@k" git commit -m "Squashed" - just FYI.

PR #2159, opened 2023-01-09
Add explicit `permissions` to GHA workflows

TimWolla:
see php/php-src@1d45ca58c86ac103ea624c3af3c81c20863950ad


cmb69 (approved):
Ah, thank you!

PR #2160, opened 2023-01-09
Add missing class constants with empty description - part 4

kocsismate:
Only 7 false positive class synopsis page diffs remain with the stubs after this PR!


cmb69 (approved):
Thank you!

PR #2165, opened 2023-01-10
Add SensitiveParameterValue documentation

TimWolla:
Some parts (long description for `getValue()` and general refining) are still missing, but having something is better than having nothing and these parts should allows folks to build onto it.

### DIFF
--- a/language/predefined/sensitiveparametervalue.xml
+++ b/language/predefined/sensitiveparametervalue.xml
@@ -9,7 +9,13 @@
   <section xml:id="sensitiveparametervalue.intro">
    &reftitle.intro;
    <para>
-
+    The <classname>SensitiveParameterValue</classname> class allows wrapping sensitive
+    values to protect them against accidental exposure.
+   </para>
+   <para>
+    Values of parameters having the <classname>SensitiveParameter</classname> attribute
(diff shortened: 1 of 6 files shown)


language/predefined/sensitiveparametervalue.xml
```diff
@@ -0,0 +1,101 @@
...
+    </varlistentry>
+   </variablelist>
+  </section>
+<!-- }}} -->
+
+ </partintro>
+
+ &language.predefined.sensitiveparametervalue.construct;
+ &language.predefined.sensitiveparametervalue.debuginfo;
+ &language.predefined.sensitiveparametervalue.getvalue;
```
TimWolla, line 79: FWIW: It's not great that no automated entity file is generated for classes outside of `reference/` :-(

cmb69: Indeed. Should be possible to also [generate `entities.*.xml` for `language/`](https://github.com/php/doc-base/blob/9965c2b16fbeba6d2451bd300d6da85260856de9/scripts/file-entities.php.in#L161).

cmb69 (approved):
Thank you! Looks like a sensible start.

PR #2170, opened 2023-01-12
Document the __toString() and export() related changes in ext/reflection

kocsismate:


reference/reflection/reflector.xml
```diff
@@ -36,10 +36,6 @@
       <classname>Stringable</classname>
      </ooclass>
     </classsynopsisinfo>
-<!-- }}} -->
-    
-    <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
```
kocsismate, line 41: Reflector doesn't have methods on its own: the `toString` method is inherited from `Stringable` (and we should set up a redirection), the export method has already been removed, so I linked in the "see also" section.

reference/reflection/reflector.xml
```diff
@@ -64,16 +60,30 @@
...
        <entry>
-        <classname>Reflector</classname> implements
+        <methodname>Reflector::export</methodname> was removed.
+       </entry>
+      </row>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <classname>Reflector</classname> extends
         <interfacename>Stringable</interfacename> now.
```
cmb69, line 70: I think we should tell users that `::__toString()` was previously declared on `Reflector`. Other than that, removing to `Reflector::__toString()` page makes sense to me, given that there was no real documentation.

cmb69 (commented):
Thank you! Besides the commented issue, this looks good to me.

cmb69 (approved):
> WDYT about my recent commit?

Perfect! :)

PR #2176, opened 2023-01-12
Improve random fruit examples

TimWolla:
- Add more elements to be able to select reasonable subsets.
- Don't use two differently-colored apples.
- Showcase two different selections.


cmb69 (approved):
Thank you!

derickr (approved):

PR #2177, opened 2023-01-13
Update cloning.xml

bradchesney79:
The example had a typo. Both variables were "object1" instead of "object2" being a clone. It is now.


cmb69:
This modification [would change the output of the example](https://3v4l.org/4IKLL). I think the example is okay-ish either way, so I wouldn't bother to change it (except for the comment above the cloning, which refers to a non-existing property).

kamil-tekiela:
Sorry, but `$this->object1` is correct. What you are proposing isn't right and not what the author intended.

PR #2181, opened 2023-01-14
Dynamic Properties - Clarify scope of deprecation

totten:
The phrasing in the manual is confusing/misleading. 

Let's go to the RFC that defined the current policy:

> "This RFC proposes to deprecate and later remove the creation of dynamic properties, unless the class explicitly allows dynamic properties."
>
> (https://wiki.php.net/rfc/deprecate_dynamic_properties)

The sentence structure is awkward but clear -- it allows for *explicitly* enabled dynamic-properties.  This interpretation aligns with the actual behavior that was implemented.  Compare:

1. At the time of the RFC's writing, there was implicit/universal support for dynamic-properties. PHP 8.2 respects this mechanism but generates a deprecation warning.
2. As a replacement, the RFC added a new explicit/opt-in mechanism for dynamic-properties (`AllowDynamicProperties`). PHP 8.2 was the first version to have this. It does not produce a deprecation warning.

The phrasing in the manual is confusing -- it says "dynamic properties are deprecated" without distinguishing between the truly deprecated mechanism (*the implicit/universal form*) and the new/supported mechanism (*the explicit/opt-in form*).


language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -11,6 +11,14 @@
     This attribute is used to mark classes that allow
     <link linkend="language.oop5.properties.dynamic-properties">dynamic properties</link>.
    </para>
+
+   <note>
+     <para>
+       <literal>AllowDynamicProperties</literal> was introduced in PHP 8.2.
+       In prior versions, all classes implicitly allowed dynamic properties.
+       Implicit dynamism is now deprecated.
```
kamil-tekiela, line 19:
```suggestion
       Implicit creation of dynamic properties is now deprecated.
```
I don't like this sentence.

language/oop5/properties.xml
```diff
@@ -337,13 +337,12 @@ $test->obj = new stdClass;
...
    <warning>
     <simpara>
-     Dynamic properties are deprecated as of PHP 8.2.0.
-     It is recommended to declare the property instead.
-     To handle arbitrary property names, the class should implement the magic
-     methods <link linkend="object.get">__get()</link> and
-     <link linkend="object.set">__set()</link>.
-     At last resort the class can be marked with the
-     <code>#[\AllowDynamicProperties]</code> attribute.
+     Prior to PHP 8.2, all objects implicitly supported dynamic
```
kamil-tekiela, line 340:
```suggestion
     Prior to PHP 8.2.0, all classes implicitly supported dynamic
```

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -38,35 +46,55 @@
...
+$obj->myDynamicProp = 1234;
+?>
+]]>
+    </programlisting>
+   </example>
+
+   <example>
+    <para>
+     On PHP 8.2+, the declaration is required. If you omit the declaration
+     and access an undeclared property, it will raise a warning.
```
kamil-tekiela, line 71:
```suggestion
     As of PHP 8.2.0, the declaration is required. If the declaration is omitted 
     and an undeclared property is accessed, PHP will raise a warning.
```

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -38,35 +46,55 @@
...
     <screen>
-<![CDATA[
-Deprecated: Creation of dynamic property DefaultBehaviour::$nonExistingProp is deprecated in file on line 10
+     <![CDATA[
+Deprecated: Creation of dynamic property InvalidExample::$myUnknownProperty is deprecated in file on line 5
 ]]>
     </screen>
+
+    <para>
+     In future versions, <literal>InvalidExample</literal> may create a hard failure.
```
kamil-tekiela, line 93: What are "future versions"? Please be specific.

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -38,35 +46,55 @@
...
+$obj->myUnknownProperty = 1234;
 ?>
 ]]>
     </programlisting>
+
     &example.outputs.82;
     <screen>
-<![CDATA[
-Deprecated: Creation of dynamic property DefaultBehaviour::$nonExistingProp is deprecated in file on line 10
+     <![CDATA[
```
kamil-tekiela, line 87:
```suggestion
<![CDATA[
```

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -38,35 +46,55 @@
...
+   <example>
+    <para>
+     On PHP 8.2+, the declaration is required. If you omit the declaration
+     and access an undeclared property, it will raise a warning.
+    </para>
 
-$o1 = new DefaultBehaviour();
-$o2 = new ClassAllowsDynamicProperties();
+    <programlisting role="php">
+     <![CDATA[
```
kamil-tekiela, line 75:
```suggestion
<![CDATA[
```

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -38,35 +46,55 @@
...
-Deprecated: Creation of dynamic property DefaultBehaviour::$nonExistingProp is deprecated in file on line 10
+     <![CDATA[
+Deprecated: Creation of dynamic property InvalidExample::$myUnknownProperty is deprecated in file on line 5
 ]]>
     </screen>
+
+    <para>
+     In future versions, <literal>InvalidExample</literal> may create a hard failure.
+    </para>
+
```
kamil-tekiela, line 95: What's the purpose of these empty lines?

kamil-tekiela (changes requested):
I think this is an overall improvement. Please see my nits though.

language/predefined/attributes/allowdynamicproperties.xml
totten: Oh, good point. I had the soft impression "It's probably going to be an error in 9.0" -- but the RFC does explicitly say this will be an `Error` in 9.0. Pushed up the change to make it clearer.

language/predefined/attributes/allowdynamicproperties.xml
totten: No particular purpose. It's my first time in the repo and still learning the dialect.

TBH, I skimmed a couple random files and couldn't discern a convention. (Sometimes `<para>`s are separated by newlines; sometimes they aren't.)

language/oop5/properties.xml
```diff
@@ -337,13 +337,12 @@ $test->obj = new stdClass;
...
-     methods <link linkend="object.get">__get()</link> and
-     <link linkend="object.set">__set()</link>.
-     At last resort the class can be marked with the
-     <code>#[\AllowDynamicProperties]</code> attribute.
+     Prior to PHP 8.2.0, all classes implicitly supported dynamic
+     properties. This is deprecated. Instead, dynamic
+     properties should be explicitly implemented with
+     <link linkend="allowdynamicproperties.intro">AllowDynamicProperties</link>
+     or with magic methods (<link linkend="object.get">__get()</link>
+     and <link linkend="object.set">__set()</link>).
```
cmb69, line 345: We should also mention that `stdClass` and its subclasses also support dynamic properties.

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -11,6 +11,14 @@
     This attribute is used to mark classes that allow
     <link linkend="language.oop5.properties.dynamic-properties">dynamic properties</link>.
    </para>
+
+   <note>
+     <para>
+       <literal>AllowDynamicProperties</literal> was introduced in PHP 8.2.
```
cmb69, line 17: This is already documented (just below the page title), and as such is superfluous.

language/predefined/attributes/allowdynamicproperties.xml
```diff
@@ -38,35 +46,55 @@
...
     <screen>
 <![CDATA[
-Deprecated: Creation of dynamic property DefaultBehaviour::$nonExistingProp is deprecated in file on line 10
+Deprecated: Creation of dynamic property InvalidExample::$myUnknownProperty is deprecated in file on line 5
 ]]>
     </screen>
+
+    <para>
+     In PHP 9.0.0, <literal>InvalidExample</literal> will throw an <literal>Error</literal> exception.
+    </para>
```
cmb69, line 94: We do not document what is supposed to happen in the future; it's just to uncertain. :)

cmb69 (commented):
Thanks for the PR!

language/predefined/attributes/allowdynamicproperties.xml
kamil-tekiela: Yeah, newlines can creep in easily when editing the file. I would advise avoiding adding them unless you are changing the structure of the file. It's no big deal, it just makes reviewing a little more confusing.

language/predefined/attributes/allowdynamicproperties.xml
kamil-tekiela: So, maybe remove this sentence and start the next one with "Prior to PHP 8.2.0"

Girgias (commented):
Haven't spent any time on the wording, but I dislike the current form as it seems to encourage users to just add the attribute, when this really should be the option of last resort.

I'll see if I can come up with better wording that I don't hate.

totten:
Added some updates to incorporate suggestions from @cmb69 and @kamil-tekiela.

Rebased/squashed a little bit to reduce some of the flipflopping on whitespace .

> (@Girgias) Haven't spent any time on the wording, but I dislike the current form as it seems to encourage users to just add the attribute, when this really should be the option of last resort.
>
> I'll see if I can come up with better wording that I don't hate.

That sounds like a high bar! 🙃

More seriously... I don't think it's a first resort or a last resort. As far as I can see, it's just a thing that the manual should describe accurately. If there are downsides to using it, then the manual should describe those.

Girgias:
> Added some updates to incorporate suggestions from @cmb69 and @kamil-tekiela.
> 
> Rebased/squashed a little bit to reduce some of the flipflopping on whitespace .
> 
> > (@Girgias) Haven't spent any time on the wording, but I dislike the current form as it seems to encourage users to just add the attribute, when this really should be the option of last resort.
> > I'll see if I can come up with better wording that I don't hate.
> 
> That sounds like a high bar! upside_down_face
> 
> More seriously... I don't think it's a first resort or a last resort. As far as I can see, it's just a thing that the manual should describe accurately. If there are downsides to using it, then the manual should describe those.

There are serious downsides to it, mainly performance and memory consumption. Moreover, dynamic properties makes understanding the relevant code hard and also bug prone as the engine (or static analysis) cannot catch typos in property names.

totten:
> There are serious downsides to it, mainly performance and memory consumption. 

OK. Let's start with that criterion -- performance/memory -- and go through the existing advice from `doc-en@HEAD`. Here's what it said:

>  ... To handle arbitrary property names, the class should implement the magic methods... At last resort the class can be marked with the `AllowDynamicProperties` attribute.

So magic-methods are suggested, and `AllowDynamicProperties` is the "last resort" -- because (presumably) `AllowDynamicProperties` provides worse performance?

But the presumption appears to be wrong. AFAICS, `AllowDynamicProperties` has *better* performance. I borrowed [a naive benchmark script that seemed popular on Stackexchange](https://stackoverflow.com/questions/2193049/php-objects-vs-arrays-performance-comparison-while-iterating) and [adapted it](https://gist.github.com/totten/2e345fe6960a68862fd7fb826cd9b012) to compare two classes:

```php
#[AllowDynamicProperties]
class DynamicClass {
}

class MagicClass {
    private $data = [];
    public function __get($name) {
      return $this->data[$name] ?? NULL;
    }
    public function __set($name, $value) {
      $this->data[$name] = $value;
    }
}
```

`DynamicClass` takes ~90ms and ~460k (*M1/macOS/php82*). `MagicClass` takes ~980ms and ~508k (*same configuration*).  980ms/90ms=~10.9 In other words, `MagicClass` is ten times slower than `DynamicClass`. (Memory usage for `MagicClass` is only 8% worse, though.)

>  Moreover, dynamic properties makes understanding the relevant code hard and also bug prone as the engine (or static analysis) cannot catch typos in property names.

Right, that's a fair knock against all dynamic/open models. (FWIW, I'm very glad that PHP has improved its support for static models by cleaning up the default behavior...)

But I don't see how this makes `#[AllowDynamicProperties]` into a "last resort". The same issues apply to models that use `array`, `stdClass`, `ArrayObject`, `__get()`, `__set()`, `__call()`, `call_user_func()`,the "map"/"bag" patterns, the "registry/locator/container" patterns, and so on -- they can all be hard to read; they all frustrate static-analysis; they're all susceptible to typos. It just feels a bit arbitrary to call-out `AllowDynamicProperties` for that.

PR #2186, opened 2023-01-15
Add missing curl_multi_getcontent() return null description

othercorey:


kamil-tekiela (approved):
It sounds right to me.

PR #2187, opened 2023-01-15
Add missing DOMCharacterData::appendData() return true description

othercorey:
The return type is bool not void and always returns true.

### DIFF
--- a/reference/dom/domcharacterdata/appenddata.xml
+++ b/reference/dom/domcharacterdata/appenddata.xml
@@ -35,7 +35,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.void;
+   Always returns &true;.
   </para>
  </refsect1>
  <refsect1 role="seealso">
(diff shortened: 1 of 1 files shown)


othercorey:
@cmb69 Can I poke you to finish these last PRs?

reference/dom/domcharacterdata/appenddata.xml
```diff
@@ -35,7 +35,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.void;
+   Always returns &true;.
```
Girgias, line 38:
```suggestion
   &return.true.always;
```

Girgias:
If this is the case the stubs in php-src should also be updated to reflect this.

othercorey:
The stub is type `bool`. Wasn't sure if everything currently returning `true` should be `true` in the stubs.

Girgias:
> The stub is type `bool`. Wasn't sure if everything currently returning `true` should be `true` in the stubs.

IMHO it should, do you want to make the PR to php-src?

tstarling:
Guess I should have rebased that manually. Apparently the change had already been made, and the commit was a no-op.

Girgias:
Ah yeah, must have gone through in one of the other stubs syncs.

PR #2190, opened 2023-01-15
Add missing DOMCharacterData::insertData() return false description

othercorey:


othercorey:
@cmb69 I will just close if these aren't acceptable. I assume you wanted more info. I was just dropping missing return descriptions.

reference/dom/domcharacterdata/insertdata.xml
```diff
@@ -44,7 +44,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.void;
+   Returns &false; on error.
```
cmb69, line 47: And `true` on success, I think, in which case we should have:
```suggestion
   &return.success;
```

cmb69 (commented):
Thank you!

reference/dom/domcharacterdata/insertdata.xml
othercorey: Updated all 3 DOMCharacterData PRs with this change.

othercorey:
@cmb69 I think these are all ready to go.

PR #2197, opened 2023-01-16
random: Add examples for the engine constructors

TimWolla:


reference/random/random/engine/pcgoneseq128xslrr64/construct.xml
```diff
@@ -97,16 +97,36 @@
...
+   <programlisting role="php">
+<![CDATA[
+<?php
+$string = "My string seed";
+
+// Hash the string with truncated SHA-256 using binary output
+// to turn the $string into a 128 Bit seed. Using the same
+// string will result in the same sequence of randomness.
+$e = new \Random\Engine\PcgOneseq128XslRr64(
+    substr( hash( 'sha256', $string, binary: true ), 0, 16 )
```
TimWolla, line 119: I considered using MD5 here, because it's still secure for this type of usage (the engine is not cryptographically safe, no matter the seed, especially not with a fixed seed), but this likely would be unnecessarily flagged by dumb vulnerability scanner.

zeriyoshi: What exactly would be the vulnerability scanner described here?

TimWolla: I'm not thinking about something specific here, but to the general class of dumb vulnerability scanners that effectively perform a textual search to report "security issues".

As an example, some companies might have a blanket ban for "using MD5", because "MD5 is insecure" without taking the context into account. So them using an MD5-based example to seed the engine would needlessly be flagged by such a scanner, possibly also causing issues to be filed about the documentation being "insecure".

zeriyoshi: I see. Thank you very much. That is certainly a possibility.
It might be safer not to combine the word of "security" and "MD5" :)

zeriyoshi (approved):

reference/random/random/engine/pcgoneseq128xslrr64/construct.xml
cmb69: Yeah, f6108da0b5a4be28358b1b958a54f2ad77085f68 …

cmb69 (approved):
Thank you!

PR #2204, opened 2023-01-17
Drop ldap_get_attributes() return false description

othercorey:


kamil-tekiela (approved):
Yeah, it looks like it never returned false.

cmb69:
It returned `false` prior to PHP 8.0.0, if an invalid resource (completely wrong type, or already closed) has been passed. This is actually the same as ZPP failure, so doesn't really need to be documented (generally "undefined behavior"), but I just pushed cf18a02ae8d0fca239cc876be841f15ee7b39771 to clarify a bit.

PR #2206, opened 2023-01-17
Remove some mysqli_driver properties and methods

kocsismate:


reference/mysqli/mysqli_driver.xml
```diff
@@ -52,26 +52,11 @@
      <type>int</type>
      <varname linkend="mysqli-driver.props.driver-version">driver_version</varname>
     </fieldsynopsis>
-    <fieldsynopsis>
```
kamil-tekiela, line 55: Are we not documenting old information in class synopsis anymore? Is it representing PHP 8.2 state only?

reference/mysqli/mysqli_driver.xml
```diff
@@ -113,6 +98,8 @@
      <term><varname>reconnect</varname></term>
      <listitem>
       <para>Allow or prevent reconnect (see the mysqli.reconnect INI directive)</para>
+      <warning><simpara>This property has
+       been <emphasis>removed</emphasis> as of PHP 8.0.0.</simpara></warning>
```
kamil-tekiela, line 102: It has been removed as of PHP 8.2.

kamil-tekiela: Also, it should be mentioned that the INI setting has also been removed. https://www.php.net/manual/en/mysqli.configuration.php#ini.mysqli.reconnect

reference/mysqli/mysqli_driver.xml
kocsismate: Good question! Syncronization with stubs only works well if class/method synopses represent the latest state only. As we always update all signatures with the recent PHP releases, IMO class signatures shouldn't be an exception. It's just very rare that we remove something in a minor release. BTW I kept the description of the removed properties. @cmb69 What do you think?

reference/mysqli/mysqli_driver.xml
kocsismate: Thanks for the suggestion! I pushed a commit

reference/mysqli/ini.xml
```diff
@@ -244,7 +244,9 @@
      Automatically reconnect if the connection was lost.
     </para>
     <note>
-     <simpara>This &php.ini; setting is ignored by the mysqlnd driver.</simpara>
+     <simpara>This &php.ini; setting had been ignored by the mysqlnd driver as of PHP 7.0.0, and was removed
```
kocsismate, line 247: @kamil-tekiela Do you know since when the option is ignored? Does this description make sense or do you have a better suggestion?

kamil-tekiela: Since always. It was never part of the spec and it was never implemented. It only existed because of libmysql that had this feature. When mysqli was compiled with mysqlnd, it was simply ignored. You can't compile mysqli with libmysql anymore so this setting is gone now too.

reference/mysqli/ini.xml
```diff
@@ -244,7 +244,9 @@
...
+      as of PHP 8.2.0.
```
cmb69, line 248: So I think we should have:
```suggestion
     <simpara>This &php.ini; setting had been ignored by the mysqlnd driver, and was removed
      as of PHP 8.2.0.
```

cmb69 (approved):
Thank you! Besides the nit below, this looks good to me.

reference/mysqli/ini.xml
kamil-tekiela: Or invert the sentence and say "This setting has been only applicable with libmysql driver, and was removed as of PHP 8.2.0."

reference/mysqli/mysqli_driver.xml
```diff
@@ -131,6 +121,42 @@
...
+       <entry>8.1.0</entry>
+       <entry>
+        <property>mysqli_driver::$driver_version</property> has been deprecated.
+       </entry>
+      </row>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        <property>mysqli_driver::$embedded</property> and
+        <property>mysqli_driver::$reconnect</property> were removed.
```
kamil-tekiela, line 146: Please don't forget to adjust this.

kocsismate: Done! Thanks for the remainder!

kamil-tekiela (approved):
I think it looks fine now. @cmb69 ?

cmb69:
Yes, looks good.

PR #2208, opened 2023-01-17
Drop SplFileObject::fgets() return false description

othercorey:
This was converted to an exception, but looks like that's already in the errors section.


kamil-tekiela (approved):
LGTM. Thanks!

PR #2210, opened 2023-01-17
Drop xml_get_current_byte_index() return false description

othercorey:
This was dropped during conversion to object in https://github.com/php/php-src/commit/e35a3cb28cd663489b27630e4dca2a5e61898f2e so no new exceptions were added that require a changelog.


cmb69:
Thank you!

PR #2212, opened 2023-01-17
Drop xml_get_current_line_number() return false description

othercorey:
Same reason as https://github.com/php/doc-en/pull/2210


cmb69:
Thank you!

PR #2214, opened 2023-01-17
Drop xml_parser_get_option() return false description

othercorey:
Same reason as https://github.com/php/doc-en/pull/2210


othercorey:
@cmb69 This goes with the others you merged.

cmb69:
> This goes with the others you merged.

But this should document the `ValueError` which is thrown in case the `$option` is invalid, and also have a changelog entry about the change as of PHP 8.0.0 (previously, it emitted `E_WARNING` and returned `false`).

othercorey:
Ah, I missed this changed further. Th commit I referenced didn't fully remove return false for this function anyway.

PR #2216, opened 2023-01-17
random: Drop the range constraints from getInt()/random_int()

TimWolla:
It’s impossible to pass an integer that does not fall into the documented range, making this useless information. Integers outside these ranges will be converted into floats, making those a TypeError.


cmb69 (approved):
Thank you!

Issue #2185, opened 2023-01-15
spl_autoload_functions missing changelog for return type


othercorey:
In https://github.com/php/php-src/commit/5b59d4915cfa7c12ddf65281e4f7f81d6985d27, the false return type was dropped. I don't see any new errors thrown so the error conditions must have been eliminated.

PR #2217, opened 2023-01-18
Added changelog to reflect changed return type of spl_autoload_functions

ollieread:
I've updated the return description and added a changelog for `spl_autoload_functions` to reflect the changes made in php/php-src@5b59d49.

This closes php/doc-en#2185 and ollieread/php-doc-en#14.


kamil-tekiela (approved):
LGTM. Thanks!

PR #2222, opened 2023-01-19
Fix example in str_ends_with()

ProtocolLive:


TimWolla (approved):
Thank you! I've also checked the examples in `str_starts_with()` and `str_contains()` and they are already good.

PR #2223, opened 2023-01-19
Improve documentation for mysqli_change_user

kamil-tekiela:
I opened a bug report https://github.com/php/php-src/issues/10382 but then I realized that I was actually confused by the documentation. I tried to explain it better and also provide clearer code examples. The function works how it was designed, but the documentation was unclear.

### DIFF
--- a/reference/mysqli/mysqli/change-user.xml
+++ b/reference/mysqli/mysqli/change-user.xml
@@ -4,7 +4,7 @@
  <refnamediv>
   <refname>mysqli::change_user</refname>
   <refname>mysqli_change_user</refname>
-  <refpurpose>Changes the user of the specified database connection</refpurpose>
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli/change-user.xml
```diff
@@ -84,89 +83,82 @@
  <refsect1 role="examples">
   &reftitle.examples;
   <example>
-   <title><methodname>mysqli::change_user</methodname> example</title>
+   <title>The following example shows how this function resets the connection session</title>
```
jimwins, line 86:
```suggestion
   <title>Resetting the connection session</title>
```

I don't think the `<title>` should ready as a sentence like that.

jimwins (changes requested):

jimwins (approved):
LGTM.

PR #2235, opened 2023-01-20
Automatic leading dot in setcookie.xml

fullbl:
PHP adds a leading . when using a subdomain in setcookie

### DIFF
--- a/reference/network/functions/setcookie.xml
+++ b/reference/network/functions/setcookie.xml
@@ -120,6 +120,9 @@
        name (<literal>'example.com'</literal>, in this case).
       </para>
       <para>
+       PHP will add a leading <literal>.</literal> when using a subdomain as domain.
+      </para>
+      <para>
        Older browsers still implementing the deprecated
        <link xlink:href="&url.rfc;2109">RFC 2109</link> may require a leading
        <literal>.</literal> to match all subdomains.


reference/network/functions/setcookie.xml
```diff
@@ -119,6 +119,9 @@
        (including all subdomains of it), simply set the value to the domain
        name (<literal>'example.com'</literal>, in this case).
       </para>
+      <para>
+       PHP will add a leading <literal>.</literal> when using a subdomain as domain.
```
TimWolla, line 123: What does “subdomain” refer to here? When is the given domain not a subdomain – when will PHP not add a leading `.`?

fullbl: A subdomain is a domain that is a part of another domain.

If you are browsing sub.domain.com:
if you set the cookie domain as sub.domain.com, php will set it as .sub.domain.com
if you set the cookie domain as domain.com, php will set it as domain.com
if you leave it blank, php will set it as sub.domain.com

TimWolla: So it's based on the hostname that is accessed? I'm not able to reproduce this, PHP always emits the `domain` value as it is given in the program code and the source code does not contain any such dot-prepending logic either, as far as I can tell.

cmb69: I wonder whether this is supposed to be a feature request.

fullbl: I don't know if it's allowed, but I put a test on a subdomain I own. This is the code:

```
<?php

setcookie('domain', 'hi', 0, '/', 'fullbl.it');
setcookie('domain-dot', 'hi', 0, '/', '.fullbl.it');
setcookie('subdomain', 'hi', 0, '/', 'shop.fullbl.it');
setcookie('subdomain-dot', 'hi', 0, '/', '.shop.fullbl.it');
setcookie('empty', 'hi', 0, '/');
```

This is what you find on domain shop.fullbl.it (PHP 8.1.13):
```
domain: .fullbl.it
domain-dot: .fullbl-it
subdomain: .shop.fullbl.it
subdomain-dot: .shop.fullbl.it
empty: shop.fullbl.it
```

So you can see it for yourself: the only way to set the subdomain without dot is leaving it empty!

tstarling: https://shop.fullbl.it/cookies.php seems to be a 404 now, but here's what I get with the script above:

```
$ curl -kI '.../setcookie-dot.php'
HTTP/1.1 200 OK
Date: Tue, 06 Jun 2023 04:28:52 GMT
Server: Apache/2.4.52 (Ubuntu)
X-Powered-By: PHP/8.1.18
Set-Cookie: domain=hi; path=/; domain=fullbl.it
Set-Cookie: domain-dot=hi; path=/; domain=.fullbl.it
Set-Cookie: subdomain=hi; path=/; domain=shop.fullbl.it
Set-Cookie: subdomain-dot=hi; path=/; domain=.shop.fullbl.it
Set-Cookie: empty=hi; path=/
Vary: Cookie
Content-Type: text/html; charset=UTF-8
```

So, PHP is not adding a leading dot. Probably some other thing adds the dot for @fullbl.

tstarling:
It looks like @fullbl is confusing client behaviour with server behaviour.

reference/network/functions/setcookie.xml
fullbl: Thanks, actually both firefox and chromium browsers add the dot. Don't know why I didn't check it with curl!

Issue #2142, opened 2023-01-05
Traits Conflict Resolution Example Error


Sunthief:
From manual page: https://php.net/language.oop5.traits
---
Example # 13 Conflict Resolution:
```
trait PropertiesTrait {
    public $same = true;
    public $different1 = false;
    public bool $different2;
    readonly public bool $different3;
}

class PropertiesExample {
    use PropertiesTrait;
    public $same = true;
    public $different1 = true; // Fatal error
    public string $different2; // Fatal error
    readonly public bool $different3; // Fatal error
}
```
The last example `readonly public bool $different3; // Fatal error`  does in fact not create a fatal error as both properties in the trait and class are exactly the same and thus compatible.

cmb69:
Indeed, that comment is wrong. I assume the intention was to have the `readonly` only on the class or on the trait. @Girgias?

Girgias:
Ah yes otherwise this doesn't make a lot of sense.

PR #2237, opened 2023-01-20
Fix: #2142 Traits conflict resolution example error

syl20b:
Hi, 

this should fix this issue #2142 

If it’s OK for you, I can do the same for the others languages.

Regards,

Sylvain


Girgias (approved):

Girgias:
Thank you!

PR #2240, opened 2023-01-21
tmpfile() uses binary mode.

divinity76:
ref https://github.com/php/php-src/blob/master/main/php_open_temporary_file.c#L103


cmb69:
Ah, right! Thank you!

(It might be a good idea to get rid of "t" mode altogether; https://wiki.php.net/rfc/deprecations_php_8_1?s[]=binary&s[]=text#t_fopen_mode was declined, though.)

PR #2243, opened 2023-01-22
Code style fixes

kamil-tekiela:


cmb69 (commented):
Removing the parens around the `echo` arguments is a good idea, but I'm not sure if we should touch the cubrid examples, without removing the HTML cruft, and replacing `each()` with a `foreach` loop (and may some more).

kamil-tekiela:
I think you are right. I will merge only the first commit and close this PR.

PR #2247, opened 2023-01-23
random: Update list of CSPRNG sources

TimWolla:
The result is absolutely not pretty, but it should hopefully be accurate for all PHP versions supported by the manual (i.e. >= 7.0):

image

It's a little messy, because new sources have often been added in patch versions. At least the NetBSD getrandom(2) support will only appear starting in 8.3.


TimWolla:
I would also be fine with only documenting the current (PHP 8.2) state which should simplify this quite a bit.

PR #2249, opened 2023-01-24
random: Drop example section from the `generate()` method of the Engines

TimWolla:
This method is not meant to be used directly and showcasing some random bytes of fixed length does not provide a value-add for the reader.


cmb69 (approved):
Makes sense. Thank you!

PR #2255, opened 2023-01-27
mysqli functions that always return true

Parsonswy:
Moved from #1413.

Reference;
[`mysqli_close`](https://github.com/php/php-src/blob/master/ext/mysqli/mysqli_api.c#L282)
[`mysqli_stmt_close`](https://github.com/php/php-src/blob/master/ext/mysqli/mysqli_api.c#L1480)
[`mysqli_field_seek`](https://github.com/php/php-src/blob/master/ext/mysqli/mysqli_api.c#L810)

### DIFF
--- a/reference/mysqli/mysqli_result/field-seek.xml
+++ b/reference/mysqli/mysqli_result/field-seek.xml
@@ -57,6 +57,30 @@
   </para>
  </refsect1>
 
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <para>
+   <informaltable>
+    <tgroup cols="2">
+     <thead>
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function now always returns &true;. Previously it returned &false; on failure.
+       </entry>
+      </row>
+     </tbody>
+    </tgroup>
+   </informaltable>
+  </para>
+ </refsect1>
+
  <refsect1 role="examples">
   &reftitle.examples;
   <example>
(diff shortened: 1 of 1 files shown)


cmb69:
Hmm, prior to PHP 8.0.0, these functions emitted a warning and returned `null` or `false` on failure; as of PHP 8.0.0, in that case an `Error` is thrown. Not sure how to handle that in the docs; a `TypeError` would not be documented at all; a `ValueError` would, though. This seems more related to `ValueErrors` – at least the `false` return.

Girgias:
Needs to be rebases, and a Changelog section added to describe the change in behaviour between PHP 7 and 8

reference/mysqli/mysqli_result/field-seek.xml
```diff
@@ -53,7 +53,31 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.success;
+   &return.true.always;
```
kamil-tekiela, line 56: Strange, but in php-src it is still marked as return type bool

Girgias: This should be fixed then, and have a changelog added, although the entity I created (``return.type.true``) for that is specific to 8.2.0... which was probably a bit of a mistake

Girgias: Probably makes sense to actually add that changelog to the 2 other functions if it was enacted in 8.2

Girgias (approved):

reference/mysqli/mysqli_result/field-seek.xml
kamil-tekiela: The *value* has changed in 8.0, but the return *type* has changed in 8.2. I will follow up on the return type separately.

PR #2259, opened 2023-01-29
Modernize note about session_unset

kamil-tekiela:
I believe this was written back when PHP 5.4 was released. The old note isn't very clear now unless someone remembers the old days. However, the note still makes sense on the `session_unset` page with a clear explanation of why using it may be problematic. 

P.S. Help wanted for a good example code for this page.


Girgias (approved):
Looks sensible? But yeah the wonders of ext-session.

PR #2261, opened 2023-01-30
random: Showcase turning picked array keys into values

TimWolla:
see php/php-src#10412


cmb69 (approved):
That example makes sense. Thank you!

PR #2262, opened 2023-01-30
Update variables.xml

odsantos:
Suggesting a text change. Thanks.


kamil-tekiela:
Honestly, I don't know about this one. It looks like an improvement, but that sentence (nay, the whole page) is very confusing.

Girgias:
Yeah, I'm not convinced either about this change.

This page is pretty complex and better trying to reword it properly then drop a word here or there.

PR #2264, opened 2023-02-02
Rewrite Sql injection page

kamil-tekiela:
The page was full of bloat and unclear text. The most important thing: prepared statements wasn't even given a good example. This PR attempts to improve this page by rewrite large chunks of it, moving some information around, and emphasising important points. 

@colshrapnel You might be interested in reviewing this PR too.

### DIFF
--- a/security/database.xml
+++ b/security/database.xml
@@ -350,7 +350,7 @@ $result = mssql_query($query);
     <sect2 xml:id="security.database.avoiding">
      <title>Avoidance Techniques</title>
      <para>
-      The recommended way to avoid SQL injection is by binding all data via the
+      The recommended way to avoid SQL injection is by binding all data via
       prepared statements. Using parameterized queries isn't enough to entirely
       avoid SQL injection, but it is the easiest and safest way to provide input
       to SQL statements. All dynamic data literals in <literal>WHERE</literal>,
@@ -359,7 +359,7 @@ $result = mssql_query($query);
       execution and sent separately from the SQL command.
      </para>
      <para>
-      Parameter binding can be used only for data. Other dynamic parts of the
+      Parameter binding can only be used for data. Other dynamic parts of the
       SQL query must be filtered against a known list of allowed values.
      </para>
      <para>
@@ -382,6 +382,14 @@ $stmt->execute(["%$productId%"]);
        </programlisting>
       </example>
      </para>
+
+     <simpara>
+      Prepared statements are provided
+      <link linkend="pdo.prepared-statements">by PDO</link>,
+      <link linkend="mysqli.quickstart.prepared-statements">by MySQLi</link>,
+      and by other database libraries.
+     </simpara>
+
      <simpara>
       SQL injection attacks are mainly based on exploiting the code not being written
       with security in mind. Never trust any input, especially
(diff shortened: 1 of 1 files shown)


security/database.xml
```diff
@@ -349,114 +349,113 @@ $result = mssql_query($query);
...
-// please note %d in the format string, using %s would be meaningless
-$query = sprintf("SELECT id, name FROM products ORDER BY name LIMIT 20 OFFSET %d;",
-                 $offset);
+// The dynamic SQL part is validated against expected values
+$sortingOrder = $_GET['sortingOrder'] === 'DESC' ? 'DESC' : 'ASC';
+$productId = $_GET['productId'];
+// The SQL is prepared with a placeholder
+$stmt = $pdo->prepare("SELECT * FROM products WHERE id LIKE ? ORDER BY price $sortingOrder");
+// The value is provided with LIKE wildcards
+$stmt->execute(["%$productId%"]);
```
TimWolla, line 378:
```suggestion
$stmt = $pdo->prepare("SELECT * FROM products WHERE id LIKE ? ORDER BY price {$sortingOrder}");
// The value is provided with LIKE wildcards
$stmt->execute(["%{$productId}%"]);
```
I'd use this type of interpolation in the "good" example, as it makes the locations where a variable is inserted much more obvious. I believe it should be preferred for that reason.

security/database.xml
```diff
@@ -349,114 +349,113 @@ $result = mssql_query($query);
...
+         in <link linkend="ref.ctype">Character Type Functions</link>
+         (e.g. <function>is_numeric</function>, <function>ctype_digit</function>
+         respectively) and onwards to the
+         <link linkend="ref.pcre">Perl Compatible Regular Expressions</link>
+         support.
+        </simpara>
+       </listitem>
+       <listitem>
+        <simpara>
+         If the application waits for numerical input, consider verifying data
```
TimWolla, line 424:
```suggestion
         If the application expects numerical input, consider verifying data
```

security/database.xml
```diff
@@ -349,114 +349,113 @@ $result = mssql_query($query);
...
+         using <function>settype</function>, or use its numeric representation
+         by <function>sprintf</function>.
+        </simpara>
+       </listitem>
+       <listitem>
+        <simpara>
+         If the database layer doesn't support binding variables then
+         quote each non-numeric user-supplied value that is passed to the
+         database with the database-specific string escape function (e.g.
+         <function>mysql_real_escape_string</function>,
```
TimWolla, line 435: I believe we can drop that. ext/mysql is luckily gone.

TimWolla (commented):
Some comments, but looking at the diff this is a clear improvement.

security/database.xml
kamil-tekiela: Last time I checked it was still available for PHP 8.0. cmb69 said that we still keep stuff that works with PHP 7.0, so I would keep it. The sqlite function could be removed, but I didn't know what to replace it with.

TimWolla: Then the docs are incorrect: https://www.php.net/manual/en/function.mysql-real-escape-string.php "This extension was deprecated in PHP 5.5.0, and it was removed in PHP 7.0.0"

kamil-tekiela: @cmb69 Correct me if I am wrong, but when an extension is removed from PHP we document it like it is now, but we do not mention that it has been moved to PECL, correct?

Girgias: Normally, the fact it lives in PECL should be reflected via the ``versions.xml`` file.

However, in this context it may make sense to not point to the old MySQL extension but to ``mysqli`` if applicable?

Girgias (approved):

security/database.xml
kamil-tekiela: Ok, so I am not sure what to do with this paragraph. I would keep it as it is now. Yes, the sqlite function is long gone and the mysql_* API is in Pecl only but the idea here is to let users know what to do when they are using these very old extensions.

Girgias: I think it's fine, we can always fix this later.

PR #2267, opened 2023-02-02
Fix minor typo in cloning.xml leading paragraph

xSavitar:
The leading paragraph didn't quite read well due to the typo. This  patch fixes it.

### DIFF
--- a/language/oop5/cloning.xml
+++ b/language/oop5/cloning.xml
@@ -5,14 +5,14 @@
   
   <para>
    Creating a copy of an object with fully replicated properties is not
-   always the wanted behavior. A good example of the need for copy
-   constructors, is if you have an object which represents a GTK window and the
-   object holds the resource of this GTK window, when you create a duplicate
-   you might want to create a new window with the same properties and have the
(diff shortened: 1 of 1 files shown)


TimWolla (commented):
I do not believe this change is correct. I believe "copy" is not a verb here and instead is part of the compound noun "copy constructor".

xSavitar:
Compound nouns must be really interesting then :). It didn't quite read well to me, but if you say it's correct then that is fine. Thank you for reviewing.

Could you link me to some examples of how "copy" is used as a compound noun? Thanks!

TimWolla:
> Could you link me to some examples of how "copy" is used as a compound noun? Thanks!

https://en.wikipedia.org/wiki/Copy_constructor_(C%2B%2B) is likely an appropriate reference for the concept.

xSavitar:
> > Could you link me to some examples of how "copy" is used as a compound noun? Thanks!
> 
> https://en.wikipedia.org/wiki/Copy_constructor_(C%2B%2B) is likely an appropriate reference for the concept of a “copy constructor”.

Thank you very much!

PR #2270, opened 2023-02-05
Remove curl context

KapitanOczywisty:
Removed in [PHP 5.5](https://wiki.php.net/rfc/curl-wrappers-removal-rfc) and seems like never made into pecl.


Girgias:
Makes sense. Thank you!

PR #2271, opened 2023-02-06
Fix typo in php8 example

lode:
The php8 example contains an additional `]` at the end. Probably a mistake.

afbeelding


kamil-tekiela (approved):
Indeed looks like a typo. Thanks!

PR #2274, opened 2023-02-07
Document `Spoofchecker::setRestrictionLevel`

MatTheCat:
Part of #2163

Stumbled upon this (amongst other things) while implementing [Symfony’s `NoSuspiciousCharacters` constraint](https://symfony.com/doc/current/reference/constraints/NoSuspiciousCharacters.html).

### DIFF
--- a/reference/intl/spoofchecker/setrestrictionlevel.xml
+++ b/reference/intl/spoofchecker/setrestrictionlevel.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="spoofchecker.setrestrictionlevel" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>Spoofchecker::setRestrictionLevel</refname>
+  <refpurpose>Set the restriction level</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis role="Spoofchecker">
+   <modifier>public</modifier> <type>void</type><methodname>Spoofchecker::setRestrictionLevel</methodname>
+   <methodparam><type>int</type><parameter>restrictionLevel</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Sets the restriction level of <methodname>SpoofChecker::isSuspicious</methodname>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>restrictionLevel</parameter></term>
+    <listitem>
+     <para>
+      The restriction level of <methodname>SpoofChecker::isSuspicious</methodname>.
+      One of
+      <constant>Spoofchecker::ASCII</constant>,
+      <constant>Spoofchecker::SINGLE_SCRIPT_RESTRICTIVE</constant>,
+      <constant>Spoofchecker::HIGHLY_RESTRICTIVE</constant>,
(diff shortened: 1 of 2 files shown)


tcitworld (commented):
Came here to contribute this very function, looks good.

reference/intl/spoofchecker/setrestrictionlevel.xml
```diff
@@ -0,0 +1,70 @@
...
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>restrictionLevel</parameter></term>
+    <listitem>
+     <para>
+      The restriction level of <methodname>SpoofChecker::isSuspicious</methodname>.
+      One of
+      <constant>Spoofchecker::ASCII</constant>,
```
kocsismate, line 29: It would be nice to describe what the actual codes actually do (not necessarily here, but e.g. on the class synopsis page where they already have a(n empty) section.

reference/intl/spoofchecker/setrestrictionlevel.xml
```diff
@@ -0,0 +1,70 @@
...
+  </variablelist>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   &return.void;
+  </para>
+ </refsect1>
+
```
kocsismate, line 48: nit: looks like an unnecessary newline

reference/intl/spoofchecker/setrestrictionlevel.xml
```diff
@@ -0,0 +1,70 @@
...
+ <refnamediv>
+  <refname>Spoofchecker::setRestrictionLevel</refname>
+  <refpurpose>Set the restriction level</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis role="Spoofchecker">
+   <modifier>public</modifier> <type>void</type><methodname>Spoofchecker::setRestrictionLevel</methodname>
+   <methodparam><type>int</type><parameter>restrictionLevel</parameter></methodparam>
```
kocsismate, line 13: it also has a second, optional parameter passed by reference, called `$errorCode`.

In order to avoid inconsistencies like this, it is very much suggested to run `./build/gen_stub.php --replace-methodsynopses ./ TARGET_DIR` from php-src (https://github.com/kocsismate/php-src/blob/cc2a68e5882e96bdc545588a75c56748860ce808/build/gen_stub.php) where the target directory refers to a path in doc-en which you want to update.

MatTheCat: > it also has a second, optional parameter passed by reference, called `$errorCode`.

I couldn’t find any source to check this; aren’t you confusing `setRestrictionLevel` with `isSuspicious` or `areConfusable`?

reference/intl/spoofchecker/setrestrictionlevel.xml
MatTheCat: Indeed. This can be found at https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/text/SpoofChecker.RestrictionLevel.html#enum.constant.summary e.g.

reference/intl/spoofchecker/setrestrictionlevel.xml
MatTheCat: I can remove one or the two empty lines, but I’m not sure this would make sense as there is no consistency in the docs :confused:

reference/intl/spoofchecker/setrestrictionlevel.xml
kocsismate: Ah sorry, I somehow indeed confused the two functions 😵

PR #2277, opened 2023-02-08
timezones.xml - Kyiv changed to Kiev

tomem92:
tz data says the timezone is `Europe/Kiev`. also `\DateTimeZone::listIdentifiers()` function list that timezone as `Europe/Kiev`, so doc is missleading

### DIFF
--- a/reference/datetime/timezones.xml
+++ b/reference/datetime/timezones.xml
@@ -605,7 +605,7 @@
      <row>
       <entry>Europe/Kaliningrad</entry>
       <entry>Europe/Kirov</entry>
-      <entry>Europe/Kyiv</entry>
+      <entry>Europe/Kiev</entry>
       <entry>Europe/Lisbon</entry>
      </row>
      <row>


tomem92:
quick test:
`echo \DateTimeZone::listIdentifiers()[336];`

Danack:
The top of the file says:

>  AUTO GENERATED, DO NOT TRANSLATE OR MODIFY BY HAND 

Though inconveniently doesn't say how to generate it...I'll ask around...

derickr:
@tomem92 You need to upgrade your timezone database, either by upgrading PHP, or installing PECL timezonedb: pecl.php.net/timezonedb

tomem92:
@derickr you're right. After `timezonedb` update I have the same as in the doc.
also double checked tz data raw files and in fact `Europe/Kiev` is an alias to `Europe/Kyiv` in the current tz db.

BTW. Thank you very much for Xdebug :star2: :beer: :beers:

PR #2278, opened 2023-02-08
Fixes iterator_to_array parameter name

rodrigopedra:
`iterator_to_array()` second parameter's name is `preserve_keys` and not `use_keys` as it is currently stated.

This is useful when using named parameters.


kamil-tekiela (approved):
That's absolutely true. Thanks!

PR #2286, opened 2023-02-09
Fix broken method reference, `push` doesn't exist

leinster:
There is no `push` method, `add` seems equivalent here.


Girgias (approved):
Indeed! Thank you, and apologies for the delay.

PR #2290, opened 2023-02-12
Fix sort true return type

othercorey:
These were updated to the new `true`, they could have never returned false according to https://github.com/php/php-src/commit/e86cdce549c9e81847584d2caa2dff7f89eac682


Girgias:
This should also apply to the SPL array methods.

Girgias:
Actually this is covered by #2154 already

PR #2296, opened 2023-02-14
Add format RFC 7231 warning

jorgsowa:
Format RFC 7231 doesn't take into consideration the timezone of the DateTime object. It shows always `GMT` designation. It should be reflected in documentation.

### DIFF
--- a/reference/datetime/datetimeinterface.xml
+++ b/reference/datetime/datetimeinterface.xml
@@ -260,6 +260,11 @@
        <simpara>
         RFC 7231 (since PHP 7.0.19 and 7.1.5) (example: Sat, 30 Apr 2016 17:52:13 GMT)
        </simpara>
+      <note>
+       <simpara>
+        This format ignores the given timezone and always displays hardcoded GMT timezone designation.
+       </simpara>
+      </note>
       </listitem>
     </varlistentry>
 


derickr:
Hmm, I am not sure if this is wise. The implementation **discards** the timezone, which isn't quite the same as showing it in **GMT**. The constant is therefore a little misleading.

derickr:
I would argue that this [patch](https://github.com/php/php-src/commit/618d97962a8ac59d981c8307cc704790f46090c8) in php-src is therefore wrong.

jorgsowa:
Hey @derickr. What do you think about merging this warning and implement fix for the constant later. This way developers are informed until the fix is ready.

derickr:
I think that the text should indicate that this is currently a known bug, and you should also indent your insertion right (misses an extra space).

derickr:
The bug is that the constant just makes it show as it is UTC/GMT without actually adjusting the timezone of the DateTime object that is being formatted.

Crell:
What are the next steps here, if any?  Should we close this PR or how should it be revised/corrected?

reference/datetime/datetimeinterface.xml
```diff
@@ -272,23 +254,22 @@
...
-      </term>
+      <term><constant>DateTimeInterface::RFC7231</constant></term>
       <term><constant>DATE_RFC7231</constant></term>
       <listitem>
        <simpara>
         RFC 7231 (since PHP 7.0.19 and 7.1.5) (example: Sat, 30 Apr 2016 17:52:13 GMT)
        </simpara>
+       <note>
+        <simpara>
+         This format ignores the given timezone and always displays hardcoded GMT timezone designation.
```
derickr, line 265: This section needs to now not be here, but instead the "Changelog" section should mention that this constant is deprecated in 8.5.0

derickr (changes requested):

PR #2300, opened 2023-02-15
[ibm_db2] Update param and return type for `db2_autocommit()`

phansys:
Prior to version ibm_db2 2.1, the `db2_autocommit()` function was expecting a boolean as argument 2.

See:
* https://bugs.php.net/bug.php?id=77591
* https://github.com/php/pecl-database-ibm_db2/commit/e9fb4b173046ab4fa1bb69125adf3de8d3818b97

Additionally, the return type of this function can be `int|bool` depending on the received arguments, so `mixed` was narrowed to `int|bool`.


Girgias:
Closing in favour of #2316

PR #2301, opened 2023-02-16
Don't mention sorting but comparison

DanielSiepmann:
The array_unique function doesn't sort, but compares. It uses the same constants but terms are mixed up. The text uses "sorting" while each constant uses "compare".

This is now streamlined in order to reduce friction while reading.


TimWolla:
Thank you!

PR #2306, opened 2023-02-17
[PDO] Fix param name "cursorOffset" for `PDOStatement::fetch()`

phansys:


msamesch:
I was also just about to do that along with some additions. There's <literal>cursor_orientation</literal> two times, which probably should be <parameter>cursorOrientation</parameter> and the constants are tagged as literal instead of constant.

I tried to add a patch, but it didn't work.

phansys:
Thank you for the hints. I'll update the PR based on your findings.

Girgias (approved):

PR #2307, opened 2023-02-18
Rename language/predefined/iteratoraggregate.xml to language/predefin…

jaenDeyPlaidne:
…ed/comparitive/ia’query sample /.xml


TimWolla:
This looks like it was submitted in error.

PR #2311, opened 2023-02-19
Clarify return value when array is inputted

Loyen:
It's not made clear that inputting an array as `value` causes each entry to be filtered rather than the array itself.

### DIFF
--- a/reference/filter/functions/filter-var.xml
+++ b/reference/filter/functions/filter-var.xml
@@ -26,9 +26,12 @@
       <para>
        Value to filter. Note that scalar values are
        <link linkend="language.types.string.casting">converted to string</link>
-       internally before they are filtered. If value is an array, each scalar
-       entry found within it will have the inputted filter applied to them
-       recursively.
+       internally before they are filtered.
+      </para>
+      <para>
+       If value is an array together with <constant>FILTER_REQUIRE_ARRAY</constant>
+       as <parameter>options</parameter>, each scalar entry found within it will
+       have the inputted filter applied to them.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/filter/functions/filter-var.xml
+++ b/reference/filter/functions/filter-var.xml
@@ -24,14 +24,9 @@
      <term><parameter>value</parameter></term>
      <listitem>
       <para>
-       Value to filter. Note that scalar values are
-       <link linkend="language.types.string.casting">converted to string</link>
-       internally before they are filtered.
-      </para>
-      <para>
-       If value is an array together with <constant>FILTER_REQUIRE_ARRAY</constant>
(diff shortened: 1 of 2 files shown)


Girgias:
Can you provide an example, as I don't understand what you're describing here.

Because using an array seems to fail for me: https://3v4l.org/BY1aX

Loyen:
@Girgias, ah I didn't realize that I only tested this using the  `FILTER_REQUIRE_ARRAY` flag.

Here's an example with when using that flag: https://3v4l.org/uOTvE

Girgias:
So basically using the ``FILTER_REQUIRE_ARRAY`` flag is the same as using the ``filter_var_array()`` function?

In that case, I would just redirect users to the corresponding function for examples (also what business has this flag existing in the first place...)

Loyen:
Not exactly. From my understand, `filter_var_array()` lets you define a filter and options per key in an array, while `FILTER_REQUIRE_ARRAY` applies the chosen filter across _all_ scalar values within the value given recursively. 

A better example would be https://3v4l.org/2X9rZ (both `filter_var()` and `filter_var_array()` included) where you can see it apply the filter to the array and sub-arrays when the flag is set.

Maybe a better idea would be to clarify the [FILTER_REQUIRE_FLAG](https://github.com/php/doc-en/blob/master/reference/filter/filters.xml#L800-L807) flag instead of updating `filter_var()`, since you can get the same effect with `filter_var_array()` if you choose to use it there.

reference/filter/functions/filter-var.xml
```diff
@@ -24,14 +24,9 @@
...
-       <link linkend="language.types.string.casting">converted to string</link>
-       internally before they are filtered.
-      </para>
-      <para>
-       If value is an array together with <constant>FILTER_REQUIRE_ARRAY</constant>
-       as <parameter>options</parameter>, each scalar entry found within it will
-       have the inputted filter applied to them.
+       Value to filter. Note that scalar values are <link
+       linkend="language.types.string.casting">converted to
+       string</link> internally before they are filtered.
```
Girgias: Is it just scalars or will it try to convert anything to string? As arrays casted to string "works" and if it attempts blindly object may throw a type error

Loyen: Looking at the example in https://github.com/php/doc-en/pull/2311#issuecomment-1439177521, it seems to not attempt to cast arrays to strings but rather recursively finds any scalar values. What I did in this commit (7630950) was revert my changes to the `filter_var()` description specifically and move it to the flag itself.

Girgias:
Very late on this as I was somewhat on holiday and conferences around that time.

But thank you!

Loyen:
No worries! Just happy to be able to contribute when possible.

PR #2314, opened 2023-02-20
Fix SSLVERSION missing CURL_SSLVERSION_TLSv1_3

Sy-Dante:


devnexen:
Appreciated effort. However, would you mind updating [somewhere here too](https://github.com/php/doc-en/blob/f0261e36dc250410f352fe33ad4c4e699cb18b02/reference/curl/constants.xml#L844) ?

Sy-Dante:
> Appreciated effort. However, would you mind updating [somewhere here too](https://github.com/php/doc-en/blob/f0261e36dc250410f352fe33ad4c4e699cb18b02/reference/curl/constants.xml#L844) ?

My pleasure. Already updated.

devnexen:
Thanks !

PR #2317, opened 2023-02-21
cloning Exception hasn't been fatal since PHP 5.6

divinity76:
https://3v4l.org/blOeC

### DIFF
--- a/language/predefined/exception/clone.xml
+++ b/language/predefined/exception/clone.xml
@@ -13,7 +13,8 @@
    <void/>
   </methodsynopsis>
   <para>
-   Tries to clone the Exception, which results in an Error.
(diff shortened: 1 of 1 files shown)


language/predefined/exception/clone.xml
```diff
@@ -13,7 +13,7 @@
    <void/>
   </methodsynopsis>
   <para>
-   Tries to clone the Exception, which results in Fatal error.
+   Tries to clone the Exception, which results in an Error.
```
Girgias, line 16:
```suggestion
   <classname>Exception</classname>s cannot be cloned,
   and attempting to do so will throw an <classname>Error</classname>.
```

PR #2318, opened 2023-02-21
random: Unify code style in examples

TimWolla:
I apologize to the translators in advance :see_no_evil: The changes should be able to be trivially copied over, though.

-------------------

The spaces within a function call have been introduced in ad458c5f01db4bed979d1623267738ccc7a31962 and I have been trying to emulate them, but failed due to muscle memory, resulting in very inconsistently formatted examples.

Adjust the code examples purely stylistically, without changing their output:

- Remove the extra spaces within the function calls to match the commonly accepted code style (e.g. PSR-12).
- Use `implode()` instead of the `join()` alias.
- Use comma instead of string concatenation when echo-ing.
- Consistently use "\n" to emit newlines, not `PHP_EOL`.
- Unify to double quotes.

### DIFF
--- a/reference/random/functions/mt-rand.xml
+++ b/reference/random/functions/mt-rand.xml
@@ -111,10 +111,10 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-echo mt_rand() . "\n";
-echo mt_rand() . "\n";
+echo mt_rand(), "\n";
+echo mt_rand(), "\n";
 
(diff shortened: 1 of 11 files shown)


Girgias (approved):
Well I'm pretty sure this hasn't been translated into French and only @saundefined and @mumumu might be on top with Russian/Japanese respectively.

PR #2321, opened 2023-02-22
Fix wording of openssl-pkey-get-details to be consistent with other functions

martinssipenko:
Changed wording to use `on` instead of `in` to be consistent with other functions.


devnexen (approved):
LGTM

devnexen:
Thanks !

PR #2324, opened 2023-02-23
Update inheritance.xml

Yigaue:
Nicely formatted text that flows out of the page to make reading friendly.


kamil-tekiela:
Thanks for the PR. Sorry, but I am afraid we do this on purpose.

PR #2330, opened 2023-02-28
Explicitly mention that spl_autoload_register will drop any leading backslash

TimWolla:
Due to this guarantee there is no need to perform this unnecessary check within the autoloader.

### DIFF
--- a/reference/spl/functions/spl-autoload-register.xml
+++ b/reference/spl/functions/spl-autoload-register.xml
@@ -51,7 +51,7 @@
       </methodsynopsis>
       <para>
        The <parameter>class</parameter> will not contain the leading
-       backslash of a fully-qualified classname.
+       backslash of a fully-qualified identifier.
       </para>
      </listitem>
     </varlistentry>
@@ -168,7 +168,7 @@ Fatal error: Class 'Foobar\InexistentClass' not found in ...
     </screen>
    </example>
    <example>
-    <title>The classname will be provided without the leading backslash</title>
+    <title>The identifier will be provided without the leading backslash</title>
     <programlisting role="php">
 <![CDATA[
 <?php
--- a/reference/spl/functions/spl-autoload-register.xml
+++ b/reference/spl/functions/spl-autoload-register.xml
@@ -47,10 +47,10 @@
       </para>
       <methodsynopsis>
        <type>void</type><methodname><replaceable>callback</replaceable></methodname>
-       <methodparam><type>string</type><parameter>class_name</parameter></methodparam>
+       <methodparam><type>string</type><parameter>class</parameter></methodparam>
       </methodsynopsis>
       <para>
-       The <parameter>class_name</parameter> will not contain the leading
+       The <parameter>class</parameter> will not contain the leading
        backslash of a fully-qualified classname.
       </para>
      </listitem>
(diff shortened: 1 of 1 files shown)


reference/spl/functions/spl-autoload-register.xml
```diff
@@ -160,6 +164,34 @@ new InexistentClass;
...
+<?php
+
+spl_autoload_register(static function ($class_name) {
+    var_dump($class_name);
+});
+
+class_exists(RelativeName::class);
+class_exists(RelativeName\WithNamespace::class);
+class_exists(\AbsoluteName::class);
+class_exists(\AbsoluteName\WithNamespace::class);
```
kamil-tekiela, line 183: While your PR makes senses to me, your example is misleading. Please see `var_dump(\AbsoluteName\WithNamespace::class);`. https://3v4l.org/mBbet

TimWolla: Good catch, thank you. Fixed.

reference/spl/functions/spl-autoload-register.xml
```diff
@@ -49,6 +49,10 @@
        <type>void</type><methodname><replaceable>callback</replaceable></methodname>
        <methodparam><type>string</type><parameter>class_name</parameter></methodparam>
       </methodsynopsis>
+      <para>
+       The <parameter>class_name</parameter> will not contain the leading
+       backslash of a fully-qualified classname.
```
kamil-tekiela, line 54:
```suggestion
       The <parameter>class_name</parameter> will not contain the leading
       backslash of a fully-qualified identifier.
```

`classname` is not actually a word. Besides, it's not just classes that get loaded here. Replace this word with either `identifier` or `name`. See https://www.php.net/manual/en/language.namespaces.rules.php

You also touch on another point here, why is the param called `$class_name` at all? [`spl_autoload()`](https://www.php.net/manual/en/function.spl-autoload.php) lists it as `$class`. Which name is correct?

reference/spl/functions/spl-autoload-register.xml
```diff
@@ -160,6 +164,34 @@ new InexistentClass;
 <![CDATA[
 [[Foobar\InexistentClass]]
 Fatal error: Class 'Foobar\InexistentClass' not found in ...
+]]>
+    </screen>
+   </example>
+   <example>
+    <title>The classname will be provided without the leading backslash</title>
```
kamil-tekiela, line 171:
```suggestion
    <title>The identifier will be provided without the leading backslash</title>
```

kamil-tekiela (changes requested):

reference/spl/functions/spl-autoload-register.xml
Girgias: ``$class`` is the correct name if you look at the stubs. So that needs to be fixed.

Girgias (changes requested):
I agree with @kamil-tekiela's comments, so addressing those would be great!

reference/spl/functions/spl-autoload-register.xml
TimWolla: All done, I believe. PTAL (and feel free to merge directly if it is good).

kamil-tekiela (approved):
LGTM

PR #2331, opened 2023-02-28
Fix function signature

m157y:
Missed false on return in case of failure

https://github.com/php/php-src/blob/4177257178d6a1a44f0aa6d6b23d02b91e0a58d3/ext/ftp/php_ftp.c#L797


phansys:
IMO, this should be updated in the stub:
https://github.com/php/php-src/blob/4177257178d6a1a44f0aa6d6b23d02b91e0a58d3/ext/ftp/ftp.stub.php#L113

Girgias:
Yeah that's a php-src bug that needs to be fixed.

kocsismate (approved):

PR #2348, opened 2023-03-08
update length verbiage to differentiate behavior between BLOB bytes and CLOB characters

ashnazg:
This updates the doc verbiage based on @cjbj 's comments on https://bugs.php.net/bug.php?id=70700.

### DIFF
--- a/reference/oci8/ocilob/read.xml
+++ b/reference/oci8/ocilob/read.xml
@@ -13,13 +13,14 @@
    <methodparam><type>int</type><parameter>length</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Reads <parameter>length</parameter> bytes from the current position of
-   LOB's internal pointer.
+   Reads <parameter>length</parameter> of bytes (BLOB) or characters (CLOB)
+   from the current position of LOB's internal pointer.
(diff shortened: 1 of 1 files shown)


reference/oci8/ocilob/read.xml
```diff
@@ -31,7 +32,8 @@
      <term><parameter>length</parameter></term>
      <listitem>
       <para>
-       The length of data to read, in bytes.  Large values will be rounded down to 1 MB.
+       The length of data to read, in bytes (BLOB) or characters (CLOB).
+       Large values will be rounded down to 1 MB.
```
ashnazg, line 36: @cjbj , does this line about `rounded down to 1 MB` need to be adjusted based on the byte/character difference?

cjbj: I expect so but would need to test.

ashnazg: So then, is it reasonable to merge this PR, and I'll _separately_ keep that `1 MB` conversation going?

Girgias: Merged for the time being.

PR #2350, opened 2023-03-10
Delete DirectoryIterator methods that are inherited

othercorey:
https://github.com/php/php-src/blob/master/ext/spl/spl_directory.stub.php#L107

These methods are inherited from SplFileInfo and have not been maintained. Many have incorrect return types.

I believe the current practice is to show the inherited methods in the class listing and link where they're defined not maintain copies of the method.


othercorey:
@Girgias Is this the correct way to fix inherited methods?

Girgias:
> @Girgias Is this the correct way to fix inherited methods?

It would be yes, but there seems to be some examples compared to SPLFileInfo, maybe trying to pull the examples up is a good idea?

othercorey:
I'm not sure I understand. You want to maintain the copies of the inherited methods so the example say DirectoryIterator instead of SPLFileInfo?

Girgias:
> I'm not sure I understand. You want to maintain the copies of the inherited methods so the example say DirectoryIterator instead of SPLFileInfo?

No, I would like for you to take the DirectoryIterator examples, and adapt them so they can be put into the SPLFileInfo docs and continue to remove the existing docs for the inherited methods on DirectoryIterator

othercorey:
Harvested useful parts of examples and updated SplFileInfo.

reference/spl/splfileinfo/getctime.xml
```diff
@@ -32,19 +32,19 @@
...
+   Throws <classname>RuntimeException</classname> on error.
   </para>
  </refsect1>
 
  <refsect1 role="examples">
   &reftitle.examples;
   <para>
    <example>
-    <title><function>SplFileInfo::getCTime</function> example</title>
+    <title><methodname>SplFileInfo::getCTime</methodname> example</title>
```
othercorey, line 43: DirectoryIterator was using `methodname`. I'm not sure how useful that is since it justs to the same file we're already in. However, `function` doesn't seem correct either as it does nothing.

Girgias: ``<function>`` is not correct, but PhD should handle it the same way as methodname, but it's a mess. And better to be explicit with the method name for the moment can always decide to drop it later.

reference/spl/splfileinfo/getowner.xml
```diff
@@ -28,27 +28,44 @@
...
+
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
    Throws <classname>RuntimeException</classname> on error.
   </para>
  </refsect1>
 
- <refsect1 role="examples">
+  <refsect1 role="examples">
```
Girgias, line 39: Nit: unintentionnal indent change?

reference/spl/splfileinfo/isreadable.xml
```diff
@@ -33,23 +33,21 @@
...
 <![CDATA[
 <?php
-$info = new SplFileInfo(__FILE__);
-var_dump($info->isReadable());
-
-$info = new SplFileInfo('foo');
-var_dump($info->isReadable());
+$info = new SplFileInfo('readable.jpg');
+if ($info->isReadable()) {
+ echo $info->getFilename() . ' is readable';
```
Girgias, line 42: Nit: 4 spaces indent in code examples
```suggestion
    echo $info->getFilename() . ' is readable';
```

reference/spl/splfileinfo/iswritable.xml
```diff
@@ -29,6 +29,31 @@
...
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><methodname>SplFileInfo::isWriteable</methodname> example</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$info = new SplFileInfo('locked.jpg');
+if (!$info->isWriteable()) {
+ echo $info->getFilename() . ' is not writeable';
```
Girgias, line 42: Nit: Ditto
```suggestion
    echo $info->getFilename() . ' is not writeable';
```

Girgias (approved):
Small nits but LGTM! Thank you!

reference/spl/splfileinfo/getowner.xml
othercorey: DirectoryIterator had some odd indents

Girgias:
Thank you!

PR #2353, opened 2023-03-12
`str-split`: precise behavior change for PHP 8.2.0

homersimpsons:
It looks like `str_split` behaviour's did change in PHP8.2.0: https://3v4l.org/mHMvb

This document the change. I am unsure about the wording, also an example may be clearer.

### DIFF
--- a/reference/strings/functions/str-split.xml
+++ b/reference/strings/functions/str-split.xml
@@ -73,6 +73,13 @@
     </thead>
     <tbody>
      <row>
+      <entry>8.2.0</entry>
+      <entry>
+       If <parameter>string</parameter> is empty an empty &array; is now returned.
+       Previously an &array; containing a single empty string was returned.
+      </entry>
+     </row>
+     <row>
       <entry>8.0.0</entry>
       <entry>
        If <parameter>length</parameter> is less than <literal>1</literal>,


homersimpsons:
I do not know who should be assigned, maybe @Girgias you could look into this ?

reference/strings/functions/str-split.xml
```diff
@@ -81,6 +81,13 @@
        has been raised instead, and the function returned &false;.
       </entry>
      </row>
+     <row>
+      <entry>8.2.0</entry>
+      <entry>
+       If <parameter>string</parameter> is empty an empty array will be returned.
+       Previously an array containing a single empty string was returned.
+      </entry>
+     </row>
```
Girgias, line 90: Nits:
Should be above the 8.0 entry.
```suggestion
     <row>
      <entry>8.2.0</entry>
      <entry>
       If <parameter>string</parameter> is empty an empty &array; is now returned.
       Previously an &array; containing a single empty string was returned.
      </entry>
     </row>
```

Girgias (approved):
Thank you! Minor nits and it's good to go.

reference/strings/functions/str-split.xml
homersimpsons: (done)

Girgias:
Thank you!

PR #2355, opened 2023-03-16
Add execute_query() to the See also section

colshrapnel:
To give it more traction.


kamil-tekiela (approved):
Seems reasonable to me.

PR #2360, opened 2023-03-16
Re-phrase sentence to improve its meaning

odsantos:


language/types/string.xml
```diff
@@ -395,7 +395,7 @@ PHP Parse error:  syntax error, unexpected identifier "ING", expecting "]" in ex
 
     <simpara>
      To avoid this problem, it is safe for you to follow the simple rule:
-     <emphasis>do not choose the closing identifier that appears in the body
+     <emphasis>do not choose as closing identifier if it appears in the body
```
Girgias, line 398:
```suggestion
     <emphasis>do not choose as a closing identifier if it appears in the body
```
More? because it really doesn't read well otherwise.

Ideally the personalization (usage of "you") in the above line would also be removed.

odsantos: I agree with your suggestion.

language/types/string.xml
```diff
@@ -394,8 +394,8 @@ PHP Parse error:  syntax error, unexpected identifier "ING", expecting "]" in ex
     </example>
 
     <simpara>
-     To avoid this problem, it is safe for you to follow the simple rule:
-     <emphasis>do not choose the closing identifier that appears in the body
+     To avoid this problem, it is safe to follow the simple rule:
+     <emphasis>do not choose a closing identifier if it appears in the body
```
Girgias, line 398:
```suggestion
     <emphasis>do not choose as a closing identifier if it appears in the body
```

PR #2361, opened 2023-03-16
Fix: Remove class declaration

localheinz:
This pull request

- [x] removes a class declaration for a `SpecificException` that is not used in the example

See https://www.php.net/manual/en/language.exceptions.php#example-336:

CleanShot 2023-03-16 at 23 55 29@2x


TimWolla (approved):
Thank you!

Issue #2362, opened 2023-03-17
Undocumented $mode


jdarwood007:
From manual page: https://php.net/phar.count

---

This page indicates count() has a parameter called $mode.  But no documentation exists explaining its usage.

PR #2363, opened 2023-03-17
added: Phar::count parameter documentation

aarkaynegi:
I have updated the documentation for the Phar::count method to include information on its $mode parameter. This should fix issue #2362.

### DIFF
--- a/reference/phar/Phar/count.xml
+++ b/reference/phar/Phar/count.xml
@@ -18,20 +18,20 @@
  </refsect1>
  <refsect1 role="parameters">
   &reftitle.parameters;
-  <para>
-    <variablelist>
-      <varlistentry>
-        <term>
-          <parameter>mode</parameter>
-        </term>
-        <listitem>
-          <para>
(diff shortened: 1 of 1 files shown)


reference/phar/Phar/count.xml
```diff
@@ -19,6 +19,18 @@
...
+          <parameter>mode</parameter>
+        </term>
+        <listitem>
+          <para>
+            <parameter>mode</parameter> is an integer value specifying the counting mode to be used. By default, it is set to <constant>COUNT_NORMAL</constant>, which counts only the number of items in the archive that have not been deleted or hidden. When set to <constant>COUNT_RECURSIVE</constant>, it counts all items in the archive, including those that have been deleted or hidden.
+          </para>
+        </listitem>
+      </varlistentry>
+    </variablelist>
   </para>
```
Girgias, line 34: Those external para tags look wrong.
```suggestion
  <variablelist>
   <varlistentry>
    <term><parameter>mode</parameter></term>
    <listitem>
     <para>
      <parameter>mode</parameter> is an integer value specifying the counting mode to be used.
      By default, it is set to <constant>COUNT_NORMAL</constant>,
      which counts only the number of items in the archive that have not been deleted or hidden.
      When set to <constant>COUNT_RECURSIVE</constant>, it counts all items in the archive,
      including those that have been deleted or hidden.
     </para>
    </listitem>
   </varlistentry>
  </variablelist>
```

Also added some line breaks which is helpful for translations.

PR #2364, opened 2023-03-17
Add error_log() to See Also for log()

larsnystrom:
A little bit of help to those who might be looking for a way to log error messages.


Girgias:
Thank you!

PR #2385, opened 2023-03-26
call_user_func_array $args need not be indexed

mjec:
Since PHP 8.0, the keys can be used as argument names.

### DIFF
--- a/reference/funchand/functions/call-user-func-array.xml
+++ b/reference/funchand/functions/call-user-func-array.xml
@@ -35,7 +35,23 @@
      <term><parameter>args</parameter></term>
      <listitem>
       <para>
-       The parameters to be passed to the callback, as an indexed array.
+       The parameters to be passed to the callback, as an array.
+      </para>
+      <para>
+        If the keys of <parameter>args</parameter> are all numeric,
+        the keys are ignored and each element will be passed to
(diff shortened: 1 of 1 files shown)


Girgias (changes requested):
This needs a changelog entry at minimum. The PHP docs must document PHP 7 + 8 behaviour.

mjec:
> This needs a changelog entry at minimum. The PHP docs must document PHP 7 + 8 behaviour.

The changelog already has an entry for this:

> 8.0.0 	args keys will now be interpreted as parameter names, instead of being silently ignored.

Is that sufficient or should the description of args be expanded to explicitly specify 7 + 8 behaviors?

Girgias:
> > This needs a changelog entry at minimum. The PHP docs must document PHP 7 + 8 behaviour.
> 
> The changelog already has an entry for this:
> 
> > 8.0.0 	args keys will now be interpreted as parameter names, instead of being silently ignored.
> 
> Is that sufficient or should the description of args be expanded to explicitly specify 7 + 8 behaviors?

I think the PHP 8.0 behaviour should be described in the description of the args param, and the current changelog entry should be sufficient to say well the PHP 7 behaviour is different because the keys are ignored and only positional args where accepted

Girgias (approved):
Brilliant, thank you!

PR #2387, opened 2023-03-26
Make CTRL+C clearer in documentation

adoy:


features/dtrace.xml
```diff
@@ -369,7 +369,7 @@ php*:::request-startup
 
    <para>
     When monitoring is complete, the D script can be terminated with a
-    <literal>^C</literal>.
+    <literal>CTRL+C</literal>.
```
Girgias, line 372: So, looking at [DocBooks](https://tdg.docbook.org/tdg/5.1/keycombo.html) I would expect that we should mark it up as:
```suggestion
     <keycombo action='simul'><keycap>CTRL</keycap> <keycap>C</keycap></keycombo>.
```
But I don't think PhD has support for rendering this :-/

adoy: Should I change it then ? I don't know DocBooks/PhD that much so I live it in your hands.

Girgias: PhD doesn't support it, let me see if I can add support quickly or not...

Girgias: Okay let's wait for a bit on this front and see if my PhD changes land.

salathe: If your PhD change lands, suggest not having any whitespace between sibling `keycap` elements (so that that whitespace doesn't end up in the resulting HTML).

Girgias (approved):

PR #2398, opened 2023-03-31
Issue #2397: Bring reserved.xml in sync with RFC 

### LINKED ISSUE #2397: Documentation about other reserved words in namespaces is out of date
The [list of other reserved words](https://www.php.net/manual/en/reserved.other-reserved-words.php) states that they cannot be used in namespaces. However [_PHP RFC: Treat namespaced names as single token_](PHP RFC: Treat namespaced names as single token) changed that behavior and that should be removed here.

FatherShawn:
Fixes #2397 and brings documentation in sync with the RFC on [namespaces as a single token](https://wiki.php.net/rfc/namespaced_names_as_token) which was [implemented in 8.0](https://github.com/php/php-src/milestone/1?closed=1).

### DIFF
--- a/appendices/reserved.xml
+++ b/appendices/reserved.xml
@@ -508,6 +508,7 @@
    <title>List of other reserved words</title>
    <simpara>
     The following words cannot be used to name a class, interface or trait.
+    Prior to PHP 8.0, they are also prohibited from being used in namespaces.
    </simpara>
    <para>
     <table>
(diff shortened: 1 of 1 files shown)


Girgias (changes requested):
The documentation describes PHP 7 and PHP 8 behaviour so there needs to be a changelog entry and/or paragraphs duplications.

For example, start the first paragraph with "As of PHP 8.X.0," and the second one "Prior to PHP 8.X.0," describing the behaviour.

Girgias (approved):

PR #2409, opened 2023-04-08
Remove entities.functions.xml

mowangjuanzi:


Girgias:
That's a generated file right?

mowangjuanzi:
It looks like it is. It may have been submitted accidentally.

Issue #696, opened 2021-06-16
Meaningless caution on generator syntax page since the related example was removed


a-menshchikov:
From manual page: https://php.net/language.generators.syntax

---

The following block seems like an artifact:

> Caution
> The value that will be assigned to $data is the value passed to Generator::send(), or null if Generator::next() is called instead.

cmb69:
That has been added as part of fixing [bug 76104](https://bugs.php.net/bug.php?id=76104). Do you have a suggestion on how to improve this?

nikic:
At least the caution doesn't make sense in context, because it refers to a `$data` variable that doesn't exist. This was probably moved around to a place where it no longer makes sense.

nikic:
Same for the second one:

> Caution
> As with the simple value yields shown earlier, yielding a key/value pair in an expression context requires the yield statement to be parenthesised: 

This is not "shown earlier" and the parentheses requirement is gone in PHP 7, so this one can just be dropped entirely.

a-menshchikov:
I think that usage of `yield` keyword as expression that returns a value should be described in a separate subsection.

cmb69:
Oh, https://github.com/php/doc-en/commit/cdaea0421544885f02ff3d36bd203dc01b78299e#diff-9e27d08e700634a0fe443f7249fd2bcbcfc7f5c788e96dcd298843b78010ddbc removed the example to which the note was referring to.

PR #2416, opened 2023-04-10
Drop outdated cautions in generator syntax examples

othercorey:
fixes https://github.com/php/doc-en/issues/696


kamil-tekiela (approved):
Makes sense. Thanks!

Issue #1351, opened 2022-01-19
DateTime::getLastErrors only returns false when no date functions have been called


marcinmdev:
From manual page: https://php.net/datetime.getlasterrors

---
```
PHP 8.0.2 (cli) (built: Feb  9 2021 19:20:59) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.2, Copyright (c) Zend Technologies
```

```
php > var_dump(DateTime::createFromFormat('d/m/Y','01/01/2022')::getLastErrors());
array(4) {
  ["warning_count"]=>
  int(0)
  ["warnings"]=>
  array(0) {
  }
  ["error_count"]=>
  int(0)
  ["errors"]=>
  array(0) {
  }
}

```

cmb69:
If no date function had been called prior to calling `date_get_last_errors()`, [`false` is returned](https://3v4l.org/rvMCX). It's probably a good idea to document that (and maybe change the behavior sometime in the future).

othercorey:
@kamil-tekiela This should be fixed with the last patch

derickr:
We fixed this behaviour in PHP 8.2, at which time we also documented this in the changelog: https://www.php.net/manual/en/datetimeimmutable.getlasterrors.php#refsect1-datetimeimmutable.getlasterrors-changelog

PR #2422, opened 2023-04-13
Update DateTime::getLastErrors() return false description

othercorey:
fixes https://github.com/php/doc-en/issues/1351


kamil-tekiela:
Are you sure this is correct? It seems to return false even when I create DateTime object before.

othercorey:
I understood cmb's explanation to mean it returns false until a function is called.

PR #2429, opened 2023-04-16
spl_autoload_register ignores throw parameter

KapitanOczywisty:
I've added note about ignored parameter

https://github.com/php/php-src/commit/2302b14aab3bf01a9a87d2ee87951a13bc7752a0


kamil-tekiela (approved):
LGTM

PR #2430, opened 2023-04-16
Replace links to __autoload

### LINKED ISSUE #1456: __autoload deprecated
From manual page: https://php.net/function.class-exists

---

Whether to call [[spl_autoload_register](https://www.php.net/manual/en/function.spl-autoload-register.php)](https://www.php.net/manual/en/language.oop5.autoload.php) by default.

KapitanOczywisty:
I've replaced some links to removed [__autoload](https://www.php.net/manual/en/function.autoload.php) with [autoloading classes](https://www.php.net/manual/en/language.oop5.autoload.php). There are also a few in SPL, but I've not updated these.

Fixes #1456

### DIFF
--- a/reference/classobj/functions/class-alias.xml
+++ b/reference/classobj/functions/class-alias.xml
@@ -43,7 +43,8 @@
      <term><parameter>autoload</parameter></term>
      <listitem>
       <para>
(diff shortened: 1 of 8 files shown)


kamil-tekiela (approved):
LGTM

PR #2432, opened 2023-04-18
Update generators.xml

MaxiCom:


derickr:
Why are you changing this `+` to `-`? Please add a description to your PR.

MaxiCom:
My bad, i just realized that if the step is negative (mandatory to reach the for statement), the `$i += $step`, will indeed decrease the index!

Sorry for your inconvenience 😞

PR #2442, opened 2023-04-24
chore: updated code example to use PHP 8 feature

thorsten:


Girgias:
The docs support PHP 7 and 8.

PR #2448, opened 2023-04-26
Optimize reference code

LRainner:
PHP 7.4 custom object serialization reference code optimization


kamil-tekiela:
I'm sorry, but what's the improvement? I can't see anything other then you wrapping this in `A` class.

LRainner:
I have read all the reference codes of PHP new features, only this one cannot be executed directly, I don’t think the reference code should be like this. Besides, I see https://github.com/php/php-src/blob/master/Zend/zend_language_parser.y#L938, it shows `public` reserved word must be in a class. So, I think it is reasonable.

PR #2451, opened 2023-04-27
Update fetch-fields.xml

mattkolb:
Docs should be updated according to this:


kamil-tekiela:
Can you merge this PR with the other one?

mattkolb:
Was that a request for me? If so, I can, but I'd need some instruction.

kamil-tekiela:
Yes. Please just make the change in the other PR as requested in the comment. https://github.com/php/doc-en/pull/2452 then you can close this one.

PR #2452, opened 2023-04-27
Update fetch-field-direct.xml

mattkolb:

### DIFF
--- a/reference/mysqli/mysqli_result/fetch-field-direct.xml
+++ b/reference/mysqli/mysqli_result/fetch-field-direct.xml
@@ -84,7 +84,7 @@
      </row>
      <row>
       <entry>max_length</entry>
-      <entry>The maximum width of the field for the result set.</entry>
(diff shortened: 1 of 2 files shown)


reference/mysqli/mysqli_result/fetch-field-direct.xml
```diff
@@ -84,7 +84,7 @@
      </row>
      <row>
       <entry>max_length</entry>
-      <entry>The maximum width of the field for the result set.</entry>
+      <entry>The maximum width of the field for the result set. As of PHP 8.1, this will always return 0.</entry>
```
kamil-tekiela, line 87:
```suggestion
      <entry>The maximum width of the field for the result set. As of PHP 8.1, this value is always <literal>0</literal>.</entry>
```

kamil-tekiela (changes requested):

kamil-tekiela (approved):

PR #2453, opened 2023-04-27
Update store-result.xml

mattkolb:


kamil-tekiela (commented):
This is good, but a changelog entry should also be added.

PR #2454, opened 2023-04-27
Update Mysqli::construct() docs for return value as of PHP 8.1

mattkolb:

### DIFF
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -128,8 +128,8 @@
    &return.falseforfailure;.
   </para>
   <para>
-   <methodname>mysqli::connect</methodname> returns &null; on success&return.falseforfailure;.
-   As of PHP 8.1.0, returns &true; on success.
+   <methodname>mysqli::connect</methodname> returns &true; on success&return.falseforfailure;.
+   Prior to PHP 8.1.0, returns &null; on success.
   </para>
  </refsect1>
 
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -124,10 +124,6 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   <methodname>mysqli::__construct</methodname> always returns an object which represents the connection to a MySQL Server, 
-   regardless of it being successful or not.
-  </para>
-  <para>
    <function>mysqli_connect</function> returns an object which represents the connection to a MySQL Server,
    &return.falseforfailure;.
   </para>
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli/construct.xml
```diff
@@ -132,7 +132,7 @@
    &return.falseforfailure;.
   </para>
   <para>
-   <methodname>mysqli::connect</methodname> returns &null; on success&return.falseforfailure;.
+   <methodname>mysqli::connect</methodname> returns &null; on success&return.falseforfailure;. As of PHP 8.1, returns true on success.
```
kamil-tekiela, line 135:
```suggestion
   <methodname>mysqli::connect</methodname> returns &null; on success&return.falseforfailure;. As of PHP 8.1, returns &true; on success.
```

Girgias (changes requested):
This is missing a Changelog section.

reference/mysqli/mysqli/construct.xml
```diff
@@ -159,6 +160,12 @@
         All parameters are now nullable.
        </entry>
       </row>
+      <row>
+       <entry>8.1.0</entry>
+       <entry>
+        <methodname>mysqli::connect</methodname> now returns &true; instead of &null; on success.
+       </entry>
+      </row>
```
Girgias, line 168: Must be above the 7.4.0 change log entry

reference/mysqli/mysqli/construct.xml
```diff
@@ -133,6 +133,7 @@
   </para>
   <para>
    <methodname>mysqli::connect</methodname> returns &null; on success&return.falseforfailure;.
+   As of PHP 8.1.0, returns &true; on success.
```
Girgias, line 136:
```suggestion
   <methodname>mysqli::connect</methodname> &return.success;.
```
Is what I want to use, but the entity starts with an uppercase letter. :-/

Also, the "return value" for ``__construct()`` doesn't really make sense looking at it

Girgias (changes requested):

reference/mysqli/mysqli/construct.xml
kamil-tekiela: I don't think I can use this entity here. This page should probably be split into 3 separate pages. For now, I would suggest to keep it as it is in the PR.

Girgias: Possibly, or maybe just into 2 where the constructor is seperate.

But I would use more the wording:

```xml
   <methodname>mysqli::connect</methodname> returns &true; on success&return.falseforfailure;.
   Prior to PHP 8.1.0, returns &null; on success.
```

Girgias (approved):

PR #2458, opened 2023-05-02
Narrow bool return types to true when possible

kocsismate:


Girgias (commented):
The text in the return value must be amended for XML at least.

Also the XML functions did return false on an invalid resource type in PHP 7, but I think we consider this to be UB so no need to document.

@kamil-tekiela do you know since when the Mysqli functions only return true?

kamil-tekiela:
It was changed in https://github.com/php/php-src/pull/5058, but as I understand standalone true is only possible since 8.2. 

`mysqli_debug` always returned `true` but it could only be typed as bool due to lack of standalone true type.

Girgias:
> It was changed in [php/php-src#5058](https://github.com/php/php-src/pull/5058), but as I understand standalone true is only possible since 8.2.
> 
> `mysqli_debug` always returned `true` but it could only be typed as bool due to lack of standalone true type.

Okay that's just what I wanted to know if we need a changelog, which seems like we do for PHP 8.0 if it isn't already there. :)

reference/network/functions/openlog.xml
```diff
@@ -181,7 +181,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &return.success;
+   &return.true.always;
```
Girgias, line 184: So this used to return false in 8.0, so this needs some more investigation of what version this changed and why.

Girgias (approved):
Other than ``openlog()`` everything LGTM

reference/network/functions/openlog.xml
kocsismate: Ah it was Ilija: https://github.com/php/php-src/pull/9295/files#diff-96c05a0551ccbeca3ff610c111f88b54a0d727888cf353f477ea8dc951a18637R685

kocsismate: But since it's a PHP 8.3 change (he only committed it to master in last August), it should revert this change :S

PR #2461, opened 2023-05-05
stream_get_contents: Note buffer of max length size will always be allocated

naderman:
Clarify in a note that even if a stream's contents are much smaller a buffer of the maximum length will be allocated if a maximum length is provided. This may be surprising if you otherwise expect the function to dynamically grow a buffer depending on the actual stream contents, as it does with the default null value.

### DIFF
--- a/reference/stream/functions/stream-get-contents.xml
+++ b/reference/stream/functions/stream-get-contents.xml
@@ -119,6 +119,13 @@ if ($stream = fopen('http://www.example.net', 'r')) {
  <refsect1 role="notes">
   &reftitle.notes;
   &note.bin-safe;
+  <note>
+   <para>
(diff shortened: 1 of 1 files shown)


reference/stream/functions/stream-get-contents.xml
```diff
@@ -119,6 +119,13 @@ if ($stream = fopen('http://www.example.net', 'r')) {
  <refsect1 role="notes">
   &reftitle.notes;
   &note.bin-safe;
+  <note>
+   <para>
+    When specifying a length value other than null, this function will
```
Girgias, line 124:
```suggestion
    When specifying a length value other than &null;, this function will
```

Girgias:
Thank you!

PR #2462, opened 2023-05-07
Fix #2242 : Wrong parameters names for mysqli::real_connect

### LINKED ISSUE #2242: Wrong parameters names for mysqli::real_connect
Like [bug 81232](https://bugs.php.net/bug.php?id=81232), parameters names for `mysqli::real_connect()` are wrong, and leads to confusion when using named parameters.

From file `ext/mysqli/mysqli.stub.php` in the source:

```
    public function real_connect(
        ?string $hostname = null,
        ?string $username = null,
        #[\SensitiveParameter] ?string $password = null,
        ?string $database = null,
        ?int $port = null,
        ?string $socket = null,
        int $flags = 0
    ): bool {}
```

From [https://www.php.net/manual/en/mysqli.real-connect.php](https://www.php.net/manual/en/mysqli.real-connect.php):

```
 public mysqli::real_connect(
    string $host = ?,
    string $username = ?,
    string $passwd = ?,
    string $dbname = ?,
    int $port = ?,
    string $socket = ?,
    int $flags = ?
): bool
```

`$hostname`, `$password`, and `$database` are wrong in the documentation.

kocsismate:


Girgias (approved):
This looks good to me, would just to like @kamil-tekiela to have a quick look. :)

kocsismate:
> This looks good to me, would just to like @kamil-tekiela to have a quick look. :)

Yeah, I appreciate it as some of the parameter descriptions didn't make much sense to me

reference/mysqli/mysqli/construct.xml
```diff
@@ -68,15 +68,15 @@
      <term><parameter>username</parameter></term>
      <listitem>
       <para>
-       The MySQL user name.
+       The MySQL username or &null; to assume the username based on the <literal>mysqli.default_user</literal> ini option.
```
kamil-tekiela, line 71: Shouldn't this link to https://www.php.net/manual/en/mysqli.configuration.php?

reference/mysqli/mysqli/construct.xml
```diff
@@ -68,15 +68,15 @@
...
       </para>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><parameter>password</parameter></term>
      <listitem>
       <para>
-       If not provided or &null;, the MySQL server will attempt to authenticate
+       When &null; is provided, the MySQL server will attempt to authenticate
```
kamil-tekiela, line 79: That's not correct. When `null` is provided it takes the value of `mysqli.default_pw` ini setting. To authenticate with no password use `''`. The ini setting by default is an empty string.

reference/mysqli/mysqli/construct.xml
```diff
@@ -96,15 +96,17 @@
      <term><parameter>port</parameter></term>
      <listitem>
       <para>
-       Specifies the port number to attempt to connect to the MySQL server.
+       When a &string; is provided, it specifies the port number to attempt to connect to the MySQL server.
```
kamil-tekiela, line 99: It's an `int` not a `string`

reference/mysqli/mysqli/construct.xml
```diff
@@ -96,15 +96,17 @@
...
+       Otherwise the default ini option is used (<literal>mysqli.default_port</literal>).
```
kamil-tekiela, line 100: Either put otherwise at the end, or add a comma after otherwise.

reference/mysqli/mysqli/real-connect.xml
```diff
@@ -65,7 +65,7 @@
    <variablelist>
     &mysqli.link.description;
     <varlistentry>
-     <term><parameter>host</parameter></term>
+     <term><parameter>hostname</parameter></term>
      <listitem>
       <para>
        Can be either a host name or an IP address. Passing the &null; value
```
kamil-tekiela, line 71: This description is incorrect and should be identical to the one from `mysqli_connect`. Same goes for other params.

kamil-tekiela (changes requested):

reference/mysqli/mysqli/construct.xml
kocsismate: Yeah this description didn't make sense to me

PR #2467, opened 2023-05-10
Update assert() docs for PHP 8

Girgias:
``assert()`` is pretty wild due to all those options and the behaviour was changed in-between 7 and 8.

### DIFF
--- a/reference/info/functions/assert-options.xml
+++ b/reference/info/functions/assert-options.xml
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis>
    <type>mixed</type><methodname>assert_options</methodname>
-   <methodparam><type>int</type><parameter>what</parameter></methodparam>
+   <methodparam><type>int</type><parameter>option</parameter></methodparam>
    <methodparam choice="opt"><type>mixed</type><parameter>value</parameter></methodparam>
   </methodsynopsis>
   <para>
@@ -33,7 +33,7 @@
   <para>
    <variablelist>
     <varlistentry>
-     <term><parameter>what</parameter></term>
+     <term><parameter>option</parameter></term>
      <listitem>
       <para>
        <table>
@@ -53,6 +53,12 @@
            <entry>assert.active</entry>
            <entry>1</entry>
            <entry>enable <function>assert</function> evaluation</entry>
+          </row>
+          <row>
+           <entry>ASSERT_EXCEPTION</entry>
+           <entry>assert.exception</entry>
+           <entry>1</entry>
+           <entry>throws a <classname>AssertionError</classname> for each failed assertions</entry>
           </row>
           <row>
            <entry>ASSERT_WARNING</entry>
@@ -151,7 +157,41 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns the original setting of any option or &false; on errors.
+   Returns the original setting of any option.
+  </para>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   If <parameter>option</parameter> is not a valid option a
+   <classname>ValueError</classname> is thrown.
+  </para>
+ </refsect1>
+
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <para>
+   <informaltable>
+    <tgroup cols="2">
+     <thead>
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        If <parameter>option</parameter> is not a valid option,
+        a <classname>ValueError</classname> is now thrown.
+        Previously &false; was returned.
+       </entry>
+      </row>
+     </tbody>
+    </tgroup>
+   </informaltable>
   </para>
  </refsect1>
 
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -3,113 +3,47 @@
 <refentry xml:id="function.assert" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>assert</refname>
-  <refpurpose>Checks if assertion is &false;</refpurpose>
+  <refpurpose>Checks an assertion</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
   &reftitle.description;
-  <para>PHP 5 and 7</para>
   <methodsynopsis>
    <type>bool</type><methodname>assert</methodname>
    <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
-   <methodparam choice="opt"><type>string</type><parameter>description</parameter></methodparam>
-  </methodsynopsis>
-  <para>PHP 7</para>
-  <methodsynopsis>
-   <type>bool</type><methodname>assert</methodname>
-   <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
-   <methodparam choice="opt"><type>Throwable</type><parameter>exception</parameter></methodparam>
+   <methodparam choice="opt"><type class="union"><type>Throwable</type><type>string</type><type>null</type></type><parameter>description</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   <function>assert</function> will check the given
-   <parameter>assertion</parameter> and take appropriate action if
-   its result is &false;.
+   <function>assert</function> is not a function but a language construct.
+   allowing for
+   the definition of expectations: assertions that take effect in development
+   and testing environments, but are optimised away to have zero cost in
+   production.
   </para>
-  <refsect2>
-   <title>Traditional assertions (PHP 5 and 7)</title>
-   <para>
-    If the <parameter>assertion</parameter> is given as a string it
-    will be evaluated as PHP code by <function>assert</function>.
-    If you pass a boolean condition as <parameter>assertion</parameter>,
-    this condition will not be passed as a parameter to the assertion callback
-    which you may have defined with <function>assert_options</function>.
-    Rather, the callback will receive an empty string.
-   </para>
-   <para>
-    Assertions should be used as a debugging feature only. You may
-    use them for sanity-checks that test for conditions that should
-    always be &true; and that indicate some programming errors if not
-    or to check for the presence of certain features like extension
-    functions or certain system limits and features.
-   </para>
-   <para>
-    Assertions should not be used for normal runtime operations like
-    input parameter checks. As a rule of thumb your code should
-    always be able to work correctly if assertion checking is not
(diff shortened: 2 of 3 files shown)


reference/info/functions/assert-options.xml
```diff
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis>
    <type>mixed</type><methodname>assert_options</methodname>
-   <methodparam><type>int</type><parameter>what</parameter></methodparam>
+   <methodparam><type>int</type><parameter>option</parameter></methodparam>
```
Crell, line 13: I sincerely doubt anyone is calling this with named args, but should we be worried about this as a BC issue?  What does the actual function call it?

Girgias: The actual name in the stubs is ``option`` that's why I changed it.

reference/info/functions/assert.xml
```diff
@@ -125,32 +59,102 @@
...
+         <methodparam><type>string</type><parameter>assertion</parameter></methodparam>
+         <methodparam choice="opt"><type>string</type><parameter>description</parameter></methodparam>
+        </methodsynopsis>
        </entry>
       </row>
+      <row>
+       <entry><link linkend="ini.assert.exception">assert.exception</link></entry>
+       <entry>&true;</entry>
+       <entry>
+        If &true; will throw a <classname>AssertionError</classname> if the
```
Crell, line 102:
```suggestion
        If &true; will throw an <classname>AssertionError</classname> if the
```

reference/info/functions/assert.xml
```diff
@@ -125,32 +59,102 @@
...
+        If &true; will abort execution of the PHP script if the
+        expectation isn't upheld.
+       </entry>
+       <entry/>
+      </row>
+      <row>
+       <entry><link linkend="ini.assert.warning">assert.warning</link></entry>
+       <entry>&ture;</entry>
+       <entry>
+        If &true; will emit an <constant>E_WARNING</constant> if the
```
Crell, line 120:
```suggestion
        If &true;, will emit an <constant>E_WARNING</constant> if the
```

reference/info/functions/assert.xml
```diff
@@ -125,32 +59,102 @@
...
+    activated.
+   </para>
+   <para>
+    <function>assert</function> callbacks are particularly useful for
+    building automated test suites because they allow you to easily
+    capture the code passed to the assertion, along with information
+    on where the assertion was made. While this information can be
+    captured via other methods, using assertions makes it much faster
+    and easier!
+   </para>
```
Crell, line 157: Wait, really?  I need to know more here. :-)  Which "you" is intended?  The test suite author, or the test author?  (Viz, Sebastian or me?)

reference/info/functions/assert.xml
```diff
@@ -162,16 +166,18 @@
      <term><parameter>assertion</parameter></term>
      <listitem>
       <para>
-       The assertion. In PHP 5, this must be either a <type>string</type> to
-       be evaluated or a <type>bool</type> to be tested. In PHP 7, this may
-       also be any expression that returns a value, which will be executed and
-       the result used to indicate whether the assertion succeeded or failed.
+       This may be any expression that returns a value, which will be executed
```
Crell, line 169:
```suggestion
       This is any expression that returns a value, which will be executed
```

reference/info/functions/assert.xml
```diff
@@ -180,23 +186,28 @@
...
+       <classname>Throwable</classname>, it will be thrown if the
+       <parameter>assertion</parameter> fails.
+       This is done <emphasis>prior</emphasis> to calling the potentially
+       defined assertion callback.
+       As of PHP 8.0.0, the &object; will be thrown regardless of the configuration of
+       <link linkend="ini.assert.exception">assert.exception</link>.
+       As of PHP 8.0.0, the
+       <link linkend="ini.assert.bail">assert.bail</link>
+       setting has no effect in this case.
+      </para>
```
Crell, line 199: Important question: Is the throwable creation evaluated unconditionally?  Viz, if the assertion passes, do we still spend the CPU cycles to generate a backtrace for the exception that never gets thrown?  (Which is one of the slowest operations you can do in PHP.)  What about if `zend.assertions` is set to 0, so the expression is never evaluated?

reference/info/functions/assert.xml
```diff
@@ -277,81 +294,8 @@
...
- Assertion failed at test.php:21: 2 < 1
- Assertion failed at test.php:22: 2 < 1: Two is less than one
- ]]>
-     </screen>
-    </example>
-   </para>
-  </refsect2>
-  <refsect2>
-   <title>Expectations (PHP 7 only)</title>
+   <title>Expectations</title>
```
Crell, line 297: Do we still need to use this name?  I can't recall anyone using it since 7.0 came out, to differentiate from old-skool string assertions.

reference/info/functions/assert.xml
```diff
@@ -444,6 +388,78 @@ Stack trace:
...
+
+// Set up the callback
+assert_options(ASSERT_CALLBACK, 'my_assert_handler');
+
+// Make an assertion that should fail
+assert('mysql_query("")');
+?>
+]]>
+     </programlisting>
+    </example>
```
Crell, line 421: What's the output for this example?

reference/info/functions/assert.xml
```diff
@@ -444,6 +388,78 @@ Stack trace:
...
+   </para>
+   <para>
+    <example>
+     <title>Using a custom handler to print a description</title>
+     <programlisting role="php">
+      <![CDATA[
+<?php
+// Active assert and make it quiet
```
Crell, line 429:
```suggestion
// Activate assert and make it quiet
```

reference/info/functions/assert.xml
```diff
@@ -444,6 +388,78 @@ Stack trace:
...
   </refsect2>
+  <refsect2>
+   <title>Evaluated code assertions (PHP 7 only)</title>
+   <para>
+    <example>
+     <title>Handle a failed assertion with a custom handler</title>
+     <programlisting role="php">
+      <![CDATA[
+<?php
+// Active assert and make it quiet
```
Crell, line 399:
```suggestion
// Activate assert and make it quiet
```

reference/info/functions/assert.xml
Crell, line 429:
```suggestion
// Activate assert and make it quiet
```

reference/info/functions/assert.xml
```diff
@@ -444,6 +388,78 @@ Stack trace:
     </screen>
    </example>
   </refsect2>
+  <refsect2>
+   <title>Evaluated code assertions (PHP 7 only)</title>
```
Crell, line 392: Is the "PHP 7 only" still relevant?

reference/info/functions/assert.xml
```diff
@@ -125,32 +59,102 @@
...
-          (compatible with PHP 5 behaviour)
-         </member>
-        </simplelist>
+        A user defined function to call when an assertion fails.
+        It's signature should be:
+        <methodsynopsis role="procedural">
+         <type>void</type><methodname>assert_callback</methodname>
+         <methodparam><type>string</type><parameter>file</parameter></methodparam>
+         <methodparam><type>int</type><parameter>line</parameter></methodparam>
+         <methodparam><type>null</type><parameter>assertion</parameter></methodparam>
```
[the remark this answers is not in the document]
Girgias: Because it can as of PHP 8.2.

The reason being is that prior to PHP 8.0.0, to support the "legacy" PHP 5 assertion it would pass in the ``eval``ed code, and if there was no code it would pass an empty string. This support was dropped and since 8.0.0 ``null`` is the only value ever passed

reference/info/functions/assert.xml
Girgias: No, the backtrace is only created if the assertion is ``false``, and as far as I can see if the ``assert`` function is not actually called then it won't be generated either.

Although I struggle to see the point in ``zend.assertion=0``

reference/info/functions/assert.xml
Girgias: Pre-existing example so I don't know. But I will rewrite this and check the output via 3v4l.org

reference/info/functions/assert.xml
Crell: Good to know.  Let's include that explicitly in the page, because that's always made me scared. 😄 

So the internal logic is effectively:

```php
if (code snippet from first param) {
  throw code snippet from second param, which creates an exception/error.
}
```

Let's say that explicitly.

reference/info/functions/assert.xml
Crell: So, we can fix the terminology in 9.  Oh well.

reference/info/functions/assert.xml
Girgias: I don't know how you want me to clarify this as it already said "it will be thrown if the <parameter>assertion</parameter> fails", I've slightly amended the wording but this is pretty clear to me.

Crell: It will only be thrown then, but the point is that it's also only *created* if the test fails.  It's a small distinction but IMO an important one: it's instantiating an exception that is expensive, not throwing it, and you can throw exceptions from places other than where they're created.  (Which is annoying when it happens, but systems will do that.)

Girgias: This is not how exceptions work. Instantiating an exception is creating a base object, see: https://3v4l.org/UGLYO.
The instantiation still happens as usual, as the object passed needs to be created.

The expensive operation of generating the backtrace *only* happens when an object is thrown.

Crell: https://3v4l.org/pApFY

Then what's going on here?

Girgias: TIL it does create the trace on creation: https://heap.space/xref/php-src/Zend/zend_exceptions.c?r=9d5f2f13#255

But I really don't see how this should impact the description of assert(), it is "just" a normal function with some minor compile time magic happening to propagate the second param if none is provided. So it behaves like any other function expecting a Throwable.

Crell: Mainly I'm looking for a warning that the cost of the exception creation (which is not small) is always carried in an assertion, even one that passes, which is different than if you had written your own conditional with a throw.  (Which, incidentally, also makes me vastly less likely to ever use an assertion.  This is why they're disableable, I guess.  And gross.  Sigh.)

Girgias: I still don't see the point of such a warning.

The point of ``assert()`` is for debug/dev "builds" to check assumptions are indeed uphold. Yes a lot of those can be moved to the type system nowadays but not all. So it being "costly" is ... a non-issue.

Assertions are *not* guards. If a certain state can happen, then assertions are *not* the tool you should use.

I don't know *why* support for custom exceptions was added, maybe to make it clearer that when it does fail you can have a stack trace with a good description just by the name of the exception.

PR #2473, opened 2023-05-13
Fold in comments for pathinfo()

Crell:
These are the only two changes worth keeping from the comments section, most of which I have now purged.

### DIFF
--- a/reference/filesystem/functions/pathinfo.xml
+++ b/reference/filesystem/functions/pathinfo.xml
@@ -197,6 +197,11 @@ Array
   <para>
    <example>
     <title><function>pathinfo</function> example with array dereferencing</title>
+    <para>
+     The <parameter>$flags</parameter> parameter is not a bitmask. Only a single value
+     may be provided.  To select only a select set of parsed values, use array
+     dereferencing like so:
+    </para>
     <programlisting role="php">
 <![CDATA[
 <?php
--- a/reference/filesystem/functions/pathinfo.xml
+++ b/reference/filesystem/functions/pathinfo.xml
@@ -183,10 +183,10 @@ print_r(pathinfo('/some/path/.test'));
 <![CDATA[
 Array
 (
-	   [dirname] => /some/path
-	   [basename] => .test
-	   [extension] => test
-	   [filename] =>
+    [dirname] => /some/path
+    [basename] => .test
+    [extension] => test
+    [filename] =>
 )
 ]]>
     </screen>
@@ -228,23 +228,23 @@ string(15) "/www/htdocs/inc"
  </refsect1>
 
 </refentry>
- <!-- Keep this comment at the end of the file
- Local variables:
- mode: sgml
- sgml-omittag:t
- sgml-shorttag:t
- sgml-minimize-attributes:nil
- sgml-always-quote-attributes:t
- sgml-indent-step:1
- sgml-indent-data:t
- indent-tabs-mode:nil
- sgml-parent-document:nil
- sgml-default-dtd-file:"~/.phpdoc/manual.ced"
- sgml-exposed-tags:nil
- sgml-local-catalogs:nil
- sgml-local-ecat-files:nil
- End:
- vim600: syn=xml fen fdm=syntax fdl=2 si
- vim: et tw=78 syn=sgml
- vi: ts=1 sw=1
- -->
+<!-- Keep this comment at the end of the file
+Local variables:
+mode: sgml
+sgml-omittag:t
+sgml-shorttag:t
+sgml-minimize-attributes:nil
+sgml-always-quote-attributes:t
+sgml-indent-step:1
+sgml-indent-data:t
+indent-tabs-mode:nil
+sgml-parent-document:nil
+sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+sgml-exposed-tags:nil
+sgml-local-catalogs:nil
+sgml-local-ecat-files:nil
+End:
+vim600: syn=xml fen fdm=syntax fdl=2 si
+vim: et tw=78 syn=sgml
+vi: ts=1 sw=1
+-->
--- a/reference/filesystem/functions/pathinfo.xml
+++ b/reference/filesystem/functions/pathinfo.xml
@@ -108,6 +108,11 @@
    If <parameter>flags</parameter> is present, returns a
    <type>string</type> containing the requested element.
   </para>
+  <para>
+   On Windows systems only, the <literal>\</literal> character will
+   be interpreted as a directory separator. On Unix systems it will
+   be treated like any other character.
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/pathinfo.xml
```diff
@@ -202,23 +228,23 @@ Array
...
-sgml-default-dtd-file:"~/.phpdoc/manual.ced"
-sgml-exposed-tags:nil
-sgml-local-catalogs:nil
-sgml-local-ecat-files:nil
-End:
-vim600: syn=xml fen fdm=syntax fdl=2 si
-vim: et tw=78 syn=sgml
-vi: ts=1 sw=1
--->
+ <!-- Keep this comment at the end of the file
```
kamil-tekiela, line 231: Why does this need indentation?

kamil-tekiela (approved):

reference/filesystem/functions/pathinfo.xml
```diff
@@ -108,6 +108,11 @@
    If <parameter>flags</parameter> is present, returns a
    <type>string</type> containing the requested element.
   </para>
+  <para>
+   On Windows systems only, the <literal>\</literal> character will
+   be interpreted as a directory separator. On Unix systems it will
+   be treated like any other character.
+  </para>
```
salathe, line 115: Why is paragraph this added to the "Return Values" section?

reference/filesystem/functions/pathinfo.xml
```diff
@@ -178,11 +183,32 @@ print_r(pathinfo('/some/path/.test'));
...
+	   [extension] => test
+	   [filename] =>
 )
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>pathinfo</function> example with array dereferencing</title>
```
salathe, line 197: Why do we want/need an example using array dereferencing for this specific function?

reference/filesystem/functions/pathinfo.xml
```diff
@@ -178,11 +183,32 @@ print_r(pathinfo('/some/path/.test'));
...
 Array
 (
-    [dirname] => /some/path
-    [basename] => .test
-    [extension] => test
-    [filename] => 
+	   [dirname] => /some/path
+	   [basename] => .test
+	   [extension] => test
+	   [filename] =>
```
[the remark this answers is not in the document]
Crell: Ugh.  Damned IDE.

reference/filesystem/functions/pathinfo.xml
Crell: *sigh* Because my IDE insists on indenting things differently than current conventions.

reference/filesystem/functions/pathinfo.xml
Crell: ... Because I just saw a group of note elements and put it with them without thinking. 😄   I'll move it.

reference/filesystem/functions/pathinfo.xml
salathe: If more any examples are to be added to this page, I'd take one showing some use of the `$flags` parameter rather than this one.  

Perhaps this example could be more useful in the context of explaining that the `$flags` parameter is not a bitmask (i.e. `PATHINFO_DIRNAME | PATHINFO_FILENAME` won't return just those).  Just pondering.

Crell: Added in descriptive text along these lines.

reference/filesystem/functions/pathinfo.xml
```diff
@@ -32,6 +32,13 @@
     as "<literal>..</literal>".
    </para>
   </note>
+  <note>
+   <para>
+    On Windows systems only, the <literal>\</literal> character will
+    be interpreted as a directory separator. On Unix systems it will
+    be treated like any other character.
```
salathe, line 39: The "On Unix systems..." is redundant and leads to the question, "what about other systems?".

reference/filesystem/functions/pathinfo.xml
```diff
@@ -181,8 +188,34 @@ Array
...
 )
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>pathinfo</function> example with array dereferencing</title>
+    <para>
+     The <parameter>$flags</parameter> parameter is not a bitmask. Only a single value
```
salathe, line 201:
```suggestion
     The <parameter>flags</parameter> parameter is not a bitmask. Only a single value
```

reference/filesystem/functions/pathinfo.xml
```diff
@@ -181,8 +188,34 @@ Array
...
+     may be provided.  To select only a select set of parsed values, use array
```
salathe, line 202: "To select only a select…" is clunky.

reference/filesystem/functions/pathinfo.xml
```diff
@@ -181,8 +188,34 @@ Array
...
+     dereferencing like so:
+    </para>
+    <programlisting role="php">
+<![CDATA[
+<?php
+[ 'basename' => $basename, 'dirname' => $dirname ] = pathinfo('/www/htdocs/inc/lib.inc.php');
```
salathe, line 208:
```suggestion
['basename' => $basename, 'dirname' => $dirname] = pathinfo('/www/htdocs/inc/lib.inc.php');
```

reference/filesystem/functions/pathinfo.xml
```diff
@@ -181,8 +188,34 @@ Array
...
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>pathinfo</function> example with array dereferencing</title>
+    <para>
+     The <parameter>$flags</parameter> parameter is not a bitmask. Only a single value
+     may be provided.  To select only a select set of parsed values, use array
+     dereferencing like so:
```
salathe, line 203:
```suggestion
     may be provided.  To select only a select set of parsed values, use array
     destructuring like so:
```

reference/filesystem/functions/pathinfo.xml
```diff
@@ -32,6 +32,13 @@
     as "<literal>..</literal>".
    </para>
   </note>
+  <note>
+   <para>
+    On Windows systems only, the <literal>\</literal> character will
+    be interpreted as a directory separator. On Unix systems it will
```
Crell, line 38:
```suggestion
    be interpreted as a directory separator. On other systems it will
```

reference/filesystem/functions/pathinfo.xml
Crell:
```suggestion
     may be provided.  To select only a limited set of parsed values, use array
```

Issue #2423, opened 2023-04-13
printf formatting specifiers; no documentation of argument-provided padding.


MorganLOCode:
Since PHP 8.0, a [width] specifier of `*` has been allowed, which is replaced by the corresponding next argument used (or however that should be phrased).
```php
printf("[%'#*s]\n", 10, 'monkey');
```
being equivalent to
```php
printf("[%'#10s]\n", 'monkey');
```
results in
```text
[####monkey]
```
but this is not yet documented. The description for "width" is merely that it is an integer specifying a minimum width, not that it can be parameterised in the above fashion.

Ditto for the other *printf functions.
http://www.php.net/printf
http://www.php.net/sprintf
http://www.php.net/fprintf
http://www.php.net/vprintf
http://www.php.net/vsprintf
http://www.php.net/vfprintf

MorganLOCode:
Just checked, and confirmed that [precision] can also be `*`d.

---

**Width**
Either an integer that says how many characters (minimum) this conversion should result in, or `*`. If `*` is used, then the width is supplied as an additional integer value preceding the one formatted by the specifier.

Girgias:
Fixed via #2474

PR #2474, opened 2023-05-14
Mention the use of '*' in printf-formatting specifiers

MorganLOCode:
When '*' is used to indicate width or precision in a formatting specifier, an additional integer argument is to be supplied which will provide the width/precision to be used.

#2423


Girgias:
Thank you!

PR #2475, opened 2023-05-14
Update timezones.xml

Velman04:
Rename Europe/Kyiv to Europe/Kiev


MorganLOCode:
Isn't `Europe/Kiev` already listed under "Other" timezones for BC purposes?

PR #2476, opened 2023-05-15
parallel\Runtime methods descriptions were obviously wrong

pozitronik:
`parallel\Runtime::close` and `parallel\Runtime::kill` methods descriptions fixed ('join' => 'shutdown')

### DIFF
--- a/reference/parallel/parallel.runtime.xml
+++ b/reference/parallel/parallel.runtime.xml
@@ -64,7 +64,7 @@
      <xi:fallback />
     </xi:include>
 
-    <classsynopsisinfo role="comment">Join</classsynopsisinfo>
+    <classsynopsisinfo role="comment">Shutdown</classsynopsisinfo>
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.parallel-runtime')/db:refentry/db:refsect1[@audience='shutdown']/descendant::db:methodsynopsis[not(@role='procedural')])">
      <xi:fallback />
     </xi:include>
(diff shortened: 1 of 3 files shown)


reference/parallel/parallel.runtime.xml
```diff
@@ -65,7 +65,7 @@
     </xi:include>
 
     <classsynopsisinfo role="comment">Join</classsynopsisinfo>
```
Girgias, line 67: This needs to be changed as well.

But I don't know if it isn't accurate, as the purpose of closing runtimes might be to join results together?

pozitronik: >This needs to be changed as well.

Done.

>But I don't know if it isn't accurate, as the purpose of closing runtimes might be to join results together?

The logic of those calls and this extension itself does not assume any joins at all, as I see. Since parallel\Runtime isn't well-documented, let's call the author to be certain: @krakjoe

lacatoire:
Thanks for spotting this, and sorry for the very long delay.

The grouping is indeed misleading: `close()` and `kill()` shut the runtime down, while `join()` is a separate method, so labelling them "Join" is confusing.

Unfortunately this cannot be merged here: `ext/parallel` is being migrated to [php/doc-extensions](https://github.com/php/doc-extensions/), and changes made in doc-en after the migration notice was added are discarded. The files still carry the old wording there, so the fix is still worth doing, against php/doc-extensions instead.

If you reopen it there, the `refpurpose` of both pages would be worth changing too ("Runtime Graceful Join" and "Runtime Join"), otherwise the wording stays half-changed.

Closing here for that reason, not because the change is wrong.

PR #2478, opened 2023-05-15
tutorial: Promote lowercase boolean to beginners

Krinkle:
Update the example to a browser that people can actually download and install from official websites on macOS, Linux, and Windows today.

MSIE is EOL and was Windows-only (apart from [IE 5.5 for Mac](https://www.versionmuseum.com/history-of/internet-explorer), back in the day). Note that even IE11 no longer identified as MSIE, it identified as Trident instead.


Girgias:
The tutorial section is very outdated, I'm not sure doing such tiny improvements is worth it.

Maybe there should be a discussion about ripping it apart, don't know.

PR #2481, opened 2023-05-17
Update the documentation of getenv

kocsismate:


reference/info/functions/getenv.xml
```diff
@@ -33,18 +29,19 @@
...
       </para>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><parameter>local_only</parameter></term>
      <listitem>
       <para>
-       Set to true to only return local environment variables (set by the operating system or putenv).
+       When set to &true;, only local environment variables are returned, set by the operating system or putenv. It only has
+       an effect when <parameter>name</parameter> is a &string;.
```
Girgias, line 44: Is that what really happens [sapi_getenv](https://heap.space/s?defs=sapi_getenv&project=php-src) gets called?

reference/info/functions/getenv.xml
```diff
@@ -56,13 +53,13 @@
   &reftitle.returnvalues;
   <para>
    Returns the value of the environment variable
-   <parameter>varname</parameter>, or &false; if the environment
-   variable <parameter>varname</parameter> does not exist.
-   If <parameter>varname</parameter> is omitted, all environment variables are
+   <parameter>name</parameter>, or &false; if the environment
+   variable <parameter>name</parameter> does not exist.
+   If <parameter>name</parameter> is &null;, all environment variables are
    returned as associative <type>array</type>.
```
Girgias, line 59: Nit: while you're at it
```suggestion
   returned as an associative <type>array</type>.
```

Girgias (commented):
Just a small questions but LGTM

reference/info/functions/getenv.xml
kocsismate: I'm not completely sure, but I think so. E.g. when `sapi_fcgi_getenv` is triggered, it seems to return env vars from the request. The comment there also indicates this:
>/* when php is started by mod_fastcgi, no regular environment
	 * is provided to PHP.  It is always sent to PHP at the start
	 * of a request.  So we have to do our own lookup to get env
	 * vars.  This could probably be faster somehow.  */

kocsismate: I merged the PR in the meanwhile, but feel free to ping if there's anything you miss.

PR #2486, opened 2023-05-19
Clarify mktime and gmmktime without args

camporter:
This just adds some clarification for mktime and gmmktime about whether passing no arguments is still deprecated or not, and makes it obvious that as of PHP 8.0, it will throw an error.

### DIFF
--- a/reference/datetime/functions/gmmktime.xml
+++ b/reference/datetime/functions/gmmktime.xml
@@ -23,14 +23,17 @@
    so only times valid in derived local time can be used.
   </para>
   <para>
-   Like <function>mktime</function>, arguments may be left out in order
+   Like <function>mktime</function>, any arguments (other than the first) may be left out in order
    from right to left, with any omitted arguments being set to the
    current corresponding GMT value.
   </para>
-  <simpara>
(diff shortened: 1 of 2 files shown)


reference/datetime/functions/gmmktime.xml
```diff
@@ -23,14 +23,17 @@
...
    current corresponding GMT value.
   </para>
-  <simpara>
-   Calling <function>gmmktime</function> without arguments is deprecated.
-   <function>time</function> can be used to get the current timestamp.
-  </simpara>
+  <warning>
+   <simpara>
+    As of PHP 8.0.0, an <classname>ArgumentCountError</classname> is thrown if
+    the number of arguments is zero.
```
Girgias, line 33:
```suggestion
    As of PHP 8.0.0, <parameter>hour</parameter> is no longer optional.
```

Same change for the other function

camporter: Thanks! Updated.

derickr:
Thanks for this! I've updated it with a slightly different text. Convention is to mention changes only in the "changelog" section, and not in the main body which should document the functions as they work in the latest version (8.0+, in our case).

PR #2487, opened 2023-05-20
Create SECURITY.md

RZ7272:


sy-records (changes requested):
Thank you for your PR. 
But I don't think this file should be added. And the version number is wrong.

Girgias:
We have such a file for php-src: https://github.com/php/php-src/blob/master/SECURITY.md

This file makes no sense on the doc repo.

PR #2491, opened 2023-05-23
Fix typo and expected bool value

stchr:


reference/soap/soapclient/dorequest.xml
```diff
@@ -65,7 +65,7 @@
      <term><parameter>oneWay</parameter></term>
      <listitem>
       <para>
-       If <literal>one_way</literal> is set to 1, this method returns nothing. 
+       If <literal>oneWay</literal> is set to true, this method returns nothing.
```
Girgias, line 68:
```suggestion
       If <parameter>oneWay</parameter> is set to &true;, this method returns nothing.
```

Girgias:
Thank you!

PR #2496, opened 2023-05-26
Adding example of why enum inheritance is not allowed.

Danack:
Hoping the CI works to test it's formatted okay....

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -753,7 +753,7 @@ function bar(B $b) {
 
  <simpara>
   This code is type-safe, as B follows the contract of A, and through the magic of
-  co/contra-variance, any expectation you may have of the methods will be
+  co/contra-variance, any expectation one may have of the methods will be
   preserved, exceptions excepted.
  </simpara>
 
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -752,7 +752,7 @@ function bar(B $b) {
  </programlisting>
 
  <simpara>
-  This is safe, as B follows the contract of A, and through the magic of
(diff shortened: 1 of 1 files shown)


language/enumerations.xml
```diff
@@ -727,6 +727,103 @@ Baz Enum:int {
...
+    foo($b);
+}
+?>
+]]>
+ </programlisting>
+
+ <simpara>
+  This code is type-safe, as B follows the contract of A, and through the magic of
+  co/contra-variance, any expectation you may have of the methods will be
+  preserved, exceptions excepted.
```
Girgias, line 757: The manual shouldn't use personalisation (i.e. using "you")
So maybe something like:
```suggestion
  This code is type-safe, as B follows the contract of A, and through the magic of
  co/contra-variance, any expectation that is hold about methods will be
  preserved, exceptions excepted.
```
?

Girgias (commented):
The indentation is also somewhat off at the beginning where it should have 2 spaces instead of 1

language/enumerations.xml
Danack: "The manual shouldn't use personalisation"

Kind of hard to refer to someone's state of mind without doing that.....impersonal pronoun it is then.

Girgias (approved):
Thank you!

language/enumerations.xml
```diff
@@ -727,6 +727,103 @@ Baz Enum:int {
...
+  <simpara>
+   Classes have contracts on their methods:
+  </simpara>
+
+  <programlisting role="php">
+<![CDATA[
+<?php
+
+class A {}
+class B {}
```
TimWolla, line 743: Is extends missing here?

PR #2498, opened 2023-05-28
may also return DOMElement

divinity76:
Not explicitly documenting the possibility of returning DOMElement causes the Intelephense linter (a popular PHP linter with ~9 million downloads: https://marketplace.visualstudio.com/items?itemName=bmewburn.vscode-intelephense-client ) to think this code is bad:

$xp->query("whatever")->item(0)->getAttribute("foo");

because DOMNode does not have getAttribute (while DOMElement does). documenting the DOMElement return type should fix Intelephense's linter.

### DIFF
--- a/reference/dom/domnodelist/item.xml
+++ b/reference/dom/domnodelist/item.xml
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis role="DOMNodeList">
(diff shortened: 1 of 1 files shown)


Girgias (changes requested):
This should be fixed in the php-src stubs and then synced back.

divinity76:
@Girgias interesting. made a PR: https://github.com/php/php-src/pull/11339

PR #2501, opened 2023-05-29
Clarifications about getElementsByTagNameNS

nielsdos:
And while we're here, make the documentation consistent.

Never contributed to the docs before, so not sure about wording etc :P

### DIFF
--- a/reference/dom/domdocument/getelementsbytagnamens.xml
+++ b/reference/dom/domdocument/getelementsbytagnamens.xml
@@ -28,7 +28,8 @@
      <listitem>
       <para>
        The namespace URI of the elements to match on. 
-       The special value <literal>*</literal> matches all namespaces.
+       The special value <literal>*</literal> matches all namespaces. 
+       Passing <type>null</type> matches the empty namespace.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/dom/domdocument/getelementsbytagnamens.xml
+++ b/reference/dom/domdocument/getelementsbytagnamens.xml
@@ -28,8 +28,8 @@
      <listitem>
       <para>
        The namespace URI of the elements to match on.
-       The special value <literal>*</literal> matches all namespaces.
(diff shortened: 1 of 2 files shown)


reference/dom/domdocument/getelementsbytagnamens.xml
```diff
@@ -27,8 +27,9 @@
      <term><parameter>namespace</parameter></term>
      <listitem>
       <para>
-       The namespace URI of the elements to match on. 
+       The namespace URI of the elements to match on.
        The special value <literal>*</literal> matches all namespaces.
+       Passing <type>null</type> matches the empty namespace.
```
Girgias, line 32:
```suggestion
       Passing &null; matches the empty namespace.
```
For the value `null` we use the entity (we have some similarly for true and false, you can have a look in `language-snippets.ent` and another `.ent` file the ones we define)

reference/dom/domdocument/getelementsbytagnamens.xml
```diff
@@ -27,8 +27,9 @@
      <term><parameter>namespace</parameter></term>
      <listitem>
       <para>
-       The namespace URI of the elements to match on. 
+       The namespace URI of the elements to match on.
        The special value <literal>*</literal> matches all namespaces.
```
Girgias, line 31:
```suggestion
       The special value <literal>"*"</literal> matches all namespaces.
```
I would enclose it in quotes to show that we mean the string with an asteriks

Girgias (changes requested):
Ditto the changes onto the other page.

nielsdos:
Thanks, applied the suggestions.

Girgias (approved):
I wonder if it makes sense to deprecate the behaviour of passing ``null`` to these functions as it is rather confusing.

Girgias:
So the spec is just weird :| yeah let's follow the spec.

PR #2504, opened 2023-05-31
Add an example for str_getcsv with empty input

apeschar:
It's surprising that `str_getcsv("")` returns `[null]`. This adds an example to warn users of this behavior.

### DIFF
--- a/reference/strings/functions/str-getcsv.xml
+++ b/reference/strings/functions/str-getcsv.xml
@@ -152,6 +152,31 @@ array(5) {
     </screen>
    </example>
   </para>
+  <para>
+   <example>
+    <title><function>str_getcsv</function> example with empty input</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+$string = '';
+$data = str_getcsv($string);
+
+var_dump($data);
+?>
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+array(1) {
+  [0]=>
+  NULL
+}
+]]>
+    </screen>
+   </example>
+  </para>
  </refsect1>
 
  <refsect1 role="seealso">
--- a/reference/strings/functions/str-getcsv.xml
+++ b/reference/strings/functions/str-getcsv.xml
@@ -154,7 +154,13 @@ array(5) {
   </para>
   <para>
    <example>
-    <title><function>str_getcsv</function> example with empty input</title>
+    <title><function>str_getcsv</function> example with an empty string</title>
+    <caution>
+     <simpara>
+      On an empty string this function returns the value <code>[null]</code>
+      instead of an empty array.
+     </simpara>
+    </caution>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
I think an example on its own is insufficient. I wonder why is the value `NULL`. Is this by design? Do you have any supporting evidence that this behaviour should be documented?

apeschar:
> I think an example on its own is insufficient.

I've added a note to the return values section to match the note that is attached to [fgetcsv](https://www.php.net/fgetcsv). Without the error part, since str_getcsv never fails.

> I wonder why is the value NULL. Is this by design?

It's like this because of the original implementation of fgetcsv.

https://github.com/php/php-src/blob/master/ext/standard/string.c#L5139-L5143

There's this commit which makes the "weird" behaviour more explicit:

https://github.com/php/php-src/commit/4ccf0b01819677dbdd3fbb330f9220606866028b

I assume that at this point it's unlikely for this behaviour to be changed?

> Do you have any supporting evidence that this behaviour should be documented?

I think the fact that it's noted in the docs for fgetcsv, is mentioned as weird in the above commit and that it surprised me (and you?) should indicate that documenting this behaviour is useful.

reference/strings/functions/str-getcsv.xml
```diff
@@ -148,6 +148,31 @@ array(5) {
   [4]=>
   string(5) "Swift"
 }
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>str_getcsv</function> example with empty input</title>
```
Girgias, line 157:
```suggestion
    <title><function>str_getcsv</function> example with an empty string</title>
    <caution>
     <simpara>
      On an empty string this function returns the value <code>[null]</code>
      instead of an empty array.
     </simpara>
    </caution>
```

Girgias (changes requested):
This function is cursed, and I don't think it will ever change.

Moreover, for some reason, it also right trims newlines: https://3v4l.org/UJ8JN

Pure joy.

apeschar:
For your enjoyment, I'll share another example where it introduces a null byte when you give it an unterminated quoted string:

~~~
> str_getcsv('""')
= [
    "",
  ]

> str_getcsv('"')
= [
    "\0",
  ]
~~~

https://3v4l.org/NsGVO

kamil-tekiela:
> For your enjoyment, I'll share another example where it introduces a null byte when you give it an unterminated quoted string:
> 
> ```
> > str_getcsv('""')
> = [
>     "",
>   ]
> 
> > str_getcsv('"')
> = [
>     "\0",
>   ]
> ```
> 
> https://3v4l.org/NsGVO

This must be a bug.

apeschar:
> This must be a bug.

Created an issue for it:
https://github.com/php/php-src/issues/11982

Girgias (approved):

PR #2516, opened 2023-06-06
Reflection: fix some "undocumented" methods, and some other edits for consistency

tstarling:
* If a method has a refpurpose, description, parameters and return
  values, it's documented, there's no need for the warning box.
* Add description where there was none. It is usually very similar
  to the refpurpose.
* Document ReflectionZendExtension::__construct(), and update
  ReflectionExtension::__construct() for consistency.
* In ReflectionFunction::__toString()
* Remove some references to deprecated export() methods.
* Add example for ReflectionParameter::__toString().

### DIFF
--- a/reference/reflection/reflectionclass/istrait.xml
+++ b/reference/reflection/reflectionclass/istrait.xml
@@ -26,7 +26,6 @@
   &reftitle.returnvalues;
   <para>
    Returns &true; if this is a trait, &false; otherwise.
-   Returns &null; in case of an error.
   </para>
  </refsect1>
 
--- a/reference/reflection/reflectionclass/istrait.xml
+++ b/reference/reflection/reflectionclass/istrait.xml
@@ -13,11 +13,8 @@
    <void/>
   </methodsynopsis>
   <para>
-
+   Check whether this <type>ReflectionClass</type> refers to a trait.
   </para>
-
-  &warn.undocumented.func;
-
  </refsect1>
 
  <refsect1 role="parameters">
--- a/reference/reflection/reflectionfunction/getclosure.xml
+++ b/reference/reflection/reflectionfunction/getclosure.xml
@@ -13,10 +13,8 @@
    <void/>
   </methodsynopsis>
   <para>
-
+   Get a dynamically created closure for the function.
   </para>
-
-  &warn.undocumented.func;
 
  </refsect1>
 
--- a/reference/reflection/reflectionfunction/getclosure.xml
+++ b/reference/reflection/reflectionfunction/getclosure.xml
@@ -26,10 +26,16 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns <classname>Closure</classname>.
+   Returns the newly created <classname>Closure</classname>.
   </para>
  </refsect1>
 
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <simplelist>
+   <member><link linkend="functions.first_class_callable_syntax">First class callable syntax</link></member>
+  </simplelist>
+ </refsect1>
 
 </refentry>
 <!-- Keep this comment at the end of the file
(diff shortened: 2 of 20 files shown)


reference/reflection/reflectionclass/istrait.xml
```diff
@@ -13,11 +13,8 @@
    <void/>
   </methodsynopsis>
   <para>
-
+   Check whether this <type>ReflectionClass</type> refers to a trait.
```
Girgias, line 16: Unrelated: Return value section seems to say NULL can be returned when the return type of the function says it only returns `bool`.

If it does indeed return NULL there seems to be an issue with the stubs, if this changed in PHP 8.0 (outside of ZPP checks) this will also need a changelog.

reference/reflection/reflectionfunction/getclosure.xml
```diff
@@ -13,11 +13,9 @@
    <void/>
   </methodsynopsis>
   <para>
-
+   Get a dynamically created closure for the function.
```
Girgias, line 16: Maybe point out that this can also be achieved by using the new ``function_name(...)``  syntax introduced in PHP 8.1 (? or was it 8.0).

Also, the return value section is a bit crude.

reference/reflection/reflectionparameter/tostring.xml
```diff
@@ -28,15 +26,39 @@
...
+ </refsect1>
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><methodname>ReflectionParameter::__toString</methodname> example</title>
+    <programlisting role="php">
+<![CDATA[
+     <?php
```
Girgias, line 40: Nit:
```suggestion
<?php
```

Girgias (commented):
Minor comments, LGTM overall

reference/reflection/reflectionparameter/tostring.xml
```diff
@@ -28,15 +26,39 @@
...
+    <screen>
+<![CDATA[
+Parameter #0 [ <required> string $string ]
+]]>
+    </screen>
+   </example>
   </para>
  </refsect1>
 
+
```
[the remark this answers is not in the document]
tstarling: Fixed.

reference/reflection/reflectionparameter/tostring.xml
tstarling: Done.

reference/reflection/reflectionfunction/getclosure.xml
tstarling: It was 8.1, see https://www.php.net/manual/en/functions.first_class_callable_syntax.php

I see that Closure::fromCallable() has a note box with a link in its description. But I think that here, only a "see also" is warranted. If you have a ReflectionFunction already, then getClosure() is what you want. First class callable syntax is only a replacement if you are considering constructing a ReflectionFunction from a callable just to get the closure.

I expanded the return value section slightly, inspired by Closure::fromCallable().

reference/reflection/reflectionmethod/getclosure.xml
```diff
@@ -13,11 +13,9 @@
    <methodparam choice="opt"><type class="union"><type>object</type><type>null</type></type><parameter>object</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-
+   Create a closure which will call the method.
```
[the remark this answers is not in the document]
tstarling: As for ReflectionFunction, I added a "see also" and made the return values section into a grammatical sentence.

I also removed the bit about it returning null, since I believe it can't do that anymore, based on code review. It returns a Closure or throws an exception.

I added an errors section, since there are some moderately interesting exceptions it can throw.

Girgias (approved):

Girgias:
Thank you!

PR #2517, opened 2023-06-06
Improve documentation of a few functions which had &warn.undocumented.func;

tstarling:


reference/curl/functions/curl-multi-setopt.xml
```diff
@@ -15,11 +15,9 @@
    <methodparam><type>mixed</type><parameter>value</parameter></methodparam>
   </methodsynopsis>
   <para>
-
+   Sets an option on the given cURL multi handle.
```
Girgias, line 18: Don't we have an XML entity for the ``multi_handle`` parameter? Moreover, that parameter is still lacking a description.

reference/mbstring/functions/mb-scrub.xml
```diff
@@ -28,15 +28,19 @@
...
      </para>
     </listitem>
    </varlistentry>
    <varlistentry>
     <term><parameter>encoding</parameter></term>
     <listitem>
      <para>
-      
+      The encoding used to interpret <parameter>string</parameter>.
+      If it is &null; or not specified, the
```
Girgias, line 40: Not sure if we need to say "not specified" as this is how a default value for a parameter works, but I don't mind either way.
```suggestion
      If it is &null;, the
```

Girgias (commented):
Minor issue with the cURL page but LGTM otherwise.

reference/curl/functions/curl-multi-setopt.xml
tstarling: Yes, there is `&curl.mh.description;` which I will put in the parameters section. I'm not sure if you're also asking for a change to the description section. The description I added here is consistent with the one on curl_setopt.

Girgias: No I was just asking to document the ``multi_handle`` parameter :)

reference/mbstring/functions/mb-scrub.xml
tstarling: Here's how the encoding parameter is described in other mbstring functions:

* Omitted or null: mb_chr, mb_convert_case, mb_convert_kana, mb_decode_numericentity, mb_detect_order, mb_encode_numericentity, mb_ord, mb_regex_encoding, mb_str_split, mb_strcut, mb_strimwidth, mb_strlen, mb_strpos, mb_strrpos, mb_strtolower, mb_strtoupper, mb_strwidth, mb_substr_count, mb_substr
* Omitted: mb_convert_variables, mb_http_output, mb_stripos, mb_stristr, mb_strrchr, mb_strrichr, mb_strripos, mb_strstr
* Null or not specified: mb_convert_encoding

You can guess which one I copied from. mb_scrub() shares its implementation with mb_convert_encoding().

I think I will change them both to "omitted or null", I think that is clearer than just "null" or just "omitted". Yes that's how defaults work, but it's a convention, the defaults don't really exist in the C code and it's easy to mess up this detail.

Girgias (approved):

reference/mbstring/functions/mb-scrub.xml
Girgias: Makes sense

PR #2518, opened 2023-06-06
Fix wording

ajgarlag:
Sorry. I made a mistake in my previously merged PR #2508

### DIFF
--- a/language/wrappers/php.xml
+++ b/language/wrappers/php.xml
@@ -44,8 +44,7 @@
     <filename>php://input</filename> is a read-only stream that allows you to
     read raw data from the request body.
     <filename>php://input</filename> is not available in POST requests with
-    <literal>enctype="multipart/form-data"</literal>.
-    <literal>enctype="multipart/form-data"</literal> is available if
+    <literal>enctype="multipart/form-data"</literal> if
     <link linkend="ini.enable-post-data-reading">enable_post_data_reading</link>
     option is enabled.
    </simpara>


ajgarlag:
I see that @mumumu tried to fix it in 381714c59c090a9ccd58cf6cc393fbcc25685b37 but the text still needs some change.

The original text was:
> `php://input` is a read-only stream that allows you to read raw data from the request body. `php://input` is not available with `enctype="multipart/form-data"`. 

After #2508  was merged, the text was:
> `php://input` is a read-only stream that allows you to read raw data from the request body. `php://input` is not available in POST requests with `enctype="multipart/form-data"`. `enctype="multipart/form-data"` if [enable_post_data_reading](https://www.php.net/manual/en/ini.core.php#ini.enable-post-data-reading) option is enabled.

Then, @mumumu changed it in 381714c59c090a9ccd58cf6cc393fbcc25685b37 to:
> `php://input` is a read-only stream that allows you to read raw data from the request body. `php://input` is not available in POST requests with `enctype="multipart/form-data"`. `enctype="multipart/form-data"` is available if [enable_post_data_reading](https://www.php.net/manual/en/ini.core.php#ini.enable-post-data-reading) option is enabled.

Now, this PR tries to change it to my desired original description:
> `php://input` is a read-only stream that allows you to read raw data from the request body. `php://input` is not available in POST requests with `enctype="multipart/form-data"` if [enable_post_data_reading](https://www.php.net/manual/en/ini.core.php#ini.enable-post-data-reading) option is enabled.

As I wrote in my first PR, I have to read the raw request body to develop my reverse proxy properly. I thought it was impossible after reading the original description, so I proposed to link `enable_post_data_reading` option.

Girgias:
Thank you!

PR #2519, opened 2023-06-06
Add note to headers_sent() about empty filename

KapitanOczywisty:
https://github.com/php/doc-en/issues/2523

### DIFF
--- a/reference/network/functions/headers-sent.xml
+++ b/reference/network/functions/headers-sent.xml
@@ -38,6 +38,12 @@
        and line number where output started in the <parameter>filename</parameter>
        and <parameter>line</parameter> variables.
       </para>
+      <note>
+        <para>
+          If the output has started before executing the PHP source file (for example due to a startup error),
+          <parameter>filename</parameter> parameter will be set to an empty string.
+        </para>
+      </note>
      </listitem>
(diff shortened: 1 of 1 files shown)


reference/network/functions/headers-sent.xml
```diff
@@ -38,6 +38,12 @@
        and line number where output started in the <parameter>filename</parameter>
        and <parameter>line</parameter> variables.
       </para>
+      <note>
+        <para>
+          If the output has started before executing the PHP source file (for example due to a startup error),
+          <parameter>filename</parameter> parameter will be set to an empty string.
+        </para>
+      </note>
```
Girgias, line 46:
```suggestion
      <note>
       <para>
         If the output has started before executing the PHP source file (for example due to a startup error),
         the <parameter>filename</parameter> parameter will be set to an empty string.
       </para>
      </note>
```

Girgias (approved):

reference/network/functions/headers-sent.xml
Girgias: Great GitHub doesn't want me to commit from here directly.

Girgias:
Thank you!

PR #2524, opened 2023-06-08
Add "iterable" to the list of built-in types within the Introduction

adrianguenter:
Hello,

I noticed `iterable` wasn't listed with the other built-ins at the top of this page, so I've added it.

Thanks!


Girgias:
This is intended as ``iterable`` is not a built-in type any more since 8.2 where it is a compile time type alias to ``Traversable|array``

PR #2534, opened 2023-06-20
docs: add a reference to array_map from array_combine

dritter:
Last time I searched how to zip up an array, and all I came up with was `array_combine`. Hence, I guess a link from `array_combine` to `array_map` would be helpful.
WDYT?


Girgias:
Thank you!

PR #2535, opened 2023-06-21
Improve str_pad documentation

houssemz:
IIMHO, we need to specify that `length` is the desired length of the final padded chain. In my case, it was unclear.

### DIFF
--- a/reference/strings/functions/str-pad.xml
+++ b/reference/strings/functions/str-pad.xml
@@ -42,6 +42,7 @@
      <term><parameter>length</parameter></term>
      <listitem>
       <para>
+       This is the desired length of the final padded string.
        If the value of <parameter>length</parameter> is negative,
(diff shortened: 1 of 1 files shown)


reference/strings/functions/str-pad.xml
```diff
@@ -42,6 +42,7 @@
      <term><parameter>length</parameter></term>
      <listitem>
       <para>
+       This is the desired length of the final padded string.
```
kamil-tekiela, line 45:
```suggestion
       The desired length of the final padded string.
```

kamil-tekiela (changes requested):
Right, but this is obvious, isn't it? What else would it be?

houssemz:
> 

Thank you for the review.
Not so obvious, IMHO :)

kamil-tekiela:
Thanks.

PR #2540, opened 2023-06-23
Added missing `FILE_NO_DEFAULT_CONTEXT` for file()

staabm:
it seems this flag is available but missing from the docs

### DIFF
--- a/reference/filesystem/functions/file.xml
+++ b/reference/filesystem/functions/file.xml
@@ -76,6 +76,16 @@
           </simpara>
          </listitem>
         </varlistentry>
+        <varlistentry>
+         <term>
+          <constant>FILE_NO_DEFAULT_CONTEXT</constant>
+         </term>
+         <listitem>
+          <simpara>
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/file.xml
```diff
@@ -76,6 +76,16 @@
           </simpara>
          </listitem>
         </varlistentry>
+        <varlistentry>
+         <term>
+          <constant>FILE_NO_DEFAULT_CONTEXT</constant>
+         </term>
+         <listitem>
+          <simpara>
+           Don't use the default context
```
staabm, line 85: I wasn't sure what other implications this flag has. maybe this description could be more useful

Girgias: I think that's the only one it has looking at https://heap.space/xref/php-src/main/streams/php_stream_context.h?r=d22d0e26#31

Girgias (approved):

PR #2544, opened 2023-06-26
Updated ImagickKernel method parameters from stubs.

Danack:
Making PR to test build.


vrana:
Why is this a draft?

lacatoire:
Sorry this sat for so long.

This needs to go to [php/doc-extensions](https://github.com/php/doc-extensions) rather than here: `ext/imagick` is migrating there and carries a `MOVING_TO_DOC_EXTENSIONS_DO_NOT_CHANGE_HERE.md` notice, so anything merged in doc-en would be discarded. The four pages still have the old parameter names over there, so the fix is still needed.

PR #2550, opened 2023-06-29
Fix current build issues

mallardduck:
To address the error in: https://github.com/php/doc-en/commit/6b02ef7aa875ea22b489c8fa2683d87f12da8a1c this PR duplicates the original listitem into the new varlistentry elements.


mallardduck:
Not sure if necessary or not - the commit from @remicollet may have been pushed accidentally. However in either case this will get the builds working again and such.

mallardduck:
Fixed by dd9775e98b7368f4d04847f343250c9627bb8da8

PR #2552, opened 2023-06-30
Update urlencode.xml

NathanGibbs3:
Code examples with output for urlencode()


reference/url/functions/urlencode.xml
```diff
@@ -69,11 +78,23 @@ echo '<a href="mycgi?foo=', urlencode($userinput), '">';
     <programlisting role="php">
 <![CDATA[
 <?php
+$foo = 'Data123!@-_ +';
+$bar = $foo;
```
Girgias, line 82: Probably should have $bar not be the same content as $foo.

Girgias:
Thank you!

PR #2560, opened 2023-07-05
Various fixes for even more interfaces

mallardduck:
Fixes:

- Missing spaces on: a few predefined interfaces,
- Interfaces displaying as classes: cmark, json, pthreads

### DIFF
--- a/language/predefined/backedenum.xml
+++ b/language/predefined/backedenum.xml
@@ -27,9 +27,7 @@
     </ooclass>
 
     <classsynopsisinfo>
-     <oointerface>
-      <interfacename>BackedEnum</interfacename>
-     </oointerface>
+     <ooclass><classname>BackedEnum</classname></ooclass>
 
      <oointerface>
       <interfacename>UnitEnum</interfacename>
--- a/language/predefined/throwable.xml
+++ b/language/predefined/throwable.xml
@@ -35,9 +35,7 @@
     </ooclass>
 
     <classsynopsisinfo>
-     <oointerface>
-      <interfacename>Throwable</interfacename>
-     </oointerface>
+     <ooclass><classname>Throwable</classname></ooclass>
 
      <oointerface>
       <interfacename>Stringable</interfacename>
--- a/language/predefined/iterator.xml
+++ b/language/predefined/iterator.xml
@@ -27,9 +27,7 @@
     </ooclass>
 
     <classsynopsisinfo>
-     <oointerface>
-      <interfacename>Iterator</interfacename>
-     </oointerface>
+     <ooclass><classname>Iterator</classname></ooclass>
 
      <oointerface>
       <interfacename>Traversable</interfacename>
--- a/language/predefined/iteratoraggregate.xml
+++ b/language/predefined/iteratoraggregate.xml
@@ -26,9 +26,7 @@
     </ooclass>
 
     <classsynopsisinfo>
-     <oointerface>
-      <interfacename>IteratorAggregate</interfacename>
-     </oointerface>
+     <ooclass><classname>IteratorAggregate</classname></ooclass>
 
      <oointerface>
       <interfacename>Traversable</interfacename>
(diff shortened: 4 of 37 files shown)


Girgias:
@kocsismate (knowing that you're busy) do you know why the JSON and ``/language/`` files were not updated by your previous stub sync? Is this an issue in our stubs or just that you forgot?

language/predefined/throwable.xml
```diff
@@ -35,9 +35,7 @@
     </ooclass>
 
     <classsynopsisinfo>
```
Girgias, line 37:
```suggestion
    <classsynopsis class="interface">
```

Isn't this missing, or am I not understanding how the new markup should look?

mallardduck: So for the new markup generation to kick in, we want it to call `format_classsynopsisinfo_ooclass_classname` - which means we need a `ooclass` inside of a `classsynopsisinfo`.

Then once that's being called we also need the class attr set to interface - which changes the render from calling it a `class` to use `interface`. Similarly, when that attr is set this way the `oointerface` that's being extended/implemented will adjust for the correct verb.

Girgias:
Okay, I'm not sure I actually really like this new expected format for interfaces, it is very confusing.

Can't we do something like:

```xml
   <classsynopsis class="interface">
    <oointerface>
     <interfacename>Throwable</interfacename>
    </oointerface>

    <classsynopsisinfo>
     <oointerface>
      <interfacename>Throwable</interfacename>
     </oointerface>

     <oointerface>
      <modifier>extends</modifier>
      <interfacename>Stringable</interfacename>
     </oointerface>

     <oointerface>
      <modifier>extends</modifier>
      <interfacename>SecondInterface</interfacename>
     </oointerface>
    </classsynopsisinfo>

    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.throwable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[@role='Throwable'])">
     <xi:fallback/>
    </xi:include>

    <classsynopsisinfo role="comment">&InheritedMethods;</classsynopsisinfo>
    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.stringable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[@role='Stringable'])">
     <xi:fallback/>
    </xi:include>
   </classsynopsis>
```

And where PhD collapses multiple ``oointerface`` tags that have an ``extends`` modifier into one list? As far as I can see this should be valid according to the DocBooks DTD

kocsismate:
> do you know why the JSON and /language/ files were not updated by your previous stub sync? Is this an issue in our stubs or just that you forgot?

Hmm, before committing my changes, I had to filter out a few files (which have unrelated diffs), and probably I missed `JsonSerializable`. But I'm not sure about why interfaces in the language directory didn't show up among the changes. That may be a bug in gen_stub.php 🤔

language/predefined/backedenum.xml
```diff
@@ -27,9 +27,7 @@
     </ooclass>
 
     <classsynopsisinfo>
-     <oointerface>
-      <interfacename>BackedEnum</interfacename>
-     </oointerface>
+     <ooclass><classname>BackedEnum</classname></ooclass>
```
kocsismate, line 30: Please do not do commit this change on its own since it's contrary to the output of gen_stub.php (the script which generates these markups). Please rather fix the script itself.

Girgias:
Right, @kocsismate I think the current way we render the class synopsis is kinda bad, see #2563 (and the related PhD PR) to see what the new markup would look like.

language/predefined/backedenum.xml
mallardduck: @kocsismate Are there any docs or wiki pages somewhere detailing how to work with the `gen_stub.php` script? I have ran it before in the past on `php-src` but am not sure how I would run that script for the extensions?

mallardduck:
Going to close this per the new direction.

PR #2562, opened 2023-07-06
Randomizer::nextInt return value and 32-bit note

KapitanOczywisty:
> However, if ~~`Randomizer::getInt()` with no arguments~~ `Randomizer::nextInt()` is executed in a 32-bit environment using an Engine that generates values above 32-bit, a RuntimeException will be thrown to avoid incompatibility.
> https://wiki.php.net/rfc/rng_extension#final_class_random_randomizer

Also added return value description

### DIFF
--- a/reference/random/random/randomizer/nextint.xml
+++ b/reference/random/random/randomizer/nextint.xml
@@ -27,7 +27,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-    Positive integer starting from 0, with maximum value depending on number of bytes
+    Positive integer starting from 0, with maximum value depending on the number of bytes
     returned from <methodname>Random\Engine::generate</methodname>.
     The exact value can be calculated as <code>2**($engine_bits - 1) - 1</code>.
   </para>
@@ -38,11 +38,11 @@
   <itemizedlist>
    <listitem>
     <simpara>
-      To avoid inconsistency, 32 bit PHP will throw <classname>Random\RandomException</classname>
+      To avoid inconsistencies, 32 bit PHP will throw <classname>Random\RandomException</classname>
       if the output size of <methodname>Random\Engine::generate</methodname> exceeds 32 bits,
       as the random value cannot be returned losslessly.
       This affects 64 bit engines: <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
-      <classname>Random\Engine\Xoshiro256StarStar</classname>, as well as any userland engines
+      <classname>Random\Engine\Xoshiro256StarStar</classname>, as well for any userland engines
       returning more than 4 bytes of randomness.
     </simpara>
    </listitem>
--- a/reference/random/random/randomizer/nextint.xml
+++ b/reference/random/random/randomizer/nextint.xml
@@ -27,16 +27,27 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-
+    Positive integer starting from 0, with maximum value depending on number of bytes
+    returned from <methodname>Random\Engine::generate</methodname>.
+    The exact value can be calculated as <code>2**($engine_bits - 1) - 1</code>.
   </para>
+ </refsect1>
 
-  <caution>
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <itemizedlist>
+   <listitem>
     <simpara>
-      On 32-bit PHP compilation, this function will throw <classname>Random\RandomException</classname>
-      when used with 64-bit engines: <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
-      <classname>Random\Engine\Xoshiro256StarStar</classname>.
+      To avoid inconsistency, 32 bit PHP will throw <classname>Random\RandomException</classname>
+      if the output size of <methodname>Random\Engine::generate</methodname> exceeds 32 bits,
+      as the random value cannot be returned losslessly.
+      This affects 64 bit engines: <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
+      <classname>Random\Engine\Xoshiro256StarStar</classname>, as well as any userland engines
+      returning more than 4 bytes of randomness.
     </simpara>
-  </caution>
+   </listitem>
+   &random.engineErrors;
+  </itemizedlist>
  </refsect1>
 
  <refsect1 role="examples">
--- a/reference/random/random/randomizer/nextint.xml
+++ b/reference/random/random/randomizer/nextint.xml
@@ -27,9 +27,9 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-    Positive integer starting from 0, with maximum value depending on the number of bytes
+    A positive integer between 0 and a maximum value depending on the number of bytes
     returned from <methodname>Random\Engine::generate</methodname>.
-    The exact value can be calculated as <code>2**($engine_bits - 1) - 1</code>.
+    The exact maximum can be calculated as 2<superscript>$engine_bytes * 8 - 1</superscript> - 1.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/random/random/randomizer/nextint.xml
```diff
@@ -27,8 +27,16 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-
+    Positive integer between 0 and <constant>PHP_INT_MAX</constant>, inclusive.
```
TimWolla, line 30: This is not quite correct. The exact range depends on the underlying engine. The output range will be between `0` and `2**($engine_bits - 1) - 1`. A userland engine that returns a single byte will return a value between `0` and `127`.

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,8 +27,16 @@
...
   </para>
+
+  <caution>
+    <simpara>
+      On 32-bit PHP compilation, this function will throw <classname>Random\RandomException</classname>
+      when used with 64-bit engines: <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
+      <classname>Random\Engine\Xoshiro256StarStar</classname>.
```
TimWolla, line 37: Similarly to the other note, this isn't quite correct. It will throw for any engine larger than 32 bits. This also includes userland engines that return a string with length 5 or more.

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,8 +27,16 @@
...
+    </simpara>
+  </caution>
```
TimWolla, line 39: The `<caution>` is not entirely inappropriate for this type of failure case, but we should probably also have a `<refsect1 role="errors">` for consistency.

reference/random/random/randomizer/nextint.xml
TimWolla: Perhaps something along these lines:

```
32 bit PHP will throw a <classname>Random\RandomException</classname> if the
output size of the used <classname>Random\Engine</classname> exceeds 32 bits,
as the returned random value cannot be losslessly returned. This affects the
64 bit <classname>Random\Engine\PcgOneseq128XslRr64</classname> and <classname>Random\Engine\Xoshiro256StarStar</classname> engines and might
also affect userland engines that return more than 4 bytes of randomness.
```

TimWolla (commented):
Thank you for the PR. As you can see from my remarks … it's complicated. That's the main reason why I've not yet written documentation for the `nextInt()` method. Generally speaking the usefulness of `Randomizer::nextInt()` is very small, because the output cannot be succinctly described [1] and “some positive integer” is pretty unspecific. It primarily exists for drop-in compatibility with a parameter-less `mt_rand()`. Most users should probably either use `getInt()` or `getBytes()` (or `getBytesFromString()` as of PHP 8.3).

[1] The technical details is that the method takes the output of the engine, interprets it as an unsigned little endian integer and right shifts it by 1 bit to get rid of the what-would-be-the-sign-bit.

reference/random/random/randomizer/nextint.xml
Girgias: Yes, an Error refsect1 is more appropriate here

KapitanOczywisty:
With a bit more understanding of random extension, I've updated PR. Thanks for your feedback!

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,10 +27,29 @@
...
+    The exact value can be calculated as <code>2**($engine_bits - 1) - 1</code>.
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <itemizedlist>
+   <listitem>
+    <simpara>
+      To avoid inconsistency, 32 bit PHP will throw <classname>Random\RandomException</classname>
```
Girgias, line 41:
```suggestion
      To avoid inconsistencies, 32 bit PHP will throw <classname>Random\RandomException</classname>
```

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,10 +27,29 @@
...
+      if the output size of <methodname>Random\Engine::generate</methodname> exceeds 32 bits,
+      as the random value cannot be returned losslessly.
+      This affects 64 bit engines: <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
+      <classname>Random\Engine\Xoshiro256StarStar</classname>, as well as any userland engines
+      returning more than 4 bytes of randomness.
```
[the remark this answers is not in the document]
KapitanOczywisty: Changed, though I think this should be "as well as for" - not sure.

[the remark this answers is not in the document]
Girgias: Yeah actually that might be better... @Crell need a wordsmith expert here

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,10 +27,29 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-
+    Positive integer starting from 0, with maximum value depending on the number of bytes
+    returned from <methodname>Random\Engine::generate</methodname>.
+    The exact value can be calculated as <code>2**($engine_bits - 1) - 1</code>.
```
TimWolla, line 32: This could possibly be phrased in terms of the engine bytes, because byte lengths are much more natural with the string return value:
```suggestion
    The exact maximum can be calculated as <code>2**($engine_bytes * 8 - 1) - 1</code>.
```

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,10 +27,29 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-
+    Positive integer starting from 0, with maximum value depending on the number of bytes
```
TimWolla, line 30:
```suggestion
    A positive integer between 0 and a maximum value depending on the number of bytes
```

reference/random/random/randomizer/nextint.xml
```diff
@@ -27,10 +27,29 @@
...
+  &reftitle.errors;
+  <itemizedlist>
+   <listitem>
+    <simpara>
+      To avoid inconsistencies, 32 bit PHP will throw <classname>Random\RandomException</classname>
+      if the output size of <methodname>Random\Engine::generate</methodname> exceeds 32 bits,
+      as the random value cannot be returned losslessly.
+      This affects 64 bit engines: <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
+      <classname>Random\Engine\Xoshiro256StarStar</classname>, as well for any userland engines
+      returning more than 4 bytes of randomness.
```
TimWolla, line 46:
```suggestion
      To avoid inconsistencies, 32 bit PHP will throw <classname>Random\RandomException</classname>
      if the output size of <methodname>Random\Engine::generate</methodname> exceeds 32 bits,
      as the selected integer cannot be returned losslessly.
      This affects the native 64 bit engines <classname>Random\Engine\PcgOneseq128XslRr64</classname> and
      <classname>Random\Engine\Xoshiro256StarStar</classname>. Any userland engine
      returning more than 4 bytes of randomness is also affected.
```

TimWolla (commented):
Thank you. It's technically accurate now. I've made some phrasing suggestions that hopefully matches the style used in the remainder of the ext/random a little better.

reference/random/random/randomizer/nextint.xml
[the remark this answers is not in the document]
Crell: I think Tim's proposed text (below) works better, despite the use of passive voice. 😄

reference/random/random/randomizer/nextint.xml
TimWolla: I just remembered, it's actually possible to render this "correctly":
```suggestion
    The exact maximum can be calculated as 2<superscript>$engine_bytes * 8 - 1</superscript> - 1.
```
see: https://github.com/php/doc-en/blob/38ce6211e1da2f62833f8961511f3c980ea9093b/reference/random/random/engine/xoshiro256starstar/jump.xml#L24C4-L24C35

TimWolla:
Thanks!

KapitanOczywisty:
@TimWolla thanks for applying changes, I've got busy lately..

PR #2564, opened 2023-07-07
Remove false return hint on `curl_share_errno`

joshuaruesweg:
As stated in the changelog, this method no longer returns `false` on error.


TimWolla:
Thanks!

PR #2568, opened 2023-07-09
Corrections to get_called_class

nielsdos:


reference/classobj/functions/get-called-class.xml
```diff
@@ -24,7 +24,41 @@
...
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        Calling this function from outside a class, will
+        trigger an <classname>Error</classname>. Previously,
+        an <constant>E_WARNING</constant> was raised
+        and the function returned &false;.
```
Girgias, line 56:
```suggestion
        Calling this function from outside a class, will
        now throw an <classname>Error</classname>. Previously,
        an <constant>E_WARNING</constant> was raised
        and the function returned &false;.
```

reference/classobj/functions/get-called-class.xml
```diff
@@ -24,7 +24,41 @@
...
   <para>
-   Returns the class name.  Returns &false; if called from outside a class.
+   Returns the class name.
+  </para>
+ </refsect1>
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   If <function>get_called_class</function> is called from outside a class,
+   <classname>Error</classname> is raised. Prior to PHP 8.0.0,
```
Girgias, line 34:
```suggestion
   If <function>get_called_class</function> is called from outside a class,
   an <classname>Error</classname> is thrown. Prior to PHP 8.0.0,
```

nielsdos:
@Girgias FWIW, I took some of the wording from the get_class page, so maybe that page should also get your wording updates.

Girgias:
> @Girgias FWIW, I took some of the wording from the get_class page, so maybe that page should also get your wording updates.

Right, doc wording consistency already not great, will try to remember to sync them.

Girgias:
Done in fe33a5685ef58fd545cfacc1e19ba6411379feab

PR #2569, opened 2023-07-10
Update bind-result.xml

niemeier23:
Corrected simple grammatical error.

"This" is not referring to multiple "functions," but only this one function.


Girgias:
Thank you!

PR #2570, opened 2023-07-10
Documented ZipArchive::open return values per file-upload.xml example

mmcev106:
It'd be nice to easily look up what a `ZipArchive::open` return value means.  I mirrored the first example I found of something similar ([Handling file uploads -> Error Messages Explained](https://www.php.net/manual/en/features.file-upload.errors.php)).  If you'd prefer this be done a different way, I'm happy to accommodate.

### DIFF
--- a/reference/zip/ziparchive/open.xml
+++ b/reference/zip/ziparchive/open.xml
@@ -77,55 +77,55 @@
     <varlistentry>
      <term><constant>ZipArchive::ER_EXISTS</constant></term>
      <listitem>
-      <simpara>File already exists.</simpara>
+      <simpara>Value: 10; File already exists.</simpara>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><constant>ZipArchive::ER_INCONS</constant></term>
      <listitem>
-      <simpara>Zip archive inconsistent.</simpara>
+      <simpara>Value: 21; Zip archive inconsistent.</simpara>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><constant>ZipArchive::ER_INVAL</constant></term>
      <listitem>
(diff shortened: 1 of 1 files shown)


Girgias:
We recently got rid of displaying the value of constants, as most of the time we don't actually know them as they are defined by the underlying library. And it seems to be the same here.

So not really sure how to proceed here.

mmcev106:
@Girgias, are you saying the values are different on different systems?

Girgias:
> @Girgias, are you saying the values are different on different systems?

Yes, they could, as they are dependent on the library, which could change the value between versions.

mmcev106:
@Girgias, I understand the concern coming from lack of control of a third party library.  However, realistically speaking, it sure seems like whomever manages those libraries would most likely never change these values because they would also be concerned about the impact.

I suspect these values are the same on all systems today, for the foreseeable future, and perhaps eternity.  Why not document them for quick clarity on cases like the following?  We always remove them in the unlikely event they do become inaccurate down the road.
```
if($zip->open($path) !== true){
    throw new Exception('Zip open failed with following status: ' . $status . '. I wish the PHP docs could tell me what that means.');
}
```

mmcev106:
@Girgias, I can understand all that.  It still unfortunately leaves no elegant way of handling the example in [this comment](https://github.com/php/doc-en/pull/2570#issuecomment-1632742624).  Perhaps this problem would be better solved more generally by creating a `get_contant_name()` function that could be used like so:

```php
$result = $zip->open($path)
if($result !== true){
    throw new Exception('Zip open failed with following error: ' . get_constant_name(ZipArchive::class, $result));
}
```

I'm going to assume that this is just not priority and close this PR.  If/when there is interest in improving scenarios like this, I'd be willing to throw time at it.

PR #2576, opened 2023-07-12
Update elseif.xml

sidneylopes:
Adjusting text to be more simple and understandable.

### DIFF
--- a/language/control-structures/elseif.xml
+++ b/language/control-structures/elseif.xml
@@ -37,10 +37,10 @@ if ($a > $b) {
   There may be several <literal>elseif</literal>s within the same
   <literal>if</literal> statement.  The first
   <literal>elseif</literal> expression (if any) that evaluates to
-  &true; would be executed.  In PHP, it's possible to write 
+  &true; would be executed.  In PHP, it's possible to write
   <literal>else if</literal> (in two words) and the behavior would be identical
   to the one of <literal>elseif</literal> (in a single word).  The syntactic meaning
-  is slightly different (the same behavior as C) but the bottom line 
+  is slightly different (the same behavior as C) but the bottom line
   is that both would result in exactly the same behavior.
  </simpara>
  <simpara>
@@ -55,10 +55,10 @@ if ($a > $b) {
   <simpara>
    Note that <literal>elseif</literal> and <literal>else if</literal>
    will only be considered exactly the same when using curly brackets
-   as in the above example.  When using a colon to define 
-   <literal>if</literal>/<literal>elseif</literal> conditions, the use 
-   of <literal>elseif</literal> in a single word becomes necessary. PHP 
-   will fail with a parse error if <literal>else if</literal> 
+   as in the above example.  When using a colon to define
+   <literal>if</literal>/<literal>elseif</literal> conditions, the use
+   of <literal>elseif</literal> in a single word becomes necessary. PHP
+   will fail with a parse error if <literal>else if</literal>
    is split into two words.
   </simpara>
  </note>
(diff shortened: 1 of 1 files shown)


language/control-structures/elseif.xml
```diff
@@ -58,8 +58,8 @@ if ($a > $b) {
    will only be considered exactly the same when using curly brackets
    as in the above example.  When using a colon to define your
    <literal>if</literal>/<literal>elseif</literal> conditions, you must
-   not separate <literal>else if</literal> into two words, or PHP will
-   fail with a parse error.
+   use <literal>elseif</literal> in a single word. PHP will fail with a
+   parse error if you split <literal>else if</literal> into two words.
```
Girgias, line 62: While at it, could you remove the usage of ``you`` in the above line? As the manual should not use personalisation.

sidneylopes: Done!

language/control-structures/elseif.xml
```diff
@@ -37,12 +37,11 @@ if ($a > $b) {
   There may be several <literal>elseif</literal>s within the same
   <literal>if</literal> statement.  The first
   <literal>elseif</literal> expression (if any) that evaluates to
-  &true; would be executed.  In PHP, you can also
-  write 'else if' (in two words) and the behavior would be identical
+  &true; would be executed.  In PHP, It's possible to write 
+  'else if' (in two words) and the behavior would be identical
```
Girgias, line 41:
```suggestion
  &true; would be executed.  In PHP, it's possible to write 
  <literal>else if</literal> (in two words) and the behavior would be identical
```

language/control-structures/elseif.xml
```diff
@@ -37,12 +37,11 @@ if ($a > $b) {
...
   to the one of 'elseif' (in a single word).  The syntactic meaning
```
Girgias, line 42:
```suggestion
  to the one of <literal>elseif</literal> (in a single word).  The syntactic meaning
```

Girgias (approved):
Might as well fix the text to use proper markup

language/control-structures/elseif.xml
```diff
@@ -37,12 +37,11 @@ if ($a > $b) {
...
-  write 'else if' (in two words) and the behavior would be identical
-  to the one of 'elseif' (in a single word).  The syntactic meaning
-  is slightly different (if you're familiar with C, this is the same
-  behavior) but the bottom line is that both would result in exactly
-  the same behavior.
+  &true; would be executed.  In PHP, it's possible to write 
+  <literal>else if</literal> (in two words) and the behavior would be identical
+  to the one of <literal>elseif</literal> (in a single word).  The syntactic meaning
+  is slightly different (the same behavior as C) but the bottom line 
+  is that both would result in exactly the same behavior.
```
Girgias, line 44:
```suggestion
  &true; would be executed.  In PHP, it's possible to write
  <literal>else if</literal> (in two words) and the behavior would be identical
  to the one of <literal>elseif</literal> (in a single word).  The syntactic meaning
  is slightly different (the same behavior as C) but the bottom line
  is that both would result in exactly the same behavior.
```

language/control-structures/elseif.xml
```diff
@@ -56,10 +55,11 @@ if ($a > $b) {
...
    will only be considered exactly the same when using curly brackets
-   as in the above example.  When using a colon to define your
-   <literal>if</literal>/<literal>elseif</literal> conditions, you must
-   not separate <literal>else if</literal> into two words, or PHP will
-   fail with a parse error.
+   as in the above example.  When using a colon to define 
+   <literal>if</literal>/<literal>elseif</literal> conditions, the use 
+   of <literal>elseif</literal> in a single word becomes necessary. PHP 
+   will fail with a parse error if <literal>else if</literal> 
+   is split into two words.
```
Girgias, line 62:
```suggestion
   as in the above example.  When using a colon to define
   <literal>if</literal>/<literal>elseif</literal> conditions, the use
   of <literal>elseif</literal> in a single word becomes necessary. PHP
   will fail with a parse error if <literal>else if</literal>
   is split into two words.
```

Girgias:
Thank you!

PR #2578, opened 2023-07-15
Remove superfluous white spaces

joshuaruesweg:


Girgias:
2 spaces after a period/full stop is also correct, and this is done across the manual.
So until we have a proper style guide and CI enforcement, I don't see the point in addressing this.

See: https://www.grammarly.com/blog/spaces-after-period/

PR #2581, opened 2023-07-17
Add note on how PHP/GNU layout affects manpath

nisbet-hubbard:
Ref: https://manpages.ubuntu.com/manpages/jammy/en/man5/manpath.5.html#search%20path


appendices/configure/misc.xml
```diff
@@ -20,7 +20,7 @@
     <listitem>
      <para>
       Sets how installed files will be laid out. Type is one of PHP (default) 
-or GNU.
+or GNU. Note that if installing manpages under PREFIX (default), you need to choose the GNU style in order for them to be found in the search path of &manpath; utility.
```
Girgias, line 23: Line break and the manual must not use personalization (i.e. usage of ``you``).

Girgias (changes requested):

appendices/configure/misc.xml
Girgias: Yeah, some part of the manual use ``you`` quite a bit because if was written at a time the rule wasn't enforced. So if you want to rewrite those please feel free the help would be appreciated!

appendices/configure/misc.xml
```diff
@@ -20,7 +20,8 @@
     <listitem>
      <para>
       Sets how installed files will be laid out. Type is one of PHP (default) 
-or GNU.
+      or GNU. Note that if installing manpages under PREFIX (default), choose the
+      GNU style in order for them to be found in the search path of &manpath; utility.
```
Girgias, line 24:
```suggestion
      or GNU. Note that if installing manpages under PREFIX (default), choose the
      GNU style in order for them to be found in the search path of &manpath; utility.
```

appendices/configure/misc.xml
```diff
@@ -20,7 +20,8 @@
...
```
Girgias, line 24: We don't actually define a ``&manpath;`` entity. What would you expect it to be written there?

Girgias (approved):

Girgias:
Thank you!

PR #2591, opened 2023-07-21
curl_multi_init: fix return description

voku:
ref.: https://github.com/php/php-src/blob/master/ext/curl/multi.c#L60 (stub: `function curl_multi_init(): CurlMultiHandle {}`)

⇒ "Stubs are the authoriative source of truth. The descriptions in the manual are sometimes buggy due to them not being syncronized with the stubs via automatic tooling (unlike signatures)." - Máté Kocsis (https://twitter.com/kocsismate90/status/1681912594024525824)

### DIFF
--- a/reference/curl/functions/curl-multi-init.xml
+++ b/reference/curl/functions/curl-multi-init.xml
@@ -25,7 +25,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns a cURL multi handle object.
+   Returns a cURL multi handle.
   </para>
  </refsect1>
  
--- a/reference/curl/functions/curl-multi-init.xml
+++ b/reference/curl/functions/curl-multi-init.xml
@@ -25,7 +25,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns a cURL multi handle on success, &false; on failure.
+   Returns a cURL multi handle object.
   </para>
  </refsect1>
  


TimWolla (commented):
This should probably also adjust `curl_init`. With regard to consistency `curl_share_init` does not ist the word "object".

kocsismate (approved):

Girgias (commented):
This feels like it is missing en Error/Exception section as I imagine it now throws an error on failure? Or did that just never occur?

voku:
@TimWolla `curl_init` still returns false: https://github.com/php/php-src/blob/master/ext/curl/interface.c#L1200

Girgias:
Okay so the implementation never fails, but we don't check if it has been properly initialized looking at the cURL docs: https://curl.se/libcurl/c/curl_multi_init.html

TimWolla:
> `curl_init` still returns false: https://github.com/php/php-src/blob/master/ext/curl/interface.c#L1200

Fun. I fully expected it to throw and thus didn't check.

TimWolla:
Merging, as this is a improvement as it is. Thanks!

voku:
> Okay so the implementation never fails, but we don't check if it has been properly initialized looking at the cURL docs: https://curl.se/libcurl/c/curl_multi_init.html

Return value

If this function returns NULL, something went wrong and you cannot use the other curl functions.

---

=> Sounds like it doesn't always return a new object. 🤔

divinity76:
wouldn't exactly call this issue resolved, because now it is /undocumented/ what curl_multi_init() does on failure

divinity76:
related PR https://github.com/php/php-src/pull/13157

Girgias:
> related PR [php/php-src#13157](https://github.com/php/php-src/pull/13157)
> 
> fwiw prior to that PR (php <8.4.0-dev?) it returns a broken curl_multi handle on failure

Could you open a new issue about this, or add an entry to UPGRADING in the php-src PR so this does not get forgotten.

PR #2594, opened 2023-07-23
Document ReflectionProperty::setValue PHP 8.3 deprecation

kylekatarnls:
Document the consequence of [`11703`](https://github.com/php/php-src/pull/11703) for `ReflectionProperty::setValue`

### DIFF
--- a/reference/reflection/reflectionproperty/setvalue.xml
+++ b/reference/reflection/reflectionproperty/setvalue.xml
@@ -21,6 +21,11 @@
   <para>
    Sets (changes) the property's value.
   </para>
+  <note>
+   <simpara>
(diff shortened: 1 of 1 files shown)


Girgias (changes requested):
This requires a changelog entry, and usually we have a predefined XML entity for features that are deprecated that needs to be added

Girgias:
Ping @kylekatarnls please address review comments

kylekatarnls:
Changelog added.

> we have a predefined XML entity for features that are deprecated that needs to be added

I used the tags that I found in many places for similar changes as of a certain PHP version (in `final`, `constant` etc.) if you have another template in mind can you please give me a page that is using the last standard (with regard that the method is not fully deprecated only calling it with 1 parameter is)?

Girgias (approved):
LGTM, thank you!

Issue #2513, opened 2023-06-05
json_decode return values documentation


NanoCaiordo:
From manual page: https://php.net/function.json-decode

---
The Return Values section instead of 

> Values true, false and null are returned as true, false and null respectively.

Should be written as

> JSON encoded values true, false and null are returned as true, false and null respectively.

Girgias:
Hello,

Could you provide a PR for the wording you would like?

NanoCaiordo:
<p class="para">
   Returns the value from an encoded <code class="parameter">json</code> in the appropriate
   PHP type. Encoded values <code class="literal">true</code>, <code class="literal">false</code> and
   <code class="literal">null</code> are returned as <strong><code>true</code></strong>, <strong><code>false</code></strong> and <strong><code>null</code></strong>
   respectively. <strong><code>null</code></strong> is returned if the <code class="parameter">json</code> cannot
   be decoded or if the encoded data is deeper than the nesting limit.
  </p>

PR #2606, opened 2023-07-27
Reworded json_decode return value paragraph to be more explicit about `true`, `false` and `null`

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2513

### DIFF
--- a/reference/json/functions/json-decode.xml
+++ b/reference/json/functions/json-decode.xml
@@ -80,11 +80,12 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns the value encoded in <parameter>json</parameter> in appropriate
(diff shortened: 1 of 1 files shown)


reference/json/functions/json-decode.xml
```diff
@@ -81,10 +81,11 @@
   &reftitle.returnvalues;
   <para>
    Returns the value encoded in <parameter>json</parameter> in appropriate
```
Girgias, line 83:
```suggestion
   Returns the value encoded in <parameter>json</parameter> as an appropriate
```
Might read better

alexandre-daubois: Definitely, updated! :ok_hand:

Girgias:
Thank you!

Issue #2406, opened 2023-04-06
Add notice that output of spl_object_hash() should only be compared type safe


dev-maniac:
With pull request https://github.com/php/php-src/pull/7010 (PHP 8.1, I think) the result format of spl_object_hash() changed.
The hash is now much more simple and can be something like `0000000000000e600000000000000000`, which PHP will interpret as number `0` (exponent).
We stumbled across this, because we used the object hash to check, if we already processed an object. We stored the hash in an array and checked with a simple `in_array()`, if array contains hash. But with PHP 8.2 this suddenly returned true for some objects which lead to some "interesting" side effects...

So I think it would be worthwhile to add a notice in documentation, that the hash should always be compared type-safe.

PR #2607, opened 2023-07-27
Add a notice to `spl_object_hash()` return values comparison

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2406

### DIFF
--- a/reference/spl/functions/spl-object-hash.xml
+++ b/reference/spl/functions/spl-object-hash.xml
@@ -70,6 +70,14 @@ $storage[$id] = $object;
     When an object is destroyed, its hash may be reused for other objects.
    </para>
   </note>
+  <note>
+   <para>
+    Object hashes should be compared for identity with <code>===</code> and
+    <code>!==</code>, because the returned hash could be a
+    <link linkend="language.types.numeric-strings">numeric string</link>.
+    For example: <literal>0000000000000e600000000000000000</literal>.
+   </para>
+  </note>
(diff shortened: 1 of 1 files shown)


reference/spl/functions/spl-object-hash.xml
```diff
@@ -70,6 +70,15 @@ $storage[$id] = $object;
     When an object is destroyed, its hash may be reused for other objects.
    </para>
   </note>
+  <note>
+   <para>
+    Object hashes should always be strictly compared with <code>===</code> and
+    <code>!==</code>, because returned hashes could be a valid integer
+    representation like <code>0000000000000e600000000000000000</code> (which is
+    a valid integer representation using exponent notation). Not doing so may
+    lead to confusing and false comparisons.
```
Girgias, line 79: Improve wording to use the name of the ``===`` operator, but also I don't know if saying this can lead to confusing comparison needs more detail if we mention (and link) to numeric strings.

```suggestion
    Object hashes should be compared for identity with <code>===</code> and
    <code>!==</code>, because the returned hash could be a
    <link linkend="language.types.numeric-strings">numeric string</link>.
    For example: <literal>0000000000000e600000000000000000</literal>.
```

alexandre-daubois: That's definitely better. It's updated, thanks for the suggestion!

Girgias:
Thank you!

PR #2609, opened 2023-07-27
OPcache: clarify validate_timestamps considering other options

pereorga:
The changes aim to clarify that disabling `opcache.validate_timestamps` doesn't mean timestamps will never be checked.

In scenarios where `opcache.file_update_protection` is non-zero (it is `2` by default) or `opcache.max_file_size` is set to a non-default value, OPcache may still validate file timestamps. This behavior was not clearly indicated in the original documentation and could lead to potential confusion for users (e.g., https://github.com/amnuts/opcache-gui/issues/92#issuecomment-1261958279)

I have confirmed this is the behavior here: https://github.com/php/php-src/blob/master/ext/opcache/ZendAccelerator.c#L1772C2-L1772C2

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -482,6 +482,10 @@
       manually via <function>opcache_reset</function>,
       <function>opcache_invalidate</function> or by restarting the Web server
       for changes to the filesystem to take effect.
+      Note that OPcache may still validate the timestamp of a file if
+      <link linkend="ini.opcache.file_update_protection">opcache.file_update_protection</link>
+      or <link linkend="ini.opcache.max-file-size">opcache.max_file_size</link>
+      options are set to non-zero values.
      </para>
     </listitem>
    </varlistentry>
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -482,10 +482,14 @@
       manually via <function>opcache_reset</function>,
       <function>opcache_invalidate</function> or by restarting the Web server
       for changes to the filesystem to take effect.
-      Note that OPcache may still validate the timestamp of a file if
-      <link linkend="ini.opcache.file_update_protection">opcache.file_update_protection</link>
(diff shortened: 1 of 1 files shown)


reference/opcache/ini.xml
```diff
@@ -482,6 +482,10 @@
       manually via <function>opcache_reset</function>,
       <function>opcache_invalidate</function> or by restarting the Web server
       for changes to the filesystem to take effect.
+      Note that OPcache may still validate the timestamp of a file if
+      <link linkend="ini.opcache.file_update_protection">opcache.file_update_protection</link>
+      or <link linkend="ini.opcache.max-file-size">opcache.max_file_size</link>
+      options are set to non-zero values.
```
Girgias, line 488: Using proper markup for the note so that it stands out.
```suggestion
      <note>
       <simpara>
        OPcache may still validate the timestamp of a file if
        <link linkend="ini.opcache.file_update_protection">opcache.file_update_protection</link>
        or <link linkend="ini.opcache.max-file-size">opcache.max_file_size</link>
        options are set to non-zero values.
       </simpara>
      </note>
```

Girgias:
Thank you!

pereorga:
@Girgias thanks!

After giving it more thought, I realize that my recent changes to the documentation might have introduced a potential ambiguity between two distinct processes: validating the file timestamp at compile time, and validating it each time the code is executed.

While the existing text correctly reflects the compile-time process, the added phrasing could be misinterpreted to mean that the `file_update_protection` and `max_file_size ` settings affect runtime validation of timestamps as well, which it does not.

I apologize for introducing the confusion :(

pereorga:
Should we say instead "OPcache may still validate the timestamp of a file at compile-time, if..."

I'm happy to create a PR to address this.

Girgias:
Please do, I'm not very familiar with the intricacies of OPcache

PR #2610, opened 2023-07-27
OPcache: explain why setting file_update_protection to 0 may improve performance

pereorga:
Related: https://github.com/php/doc-en/pull/2609

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -784,7 +784,9 @@
      <para>
       Prevents caching files that are less than this number of seconds old.
       It protects from caching of incompletely updated files. In case all file
-      updates on your site are atomic, you may increase performance by setting it to "0".
+      updates on your site are atomic, you may increase performance by setting
+      it to "0". This will allow files to be cached immediately, potentially
+      eliminating the need to validate timestamps.
      </para>
     </listitem>
    </varlistentry>
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -784,9 +784,8 @@
      <para>
       Prevents caching files that are less than this number of seconds old.
       It protects from caching of incompletely updated files. In case all file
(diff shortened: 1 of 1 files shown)


reference/opcache/ini.xml
```diff
@@ -784,7 +784,9 @@
      <para>
       Prevents caching files that are less than this number of seconds old.
       It protects from caching of incompletely updated files. In case all file
-      updates on your site are atomic, you may increase performance by setting it to "0".
+      updates on your site are atomic, you may increase performance by setting
+      it to "0". This will allow files to be cached immediately, potentially
+      eliminating the need to validate timestamps.
```
Girgias, line 789: While at it, let's remove personalization (usage of you) which the manual should not use.
```suggestion
      updates are atomic, performance can be increased by setting this to <literal>0</literal>.
      This will allow files to be cached immediately, potentially eliminating the need to validate timestamps.
```

Aside, the type of this INI setting looks to be ``int`` and not string.

reference/opcache/ini.xml
```diff
@@ -784,7 +784,8 @@
      <para>
       Prevents caching files that are less than this number of seconds old.
       It protects from caching of incompletely updated files. In case all file
-      updates on your site are atomic, you may increase performance by setting it to "0".
+      updates are atomic, performance can be increased by setting this to <literal>0</literal>.
+      This will allow files to be cached immediately, potentially eliminating the need to validate timestamps.
```
pereorga, line 788:
```suggestion
      This will allow files to be cached immediately.
```

As per https://github.com/php/doc-en/pull/2609#issuecomment-1655109927.

We could still mention that timestamp validation on compile time may not occur, but I'm not sure if it really matters (regarding performance).

Girgias (approved):

Girgias:
Thank you!

PR #2611, opened 2023-07-27
Use new class synopsis markup

kocsismate:
Following https://github.com/php/phd/pull/77

### DIFF
--- a/language/predefined/argumentcounterror.xml
+++ b/language/predefined/argumentcounterror.xml
@@ -20,21 +20,15 @@
    &reftitle.classsynopsis;
  
 <!-- {{{ Synopsis -->
-   <classsynopsis>
+   <classsynopsis class="class">
+    <ooexception>
+     <exceptionname>ArgumentCountError</exceptionname>
+    </ooexception>
+
     <ooclass>
-     <classname>ArgumentCountError</classname>
+     <modifier>extends</modifier>
+     <classname>TypeError</classname>
     </ooclass>
-
-    <classsynopsisinfo>
-     <ooclass>
-      <classname>ArgumentCountError</classname>
-     </ooclass>
-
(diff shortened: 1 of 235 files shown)


TimWolla (commented):
ack for reference/random/

Girgias (changes requested):
Can we not just fix the interfaces for the time being, and possibly continue to use the legacy markup for classes?

Not sure if I really want to dump a 260 file change onto translations, but maybe @mumumu, @saundefined, @sy-records are fine with that?

saundefined:
It's not a big problem for me, thanks!

But I'd also listen to other translators' opinions as well.

kocsismate:
> an we not just fix the interfaces for the time being, and possibly continue to use the legacy markup for classes?

I'm fine with fixing the interfaces first in a separate commit, but I'd be a bit upset if classes had to use the old syntax because then I would have to conditionally generate a different markup for classes and interfaces which would complicate gen_stub even more.

Girgias:
> > an we not just fix the interfaces for the time being, and possibly continue to use the legacy markup for classes?
> 
> I'm fine with fixing the interfaces first in a separate commit, but I'd be a bit upset if classes had to use the old syntax because then I would have to conditionally generate a different markup for classes and interfaces which would complicate gen_stub even more.

Yeah I would want them to be the same too, but let's first do interfaces to fix the immediate issues, and lets merge classes a tad later

kocsismate:
OK, I'll create a new PR soon only containing the interfaces :)

kocsismate:
@Girgias Can I merge this?

Girgias:
Give me/doc-fr until the end of the week to try to get our stuff in order :)

kocsismate:
@Girgias What about now? :)

Issue #2381, opened 2023-02-08
phpversion


janhanacek1:
### Description

Function phpversion with extension parameter.

By documentation.
Returns the current PHP version as a string. If a string argument is provided for extension parameter, phpversion() returns the version of that extension, or false if there is no version information associated or the extension isn't enabled.

The following code:

```php
<?php

echo phpversion( 'intl' ); // any extension returns php version instead

```

Resulted in this output:
```
8.2.2
```

But I expected this output instead:
```
// version of extension
```

### PHP Version

php 8.2

### Operating System

Debian 11

PR #2613, opened 2023-07-28
Add a note about PHP core extension versions in `phpversion()`

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2381

### DIFF
--- a/reference/info/functions/phpversion.xml
+++ b/reference/info/functions/phpversion.xml
@@ -136,6 +136,12 @@ if (PHP_VERSION_ID < 50207) {
     constants.
    </para>
   </note>
+  <note>
+   <para>
+    Some extensions may define their own version number.
+    However, most bundled extension will use the PHP version as their version number.
+   </para>
+  </note>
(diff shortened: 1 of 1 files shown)


reference/info/functions/phpversion.xml
```diff
@@ -136,6 +136,13 @@ if (PHP_VERSION_ID < 50207) {
     constants.
    </para>
   </note>
+  <note>
+   <para>
+    Even if some extensions are defining their own version number, most
+    core extensions will return the same version as the PHP version being
+    used.
```
Girgias, line 143: We don't use "core" as there is an extension that is name "core", extension that are part of php-src are called "bundled".
```suggestion
    Some extensions may define their own version number.
    However most bundled extension will use the PHP version as their version number.
```

alexandre-daubois: Thank you for the tip, that's updated! :+1:

Girgias (approved):

Issue #2373, opened 2023-03-21
Lack of class properties types example on the "Type declarations" page


PhoneixS:
I think that the page ["Type declarations"](https://www.php.net/manual/en/language.types.declarations.php) should have an example of how to write type hinting for class properties and static class properties.

Also, the "Changelog" should have, in my opinion, the 7.4 version which added this possibility of setting types for class properties.

An example could be something as simple as:

```lang=php
class User {

    public static string $my_static = 'foo';

    public int $id;
    public string $name;
 
    public function __construct(int $id, string $name) {
        $this->id = $id;
        $this->name = $name;
    }
}
```

PhoneixS:
I agree with you, probably when you are learning you search for type declarations of properties on the page about how to make properties in general. However, if you're checking Type Declarations is a must to have at least links to the place explaining how they are used if not a little example, at least it's my point of view.

PR #2614, opened 2023-07-28
Add example for class properties type declaration

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2373

### DIFF
--- a/language/types/declarations.xml
+++ b/language/types/declarations.xml
@@ -93,6 +93,12 @@
       </entry>
      </row>
      <row>
+      <entry>7.4.0</entry>
+      <entry>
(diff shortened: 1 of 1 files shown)


language/types/declarations.xml
```diff
@@ -603,6 +609,27 @@ function get_item(): ?string {
...
 }
 ?>
+]]>
+   </programlisting>
+  </example>
+
+  <example>
+   <title>Class property type declaration</title>
+   <programlisting role="php">
+    <![CDATA[
```
Girgias, line 619: Nit: we don't put indentation before a CDATA tag
```suggestion
<![CDATA[
```

Issue #2079, opened 2022-12-13
Enum type page should indicate availability only as of PHP 8.1.


staabm:
shouldn't https://www.php.net/manual/en/language.types.enumerations.php mention that enums are available since 8.1+?

cmb69:
Indeed, that should be mentioned.

PR #2615, opened 2023-07-28
Display version required by enumerations

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2079


Girgias:
I actually don't know if that would fix it, even if this looks sensible. Did you render this with PhD just to be sure?

Girgias:
Thank you!

Issue #2155, opened 2023-01-09
missing :


Miglandz:
From manual page: https://php.net/pdostatement.bindvalue

---

hello you have missing ":"
$sth->bindValue('calories', $calories, PDO::PARAM_INT);

should be 
$sth->bindValue(':calories', $calories, PDO::PARAM_INT);

PR #2616, opened 2023-07-28
Reword parameters value binding

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2155

It is actually already documented, but I tried a bit of rewording to be clearer


Girgias:
Thank you!

Issue #2102, opened 2022-12-21
`--with-password-argon2` no longer expects argument


crackerjam:
### Description

First, I build argon2 from [source](https://github.com/P-H-C/phc-winner-argon2), with a simple `make $PREFIX=$HOME/php-temp-argon2` and `make install $PREFIX=$HOME/php-temp-argon2`. This builds successfully and places the build products in $HOME/php-temp/argon2 as expected.

Then, on the php source, I run `./configure --prefix=$HOME/php --with-password-argon2=$HOME/php-temp/argon2`. This works successfully.

However, running `make` afterwards results this error message:

```/bin/bash /media/5cd8/crackerjam/php-temp/php-8.1.13/libtool --silent --preserve-dup-deps --tag CC --mode=compile cc -Iext/standard/ -I/media/5cd8/crackerjam/php-temp/php-8.1.13/ext/standard/ -I/media/5cd8/crackerjam/php-temp/php-8.1.13/include -I/media/5cd8/crackerjam/php-temp/php-8.1.13/main -I/media/5cd8/crackerjam/php-temp/php-8.1.13 -I/media/5cd8/crackerjam/php-temp/php-8.1.13/ext/date/lib -I/usr/include/libxml2 -I/media/5cd8/crackerjam/php-temp/php-8.1.13/TSRM -I/media/5cd8/crackerjam/php-temp/php-8.1.13/Zend    -fno-common -Wstrict-prototypes -Wlogical-op -Wduplicated-cond -Wno-clobbered -Wall -Wextra -Wno-strict-aliasing -Wno-unused-parameter -Wno-sign-compare -g -O2 -fvisibility=hidden -DZEND_SIGNALS   -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 -c /media/5cd8/crackerjam/php-temp/php-8.1.13/ext/standard/password.c -o ext/standard/password.lo  -MMD -MF ext/standard/password.dep -MT ext/standard/password.lo
/media/5cd8/crackerjam/php-temp/php-8.1.13/ext/standard/password.c: In function ‘php_password_argon2_hash’:
/media/5cd8/crackerjam/php-temp/php-8.1.13/ext/standard/password.c:342:16: error: too many arguments to function ‘argon2_encodedlen’
  encoded_len = argon2_encodedlen(
                ^~~~~~~~~~~~~~~~~
In file included from /media/5cd8/crackerjam/php-temp/php-8.1.13/ext/standard/password.c:31:0:
[...]

PR #2617, opened 2023-07-28
Remove configure option argument requirement as of PHP 8.1.0

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2102

### DIFF
--- a/reference/password/setup.xml
+++ b/reference/password/setup.xml
@@ -18,10 +18,14 @@
   &reftitle.install;
   &no.install;
   <para>
-   To enable Argon2 password hashing, however, PHP must be build with libargon2
-   support using the <option role="configure">--with-password-argon2[=DIR]</option>
+   To enable Argon2 password hashing, however, PHP must be built with libargon2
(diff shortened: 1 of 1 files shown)


reference/password/setup.xml
```diff
@@ -20,8 +20,10 @@
   <para>
    To enable Argon2 password hashing, however, PHP must be build with libargon2
    support using the <option role="configure">--with-password-argon2[=DIR]</option>
-   configure option.
-  </para> 
+   configure option. As of PHP 8.1.0, the
+   <option role="configure">--with-password-argon2</option> option doesn't take
+   any argument.
+  </para>
```
Girgias, line 26: I would reword this to make it future-proof by having the current way be the "main" and leading paragraph and then have something like:

> Prior to PHP 8.1.0, it was possible to specifiy the argon2 directory with ``command``

alexandre-daubois: I updated the PR with your suggestion :slightly_smiling_face:

Girgias:
Thank you

PR #2619, opened 2023-07-28
OPcache: clarify timestamps can be checked at compile-time

pereorga:
This is to clarify that `file_update_protection` and `max_file_size` settings affect validation of timestamps at compile-time only.

Context: https://github.com/php/doc-en/pull/2609#issuecomment-1655109927


Girgias (approved):

Girgias:
Thank you!

PR #2624, opened 2023-07-29
Fix confusion around statically calling DOMDocument load methods

### LINKED ISSUE #2443: loadHTML method of DomDocument
The documentation at https://www.php.net/manual/en/domdocument.loadhtml says that the loadHTML method of DomDocument may be called statically. This is no longer the case as of PHP8. Could this be clarified in the documentation?

nielsdos:
I'm guessing we can land commit 1, but have to wait until 8.3's release to land commit 2?

Fixes #2443.


Girgias (commented):
I think all those functions need a changelog entry to say that prior to PHP 8 if called statically it would return a DOMDocument

Girgias (commented):
Actually no, the signature can land now already, as it requires the changelog for PHP 7, and it "effectively" always returned bool in PHP 8, even if the stubs were incorrect.

I probably should start writting the mass PHP 8.3 migration guide issue so that we can track PRs related to that

nielsdos:
> I think all those functions need a changelog entry to say that prior to PHP 8 if called statically it would return a DOMDocument

There's already a mention of static calling in the errors/exception section. Do we then move that to the changelog too?

Girgias:
> > I think all those functions need a changelog entry to say that prior to PHP 8 if called statically it would return a DOMDocument
> 
> There's already a mention of static calling in the errors/exception section. Do we then move that to the changelog too?

Yeah that errors/exception section should have added a changelog entry, maybe that was missed as the XML entity already existed? But now I'm wondering if it was the case for the whole of PHP 7 or if it was deprecated in on of the PHP 7.(1+x) versions >_>

Girgias:
> > Yeah that errors/exception section should have added a changelog entry, maybe that was missed as the XML entity already existed? But now I'm wondering if it was the case for the whole of PHP 7 or if it was deprecated in on of the PHP 7.(1+x) versions >_>
> 
> So the wording was changed in: [de9c65c](https://github.com/php/doc-en/commit/de9c65c91ff1710d8b2d2ec955caea0162679305) i.e. this threw an E_STRICT error in the PHP 5 era. When https://wiki.php.net/rfc/reclassify_e_strict#calling_non-static_methods_statically was applied to 7.0, this became an E_DEPRECATED error.
> 
> I suppose we remove that allowstatic entity, and add the changelog entry for 8.0 when it became impossible to call it statically?

Yeah makes sense to me :)

nielsdos:
Ok, done that now.

nielsdos:
... on the wrong branch AAAAAAAAAAAA

PR #2625, opened 2023-07-30
Update operator precedence table

### LINKED ISSUE #2622: Throw missing from operator precedence page
Since PHP 8.0, `throw`can be used as an expression, but the `throw` keyword is not listed in the operator precedence table.

Refs:
* https://www.php.net/manual/en/language.operators.precedence.php
* https://wiki.php.net/rfc/throw_expression

jrfnl:
Updates the table to match the information per the parser.

This adds entries for:
* `=>`
* `include`, `include_once`, `require`, `require_once`
* `fn`
* ~~`throw`~~

~~Ref: https://github.com/php/php-src/blob/2e2416825d2270d256f7597904e14e386cc0d854/Zend/zend_language_parser.y#L54-L82~~
Ref: https://github.com/php/php-src/blob/82a15338e298142f52854b64d803695d4e5252df/Zend/zend_language_parser.y#L53-L82 (Sept 2026)

~~Fixes~~ ++Related to++ #2622

### DIFF
--- a/language/operators/precedence.xml
+++ b/language/operators/precedence.xml
@@ -258,6 +258,15 @@
      </row>
      <row>
       <entry>(n/a)</entry>
+      <entry><literal>=&gt;</literal></entry>
+      <entry>
+       <link linkend="language.types.array">array</link>,
+       <link linkend="control-structures.match">match</link>&listendand;
+       <link linkend="functions.arrow">fn</link>
+      </entry>
+     </row>
+     <row>
+      <entry>(n/a)</entry>
       <entry><literal>yield</literal></entry>
       <entry>
        <link linkend="control-structures.yield">yield</link>
@@ -288,6 +297,26 @@
       <entry>
        <link linkend="language.operators.logical">logical</link>
       </entry>
+     </row>
+     <row>
+      <entry>(n/a)</entry>
+      <entry>
+       <literal>include</literal>
+       <literal>include_once</literal>
+       <literal>require</literal>
+       <literal>require_once</literal>
+      </entry>
+      <entry>
+       <function>include</function>,
+       <function>include_once</function>,
+       <function>require</function>&listendand;
+       <function>require_once</function>
+      </entry>
+     </row>
+     <row>
+      <entry>(n/a)</entry>
+      <entry><literal>fn</literal></entry>
+      <entry><link linkend="functions.arrow">fn</link></entry>
      </row>
      <row>
       <entry>(n/a)</entry>
--- a/language/operators/precedence.xml
+++ b/language/operators/precedence.xml
@@ -260,9 +260,7 @@
       <entry>(n/a)</entry>
       <entry><literal>=&gt;</literal></entry>
       <entry>
-       <link linkend="language.types.array">array</link>,
-       <link linkend="control-structures.match">match</link>&listendand;
-       <link linkend="functions.arrow">fn</link>
+       <link linkend="language.generators.syntax">yield with keys</link>
       </entry>
      </row>
      <row>
--- a/language/operators/precedence.xml
+++ b/language/operators/precedence.xml
@@ -313,7 +313,7 @@
      </row>
      <row>
       <entry>(n/a)</entry>
-      <entry><literal>=&gt;</literal></entry>
+      <entry><literal>fn (...) =&gt;</literal></entry>
       <entry><link linkend="functions.arrow">fn arrow</link> disambiguity</entry>
      </row>
      <row>
--- a/language/operators/precedence.xml
+++ b/language/operators/precedence.xml
@@ -313,8 +313,8 @@
      </row>
      <row>
       <entry>(n/a)</entry>
-      <entry><literal>fn</literal></entry>
-      <entry><link linkend="functions.arrow">fn</link></entry>
+      <entry><literal>=&gt;</literal></entry>
+      <entry><link linkend="functions.arrow">fn arrow</link> disambiguity</entry>
      </row>
      <row>
       <entry>(n/a)</entry>


language/operators.xml
```diff
@@ -279,6 +279,15 @@
          <link linkend="control-structures.yield.from">yield from</link>
         </entry>
        </row>
+       <row>
+        <entry>(n/a)</entry>
+        <entry><literal>=&gt;</literal></entry>
+        <entry>
+         <link linkend="language.types.array">array</link>,
+         <link linkend="control-structures.match">match</link>&listendand;
+         <link linkend="functions.arrow">fn</link>
```
iluuu1994, line 288: I think this precedence is there exclusively to disambiguate `[yield $expr => $expr]`. All other uses of `=>` are unambiguous due to braces.

language/operators.xml
```diff
@@ -312,6 +321,33 @@
...
+        <entry>
+         <function>include</function>,
+         <function>include_once</function>,
+         <function>require</function>&listendand;
+         <function>require_once</function>
+        </entry>
+       </row>
+       <row>
+        <entry>(n/a)</entry>
+        <entry><literal>fn</literal></entry>
```
iluuu1994, line 341: `PREC_ARROW_FUNCTION` is used for the backup flags, rather than the global expression precedence. I think this can be dropped.

iluuu1994: Maybe @bwoebi can verify, I don't know exactly how `%prec` works in Bison. Maybe it's there to disambiguate `fn() => (1 + 2)` vs `(fn() => 1) + 2`

bwoebi: PREC_ARROW_FUNCTION is actually defining the real precedence, but it's not about `+`, but about the logical operators. If it weren't for `%precedence PREC_ARROW_FUNCTION`, `fn() => 1 or 2` would be `(fn() => 1) or 2`. Basically overriding the precedence of the `T_DOUBLE_ARROW`.

language/operators.xml
bwoebi: That's correct.

language/operators.xml
bwoebi: Also, the precedence does not really apply to `fn`, but to the T_DOUBLE_ARROW used in a `fn()` expression.

jrfnl: Hmm.. how should I change this in that case ? Cause that sounds like it should be a double arrow, but then we'd end up with the double arrow twice in the table, once for `yield` and once for `fn`. Did I understand that correctly ? And is that what you'd like me to change it to ?

language/operators.xml
jrfnl: I've updated the link block to only mention `yield` and linked to the language page for that.

jrfnl:
Anything I can do to move this forward ?

language/operators.xml
iluuu1994: Sorry for the late response. I suppose this case is somewhat special, in that the precedence is only applied in the arrow function context. Maybe we can make that clear by using `fn (...) =>` as the operator, or we just leave this out as an implementation detail.

lacatoire:
@jrfnl Sorry for the delay. The precedence table has moved since this pull request was opened: `language/operators.xml` was split by #2799, and the table now lives in `language/operators/precedence.xml`. That is what the conflict comes from.

jrfnl:
> @jrfnl Sorry for the delay. The precedence table has moved since this pull request was opened: `language/operators.xml` was split by #2799, and the table now lives in `language/operators/precedence.xml`. That is what the conflict comes from.

Thanks for the pointer, but that didn't answer my question, which I asked well before a conflict ever got created for this PR.

Either way, I'll rebase the PR while keeping the current page split in mind to get rid of the conflict.
I'll also re-compare against the precedence table just in case something new was introduced and not documented yet.

jrfnl:
Oh and the underlying issue (`throw` missing from the precedence table) this PR was trying to solve has been solved by another PR #5083 since.... though this PR still has value for the other things it does....

language/operators/precedence.xml
```diff
@@ -289,6 +296,26 @@
...
+       <function>include</function>,
+       <function>include_once</function>,
+       <function>require</function>&listendand;
+       <function>require_once</function>
+      </entry>
+     </row>
+     <row>
+      <entry>(n/a)</entry>
+      <entry><literal>=&gt;</literal></entry>
+      <entry><link linkend="functions.arrow">fn arrow</link> disambiguity</entry>
```
lacatoire: `PREC_ARROW_FUNCTION` is attached to the empty `backup_fn_flags` rule, which sits *after* `expr` in the arrow function rule, so what it governs is where the body stops, not how `=>` binds:
At `fn() => 1 • or 2` the parser chooses between reducing that empty rule and shifting `or`; since `PREC_ARROW_FUNCTION` is below `T_LOGICAL_OR` it shifts. That is a different thing from the `=>` of `yield`, whose precedence is carried by the token itself in [`L1383`](https://github.com/php/php-src/blob/82a15338e298142f52854b64d803695d4e5252df/Zend/zend_language_parser.y#L1383) (`T_YIELD expr T_DOUBLE_ARROW expr`) and declared at [`L61`](https://github.com/php/php-src/blob/82a15338e298142f52854b64d803695d4e5252df/Zend/zend_language_parser.y#L61).

Putting `=>` in the operator column for both rows therefore reads as two contradictory precedences for one token, which is what makes the entry look odd. The effect is still worth a row, since it is the one case the table cannot currently predict:

```php
$x = 1 or 2;            // $x is 1                    -> ($x = 1) or 2
yield 'k' => 1 or 2;    // yields 1                   -> (yield 'k' => 1) or 2
fn() => 1 or 2;         // Closure, body returns true -> fn() => (1 or 2)
```

```suggestion
      <entry><literal>fn</literal></entry>
      <entry>
       The body of an <link linkend="functions.arrow">arrow function</link>
       extends as far to the right as possible
      </entry>
```

jrfnl: @lacatoire This has been discussed before above: https://github.com/php/doc-en/pull/2625#discussion_r1279669176 - also your take reads as AI generated.

lacatoire: Got it, it's the T_DOUBLE_ARROW, and it would appear twice. Let's go with iluuu1994's first suggestion. Write the operator as `fn (...) =>` rather than a bare `=>`, so the two rows can't be confused. 
The behaviour is worth the row: fn() => 1 or 2 puts the or inside the body, which nothing else in the table lets you predict. I'll merge as soon as you've rebased.

jrfnl: Fixed.

jordikroon: Thank you for your work @jrfnl. Will get to the above privately with Louis-Arnaud. Let's stay human, is my strong suggestion.

jrfnl: > Let's stay human, is my strong suggestion.

I fully support that. AI use was pretty clear from the original (unedited) comment.

iluuu1994: A heads-up: Volker and I will propose an LLM policy (hopefully) soon. It will require disclosure for LLM-generated content on the mailing list and on GitHub (for PR descriptions, comments, issues, etc., extending to all of our repos). Comments authored fully autonomously by an agent will be prohibited.

I expect that to be a reasonable middle-ground and for the proposal to be accepted.

jrfnl: @iluuu1994 Well, I have opinions on that, but I don't have a vote, so I'll keep my opinions to myself (and ignore anything and everything the mansplaining machines spit out).

iluuu1994: I'd like to completely prohibit LLM generated content, at least in comments. But we're trying to make the first step that is likely to pass.

Feel free to share your thoughts though, happy to hear other opinions.

jrfnl: > I'd like to completely prohibit LLM generated content, at least in comments. But we're trying to make the first step that is likely to pass.

@iluuu1994 I'm afraid if you open the door, you will never be able to close it again.

To me, the only acceptable stance is **_"no GenAI"_**. And to make sure we understand each other, that doesn't forbid translation assistance for non-native speakers, but anything **_generated_**, be it a review, a comment, code or anything else, is unacceptable.

I can write essays on the why behind that stance, but here is a very short version of it:
* GenAI is unethical - based on theft, slavery & killing the planet.
* The fact that the legal side of GenAI code is unclear is an existential threat to open source.
* Reviewing anything GenAI created is a waste of my time:
    - The cognitive load of reading anything AI generated is much higher than anything human written.
    - It take 10 x as much time to get it to "merge-ready" compared to a human-created PR, it always contains bugs and often doesn't actually fix what it is supposed to fix and finding these bugs takes much more of my energy than when reviewing a human created PR.
    - Reviewing in OSS is part of "training the next generation of maintainers". When an AI is involved, then a) the human wasn't interested in learning about the codebase in the first place and b) nobody is being trained, as the AI will forget all that feedback and make the same mistakes again in the next session.
* I can't stand the mansplaining. I get enough of that already without _machines_ mansplaining to me as well.
* Even more so, I can't stand machines pretending they are human, i.e. using "I" in phrases etc. AI is a "give me the next reasonable phrase in a conversation machine", not a person who can take responsibility for their actions.

iluuu1994: I share many of those views. But we're not opening the door for LLMs, the door is already wide open. Without a policy, anything goes and we have no power to take any actions. We'll make it clear our policy does not preclude a more strict LLM policy in the future.

iluuu1994: Thanks for your input, I sent it to Volker and we'll make sure to add something like the above.

jrfnl: If it helps, I have some phrasing about no-AI in the [PHPCS `CONTRIBUTING` guide](https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/4.x/.github/CONTRIBUTING.md#final-words), though I know Volker is already aware of that.

PR #2626, opened 2023-07-30
Constructors and Destructors: add note to promoted properties

### LINKED ISSUE #2586: Constructor Property promotion applies parameter name rules to properties
### Description

The following code:

```php
<?php
class Foo {
    private array $_GET; // OK.
}

class Bar {
    public function __construct( private array $_GET ) {} // Same thing, but using CPP syntax sugar results in fatal error.
}
```

Resulted in this output:
```
Fatal error: Cannot re-assign auto-global variable _GET in /in/ZA5Bk on line 8
```

_No error, fatal or otherwise._

### Additional context

Since PHP 5.4, parameter names shadowing superglobals are forbidden. However, this restriction does not apply to property names.

When using the constructor property promotion, that restriction is, however, applied to properties too.

### PHP Version

PHP 8.0+

### Operating System

_No response_

jrfnl:
Point out naming restrictions for promoted properties.

Ref: https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion

Fixes #2586

### DIFF
--- a/language/oop5/decon.xml
+++ b/language/oop5/decon.xml
@@ -157,6 +157,12 @@ class Point {
     </note>
     <note>
      <para>
+      As promoted properties are desugared to both a property as well as a function parameter, any
+      and all naming restrictions for both properties as well as parameters apply.
+     </para>
+    </note>
+    <note>
+     <para>
       <link linkend="language.attributes">Attributes</link> placed on a
       promoted constructor argument will be replicated to both the property
       and argument.  Default values on a promoted constructor argument will be replicated only to the argument and not the property.


Girgias:
TIL, never thought about naming a parameter as one of the superglobals

jrfnl:
> TIL, never thought about naming a parameter as one of the superglobals

Edge case for sure, but probably a good thing to save someone else the head-scratching ;-)

Girgias:
> > TIL, never thought about naming a parameter as one of the superglobals
> 
> Edge case for sure, but probably a good thing to save someone else the head-scratching ;-)

Indeed! Thanks for the PR :D

Issue #1383, opened 2022-02-06
Missing documentation about GD requires zlib


mpepito13:
From manual page: https://php.net/zlib.installation

PHP 7.4.27 cannot be compiled without zlib being installed (see error line 10051 in the attachment)

Is it a documentation issue or Makefile issue ?

[config.log](https://github.com/php/doc-en/files/8011271/config.log)

---

cmb69:
https://www.php.net/manual/en/zlib.requirements.php mentions zlib as requirement. Isn't that sufficient?

mpepito13:
I don't understand it the same way as you.
https://www.php.net/manual/en/zlib.requirements.php mentions that a zlib version >= 1.2.0.4 is a requirement if you want to use Zlib, not that Zlib is a requirement to compile PHP.

By the way, I don't have the error when I compile PHP 8.1.

cmb69:
Hmm, zlib is not supposed to be enabled by default (neither for PHP 7.4 nor 8.1). What was the configure line which produced the config.log above?

cmb69:
Hmm, zlib is a dependency of ext/gd (what is not documented) and maybe other components. Please try to build without `--enable-gd` to confirm.

mpepito13:
Yes it works !
I need GD so I will install zlib on my own but thanks for your help

PR #2629, opened 2023-07-31
Add `zlib` requirement to GD

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/1383

Got the [version info here](https://github.com/php/php-src/blob/655f116be57b46efe32221d7adfec6d6b81eeece/ext/zlib/config0.m4#L4), hope hat was the good place to get it :slightly_smiling_face:


Girgias (approved):
Thank you!

Issue #2425, opened 2023-04-13
Prior to PHP 8.1 2nd and 3rd parameter of IntlDateFormatter::__construct / IntlDateFormatter::create are not optional


mz-aimcom:
From manual page: https://php.net/intldateformatter.create

---

In my tests parameters no. 2 `$dateType` and no. 3 `$timeType` are not optional in PHP before version 8.1.

Tested in PHP 8.0.28 and 7.4.33.

kocsismate:
Caused by https://github.com/php/php-src/commit/331eddadc8d1876e9dfe45c9784e75ef2930bb3e

PR #2633, opened 2023-07-31
[IntlDateFormatter] Add changelog entry about making `dateType` and `timeType` optional

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2425

### DIFF
--- a/reference/intl/dateformatter/create.xml
+++ b/reference/intl/dateformatter/create.xml
@@ -68,11 +68,10 @@
      <term><parameter>dateType</parameter></term>
      <listitem>
       <para>
-       Date type to use (<constant>none</constant>, <constant>short</constant>,
-       <constant>medium</constant>, <constant>long</constant>,
-       <constant>full</constant>).  This is one of the <link
-       linkend="intl.intldateformatter-constants">IntlDateFormatter
-       constants</link>.
+       Format of the date determined by one of the
+       <link linkend="intl.intldateformatter-constants">IntlDateFormatter
+       constants</link>. The default value is
+       <constant>IntlDateFormatter::FULL</constant>.
       </para>
      </listitem>
     </varlistentry>
@@ -80,11 +79,10 @@
      <term><parameter>timeType</parameter></term>
      <listitem>
       <para>
-       Time type to use (<constant>none</constant>, <constant>short</constant>,
-       <constant>medium</constant>, <constant>long</constant>,
-       <constant>full</constant>).  This is one of the <link
-       linkend="intl.intldateformatter-constants">IntlDateFormatter
-       constants</link>.
+       Format of the time determined by one of the
+       <link linkend="intl.intldateformatter-constants">IntlDateFormatter
+       constants</link>. The default value is
+       <constant>IntlDateFormatter::FULL</constant>.
       </para>
      </listitem>
     </varlistentry>
@@ -159,25 +157,11 @@
      </thead>
      <tbody>
       <row>
-       <entry>5.5.0/PECL 3.0.0</entry>
+       <entry>8.1.0</entry>
        <entry>
         <para>
-         An <classname>IntlCalendar</classname> object is allowed for
-         <parameter>calendar</parameter>.
-        </para>
-        <para>
-         Objects of type <classname>IntlTimeZone</classname> and
-         <classname>DateTimeZone</classname> are allowed for
-         <parameter>timezone</parameter>.
-        </para>
-        <para>
-         Invalid timezone identifiers (including empty strings) are no longer
-         allowed for <parameter>timezone</parameter>.
-        </para>
-        <para>
-         If &null; is given for <parameter>timezone</parameter>, the timezone
-         identifier given by <function>date_default_timezone_get</function> will
-         be used instead of ICUʼs default.
+         Parameters <parameter>dateType</parameter> and
+         <parameter>timeType</parameter> are now optional.
         </para>
        </entry>
       </row>


reference/intl/dateformatter/create.xml
```diff
@@ -158,6 +158,15 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.1.0</entry>
+       <entry>
+        <para>
+         Parameters <parameter>dateType</parameter> and
+         <parameter>timeType</parameter> are made optional.
```
Girgias, line 166:
```suggestion
         Parameters <parameter>dateType</parameter> and
         <parameter>timeType</parameter> are now optional.
```

reference/intl/dateformatter/create.xml
```diff
@@ -70,9 +70,9 @@
       <para>
        Date type to use (<constant>none</constant>, <constant>short</constant>,
        <constant>medium</constant>, <constant>long</constant>,
-       <constant>full</constant>).  This is one of the <link
+       <constant>full</constant>). This is one of the <link
        linkend="intl.intldateformatter-constants">IntlDateFormatter
-       constants</link>.
+       constants</link>. The default value is <constant>full</constant>.
```
Girgias, line 75: Those constants need fixing, as they should actually refer to the class too e.g. ``<constant>IntlDateFormatter:FULL</constant>`` as when I first saw this I was thinking there was a mistake in the usage of the ``constant`` tags instead of ``literal``...

alexandre-daubois: Updated! Does it make sense to only mention those 5, and not the `IntlDateFormatter::RELATIVE_*` ones?

reference/intl/dateformatter/create.xml
```diff
@@ -158,6 +164,15 @@
...
+       <entry>
+        <para>
+         Parameters <parameter>dateType</parameter> and
+         <parameter>timeType</parameter> are now optional.
+        </para>
+       </entry>
+      </row>
       <row>
        <entry>5.5.0/PECL 3.0.0</entry>
        <entry>
```
Girgias, line 178: Actually while you are here the 5.5.0 entry can now be removed as we only document PHP 7.0 upwards

reference/intl/dateformatter/create.xml
Girgias: I don't know this extension, but if those are valid options I would say yes.

For the ``*`` you can do something like ``<constant>IntlDateFormatter::RELATIVE_<replaceable>*</replaceable></constant>`` as we determined in the unmerged PR: #1963

alexandre-daubois: That's updated as well with `replaceable` :+1:

reference/intl/dateformatter/create.xml
alexandre-daubois: Would it be a good idea to prepare a PR removing changelog entries older than PHP 7.0? I'll take care of it if yes :+1:

Girgias: So currently the manual _must_ document PHP 7 and 8, and _should_ remove information relating to PHP 5.

This is more QA work and I don't know how much I want QA work to happen now as this will impact translations working on catching up with the bug fixes and migration guide/infos that need to be added.

alexandre-daubois: Alright, got it

reference/intl/dateformatter/create.xml
```diff
@@ -68,23 +68,31 @@
...
-       Date type to use (<constant>none</constant>, <constant>short</constant>,
-       <constant>medium</constant>, <constant>long</constant>,
-       <constant>full</constant>).  This is one of the <link
+       Date type to use (<constant>IntlDateFormatter::NONE</constant>,
+       <constant>IntlDateFormatter::SHORT</constant>,
+       <constant>IntlDateFormatter::MEDIUM</constant>,
+       <constant>IntlDateFormatter::LONG</constant>,
+       <constant>IntlDateFormatter::FULL</constant> and
+       <constant>IntlDateFormatter::RELATIVE_<replaceable>*</replaceable></constant>
+       as of PHP 8.0.0). This is one of the <link
```
Girgias, line 77: Actually not sure how fan of the phrasing I am...

Maybe something more like:
> Format of the date determined by one of the IntlDateFormatter constants.
> By default IntlDateFormatter::FULL.

And just not show a list of formats, as that seems pretty brittle?

alexandre-daubois: Updated, will be more future-proof :+1:

reference/intl/dateformatter/create.xml
```diff
@@ -68,23 +68,21 @@
      <term><parameter>dateType</parameter></term>
      <listitem>
       <para>
-       Date type to use (<constant>none</constant>, <constant>short</constant>,
-       <constant>medium</constant>, <constant>long</constant>,
-       <constant>full</constant>).  This is one of the <link
+       Format of the date determined by one of the <link
        linkend="intl.intldateformatter-constants">IntlDateFormatter
```
Girgias, line 72:
```suggestion
       Format of the date determined by one of the
       <link linkend="intl.intldateformatter-constants">IntlDateFormatter
```
The stub generating file doesn't like link tag split up into multiple lines and that always causes weird diffs

alexandre-daubois: Well noted, updated :+1:

Girgias (approved):

PR #2636, opened 2023-08-02
Min and max now behave consistently when comparing int to string

### LINKED ISSUE #1891: max example error
https://github.com/php/doc-en/blob/master/reference/math/functions/max.xml#L100

```php
// The string 'hello' when compared to an int is treated as 0
// Since the two values are equal, the order they are provided determines the result
echo max(0, 'hello');     // 0
echo max('hello', 0);     // hello
```

actual output:

```bash
hello
hello
```

afilina:
Result in all versions: https://3v4l.org/WtQN6

Fixes #1891

### DIFF
--- a/reference/math/functions/max.xml
+++ b/reference/math/functions/max.xml
@@ -105,6 +105,13 @@
        previously, &false; was returned and an <constant>E_WARNING</constant> error was emitted.
       </entry>
      </row>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       In case of an int to string comparison, <function>max</function> no longer returns a
+       different value based on the argument order.
+      </entry>
+     </row>
     </tbody>
    </tgroup>
   </informaltable>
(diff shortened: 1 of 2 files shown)


reference/math/functions/max.xml
```diff
@@ -105,6 +105,13 @@
        previously, &false; was returned and an <constant>E_WARNING</constant> error was emitted.
       </entry>
      </row>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       In case of an int to string comparison, <function>max</function> no longer returns a
+       different value based on the argument order.
```
afilina, line 112:
```suggestion
       As string to number comparisons where changed, <function>max</function> no longer returns a
       different value based on the argument order in those cases.
```

reference/math/functions/min.xml
```diff
@@ -105,6 +105,13 @@
        previously, &false; was returned and an <constant>E_WARNING</constant> error was emitted.
       </entry>
      </row>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       In case of an int to string comparison, <function>min</function> no longer returns a
+       different value based on the argument order.
```
afilina, line 112:
```suggestion
       As string to number comparisons where changed, <function>min</function> no longer returns a
       different value based on the argument order in those cases.
```

reference/math/functions/max.xml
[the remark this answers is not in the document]
afilina: Done.

PR #2638, opened 2023-08-02
Create a docker container for a single-line setup and rendering

### LINKED ISSUE #2637: Dockerize for easier local setup of the docs
Reference: https://github.com/php/doc-en/pull/877#issuecomment-1364360439

afilina:
Use case: a casual contributor who wants to preview their changes before opening a PR.

- There is currently no way to provide custom args to phd/render.php, because then we'd be back to complicated instructions.
- I tried not to make any assumptions about the host system, so I don't use make.
- This image is tied to an older commit of phd, since a (very) recent commit causes a crash on `assert($this->cchunk["classsynopsis"]["legacy"] === true);`. I'll just assume that it's a regression, since it fails on 2 machines and 2 PHP versions. I don't want to go down that rabbit hole today.

Fixes #2637

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,15 +1,15 @@
 FROM php:8.2-cli
 
-RUN apt-get update ; \
+RUN apt-get update && \
     apt-get install -y git
 
 WORKDIR /var/www
 
-RUN git clone https://github.com/php/phd.git ;  \
-    git clone https://github.com/php/doc-base.git ; \
+RUN git clone https://github.com/php/phd.git &&  \
+    git clone https://github.com/php/doc-base.git && \
     git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5
 
 RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
 
-CMD php doc-base/configure.php --disable-segfault-error ; \
+CMD php doc-base/configure.php --disable-segfault-error && \
     php phd/render.php --docbook doc-base/.manual.xml --output=/var/www/en/output --package PHP --format xhtml
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -5,6 +5,9 @@ RUN apt-get update && \
 
 WORKDIR /var/www
 
+ADD https://api.github.com/repos/php/phd/git/refs/heads/master version-phd.json
+ADD https://api.github.com/repos/php/doc-base/git/refs/heads/master version-doc-base.json
+
 RUN git clone --depth 1 https://github.com/php/phd.git &&  \
     git clone --depth 1 https://github.com/php/doc-base.git
 
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update && \
 WORKDIR /var/www
 
 RUN git clone https://github.com/php/phd.git &&  \
-    git clone https://github.com/php/doc-base.git && \
+    git clone --depth 1 https://github.com/php/doc-base.git && \
     git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5
 
 RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,0 +1,15 @@
+FROM php:8.2-cli
+
+RUN apt-get update ; \
+    apt-get install -y git
+
+WORKDIR /var/www
+
+RUN git clone https://github.com/php/phd.git ;  \
+    git clone https://github.com/php/doc-base.git ; \
+    git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5
+
+RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
+
+CMD php doc-base/configure.php --disable-segfault-error ; \
+    php phd/render.php --docbook doc-base/.manual.xml --output=/var/www/en/output --package PHP --format xhtml
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -5,9 +5,8 @@ RUN apt-get update && \
 
 WORKDIR /var/www
 
-RUN git clone https://github.com/php/phd.git &&  \
-    git clone --depth 1 https://github.com/php/doc-base.git && \
-    git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5
+RUN git clone --depth 1 https://github.com/php/phd.git &&  \
+    git clone --depth 1 https://github.com/php/doc-base.git
 
 RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
 
--- a/README
+++ b/README
@@ -7,3 +7,8 @@ Please refer to the README file within the [doc-base repository](https://github.
 For information related to creating this setup, see:
 
   http://doc.php.net/tutorial/local-setup.php
+
+Running With Docker
+
+- Rebuild the documentation using `docker-compose up`
+- Open output/php-chunked-xhtml/ in your browser.
--- a/README
+++ b/README
@@ -10,6 +10,6 @@ For information related to creating this setup, see:
 
 Running With Docker
 
-- Install Docker (https://docs.docker.com/get-docker/
+- Install Docker (https://docs.docker.com/get-docker/)
 - Rebuild the documentation using `docker-compose up`
 - Open output/php-chunked-xhtml/ in your browser.
--- a/README
+++ b/README
@@ -10,6 +10,6 @@ For information related to creating this setup, see:
 
 Running With Docker
 
-- Install [Docker](https://docs.docker.com/get-docker/).
+- Install Docker (https://docs.docker.com/get-docker/
 - Rebuild the documentation using `docker-compose up`
 - Open output/php-chunked-xhtml/ in your browser.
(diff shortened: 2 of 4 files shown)


afilina:
@cmb69 Might be a solution to #877

.docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+WORKDIR /var/www
+
+RUN git clone https://github.com/php/phd.git ;  \
+    git clone https://github.com/php/doc-base.git ; \
+    git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5
+
+RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
+
+CMD php doc-base/configure.php --disable-segfault-error ; \
+    php phd/render.php --docbook doc-base/.manual.xml --output=/var/www/en/output --package PHP --format xhtml
```
afilina, line 15:
```suggestion
FROM php:8.2-cli

RUN apt-get update && \
    apt-get install -y git

WORKDIR /var/www

RUN git clone https://github.com/php/phd.git &&  \
    git clone https://github.com/php/doc-base.git && \
    git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5

RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini

CMD php doc-base/configure.php --disable-segfault-error && \
    php phd/render.php --docbook doc-base/.manual.xml --output=/var/www/en/output --package PHP --format xhtml
```

TimWolla (approved):
This looks absolutely reasonable for a starter.

.docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM php:8.2-cli
+
+RUN apt-get update && \
+    apt-get install -y git
+
+WORKDIR /var/www
+
+RUN git clone https://github.com/php/phd.git &&  \
+    git clone https://github.com/php/doc-base.git && \
```
Lewiscowles1986, line 9: could clone with ` --depth 1` for faster download?

.docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM php:8.2-cli
+
+RUN apt-get update && \
+    apt-get install -y git
+
+WORKDIR /var/www
+
+RUN git clone https://github.com/php/phd.git &&  \
```
Lewiscowles1986, line 8: Would be nice if someone could tag the checkout so you can clone the tag using `--depth 1` for same benefit as prior comment.

Lewiscowles1986:
https://github.com/php/phd/pull/77 seems to be what broke phpd with something explicitly asserting a true which is not true.

mallardduck:
Once https://github.com/php/doc-en/pull/2641 is merged it will fix docs-en builds.

README
```diff
@@ -7,3 +7,9 @@ Please refer to the README file within the [doc-base repository](https://github.
 For information related to creating this setup, see:
 
   http://doc.php.net/tutorial/local-setup.php
+
+Running With Docker
+
+- Install [Docker](https://docs.docker.com/get-docker/).
```
afilina, line 13:
```suggestion
- Install Docker (https://docs.docker.com/get-docker/
```

It's not a Markdown file...

afilina: Need more coffee this morning :)
Thanks.

README
```diff
@@ -7,3 +7,9 @@ Please refer to the README file within the [doc-base repository](https://github.
 For information related to creating this setup, see:
 
   http://doc.php.net/tutorial/local-setup.php
+
+Running With Docker
+
+- Install Docker (https://docs.docker.com/get-docker/
```
afilina, line 13:
```suggestion
- Install Docker (https://docs.docker.com/get-docker/)
```

.docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+    git clone https://github.com/php/doc-base.git && \
+    git -C phd checkout 5708881e233cd9979dd11bcbd9fd0bb8a714f9d5
```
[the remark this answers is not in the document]
afilina: I'll wait to see the outcome of https://github.com/php/doc-en/pull/2641 and just do a `--depth 1` instead. Thanks, @mallardduck!

Lewiscowles1986:
All builds for me

I used the following edits to test depth 1 and not having a commit.

```
FROM php:8.2-cli

RUN apt-get update && \
    apt-get install -y git

WORKDIR /var/www

ARG PHPD_SELECTOR="--branch=master"
ARG DOC_BASE_SELECTOR="--branch=master"

RUN git clone --depth 1 https://github.com/php/phd.git ${PHPD_SELECTOR} &&  \
    git clone --depth 1 https://github.com/php/doc-base.git ${DOC_BASE_SELECTOR}

RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini

CMD php doc-base/configure.php && \
    php phd/render.php -d doc-base/.manual.xml -P PHP -f xhtml -o /var/www/en/output/

```

The suggested edits have the whole thing build in under a minute (50 seconds), which is all clone depth. 

You should be free to iterate in future.

The ARGS are optional, but if someone wants to fill this out with chm support or PDF, why not right?

.docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM php:8.2-cli
+
+RUN apt-get update && \
+    apt-get install -y git
+
+WORKDIR /var/www
+
+RUN git clone --depth 1 https://github.com/php/phd.git &&  \
+    git clone --depth 1 https://github.com/php/doc-base.git
```
TimWolla, line 9: Looking at this again, this might be problematic, because Docker's build cache will not be busted when rebuilding, as the command didn't change. Perhaps the full repository should automatically be cloned when first running the container and afterwards the user is required to manually update the dependencies.

Lewiscowles1986: If dockers build cache won't change, then solving here won't solve a thing. I was running using `--build`; but I am curious to hear how having the entire repo will solve?

nielsdos: A trick I like to use to bust Docker's caching is adding something like
`ADD https://api.github.com/repos/php/phd/git/refs/heads/master version.json`
`ADD https://api.github.com/repos/php/doc-base/git/refs/heads/master version.json`
before the clones
If there were no new commits the cache will be used, otherwise it will reclone.

Lewiscowles1986: Neat trick, so it forces rebuild only when upstream changes?

I think we may be using docker for different things, but that is a cool trick and I guess it's the same as not wanting to pin to `latest` for a docker image. Not enough information to trigger a re-build, but by using a URL, you're able to circumvent that.

https://github.com/moby/buildkit/issues/3739 has some interesting writing on the subject.

Girgias (commented):
I know nothing about Docker, but this seems sensible if people who know better are OK with the docker file.
Although I might move this more to doc-base. Would it be possible to have it take an argument to be able to render translations? (just a lang thing that defaults to en)

afilina:
> Although I might move this more to doc-base. Would it be possible to have it take an argument to be able to render translations? (just a lang thing that defaults to en)

That means that people would need to also clone `doc-base` in a sensible location, bringing us back to multiple steps and complex instructions (how to organize folders, where to run the command, etc.) I'm hoping to start small here, get a few people try it for feedback, then move it to dockerhub or something so that it wouldn't need to be in _any_ repo. A bit like [`phpco`](https://github.com/fortrabbit/phpco-docker) works, solving the exact same setup problem for another project.

Girgias:
> > Although I might move this more to doc-base. Would it be possible to have it take an argument to be able to render translations? (just a lang thing that defaults to en)
> 
> That means that people would need to also clone `doc-base` in a sensible location, bringing us back to multiple steps and complex instructions (how to organize folders, where to run the command, etc.) I'm hoping to start small here, get a few people try it for feedback, then move it to dockerhub or something so that it wouldn't need to be in _any_ repo. A bit like [`phpco`](https://github.com/fortrabbit/phpco-docker) works, solving the exact same setup problem for another project.

Right... I was just thinking have people "just" clone doc-base and let the docker compose do the cloning for them. But then one can't actually edit the docs.

.docker/Dockerfile
Lewiscowles1986: @afilina are you open to including the two ADD lines to auto-break docker cache?

afilina: I have no idea what the best practice is. I'm fine with it if that gets the PR merged.

Lewiscowles1986: @TimWolla is this blocking? (can't see anything else that would, and I do think this is a nice benefit)

TimWolla: Yes, I think having a cache buster is necessary, because otherwise this is going to be confusing to the less-experienced end-user which is specifically the target group of this ready-made Dockerfile.

Please include the suggestion from nielsdos, but please use difference names for the `version.json`:

```
ADD https://api.github.com/repos/php/phd/git/refs/heads/master version-phd.json
ADD https://api.github.com/repos/php/doc-base/git/refs/heads/master version-doc-base.json
```

The rest LGTM.

.docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM php:8.2-cli
+
+RUN apt-get update && \
+    apt-get install -y git
+
+WORKDIR /var/www
+
```
Lewiscowles1986, line 7:
```suggestion

ADD https://api.github.com/repos/php/phd/git/refs/heads/master version-phd.json
ADD https://api.github.com/repos/php/doc-base/git/refs/heads/master version-doc-base.json

```

.docker/Dockerfile
Lewiscowles1986: added as a suggestion, which hopefully helps.

nielsdos (approved):
Looks legit to me

Lewiscowles1986:
🙌  very nice

PR #2563, opened 2023-07-06
Example of new markup for PhD

Girgias:
See https://github.com/php/phd/pull/77

PR #2641, opened 2023-08-03
Updates the DS interfaces to use new syntax

mallardduck:
per title, this fixes the current build issues.
this is accomplished by updating the Data Structures docs to match the examples from here: https://github.com/php/doc-en/pull/2563
once merged these interfaces will use the proper new syntax and fixes both the builds and the current render issues on php.net

### DIFF
--- a/reference/ds/ds.collection.xml
+++ b/reference/ds/ds.collection.xml
@@ -24,19 +24,18 @@
 
 <!-- {{{ Synopsis -->
    <classsynopsis class="interface">
-    <ooclass><classname>Ds\Collection</classname></ooclass>
+    <oointerface><interfacename>Ds\Collection</interfacename></oointerface>
 
-<!-- {{{ Class synopsis -->
-    <classsynopsisinfo>
-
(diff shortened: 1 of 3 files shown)


afilina:
Just to confirm, does this mean that the docs become buildable using phd's HEAD?
Reference: https://github.com/php/doc-en/pull/2638#discussion_r1283251195

mallardduck:
@afilina - yeah, this PR should allow all HEAD tags for PHD/en/src to play nice again. Sorry for the headache there on your PR, this is all kinda my fault when I created a bodge level (in hindsight) interface fix. And then @Girgias found a more proper solution which required some updates that took time - and by then I forgot to follow up with this needed PR.

Girgias (approved):
I still don't understand why I don't have the build failures :(

Lewiscowles1986:
😊  I had cloned @afilina branch... So my master was never changing to be ahead. I was internally screaming just now as it still wasn't working, but upon debugging my setup; my commit sha was not updating. `git remote -v` to the rescue. It now builds 🙌

Issue #2635, opened 2023-08-01
Examples in the Warning section are the same


nikolaynikolaevn:
From manual page: https://php.net/control-structures.alternative-syntax

---

I fail to see the difference between the two examples in the red Warning section.

Girgias:
Probably should be changed to:
```php
<?php switch ($foo): ?>

<?php case 1: ?>
...
<?php endswitch ?>
```

I expect this example to have some trailing whitespace initially that got removed at one point.

PR #2648, opened 2023-08-07
Fix switch alternative syntax

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2635

### DIFF
--- a/language/control-structures/alternative-syntax.xml
+++ b/language/control-structures/alternative-syntax.xml
@@ -67,9 +67,10 @@ endif;
    <programlisting role="php">
 <![CDATA[
 <?php switch ($foo): ?>
+
     <?php case 1: ?>
     ...
-<?php endswitch ?>
+<?php endswitch; ?>
 ]]>
    </programlisting>
   </informalexample>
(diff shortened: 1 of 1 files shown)


Girgias (approved):

language/control-structures/alternative-syntax.xml
```diff
@@ -67,9 +67,10 @@ endif;
    <programlisting role="php">
 <![CDATA[
 <?php switch ($foo): ?>
+
     <?php case 1: ?>
```
TimWolla, line 71: Isn't the actual problem that the rendered version dropped the indentation of this line?

Girgias: Maybe... that's something that should be flagged on PhD (and the joy of having no tests for it to catch regressions...) hopefully the new line break will "fix" this anyway

alexandre-daubois: I'd say yes, but this means that the doc generator (or related) _may_ be fixed (which could be way more complex than adding this blank line). I guess adding this blank line will help anyway :slightly_smiling_face:

TimWolla: It was a CSS issue, no `white-space: pre`. I also found the previous version a more "natural" mistake, because it can easily halten that someone wants to indent the next line, whereas adding an empty line is less likely.

Girgias: Right, if it's fixed then please feel free to revert.

Issue #2631, opened 2023-07-31
Example #1 array_diff_assoc() example [second arrays last value should have a key "1"]


DEarTh27:
From manual page: https://php.net/function.array-diff-assoc

---

In this example you see the "a" => "green" pair is present in both arrays and thus it is not in the output from the function. Unlike this, the pair 0 => "red" is in the output because in the second argument "red" has key which is 1.

```
<?php
$array1 = array("a" => "green", "b" => "brown", "c" => "blue", "red");
$array2 = array("a" => "green", "yellow", "red");
$result = array_diff_assoc($array1, $array2);
print_r($result);
?>
```

damianwadley:
Since your report is just a copy-and-paste of the example's description and (part of) its code, how about clarifying what you think should be changed?

DEarTh27:
`$array2 = array("a" => "green", "yellow", "red"); `

Should be

`$array2 = array("a" => "green", "yellow", "1" => "red"); `

As per the description.

salathe:
In `$array2`, the item with value `"red"` already had a key of `1` automatically assigned. 
What's the problem?

salathe:
Again, please explain the problem.

Before and after the related code change (#2634) for this example, both `$array2` arrays are identical.

What do you want to achieve? Does that require changing the example code?

DEarTh27:
Seems like a misunderstanding. May be we can add description to say that "In $array2, the item with value "red" already had a key of 1." "**That is automatically assigned**"

PR #2650, opened 2023-08-07
Clarify `array_diff_(u)assoc` examples

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2631

My last try to clarify this wasn't the good fix. I try here to clarify what happens with `array_diff_(u)assoc` and why `red` is present in the output.

### DIFF
--- a/reference/array/functions/array-diff-uassoc.xml
+++ b/reference/array/functions/array-diff-uassoc.xml
@@ -80,25 +80,24 @@
    <example>
     <title><function>array_diff_uassoc</function> example</title>
     <para>
-     The <literal>"a" =&gt; "green"</literal>
+     In this example the <literal>"a" =&gt; "green"</literal>
      pair is present in both arrays and thus it is not in the output from the
      function. Unlike this, the pair <literal>0 =&gt; "red"</literal>
-     is in the output because in the second argument <literal>"red"</literal>
-     has key which is <literal>1</literal>.
+     is in the output because the key of <literal>"red"</literal> is
+     automatically assigned to <literal>0</literal> in the first array,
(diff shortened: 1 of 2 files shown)


reference/array/functions/array-diff-uassoc.xml
```diff
@@ -80,11 +80,13 @@
    <example>
     <title><function>array_diff_uassoc</function> example</title>
     <para>
-     The <literal>"a" =&gt; "green"</literal>
+     In this example you see the <literal>"a" =&gt; "green"</literal>
```
Girgias, line 83:
```suggestion
     In this example the <literal>"a" =&gt; "green"</literal>
```
Would be better to rephrase the other wording to get rid of personalization.

Girgias (changes requested):

reference/array/functions/array-diff-uassoc.xml
alexandre-daubois: That's right! I also took the opportunity to remove "you" occurrences in these pages.

reference/array/functions/array-diff-uassoc.xml
```diff
@@ -94,11 +96,12 @@ function key_compare_func($a, $b)
     if ($a === $b) {
         return 0;
     }
-    return ($a > $b)? 1:-1;
+
+    return ($a > $b) ? 1 : -1;
```
Girgias, line 100: Could use the spaceship operator, as this exists since 7.0

alexandre-daubois: Here you go :slightly_smiling_face:

Girgias (approved):

Issue #2484, opened 2023-05-11
ArrayObject - undocumented AllowDynamicProperties?


ondrejmirtes:
### Description

The following code:

```php
<?php

$obj = new \ArrayObject([
  'abc' => true,
], \ArrayObject::ARRAY_AS_PROPS);

var_dump($obj->abc);
```

Resulted in this output:
```
bool(true)
```

But I expected this output instead:
```
Warning: Undefined property: ArrayObject::$abc
```

The [PHP RFC "Deprecate dynamic properties"](https://wiki.php.net/rfc/deprecate_dynamic_properties) says that only `stdClass`, classes with magic __get/__set methods, and classes marked as `#[AllowDynamicProperties]` are allowed to have dynamic properties.

But ArrayObject is none of those:

* https://www.php.net/manual/en/class.arrayobject.php
* https://github.com/php/php-src/blob/5820528c9c25012ad763576c7120414eb12a05f0/ext/spl/spl_array.stub.php#L5

So what are the rules about dynamic properties? Are there any other cases where they are allowed that I'm unaware of?

Please understand I'm asking from the point of a static analyser developer, to be able to describe these cases. I get it it's useful to have ArrayObject to allow them, but I need the rules to describe this behaviour.

Thank you.

### PHP Version

PHP 8.2

PR #2652, opened 2023-08-07
Add a note about accessing dynamic properties in `ArrayObject`

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2484


Girgias:
One user note links to https://stackoverflow.com/questions/14610307/spl-arrayobject-arrayobjectstd-prop-list/16619183#16619183 which explains this quite well.

@kocsismate I wonder if it makes sense to add "virtual" docs for ``__get()``/``__set()``/etc. to describe this behaviour? Maybe we even should actually add those methods on the relevant classes?

Issue #2357, opened 2023-03-16
Abstract classes can implement \Traversable as of 8.0


epixian:
Documentation for [Traversable](https://www.php.net/manual/en/class.traversable.php) is missing a change as of PHP 8.0 that allows [abstract classes to implement `\Traversable`](https://github.com/php/php-src/commit/e9ae581f024e06878b2b1991b7daed6318c811a7).  The extending class must implement Iterator or IteratorAggregate or it will cause a fatal error:

```php
abstract class AbstractTraversable implements Traversable {}

class NonAbstractTraversable extends AbstractTraversable {}

// Fatal error: Class NonAbstractTraversable must implement interface Traversable as part of either Iterator or IteratorAggregate in Unknown on line 0
```

```php
abstract class AbstractTraversable implements Traversable {}

class NonAbstractTraversable extends AbstractTraversable implements IteratorAggregate {
    public function getIterator() {
        yield "foo";
        yield "bar";
    }
}

foreach (new NonAbstractTraversable as $value) {
    echo $value, "\n";
}

// foo
// bar
```

PR #2661, opened 2023-08-08
Fix GH-2357: improve `Traversable` page

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2357

### DIFF
--- a/language/predefined/traversable.xml
+++ b/language/predefined/traversable.xml
@@ -14,29 +14,10 @@
     Interface to detect if a class is traversable using &foreach;.
    </para>
    <para>
-    Abstract base interface that cannot be implemented alone. Instead it must
+    Abstract base interface that cannot be implemented alone. Instead, it must
     be implemented by either <interfacename>IteratorAggregate</interfacename> or
     <interfacename>Iterator</interfacename>.
    </para>
-   <note>
-    <para>
-     Internal (built-in) classes that implement this interface can be used in
-     a &foreach; construct and do not need to implement
-     <interfacename>IteratorAggregate</interfacename> or
-     <interfacename>Iterator</interfacename>.
-    </para>
-   </note>
-   <note>
-    <para>
-     This is an internal engine interface which cannot be implemented in PHP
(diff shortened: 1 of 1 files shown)


language/predefined/traversable.xml
```diff
@@ -58,6 +39,54 @@
...
+     <interfacename>IteratorAggregate</interfacename> or
+     <interfacename>Iterator</interfacename>.
+    </para>
+   </note>
+   <note>
+    <para>
+     Prior to PHP 8.0.0, this internal engine interface couldn't be implemented
+     in PHP scripts. Either <interfacename>IteratorAggregate</interfacename>
+     or <interfacename>Iterator</interfacename> must be used instead.
+     When implementing an interface which extends Traversable, make sure to
```
Girgias, line 82: Yes this was already missing but might as well add it

```suggestion
     When implementing an interface which extends <interfacename>Traversable</interfacename>,
     make sure to
```

Girgias: But also is the ordering still important?

Girgias (approved):
Minor comments but LGTM

language/predefined/traversable.xml
alexandre-daubois: I tested and actually no, I updated it :slightly_smiling_face:

alexandre-daubois: Tried on 7.0, here is the result:

image

When switching order:

image

So I'm not exactly sure from when this rule applied

alexandre-daubois:
Yes it seems you're right. Sorry for the brittle check, I updated the XML :slightly_smiling_face: Set the changelog entry to 7.4.0. I'm not sure the ordering rule ever applied :thinking:

Girgias:
> Yes it seems you're right. Sorry for the brittle check, I updated the XML slightly_smiling_face Set the changelog entry to 7.4.0. I'm not sure the ordering rule ever applied thinking

The ordering rule did apply prior to PHP 7.4: https://3v4l.org/ecv2m but that is a different issue all together

Girgias (approved):

Issue #2209, opened 2023-01-17
stream_set_chunk_size() missing changelog entry for ValueError


othercorey:
Looks like the return false (and warning) were converted to a ValueError exception. I could only find part of the changes, but the exception is clearly there with RETURN_THROWS().

othercorey:
@cmb69 I don't know how else to ask you a question so I'm using this issue. Is there documentation on the return system in php so I can be more confident when reading these changes? 

Does a naked `return` always return null or can it return false depending on the stub? If something else a value error, does it always throw and never return even if there is a return statement after it?

PR #2663, opened 2023-08-08
Fix GH-2209: update return value and changelog of stream_set_chunk_size()

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/2209

### DIFF
--- a/reference/stream/functions/stream-set-chunk-size.xml
+++ b/reference/stream/functions/stream-set-chunk-size.xml
@@ -45,19 +45,43 @@
   <para>
    Returns the previous chunk size on success.
   </para>
-  <para>
-   Will return &false; if <parameter>size</parameter> is less than 1 or
-   greater than <constant>PHP_INT_MAX</constant>.
-  </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/stream/functions/stream-set-chunk-size.xml
```diff
@@ -46,19 +46,51 @@
    Returns the previous chunk size on success.
   </para>
   <para>
-   Will return &false; if <parameter>size</parameter> is less than 1 or
-   greater than <constant>PHP_INT_MAX</constant>.
+   Prior to PHP 8.0.0, the function returned &false; if
+   <parameter>size</parameter> is less than 1 or greater than
+   <constant>PHP_INT_MAX</constant>.
```
Girgias, line 51: I think this (and the same bit in the Error section) should be deleted and moved to the changelog section, with similar phrasing from other changelog entries

alexandre-daubois: Here you go :slightly_smiling_face:

Girgias (approved):

PR #2664, opened 2023-08-08
[skip-revcheck] Normalize entity usage on classsynopsisinfo.

alfsb:
These files are the last on English still using strings instead of entities on `<classsynopsisinfo>`.


Girgias (approved):
Thank you!

PR #2667, opened 2023-08-09
remove unnecessary operation in code example

aghArdeshir:
in this example, `$count--` at the end of the function helps with nothing. Its existence also does not harm anything. But because in the same file, in "Example # 4 Example demonstrating need for static variables" we were talking about how useless `$a++` is in that example, putting a useless `$count--` can be misleading in this example, because the reader may look for the philosophy/reason behind existence of this statement.

### DIFF
--- a/language/variables.xml
+++ b/language/variables.xml
@@ -455,7 +455,6 @@ function test()
     if ($count < 10) {
         test();
     }
-    $count--;
 }
 ?>
 ]]>


Girgias:
This statement _is_ needed to reset the value when doing a follow-up call to ``test()``, it would be better to add a coment, or create another without the statement and have it call ``test()`` twice. In any case showing the output of this code would be a good enhancement

aghArdeshir:
Thanks @Girgias. I agree with you and I believe your point is correct.

So I'll close this PR. Hopefully me or someone else in near/distant future improves the example in one of the ways you suggested.

PR #2673, opened 2023-08-10
datetime modify: add examples with other formats

alexislefebvre:
The current page only shows examples that use the relative formats: https://www.php.net/manual/en/datetime.modify.php

I added examples with other formats so that users can see that not only the relative format is supported.

See also:

- #2674

### DIFF
--- a/reference/datetime/datetime/modify.xml
+++ b/reference/datetime/datetime/modify.xml
@@ -100,6 +100,33 @@ echo $date->format('Y-m-d') . "\n";
 ]]>
    </screen>
   </example>
+  <example>
+   <title>All formats of Date and Time are supported</title>
+   <programlisting role="php">
+<![CDATA[
+<?php
+$date = new DateTime('2020-12-31');
+
+$date->modify('July 1st, 2023');
+echo $date->format('Y-m-d H:i') . "\n";
+
+$date->modify('Monday next week');
+echo $date->format('Y-m-d H:i') . "\n";
+
+$date->modify('17:30');
+echo $date->format('Y-m-d H:i') . "\n";
+?>
+]]>
+   </programlisting>
+   &example.outputs;
+   <screen>
+<![CDATA[
+2023-07-01 00:00
+2023-07-03 00:00
+2023-07-03 17:30
+]]>
+   </screen>
+  </example>
  </refsect1>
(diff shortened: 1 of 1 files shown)


reference/datetime/datetime/modify.xml
```diff
@@ -97,6 +97,29 @@ echo $date->format('Y-m-d') . "\n";
...
+   <title>All formats of Date and Time are supported</title>
+   <programlisting role="php">
+<![CDATA[
+<?php
+$date = new DateTime('2000-12-31');
+
+$date->modify('July 1st, 2008');
+echo $date->format('Y-m-d') . "\n";
+
+$date->modify('Monday next week');
```
derickr, line 113: `Monday next week` is a relative form though :-)

reference/datetime/datetime/modify.xml
```diff
@@ -97,6 +97,29 @@ echo $date->format('Y-m-d') . "\n";
...
 2001-03-03
+]]>
+   </screen>
+  </example>
+  <example>
+   <title>All formats of Date and Time are supported</title>
+   <programlisting role="php">
+<![CDATA[
+<?php
+$date = new DateTime('2000-12-31');
```
derickr, line 108: Can you use current dates in new examples so that the documentation doesn't look "old"?

reference/datetime/datetime/modify.xml
```diff
@@ -97,6 +97,29 @@ echo $date->format('Y-m-d') . "\n";
...
+
+$date->modify('July 1st, 2008');
+echo $date->format('Y-m-d') . "\n";
```
derickr, line 111: Maybe also add one that shows: 1. Current time, and after `modify("17:30")` 2. That the seconds and microseconds are reset?

reference/datetime/datetime/modify.xml
alexislefebvre: Yes but currently the doc only shows examples with `+X [unit]`.

reference/datetime/datetime/modify.xml
alexislefebvre: I used dates in our decade.

derickr:
Merged after a rebase.

PR #2684, opened 2023-08-15
Update DOM intro

nielsdos:
There's loadHTML(File) and saveHTML(File) to to load and save HTML documents. People use it for both XML and HTML.

### DIFF
--- a/reference/dom/book.xml
+++ b/reference/dom/book.xml
@@ -9,7 +9,7 @@
  <preface xml:id="intro.dom">
   &reftitle.intro;
   <para>
-   The DOM extension allows you to operate on XML documents through the DOM
+   The DOM extension allows you to operate on XML and HTML documents through the DOM
    API with PHP.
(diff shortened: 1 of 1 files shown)


reference/dom/book.xml
```diff
@@ -9,7 +9,7 @@
  <preface xml:id="intro.dom">
   &reftitle.intro;
   <para>
-   The DOM extension allows you to operate on XML documents through the DOM
+   The DOM extension allows you to operate on XML and HTML documents through the DOM
```
Girgias, line 12:
```suggestion
   The DOM extension allows operations on XML and HTML documents through the DOM
```
Possibly one way to get rid of personalisation?

Girgias (approved):

reference/dom/book.xml
nielsdos: OK for me

Issue #1606, opened 2022-05-31
class_implements/parents should note that instanceof and is_a are available


damianwadley:
The top-voted comment on the `class_implements` page suggests how to use the function to check whether an object/class name implements a particular interface.
https://www.php.net/manual/en/function.class-implements.php#55376

The docs should that point out `instanceof` (for objects) and `is_a` (for either) as the appropriate ways of checking if something implements an interface.

For completeness, `class_parents` should get something similar.
https://www.php.net/manual/en/function.class-parents.php

Additionally, consider adding an editor's note to that comment.

PR #2697, opened 2023-08-21
Add mention of `is_a()` and `instanceof` in `class_implements()` and `class_parents()` pages

alexandre-daubois:
Fix https://github.com/php/doc-en/issues/1606

### DIFF
--- a/reference/spl/functions/class-implements.xml
+++ b/reference/spl/functions/class-implements.xml
@@ -99,12 +99,24 @@ Array
   </para>
  </refsect1>
 
+ <refsect1 role="notes">
+  &reftitle.notes;
+  <note>
+   <simpara>
+    To check that an object implements an interface, &instanceof;
+    or the <function>is_a</function> function should be used instead.
+   </simpara>
+  </note>
+ </refsect1>
+
  <refsect1 role="seealso">
   &reftitle.seealso;
   <para>
    <simplelist>
     <member><function>class_parents</function></member>
     <member><function>get_declared_interfaces</function></member>
+    <member><function>is_a</function></member>
+    <member>&instanceof;</member>
    </simplelist>
   </para>
  </refsect1>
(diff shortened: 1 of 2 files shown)


reference/spl/functions/class-implements.xml
```diff
@@ -99,12 +99,24 @@ Array
   </para>
  </refsect1>
 
+ <refsect1 role="notes">
+  &reftitle.notes;
+  &note.no-remote;
```
Girgias, line 104: Why this entity?
Doesn't seem related at all and something from a copy/paste.

reference/spl/functions/class-implements.xml
```diff
@@ -99,12 +99,24 @@ Array
...
+  <note>
+   <simpara>
+    To check that an object implements an interface, you should either use &instanceof;
+    or the <function>is_a</function> function.
```
Girgias, line 108: Personalisation:
```suggestion
    To check that an object implements an interface, &instanceof;
    or the <function>is_a</function> function should be used instead.
```

Girgias (changes requested):

reference/spl/functions/class-implements.xml
```diff
@@ -99,12 +99,24 @@ Array
...
+  </note>
+ </refsect1>
+
  <refsect1 role="seealso">
   &reftitle.seealso;
   <para>
    <simplelist>
     <member><function>class_parents</function></member>
     <member><function>get_declared_interfaces</function></member>
+    <member><function>is_a</function></member>
```
Girgias, line 119: Instanceof should also be mentionned here

alexandre-daubois: Added :slightly_smiling_face:

reference/spl/functions/class-implements.xml
alexandre-daubois: Updated!

Girgias (approved):

PR #2707, opened 2023-08-25
Update curl-getinfo.xml

JKJameson:
Commas in the example should be dots, not commas.


Girgias:
No, commas are valid as ``echo`` is a language construct and allows to output multiple arguments in a row.

PR #2712, opened 2023-08-26
Update string.xml to remove old string interpolation

kohler:
An example used the deprecated `${varname}` syntax; fix it.

### DIFF
--- a/language/types/string.xml
+++ b/language/types/string.xml
@@ -721,10 +721,10 @@ EOT;
 $juice = "apple";
 
 echo "He drank some $juice juice.".PHP_EOL;
-// Invalid. "s" is a valid character for a variable name, but the variable is $juice.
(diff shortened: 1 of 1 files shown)


adoy:
Thanks

adoy (approved):

salathe:
:-1: The example that was changed is now in the wrong section of that page. It is now an example of the "Complex (curly) syntax" but the example is within the "Simple syntax" section.

kohler:
What's your preferred fix? Complex syntax is the only non-deprecated way to handle the example.

adoy:
@salathe You're right my bad. I went to fast without paying much attention. 

I guess we should remove the part from this example that uses the complex syntax (the 3rd echo) and add this example in the complex curly syntax.

Girgias:
I was working on a rewrite of this whole page in #2000 as it was not a straightforward fix.

PR #2713, opened 2023-08-26
Standardise Getting Started example indentation

timacdonald:
There is inconsistent indentation within the "Getting started" section of the docs including the same example across different pages.

I feel that using 4 spaces is pretty established indentation in the PHP community and should be our standard within the docs.

[This example](https://www.php.net/manual/en/intro-whatis.php
):

Screen Shot 2023-08-27 at 7 33 48 am

is then repeated on [different pages](https://www.php.net/manual/en/tutorial.firstpage.php) in the same section with different indentation.

Screen Shot 2023-08-27 at 7 35 00 am

## Result

Screen Shot 2023-08-27 at 7 40 02 am

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -371,9 +371,9 @@ if (strpos($_SERVER['HTTP_USER_AGENT'], 'Firefox') !== false) {
      <programlisting role="html">
 <![CDATA[
 <form action="action.php" method="post">
- <p>Your name: <input type="text" name="name" /></p>
- <p>Your age: <input type="text" name="age" /></p>
- <p><input type="submit" /></p>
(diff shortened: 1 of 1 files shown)


TimWolla (commented):
I agree this is an improvement. Probably https://www.php.net/manual/en/tutorial.forms.php should also be updated, though.

timacdonald:
@TimWolla I've pushed changes for that example.

TimWolla:
I've pushed a small fix to drop trailing whitespace (that wasn't your fault, but was exposed due to the change to the line) and then merged. Thank you!

timacdonald:
Appreciate it!

PR #2715, opened 2023-08-27
Use semantic HTML tags for tutorial `form` example

timacdonald:
This PR improves the form example by:

1. Using `label` tags, rather than generic inaccessible `<p>` tags.
2. Removes the redundant `type` attribute from input tags.

> If this attribute is not specified, the default type adopted is text [source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types)

3. Adds an `id` tag to the `input` so that `label` tags can reference them with their `for` attibute.
3. Use a submit `button` and provide custom text. Adding custom text between an HTML tag feels much more familiar than doing it within an attribute. This also matches how the `label` works.
4. Added some spacing to ease the flow to reading the `label` / `input` pairs.
5. Removed the "closing" slash on the inputs. We've already shown the doctype in the opening example of the PHP docs showing that we aren't using "quirks mode" or an XML type that requires closing.

I can only assume that `<p>` tags were used for formatting. I don't think that a horizontal form is an issue and something we need to "format". I also don't think wrapping things in `<p>` or `<div>` is the way we should promote formatting in HTML, over CSS

## Docs before

Screenshot 2023-08-28 at 9 05 40 am

## Docs after

Screenshot 2023-08-28 at 9 05 31 am

## Rendered HTML before

Screenshot 2023-08-28 at 8 50 14 am

## Rendered HTML after

Screenshot 2023-08-28 at 8 55 29 am

This PR contains the same commit to indent the form seen in https://github.com/php/doc-en/pull/2713. This should allow both to merged / rejected independently without causing conflicts...that is if I know how git works lol 🤞

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -371,9 +371,9 @@ if (strpos($_SERVER['HTTP_USER_AGENT'], 'Firefox') !== false) {
      <programlisting role="html">
 <![CDATA[
 <form action="action.php" method="post">
- <p>Your name: <input type="text" name="name" /></p>
- <p>Your age: <input type="text" name="age" /></p>
- <p><input type="submit" /></p>
+    <p>Your name: <input type="text" name="name" /></p>
+    <p>Your age: <input type="text" name="age" /></p>
+    <p><input type="submit" /></p>
 </form>
 ]]>
      </programlisting>
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -372,12 +372,12 @@ if (strpos($_SERVER['HTTP_USER_AGENT'], 'Firefox') !== false) {
 <![CDATA[
 <form action="action.php" method="post">
     <label for="name">Your name:</label>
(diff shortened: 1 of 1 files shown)


TimWolla:
> Removes the redundant `type` attribute from input tags.

I disagree with that change. Explicitly providing the `type` makes the HTML more readable, especially since “age” could be `type="number"`. Likewise I believe it should be `<button type="submit">`.

timacdonald:
Thanks for the review, @TimWolla.

I've restored the `type="text"` attribute on the name input and added `type="number"` to the age input. I like `type="number"` here as the casting of the `(int) $_POST['age']` later shows that you can't just trust user input - even if subtly.

I've put the `type` attribute last, as the most important part of this tutorial, in my opinion, is the how the `"name"` ends up in the `$_POST`. The `type` attribute is there as a secondary thing not specifically related to the example at hand.

TimWolla:
Thank you. LGTM now.

PR #2716, opened 2023-08-27
Use `str_contains` over `strpos` for tutorial example

timacdonald:
The first example on the PHP documentation that isn't a "hello world" is searching the `User-Agent` for the string `"Firefox"`.

This example currently uses [`strpos`](https://www.php.net/manual/en/function.strpos.php). I believe we should be using [`str_contains`](https://www.php.net/manual/en/function.str-contains).

I believe this for a few reasons, but to start with I'll let the function descriptions do the talking:

> Find the position of the first occurrence of a substring in a string

vs.

> Determine if a string contains a given substring

`str_contains` matches the semantics of what we are doing and eases the overhead of understanding what is happening.

When using `strpos` we are also introducing the reader to:

1. "Truthiness"
2. Boolean literals
3. PHP's "return `false` as a flag to indicate something" behaviour

Which is all additional overhead.

The current example also has double negatives that are hare to read left to right. The `str_contains` example feels much easier to comprehend, i.e.,

> If strpos is not equal to `false`

versus

> If str_contains

and...

> returned non-false

vs

> returned true

I understand that `str_contain` is a "newer" PHP function, however these are the documentation for the current PHP language - so we should take advantage and promote it's newer features, when it makes sense.

The `str_contains` function is available in all currently supported PHP versions.

Lastly, I removed the `<br />` from the example as it serves no purpose to introducing the reader to "control structures and functions". The next example handles mixing in HTML in the "Mixing both HTML and PHP modes". The later examples also don't reference the `<br />` anyway, so kinda feels like unneeded noise we ca drop.

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -263,8 +263,8 @@ Mozilla/5.0 (Linux) Firefox/112.0
      <programlisting role="php">
 <![CDATA[
 <?php
-if (strpos($_SERVER['HTTP_USER_AGENT'], 'Firefox') !== false) {
-    echo 'You are using Firefox.<br />';
+if (str_contains($_SERVER['HTTP_USER_AGENT'], 'Firefox')) {
+    echo 'You are using Firefox.';
 }
 ?>
 ]]>
@@ -274,7 +274,7 @@ if (strpos($_SERVER['HTTP_USER_AGENT'], 'Firefox') !== false) {
      </para>
      <screen role="html">
 <![CDATA[
-You are using Firefox.<br />
+You are using Firefox.
 ]]>
      </screen>
     </example>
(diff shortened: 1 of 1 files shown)


chapters/tutorial.xml
```diff
@@ -319,14 +318,14 @@ You are using Firefox.<br />
      <programlisting role="php">
 <![CDATA[
 <?php
-if (strpos($_SERVER['HTTP_USER_AGENT'], 'Firefox') !== false) {
+if (str_contains($_SERVER['HTTP_USER_AGENT'], 'Firefox')) {
 ?>
-<h3>strpos() must have returned non-false</h3>
```
timacdonald, line 324: The use of "must have" feels like there is a possibility that we are here and it _didn't_ return `true` or `non-false`.

Question: "Did Frank clean the car?"

Answer: "Must have".

Narrator: "It was actually the rain the cleaned the car".

TimWolla:
> I understand that `str_contain` is a "newer" PHP function, however these are the documentation for the current PHP language - so we should take advantage and promote it's newer features, when it makes sense.

Agreed. This change LGTM, thank you!

PR #2721, opened 2023-08-29
PDO::lastInsertId of pdo_pgsql no longer requires sequence name

takaram:
`name` parameter is not mandatory since 7.0.11.
https://bugs.php.net/bug.php?id=72633


Girgias:
The manual documents everything from PHP 7.0.0 onwards currently, so this requires a changelog entry.

takaram:
Closing in favor of #2986

PR #2731, opened 2023-09-02
Standardise "open source" capitalisation

timacdonald:
I don't  have strong feelings on if this should be capitalised or not, but I do think we should reference it consistently.

I appears 8 times throughout the docs. Only two are capitalised.

Additionally, a similar sentence on the Preface and Getting Started pages uses differing casing.

Screen Shot 2023-09-02 at 1 13 30 pm

Screen Shot 2023-09-02 at 1 13 40 pm

### DIFF
--- a/preface.xml
+++ b/preface.xml
@@ -6,7 +6,7 @@
  <abstract>
   <simpara>
    <acronym>PHP</acronym>, which stands for &quot;<literal>PHP: Hypertext
-   Preprocessor</literal>&quot; is a widely-used Open Source general-purpose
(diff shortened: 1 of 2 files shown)


Girgias (approved):

Girgias:
Going to wait before merging this to wait for translations to catch up with some large markup changes.

salathe:
There is also 1 usage of `open-source` present in `security/database.xml`.

salathe:
This might also be worth adding to the [Style guidelines](http://doc.php.net/tutorial/style.php) ([source](https://github.com/php/web-doc/blob/master/tutorial/style.md)).

PR #2732, opened 2023-09-02
Wrap output string in `literal`

timacdonald:
I _think_ that we want strings like this wrapped in literal tags. Please let me know if this is not correct.

Screen Shot 2023-09-02 at 1 21 11 pm


chapters/intro.xml
```diff
@@ -38,8 +38,8 @@
    </para>
    <para>
     Instead of lots of commands to output HTML (as seen in C or Perl),
-    PHP pages contain HTML with embedded code that does
-    "something" (in this case, output "Hi, I'm a PHP script!").
+    PHP pages contain HTML with embedded code that does "something" (in this
+    case, output <literal>"Hi, I'm a PHP script!"</literal>).
```
TimWolla, line 42: Not an expert with regard to the markup, but it should likely be:
```suggestion
    case, output <literal>Hi, I'm a PHP script!</literal>).
```

Girgias: Indeed the quotation marks should be removed, but a better tag would be [``<computeroutput>``](https://tdg.docbook.org/tdg/5.1/computeroutput)

timacdonald: Updated the tag and removed the quotation marks.

chapters/tutorial.xml
```diff
@@ -281,7 +281,7 @@ You are using Firefox.
    </para>
    <para>
     Here we introduce a couple of new concepts. We have an 
-    <link linkend="control-structures.if">if</link> statement.
+    <literal><link linkend="control-structures.if">if</link></literal> statement.
```
timacdonald, line 284: https://www.php.net/manual/en/tutorial.useful.php

Screen Shot 2023-09-03 at 9 17 24 am

chapters/tutorial.xml
```diff
@@ -294,13 +294,13 @@ You are using Firefox.
...
-    linkend="control-structures.if">if</link> expression evaluates to &true;
+    the needle is found inside the haystack, the function returns &true;. Otherwise, it
+    returns &false;. If it returns &true;, the <literal><link
+    linkend="control-structures.if">if</link></literal> expression evaluates to &true;
     and the code within its {braces} is executed. Otherwise, the code is not
     run. Feel free to create similar examples, 
-    with <link linkend="control-structures.if">if</link>, 
-    <link linkend="control-structures.else">else</link>, and other 
+    with <literal><link linkend="control-structures.if">if</link></literal>,
+    <literal><link linkend="control-structures.else">else</link></literal>, and other
```
timacdonald, line 303: Screen Shot 2023-09-03 at 9 18 05 am

https://www.php.net/manual/en/tutorial.useful.php

chapters/tutorial.xml
```diff
@@ -349,7 +349,7 @@ if (str_contains($_SERVER['HTTP_USER_AGENT'], 'Firefox')) {
     to note here is that the logical flow of the script remains intact. Only
     one of the HTML blocks will end up getting sent to the viewer depending on
     the result of <function>str_contains</function>.  In other words, it depends on
-    whether the string <literal>Firefox</literal> was found or not.
+    whether the string <literal>'Firefox'</literal> was found or not.
```
timacdonald, line 352: Screen Shot 2023-09-03 at 9 18 27 am

https://www.php.net/manual/en/tutorial.useful.php

Previously on this page, we already reference the string in quotations..

Screen Shot 2023-09-03 at 9 19 14 am

timacdonald:
@TimWolla I hope you don't mind, I decided to add some more improvements to this PR.

Because they are all similar, I thought we could likely just get them going here to reduce overhead for you.

Couple of questions about contributing, while I have you here:

1. Do you mind me grouping similar changes like this or would you prefer more atomic changes?
2. Do we have a "style guide" for the docs, and if not, would you mind if I PR'd one to the repository with a few things we are improving here - so that we have internal consistency moving forward?

chapters/tutorial.xml
Girgias: This might make more sense to add then in the ``doc-base/entities/global.ent`` file, as we already declare various control structures as entities.

TimWolla:
@timacdonald I'm afraid I'm really the wrong person to ask here. I'm not an expert with regard to the documentation. Other than reviewing the ext/random stuff which I maintain, I just try handle to _obvious_ improvements so that the more knowledgeable folks can focus on the less obvious stuff.

Girgias:
> @TimWolla I hope you don't mind, I decided to add some more improvements to this PR. I think these are the last of these kinda of markup improvements I wanna make for the intro / tutorial.
> 
> We could probably add additional around acronyms - but I might leave that for now.
> 
> Couple of questions about contributing to the docs, while I have you here:
> 
>     1. Do you mind me grouping similar changes like this or would you prefer more atomic changes?
> 
>     2. Do we have a "style guide" for the docs, and if not, do you think one would be useful? I am happy to PR an small one with some of the small improvements I have been making - and we can expand on it into the future?

There is a style guide on the doc website: http://doc.php.net/tutorial/style.php

The main thing with regards to changes is ideally for them to be atomic, but also not doing unnecessary changes that needlessly affect translations.

As such basic typos should be filed separately with a ``[skip-revcheck]`` commit so that a file is not marked as outdated for translations.

Code style is also a hot topic and shouldn't be changed just for the sake of it.

chapters/tutorial.xml
Girgias: Ping @timacdonald if you still planning to work on this PR, addressing the above comment would be helpful.

timacdonald: Hey, I'll look at getting onto this shortly. Will mark this one as a draft until I get that sorted.

PR #2734, opened 2023-09-02
Add doctype to full HTML document

timacdonald:
Proposing that for **full HTML document** examples, we include a doctype.

The section's initial example includes a doctype, but then we don't include it on the subsequent examples.

Has doctype: https://www.php.net/manual/en/intro-whatis.php

Screen Shot 2023-09-02 at 4 36 03 pm

Missing doctype: https://www.php.net/manual/en/tutorial.firstpage.php

Screen Shot 2023-09-02 at 4 35 50 pm

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -69,6 +69,7 @@
      <info><title>Our first PHP script: <filename>hello.php</filename></title></info>
      <programlisting role="php">
 <![CDATA[
+<!DOCTYPE html>
 <html>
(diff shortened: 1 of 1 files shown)


TimWolla:
I think this makes sense, especially for consistency within the tutorial. Did you check whether any other places would need the DOCTYPE then?

Girgias (approved):

Girgias:
Thank you!

PR #2736, opened 2023-09-02
Wrap example output in CDATA

timacdonald:
https://www.php.net/manual/en/tutorial.useful.php

The output of the first example includes an initial new line.

Screen Shot 2023-09-02 at 5 06 48 pm

The script to generate this output will not generate a newline. Here is the output of the given script:

Screen Shot 2023-09-02 at 4 59 55 pm

Wrapping in CDATA, which seems to be the pattern used for `screen` tags, fixes this.

Screen Shot 2023-09-02 at 5 00 17 pm

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -232,7 +232,9 @@ echo $_SERVER['HTTP_USER_AGENT'];
      A sample output of this script may be:
     </para>
     <screen role="html">
+<![CDATA[
 Mozilla/5.0 (Linux) Firefox/112.0
+]]>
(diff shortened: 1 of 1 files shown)


TimWolla (approved):
Interesting. Without the CDATA the output is not wrapped in a `<pre>`. I wonder if this is a bug in PHD? In any case this change also makes sense for consistency. I'd like to have another pair of eyes on this, though.

Girgias (approved):

Girgias:
Thank you!

PR #2738, opened 2023-09-03
Added syntax highlighting for currentdirectory in 'Using Namespace' page

Stancobridge:
I was reading the documentation on the namespace and got confused when I saw the word currentdirectory, I thought it was a typo error but noticed that it is referencing a part of an initially highlighted part of the text <literal>currentdirectory/foo.txt</literal>.

Since this is used as an example, it has to be highlighted as well, I added the highlight


TimWolla (approved):

TimWolla:
Thank you!

PR #2741, opened 2023-09-05
Synchronize predefined constants with stubs - part 2

kocsismate:
Fix wrong formatting of constant types


TimWolla (approved):
ack for random

PR #2744, opened 2023-09-05
Synchronize list of reserved constants with stubs

kocsismate:
OK, this is the very last one. :)


Girgias:
Could you spin out the whitespace only change into a different PR that can be skipped by revcheck? as I didn't see the core INI file initially had a change

kocsismate:
> Could you spin out the whitespace only change into a different PR that can be skipped by revcheck?

I created two commits instead in this PR: first one fixes the indentation as well as trailing whitespaces, and the second one synchronizes the actual content.

PR #2753, opened 2023-09-10
Sort mis-ordered modifier in code example #2

xSavitar:
This patch proposes a fix to something that looks like a typo, supposed to be `public static ...` and not the other way around.

### DIFF
--- a/reference/funchand/functions/call-user-func-array.xml
+++ b/reference/funchand/functions/call-user-func-array.xml
@@ -134,7 +134,7 @@ foo::bar got three and four
 namespace Foobar;
 
 class Foo {
-    static public function test($name) {
+    public static function test($name) {
         print "Hello {$name}!\n";
     }
 }


Girgias:
That's a PSR convention, PHP doesn't care about the order of modifiers.

It may make sense to align it with PSR, but we don't really follow any coding standard for examples currently, as for a long time this would have been a huge impact for translations.

xSavitar:
> That's a PSR convention, PHP doesn't care about the order of modifiers.
> 
> It may make sense to align it with PSR, but we don't really follow any coding standard for examples currently, as for a long time this would have been a huge impact for translations.

Right, I was trying to be consistent here with what we already do in most of the other examples (on the site). And also, if it's a PSR, let's just stick to it ☺️

But let me know if you feel strongly opinionated about not changing it.

salathe:
If this is for consistency's sake, there are other places to change:

```
language/oop5/basic.xml:    static public function getNew()
reference/classobj/functions/get-called-class.xml:    static public function test() {
reference/funchand/functions/call-user-func-array.xml:    static public function test($name) {
reference/funchand/functions/call-user-func.xml:    static public function test() {
reference/spl/functions/spl-autoload-register.xml:    static public function test($class) {
```

kamil-tekiela:
We strive to be consistent only within a single page. Otherwise we would need to adopt a strict coding standard, but that would be too much of work for translations. 

Thank you for this PR, but I don't see the need to merge it. It's fine as it is.

PR #2767, opened 2023-09-13
$message in stream_notification_callback() is nullable

spaze:
At least for HTTP streams, the `$message` parameter of `stream_notification_callback()` is nullable, let's document it.

See https://onlinephp.io/c/f0a2fc for a simple script to get the types.


Girgias (changes requested):
This is not the markup of how a union type is written.

But the change is correct.

@kocsismate I wonder what we should do with those pages, as there are no stubs/interfaces for custom streams or related functions

spaze:
Oh, I'm sorry I messed it up. Corrected the markup.

Girgias (approved):

Girgias:
Thank you!

PR #2771, opened 2023-09-18
Synchronize list of predefined constants with stubs - part 6

kocsismate:


reference/math/constants.xml
```diff
@@ -17,149 +17,217 @@
     </thead>
     <tbody>
```
Girgias, line 18: Would it be possible to also change the wording in the header row to use the relevant entities?

Also, this table looks inconsistent with the other table (and the availability row seems completely pointless here and could be removed).

Girgias (commented):
I wonder if doesn't make more sense to just convert these constant pages to use the more usual formatting? Do you know how many constant pages currently use tables and how many use the other markup?

kocsismate:
The following other predefined constant pages use the table-based approach:
- tokens: https://www.php.net/manual/en/tokens
- error handling: https://www.php.net/manual/en/errorfunc.constants.php
- magic constants: https://www.php.net/manual/en/language.constants.magic.php
- built-in constants: https://www.php.net/manual/en/info.constants.php
- OCI8 constants: https://www.php.net/manual/en/oci8.constants.php (and https://www.php.net/manual/en/oci8.datatypes.php should be included as well)
- Stream constants: https://www.php.net/manual/en/stream.constants.php
- Tidy constants: `https://www.php.net/manual/en/tidy.constants.php`

(maybe there are a few more, but I could only find these after a quick search)

So all in all, conversion to the list based format is possible since there are not too many constant tables. IMO the only place where the table makes sense it the list of tokens...

kocsismate:
If we wanted to migrate https://www.php.net/manual/en/function.curl-setopt.php to a predefined constant page then using the list based format is also questionable, using tables look more appropriate 🤔 . So I don't have a good answer whether we should really do the conversion to lists.

Girgias:
> On the other hand, if we wanted to migrate https://www.php.net/manual/en/function.curl-setopt.php to a predefined constant page then using the list based format is also questionable, using tables look more appropriate 🤔 . So I don't have a good answer whether we should really do the conversion to lists, since there are a very few exceptions.

The cURL setopt API is just badly designed that I have no idea how to reasonably document it.

But yes I don't know what is the best format for documenting them, the list based approach is good for long descriptions, the table one is good for an overview.

Maybe doing something like INI pages where there is first a table that gives an overview and availability and then below a a list of them with individual descriptions?

See https://www.php.net/manual/en/ini.core.php#ini.sect.language-options

PR #2776, opened 2023-09-19
`compact()` now issues an E_WARNING error.(PHP8.0)

o0h:
Since PHP 8.0: "[A number of notices have been converted into warnings](https://www.php.net/manual/en/migration80.incompatible.php)"

Usually, in the PHP manual, even if the error level is raised from E_NOTICE to E_WARNING, a changelog isn't kept within individual function pages.
However, on the compact() page, there's already an entry in the changelog regarding the previous change to 'now issues an E_NOTICE'. Therefore, I believe that adding a note explicitly indicating the recent change in the issued error might prevent confusion.

### DIFF
--- a/reference/array/functions/compact.xml
+++ b/reference/array/functions/compact.xml
@@ -58,7 +58,7 @@
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   <function>compact</function> issues an E_NOTICE level error if a given string
+   <function>compact</function> issues an E_WARNING level error if a given string
    refers to an unset variable.
   </para>
  </refsect1>
@@ -74,6 +74,12 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       If a given string references an unset variable, an E_WARNING level error is now issued.
+      </entry>
+     </row>
      <row>
       <entry>7.3.0</entry>
       <entry>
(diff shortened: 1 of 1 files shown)


reference/array/functions/compact.xml
```diff
@@ -58,7 +58,7 @@
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   <function>compact</function> issues an E_NOTICE level error if a given string
+   <function>compact</function> issues an E_WARNING level error if a given string
```
Girgias, line 61:
```suggestion
   <function>compact</function> issues an <constant>E_WARNING</constant> level error if a given string
```

reference/array/functions/compact.xml
```diff
@@ -74,6 +74,12 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       If a given string references an unset variable, an E_WARNING level error is now issued.
```
Girgias, line 80:
```suggestion
       If a given string references an unset variable, an <constant>E_WARNING</constant> level error is now issued.
```

Girgias (approved):
Thank you, could you just add the missing markup to the E_WANRING and E_NOTICE constants?

o0h:
Thanks! I corrected the markup.

PR #2777, opened 2023-09-20
Corrections to xml_parser_[gs]et_option

nielsdos:
Reference: https://github.com/php/php-src/commit/81e59c64979c458b658695aac85fb8dbaad4a7ef
Came across this by coincidence and noticed the manpage is not yet up-to-date for 8.3 and there was no PR, so here we go.

### DIFF
--- a/reference/xml/functions/xml-parser-set-option.xml
+++ b/reference/xml/functions/xml-parser-set-option.xml
@@ -12,7 +12,7 @@
    <type>bool</type><methodname>xml_parser_set_option</methodname>
    <methodparam><type>XMLParser</type><parameter>parser</parameter></methodparam>
    <methodparam><type>int</type><parameter>option</parameter></methodparam>
-   <methodparam><type class="union"><type>string</type><type>int</type></type><parameter>value</parameter></methodparam>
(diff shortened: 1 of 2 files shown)


reference/xml/functions/xml-parser-set-option.xml
```diff
@@ -106,8 +106,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   This function returns &false; if <parameter>parser</parameter> does not
-   refer to a valid parser. Else the option's is set and &true; is returned.
+   Returns &true; on success or &false; on failure.
```
Girgias, line 109: Does it even fail nowadays?

Girgias (approved):
LGTM

reference/xml/functions/xml-parser-set-option.xml
nielsdos: It can return false for PHP_XML_OPTION_SKIP_TAGSTART if you give a negative value.

Girgias: Ignore me... I was nice and made it a warning and not a type error

PR #2781, opened 2023-09-21
Fix bug #61907: XML Error Codes constants should list integer value

ndossche:
Fixes https://bugs.php.net/bug.php?id=61907.
Stole the table from exif image types page and adapted for ext/xml.

### DIFF
--- a/reference/xml/error-codes.xml
+++ b/reference/xml/error-codes.xml
@@ -5,30 +5,107 @@
  <para>
   The following constants are defined for XML error codes (as
   returned by <function>xml_parse</function>):
-  <simplelist>
-   <member><constant>XML_ERROR_NONE</constant></member>
-   <member><constant>XML_ERROR_NO_MEMORY</constant></member>
-   <member><constant>XML_ERROR_SYNTAX</constant></member>
-   <member><constant>XML_ERROR_NO_ELEMENTS</constant></member>
-   <member><constant>XML_ERROR_INVALID_TOKEN</constant></member>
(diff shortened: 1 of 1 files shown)


Girgias:
@kocsismate didn't we kind of decide to stop showing unknown constant values (as they come from external libraries)?

nielsdos:
These aren't really unknown, as they are also emulated by the expat compatibility layer in PHP, and thus must match the library's values.

Girgias:
> These aren't really unknown, as they are also emulated by the expat compatibility layer in PHP, and thus must match the library's values.

Right, but they are not defined as such in the stub file? As this is how we generate the constant pages now.

nielsdos:
> > These aren't really unknown, as they are also emulated by the expat compatibility layer in PHP, and thus must match the library's values.
> 
> Right, but they are not defined as such in the stub file? As this is how we generate the constant pages now.

Ah, well, they're defined UNKNOWN like this:
```php
/**
 * @var int
 * @cvalue XML_ERROR_EXTERNAL_ENTITY_HANDLING
 */
const XML_ERROR_EXTERNAL_ENTITY_HANDLING = UNKNOWN;
```

Girgias:
We might need to open up a proper discussion on the doc ML about documenting constant values.

I personally don't find it _super_ useful, but that's just my opinion.

jordikroon:
Given this is marked as UNKNOWN in the stubs it's kinda hard to maintain. Also I don't think it adds any value as I wouldn't encourage anyone to depend on the integer values. Therefore I am closing this issue. If you want to pursue on the idea I suggest to open a discussion on the mailing list as Gina suggested.

PR #2782, opened 2023-09-22
modernize ErrorException example code

divinity76:
IMO the old example should have been 
```
set_error_handler(__NAMESPACE__ . "\\exception_error_handler");
```
so it would work in any namespace, because the old example just worked in the global namespace, and crashed-at-runtime if used in any other namespace,  but as of PHP8.1.0 we have an even better way to do it: set_error_handler(exception_error_handler(...));

### DIFF
--- a/language/predefined/errorexception.xml
+++ b/language/predefined/errorexception.xml
@@ -89,7 +89,7 @@ function exception_error_handler(int $errno, string $errstr, ?string $errfile, i
     throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
 }
 set_error_handler(__NAMESPACE__ . "\\exception_error_handler");
-// or in PHP>=8.1: set_error_handler(exception_error_handler(...));
+// PHP>=8.1: set_error_handler(exception_error_handler(...));
 
 /* Trigger exception */
 strpos();
--- a/language/predefined/errorexception.xml
+++ b/language/predefined/errorexception.xml
@@ -86,7 +86,7 @@ function exception_error_handler(int $errno, string $errstr, string $errfile, in
         // This error code is not included in error_reporting
         return;
     }
-    throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
+    throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
 }
 set_error_handler(exception_error_handler(...));
 
--- a/language/predefined/errorexception.xml
+++ b/language/predefined/errorexception.xml
@@ -81,7 +81,7 @@
      <programlisting role="php">
  <![CDATA[
 <?php
-function exception_error_handler(int $errno, string $errstr, string $errfile == null, int $errline) {
+function exception_error_handler(int $errno, string $errstr, string $errfile = null, int $errline) {
     if (!(error_reporting() & $errno)) {
         // This error code is not included in error_reporting
         return;
@@ -89,7 +89,7 @@ function exception_error_handler(int $errno, string $errstr, string $errfile == 
     throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
 }
 set_error_handler(exception_error_handler(...));
-// Prior to PHP 8.1.0 and the introduction of the first class callable syntax, the following call must be used to 
+// Prior to PHP 8.1.0 and the introduction of the first class callable syntax, the following call must be used instead 
 // set_error_handler(__NAMESPACE__ . "\\exception_error_handler");
 
 /* Trigger exception */
--- a/language/predefined/errorexception.xml
+++ b/language/predefined/errorexception.xml
@@ -81,14 +81,14 @@
      <programlisting role="php">
  <![CDATA[
 <?php
-function exception_error_handler($severity, $message, $file, $line) {
-    if (!(error_reporting() & $severity)) {
+function exception_error_handler(int $errno, string $errstr, string $errfile, int $errline) {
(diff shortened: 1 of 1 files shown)


language/predefined/errorexception.xml
```diff
@@ -81,14 +81,14 @@
...
+function exception_error_handler(int $errno, string $errstr, string $errfile, int $errline) {
+    if (!(error_reporting() & $errno)) {
         // This error code is not included in error_reporting
         return;
     }
-    throw new ErrorException($message, 0, $severity, $file, $line);
+    throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
 }
-set_error_handler("exception_error_handler");
+set_error_handler(exception_error_handler(...));
```
Girgias, line 91: Please revert this, the documentation must cover PHP 7 and 8, and unless an explanation is somewhere given about the syntax and how to pass the callable in PHP 7, it just simpler to let the callable be a string.

language/predefined/errorexception.xml
```diff
@@ -81,14 +81,14 @@
      <programlisting role="php">
  <![CDATA[
 <?php
-function exception_error_handler($severity, $message, $file, $line) {
-    if (!(error_reporting() & $severity)) {
+function exception_error_handler(int $errno, string $errstr, string $errfile, int $errline) {
```
Girgias, line 84:
```suggestion
function exception_error_handler(int $errno, string $errstr, ?string $errfile, int $errline) {
```
The file can be NULL: https://heap.space/xref/php-src/Zend/zend.c?r=692cea5c#1452

divinity76: you brought up that PHP7 must still be documented. I wonder where the limit goes, `?string $errfile` is valid as of PHP7.1 but not PHP7.0, but `string $errfile=null` would also be valid in PHP7.0. Should we care? My personal opinion is to disregard PHP<7.4 compatibility, but what's your opinion?

language/predefined/errorexception.xml
divinity76: please re-review

language/predefined/errorexception.xml
Girgias: The policy of the docs is to document everything from PHP 7.0, your opinion is not very relevant here.

language/predefined/errorexception.xml
```diff
@@ -81,14 +81,15 @@
      <programlisting role="php">
  <![CDATA[
 <?php
-function exception_error_handler($severity, $message, $file, $line) {
-    if (!(error_reporting() & $severity)) {
+function exception_error_handler(int $errno, string $errstr, ?string $errfile, int $errline) {
+    if (!(error_reporting() & $errno)) {
```
Girgias, line 85:
```suggestion
function exception_error_handler(int $errno, string $errstr, string $errfile == null, int $errline) {
    if (!(error_reporting() & $errno)) {
```

language/predefined/errorexception.xml
```diff
@@ -81,14 +81,15 @@
...
         // This error code is not included in error_reporting
         return;
     }
-    throw new ErrorException($message, 0, $severity, $file, $line);
+    throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
 }
-set_error_handler("exception_error_handler");
+set_error_handler(__NAMESPACE__ . "\\exception_error_handler");
+// PHP>=8.1: set_error_handler(exception_error_handler(...));
```
Girgias, line 92:
```suggestion
set_error_handler(exception_error_handler(...));
// Prior to PHP 8.1.0 and the introduction of the first class callable syntax, the following call must be used to 
// set_error_handler(__NAMESPACE__ . "\\exception_error_handler");
```

Girgias (commented):
I'm not sure I actually like the variable name changes but that's somewhat whatever

language/predefined/errorexception.xml
divinity76: @Girgias in that case we need to change `?string $errfile` to `string $errfile=null` to remain PHP7.0-compatible yes?

divinity76:
>I'm not sure I actually like the variable name changes but that's somewhat whatever

That was actually fixing a consistency issue - other parts of the documentation uses the variable names I switched to here. (Am on phone right now so can finding out where in the docs, but shouldn't be hard to find)

language/predefined/errorexception.xml
```diff
@@ -81,14 +81,16 @@
...
+    if (!(error_reporting() & $errno)) {
         // This error code is not included in error_reporting
         return;
     }
-    throw new ErrorException($message, 0, $severity, $file, $line);
+    throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
 }
-set_error_handler("exception_error_handler");
+set_error_handler(exception_error_handler(...));
+// Prior to PHP 8.1.0 and the introduction of the first class callable syntax, the following call must be used instead
```
Girgias, line 92:
```suggestion
// Prior to PHP 8.1.0 and the introduction of the first class callable syntax, the following call must be used instead
```

Girgias:
> > I'm not sure I actually like the variable name changes but that's somewhat whatever
> 
> That was actually fixing a consistency issue - other parts of the documentation uses the variable names I switched to here. (Am on phone right now so cba finding out where in the docs, but shouldn't be hard to find)
> 
> Edit: I think it's the set_error_handler documentation.

Right, consistency is best.

Girgias:
Thank you!

PR #2784, opened 2023-09-22
password_hash: Update script to generate cost recommendations

TimWolla:
50 ms is way to short to generate actually secure hashes. Update the recommendation to 350 ms with a starting cost of 10 (the current default) and example output of 12.

see https://wiki.php.net/rfc/bcrypt_cost_2023

/cc @roycewilliams


Girgias (approved):
It would be nice to get rid of the personalisation too (usage of ``you`` in the docs)

PR #2788, opened 2023-09-22
Clarify that the constructor is not called with classes registered via registerNodeClass

nielsdos:


TimWolla (approved):

Girgias (approved):
As a follow-up it would be nice to get rid of the personalisation (usage of ``you``)

Issue #2796, opened 2023-09-26
PHP 8.3 documentation tracker


Girgias:
Pages which need to be added/modified for complete documentation of PHP 8.3:

Based of the migration guide: https://www.php.net/manual/en/migration83.php (#2780):

> **Note**
> This issue is still a Work In Progress

## Core

- [x] New features:
   - [x] Readonly Amendments
     - [x] Anonymous classes can be marked as readonly (#4171)
     - [x] Readonly Properties can be reinitialized during cloning (#4172)
   - [x] Override attribute (#2996)
   - [x] Typed Class constants (Part #2798)
   - [x] #4193
   - [x] final methods on trait (#4167)
   - [x] Fetch class constant syntax (#4168)
   - [x] Static Variable Initializer (#4169)
   - [x] #4191
- [x] Deprecated
  - [x] Saner Increment/Decrement operators (#2841)
    - [x] Using the [increment](https://www.php.net/manual/en/language.operators.increment.php) operator (++) on empty, non-numeric, or non-alphanumeric strings is now deprecated. Moreover, incrementing non-numeric strings is soft deprecated and the new str_increment() function should be used instead.
    - [x] Using the [decrement](https://www.php.net/manual/en/language.operators.increment.php) operator (--) on empty or non-numeric strings is now deprecated.
  - [x] get_class()/get_parent_class() call without arguments (#3186)
- [x] Windows Support:
  - [x] #3810
- [x] BC Break:
  - [x] proc_get_status() (https://github.com/php/doc-en/pull/4112)
  - [x] Uses of traits with static properties (#4182)
  - [x] Assigning a negative index to an empty array (#4173)
  - [x] Class constant visibility variance check (#4183)

## SAPI

- [x] CLI:
   - [x] It is now possible to lint multiple files. (#2917)

## Extensions

 - [x] DateTime:
   - [x] New methods:
     - [x] DatePeriod::createFromISO8601String() (7d81260767f008218ffd338b365cfa3d10eb3d15)
   - [x] New error classes 

 - [x] DBA:
   - [x] Deprecated:
     - [x] Calling [dba_fetch()](https://www.php.net/manual/en/function.dba-fetch.php) with $dba as the 3rd argument is now deprecated. (#2911)

 - [x] DOM:
   - [x] New properties:
     - [x] ``DOMElement::$className``
     - [x] ``DOMElement::$id``
     - [x] ``DOMNode::$isConnected``
[...]

PR #2797, opened 2023-09-26
Document str_increment() and str_decrement()

Girgias:
Part of #2796

### DIFF
--- a/reference/strings/functions/str-decrement.xml
+++ b/reference/strings/functions/str-decrement.xml
@@ -1,0 +1,135 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xmlns="http://docbook.org/ns/docbook" xml:id="function.str-decrement">
+ <refnamediv>
+  <refname>str_decrement</refname>
+  <refpurpose>Decrement an alphanumeric string</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>string</type><methodname>str_decrement</methodname>
+   <methodparam><type>string</type><parameter>string</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Returns the decremented alphanumeric <acronym>ASCII</acronym>
+   <parameter>string</parameter>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>string</parameter></term>
+     <listitem>
+      <para>
+       The input string.
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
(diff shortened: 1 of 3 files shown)


saundefined (approved):
Looks good, thanks!

reference/strings/functions/str-decrement.xml
```diff
@@ -0,0 +1,135 @@
...
+   <parameter>string</parameter> is empty.
+  </para>
+  <para>
+   A <classname>ValueError</classname> is thrown if
+   <parameter>string</parameter> is not an alphanumeric
+   <acronym>ASCII</acronym> string.
+  </para>
+  <para>
+   A <classname>ValueError</classname> is thrown if
+   <parameter>string</parameter> is out of the decrement range.
```
TimWolla, line 56: When is that? At `"0"`?

mumumu: > When is that? At "0"?

Seems to At "1".
ValueError is thrown if we try to decrement "1".

```
<?php

$a = 'a5';
for ($i = 0; $i < 1000; $i++) {
    var_dump("decrementing $a");
    $a = str_decrement($a);
    var_dump($a);
}
```
```
sapi/cli/php test.php
string(15) "decrementing a1"
string(2) "a0"
string(15) "decrementing a0"
string(1) "9"
string(14) "decrementing 9"
string(1) "8"
string(14) "decrementing 8"
string(1) "7"
string(14) "decrementing 7"
string(1) "6"
string(14) "decrementing 6"
string(1) "5"
string(14) "decrementing 5"
string(1) "4"
string(14) "decrementing 4"
string(1) "3"
string(14) "decrementing 3"
string(1) "2"
string(14) "decrementing 2"
string(1) "1"
string(14) "decrementing 1"

Fatal error: Uncaught ValueError: str_decrement(): Argument #1 ($string) "1" is out of decrement range in /home/mumumu/build/php-src/test.php:6
```

mumumu: It may be better to describe the decrement range by example. @Girgias

Girgias: I might need to ask @iluuu1994 as he changed the decrement function after merging the RFC PR.

iluuu1994: Right, "1" erroring is a bug. The fix is here: https://github.com/php/php-src/pull/12339

This error occurs when there's a remaining carry after all digits have been decremented. This can occur for string like "000", "0a", etc. The leading digit is also removed if it is a 0 (e.g. when going from "10" to "09"). However, I have not considered the case of "1" going to "0", with only a single digit, which does not need to be removed.

mumumu, line 56: "1" erroring is now fixed, so we can add the exact range description.

```suggestion
   <parameter>string</parameter> is out of the decrement range (i.e. <literal>"0"</literal>).
```

mumumu (approved):

PR #2799, opened 2023-09-26
Split operator XML file into multiple files

Girgias:
The documentation already renders it as different pages as each operator type is wrapped in a sect1 tag.

This will make it easier to update specific sections without making it a pain for translations to need to sync a huge file.


Girgias:
If there are no comments, I will merge this on Monday

mumumu (approved):
I've already experienced this type of BIG change in #1726, np for me.

saundefined (approved):

PR #2801, opened 2023-09-27
range: Update for PHP 8.3

Girgias:
~~TODO Better examples~~ I can't come up with them, and they can always be added later.

### DIFF
--- a/reference/array/functions/range.xml
+++ b/reference/array/functions/range.xml
@@ -77,12 +77,15 @@
    <parameter>step</parameter> values apart.
   </para>
   <para>
-   The last element of the returned array is either <parameter>end</parameter> or the previous element of the sequence, depending on the value of <parameter>step</parameter>
+   The last element of the returned array is either <parameter>end</parameter>
+   or the previous element of the sequence,
+   depending on the value of <parameter>step</parameter>.
   </para>
   <para>
    If both <parameter>start</parameter> and <parameter>end</parameter> are
    <type>string</type>s, and <parameter>step</parameter> is <type>int</type>
-   the produced array will be a sequence of characters.
+   the produced array will be a sequence of bytes,
+   generally latin <abbrev>ASCII</abbrev> characters.
   </para>
   <para>
    If at least one of <parameter>start</parameter>, <parameter>end</parameter>,
@@ -102,7 +105,7 @@
   </para>
   <para>
    If <parameter>start</parameter>, <parameter>end</parameter>,
-   or <parameter>step</parameter> is not finite,
+   or <parameter>step</parameter> is not <function>is_finite</function>,
    a <classname>ValueError</classname> is thrown.
   </para>
   <para>
--- a/reference/array/functions/range.xml
+++ b/reference/array/functions/range.xml
@@ -18,7 +18,7 @@
   <para>
    If both <parameter>start</parameter> and <parameter>end</parameter> are
    <type>string</type>s, and <parameter>step</parameter> is <type>int</type>
-   the produced array will be a sequence of characters.
+   the produced array will be a sequence of bytes.
    Otherwise, the produced array will be a sequence of numbers.
   </para>
   <para>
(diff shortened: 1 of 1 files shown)


reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
     </varlistentry>
     <varlistentry>
      <term><parameter>step</parameter></term>
      <listitem>
       <para>
-       If a <parameter>step</parameter> value is given, it will be used as the
-       increment (or decrement) between elements in the sequence.  <parameter>step</parameter>
-       must not equal <literal>0</literal> and must not exceed the specified range.  If not specified,
-       <parameter>step</parameter> will default to 1.
+       By default <literal>1</literal>.
```
TimWolla, line 55:
```suggestion
```
This is already given in the signature, I don't think it is necessary to repeat that here.

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
     </varlistentry>
    </variablelist>
   </para>
  </refsect1>
+
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
    Returns an array of elements from <parameter>start</parameter> to
    <parameter>end</parameter>, inclusive.
```
TimWolla, line 76: For ext/random ranges I generally use the term “closed interval“ + an explanation what that means. Should this be unified:
```suggestion
   Returns an array of elements from the closed interval [<parameter>start</parameter>,<parameter>end</parameter>]. Both <parameter>start</parameter> and <parameter>end</parameter> are possible return values.
```

Possibly with a caveat regarding the step size: `$end` is not guaranteed to be returned AFAICT.

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
   </para>
+  <para>
+   If both <parameter>start</parameter> and <parameter>end</parameter> are
+   <type>string</type>s, and <parameter>step</parameter> is <type>int</type>
+   the produced array will be a sequence of characters.
+  </para>
+  <para>
+   If <parameter>start</parameter>, <parameter>end</parameter>,
```
TimWolla, line 84: “at least one of”?

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
    Returns an array of elements from <parameter>start</parameter> to
    <parameter>end</parameter>, inclusive.
   </para>
+  <para>
+   If both <parameter>start</parameter> and <parameter>end</parameter> are
+   <type>string</type>s, and <parameter>step</parameter> is <type>int</type>
+   the produced array will be a sequence of characters.
```
TimWolla, line 81: Characters should be clarified as either “(latin) letters” or “bytes”.

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
+  <para>
+   Otherwise, the produced array will be a sequence of <type>int</type>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   A <classname>ValueError</classname> is thrown if
+   <parameter>step</parameter> is <literal>0</literal>.
```
TimWolla, line 97: For ext/random I inversed the order, because the error condition is more important than the type of error: “If step is 0, a ValueError will be thrown”.

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
+  </para>
+  <para>
+   A <classname>ValueError</classname> is thrown if the value of
+   <parameter>start</parameter>, <parameter>end</parameter>,
+   or, as of PHP 8.3.0, <parameter>step</parameter> is not finite.
+   I.e. the value is infinity or <constant>NAN</constant>.
```
TimWolla, line 103: Finite also appears in the documentation for Randomizer::getFloat(). Should we create some documentation we can just reference for the definition?

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
+  </para>
+  <para>
+   As of PHP 8.3.0, a <classname>ValueError</classname> is thrown if the
```
TimWolla, line 106: Is it necessary to mention the version here? It should become clear from the changelog and IMO the documentation should document the current state, with the differences being resolved using the changelog.

TimWolla (commented):
Some remarks. Will probably require a second pass depending on the changes you do.

reference/array/functions/range.xml
Girgias: Yes end may not be in the range e.g. ``range(0, 5, 2)``

reference/array/functions/range.xml
Girgias: That might make sense to add to the language/types/float page, or link to is_finite?

reference/array/functions/range.xml
Girgias: Probably not

reference/array/functions/range.xml
Girgias: I also don't like using the ``[start, end]`` notation as it could be confused as an array declaration.

reference/array/functions/range.xml
TimWolla: I'd put a link to `is_finite`. I'll put onto my ToDo list that that doc page needs to be rewritten to clarify what `is_finite` means.

reference/array/functions/range.xml
```diff
@@ -33,32 +44,179 @@
...
   <para>
-   Returns an array of elements from <parameter>start</parameter> to
-   <parameter>end</parameter>, inclusive.
+   Returns a sequence of elements as an <type>array</type> with the first
+   element being <parameter>start</parameter> going up to
+   <parameter>end</parameter>, with each values of the sequence being
+   <parameter>step</parameter> values apart.
+  </para>
+  <para>
+   <parameter>end</parameter> may be the last element of the returned array.
```
TimWolla, line 80: Perhaps:

```suggestion
   The last element of the returned array is either <parameter>end</parameter> or the previous element of the sequence, depending on the value of <parameter>step</parameter>
```

This spells out more explicitly that $end may or may not be included.

TimWolla (commented):
LGTM once the three remaining remarks are resolved.

reference/array/functions/range.xml
TimWolla: I've replaced this with an updated suggestion, thus resolving.

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
+   If <parameter>step</parameter> is negative,
+   but the produced range is increasing,
+   a <classname>ValueError</classname> is thrown.
+   I.e. <code>$start &lt;= $end</code>.
+  </para>
+  <para>
+   If <parameter>start</parameter> or <parameter>end</parameter>
+   is the empty string <literal>''</literal>,
+   an <constant>E_WARNING</constant> is emitted and
+   The empty string will be interpreted as <literal>0</literal>.
```
TimWolla, line 121:
```suggestion
   the empty string will be interpreted as <literal>0</literal>.
```

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
+  <para>
+   If <parameter>start</parameter>, <parameter>end</parameter>,
+   or <parameter>step</parameter> is not <function>is_finite</function>,
+   a <classname>ValueError</classname> is thrown.
+  </para>
+  <para>
+   If <parameter>step</parameter> is negative,
+   but the produced range is increasing,
+   a <classname>ValueError</classname> is thrown.
+   I.e. <code>$start &lt;= $end</code>.
```
TimWolla, line 115:
```suggestion
   but the produced range is increasing (i.e. <code>$start &lt;= $end</code>),
   a <classname>ValueError</classname> is thrown.
```

reference/array/functions/range.xml
```diff
@@ -33,32 +44,182 @@
...
+   the produced array will be a sequence of <type>float</type>.
+  </para>
+  <para>
+   Otherwise, the produced array will be a sequence of <type>int</type>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
```
TimWolla, line 102: For ext/random the list of errors is an `<itemizedlist>`. Not sure where I got this from, but it looks nice to me.

reference/array/functions/range.xml
```diff
@@ -16,7 +15,19 @@
   <para>
    Create an array containing a range of elements.
   </para>
+  <para>
+   If both <parameter>start</parameter> and <parameter>end</parameter> are
+   <type>string</type>s, and <parameter>step</parameter> is <type>int</type>
+   the produced array will be a sequence of characters.
```
TimWolla, line 21:
```suggestion
   the produced array will be a sequence of bytes.
```

TimWolla (commented):
I'Ve now had a look at the rendered version. In addition to the new review marks, consider changing the example to:

```
<?php
echo implode(', ', range(0, 12)), PHP_EOL;

echo implode(', ', range(0, 100, 10)), PHP_EOL;

echo implode(', ', range('a', 'i')), PHP_EOL;

echo implode(', ', range('c', 'a')), PHP_EOL;

echo implode(', ', range('A', 'z')), PHP_EOL;
?>
```

with the output:

```
0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12
0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100
a, b, c, d, e, f, g, h, i
c, b, a
A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X, Y, Z, [, \, ], ^, _, `, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y, z
```

The current example lacks the output entirely is is unncessarily verbose with the loops. The new example also showcases non-letter bytes.

reference/array/functions/range.xml
```diff
@@ -16,7 +15,19 @@
...
   </para>
+  <para>
+   If both <parameter>start</parameter> and <parameter>end</parameter> are
+   <type>string</type>s, and <parameter>step</parameter> is <type>int</type>
+   the produced array will be a sequence of bytes.
+   Otherwise, the produced array will be a sequence of numbers.
+  </para>
+  <para>
+   The sequence is increasing if <parameter>start</parameter> is less
+   or equal than <parameter>end</parameter>.
```
TimWolla, line 26:
```suggestion
   The sequence is increasing if <parameter>start</parameter> is less than
   equal to <parameter>end</parameter>.
```

reference/array/functions/range.xml
```diff
@@ -33,32 +44,198 @@
...
+
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an array of elements from <parameter>start</parameter> to
-   <parameter>end</parameter>, inclusive.
+   Returns a sequence of elements as an <type>array</type> with the first
+   element being <parameter>start</parameter> going up to
+   <parameter>end</parameter>, with each values of the sequence being
+   <parameter>step</parameter> values apart.
```
TimWolla, line 77:
```suggestion
   <parameter>end</parameter>, with each value of the sequence being
   <parameter>step</parameter> values apart.
```

TimWolla:
I've now merged this, it's not getting better when it sits around for longer.

PR #2806, opened 2023-09-27
Explain error control operator set `error_level` to bit-expression

imme-emosol:


imme-emosol:
The verbose wording aims to prevent misinterpretation of `|` as "or".

An other/extra option could be to change the `<literal>` to make it have a different background for the whole expression, to signal to readers that it is an expression, not some sort of listing.

Girgias:
Please rebase this as the operators page got split into different pages.

Girgias (approved):

Girgias:
Thank you!

PR #2810, opened 2023-09-29
Update transformToDoc description

nielsdos:
Mainly, this is to fix https://bugs.php.net/bug.php?id=53693, which is about the undocumented returnClass argument.
However, I found that the description on this page was in general suboptimal, so this patch fixes that.

### DIFF
--- a/reference/xsl/xsltprocessor/transformtodoc.xml
+++ b/reference/xsl/xsltprocessor/transformtodoc.xml
@@ -3,17 +3,17 @@
 <refentry xml:id="xsltprocessor.transformtodoc" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>XSLTProcessor::transformToDoc</refname>
-  <refpurpose>Transform to a DOMDocument</refpurpose>
+  <refpurpose>Transform to a document</refpurpose>
  </refnamediv>
  <refsect1 role="description">
   &reftitle.description;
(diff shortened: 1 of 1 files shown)


reference/xsl/xsltprocessor/transformtodoc.xml
```diff
@@ -26,8 +26,17 @@
...
+       The <classname>DOMDocument</classname> or <classname>SimpleXMLElement</classname> or libxml-compatible
+       object to be transformed.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>returnClass</parameter></term>
+     <listitem>
+      <para>
+       This optional parameter may be used so that <function>transformToDoc</function> will return an object of the specified class.
```
Girgias, line 38:
```suggestion
       This optional parameter may be used so that
       <methodname>XSLTProcessor::transformToDoc</methodname>
       will return an object of the specified class.
```

Girgias (approved):
Nit for the markup of the method name, otherwise LGTM

reference/xsl/xsltprocessor/transformtodoc.xml
nielsdos: Change seems okay, thanks for this suggestion

Girgias (approved):

PR #2811, opened 2023-09-29
Cleanup options parameter description

nielsdos:
PHP 7.0 requires at least libxml2 2.6.11. So any note about an older version is simply irrelevant.

Use an entity to use consistent, more useful wording.

This patch also gets rid of some personification.


Girgias (approved):

Girgias:
Thank you!

PR #2812, opened 2023-09-29
PHP 8.3 gen_stub.php --replace-classsynopses

Girgias:
Created by ``./build/gen_stub.php --replace-classsynopses ./ ./../docs-php/en/`` with some omissions due to false positives in DateTime.

### DIFF
--- a/reference/gmp/gmp.xml
+++ b/reference/gmp/gmp.xml
@@ -30,6 +30,9 @@
     </ooclass>
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.gmp')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[@role='GMP'])">
+     <xi:fallback/>
+    </xi:include>
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.gmp')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[@role='GMP'])">
      <xi:fallback/>
     </xi:include>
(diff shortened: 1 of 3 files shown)


Girgias:
So GMP doesn't actually document the constructor _yay_

nielsdos (approved):
OK for ext/dom, thx

reference/dom/domelement.xml
```diff
@@ -47,6 +47,16 @@
      <type>string</type>
      <varname linkend="domelement.props.tagname">tagName</varname>
     </fieldsynopsis>
+    <fieldsynopsis>
+     <modifier>public</modifier>
+     <type>string</type>
+     <varname linkend="domelement.props.classname">className</varname>
```
kocsismate, line 53: the `domelement.props.classname` description is needed (the same issue below)

kocsismate:
> So GMP doesn't actually document the constructor yay

Yeah, because the constructor was only added recently by Niels :)

reference/dom/domelement.xml
Girgias: I just ran ``--regenerate-class-synopses``, would be nice if that also would generate the description section :(

reference/gmp/gmp.xml
```diff
@@ -30,6 +30,9 @@
     </ooclass>
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.gmp')/db:refentry/db:refsect1[@role='description']/descendant::db:constructorsynopsis[@role='GMP'])">
```
kocsismate, line 33: The `constructorsynopsis` page should be added in order for it to work.

kocsismate:
Relevant "fix" for the ext/date false positive diff: https://github.com/php/php-src/pull/12566

reference/gmp/gmp.xml
Girgias: I know I thought I had merged my GMP changes already

PR #2813, opened 2023-09-30
dom: Add stubs generated from php-src stubs

Girgias:
Generate using https://github.com/php/php-src/pull/12330 and running the following command (for my folder hierarchy):

```shell
../php-src/build/gen_stub.php --generate-methodsynopses ../PHP-8.3/ext/dom/ ../docs-php/en/reference/dom/
```


reference/dom/domcharacterdata/after.xml
```diff
@@ -0,0 +1,91 @@
...
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist><varlistentry>
+     <term><parameter>nodes</parameter></term>
+     <listitem>
+      <para>
+       Description.
```
kocsismate, line 27: Wouldn't it be better to leave these empty instead?

Girgias: Well, ideally, people would work on them directly after generating the stubs. And if no text is generated, one gets an empty tag, which I don't think is great.

nielsdos:
A question.
The after() method on DOMElement etc come from the interface DOMChildNode. Looking at that page I see after() listed there: https://www.php.net/manual/en/class.domchildnode.php
Clicking on after() reveals its documentation, albeit a bit lackluster. It's also hard to discover these methods right now.

The after() method (and others coming from the interface) is not listed on https://www.php.net/manual/en/class.domelement.php
So how do we go about it? I guess the description should be the same for all of them. It's a bit weird that the interface has a page tbh?

Girgias:
Oh so that's where they are documented.

I mean one solution is to add:
```
    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.domnode')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[@role='DOMChildNode'])">
     <xi:fallback/>
    </xi:include>
```
In the inherited method section, maybe having a look at how countable classes do it? As Countable::count also has a page

TimWolla:
> It's also hard to discover these methods right now.

Indeed. I've ran into this exact issue myself before: https://chat.stackoverflow.com/transcript/message/56491752#56491752

reference/dom/domelement/before.xml
```diff
@@ -0,0 +1,83 @@
...
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <simplelist>
+   <member><methodname>ClassName::otherMethodName</methodname></member>
+   <member><function>some_function</function></member>
+   <!--<member><link linkend="some.id.chunk.to.link">something appendix</link></member>-->
+  </simplelist>
+ </refsect1>
```
Girgias, line 61: Possibly?
```suggestion
 <refsect1 role="seealso">
  &reftitle.seealso;
  <simplelist>
   <member><methodname>DOMChildNode::before</methodname></member>
   <member><methodname>DOMElement::after</methodname></member>
  </simplelist>
 </refsect1>
```

reference/dom/domdocumentfragment/append.xml
```diff
@@ -0,0 +1,80 @@
...
+ <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('domparentnode.append')/db:refsect1[@role='returnvalues'])" />
+ <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('domparentnode.append')/db:refsect1[@role='errors'])" />
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <example xml:id="domdocumentfragment.append.example.basic">
+   <title><methodname>DOMDocumentFragment::append</methodname> example</title>
+   <para>
+    Description.
+   </para>
```
Girgias, line 29: Seems this was not filled in/removed. :)

nielsdos:
These still remain and I'll try to finish them soon:
- replaceWith
- replaceChildren
- getIterator

nielsdos:
Looks like I'm getting some errors because of the xincludes:

```
File /home/runner/work/doc-en/doc-en/en/reference/dom/domcharacterdata/after.xml No parameters sections.
File /home/runner/work/doc-en/doc-en/en/reference/dom/domcharacterdata/after.xml No returnvalues sections.
```

are those checked before resolving the xincludes by any chance?

Girgias:
> Looks like I'm getting some errors because of the xincludes:
> 
> ```
> File /home/runner/work/doc-en/doc-en/en/reference/dom/domcharacterdata/after.xml No parameters sections.
> File /home/runner/work/doc-en/doc-en/en/reference/dom/domcharacterdata/after.xml No returnvalues sections.
> ```
> 
> are those checked before resolving the xincludes by any chance?

Yes, they don't actually work on the manual but individual files. Just go and add exceptions to the sections_check QA script. As it probably should be written to work on the full manual.

nielsdos:
I think this is ready for review now.
I submitted https://github.com/php/doc-base/pull/106 to fix the integration.

reference/dom/domchildnode/after.xml
```diff
@@ -39,6 +40,33 @@
...
+  &reftitle.errors;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><constant>DOM_HIERARCHY_REQUEST_ERR</constant></term>
+     <listitem>
+      <para>
+       Raised if the parent is of a type that does not allow children of the
+       type of one of the passed <parameter>nodes</parameter>, or if the node to
+       put in is one of this node's ancestors or this node itself.
```
Girgias, line 53: This might make sense to either use an entity in language snippets, or possibly add a file to XInclude with an URL to a local file? (maybe for translations one want's a fallback that traverses up to directory structure then go down with ``../en/reference/dom/``

reference/dom/domdocument/adoptnode.xml
```diff
@@ -0,0 +1,121 @@
...
+     <varlistentry>
+      <term><constant>DOM_NOT_SUPPORTED_ERR</constant></term>
+      <listitem>
+       <para>
+        Raised if the node type is not supported for document transfers.
+       </para>
+      </listitem>
+     </varlistentry>
+    </variablelist>
+   </para>
```
Girgias, line 55: The para tag is unnecessary (yes I know parameters lists use them...)

Girgias (commented):
LGTM except the formatting of the error sections which I am not sure.

``variablelist`` feels inappropriate, I usually think if doing it that way we use ``itemizedlist`` but DocBook seems to be OK with using it that way.

Now the wrapping para tag is unnecessary and should be removed.

The last thin is that it would be nice to have the descriptions for the errors somehow only written once and included so that they don't go out of sync.

reference/dom/domchildnode/after.xml
TimWolla: For ext/random, I have this: https://github.com/php/doc-en/blob/32c8df6e0abb023bd332d0ac0dc94631192239af/reference/random/random/randomizer/getint.xml#L64

nielsdos: Fixed via entity

Girgias (commented):
LGTM! Thank you!

PR #2816, opened 2023-10-01
deleted non-existent mb_strtotitle description

mumumu:
It seems that `mb_strtotitle` function does not exist in the latest php-src, and the past implementation.


Girgias (approved):

Girgias:
I think it has never existed

PR #2817, opened 2023-10-01
Clarified the meaning of soft deprecation

mumumu:
refs: [PEP387#soft-deprecation](https://peps.python.org/pep-0387/#soft-deprecation)

### DIFF
--- a/appendices/migration83/deprecated.xml
+++ b/appendices/migration83/deprecated.xml
@@ -13,7 +13,7 @@
     operator (<literal>++</literal>) on empty, non-numeric,
     or non-alphanumeric strings is now deprecated.
     Moreover, incrementing non-numeric strings is soft deprecated.
-    Meaning no <constant>E_DEPRECATED</constant> diagnostic is emitted,
+    That means no <constant>E_DEPRECATED</constant> diagnostic is emitted,
     but this feature should not be used when producing new code.
     The new <function>str_increment</function> function should be used instead.
    </para>
--- a/appendices/migration83/deprecated.xml
+++ b/appendices/migration83/deprecated.xml
@@ -12,8 +12,10 @@
     Using the <link linkend="language.operators.increment">increment</link>
     operator (<literal>++</literal>) on empty, non-numeric,
     or non-alphanumeric strings is now deprecated.
-    Moreover, incrementing non-numeric strings is soft deprecated and the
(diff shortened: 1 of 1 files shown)


appendices/migration83/deprecated.xml
```diff
@@ -12,8 +12,10 @@
     Using the <link linkend="language.operators.increment">increment</link>
     operator (<literal>++</literal>) on empty, non-numeric,
     or non-alphanumeric strings is now deprecated.
-    Moreover, incrementing non-numeric strings is soft deprecated and the
-    new <function>str_increment</function> function should be used instead.
+    Moreover, incrementing non-numeric strings is soft deprecated (no longer be
+    used to write new code, but it remains safe to continue using it in existing
+    code) and the new <function>str_increment</function> function should be used
+    instead.
```
Girgias, line 18:
```suggestion
    Moreover, incrementing non-numeric strings is soft deprecated.
    Meaning no <constant>E_DEPRECATED</constant> diagnostic is emitted,
    but this feature should not be used when producing new code.
    The new <function>str_increment</function> function should be used instead.
```

Possibly reads better? @Crell @TimWolla might be good to have your opinion

TimWolla: Perhaps “writing” rather than “producing”, other than that your suggestion looks good to me.

Crell: "Meaning" should be "That means", since it begins a new sentence.  Writing or producing are both fine for me, but otherwise I like these suggestions.

appendices/migration83/deprecated.xml
```diff
@@ -12,8 +12,10 @@
     Using the <link linkend="language.operators.increment">increment</link>
     operator (<literal>++</literal>) on empty, non-numeric,
     or non-alphanumeric strings is now deprecated.
-    Moreover, incrementing non-numeric strings is soft deprecated and the
-    new <function>str_increment</function> function should be used instead.
+    Moreover, incrementing non-numeric strings is soft deprecated.
+    Meaning no <constant>E_DEPRECATED</constant> diagnostic is emitted,
```
TimWolla, line 16:
```suggestion
    That means no <constant>E_DEPRECATED</constant> diagnostic is emitted,
```
As per Crell's comment.

Crell (approved):

PR #2841, opened 2023-10-05
PHP 8.3: Describe increment/decrement operator changes

Girgias:


language/operators/increment.xml
```diff
@@ -47,48 +39,116 @@
    </tbody>
   </tgroup>
  </table>
+
+ <warning>
+  <para>
+   The increment and decrement operators have no effect on values
+   of type <type>bool</type>.
+   A <constant>E_WARNING</constant> is emitted, as of PHP 8.3.0,
```
TimWolla, line 47:
```suggestion
   A <constant>E_WARNING</constant> is emitted as of PHP 8.3.0,
```
The comma looks wrong (also applies to the other warnings).

language/operators/increment.xml
```diff
@@ -47,48 +39,116 @@
...
+   because this will implicitly cast the value to <type>int</type> in the future.
+  </para>
+ </warning>
+ <warning>
```
TimWolla, line 51: The rendering of this is not exactly pretty. Maybe merge them into a singular warning block with multiple paragraphs, because they are so similar?

language/operators/increment.xml
```diff
@@ -47,48 +39,116 @@
    </tbody>
   </tgroup>
  </table>
+
+ <warning>
```
TimWolla, line 43: I'd move this below the examples. Folks should see how the operator works in general, because telling them various cases where it won't work.

TimWolla (commented):
LGTM overall.

TimWolla (approved):

PR #2843, opened 2023-10-05
Fix error

msamesch:
A warning had been issued previously.


derickr:
This change is not OK. The original was correct. The PR seems to just remove the "before" section without any reason as to why.

msamesch:
The changelog says "Now throws       <exceptionname>DateMalformedStringException</exceptionname> if an invalid string is passed, instead of a warning."

derickr:
Ok. I missed that. I'll investigate early next week.

PR #2844, opened 2023-10-06
Stringable can't be passed to argument expecting string, when strict types enabled

NickSdot:
A method that expects a `string` type argument will not allow a `stringable` object to be passed. This behaviour surprised me today and recently someone else here https://github.com/php/php-src/issues/12097

I tend to think that PHP should have an exception here, but I also understand the arguments against it. So, since the behaviour is what it is, I would like to propose making it clearer in the docs.

### DIFF
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -273,7 +273,9 @@ class Connection
      As of PHP 8.0.0, the return value follows standard PHP type semantics,
      meaning it will be coerced into a <type>string</type> if possible and if
      <link linkend="language.types.declarations.strict">strict typing</link>
-     is disabled.
+     is disabled. If strict typing is enabled, a stringable object cannot be 
+     passed to a method that expects only a string, the argument must be 
+     typed as <interfacename>Stringable</interfacename>.
     </para>
     <para>
      As of PHP 8.0.0, any class that contains a <link linkend="object.tostring">__toString()</link>
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -273,9 +273,14 @@ class Connection
      As of PHP 8.0.0, the return value follows standard PHP type semantics,
      meaning it will be coerced into a <type>string</type> if possible and if
      <link linkend="language.types.declarations.strict">strict typing</link>
-     is disabled. If strict typing is enabled, a stringable object cannot be 
-     passed to a method that expects only a string, the argument must be 
-     typed as <interfacename>Stringable</interfacename>.
+     is disabled.
+    </para>
+    <para>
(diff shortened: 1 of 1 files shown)


Girgias:
This is already explained somewhat in https://www.php.net/manual/en/language.types.type-juggling.php#language.types.type-juggling.function and https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict

And the solution to this is to not use strict typing, as my biased opinion is that strict types was a mistake in the first place.

NickSdot:
Hey @Girgias, thanks for your answer. 

I believe the links you provided achieve not the same as my proposed PR. For two reasons:

a) it is not where I would expect it to be.

b) it does not cover the specific case, which I personally, consider a special becauae we are dealing with a magic functionality.

`_toString()` docs say:

`The __toString() method allows a class to decide how it will react when it is treated like a string.`

https://www.php.net/manual/en/language.oop5.magic.php#object.tostring

If I pass a Stringable object to a method attribute typed as a string, it is - from my point of view - treated as a string. But it doesn't behave as expected.

Just my two cents. If you see it differently, feel free to close my PR.

language/oop5/magic.xml
```diff
@@ -273,7 +273,9 @@ class Connection
      As of PHP 8.0.0, the return value follows standard PHP type semantics,
      meaning it will be coerced into a <type>string</type> if possible and if
      <link linkend="language.types.declarations.strict">strict typing</link>
-     is disabled.
+     is disabled. If strict typing is enabled, a stringable object cannot be 
+     passed to a method that expects only a string, the argument must be 
+     typed as <interfacename>Stringable</interfacename>.
     </para>
     <para>
      As of PHP 8.0.0, any class that contains a <link linkend="object.tostring">__toString()</link>
```
Girgias, line 281: The issue is that this paragraph talks about the ***return*** value of the ``__toString()`` method, not about using it.

At best, it should be in another paragraph, but I'm unconvinced this is the right place.
```suggestion
     is disabled.
    </para>
    <para>
     A <interfacename>Stringable</interfacename> object will
     <emphasis>not</emphasis> be accepted by a <type>string</type> type declaration if
     <link linkend="language.types.declarations.strict">strict typing</link>
     is enabled. If such behaviour is wanted the type declaration must accept
     <interfacename>Stringable</interfacename> and <type>string</type> via a union type.
    </para>
    <para>
     As of PHP 8.0.0, any class that contains a <link linkend="object.tostring">__toString()</link>
```

NickSdot:
@Girgias changed as requested.

Girgias:
Thank you!

PR #2846, opened 2023-10-06
is_(finite|infinite|nan): Rewrite documentation

TimWolla:


reference/math/functions/is-finite.xml
```diff
@@ -24,21 +29,55 @@
...
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if <parameter>num</parameter> is a legal finite
-   number within the allowed range for a PHP float on this platform,
+   &true; if <parameter>num</parameter> is none of <literal>NAN</literal>, <literal>INF</literal>, <literal>-INF</literal>,
    else &false;.
   </para>
  </refsect1>
+ <refsect1 role="examples">
```
Girgias, line 47: Nit: as you added line breaks between all the other sections
```suggestion
 </refsect1>

 <refsect1 role="examples">
```

reference/math/functions/is-finite.xml
```diff
@@ -24,21 +29,55 @@
...
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>is_nan</function> example</title>
```
Girgias, line 51:
```suggestion
    <title><function>is_finite</function> example</title>
```

reference/math/functions/is-infinite.xml
```diff
@@ -25,19 +26,52 @@
...
+
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if <parameter>num</parameter> is infinite, else &false;.
+   &true; if <parameter>num</parameter> is either <literal>INF</literal>
+   or <literal>-INF</literal>, else &false;.
   </para>
  </refsect1>
+ <refsect1 role="examples">
```
Girgias, line 44: Nit: ditto
```suggestion
 </refsect1>

 <refsect1 role="examples">
```

reference/math/functions/is-infinite.xml
```diff
@@ -25,19 +26,52 @@
...
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>is_nan</function> example</title>
```
Girgias, line 48:
```suggestion
    <title><function>is_infinite</function> example</title>
```

reference/math/functions/is-nan.xml
```diff
@@ -13,9 +13,28 @@
...
+   Returns whether the given <parameter>num</parameter> is <literal>NAN</literal> (Not A Number).
   </para>
+  <para>
+   <literal>NAN</literal> is returned from mathematical operations that are undefined,
+   for example when passing parameters outside of function’s input domain. The square root
+   (<function>sqrt</function>) is only defined for positive numbers, passing a negative number
+   will result in <literal>NAN</literal>. Other examples of operations returning <literal>NAN</literal>
+   are dividing <literal>INF</literal> by <literal>INF</literal> and any operation involving an
+   existing <literal>NAN</literal> value.
+  </para>
```
Girgias, line 25: TIL we don't check the input value of sqrt (and a lot of other maths function but we do for ``log()`` (and not even the other variants.

Might make sense to RFC this to add warnings.

fdiv with dividing by 0 would be a sure way to always return NAN

reference/math/functions/is-nan.xml
```diff
@@ -13,9 +13,28 @@
...
+  <note>
+   <para>
+    Despite its name of “Not A Number”, <literal>NAN</literal> is a valid value of type &float;.
```
Girgias, line 28: Possibly use the ``<quote>`` tag around "not a number" instead?

Girgias (changes requested):

reference/math/functions/is-nan.xml
TimWolla: > but we do for log()

`log()` for a negative input also returns NAN for me.

> fdiv with dividing by 0 would be a sure way to always return NAN

fdiv with divisor 0 is INF (unless you're dividing zero by zero, then its NAN). I tested with the regular `/` operator, but that threw, so I searched for another example :grin:

Girgias: > > but we do for log()
> 
> `log()` for a negative input also returns NAN for me.
> 
> > fdiv with dividing by 0 would be a sure way to always return NAN
> 
> fdiv with divisor 0 is INF (unless you're dividing zero by zero, then its NAN). I tested with the regular `/` operator, but that threw, so I searched for another example 😁

Yeah ``fdiv(0, 0)``, fdiv got introduced after the decision to make division by 0 throw an exception in 8.0.

And for log() I read the code too quickly, we only check the value of the base arg.
But still might be a good idea to add warnings dunno.

Girgias (approved):

Girgias:
Thank you!

reference/math/functions/is-nan.xml
TimWolla: > But still might be a good idea to add warnings dunno.

To completely eliminate NAN you would need to perform NAN checking for almost every mathematical operation involving floats. This is probably unacceptably slow. The behavior of NAN is well-defined in the context of floats.

PR #2849, opened 2023-10-09
array-change-key-case exception if not array

dreeh:
array-change-key-case exception if not array

### DIFF
--- a/reference/array/functions/array-change-key-case.xml
+++ b/reference/array/functions/array-change-key-case.xml
@@ -50,11 +50,18 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns an array with its keys lower or uppercased, or &null; if
-   <parameter>array</parameter> is not an array.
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-change-key-case.xml
```diff
@@ -50,11 +50,18 @@
...
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   If <parameter>array</parameter> is not an array,
+   a <classname>TypeError</classname> will be thrown.
+  </para>
+ </refsect1>
```
Girgias, line 63: Please remove this, we do not mention ZPP violations. (also, this is inaccurate for 7.X)

Girgias (changes requested):

reference/array/functions/array-change-key-case.xml
dreeh: all right

Girgias:
Thank you!

PR #2850, opened 2023-10-11
copy a modified scoping note from include to eval

chx:


Girgias (approved):

reference/misc/functions/eval.xml
```diff
@@ -15,6 +15,16 @@
...
   </para>
+  <para>
+   The code being evaluated inherits the
+   <link linkend="language.variables.scope">variable scope</link> of the
+   line on which the <function>eval</function> call occurs. Any variables
+   available at that line will be available for reading and modification in
+   the evaluated code. However, all functions and classes defined will be
+   defined in the global namespace. In other words, the compiler considers
+   the evaluated code as if it were a separate
+   <link linkend="function.include">included</link> file.
```
Girgias, line 26:
```suggestion
   The code being evaluated inherits the
   <link linkend="language.variables.scope">variable scope</link>
   of the line on which the <function>eval</function> call occurs.
   Any variables available at that line will be available for reading and
   modification in the evaluated code.
   However, all functions and classes defined will be defined in the global namespace.
   In other words, the compiler considers the evaluated code as if it were a
   separate <link linkend="function.include">included</link> file.
```

PR #2851, opened 2023-10-12
Fix example of get_headers() usage with context

vkhonin:
Function get_headers() called with wrong signature in example of usage with context. Add missing $associative argument.


jhdxr (approved):
LGTM

Girgias (approved):

PR #2856, opened 2023-10-14
Randomizer::getFloat(): Add documentation

TimWolla:
Some first words explaining the problems of selecting a random float within a given interval and how `getFloat()` does it correctly. I plan to expand the explanation a little more in a second iteration, but wanted to put up for review what I already have to make sure that it is any good and not completely incomprehensible before spending more time on it. It was hard enough to write these words :grin:

### DIFF
--- a/reference/random/random/randomizer/getfloat.xml
+++ b/reference/random/random/randomizer/getfloat.xml
@@ -34,16 +34,17 @@
    <methodname>Random\Randomizer::getFloat</methodname> implements an algorithm that will
    return a uniformly selected float from the largest possible set of exactly representable
    and equidistributed floats within the requested interval. The distance between the selectable
-   floats matches the distance between the floats with the lowest density, i.e. the distance
-   between floats for the interval boundary with the larger absolute value. This means that
+   floats (“step size”) matches the distance between the floats with the lowest density, i.e. the distance
+   between floats at interval boundary with the larger absolute value. This means that
    not all representable floats within a given interval may be returned if the interval crosses
-   a power of two.
+   one or more powers of two. Stepping will start from the interval boundary with the larger absolute value
+   to ensure the steps align with the exactly representable floats.
   </para>
   <para>
-   The interval boundaries will always be included in the set of selectable floats. Thus, if
-   the size of the interval is not an exact multiple of the distance between each float,
-   the distance between <emphasis>one</emphasis> of the boundary values to its nearest value
-   will be smaller than the distance between all the other values.
+   Closed interval boundaries will always be included in the set of selectable floats. Thus,
+   if the size of the interval is not an exact multiple of the step size
+   and the boundary with the smaller absolute value is a closed boundary, the distance between
+   that boundary and its nearest selectable float will be smaller than the step size.
   </para>
 
   <caution>
--- a/reference/random/random/randomizer/getfloat.xml
+++ b/reference/random/random/randomizer/getfloat.xml
@@ -19,7 +19,7 @@
   <para>
    Due to the limited precision, not all real numbers can be exactly represented as
    a floating point number. If a number cannot be represented exactly, it is rounded to
-   the nearest representable exact value. Furthermore float are not equally dense across
+   the nearest representable exact value. Furthermore, floats are not equally dense across
    the whole number line. At each power of two, the distance between two neighboring
    floats doubles. In other words: There are the same number of floats between 1 and 2
    as they are between 2 and 4, 4 and 8, 8 and 16, and so on.
@@ -40,7 +40,7 @@
    a power of two.
   </para>
   <para>
-   The interval boundaries will always be included in the set of selectable floats. Thus if
+   The interval boundaries will always be included in the set of selectable floats. Thus, if
    the size of the interval is not an exact multiple of the distance between each float,
    the distance between <emphasis>one</emphasis> of the boundary values to its nearest value
    will be smaller than the distance between all the other values.
--- a/reference/random/random/randomizer/getfloat.xml
+++ b/reference/random/random/randomizer/getfloat.xml
@@ -21,7 +21,7 @@
    a floating point number. If a number cannot be represented exactly, it is rounded to
    the nearest representable exact value. Furthermore, floats are not equally dense across
    the whole number line. At each power of two, the distance between two neighboring
-   floats doubles. In other words: There are the same number of floats between 1 and 2
+   floats doubles. In other words: There are the same number of representable floats between 1 and 2
    as they are between 2 and 4, 4 and 8, 8 and 16, and so on.
   </para>
   <para>
--- a/reference/random/random/randomizer/getfloat.xml
+++ b/reference/random/random/randomizer/getfloat.xml
@@ -14,11 +14,47 @@
    <methodparam choice="opt"><type>Random\IntervalBoundary</type><parameter>boundary</parameter><initializer>Random\IntervalBoundary::ClosedOpen</initializer></methodparam>
   </methodsynopsis>
   <para>
-
+   Returns a uniformly selected, equidistributed float from a requested interval.
+  </para>
+  <para>
+   Due to the limited precision, not all real numbers can be exactly represented as
+   a floating point number. If a number cannot be represented exactly, it is rounded to
+   the nearest representable exact value. Furthermore float are not equally dense across
+   the whole number line. At each power of two, the distance between two neighboring
+   floats doubles. In other words: There are the same number of floats between 1 and 2
+   as they are between 2 and 4, 4 and 8, 8 and 16, and so on.
+  </para>
+  <para>
+   Randomly sampling an arbitrary number within the requested interval, for example by
+   dividing two integers, might result in a biased distribution for this reason. The
+   necessary rounding will cause some floats to be returned more often than others,
+   especially around powers of two when the density of floats changes.
+  </para>
+  <para>
+   <methodname>Random\Randomizer::getFloat</methodname> implements an algorithm that will
+   return a uniformly selected float from the largest possible set of exactly representable
+   and equidistributed floats within the requested interval. The distance between the selectable
+   floats matches the distance between the floats with the lowest density, i.e. the distance
+   between floats for the interval boundary with the larger absolute value. This means that
+   not all representable floats within a given interval may be returned if the interval crosses
+   a power of two.
+  </para>
+  <para>
+   The interval boundaries will always be included in the set of selectable floats. Thus if
+   the size of the interval is not an exact multiple of the distance between each float,
+   the distance between <emphasis>one</emphasis> of the boundary values to its nearest value
+   will be smaller than the distance between all the other values.
   </para>
 
-  &warn.undocumented.func;
-
+  <caution>
+   <para>
+    Post-processing the returned floats is likely going to break the uniform equidistribution,
+    because the intermediate floats within a mathematical operation are experiencing implicit
+    rounding. The requested interval should match the desired interval as closely as possible
+    and rounding should only be performed as an explicit operation right before displaying the
+    selected number to a user.
+   </para>
+  </caution>
  </refsect1>
 
  <refsect1 role="parameters">
@@ -45,7 +81,7 @@
      <term><parameter>boundary</parameter></term>
      <listitem>
       <para>
-       Specifies if the interval boundaries are possible return values.
+       Specifies whether the interval boundaries are possible return values.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/random/random/randomizer/getfloat.xml
```diff
@@ -14,11 +14,47 @@
    <methodparam choice="opt"><type>Random\IntervalBoundary</type><parameter>boundary</parameter><initializer>Random\IntervalBoundary::ClosedOpen</initializer></methodparam>
   </methodsynopsis>
   <para>
-
+   Returns a uniformly selected, equidistributed float from a requested interval.
+  </para>
+  <para>
+   Due to the limited precision, not all real numbers can be exactly represented as
+   a floating point number. If a number cannot be represented exactly, it is rounded to
+   the nearest representable exact value. Furthermore float are not equally dense across
```
nielsdos, line 22:
```suggestion
   the nearest representable exact value. Furthermore, floats are not equally dense across
```
Also perhaps it is not entirely clear what dense means in this context for people unfamiliar, which is probably the target audience. I don't have a better suggestion though. You explain it below, but I don't think people realize you're still talking about density.

reference/random/random/randomizer/getfloat.xml
```diff
@@ -14,11 +14,47 @@
    <methodparam choice="opt"><type>Random\IntervalBoundary</type><parameter>boundary</parameter><initializer>Random\IntervalBoundary::ClosedOpen</initializer></methodparam>
   </methodsynopsis>
   <para>
-
+   Returns a uniformly selected, equidistributed float from a requested interval.
```
Crell, line 17:
```suggestion
   Returns a uniformly selected, equi-distributed float from a requested interval.
```

I don't know if "equidistributed" is actually a word on its own, but it certainly doesn't sound like it.  "Equally distributed" may be even better for the layman.

reference/random/random/randomizer/getfloat.xml
Crell, line 22:
```suggestion
   the nearest representable exact value. Furthermore, float are not equally dense across
```

reference/random/random/randomizer/getfloat.xml
```diff
@@ -14,11 +14,47 @@
...
+  </para>
+  <para>
+   Due to the limited precision, not all real numbers can be exactly represented as
+   a floating point number. If a number cannot be represented exactly, it is rounded to
+   the nearest representable exact value. Furthermore float are not equally dense across
+   the whole number line. At each power of two, the distance between two neighboring
+   floats doubles. In other words: There are the same number of floats between 1 and 2
+   as they are between 2 and 4, 4 and 8, 8 and 16, and so on.
```
Crell, line 25: Wait, what?  There are an infinite number of floats between any two values.  That's why there are an uncountably infinite number of floats.  I am admittedly not a mathematician, but this needs a citation because it's going to confuse the crap out of people.

reference/random/random/randomizer/getfloat.xml
```diff
@@ -14,11 +14,47 @@
...
+   <methodname>Random\Randomizer::getFloat</methodname> implements an algorithm that will
+   return a uniformly selected float from the largest possible set of exactly representable
+   and equidistributed floats within the requested interval. The distance between the selectable
+   floats matches the distance between the floats with the lowest density, i.e. the distance
+   between floats for the interval boundary with the larger absolute value. This means that
+   not all representable floats within a given interval may be returned if the interval crosses
+   a power of two.
+  </para>
+  <para>
+   The interval boundaries will always be included in the set of selectable floats. Thus if
```
Crell, line 43:
```suggestion
   The interval boundaries will always be included in the set of selectable floats. Thus, if
```

reference/random/random/randomizer/getfloat.xml
Crell: I feel like this whole section needs a preamble of some kind to explain "math doesn't work like you think it does, randomness is funky, floats are just plain weird, and they combine into a total mess.  Buckle up."  Or something. :smile:

reference/random/random/randomizer/getfloat.xml
Crell, line 43: Wait, is this true?  Isn't that the point of the `boundary` parameter, to control if that's the case?

reference/random/random/randomizer/getfloat.xml
Girgias: It is a word: https://en.wikipedia.org/wiki/Equidistributed_sequence

reference/random/random/randomizer/getfloat.xml
Girgias: Binary floating point maths is like this, we could reuse the fat warning in the ``float`` type docs, but then one can argue this should be anywhere were floating points are mentioned

Girgias: > Wait, what? There are an infinite number of floats between any two values. That's why there are an uncountably infinite number of floats. I am admittedly not a mathematician, but this needs a citation because it's going to confuse the crap out of people.

There are infinite amounts of ***real*** numbers between any two values, not floats. There are a finite number of floats, because they need to fit into 64 bits.

reference/random/random/randomizer/getfloat.xml
Girgias: Yeah this seems incorrect considering the return value description.

reference/random/random/randomizer/getfloat.xml
Crell: Ahhh...  Those (binary and float vs. real) are critical points that should be made explicit up front, because it's not at all obvious otherwise.

reference/random/random/randomizer/getfloat.xml
TimWolla: Already mentioned by Niels (together with a missing plural).

reference/random/random/randomizer/getfloat.xml
TimWolla: It's badly phrased. What's always included is the smallest and largest value that lies within the requested interval.

Let me give an example. Consider that between 1 and 2 all steps `0.1` are exactly representable and between 2 and 4 all steps `0.2` are exactly representable. Thus the exact values are 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2.0, 2.2, 2.4, 2.6, 2.8, 3.0, 3.2, 3.4, 3.6, 3.8, 4.0.

Now if I want to sample a random float within the ClosedOpen interval [1.5, 2.6) (i.e. including 1.5 until, but not including 2.6). The algorithm now needs to select the *larger* step size of `0.2` to ensure the equidistribution. This interval is not an exact multiple of 0.2 (its size is 1.1). Then the following happens:

1. It needs to start stepping from 2.6, not from 1.5, because otherwise it would be shifted by `0.1` once reaching 2.0.
2. It always includes the minimum and maximum within the interval.

Thus the selectable floats are: 2.4 (this is the largest float smaller than 2.6, which is not included), 2.2, 2.0, 1.8, 1.6. And finally 1.5 is also included, even though its distance to 1.6 is only 0.1.

Now writing this all out, I need to check the behavior of the OpenOpen boundaries, it could be that only the *Closed* edge values are guaranteed to always be included.

reference/random/random/randomizer/getfloat.xml
TimWolla: I don't think it is useful to include the full detail of how floats work here, but only the parts that are relevant to understand why randomly selecting a float is hard and only on a high level.

It would probably be useful to expand this page and reference it for further reading, though: https://www.php.net/manual/en/language.types.float.php

TimWolla: I'm totally accepting concrete phrasing suggestions here, though. Writing the introduction is always the hard part :sweat_smile:

reference/random/random/randomizer/getfloat.xml
Crell: This is an excellent explanation of both the boundaries and the equidistributed stuff, and should be included in the doc page almost verbatim.

TimWolla: > it could be that only the _Closed_ edge values are guaranteed to always be included.

Yes: Only the Closed edges will always be included. An extreme example would be the following:

```
<?php

use Random\Engine;
use Random\Randomizer;
use Random\IntervalBoundary;

final class TestCountingEngine64 implements Engine
{
    private int $count = 9007199254740980;

    public function generate(): string
    {
        return pack('P', $this->count++);
    }
}

function run_test(Randomizer $r, float $min, float $max, int $count) {
	printf("[%.17f, %.17f]\n", $min, $max);

	$prev = null;
	for ($i = 0; $i < $count; $i++) {
	$float = $r->getFloat($min, $max, IntervalBoundary::OpenOpen);
	printf(" %.17f", $float);
	if ($prev !== null) {
	printf(" (%+.17g)", ($float - $prev));

	if ($prev < $float) {
		printf("\nRepeat");
	}
	}
	printf("\n");

	$prev = $float;
	}
}

run_test(
	new Randomizer(new TestCountingEngine64()),
	1.99999999999999889,
	9007199254740991,
	15,
);
```

This prints:

```
[1.99999999999999889, 9007199254740991.00000000000000000]
 10.00000000000000000
 9.00000000000000000 (-1)
 8.00000000000000000 (-1)
 7.00000000000000000 (-1)
 6.00000000000000000 (-1)
 5.00000000000000000 (-1)
 4.00000000000000000 (-1)
 3.00000000000000000 (-1)
 2.00000000000000000 (-1)
 9007199254740990.00000000000000000 (+9007199254740988)
Repeat
 9007199254740989.00000000000000000 (-1)
 9007199254740988.00000000000000000 (-1)
 9007199254740987.00000000000000000 (-1)
 9007199254740986.00000000000000000 (-1)
 9007199254740985.00000000000000000 (-1)
```

With `ClosedOpen` instead of `OpenOpen` it prints:

```
[1.99999999999999889, 9007199254740991.00000000000000000]
 10.00000000000000000
 9.00000000000000000 (-1)
 8.00000000000000000 (-1)
 7.00000000000000000 (-1)
 6.00000000000000000 (-1)
 5.00000000000000000 (-1)
 4.00000000000000000 (-1)
 3.00000000000000000 (-1)
 2.00000000000000000 (-1)
 1.99999999999999889 (-1.1102230246251565e-15)
 9007199254740990.00000000000000000 (+9007199254740988)
Repeat
 9007199254740989.00000000000000000 (-1)
 9007199254740988.00000000000000000 (-1)
 9007199254740987.00000000000000000 (-1)
 9007199254740986.00000000000000000 (-1)
```

Note how the last step from 2.0 to 1.99999999999999889 is significantly smaller than all the other steps.

TimWolla: I've pushed an update with a hopefully clearer phrasing. I'll rewrap before merging to keep the diff of that commit small.

reference/random/random/randomizer/getfloat.xml
Girgias: > Ahhh... Those (binary and float vs. real) are critical points that should be made explicit up front, because it's not at all obvious otherwise.

I disagree, we are working with floating point numbers, which are always a mess. This is a thing one _must_ be aware of all the time when dealing with floating point numbers. They are not real numbers and thus have finite precision. In the same way that the ``int`` type has finite precision, even though there are an infinite number of integers.

Now, improving the float type page Tim linked is fine and adequate, and possibly even creating an XML entity to emphasis knowing about the problems with floating point numbers. But then this warning should probably displayed on more pages.

Girgias (approved):
LGTM, could you break up the lines at the ending of a sentence?

This makes it easier for translations.

TimWolla:
@Crell Do you have any further remarks for this initial version or is it looking good to you? I'll add the example explanation of the algorithm (https://github.com/php/doc-en/pull/2856#discussion_r1359909874) in a follow-up together with some other constraints (e.g. the fact that the boundaries need to be greater than 2^-1020 for it to work correctly).

reference/random/random/randomizer/getfloat.xml
```diff
@@ -14,11 +14,48 @@
...
   </methodsynopsis>
   <para>
-
+   Returns a uniformly selected, equidistributed float from a requested interval.
+  </para>
+  <para>
+   Due to the limited precision, not all real numbers can be exactly represented as
+   a floating point number. If a number cannot be represented exactly, it is rounded to
+   the nearest representable exact value. Furthermore, floats are not equally dense across
+   the whole number line. At each power of two, the distance between two neighboring
```
Crell, line 23:
```suggestion
   the whole number line. Because computers use binary, at each power of two the distance between two neighboring
```

To make it clear why the power of two issue matters.

reference/random/random/randomizer/getfloat.xml
Crell: The updated intro I think makes it sufficiently clear, so closing this thread.

Crell:
Left a few smaller clarifying suggestions, but otherwise this is pretty good now, I think.  I'd still like to see Tim's example comment above captured somewhere in the docs; if not here, then in some part of the float documentation.  It's a very lucid explanation of the issues involved.

reference/random/random/randomizer/getfloat.xml
TimWolla:
```suggestion
   the whole number line. Because floats use a binary exponent, the distance between two neighboring floats doubles at each power of two
```
It's not so much computers per se, but rather the binary exponent of the IEEE-754 double precision float. If floats would use, say, a binary coded decimal (https://en.wikipedia.org/wiki/Binary-coded_decimal), the behavior would be different.

This should be explained in more detail at the general float page, which I plan to rewrite time permitting.

TimWolla:
Thank you all. Now merged. I'll wait for the rendered docs to appear to see how they look in the live version and then I'll make the follow-ups.

PR #2863, opened 2023-10-16
assert: Fix return value explanation

TimWolla:
Specifically `assert()` will always return `true` if assertions are disabled. Ideally the return value would never be used anyway …

### DIFF
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -237,7 +237,7 @@
    <member>A custom exception object is passed to <parameter>description</parameter>.</member>
   </simplelist>
   <para>
-   If none of the conditions is true <function>assert</function> will return &true; if
+   If none of the conditions are true <function>assert</function> will return &true; if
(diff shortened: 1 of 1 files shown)


reference/info/functions/assert.xml
```diff
@@ -227,7 +227,18 @@
...
+  <simplelist>
+   <member><literal>zend.assertions=0</literal></member>
+   <member><literal>zend.assertions=-1</literal></member>
+   <member><literal>assert.exception=1</literal></member>
+   <member><literal>assert.bail=1</literal></member>
+   <member>A custom exception object is passed to <parameter>description</parameter>.</member>
+  </simplelist>
+  <para>
+   If none of the conditions is true <function>assert</function> will return &true; if
+   <parameter>assertion</parameter> is truthy and &false; otherwise.
```
Girgias, line 241:
```suggestion
   If none of the conditions are true <function>assert</function> will return &true; if
   <parameter>assertion</parameter> is truthy and &false; otherwise.
```

PR #2870, opened 2023-10-18
Randomizer::getFloat(): Add step-by-step explanation of γ-section

TimWolla:
This adds the explanation from https://github.com/php/doc-en/pull/2856#discussion_r1359906990 with different values: I opted to use binary steps (0.125 and 0.25) instead of decimal steps (0.1 and 0.2) to more closely match the reality. I've also explained how the affine transformation `(max - min) * x + min` fails. A caution note and example advising against this affine transformation already exists in the documentation of `nextFloat()`.


zeriyoshi (approved):

reference/random/random/randomizer/getfloat.xml
```diff
@@ -74,6 +78,114 @@
     displaying the selected number to a user.
    </para>
   </caution>
+  <refsect2>
+   <title>Explanation of the Algorithm Using Example Values</title>
+   <para>
+    To give an example of how the algorithm works, consider that between
+    <literal>1.0</literal> and <literal>2.0</literal> all steps of size <literal>0.125</literal>
+    are exactly representable and between <literal>2.0</literal> and <literal>4.0</literal>
```
Crell, line 86:
```suggestion
    are exactly representable in binary and between <literal>2.0</literal> and <literal>4.0</literal>
```

reference/random/random/randomizer/getfloat.xml
```diff
@@ -74,6 +78,114 @@
...
+    which is smaller than the previously determined step size of <literal>0.25</literal>.
+    The reason for that is that the requested interval is closed at the lower boundary
+    (<literal>1.625</literal>) and closed boundaries are always included.
+   </para>
+   <para>
+    Finally the algorithm uniformly selects one of the four selectable values at random
+    and returns it.
+   </para>
+   <para>
+    In this example, there are eight representable floating point numbers between each
```
Crell, line 150: This paragraph onward is describing a rejected algorithm.  First, I'm not clear if that is necessary or appropriate, since it's describing something PHP doesn't do.  If it is kept, though, it should be clearly set aside with its own section header to make it clear that it's not what PHP does; I was 2/3 of the way through it before I realized it was describing a bad approach, not an alternate, equally valid approach.

reference/random/random/randomizer/getfloat.xml
```diff
@@ -74,6 +78,114 @@
...
+     <member><literal>2.5</literal></member>
+     <member><literal>2.75</literal></member>
+     <member><literal>3.0</literal></member>
+     <member><literal>3.25</literal></member>
+     <member><literal>3.5</literal></member>
+     <member><literal>3.75</literal></member>
+     <member><literal>4.0</literal></member>
+    </simplelist>
+    are the exactly representable floats between <literal>1.0</literal>
+    and <literal>4.0</literal>.
```
Crell, line 110:
```suggestion
    and <literal>4.0</literal>.  Note that there are twice as many that start with 1 as start with 2 or 3, so selecting randomly from this list would bias toward values between 1 and 2.
```

Crell:
Left some clarifying suggestions; please confirm they're still accurate.  Overall this is really good, so thanks Tim for including it!  It was quite educational.

reference/random/random/randomizer/getfloat.xml
TimWolla: I'm not sure this is useful to include here, because we don't want to select between 1 and 4. This explanation is just to set-up the “environment”, referring back to the

> In other words: There are the same number of representable floats between 1 and 2 as they are between 2 and 4, 4 and 8, 8 and 16, and so on.

of the initial explanation.

reference/random/random/randomizer/getfloat.xml
TimWolla: This is referring back to:

> Randomly sampling an arbitrary number within the requested interval, for example by dividing two integers, might result in a biased distribution for this reason. The necessary rounding will cause some floats to be returned more often than others, especially around powers of two when the density of floats changes.

and the very common (but wrong) solution of using `(mt_rand() / mt_getrandmax()) * ($max - $min) + $min` which also relates to the “Caution note” and Example 2 you're seeing here:

https://www.php.net/manual/en/random-randomizer.nextfloat.php

------------------

I'm not overly attached to the explanation here. If you think it is not helpful, then I'm happy to drop it.

reference/random/random/randomizer/getfloat.xml
```diff
@@ -74,6 +78,114 @@
     displaying the selected number to a user.
    </para>
   </caution>
+  <refsect2>
+   <title>Explanation of the Algorithm Using Example Values</title>
+   <para>
+    To give an example of how the algorithm works, consider that between
+    <literal>1.0</literal> and <literal>2.0</literal> all steps of size <literal>0.125</literal>
+    are exactly representable and between <literal>2.0</literal> and <literal>4.0</literal>
+    all steps of size <literal>0.25</literal> are exactly representable.
```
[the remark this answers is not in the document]
TimWolla: I'm not a fan of including that “in binary”, because it is partly reverted by the suggested part within the parentheses. However we could rephrase that a little more:

> To give an example of how the algorithm works, consider a floating point representation that uses a 3-bit mantissa. This representation is capable of representing 8 different floating point values between consecutive powers of two. This means that between `1.0` and `2.0` all steps of size `0.125` are exactly representable and between `2.0` and `4.0` all steps of size `0.25` are exactly representable. In reality PHP’s floats use a 52-bit mantissa and can represent 2<sup>52</sup> different values between each power of two.

WDYT?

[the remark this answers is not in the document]
Crell: If you can explain what a mantissa is, that could work. 😄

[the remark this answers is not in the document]
Crell: Though I think some form of "the real thing is way more precise" is helpful.

reference/random/random/randomizer/getfloat.xml
Crell: I'm not entirely opposed to it, but if kept it needs to be in a separate block to make it clear that it's a negative example, explaining why not to do the thing in that code block.

Actually, with that negative example that's already there, including an explanation of why it's bad is probably a good thing.  It just needs to be set off more so it's not confusing.

reference/random/random/randomizer/getfloat.xml
Crell: Repetition and explicitness are good things in documentation.  It loops back to the previous point and reinforces it by referencing a previous abstract statement and giving a concrete example, which also serves to make explicit something implied by the previous list.

reference/random/random/randomizer/getfloat.xml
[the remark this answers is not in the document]
TimWolla: > If you can explain what a mantissa is, that could work. 😄

The second sentence about “This representation is capable of representing 8 different floating point values” should sufficiently explain this for *this* documentation page. Should I make this more explicit as `2^3 = 8`?

The floating point representation should be explained in a more general way within https://www.php.net/manual/en/language.types.float.php. I plan to look into that after this is done, time permitting.

> Though I think some form of "the real thing is way more precise" is helpful.

That's included at the end: 

> In reality PHP’s floats use a 52-bit mantissa and can represent 252 different values between each power of two.

reference/random/random/randomizer/getfloat.xml
TimWolla: @Girgias?

reference/random/random/randomizer/getfloat.xml
```diff
@@ -74,6 +78,114 @@
...
+
+    This means that
+    <simplelist><!-- This would benefit from type="inline" -->
```
Girgias, line 90: Can you open an issue on PhD for support for this, it really shouldn't be difficult to add.

reference/random/random/randomizer/getfloat.xml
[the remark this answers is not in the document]
Girgias: The point of this page is to not describe floats, that's the task of the float type page.

Also forcing everything to be defined "in-line" is kinda ridiculous.

I'm in favour of @TimWolla suggested updated description, as I'm also not a massive fan of having ``in binary`` being mentioned.

reference/random/random/randomizer/getfloat.xml
Girgias: This is _already_ verbose and technical, and having such a note in the main content flow is distracting.

If I were writing a LaTeX document, that "note" would be in a remark section, so it is separated from the main content flow.

reference/random/random/randomizer/getfloat.xml
Girgias: I don't know how you would go of separating this?

The whole point of this wall of text is to get an in-depth explanation of the inner working and what happens if one uses a "naive" implementation.

But crucially, this wall of text is totally optional and ignorable.

reference/random/random/randomizer/getfloat.xml
TimWolla: https://github.com/php/phd/issues/83

reference/random/random/randomizer/getfloat.xml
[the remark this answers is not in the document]
TimWolla: Okay, I've pushed my suggested updated introduction.

reference/random/random/randomizer/getfloat.xml
TimWolla: I've wrapped this into another `<refsect3>` to give it a dedicated `<title>`. I've also fixed the example: There are 16 equally spaced floats between 0 and 1 for a 3-bit mantissa, because 0.5 itself is a power of two and thus there are already 8 between 0.5 and 1.

For the same reason `nextFloat()` returns 2^53 different values, despite doubles using a 52-bit mantissa.

Crell: Where is the line between the "this is how it works" and "this is why you shouldn't do this other thing"?  In the original text that's not clear, and that's my main issue.

The added refsect3 does resolve this issue, thanks.

Girgias (approved):

PR #2875, opened 2023-10-20
Update functions.xml

aleksandr-shevchenko:


TimWolla:
The previous version was correct, because it refers to a previous example that uses new syntax and is now showcased using older syntax.

PR #2881, opened 2023-10-20
Use standard DocBook tag for predefined variables

Girgias:
In conjunction with https://github.com/php/phd/pull/85

### DIFF
--- a/language/predefined/variables/argc.xml
+++ b/language/predefined/variables/argc.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-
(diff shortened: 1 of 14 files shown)


Girgias:
@mumumu @saundefined I might run this as a ``[skip-revcheck]`` commit to first see if the indexing and search still behaves as expected.

Also, if I actually change all the ``<phpdoc:classref>`` it would mark the whole docs as outdated when it's not true. But I would make the changes on every single translation repo when this change happens.

Girgias:
I'm actually testing this first on doc-fr and writing a shell script which does a bunch of sed commands to migrate this :')

saundefined:
Can do without `[skip-revcheck]`, not many changes here,
thanks!

Girgias:
> Can do without `[skip-revcheck]`, not many changes here, thanks!

Yeah but I can automate them and push them to translations without messing stuff up. Considering I need to change a lot of classes too :|

Girgias:
Merged with skip revcheck already

PR #2885, opened 2023-10-25
Add `IMAGETYPE_AVIF` constant to documentation

joshuaruesweg:
This constant is introduced with PHP 8.1 and is not documented yet.

For reference: php/doc-de#163

### DIFF
--- a/reference/exif/functions/exif-imagetype.xml
+++ b/reference/exif/functions/exif-imagetype.xml
@@ -133,6 +133,10 @@
         <entry>18</entry>
         <entry><constant>IMAGETYPE_WEBP</constant></entry>
        </row>
+       <row>
+        <entry>19</entry>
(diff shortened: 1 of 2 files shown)


TimWolla:
Related: #2628

reference/exif/functions/exif-imagetype.xml
```diff
@@ -133,6 +133,10 @@
         <entry>18</entry>
         <entry><constant>IMAGETYPE_WEBP</constant></entry>
        </row>
+       <row>
+        <entry>19</entry>
```
TimWolla, line 137: I'm not sure how useful it is to document the exact values here (or anywhere). It might make sense to clean this up in a follow-up to only mention the constants and possibly a link to https://www.php.net/manual/de/image.constants.php. /cc @Girgias

TimWolla:
LGTM, thanks!

Issue #2882, opened 2023-10-21
Performance Considerations - Reduced Memory Usage - Invalid PHP code in an example


krzotr:
From the manual page: https://php.net/features.gc.performance-considerations

In section **Performance Considerations -> Reduced Memory Usage -> Example Memory usage example**

I copied, pasted, and ran PHP example code:

```php
<?php
class Foo
{
    public $var = '3.14159265359';
}

$baseMemory = memory_get_usage();

for ( $i = 0; $i <= 100000; $i++ )
{
    $a = new Foo;
    $a->self = $a;
    if ( $i % 500 === 0 )
    {
        echo sprintf( '%8d: ', $i ), memory_get_usage() - $baseMemory, "\n";
    }
}
```

I got an error:

```
Deprecated: Creation of dynamic property Foo::$self is deprecated in /app/app/webroot/x.php on line 12
```

After replaced `$a->self = $a;` with `$a->var = $a;` error gone

Girgias:
I would say declaring a new public property named ``$self`` would be the proper fix.

PR #2887, opened 2023-10-26
Fixed deprecation warning for dynamic property creation

philipp-check24:
Fixes https://github.com/php/doc-en/issues/2882


Girgias:
Thank you!

PR #2890, opened 2023-10-28
Synchronize ext/dom class synopses with stubs based on PHP 8.3 changes

kocsismate:
Most descriptions are empty, so I'm happy to receive suggestions :)

### DIFF
--- a/reference/dom/domelement.xml
+++ b/reference/dom/domelement.xml
@@ -171,7 +171,7 @@
     <varlistentry xml:id="domelement.props.classname">
      <term><varname>className</varname></term>
      <listitem>
-      <para>The element class name</para>
+      <para>A string representing the classes of the element separated by spaces</para>
      </listitem>
     </varlistentry>
     <varlistentry xml:id="domelement.props.id">
--- a/reference/dom/domelement.xml
+++ b/reference/dom/domelement.xml
@@ -40,12 +40,27 @@
      <interfacename>DOMChildNode</interfacename>
     </oointerface>
 
+    <classsynopsisinfo role="comment">&InheritedConstants;</classsynopsisinfo>
+    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.domnode')/db:partintro/db:section/db:classsynopsis/db:fieldsynopsis[preceding-sibling::db:classsynopsisinfo[1][@role='comment' and text()='&Constants;']]))">
+     <xi:fallback/>
+    </xi:include>
+
     <classsynopsisinfo role="comment">&Properties;</classsynopsisinfo>
     <fieldsynopsis>
      <modifier>public</modifier>
      <modifier>readonly</modifier>
      <type>string</type>
      <varname linkend="domelement.props.tagname">tagName</varname>
+    </fieldsynopsis>
+    <fieldsynopsis>
+     <modifier>public</modifier>
+     <type>string</type>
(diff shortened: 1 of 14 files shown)


TimWolla:
Duplicate of #2813?

kocsismate:
Nope, fortunately this one synchronizes the class synopses. 🙂

Girgias (commented):
I'll wait for @nielsdos to comment about the descriptions but looks good to me otherwise

reference/dom/domelement.xml
```diff
@@ -153,6 +168,18 @@
       <para>The element name</para>
      </listitem>
     </varlistentry>
+    <varlistentry xml:id="domelement.props.classname">
+     <term><varname>className</varname></term>
+     <listitem>
+      <para>The element class name</para>
```
nielsdos, line 174:
```suggestion
      <para>A string representing the classes of the element separated by spaces</para>
```

reference/dom/domnode.xml
```diff
@@ -24,6 +24,50 @@
      <classname>DOMNode</classname>
     </ooclass>
 
+    <classsynopsisinfo role="comment">&Constants;</classsynopsisinfo>
```
nielsdos, line 27: I only added these constants in 8.**4**, they are here to support the compareDocumentPosition method that was added in 8.4.

reference/dom/domnode.xml
```diff
@@ -129,6 +185,48 @@
   </section>
  
 <!-- {{{ DOMNode properties -->
+  <section xml:id="domnode.constants">
```
nielsdos, line 188: Same as earlier: they're 8.4 material

reference/dom/domnode.xml
```diff
@@ -161,6 +259,12 @@
       <para>The parent of this node. If there is no such node, this returns &null;.</para>
      </listitem>
     </varlistentry>
+    <varlistentry xml:id="domnode.props.parentelement">
+     <term><varname>parentElement</varname></term>
+     <listitem>
+      <para>The parent of this element. If there is no such element, this returns &null;.</para>
```
nielsdos, line 265:
```suggestion
      <para>The parent element of this element. If there is no such element, this returns &null;.</para>
```

Otherwise it might be confused with parent**Node**, which is the parent node of the element.

reference/dom/domnode.xml
```diff
@@ -214,6 +318,12 @@
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry xml:id="domnode.props.isconnected">
+     <term><varname>isConnected</varname></term>
+     <listitem>
+      <para></para>
```
nielsdos, line 324:
```suggestion
      <para>Whether the node is connected to a document</para>
```

reference/dom/domnode.xml
```diff
@@ -268,6 +378,26 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.3.0</entry>
```
nielsdos, line 382: Same remark about the constants as earlier.

nielsdos (changes requested):
Thank you for working on this Máté!
I have some remarks and description suggestions :slightly_smiling_face:

reference/dom/domnode.xml
kocsismate: Ohh, I totally missed that :/ And thanks for the heads up!

PR #2906, opened 2023-11-02
[PHP 8.3] Added json_validate function description.

mumumu:
Based on [RFC](https://wiki.php.net/rfc/json_validate) and [json_validate source code](https://github.com/php/php-src/blob/ea299d44a155e0ef39c24a4d79b0e51accf8d9d0/ext/json/json.c#L313-L353).

### DIFF
--- a/reference/json/functions/json-validate.xml
+++ b/reference/json/functions/json-validate.xml
@@ -76,10 +76,9 @@
      <term><parameter>flags</parameter></term>
      <listitem>
       <para>
-       Bitmask of
-       <constant>JSON_INVALID_UTF8_IGNORE</constant>.
-       The behaviour of these constants is described on the
-       <link linkend="json.constants">JSON constants</link> page.
+       Currently only
+       <constant>JSON_INVALID_UTF8_IGNORE</constant>
+       is accepted.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/json/functions/json-validate.xml
+++ b/reference/json/functions/json-validate.xml
@@ -1,0 +1,140 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.json-validate" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>json_validate</refname>
+  <refpurpose>Validates a JSON string</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>bool</type><methodname>json_validate</methodname>
+   <methodparam><type>string</type><parameter>json</parameter></methodparam>
+   <methodparam><type>int</type><parameter>depth</parameter><initializer>512</initializer></methodparam>
+   <methodparam><type>int</type><parameter>flags</parameter><initializer>0</initializer></methodparam>
+  </methodsynopsis>
+  <para>
+   Check if a specified string contains a valid json.
+   Errors during validation can be fetched by using
+   <function>json_last_error</function> and/or
+   <function>json_last_error_msg</function>.
+  </para>
+  <para>
+   It's guaranteed that the json valid in <function>json_validate</function>
+   is also valid in <function>json_decode</function>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>json</parameter></term>
+     <listitem>
+      <para>
+       The json string being analyzed.
+      </para>
+      <para>
+       This function only works with UTF-8 encoded strings.
+      </para>
+      &json.implementation.superset;
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>depth</parameter></term>
+     <listitem>
+      <para>
+       Maximum nesting depth of the structure being decoded.
+       The value must be greater than <literal>0</literal>,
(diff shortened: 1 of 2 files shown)


reference/json/functions/json-validate.xml
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.json-validate" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>json_validate</refname>
+  <refpurpose>Validates a JSON string</refpurpose>
```
TimWolla, line 6:
```suggestion
  <refpurpose>Checks if a string contains valid JSON</refpurpose>
```

reference/json/functions/json-validate.xml
```diff
@@ -0,0 +1,140 @@
...
+  <para>
+   Check if a specified string contains a valid json.
+   Errors during validation can be fetched by using
+   <function>json_last_error</function> and/or
+   <function>json_last_error_msg</function>.
+  </para>
+  <para>
+   It's guaranteed that the json valid in <function>json_validate</function>
+   is also valid in <function>json_decode</function>.
+  </para>
```
TimWolla, line 26:
```xml
<para>
 Returns whether the given &string; is syntactically valid JSON.
 If <function>json_validate</function> returns &true;, <function>json_decode</function>
 will successfully decode the given string when using the same
 <parameter>depth</parameter> and <parameter>flags</parameters>.
</para>
<para>
 If <function>json_validate</function> returns &false;, the cause
 can be retrieved using <function>json_last_error</function> and
 <function>json_last_error_msg</function>.
</para>
<para>
 <function>json_validate</function> uses less memory than
 <function>json_decode</function> if decoded JSON payload is
 not used, because it does not need to build the array or
 object structure containing the payload.
</para>
<caution>
 <para>
  Calling <function>json_validate</function> immediately before
  <function>json_decode</function> will unnecessarily parse the string
  twice, as <function>json_decode</function> implicitly performs
  validation during decoding.
 </para>
 <para>
  <function>json_validate</function> should therefore only be used
  if the decode JSON payload is not immediately used and knowing whether
  the string contains valid JSON is needed.
 </para>
</caution>
```

reference/json/functions/json-validate.xml
```diff
@@ -0,0 +1,140 @@
...
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>json</parameter></term>
+     <listitem>
+      <para>
+       The json string being analyzed.
```
TimWolla, line 37:
```suggestion
       The string to validate.
```
It's not guaranteed that it's valid JSON, that's the point, thus removing the “JSON” mention.

reference/json/functions/json-validate.xml
```diff
@@ -0,0 +1,140 @@
...
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns &true; if the string passed contains a valid json, otherwise
+   returns &false;.
```
TimWolla, line 74:
```suggestion
   Returns &true; if the given string is syntactically valid JSON, otherwise
   returns &false;.
```

reference/json/functions/json-validate.xml
Girgias: Small nit to the above suggestion:
> <function>json_validate</function> uses less memory than
 <function>json_decode</function> if ***the*** decoded JSON payload is
 not used, because it does not need to build the array or
 object structure containing the payload.

reference/json/functions/json-validate.xml
```diff
@@ -0,0 +1,140 @@
...
+       and less than or equal to <literal>2147483647</literal>.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>flags</parameter></term>
+     <listitem>
+      <para>
+       Bitmask of
+       <constant>JSON_INVALID_UTF8_IGNORE</constant>.
```
Girgias, line 60: This is confusing?

Should be a bitmask of `JSON_*` constatns no?

Girgias (commented):
Mostly agree with @TimWolla's suggestions

reference/json/functions/json-validate.xml
mumumu: It seems that json_validate accepts `PHP_JSON_INVALID_UTF8_IGNORE` only.

```c
	if ((options != 0) && (options != PHP_JSON_INVALID_UTF8_IGNORE)) {
		zend_argument_value_error(3, "must be a valid flag (allowed flags: JSON_INVALID_UTF8_IGNORE)");
		RETURN_THROWS();
	}
```

https://github.com/php/php-src/blob/ea299d44a155e0ef39c24a4d79b0e51accf8d9d0/ext/json/json.c#L329-L332

We should change as the following?

```
    <varlistentry>
     <term><parameter>flags</parameter></term>
     <listitem>
      <para>
       Currently only 
       <constant>JSON_INVALID_UTF8_IGNORE</constant>
       is accepted.
      </para>
     </listitem>
    </varlistentry>
```

Girgias: Yes I think this makes more sense. :)

TimWolla (commented):
LGTM now, except for Girgias' remark regarding the `flags` parameter.

Girgias (approved):

PR #2908, opened 2023-11-02
add link to multibyte version of str_pad

jawira:
`mb_str_pad` is a new function added in PHP 8.3, so I added a link in `str_pad` page.


mumumu (approved):
LGTM

We can link multibyte version like [strlen](https://www.php.net/manual/en/function.strlen.php).

PR #2910, opened 2023-11-03
Correct LDAP paging example

marksteward:
As I've noted below the documentation, `empty('0')` is true in PHP, but `$cookie` is an opaque string and can legitimately be `'0'` (e.g. python-ldapserver returns this for the first response to every query).

I think `strlen` should work in all the same places, and also for `'0'`.


Girgias (approved):
Thank you!

PR #2914, opened 2023-11-05
[PHP 8.3] The NumberFormatter::TYPE_CURRENCY constant has been deprecated.

mumumu:
refs: https://www.php.net/manual/en/migration83.deprecated.php#migration83.deprecated.intl


Girgias (approved):
Thank you!

PR #2916, opened 2023-11-05
[PHP 8.3] Calling Phar::setStub() with a resource and a $length is now deprecated.

mumumu:
based on [Other Changes](https://www.php.net/manual/en/migration83.deprecated.php#migration83.deprecated.phar) and [Phar object stub](https://github.com/php/php-src/blob/PHP-8.3/ext/phar/phar_object.stub.php#L201-L205).


Girgias (approved):
Thank you!

PR #2917, opened 2023-11-05
[PHP 8.3] It is now possible to lint multiple files.

mumumu:
refs: https://www.php.net/manual/en/migration83.new-features.php#migration83.new-features.cli

### DIFF
--- a/features/commandline.xml
+++ b/features/commandline.xml
@@ -548,6 +548,11 @@ string(15) "doesntmakesense"
         </para>
         <note>
          <para>
+          As of PHP 8.3.0, it is possible to check multiple files.
+         </para>
+        </note>
+        <note>
+         <para>
           This option does not work together with the <option>-r</option>
           option.
          </para>
--- a/features/commandline.xml
+++ b/features/commandline.xml
@@ -533,9 +533,9 @@ string(15) "doesntmakesense"
        <entry>--syntax-check</entry>
        <entry>
         <para>
-         Syntax check but do not execute the given PHP code. The input from
-         standard input will be processed if no filenames are specified, otherwise
-         each filename will be checked.
+         Syntax check but do not execute the given PHP code.
+         The input from standard input will be processed if no filenames are specified,
+         otherwise each filename will be checked.
          On success, the text
          <literal>No syntax errors detected in &lt;filename&gt;</literal> is
          written to standard output.
@@ -551,7 +551,7 @@ string(15) "doesntmakesense"
         </para>
         <note>
          <para>
-          Before PHP 8.3.0, it was only possible to specify one filename to
+          Prior to PHP 8.3.0, it was only possible to specify one filename to
           be checked.
          </para>
         </note>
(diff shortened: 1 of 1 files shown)


features/commandline.xml
```diff
@@ -546,6 +546,11 @@ string(15) "doesntmakesense"
          This option won't find fatal errors (like undefined functions). Use
          the <option>-f</option> to test for fatal errors too.
         </para>
+        <note>
+         <para>
+          As of PHP 8.3.0, it is possible to check multiple files.
+         </para>
+        </note>
```
Girgias, line 553: I wonder if it better to write something more generic in the main paragraph and add a warning that the feature is not available prior to 8.3, would be more future proof

features/commandline.xml
```diff
@@ -546,6 +546,11 @@ string(15) "doesntmakesense"
          This option won't find fatal errors (like undefined functions). Use
          the <option>-f</option> to test for fatal errors too.
         </para>
+        <note>
+         <para>
+          As of PHP 8.3.0, it is possible to check multiple files.
```
cookieguru, line 551: "It is possible to check multiple files" doesn't explain _how_
```suggestion
          As of PHP 8.3.0, it is possible to specify multiple file names.
```

nielsdos: Yeah I agree that should be described.
For future reference: `php -l file1 file2 file3 ...`

jimwins:
I think my re-write addresses the review feedback, and also stops giving the questionable guidance to just use `-f` to get fatal syntax errors.

features/commandline.xml
```diff
@@ -533,22 +533,26 @@ string(15) "doesntmakesense"
...
         <para>
-         This option won't find fatal errors (like undefined functions). Use
-         the <option>-f</option> to test for fatal errors too.
+         This option won't find fatal errors (like undefined functions) that
+         require executing the code.
         </para>
         <note>
          <para>
-          As of PHP 8.3.0, it is possible to check multiple files.
+          Before PHP 8.3.0, it was only possible to specify one filename to
```
Girgias, line 554: Manual style guide dictates the following:
```suggestion
          Prior to PHP 8.3.0, it was only possible to specify one filename to
```

features/commandline.xml
```diff
@@ -533,22 +533,26 @@ string(15) "doesntmakesense"
        <entry>--syntax-check</entry>
        <entry>
         <para>
-         Provides a convenient way to perform only a syntax check
-         on the given PHP code. On success, the text
+         Syntax check but do not execute the given PHP code. The input from
+         standard input will be processed if no filenames are specified, otherwise
+         each filename will be checked.
```
Girgias, line 538: I don't know if this all fits in 80 chars, but breaking on punctuation is IMHO better (especially if we are rewritting stuff)

```suggestion
         Syntax check but do not execute the given PHP code.
         The input from standard input will be processed if no filenames are specified,
         otherwise each filename will be checked.
```

Girgias (approved):
Two nits but LGTM

Girgias (approved):

PR #2919, opened 2023-11-06
[PHP 8.3] IntlCalendar setDate / setDateTime

saundefined:
Ref: https://github.com/php/php-src/commit/f236eb83b4


Girgias (approved):
This looks reasonable even if I don't know ICU really

Girgias:
Just remembered, you need to update the versions.xml file

saundefined:
Yeah, just remembered that too :)
Thanks, I'll do it now!

PR #2920, opened 2023-11-06
[PHP 8.3] Return type of ReflectionClass::getStaticProperties() is no longer nullable.

mumumu:
refs: [php-src](https://github.com/php/php-src/blob/PHP-8.3/ext/reflection/php_reflection.stub.php#L408-L409) and [other changes](https://www.php.net/manual/en/migration83.other-changes.php#migration83.other-changes.functions.reflection)

### DIFF
--- a/reference/reflection/reflectionclass/getstaticproperties.xml
+++ b/reference/reflection/reflectionclass/getstaticproperties.xml
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis role="ReflectionClass">
-   <modifier>public</modifier> <type class="union"><type>array</type><type>null</type></type><methodname>ReflectionClass::getStaticProperties</methodname>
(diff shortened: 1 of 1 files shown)


reference/reflection/reflectionclass/getstaticproperties.xml
```diff
@@ -26,10 +26,33 @@
...
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.3.0</entry>
+      <entry>
+        Return type of <methodname>ReflectionClass::getStaticProperties</methodname>
+        is no longer nullable.
+      </entry>
```
Girgias, line 49: Maybe something more along the lines of:
```suggestion
      <entry>
       The return type of <methodname>ReflectionClass::getStaticProperties</methodname>
       has been changed to <type>array</type> from <literal>?array</literal>.
      </entry>
```
is better?

Girgias (approved):

PR #2923, opened 2023-11-07
[PHP 8.3]: The ReflectionProperty::setValue $object is now required.

andrewnicols:
I noticed this while updating code for 8.3.

Whilst the deprecation ins referenced as a Note, it was not reflected in the parameter descriptions.

Whilst `$object` is _optional_ before 8.3, it has supported a `null` value to have the same meaning for a long time (probably always). I felt it better to not suggest the alternative as it will lead to future compatability issues.

### DIFF
--- a/reference/reflection/reflectionproperty/setvalue.xml
+++ b/reference/reflection/reflectionproperty/setvalue.xml
@@ -37,8 +37,7 @@
      <listitem>
       <para>
        If the property is non-static an object must be provided to change
-       the property on. If the property is static this parameter is left
-       out and only <parameter>value</parameter> needs to be provided.
+       the property on. If the property is static a value of &null; should be to be provided.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/reflection/reflectionproperty/setvalue.xml
+++ b/reference/reflection/reflectionproperty/setvalue.xml
@@ -37,7 +37,8 @@
      <listitem>
       <para>
        If the property is non-static an object must be provided to change
-       the property on. If the property is static a value of &null; should be to be provided.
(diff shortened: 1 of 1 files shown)


Girgias:
This does not seem correct from looking at the implementation.

From what version are you upgrading from?

andrewnicols:
This is the documented backwards compatibility approach per the RFC:
https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures

> *Suggested backward compatible alternative:* code which invokes
ReflectionProperty::setValue() for a static property with a single argument
should be modified to pass null as the first parameter, code which invokes
ReflectionProperty::setValue() for a static property with an incorrect
first argument (which is neither object, nor null) should pass null instead.


See this 3v4l example which works for all versions form PHP 5.5.0 to
master:
https://3v4l.org/jJFmm#v5.5.0

reference/reflection/reflectionproperty/setvalue.xml
```diff
@@ -37,8 +37,7 @@
      <listitem>
       <para>
        If the property is non-static an object must be provided to change
-       the property on. If the property is static this parameter is left
-       out and only <parameter>value</parameter> needs to be provided.
+       the property on. If the property is static a value of &null; should be to be provided.
```
Girgias, line 40:
```suggestion
       the property on.
       If the property is static a value of &null; <emphasis>must</emphasis> be to be provided.
```

Girgias (commented):
Right yes, I was confused as the new implementation is indeed convoluted

Girgias:
Thank you!

PR #2924, opened 2023-11-07
Add stubs/documentation for missing GMP functions

Girgias:
Reference:
 - https://github.com/php/php-src/commit/4ea85d4044ae23dd91f0c5f6ede83a7da693d7a6

I wonder if it makes sense to deprecate the GMP random functions as ``gmp_random()`` was deprecated during the 7.X release cycle (we should then also deprecate the ``gmp_random_seed()`` function)


Girgias:
Note: need to fix gen stub file name generation to make underscore become ``-``

nielsdos (commented):
Ok for constructor

reference/gmp/functions/gmp-random-bits.xml
```diff
@@ -14,10 +14,12 @@
   </methodsynopsis>
   <para>
    Generate a random number. The number will be between
-   0 and (2 ** <parameter>bits</parameter>) - 1.
+   <literal>0</literal> and
+   <code>(2 ** <parameter>bits</parameter>) - 1</code>.
```
TimWolla, line 18: As that changed: That could use `<superscript>`. See: https://github.com/php/doc-en/blob/5b5736079568ff7c0e79ceedc72580ef7c39b530/reference/random/random/randomizer/nextint.xml#L32

TimWolla:
> I wonder if it makes sense to deprecate the GMP random functions as `gmp_random()` was deprecated during the 7.X release cycle (we should then also deprecate the `gmp_random_seed()` function)

I don't think there's a “drop-in” replacement functionality for them, though?

Girgias:
> > I wonder if it makes sense to deprecate the GMP random functions as `gmp_random()` was deprecated during the 7.X release cycle (we should then also deprecate the `gmp_random_seed()` function)
> 
> I don't think there's a “drop-in” replacement functionality for them, though?

Right, I just find the "bits" variant a bit strange but they do work with arbitrary large integer numbers.

TimWolla (approved):

PR #2938, opened 2023-11-13
Documentation updates for ext-mongodb 1.17

jmikola:
https://jira.mongodb.org/browse/PHPC-2219
https://jira.mongodb.org/browse/PHPC-2180
https://jira.mongodb.org/browse/PHPC-2289
https://jira.mongodb.org/browse/PHPC-1017
https://jira.mongodb.org/browse/PHPC-2311
https://jira.mongodb.org/browse/PHPC-2280

### DIFF
--- a/reference/mongodb/bson/packedarray.xml
+++ b/reference/mongodb/bson/packedarray.xml
@@ -68,7 +68,14 @@
        <row>
         <entry>PECL mongodb 1.17.0</entry>
         <entry>
-         Implements <interfacename>MongoDB\BSON\Type</interfacename>.
+         <para>
+          Implements <interfacename>MongoDB\BSON\Type</interfacename>.
+         </para>
+         <para>
+          <classname>MongoDB\BSON\PackedArray</classname> cannot be serialized
+          in contexts where a BSON document is expected. In earlier versions,
+          the BSON array would have been converted to a document.
+         </para>
         </entry>
        </row>
       </tbody>
--- a/reference/mongodb/bson/packedarray.xml
+++ b/reference/mongodb/bson/packedarray.xml
@@ -68,14 +68,15 @@
        <row>
         <entry>PECL mongodb 1.17.0</entry>
         <entry>
-         <para>
-          Implements <interfacename>MongoDB\BSON\Type</interfacename>.
-         </para>
-         <para>
-          <classname>MongoDB\BSON\PackedArray</classname> cannot be serialized
-          in contexts where a BSON document is expected. In earlier versions,
-          the BSON array would have been converted to a document.
-         </para>
+         Implements <interfacename>MongoDB\BSON\Type</interfacename>.
+        </entry>
+       </row>
+       <row>
+        <entry>PECL mongodb 1.17.0</entry>
+        <entry>
+         <classname>MongoDB\BSON\PackedArray</classname> cannot be serialized
+         in contexts where a BSON document is expected. In earlier versions,
+         the BSON array would have been converted to a document.
         </entry>
        </row>
       </tbody>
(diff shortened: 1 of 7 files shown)


reference/mongodb/bson/packedarray.xml
```diff
@@ -68,7 +68,14 @@
...
-         Implements <interfacename>MongoDB\BSON\Type</interfacename>.
+         <para>
+          Implements <interfacename>MongoDB\BSON\Type</interfacename>.
+         </para>
+         <para>
+          <classname>MongoDB\BSON\PackedArray</classname> cannot be serialized
+          in contexts where a BSON document is expected. In earlier versions,
+          the BSON array would have been converted to a document.
+         </para>
         </entry>
```
Girgias, line 79: Multiple entries for a single version should be split into two different rows.

```suggestion
         Implements <interfacename>MongoDB\BSON\Type</interfacename>.
        </entry>
       </row>
       <row>
        <entry>PECL mongodb 1.17.0</entry>
        <entry>
         <classname>MongoDB\BSON\PackedArray</classname> cannot be serialized
         in contexts where a BSON document is expected. In earlier versions,
         the BSON array would have been converted to a document.
        </entry>
```

reference/mongodb/configure.xml
```diff
@@ -31,9 +31,36 @@ $ sudo pecl install mongodb
...
 
   <para>
-   Installing the driver via PECL will use bundled versions of
-   <link xlink:href="&url.mongodb.libbson;">libbson</link> and
-   <link xlink:href="&url.mongodb.libmongoc;">libmongoc</link> and attempt to
+   As of driver version 1.17.0, PECL will prompt for various
+   <literal>configure</literal> options. If you wish to install the driver with
+   default options in a non-interactive script, you can do so by piping empty
+   strings from the <literal>yes</literal> command:
+   <programlisting role="shell">
```
Girgias, line 38: The manual should not use personalisation (i.e. usage of ``you``).

reference/mongodb/mongodb/driver/manager/addsubscriber.xml
```diff
@@ -52,6 +52,7 @@
   &reftitle.errors;
   <simplelist>
    &mongodb.throws.argumentparsing;
+   <member>Throws <classname>MongoDB\Driver\Exception\InvalidArgumentException</classname> if <parameter>subscriber</parameter> is a <classname>MongoDB\Driver\Monitoring\LogSubscriber</classname>, since loggers can only be registered globally.</member>
```
Girgias, line 55: The manual _should_ try to limit line length to 80 characters for ease of translators

Girgias (commented):
Manual style comments.

alcaeus (approved):

reference/mongodb/configure.xml
jmikola: Just dug up: http://doc.php.net/tutorial/style.php

Thanks for the reminder.

reference/mongodb/bson/packedarray.xml
jmikola: I don't see this mentioned in http://doc.php.net/tutorial/style.php, but I did find prior art for it in the manual: https://www.php.net/manual/en/function.mt-rand.php#refsect1-function.mt-rand-changelog

Is this done to ease the work of translators? From a UI standpoint, it seems this would make it more likely to miss a change if a reader wasn't expecting version numbers to be repeated (although I can't imagine many people read this anyway).

Girgias: I have no idea the reason as to why, but AFAIK the manual has always been documented like this, so better be consistent.

Also, each row is one specific "changelog" entry, which can be varied.

PR #2939, opened 2023-11-13
Add MongoDB\BSON\Binary::TYPE_SENSITIVE constant

jmikola:
https://jira.mongodb.org/browse/PHPC-2267

Split this off from https://github.com/php/doc-en/pull/2938 since it relates to an unmerged PR: https://github.com/mongodb/mongo-php-driver/pull/1488

### DIFF
--- a/reference/mongodb/bson/binary.xml
+++ b/reference/mongodb/bson/binary.xml
@@ -107,6 +107,12 @@
     <fieldsynopsis>
      <modifier>const</modifier>
      <type>int</type>
+     <varname linkend="mongodb-bson-binary.constants.type-sensitive">MongoDB\BSON\Binary::TYPE_SENSITIVE</varname>
+     <initializer>8</initializer>
+    </fieldsynopsis>
+    <fieldsynopsis>
+     <modifier>const</modifier>
+     <type>int</type>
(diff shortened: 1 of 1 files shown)


reference/mongodb/bson/binary.xml
```diff
@@ -180,6 +186,13 @@
      </listitem>
     </varlistentry>
 
+    <varlistentry xml:id="mongodb-bson-binary.constants.type-sensitive">
+     <term><constant>MongoDB\BSON\Binary::TYPE_SENSITIVE</constant></term>
+     <listitem>
+      <para>Sensitive data. This subtype is used for sensitive data that should be excluded from server-side logging when possible.</para>
```
Girgias, line 192: Nit for ease of translators.
```suggestion
      <para>
       Sensitive data.
       This subtype is used for sensitive data that should be excluded from server-side logging when possible.
      </para>
```

alcaeus (approved):

reference/mongodb/bson/binary.xml
jmikola: Thanks. I re-wrapped the text in an amended commit.

I considered fixing the wrapping on other, existing constants but that'd be beyond the scope of this PR and would likely only add more work for translators (assuming that content has already been translated).

PR #2943, opened 2023-11-15
[PHP 8.3] New IntlGregorianCalendar methods

saundefined:
Stubs: https://github.com/php/php-src/blob/PHP-8.3/ext/intl/calendar/calendar.stub.php#L491-L493

### DIFF
--- a/reference/intl/intlgregoriancalendar/createfromdate.xml
+++ b/reference/intl/intlgregoriancalendar/createfromdate.xml
@@ -62,7 +62,7 @@
   &reftitle.examples;
   <para>
    <example>
-    <title><function>IntlGregorianCalendar::createFromDate</function> example</title>
+    <title><methodname>IntlGregorianCalendar::createFromDate</methodname> example</title>
     <programlisting role="php">
 <![CDATA[
 <?php
--- a/reference/intl/intlgregoriancalendar/createfromdate.xml
+++ b/reference/intl/intlgregoriancalendar/createfromdate.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="intlgregoriancalendar.createfromdate" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>IntlGregorianCalendar::createFromDate</refname>
+  <refpurpose>Create a new IntlGregorianCalendar instance from date</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis role="IntlGregorianCalendar">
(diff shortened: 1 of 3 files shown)


reference/intl/intlgregoriancalendar/createfromdate.xml
```diff
@@ -0,0 +1,81 @@
...
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   Returns a new <classname>IntlGregorianCalendar</classname> instance.
+  </para>
+ </refsect1>
+
+</refentry>
```
Girgias, line 61: Add a see also section for the createFromDateTime method (and vice versa on the other page).

Girgias (approved):
Missing a See Also section but LGTM otherwise (well maybe an example would be good but oh well)

reference/intl/intlgregoriancalendar/createfromdate.xml
saundefined: Thanks! Added an example and see also, but not sure about the output of the example..
I think it might be redundant.

What do you think?

reference/intl/intlgregoriancalendar/createfromdate.xml
```diff
@@ -0,0 +1,181 @@
...
+  <para>
+   Returns a new <classname>IntlGregorianCalendar</classname> instance.
+  </para>
+ </refsect1>
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>IntlGregorianCalendar::createFromDate</function> example</title>
```
Girgias, line 65:
```suggestion
    <title><methodname>IntlGregorianCalendar::createFromDate</methodname> example</title>
```

reference/intl/intlgregoriancalendar/createfromdatetime.xml
```diff
@@ -0,0 +1,208 @@
...
+  <para>
+   Returns a new <classname>IntlGregorianCalendar</classname> instance.
+  </para>
+ </refsect1>
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>IntlCalendar::setDateTime</function> example</title>
```
Girgias, line 92:
```suggestion
    <title><methodname>IntlCalendar::setDateTime</methodname> example</title>
```

reference/intl/intlgregoriancalendar/createfromdate.xml
Girgias: It might be, but that can always be removed later :)

Girgias (approved):

PR #2944, opened 2023-11-15
[PHP 8.3] New LDAP functions

saundefined:
Undocumented, based on stubs:

- https://github.com/php/php-src/blob/PHP-8.3/ext/ldap/ldap.stub.php#L784
- https://github.com/php/php-src/blob/PHP-8.3/ext/ldap/ldap.stub.php#L615


Girgias:
@MCMic would you be able to document this?

PR #2949, opened 2023-11-18
cURL: A curl_close() call will not write the cookies out to the file …

deang:
…specified by CURLOPT_COOKIEJAR as of PHP 8.0.  Updating the CURLOPT_COOKIEJAR description to recommend calling unset on the handle if the user needs the cookies written.

### DIFF
--- a/reference/curl/functions/curl-setopt.xml
+++ b/reference/curl/functions/curl-setopt.xml
@@ -1479,8 +1479,10 @@
           <row>
            <entry valign="top"><constant>CURLOPT_COOKIEJAR</constant></entry>
            <entry valign="top">
-            The name of a file to save all internal cookies to when the handle is closed, 
-            e.g. after a call to curl_close.
+            The name of a file to save all internal cookies to when the handle's 
+            destructor is called.  Note that curl_close will not do this as of 
+            PHP 8.0.  If you need the cookies written before the handle goes 
+            out of scope, call <function>unset</function> on the handle.
            </entry>
            <entry valign="top">
            </entry>
--- a/reference/curl/functions/curl-setopt.xml
+++ b/reference/curl/functions/curl-setopt.xml
@@ -1480,9 +1480,15 @@
            <entry valign="top"><constant>CURLOPT_COOKIEJAR</constant></entry>
            <entry valign="top">
             The name of a file to save all internal cookies to when the handle's 
-            destructor is called.  Note that curl_close will not do this as of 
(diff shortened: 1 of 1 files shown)


reference/curl/functions/curl-setopt.xml
```diff
@@ -1479,8 +1479,10 @@
           <row>
            <entry valign="top"><constant>CURLOPT_COOKIEJAR</constant></entry>
            <entry valign="top">
-            The name of a file to save all internal cookies to when the handle is closed, 
-            e.g. after a call to curl_close.
+            The name of a file to save all internal cookies to when the handle's 
+            destructor is called.  Note that curl_close will not do this as of 
+            PHP 8.0.  If you need the cookies written before the handle goes 
+            out of scope, call <function>unset</function> on the handle.
```
Girgias, line 1485:
```suggestion
            The name of a file to save all internal cookies to when the handle's 
            destructor is called.
            <warning>
             <simpara>
              As of PHP 8.0.0, <function>curl_close</function> is a no-op
              and does <emphasis>not</emphasis> destroy the handle.
              If cookies need to be written prior to the handle being automatically
              destroyed, call <function>unset</function> on the handle.
             </simpara>
            </warning>
```

reference/curl/functions/curl-setopt.xml
```diff
@@ -1479,8 +1479,16 @@
           <row>
            <entry valign="top"><constant>CURLOPT_COOKIEJAR</constant></entry>
            <entry valign="top">
-            The name of a file to save all internal cookies to when the handle is closed, 
-            e.g. after a call to curl_close.
+            The name of a file to save all internal cookies to when the handle's
```
Girgias, line 1482:
```suggestion
            The name of a file to save all internal cookies to when the handle's
```

Girgias:
I've amended the wording and markup to be more in line with the manual, let me know if you disagree with the changes.

Girgias:
Apologies for the delay

divinity76:
unset does not guarantee cookies get written.
To force cookie writing, run
```php
curl_setopt($ch, CURLOPT_COOKIELIST, "FLUSH");
```
I tried adding this to the curl_close doc, but it was rejected: https://github.com/php/doc-en/pull/5791

Only when the last reference gets deleted, does the cookies get written automatically. Granted, unset probably deletes the last reference, but you'd have to read the rest of the code to know for sure.

PR #2953, opened 2023-11-20
Revert pgsql 83

devnexen:
it s simpler to add back the `pg-set-error-context-visibility` part.


Girgias (approved):
I wonder if already opening a new PR for 8.4 would be a good idea

PR #2958, opened 2023-11-21
Update command line features

cookieguru:
Add a note to indicate that multiple files can be linted at once in >=8.3


cookieguru:
Duplicate of #2917

PR #2967, opened 2023-11-24
Update floating point number syntax in float.xml

diesirae68:
The update makes numbers like `_123.456` and `123._456` incorrect.


Girgias (approved):

Girgias:
Thank you!

PR #2969, opened 2023-11-25
Security update mail.xml

TCGeek:
Added additional security warnings for unsanitized inputs when using the `mail()` function.

### DIFF
--- a/reference/mail/functions/mail.xml
+++ b/reference/mail/functions/mail.xml
@@ -41,6 +41,11 @@
         <member>User &lt;user@example.com&gt;, Another User &lt;anotheruser@example.com&gt;</member>
        </simplelist>
       </para>
+      <caution>
+       <para>
+         For security reasons, it is recommended for the user to sanitize this parameter.
+       </para>
+      </caution>
      </listitem>
     </varlistentry>
     <varlistentry>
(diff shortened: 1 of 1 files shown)


reference/mail/functions/mail.xml
```diff
@@ -41,6 +41,11 @@
         <member>User &lt;user@example.com&gt;, Another User &lt;anotheruser@example.com&gt;</member>
        </simplelist>
       </para>
+      <caution>
+       <para>
+         For security reasons, it is recommended for the user to sanitize this parameter.
```
kamil-tekiela, line 46: "sanitize" how? What values should be removed?

reference/mail/functions/mail.xml
```diff
@@ -54,6 +59,13 @@
         Subject must satisfy <link xlink:href="&url.rfc;2047">RFC 2047</link>.
        </para>
       </caution>
+      <caution>
+       <para>
+         For security reasons, it is recommended for the user to sanitize this parameter.
+         Using an unsanitized user controlled parameter in conjunction with the
```
kamil-tekiela, line 65: What if it is not user controlled? I belive it could also lead to unwanted command execution, right?

PR #2977, opened 2023-11-30
Correction to variable name - closes #2976

### LINKED ISSUE #2976: Variable name needs correction in some pages of reference/math/functions
Line 18 inside all three files acos.xml, asin.xml and atan.xml should read:
"for every value of num that is within"
instead of
"for every value of a that is within"

This was caused by the following change in a previous revision:
from:
<literal>a==sin(asin(a))</literal>
to
<code>$num == sin(asin($num))</code>

leonardolara:
Correction to variable name (from `a` to `num`)
in files reference/math/asin.xml, acos.xml, atan.xml - closes #2976

### DIFF
--- a/reference/math/functions/acos.xml
+++ b/reference/math/functions/acos.xml
@@ -15,7 +15,7 @@
    Returns the arc cosine of <parameter>num</parameter> in radians.
    <function>acos</function> is the inverse function of
    <function>cos</function>, which means that
-   <code>$num == cos(acos($num))</code> for every value of a that is within
+   <code>$num == cos(acos($num))</code> for every value of num that is within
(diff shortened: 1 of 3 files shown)


reference/dbase/functions/dbase-get-record-with-names.xml
```diff
@@ -32,7 +32,7 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>_number</parameter></term>
+     <term><parameter>number</parameter></term>
```
Girgias, line 35: This change is correct, however can you split this into a second PR?

PR #2981, opened 2023-12-01
Correction to variable name - for issue #2976

### LINKED ISSUE #2976: Variable name needs correction in some pages of reference/math/functions

leonardolara:
In accordance with @Girgias suggestions.
Closes #2976


Girgias:
Thank you!

PR #2983, opened 2023-12-01
Removed references to deprecated mysqli_init function

machavity:
[mysqli_init](https://www.php.net/manual/en/mysqli.init.php) was [deprecated in 8.1](https://wiki.php.net/rfc/deprecations_php_8_1) and will generate deprecated notices. While the function page shows it was deprecated, the rest of the manual still recommended it. Removed it from both the procedural notes (where `mysqli_connect` is preferred) and updated examples for [mysqli_real_connect](https://www.php.net/manual/en/mysqli.real-connect.php)

### DIFF
--- a/reference/mysqli/mysqli/real-connect.xml
+++ b/reference/mysqli/mysqli/real-connect.xml
@@ -41,8 +41,8 @@
   <itemizedlist>
    <listitem>
     <para>
-     <function>mysqli_real_connect</function> needs a valid object which has
-     to be created by function <function>mysqli_init</function>.
+     <function>mysqli_real_connect</function> needs an instance of the <classname>mysqli</classname> object. 
+     This can be obtained by calling <methodname>mysqli::__construct</methodname> or <function>mysqli_connect</function> with no arguments
     </para>
    </listitem>
    <listitem>
@@ -236,9 +236,9 @@
 <![CDATA[
 <?php
 
-$mysqli = mysqli_init();
+$mysqli = new mysqli();
 if (!$mysqli) {
-    die('mysqli_init failed');
+    die('mysqli connection failed');
 }
 
 if (!$mysqli->options(MYSQLI_INIT_COMMAND, 'SET AUTOCOMMIT = 0')) {
@@ -267,7 +267,8 @@ $mysqli->close();
 
 class foo_mysqli extends mysqli {
     public function __construct($host, $user, $pass, $db) {
-        parent::init();
+        // The parent constructor must be called directly, or an E_FATAL error will be generated
+        parent::__construct();
 
         if (!parent::options(MYSQLI_INIT_COMMAND, 'SET AUTOCOMMIT = 0')) {
             die('Setting MYSQLI_INIT_COMMAND failed');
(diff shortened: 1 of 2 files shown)


reference/mysqli/mysqli/real-connect.xml
```diff
@@ -41,8 +41,7 @@
   <itemizedlist>
    <listitem>
     <para>
-     <function>mysqli_real_connect</function> needs a valid object which has
-     to be created by function <function>mysqli_init</function>.
+     <function>mysqli_real_connect</function> needs an instance of the <classname>mysqli</classname> object. This can be obtained by calling <methodname>mysqli::__construct</methodname> or <function>mysqli_connect</function> with no arguments
```
Girgias, line 44: Nit: we try to keep lines less than 80 chars for translators
```suggestion
     <function>mysqli_real_connect</function> needs an instance of the
     <classname>mysqli</classname> object.
     This can be obtained by calling
     <methodname>mysqli::__construct</methodname> or
     <function>mysqli_connect</function> with no arguments.
```

Girgias (approved):
Minor nit but LGTM, thank you!

reference/mysqli/mysqli/real-connect.xml
```diff
@@ -267,7 +270,8 @@ $mysqli->close();
 
 class foo_mysqli extends mysqli {
     public function __construct($host, $user, $pass, $db) {
-        parent::init();
+        // The parent constructor must be called directly, or an E_FATAL error will be generated
```
kamil-tekiela, line 273: What is an `E_FATAL`?

kamil-tekiela (commented):
I will roll back this commit as it does more harm than good. I see one line here that is a valid change so I will recommit it separately.

kamil-tekiela:
Yes, `mysqli::init()` is deprecated, but `mysqli_init()` isn't. It's a difference between a function and a method in the mysqli class. The method is deprecated. The function isn't. The purpose of the function is to be the procedural alternative to `new mysqli()`. Without arguments creation of a new `mysqli` object creates the object but doesn't attempt connection. The `mysqli_init()` function does the same. The purpose of `mysqli::init()` method was less clear and much more confusing. 

Calling `mysqli_connect()` without arguments is not the same as `new mysqli()`. `mysqli_connect()` will read the default values from INI and attempt the connection. 

Some code examples:
```diff
-$mysqli = mysqli_connect();
+$mysqli = mysqli_init();
+$mysqli->connect();
```
```diff
-$mysqli = new mysqli();
+$mysqli = mysqli_init();
```
```diff
class my_mysqli extends mysqli {
    public function __construct() {
-        parent::init();
+        parent::__construct();
    }
}

kamil-tekiela:
> You might want to remove `init()` from the deprecated list on the right side, then. Thanks for fixing it. I misunderstood the errors.

Good suggestion but unfortunately we can't. The way that PHP manual is structured, we prioritize classes and their methods. If a single page contains documentation for method and its corresponding function then it's listed under the class. And in this case, the class method is deprecated. It's confusing to everyone. Changing this would be a large undertaking. 

I just wanted to say that I appreciate you raising this PR nonetheless. It did contain one correct fix. I reapplied it after reverting this PR. It's good that you found it.

Girgias:
The "easy" fix for that is to split the procedural function into its own page. And one can use XIncludes to "import" the documentation from the OO docs.

PR #2984, opened 2023-12-02
Document DOM BC breaks in 8.3

nielsdos:
Basically these are BC-breaking bugfixes

### DIFF
--- a/reference/dom/domparentnode/append.xml
+++ b/reference/dom/domparentnode/append.xml
@@ -48,6 +48,29 @@
   </variablelist>
  </refsect1>
 
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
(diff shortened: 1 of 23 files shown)


reference/dom/domparentnode/append.xml
```diff
@@ -48,6 +48,29 @@
...
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.3.0</entry>
+      <entry>
+       Calling this method on a node without an owner document now works.
+       Previously this threw a <classname>DOMException</classname> with code <constant>DOM_HIERARCHY_REQUEST_ERR</constant>.
```
Girgias, line 66: This text seems to be repeated a _lot_ maybe making this an entity would be helpful?

nielsdos: Ait, done that now

Girgias (approved):

PR #2996, opened 2023-12-06
Add Override Attribute stub

Girgias:
So... I don't remember what the reason was to move the ``AllowDynamicProperties`` Class methods definition out of the subfolder like I did it this time.

But I can't see any issues with it.


TimWolla:
> But I can't see any issues with it.

Should probably be unified then.

TimWolla (approved):

PR #2998, opened 2023-12-06
Update ini.xml

mmalferov:
Added an literal tags to literal values and amended some punctuations

### DIFF
--- a/reference/memcached/ini.xml
+++ b/reference/memcached/ini.xml
@@ -222,7 +222,10 @@
      </term>
      <listitem>
       <para>
-       If set to <literal>On</literal>, consistent hashing (libketama) is used for session handling. When consistent hashing is used, one can add or remove cache node(s) without messing up too much with existing keys. Default is <literal>On</literal>.
+       If set to <literal>On</literal>, consistent hashing (libketama) is used for session handling.
+       When consistent hashing is used,
+       one can add or remove cache node(s) without messing up too much with existing keys.
+       Default is <literal>On</literal>.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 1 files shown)


Girgias (approved):
Thank you!

reference/memcached/ini.xml
```diff
@@ -552,7 +562,7 @@
      </term>
      <listitem>
       <para>
-       Set this value to enable the server be removed after configured number of continuous times connection failure. Default is 0.
+       Set this value to enable the server be removed after configured number of continuous times connection failure. Default is <literal>0</literal>.
```
Girgias, line 565:
```suggestion
       Set this value to enable the server be removed after configured number of continuous times connection failure.
       Default is <literal>0</literal>.
```

reference/memcached/ini.xml
```diff
@@ -541,7 +551,7 @@
      </term>
      <listitem>
       <para>
-       Allow failed memcached server to automatically be removed. Default is Off. (In previous versions, this setting was called memcached.sess_remove_failed)
+       Allow failed memcached server to automatically be removed. Default is <literal>Off</literal>. (In previous versions, this setting was called <literal>memcached.sess_remove_failed</literal>).
```
Girgias, line 554:
```suggestion
       Allow failed memcached server to automatically be removed.
       Default is <literal>Off</literal>.
       (In previous versions, this setting was called <literal>memcached.sess_remove_failed</literal>).
```
I didn't see there were more changes then the single commit the email sent me.
Could you move the previous version thing to a ``<note>`` tag like I did?

reference/memcached/ini.xml
```diff
@@ -530,7 +540,7 @@
      </term>
      <listitem>
       <para>
-       Whether or not to re-use the memcached connections corresponding to the value(s) of session.save_path after the execution of the script ends. Don't use this if certain settings (e.g. SASL settings, sess_binary_protocol) would be overridden between requests. Default is Off.
+       Whether or not to re-use the memcached connections corresponding to the value(s) of <literal>session.save_path</literal> after the execution of the script ends. Don't use this if certain settings (e.g. SASL settings, sess_binary_protocol) would be overridden between requests. Default is <literal>Off</literal>.
```
Girgias, line 543: Breaking up the lines
```suggestion
       Whether or not to re-use the memcached connections corresponding to the value(s) of
       <literal>session.save_path</literal> after the execution of the script ends.
       Don't use this if certain settings (e.g. SASL settings, sess_binary_protocol)
       would be overridden between requests.
       Default is <literal>Off</literal>.
```

Girgias (commented):
Didn't realise there were other changes sorry >-<

Could you break up the lines so that they don't span for ages, this will make it easier for translators in the long run.

PR #3000, opened 2023-12-07
Add method_exists errors and changelog

leveneg:
Add information about a change to method_exists made in 8.0.0 which causes the function to now raise an error when the first argument is not a string or object.

1. [bug 79462](https://bugs.php.net/bug.php?id=79462)
2. [8.0.0 changelog (core section)](https://www.php.net/ChangeLog-8.php)

### DIFF
--- a/reference/classobj/functions/method-exists.xml
+++ b/reference/classobj/functions/method-exists.xml
@@ -49,6 +49,40 @@
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


Girgias:
Those are standard TypeError/ZPP issues that got all fixed in PHP 8.0 and that we do not document.

PR #3001, opened 2023-12-07
Update mb-str-pad.xml

roland-d:
Remove extraneous parenthesis in last example. There are 3 closing parenthesis but there are only 2 opening parenthesis. As a result the example code won't run, this PR fixes that.


nielsdos (approved):
Thanks!

PR #3007, opened 2023-12-11
Update return values for `floor()` function

phansys:


reference/math/functions/cos.xml
```diff
@@ -37,6 +38,31 @@
...
+      <entry>
+       <parameter>num</parameter> no longer accepts internal objects which support
+       numeric conversion. Previous to this version, this function returns &false;
+       in case of an error and an <constant>E_WARNING</constant> error is emitted.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
+ </refsect1>
```
Girgias, line 64: This is invalid, it never supported internal objects.

reference/math/functions/ceil.xml
```diff
@@ -57,7 +57,8 @@
       <entry>8.0.0</entry>
       <entry>
        <parameter>num</parameter> no longer accepts internal objects which support
-       numeric conversion.
+       numeric conversion. Previous to this version, this function returns &false;
+       in case of an error and an <constant>E_WARNING</constant> error is emitted.
```
Girgias, line 61: An E_WARNING is not always emitted, an E_NOTICE is emitted for objects see: https://3v4l.org/SGfsd a warning is only emitted for non-numeric strings as far as I can tell.

And other types such as arrays do not emit any diagnostic at all: https://3v4l.org/FCHmh

Moreover, I consider those to be ZPP/Type failures and something we do not document.

Girgias (changes requested):
Please remove the changelogs changes, we do not document the behaviour of passing invalid types to functions as this was never standard.

phansys:
>Please remove the changelogs changes, we do not document the behaviour of passing invalid types to functions as this was never standard.

Thanks for the feedback.
Based on your insights, I reverted all the changes except for `floor()`.

Girgias (approved):

Girgias:
Thank you!

PR #3011, opened 2023-12-13
Curl setopt.xml

mmalferov:
When translating into another language, the question arises what does mean in English: "The timeout for Expect: 100-continue responses…"?!

A-a-ah! The timeout for «Expect: 100-continue» responses!


Girgias:
I think you need to rebase the PR to get rid of the memcached changes

PR #3016, opened 2023-12-14
mysqli_result::fetch_field_direct correction to parameter name

### LINKED ISSUE #3015: mysqli_result::fetch_field_direct incorrect parameter name in returnvalues section
mysqli_result::fetch_field_direct
file: reference/mysqli/msqli_result/fetch-field-direct.xml

The "returnvalues" section currently reads:
   Returns an object which contains field definition information or &false;
   if no field information for specified <literal>fieldnr</literal> is 
   available.

However, there is no 'fieldnr' parameter, It should be 'index'. This is my suggestion:
   Returns an object which contains field definition information or &false;
   if no field information for specified <parameter>index</parameter> is 
   available.

leonardolara:
Closes #3015

### DIFF
--- a/reference/mysqli/mysqli_result/fetch-field-direct.xml
+++ b/reference/mysqli/mysqli_result/fetch-field-direct.xml
@@ -48,7 +48,7 @@
   &reftitle.returnvalues;
   <para>
    Returns an object which contains field definition information or &false;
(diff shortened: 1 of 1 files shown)


Girgias (approved):
LGTM.

By the way, no need to open an issue, you can just send a PR to fix the issue right away :)

PR #3018, opened 2023-12-15
Fixes found from doc fr qa tools

Girgias:


alfsb:
Yes, these are the type of improvements on en manual that originates by structurally comparing with a particular translation.

These QA tools generates a direct and indirect review of the two languages they compare.

alfsb (commented):
LGTM, but I not really a editor or codeowner on doc-en.

PR #3023, opened 2023-12-15
MYSQLI_OPT_INT_AND_FLOAT_NATIVE has no effect on prepared statements

janedbal:
As documented [here](https://www.php.net/manual/en/mysqli.quickstart.prepared-statements.php#example-4303), this option does nothing when prepared statements are used. I believe it is handy to have this mentioned right next to the option.

### DIFF
--- a/reference/mysqli/mysqli/options.xml
+++ b/reference/mysqli/mysqli/options.xml
@@ -113,8 +113,8 @@
           <row>
            <entry><constant>MYSQLI_OPT_INT_AND_FLOAT_NATIVE</constant></entry>
            <entry>
-             Convert integer and float columns back to PHP numbers. Only valid
-             for mysqlnd. Applies only for non-prepared statements.
+             Convert integer and float columns back to PHP numbers when using non-prepared statements. 
+             Only valid for mysqlnd.
            </entry>
           </row>
           <row>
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli/options.xml
```diff
@@ -114,7 +114,7 @@
            <entry><constant>MYSQLI_OPT_INT_AND_FLOAT_NATIVE</constant></entry>
            <entry>
              Convert integer and float columns back to PHP numbers. Only valid
```
kamil-tekiela, line 116:
```suggestion
             Convert integer and float columns back to PHP numbers when using non-prepared statements. Only valid
```
And then you don't need to add another sentence at the end.

janedbal: Changed

kamil-tekiela (approved):

reference/mysqli/mysqli/options.xml
```diff
@@ -113,8 +113,8 @@
           <row>
            <entry><constant>MYSQLI_OPT_INT_AND_FLOAT_NATIVE</constant></entry>
            <entry>
-             Convert integer and float columns back to PHP numbers. Only valid
-             for mysqlnd.
+             Convert integer and float columns back to PHP numbers when using non-prepared statements.
```
Girgias, line 116:
```suggestion
             Convert integer and float columns back to PHP numbers when using non-prepared statements.
```

PR #3027, opened 2023-12-18
included revision line

fernandowobeto:


Girgias:
We don't need those in EN so I'm not sure those should be kept in the first place

PR #3029, opened 2023-12-18
PDO::getAttribute may throw PDOException

janedbal:
This is currently undocumented behaviour, example [here](https://github.com/php/php-src/issues/12969)

### DIFF
--- a/reference/pdo/pdo/getattribute.xml
+++ b/reference/pdo/pdo/getattribute.xml
@@ -74,7 +74,7 @@
  <refsect1 role="errors">
   &reftitle.errors;
   <para>
-   <methodname>PDO::getAttribute</methodname> may throw <classname>PDOException</classname>
+   <methodname>PDO::getAttribute</methodname> may throw a <exceptionname>PDOException</exceptionname>
    when driver does not support such attribute.
   </para>
  </refsect1>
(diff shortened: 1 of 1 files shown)


SakiTakamachi:
This is also a known issue. I was a little wondering if I should just add it to the documentation, since even if I set the error mode to silent, I still get the warning.

https://github.com/php/php-src/issues/12032

reference/pdo/pdo/getattribute.xml
```diff
@@ -71,6 +71,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   <methodname>PDO::getAttribute</methodname> may throw <classname>PDOException</classname>
```
Girgias, line 77:
```suggestion
   <methodname>PDO::getAttribute</methodname> may throw a <exceptionname>PDOException</exceptionname>
```

reference/pdo/pdo/getattribute.xml
```diff
@@ -71,6 +71,14 @@
...
+   when driver does not support such attribute.
```
Girgias, line 78:
```suggestion
   when the underlying driver does not support the requested <parameter>attribute</parameter>.
```

Girgias (commented):
Maybe needs to be slightly reordered to ideally be below the 80 characters by line guideline

Girgias (approved):

Issue #2747, opened 2023-09-07
Document missing internal global constants


kocsismate:
The following internal global constants are not yet documented:

- [ ] `ZEND_THREAD_SAFE`
- [ ] `ZEND_DEBUG_BUILD`
- [x] `LOCALE_SYSTEM_DEFAULT` (#3043)
- [x] `LOCALE_NEUTRAL` (#3043)
- [x] `DISP_E_PARAMNOTFOUND` (#3043)
- [ ] `CURLOPT_PREQUOTE`
- [ ] `CURLOPT_READDATA`
- [ ] `CURLOPT_SHARE`
- [ ] `CURLOPT_TELNETOPTIONS`
- [ ] `CURLOPT_XFERINFOFUNCTION`
- [ ] `CURLE_BAD_DOWNLOAD_RESUME`
- [ ] `CURLE_FTP_PARTIAL_FILE`
- [ ] `CURLE_HTTP_RETURNED_ERROR`
- [ ] `CURLE_OPERATION_TIMEDOUT`
- [ ] `CURLE_SSL_PINNEDPUBKEYNOTMATCH`
- [ ] `CURLINFO_LASTONE`
- [ ] `CURLINFO_EFFECTIVE_METHOD`
- [ ] `CURLINFO_CAPATH`
- [ ] `CURLINFO_CAINFO`
- [ ] `CURLM_ADDED_ALREADY`
- [ ] `CURLSHOPT_NONE`
- [ ] `CURLSHOPT_SHARE`
- [ ] `CURLSHOPT_UNSHARE`
- [ ] `CURL_LOCK_DATA_COOKIE`
- [ ] `CURL_LOCK_DATA_DNS`
- [ ] `CURL_LOCK_DATA_SSL_SESSION`
- [ ] `CURLAUTH_NONE`
- [ ] `CURLINFO_HTTP_CONNECTCODE`
- [ ] `CURLINFO_HTTPAUTH_AVAIL`
- [ ] `CURLINFO_PROXYAUTH_AVAIL`
- [ ] `CURLOPT_FTP_RESPONSE_TIMEOUT`
- [ ] `CURLOPT_IPRESOLVE`
- [ ] `CURLOPT_MAXFILESIZE`
- [ ] `CURL_IPRESOLVE_V4`
- [ ] `CURL_IPRESOLVE_V6`
- [ ] `CURL_IPRESOLVE_WHATEVER`
- [ ] `CURLOPT_NETRC_FILE`
[...]

PR #3039, opened 2023-12-25
Update output control constants

haszi:
Added additional section headers above each set of constants to clarify what they are and how to access them, clarified which functions the buffer control flags have an effect on, and added a section for the missing handler status constants. 

These last three constants include PHP_OUTPUT_HANDLER_PROCESSED which is returned by ob_get_status() but it is, IMO by mistake, not exposed to userland (PR submitted php/php-src/pull/13019).

The other two, PHP_OUTPUT_HANDLER_STARTED and PHP_OUTPUT_HANDLER_DISABLED, are mentioned as needing documentation in #2747 .

### DIFF
--- a/reference/outcontrol/constants.xml
+++ b/reference/outcontrol/constants.xml
@@ -4,6 +4,10 @@
  &reftitle.constants;
  &extension.constants.core;
  <variablelist>
+  <title>Status flags passed to output handler</title>
+  <para>
+   The following flags are passed to the second (<parameter>phase</parameter>) parameter of the output handler set by <function>ob_start</function> as part of a bitmask:
+  </para>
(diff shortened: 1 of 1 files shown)


reference/outcontrol/constants.xml
```diff
@@ -90,6 +94,12 @@
     </para>
    </listitem>
   </varlistentry>
+</variablelist>
+<variablelist>
+  <title>Output buffer control flags</title>
```
Girgias, line 99: Nit: the indent is off here
```suggestion
  </varlistentry>
 </variablelist>
 <variablelist>
  <title>Output buffer control flags</title>
```

Girgias (approved):
This looks good to me, thank you!

PR #3041, opened 2023-12-26
Fix MongoDB links in the library tutorial page

mmalferov:
Additional slashes result in a 404 pages. E. g.:

https://www.mongodb.com/docs/php-library/master/reference//class/MongoDBClient/

link contains a double-slash before "class" path's part.

https://www.php.net/manual/en/mongodb.tutorial.library.php

### DIFF
--- a/reference/mongodb/tutorial/library.xml
+++ b/reference/mongodb/tutorial/library.xml
@@ -83,12 +83,12 @@ require 'vendor/autoload.php';
    <para>
     If you have used MongoDB drivers in other languages, the library's API
     should look familiar. It contains a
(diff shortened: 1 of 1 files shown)


sy-records (approved):
LGTM.

btw, It is also possible to modify here directly so that other language translations do not have to be modified, https://github.com/php/doc-base/blob/e4890bd6dba8e9cba8cfa154589e3d506b0af2b9/entities/global.ent#L239

Girgias:
Yes, updating this in doc-base makes more sense to me too.

jmikola:
Thank you, @mmalferov. Sorry for not getting a chance to review this earlier (on holiday this week).

PR #3042, opened 2023-12-26
Update ob_list_handlers documentation

haszi:
Added more details on when "default output handler" and "Closure::__invoke" is returned, and what is returned when passing functions or classes/objects and their methods. Added examples for each of these as well.

If bugs.php.net is still maintained, this closes https://bugs.php.net/bug.php?id=65826 .

### DIFF
--- a/reference/outcontrol/functions/ob-list-handlers.xml
+++ b/reference/outcontrol/functions/ob-list-handlers.xml
@@ -106,7 +106,7 @@ class MyClass {
     public static function handle($string) {
         return $string;
     }
-    
+
     public function __invoke($string) {
         return $string;
     }
--- a/reference/outcontrol/functions/ob-list-handlers.xml
+++ b/reference/outcontrol/functions/ob-list-handlers.xml
@@ -30,34 +30,25 @@
   <para>
    If <link linkend="ini.output-buffering">output_buffering</link> is enabled
    and no <link linkend="ini.output-handler">output_handler</link> is set,
-   or no callback or <literal>null</literal> was passed to <function>ob_start</function>,
-   "default output handler" is returned.
+   or no callback or &null; was passed to <function>ob_start</function>,
+   <literal>"default output handler"</literal> is returned.
    Enabling <link linkend="ini.output-buffering">output_buffering</link>
    and setting an <link linkend="ini.output-handler">output_handler</link>
    is equivalent to passing
    an <link linkend="functions.internal">internal (built-in) function</link>
-   to <function>ob_start</function> (see below).
+   to <function>ob_start</function>.
   </para>
   <para>
-   If an <link linkend="functions.anonymous">anonymous function</link>,
-   an <link linkend="functions.arrow">arrow function</link>
-   or a <link linkend="functions.first_class_callable_syntax">first class callable</link>
-   was passed to <function>ob_start</function>,
-   "Closure::__invoke" is returned.
-  </para>
-  <para>
-   If an <link linkend="functions.internal">internal (built-in) function</link>
-   or a <link linkend="functions.user-defined">user-defined function</link>
-   callback was passed to <function>ob_start</function>,
-   then the <link linkend="language.namespaces.basics">fully qualified name</link>
-   of the function is returned.
-  </para>
-  <para>
-   If an <link linkend="language.oop5.basic">class and method name</link>
-   or an <link linkend="language.oop5.basic">object and method name</link>
-   was passed to <function>ob_start</function>,
-   then the <link linkend="language.namespaces.basics">fully qualified name</link>
-   of the method is returned.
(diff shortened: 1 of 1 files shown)


reference/outcontrol/functions/ob-list-handlers.xml
```diff
@@ -25,11 +25,39 @@
...
+   an <link linkend="functions.internal">internal (built-in) function</link>
+   to <function>ob_start</function> (see below).
+  </para>
+  <para>
+   If an <link linkend="functions.anonymous">anonymous function</link>,
+   an <link linkend="functions.arrow">arrow function</link>
+   or a <link linkend="functions.first_class_callable_syntax">first class callable</link>
+   was passed to <function>ob_start</function>,
+   "Closure::__invoke" is returned.
+  </para>
```
Girgias, line 47:
```suggestion
  <para>
   If a <classname>Closure</classname>
   was passed to <function>ob_start</function>,
   <literal>"Closure::__invoke"</literal> is returned.
  </para>
```

The way the closure was created is irrelevant, the wording also does not take into account Closures created via Reflection.

reference/outcontrol/functions/ob-list-handlers.xml
```diff
@@ -25,11 +25,39 @@
...
+   then the <link linkend="language.namespaces.basics">fully qualified name</link>
+   of the function is returned.
+  </para>
+  <para>
+   If an <link linkend="language.oop5.basic">class and method name</link>
+   or an <link linkend="language.oop5.basic">object and method name</link>
+   was passed to <function>ob_start</function>,
+   then the <link linkend="language.namespaces.basics">fully qualified name</link>
+   of the method is returned.
   </para>
```
Girgias, line 61: This could be simplified to just mention that if a callable is passed, the FQN is returned.

Possible flip the order of callable and Closures and mention that if the callable is a closure X is returned.

reference/outcontrol/functions/ob-list-handlers.xml
```diff
@@ -41,18 +69,64 @@
...
+
+// class and static method
+ob_start(['MyClass','staticHandle']);
+print_r(ob_list_handlers());
+ob_end_flush();
+
+// object and non-static method
+ob_start([new MyClass,'handle']);
+print_r(ob_list_handlers());
+ob_end_flush();
```
Girgias, line 129: I would prefer to rewrite the whole example to use ``var_dump()`` instead of ``print_r()`` as the output of print_r is not great IMHO.

Girgias (changes requested):
Wording can be simplified, also you do not cover what is the returned value for an object that implements ``__invoke()``

reference/outcontrol/functions/ob-list-handlers.xml
```diff
@@ -25,11 +25,30 @@
...
+  <para>
+   If <link linkend="ini.output-buffering">output_buffering</link> is enabled
+   and no <link linkend="ini.output-handler">output_handler</link> is set,
+   or no callback or &null; was passed to <function>ob_start</function>,
+   <literal>"default output handler"</literal> is returned.
+   Enabling <link linkend="ini.output-buffering">output_buffering</link>
+   and setting an <link linkend="ini.output-handler">output_handler</link>
+   is equivalent to passing
+   an <link linkend="functions.internal">internal (built-in) function</link>
+   to <function>ob_start</function>.
```
Girgias, line 39: Can't the default ``output_handler`` also be a user defined function, it just needs to be a string?

Girgias (approved):
One minor remark, but LGTM thank you!

Girgias:
> I tried to link the name element of the ob_get_status() docs to the return values section of this file but I couldn't figure out how. I tried adding an xml:id to the first <para> that explains what names get returned but that didn't work either. Could someone let me know how this can be done?

You probably need to add an ``xml:id`` on the ``<seglistitem>`` if I am understanding correctly what you want to do. :)

reference/outcontrol/functions/ob-list-handlers.xml
Girgias: Right, that's good to know but can always be detailled later.

Girgias: Ah yes, then this is fine

Girgias:
> > You probably need to add an `xml:id` on the `<seglistitem>` if I am understanding correctly what you want to do. :)
> 
> There is only a `<refsect1>` with three `<para>`s and the renderer removes the `xml:id`s I add to the `<para>`s. The `returnvalues` `<refsect1>` has an autogenerated `id` but I don't know how to target that with a `<link>`.

I don't understand what you are trying to achieve then :/

Girgias:
> Could you also close https://bugs.php.net/bug.php?id=65826? Thanks.

Done :)

PR #3043, opened 2023-12-27
Add missing COM constants

haszi:
Add three missing COM constants that were listed in #2747.

References for these changes:
 - [LOCALE_NEUTRAL](https://learn.microsoft.com/en-us/windows/win32/intl/locale-neutral)
 - [LOCALE_SYSTEM_DEFAULT](https://learn.microsoft.com/en-us/windows/win32/intl/locale-system-default)
 - DISP_E_PARAMNOTFOUND [1](https://learn.microsoft.com/en-us/windows/win32/com/com-error-codes-2) and [2](https://learn.microsoft.com/en-us/windows/win32/api/oaidl/nf-oaidl-idispatch-invoke)

The values of the constants were verified on PHP 8.2 running on Windows 10.

### DIFF
--- a/reference/com/constants.xml
+++ b/reference/com/constants.xml
@@ -616,6 +616,18 @@
      </entry>
      <entry>As of PHP 7.0.0, the value is <literal>2147614731</literal> on x64.</entry>
     </row>
+    <row xml:id="constant.disp-e-paramnotfound">
+     <entry>
+      <constant>DISP_E_PARAMNOTFOUND</constant>
+      (<type>int</type>)
+     </entry>
+     <entry>-2147352572</entry>
+     <entry>
+      A return value that indicates that one of the parameter IDs
+      does not correspond to a parameter on the method.
+     </entry>
+     <entry>As of PHP 7.0.0, the value is <literal>2147614724</literal> on x64.</entry>
+    </row>
     <row xml:id="constant.mk-e-unavailable">
      <entry>
       <constant>MK_E_UNAVAILABLE</constant>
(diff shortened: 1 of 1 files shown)


reference/com/constants.xml
```diff
@@ -628,6 +640,29 @@
...
+      Instead, use LOCALE_SYSTEM_DEFAULT.
+     </entry>
+     <entry></entry>
+    </row>
+    <row xml:id="constant.locale-system-default">
+     <entry>
+      <constant>LOCALE_SYSTEM_DEFAULT</constant>
+      (<type>int</type>)
+     </entry>
+     <entry>2048</entry>
```
Girgias, line 660: We stopped describing the value of constants, so possibly as a follow-up you could remove all of them from this page (usually we do this when we use gen_stub script to sync so don't bother doing it on other pages).

reference/com/constants.xml
```diff
@@ -616,6 +616,18 @@
...
+     <entry>
+      <constant>DISP_E_PARAMNOTFOUND</constant>
+      (<type>int</type>)
+     </entry>
+     <entry>-2147352572</entry>
+     <entry>
+      A return value that indicates that one of the parameter IDs
+      does not correspond to a parameter on the method.
+     </entry>
+     <entry>As of PHP 7.0.0, the value is <literal>2147614724</literal> on x64.</entry>
```
Girgias, line 629: We shouldn't be describing changes that happened in PHP 7.0.0 as this is the "base version" from which the documentation should work on.

Girgias (approved):

Girgias:
Thank you!

PR #3045, opened 2023-12-27
Update union types docs to note they are available as of 8.0

cookieguru:


Girgias:
The provided change is incorrect, the only thing available as of PHP 8.0 as generalized ***simple*** union types.

Nullable union types are available as of 7.1 and DNF types (which are a form of union types) are only available as of 8.2, moreover the changelog is described in more details in https://www.php.net/manual/en/language.types.declarations.php

PR #3046, opened 2023-12-28
Fix typos

msamesch:
Maybe there's more..

### DIFF
--- a/appendices/migration83/other-changes.xml
+++ b/appendices/migration83/other-changes.xml
@@ -48,9 +48,10 @@
    </para>
 
    <para>
-    Memory stream no longer fails if seek offset is past the end. Instead
-    the memory is increase on the next write and date between the old end and
-    offset is filled with zero bytes in the same way how it works for files.
+    A memory stream no longer fails if seek offset is past the end. Instead
+    the memory will be inceased on the next write and the data between the old
+    end and the offset is filled with zero bytes in the same way as it works
+    for files.
    </para>
 
    <para>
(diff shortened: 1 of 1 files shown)


appendices/migration83/other-changes.xml
```diff
@@ -48,9 +48,10 @@
    </para>
 
    <para>
-    Memory stream no longer fails if seek offset is past the end. Instead
-    the memory is increase on the next write and date between the old end and
-    offset is filled with zero bytes in the same way how it works for files.
+    A memory stream no longer fails if seek offset is past the end. Instead
+    the memory will be inceased on the next write and the data between the old
+    end and the offset is filled with zero bytes in the same way as it works
+    for files.
```
Girgias, line 54: Is this possibly even clearer?
```suggestion
    A memory stream no longer fails if the seek offset is past the end.
    Instead, the memory will be inceased on the next write and the data between
    the old end and the offset is filled with zero bytes, similar to how files work.
```

msamesch:
Thanks a lot for reviewing.

appendices/migration83/other-changes.xml
```diff
@@ -48,10 +48,9 @@
    </para>
 
    <para>
-    A memory stream no longer fails if seek offset is past the end. Instead
-    the memory will be inceased on the next write and the data between the old
-    end and the offset is filled with zero bytes in the same way as it works
-    for files.
+    A memory stream no longer fails if the seek offset is past the end.
+    Instead, the memory will be inceased on the next write and the data between
```
nielsdos, line 52: Still a typo left: inceased instead of increased

msamesch: Thanks. Adding typos while fixing others...

PR #3048, opened 2023-12-28
Remove unnecessary and, thus, confusing comma

omahane:
A comma is neither necessary nor conventional in a "neither/nor" construction and is confusing because it is in, though doesn't complete, the dependent clause.


Girgias:
Thank you!

PR #3049, opened 2023-12-29
Ob list handlers.xml

mmalferov:


Girgias (commented):
This seems to be a duplicate of #3042 ?

PR #3054, opened 2024-01-01
Add missing ZEND constants and descriptions of their PHP aliases

haszi:
Added the missing constants (https://github.com/php/doc-en/issues/2747) ZEND_THREAD_SAFE and ZEND_DEBUG_BUILD, and descriptions of their PHP integer aliases (PHP_ZTS and PHP_DEBUG).

### DIFF
--- a/appendices/reserved.constants.core.xml
+++ b/appendices/reserved.constants.core.xml
@@ -80,6 +80,28 @@
     </simpara>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.zend-thread-safe">
+   <term>
+    <constant>ZEND_THREAD_SAFE</constant>
+    (<type>bool</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Indicates whether the current build of PHP is thread safe.
+    </simpara>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.zend-debug-build">
+   <term>
+    <constant>ZEND_DEBUG_BUILD</constant>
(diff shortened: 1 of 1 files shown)


appendices/reserved.constants.core.xml
```diff
@@ -80,14 +80,36 @@
...
+   <listitem>
+    <simpara>
+     Indicates whether the current build of PHP is a debug build.
+    </simpara>
+   </listitem>
+  </varlistentry>
   <varlistentry xml:id="constant.php-zts">
    <term>
     <constant>PHP_ZTS</constant>
     (<type>int</type>)
```
Girgias, line 108: Is this really an int? And is this is an alias for the ZEND constant? If so maybe saying that it is an alias of ``&Alias; <constant>NAME</constant>`` ?

Girgias: Okay, could you mark the ``ZEND`` constants as aliases for the PHP ones and mark them all as bool?

I'm wondering if maybe we shouldn't deprecate the Zend ones and only keep the PHP ones, but that is out of scope for this PR (and is not a doc change anyway)

Girgias: wtf is this... can you file an issue (or PR) on php-src to mention this, as those should be consistent.

Girgias: Changing it to just be ``if (PHP_ZTS)`` makes sense

Girgias (approved):

PR #3055, opened 2024-01-01
Update namespace FAQ on overriding special constants

haszi:
Removed ZEND_THREAD_SAFE and ZEND_DEBUG_BUILD from the section about defining namespaced special, built-in constants resulting in a fatal error as this is no longer the case for these two.

This code (https://3v4l.org/iMP1P)
```php
<?php
namespace bar;
const ZEND_DEBUG_BUILD = true;
const ZEND_THREAD_SAFE = true;

var_dump(\ZEND_DEBUG_BUILD);
var_dump(\ZEND_THREAD_SAFE);

var_dump(ZEND_DEBUG_BUILD);
var_dump(ZEND_THREAD_SAFE);
```
no longer results in a fatal error but in this output

```
bool(false)
bool(false)
bool(true)
bool(true)
```

### DIFF
--- a/language/namespaces.xml
+++ b/language/namespaces.xml
@@ -1135,7 +1135,7 @@ A\B::foo();   // calls method "foo" of class "B" from namespace "A\A"
     <listitem>
      <simpara>
       <link linkend="language.namespaces.faq.builtinconst">Cannot override special
-      constants NULL, TRUE or FALSE</link>
+      constants &null;, &true; or &false;</link>
      </simpara>
     </listitem>
    </orderedlist>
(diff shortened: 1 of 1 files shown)


language/namespaces.xml
```diff
@@ -1135,7 +1135,7 @@ A\B::foo();   // calls method "foo" of class "B" from namespace "A\A"
     <listitem>
      <simpara>
       <link linkend="language.namespaces.faq.builtinconst">Cannot override special
-      constants NULL, TRUE, FALSE, ZEND_THREAD_SAFE or ZEND_DEBUG_BUILD</link>
+      constants NULL, TRUE or FALSE</link>
```
Girgias, line 1138: Maybe should use the entities ``&null;``, ``&false;`` and ``&true;``

Also @Crell is ``or`` correct here or should it be ``and``?

Girgias (approved):

Issue #2162, opened 2023-01-10
Document missing classes


kocsismate:
Currently, the following internal classes are not yet documented:

  * [x] `SensitiveParameterValue` (#2165)
  * [x] `ClosedGeneratorException` (#3065/#2997)
  * [x] `com_safearray_proxy` (#3066) 
  * [x] `DOMNameSpaceNode` (#3060)
  * [x]  `LibXMLError`  (#3063)
  * [x] `PDORow` (#3059)

cmb69:
Yay, [`com_safearray_proxy`](https://github.com/php/php-src/pull/10278)! I'll try to figure out its details.

Maybe @TimWolla wants to tackle `SensitiveParameterValue`?

TimWolla:
> Maybe @TimWolla wants to tackle `SensitiveParameterValue`?

I can write the description, but I need someone more experienced to bootstrap the docs for me. I've tried `php doc-base/scripts/docgen/docgen.php -c SensitiveParameterValue`, but that result is vastly incomplete (missing `final` on the class, missing properties, …).

kocsismate:
@TimWolla as soon aa docgen created the file, you can run gen_stub.php (with --replace-classsynopses) and it will fix signature.

TimWolla:
I've tried `php build/gen_stub.php --replace-classsynopses ../phpdoc/en/language/predefined/`, but that did not do anything. Am I holding it wrong?

The docgen results are:

```
$ g d HEAD^..HEAD --stat
 language/predefined/interfaces.xml                        |  1 +
 language/predefined/sensitiveparametervalue.xml           | 65 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 language/predefined/sensitiveparametervalue/construct.xml | 69 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 language/predefined/sensitiveparametervalue/debuginfo.xml | 55 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
 language/predefined/sensitiveparametervalue/getvalue.xml  | 55 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
 language/predefined/versions.xml                          |  4 ++++
 6 files changed, 249 insertions(+)
```

kocsismate:
@TimWolla yes, you are missing a parameter: after the --replace.... You need the stub path(s) in php-src. And the very last parameter is for the doc-en path.

TimWolla:
[...]

PR #3059, opened 2024-01-02
Add PDORow class to docs

haszi:
Add the PDORow class to the docs (#2162) and reference it from the appropriate parts of the PDO docs.

As this class has no methods and all but one the properties of this class are dynamic, I added the description of the class and it's dynamic properties to the introduction section (I've got the idea from stdClass's docs). Let me know whether this format is acceptable.

### DIFF
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -131,8 +131,10 @@
    <listitem>
     <simpara>
      Specifies that the fetch method shall return each row as an object with
-     variable names that correspond to the column names returned in the result
-     set. <constant>PDO::FETCH_LAZY</constant> creates the object variable names as they are accessed.
+     variable names that correspond to the column names returned in the result set.
+     <constant>PDO::FETCH_LAZY</constant> returns
+     a <classname>PDORow</classname> object
+     which creates the object variable names as they are accessed.
      Not valid inside <methodname>PDOStatement::fetchAll</methodname>.
     </simpara>
    </listitem>
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -131,10 +131,10 @@
    <listitem>
     <simpara>
      Specifies that the fetch method shall return each row as an object with
-     variable names that correspond to the column names returned in the result set.
+     property names that correspond to the column names returned in the result set.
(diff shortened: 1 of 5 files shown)


Girgias:
@kocsismate we should probably have the attribute somehow generated in the class synopsis. don't you think?

reference/pdo/constants.xml
```diff
@@ -131,8 +131,10 @@
    <listitem>
     <simpara>
      Specifies that the fetch method shall return each row as an object with
-     variable names that correspond to the column names returned in the result
-     set. <constant>PDO::FETCH_LAZY</constant> creates the object variable names as they are accessed.
+     variable names that correspond to the column names returned in the result set.
```
kamil-tekiela, line 134:
```suggestion
     property names that correspond to the column names returned in the result set.
```

reference/pdo/constants.xml
```diff
@@ -131,8 +131,10 @@
...
+     <constant>PDO::FETCH_LAZY</constant> returns
+     a <classname>PDORow</classname> object
+     which creates the object variable names as they are accessed.
```
kamil-tekiela, line 137:
```suggestion
     which creates the object property names as they are accessed.
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -69,7 +69,8 @@
         <listitem><para>
          <literal>PDO::FETCH_LAZY</literal>: combines
          <literal>PDO::FETCH_BOTH</literal> and <literal>PDO::FETCH_OBJ</literal>,
-         creating the object variable names as they are accessed
+         and is returning a <classname>PDORow</classname> object
+         which is creating the object variable names as they are accessed.
```
kamil-tekiela, line 73:
```suggestion
         which is creating the object property names as they are accessed.
```

kamil-tekiela (approved):
It looks good. It would be nice to add some more information from https://phpdelusions.net/pdo/fetch_modes#FETCH_LAZY but this can be done later I guess.

kamil-tekiela:
I leave it up to @kocsismate to merge. It looks good to me.

PR #3060, opened 2024-01-02
Add DOMNameSpaceNode class to docs

haszi:
Add the missing DOMNameSpaceNode class to the docs (ref #2162) and reference it from the DOM book.

### DIFF
--- a/reference/dom/domnamespacenode.xml
+++ b/reference/dom/domnamespacenode.xml
@@ -97,7 +97,7 @@
      <term><varname>nodeName</varname></term>
      <listitem>
       <para>
-       Returns the most accurate name for the current node type
+       The qualified name of this node.
       </para>
      </listitem>
     </varlistentry>
@@ -105,11 +105,7 @@
      <term><varname>nodeValue</varname></term>
      <listitem>
       <para>
-       The value of this node, depending on its type.
-       Contrary to the W3C specification, the node value of
-       <classname>DOMElement</classname> nodes is equal to
-       <link linkend="domnode.props.textcontent">DOMNode::textContent</link>
-       instead of &null;.
+       The namespace URI declared by this node, or &null; if the empty namespace.
       </para>
      </listitem>
     </varlistentry>
@@ -117,8 +113,10 @@
      <term><varname>nodeType</varname></term>
      <listitem>
       <para>
-       Gets the type of the node. One of the predefined
-       <link linkend="dom.constants">XML_xxx_NODE constants</link>
+       The type of the node. In this case it returns
+       <link linkend="dom.constants">
+        <constant>XML_NAMESPACE_DECL_NODE</constant>
+       </link>.
       </para>
      </listitem>
     </varlistentry>
@@ -126,7 +124,7 @@
      <term><varname>prefix</varname></term>
      <listitem>
       <para>
-       The namespace prefix of this node.
+       The namespace prefix declared by this node.
       </para>
      </listitem>
     </varlistentry>
@@ -134,7 +132,7 @@
      <term><varname>localName</varname></term>
      <listitem>
       <para>
-       Returns the local part of the qualified name of this node.
+       The local part of the qualified name of this node.
       </para>
      </listitem>
     </varlistentry>
@@ -142,7 +140,7 @@
      <term><varname>namespaceURI</varname></term>
      <listitem>
       <para>
-       The namespace URI of this node, or &null; if it is unspecified.
+       The namespace URI declared by this node, or &null; if it is unspecified.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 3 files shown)


reference/dom/domnamespacenode.xml
```diff
@@ -0,0 +1,233 @@
...
+       instead of &null;.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry xml:id="domnamespacenode.props.nodetype">
+     <term><varname>nodeType</varname></term>
+     <listitem>
+      <para>
+       Gets the type of the node. One of the predefined
+       <link linkend="dom.constants">XML_xxx_NODE constants</link>
```
nielsdos, line 121:
```suggestion
       Gets the type of the node. In this case it returns <link linkend="dom.constants"><constant>XML_NAMESPACE_DECL_NODE</constant></link>.
```

reference/dom/domnamespacenode.xml
```diff
@@ -0,0 +1,233 @@
...
+
+<!-- {{{ DOMNameSpaceNode properties -->
+  <section xml:id="domnamespacenode.props">
+   &reftitle.properties;
+   <variablelist>
+    <varlistentry xml:id="domnamespacenode.props.nodename">
+     <term><varname>nodeName</varname></term>
+     <listitem>
+      <para>
+       Returns the most accurate name for the current node type
```
nielsdos, line 100: You can be more specific here, and say that it returns the qualified name of the namespace.

reference/dom/domnamespacenode.xml
```diff
@@ -0,0 +1,233 @@
...
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry xml:id="domnamespacenode.props.nodevalue">
+     <term><varname>nodeValue</varname></term>
+     <listitem>
+      <para>
+       The value of this node, depending on its type.
```
nielsdos, line 108: You can be more specific and say this returns the namespace URI this node declared, or &null if the empty namespace.

reference/dom/domnamespacenode.xml
```diff
@@ -0,0 +1,233 @@
...
+       Gets the type of the node. One of the predefined
+       <link linkend="dom.constants">XML_xxx_NODE constants</link>
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry xml:id="domnamespacenode.props.prefix">
+     <term><varname>prefix</varname></term>
+     <listitem>
+      <para>
+       The namespace prefix of this node.
```
nielsdos, line 129:
```suggestion
       The namespace prefix declared by this node.
```

reference/dom/domnamespacenode.xml
```diff
@@ -0,0 +1,233 @@
...
+      <para>
+       Returns the local part of the qualified name of this node.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry xml:id="domnamespacenode.props.namespaceuri">
+     <term><varname>namespaceURI</varname></term>
+     <listitem>
+      <para>
+       The namespace URI of this node, or &null; if it is unspecified.
```
nielsdos, line 145: Actually, this is the namespace URI that this node declares instead of the namespace URI of the node itself.
That's different in semantics from the namespace URI of this node to be precise (which would be the xmlns namespace, but it's not).

```suggestion
       The namespace URI declared by this node, or &null; if it is unspecified.
```

nielsdos (changes requested):
Thanks for your interest on adding docs for this!
I have some minor remarks/suggestions.

reference/dom/domnamespacenode.xml
nielsdos: Indeed this is the namespace declared by this node.

localName is the prefix declared by this node, unless there is no prefix in which case it will be equal to "xmlns".
So a simpler way of putting it is saying that: localName _is_ actually equal to the local name of this node, not the local name _declared by_ this node.

It's all a bit inconsistent unfortunately.

Demo link: https://3v4l.org/sSv4Q

nielsdos: For nodeName it might be easiest after all to describe it as the qualified name of the node.

reference/dom/domnamespacenode.xml
```diff
@@ -0,0 +1,231 @@
...
+       The parent of this node.
+       If there is no such node, this returns &null;.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry xml:id="domnamespacenode.props.parentelement">
+     <term><varname>parentElement</varname></term>
+     <listitem>
+      <para>
+       The parent element of this element.
```
nielsdos, line 177:
```suggestion
       The parent element of this node.
```

nielsdos (changes requested):
Looks mostly good, thanks!
Just found one minor issue.

nielsdos (approved):
Applied the last typo fix myself.
Thanks for your contribution!

PR #3061, opened 2024-01-03
Update INI documentation for ibm_db2

NattyNarwhal:
A bunch of previously undocumented stuff, including things that are ugly and probably should be deprecated.

Note some of these are platform-specific; I have tried to word based on this ("on IBM i" vs. "when connecting to IBM i"), but I've considered adding a column, if that would make sense.

### DIFF
--- a/reference/ibm_db2/ini.xml
+++ b/reference/ibm_db2/ini.xml
@@ -279,7 +279,7 @@
       This options affects the internal buffer allocation strategy on IBM i.
       By default, this option is <literal>0</literal>.
       When this option is set, buffers are allocated with a much larger size,
-      in case the database is misleading about character size when converting
+      in case the database underestimates a string's size when converting
       between encodings.
       This option uses six times as much memory for buffers (to account for
       the largest possible UTF-8 sequences), but may be needed if truncated
@@ -332,20 +332,23 @@
     </term>
     <listitem>
      <para>
-      This option sets if diagnostic messages like warnings and errors are
-      always sent to the error log on IBM i.
-      Normally, only a brief message on failure is sent, as this option is set
-      to <literal>0</literal> by default. Note that you can still and should
-      call i.e. <function>db2_stmt_errormsg</function> manually.
+      This option sets if SQL diagnostic messages like warnings and errors are
+      always sent to the PHP error log on IBM i.
+      Normally, only a brief message on failure is sent (such as "statement
+      execute failed") to the PHP error log, as this option is set to
+      <literal>0</literal> by default.
+      Note that you can still and should call i.e.
+      <function>db2_stmt_errormsg</function> manually as part of checking if
+      functions fail.
       <itemizedlist>
        <listitem>
         <para>
-         0 - Only logs concise messages.
+         0 - Only logs brief messages.
         </para>
        </listitem>
        <listitem>
         <para>
-         1 - Logs the SQL diagnostic message in addition to the concise message.
+         1 - Logs the SQL diagnostic message in addition to the brief message.
         </para>
        </listitem>
       </itemizedlist>
@@ -436,14 +439,18 @@
     </term>
     <listitem>
      <para>
-      The ASCII CCSID to use for character conversions from EBCDIC on IBM i.
-      Setting this to <literal>1208</literal> will use UTF-8.
+      The PASE CCSID to use for character conversions from EBCDIC on IBM i.
       By default, this is <literal>0</literal>, which will select the default
-      ASCII job CCSID.
+      PASE job CCSID, which comes from the PASE locale settings.
+      For example, setting this to <literal>1208</literal> will use UTF-8.
+      You should only need to set this if the PASE job CCSID isn't the expected
+      CCSID, and you can't change the locale.`
      </para>
      <para>
       To learn more about CCSIDs on IBM i, consult the
       <link xlink:href="https://www.ibm.com/docs/en/i/7.5?topic=information-ccsid-reference">IBM documentation</link>.
+      To learn how locales on IBM i PASE are mapped to CCSIDs, consult the
+      <link xlink:href="https://www.ibm.com/docs/en/i/7.5?topic=ssw_ibm_i_75/apis/pase_locales.html">IBM documentation</link>.
      </para>
     </listitem>
    </varlistentry>
--- a/reference/ibm_db2/ini.xml
+++ b/reference/ibm_db2/ini.xml
@@ -182,7 +182,7 @@
     <listitem>
      <para>
       This option controls the transaction isolation mode used.
-      By default, this is set to 0, so commitment control isn't used.
+      By default, this option is <literal>0</literal>, so commitment control isn't used.
       This option can be overriden when connecting if the array key
       <parameter>i5_commit</parameter> is set in the connection options array
       passed to <function>db2_connect</function> or
--- a/reference/ibm_db2/ini.xml
+++ b/reference/ibm_db2/ini.xml
@@ -224,10 +224,12 @@
     </term>
     <listitem>
      <para>
-      This controls if a blank user ID should be allowed on IBM i. By default,
-      this option is off. Unlike <parameter>ibm_db2.i5_ignore_userid</parameter>,
-      this option doesn't force all user IDs to be empty, but simply allows an
-      empty user ID to be passed, for connecting to Db2 as the current user.
+      This controls if a blank user ID should be allowed on IBM i.
+      By default, this option is <literal>0</literal>.
+      Unlike <parameter>ibm_db2.i5_ignore_userid</parameter>, this option
+      doesn't force all user IDs to be empty or change job behaviour, but
+      simply allows an empty user ID to be passed, for connecting to Db2 as the
+      current user.
       <itemizedlist>
        <listitem>
         <para>
@@ -363,21 +365,22 @@
     <listitem>
      <para>
       This option ignores the user ID when connecting to the database when
-      running on IBM i.
+      running on IBM i, and runs SQL/CLI functionality inside of the PHP job,
+      instead of a separate job.
       By default, this option is <literal>0</literal>.
-      When enabled, a blank user ID and password is always used when
-      connecting, so it uses the current profile when connecting.
-      Historically, this provided as a convenience for simple applications.
-      It should not be used in new applications.
+      When enabled, it no longer uses a separate database server job, and
+      always uses the current user profile for the database, ignoring the
+      username and password passed to <function>db2_connect</function> and
+      <function>db2_pconnect</function>.
       <itemizedlist>
        <listitem>
         <para>
-         0 - Uses the specified credentials.
+         0 - Uses the specified credentials, and use an SQL/CLI server job.
         </para>
        </listitem>
        <listitem>
         <para>
-         1 - Always uses a blank user ID and password.
+         1 - Always use blank credentials, and run SQL/CLI in the PHP job.
         </para>
        </listitem>
       </itemizedlist>
--- a/reference/ibm_db2/ini.xml
+++ b/reference/ibm_db2/ini.xml
@@ -12,7 +12,7 @@
       <entry>&Name;</entry>
       <entry>&Default;</entry>
       <entry>&Changeable;</entry>
-      <entry>Changelog</entry>
+      <entry>&Changelog;</entry>
      </row>
     </thead>
     <tbody xml:id="ibm-db2.configuration.list">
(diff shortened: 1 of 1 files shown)


reference/ibm_db2/ini.xml
```diff
@@ -34,24 +34,78 @@
...
+      <entry><link linkend="ini.ibm-db2.i5-blank-userid">ibm_db2.i5_blank_userid</link></entry>
+      <entry>"0"</entry>
+      <entry>PHP_INI_SYSTEM</entry>
+      <entry>Available since ibm_db2 1.9.7.</entry>
+     </row>
+     <row>
+      <entry><link linkend="ini.ibm-db2.i5-char-trim">ibm_db2.i5_char_trim</link></entry>
+      <entry>"0"</entry>
+      <entry>PHP_INI_SYSTEM</entry>
+      <entry>Available since ibm_db2 2.1.0.</entry>
```
Girgias, line 47: ``since`` shouldn't be used in this context and should be ``as of ibm_db2 2.1.0`` but this is existing wording so relegating this as a QA issue.

reference/ibm_db2/ini.xml
```diff
@@ -98,21 +152,20 @@
...
-      application. When <parameter>ibm_db2.i5_all_pconnect</parameter> = 1, all db2 connections 
-      become persistent (<function>db2_pconnect</function>). On i5/OS, <function>db2_pconnect</function> performs 
-      dramatically better with lower machine stress over <function>db2_connect</function>. This 
-      is a convenience override of <function>db2_connect</function> to evoke <function>db2_pconnect</function> without 
-      PHP source code changes. 
+      This option forces all connections to be persistent on IBM i. Basically,
+      all <function>db2_connect</function> calls transparently become
+      <function>db2_pconnect</function> calls. By default, this option is 0.
+      This option is provided as a convenience for cases where persistent
+      connections are faster. It should not be used in new applications.
```
Girgias, line 159: I prefer to break lines on a full stop as this is a good marker for translations to break on too.
Also a literal tag might be useful.
```suggestion
      This option forces all connections to be persistent on IBM i.
      Basically, all <function>db2_connect</function> calls transparently become
      <function>db2_pconnect</function> calls.
      By default, this option is <literal>0</literal>.
      This option is provided as a convenience for cases where persistent
      connections are faster.
      It should not be used in new applications.
```

reference/ibm_db2/ini.xml
```diff
@@ -126,8 +179,10 @@
     </term>
     <listitem>
      <para>
-      This option controls the isolation mode used for i5 schema collections 
-      in the PHP application (see <parameter>i5_commit</parameter> for override).
+      This option controls the transaction isolation mode used. By default,
+      this is set to 0, so commitment control isn't used. You can override the
+      option when connecting when <parameter>i5_commit</parameter> is in the
+      connection options array.
```
Girgias, line 185: The docs shouldn't use personalization (i.e. `you`)

```suggestion
      this is set to 0, so commitment control isn't used.
      This option can be overriden in the connecting when
      <parameter>i5_commit</parameter> is in the connection options array.
```
Not sure the parameter tag here is valid too? Is this an array key or?

reference/ibm_db2/ini.xml
```diff
@@ -158,43 +213,133 @@
...
      </para>
     </listitem>
    </varlistentry>
-   <varlistentry xml:id="ini.ibm-db2.instance-name">
+   <varlistentry xml:id="ini.ibm-db2.i5-guard-profile">
     <term>
-     <parameter>ibm_db2.instance_name</parameter>
-     <type>string</type>
+     <parameter>ibm_db2.i5_guard_profile</parameter>
+     <type>int</type>
```
Girgias, line 298: Possibly bool?

reference/ibm_db2/ini.xml
```diff
@@ -158,43 +213,133 @@
...
+        </para>
+       </listitem>
+      </itemizedlist>
      </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="ini.ibm-db2.i5-log-verbose">
+    <term>
+     <parameter>ibm_db2.i5_log_verbose</parameter>
+     <type>int</type>
```
Girgias, line 323: bool?

reference/ibm_db2/ini.xml
```diff
@@ -158,43 +213,133 @@
      </para>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="ini.ibm-db2.i5-blank-userid">
+    <term>
+     <parameter>ibm_db2.i5_blank_userid</parameter>
+     <type>int</type>
```
Girgias, line 219: bool?

Girgias (commented):
Minor nits but looks good overall

reference/ibm_db2/ini.xml
NattyNarwhal: Yes, this is an array key passed to the optional array argument of `db2_(p)connect`. I'll make this more obvious.

reference/ibm_db2/ini.xml
NattyNarwhal: These are defined as int currently, but I should probably change them to be bool when applicable.

Girgias: ACK, then this is OK :)

reference/ibm_db2/ini.xml
```diff
@@ -158,43 +217,138 @@
...
      <type>int</type>
     </term>
     <listitem>
      <para>
-      This option controls the internal ibm_db2 allocation scheme 
-      for large DBCS column buffers.
+      This options affects the internal buffer allocation strategy on IBM i.
+      By default, this option is <literal>0</literal>.
+      When this option is set, buffers are allocated with a much larger size,
+      in case the database is misleading about character size when converting
```
alanseiden, line 282: @NattyNarwhal Perhaps replace "is misleading about" with "underestimates".

reference/ibm_db2/ini.xml
```diff
@@ -158,43 +217,138 @@
...
+   </varlistentry>
+   <varlistentry xml:id="ini.ibm-db2.i5-log-verbose">
+    <term>
+     <parameter>ibm_db2.i5_log_verbose</parameter>
+     <type>int</type>
+    </term>
+    <listitem>
      <para>
-      This option is ignored on Windows operating systems.
+      This option sets if diagnostic messages like warnings and errors are
```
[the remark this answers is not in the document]
alanseiden: @NattyNarwhal ^

reference/ibm_db2/ini.xml
```diff
@@ -158,43 +217,138 @@
...
+      always sent to the error log on IBM i.
+      Normally, only a brief message on failure is sent, as this option is set
+      to <literal>0</literal> by default. Note that you can still and should
+      call i.e. <function>db2_stmt_errormsg</function> manually.
+      <itemizedlist>
+       <listitem>
+        <para>
+         0 - Only logs concise messages.
```
alanseiden, line 343: @NattyNarwhal I think your word "brief" is better than "concise." You could write "brief" consistently. Concise implies the style of writing (i.e. not using unnecessary words), whereas brief is strictly about length.

reference/ibm_db2/ini.xml
```diff
@@ -205,30 +359,160 @@
...
-      is only for simple DB2 based websites that never require profile switching 
-      and therefore can avoid all overhead of server mode additional QSQSRVR jobs. 
-      This is a convenience override of db2_(p)connect to set the userid and password 
-      values to null without PHP source code changes. This override can be used in 
-      combination with <parameter>ibm_db2.i5_all_pconnect</parameter> = 1.
+      This option ignores the user ID when connecting to the database when
+      running on IBM i.
+      By default, this option is <literal>0</literal>.
+      When enabled, a blank user ID and password is always used when
+      connecting, so it uses the current profile when connecting.
```
alanseiden, line 366: @NattyNarwhal The historical use case for null user and password was to stay out of Server Mode, i.e. not connect to a QSQSRVR job. The database is accessed directly inside PHP. It was a performance optimization. The danger is not only being limited to the web server user, (in the case of a web application), but that database resources could potentially linger in the PHP job.

reference/ibm_db2/ini.xml
```diff
@@ -205,30 +359,160 @@
...
+    <term>
+     <parameter>ibm_db2.i5_servermode-subsystem</parameter>
+     <type>string</type>
+    </term>
+    <listitem>
+     <para>
+      This option changes which subsystem database server jobs run under on
+      IBM i.
+      By default, this option is &null;, so jobs will run under the default
+      subsystem for QSQSERVER jobs.
```
alanseiden, line 494: `QSQSRVR`

reference/ibm_db2/ini.xml
NattyNarwhal: Null username/password avoiding spawning a server job isn't documented elsewhere for `ibm_db2`; the underlying [SQLConnect](https://www.ibm.com/docs/en/i/7.2?topic=functions-sqlconnect-connect-data-source) mentions non-server mode, but implies you can just use server mode where null uid/pwd still uses a server, just with current profile. So how does it handle to use server vs. in process?

Checking ibm_db2, it does set  the environment attribute to use a here, if this option is off:
```c
            if (IBM_DB2_G(i5_ignore_userid) < 1 && !is_i5_server_mode) {
                attr = SQL_TRUE;
                rc = SQLSetEnvAttr((SQLHENV)conn_res->henv, SQL_ATTR_SERVER_MODE, (SQLPOINTER)(intptr_t)attr, SQL_IS_INTEGER);
                if (rc == SQL_SUCCESS) {
                    is_i5_server_mode=1;
                }
            }
```
...and it checks the option here for if it should set the job's subsystem if using a server:
```c
        /* 1.9.7 - IBM i consultant request switch subsystem QSQSRVR job (customer workload issues) */
        if (IBM_DB2_G(i5_servermode_subsystem) && strlen(IBM_DB2_G(i5_servermode_subsystem)) > 0 && IBM_DB2_G(i5_ignore_userid) < 1) {
            rc = SQLSetConnectAttr((SQLHDBC)conn_res->hdbc, SQL_ATTR_SERVERMODE_SUBSYSTEM, (SQLPOINTER) IBM_DB2_G(i5_servermode_subsystem), SQL_NTS);
        } 
```
...so this option is kind of misleadingly named - it should be more like "run database job in the same job as PHP". That's a lot to describe. And the funny thing is I think you can still use a different profile for same-job, it just suppresses the passed options to `SQLConnect` if so:
```c
#ifdef PASE /* 1.9.7 - IBM i cosolidate connect differences/options (code clarity) */
            /* If the string contains a =, use SQLDriverConnect */
            if ( strstr(database, "=") != NULL ) { 
                /* IBM i unqualified connect requires output buffer (long time ibm_db2 bug) */
                rc = SQLDriverConnect((SQLHDBC)conn_res->hdbc, (SQLHWND)NULL,
                        (SQLCHAR*)database, SQL_NTS, i_buffer, 255, &i_outlen, SQL_DRIVER_NOPROMPT );
            } else { /* IBM i allow/not blank db, uid, pwd (customer request vs. security issue) */
                if (IBM_DB2_G(i5_ignore_userid) > 0 || (IBM_DB2_G(i5_blank_userid) > 0 && uid_len == 0 && password_len == 0) ) {
                    rc = SQLConnect( (SQLHDBC)conn_res->hdbc, (SQLCHAR *)database,
                            (SQLSMALLINT)database_len, (SQLCHAR *)NULL, (SQLSMALLINT)0,
                            (SQLCHAR *)NULL, (SQLSMALLINT)0);
                } else { /* normal uid/pwd connection */
                    rc = SQLConnect( (SQLHDBC)conn_res->hdbc, (SQLCHAR *)database,
                        (SQLSMALLINT)database_len, (SQLCHAR *)uid, (SQLSMALLINT)uid_len,
                        (SQLCHAR *)password, (SQLSMALLINT)password_len );
                }
            }
#else /* LUW */
            /* If the string contains a =, use SQLDriverConnect */
            if ( strstr(database, "=") != NULL ) {
                rc = SQLDriverConnect((SQLHDBC)conn_res->hdbc, (SQLHWND)NULL,
                        (SQLCHAR*)database, SQL_NTS, NULL, 0, NULL, SQL_DRIVER_NOPROMPT );
            } else {
                rc = SQLConnect( (SQLHDBC)conn_res->hdbc, (SQLCHAR *)database,
                        (SQLSMALLINT)database_len, (SQLCHAR *)uid, (SQLSMALLINT)uid_len,
                        (SQLCHAR *)password, (SQLSMALLINT)password_len );
            }
#endif /* PASE */
```

alanseiden: @NattyNarwhal Thanks for confirming and clarifying. Yes, Tony sometimes named things literally in a technical sense, not so much what the purpose was. Yes, the purpose is to run the database job in the same job as PHP. Should we provide an alias called "i5-use-client-job" or something?

NattyNarwhal: I think it should probably be renamed in ibm_db2, but that's a separate issue ([tracking here](https://github.com/php/pecl-database-ibm_db2/issues/59)). For now, we should document the existing behaviour in this PR.

NattyNarwhal:
This doesn't include `i5_check_pconnect`, but that option is Not Good and contains some suspect functionality.

reference/ibm_db2/ini.xml
```diff
@@ -205,30 +364,165 @@
...
+     <type>int</type>
+    </term>
+    <listitem>
+     <para>
+      The PASE CCSID to use for character conversions from EBCDIC on IBM i.
+      By default, this is <literal>0</literal>, which will select the default
+      PASE job CCSID, which comes from the PASE locale settings.
+      For example, setting this to <literal>1208</literal> will use UTF-8.
+      You should only need to set this if the PASE job CCSID isn't the expected
+      CCSID, and you can't change the locale.`
```
Girgias, line 450: Usage of `you` here:
```suggestion
      This should only be modified if the PASE job CCSID isn't the expected
      CCSID, and the locale cannot be modified.
```
Maybe?

reference/ibm_db2/ini.xml
```diff
@@ -126,8 +181,12 @@
     </term>
     <listitem>
      <para>
-      This option controls the isolation mode used for i5 schema collections 
-      in the PHP application (see <parameter>i5_commit</parameter> for override).
+      This option controls the transaction isolation mode used.
+      By default, this is set to 0, so commitment control isn't used.
```
Girgias, line 185: Possibly to be consistent with the rest?
```suggestion
      By default, this option is <literal>0</literal>, so commitment control isn't used.
```

Girgias:
Thank you!

PR #3062, opened 2024-01-03
Add PDO_IBM INI options

NattyNarwhal:
These are basically equivalent to the ibm_db2 versions.

### DIFF
--- a/reference/pdo_ibm/ini.xml
+++ b/reference/pdo_ibm/ini.xml
@@ -1,0 +1,109 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<section xml:id="pdo-ibm.configuration" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ &reftitle.runtime;
+ &extension.runtime;
+ <para>
+  <table>
+   <title>PDO_IBM Configuration Options</title>
+   <tgroup cols="4">
+    <thead>
+     <row>
+      <entry>&Name;</entry>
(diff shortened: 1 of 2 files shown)


reference/pdo_ibm/ini.xml
```diff
@@ -0,0 +1,109 @@
...
+      <entry>&Name;</entry>
+      <entry>&Default;</entry>
+      <entry>&Changeable;</entry>
+      <entry>&Changelog;</entry>
+     </row>
+    </thead>
+    <tbody xml:id="pdo-ibm.configuration.list">
+     <row>
+      <entry><link linkend="ini.pdo-ibm.i5-dbcs-alloc">pdo_ibm.i5_dbcs_alloc</link></entry>
+      <entry>"0"</entry>
```
Girgias, line 21: Should be ``0`` maybe? Considering it is an int

reference/pdo_ibm/ini.xml
```diff
@@ -0,0 +1,109 @@
...
+      <entry>PHP_INI_SYSTEM</entry>
+      <entry>Added in PDO_IBM 1.5.0</entry>
+     </row>
+     <row>
+      <entry><link linkend="ini.pdo-ibm.i5-override-ccsid">pdo_ibm.i5_override_ccsid</link></entry>
+      <entry>"0"</entry>
```
Girgias, line 27: Ditto

reference/pdo_ibm/ini.xml
NattyNarwhal: Same as ibm_db2, these should be converted to bool in the extension, but they're int for now

PR #3066, opened 2024-01-04
Add com_safearray_proxy to docs

haszi:
Add com_safearray_proxy to docs (#2162). 

This internal class doesn't have any methods or properties and doesn't extend or implement any other class or interface hence the almost empty documentation.

The description of the class is based on the [comments](https://github.com/php/php-src/blob/0b7cd1423add534fc27030f57934cc8bb6d4c6a1/ext/com_dotnet/com_saproxy.c#L17-L22)  in the implementation.

### DIFF
--- a/reference/com/book.xml
+++ b/reference/com/book.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
- 
+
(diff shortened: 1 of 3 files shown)


Girgias (approved):
LGTM

Girgias:
Thank you!

PR #3068, opened 2024-01-05
Incorrect param name in mb_ereg_replace_callback() docs

mmalferov:
Either the $string parameter needs to be changed to $subject, or the parameter name in the function description needs to be corrected


Girgias:
It is indeed string looking at https://heap.space/xref/php-src/ext/mbstring/mbstring.stub.php?r=d8ef868b#211

PR #3073, opened 2024-01-05
Update all ob_  clean/flush functions

haszi:
Update all six `ob_ clean/flush` functions' documentation with the following:

 - describe what each function does in detail
 - differentiate between contents of the output buffer and values returned by the output handler
 - list flags passed to output handlers
 - list flags affecting or, if unexpected, not affecting each function
 - use consistent terminology for output buffer, it's contents and it's handler, and 'turn off' instead of 'destroy' an output buffer
 - add a section on notices issued

Closes the following:
https://bugs.php.net/bug.php?id=64977
https://bugs.php.net/bug.php?id=69404
https://bugs.php.net/bug.php?id=71486
https://bugs.php.net/bug.php?id=74399
https://bugs.php.net/bug.php?id=76563

### DIFF
--- a/reference/outcontrol/functions/ob-clean.xml
+++ b/reference/outcontrol/functions/ob-clean.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="function.ob-clean" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>ob_clean</refname>
-  <refpurpose>Clean (erase) the output buffer</refpurpose>
+  <refpurpose>Clean (erase) active output buffer contents</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
@@ -13,17 +13,24 @@
    <void/>
   </methodsynopsis>
   <para>
-   This function discards the contents of the output buffer.
+   This function calls the output handler (with the
+   <link linkend="constant.php-output-handler-clean">
+    <constant>PHP_OUTPUT_HANDLER_CLEAN</constant>
+   </link>
+   flag), discards it's return value
+   and cleans (erases) the contents of the active output buffer.
   </para>
   <para>
-   This function does not destroy the output buffer like
-   <function>ob_end_clean</function> does.
+   This function does not turn off the active output buffer like
+   <function>ob_end_clean</function> or <function>ob_get_clean</function> does.
   </para>
   <para>
-   The output buffer must be started by
-   <function>ob_start</function> with <link
-   linkend="constant.php-output-handler-cleanable">PHP_OUTPUT_HANDLER_CLEANABLE</link>
-   flag. Otherwise <function>ob_clean</function> will not work.
+   <function>ob_clean</function> will fail
+   without an active output buffer started with the
+   <link linkend="constant.php-output-handler-cleanable">
+    <constant>PHP_OUTPUT_HANDLER_CLEANABLE</constant>
+   </link>
+   flag.
   </para>
  </refsect1>
 
@@ -39,13 +46,21 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   If the function fails it generates an <constant>E_NOTICE</constant>.
+  </para>
+ </refsect1>
+
  <refsect1 role="seealso">
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member><function>ob_flush</function></member>
-    <member><function>ob_end_flush</function></member>
+    <member><function>ob_start</function></member>
     <member><function>ob_end_clean</function></member>
+    <member><function>ob_get_clean</function></member>
+    <member><function>ob_flush</function></member>
    </simplelist>
   </para>
  </refsect1>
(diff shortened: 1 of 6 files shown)


reference/outcontrol/functions/ob-clean.xml
```diff
@@ -3,27 +3,34 @@
 <refentry xml:id="function.ob-clean" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>ob_clean</refname>
-  <refpurpose>Clean (erase) the output buffer</refpurpose>
+  <refpurpose>Clean (erase) active output buffer contents</refpurpose>
```
Girgias, line 6: Possible better wording
```suggestion
  <refpurpose>Clean (erase) the content of the active output buffer</refpurpose>
```

reference/outcontrol/functions/ob-end-flush.xml
```diff
@@ -3,37 +3,41 @@
 <refentry xmlns="http://docbook.org/ns/docbook" xml:id="function.ob-end-flush">
  <refnamediv>
   <refname>ob_end_flush</refname>
-  <refpurpose>Flush (send) the output buffer and turn off output buffering</refpurpose>
+  <refpurpose>Flush (send) output handler return value and turn off active output buffer</refpurpose>
```
Girgias, line 6: Possible better wording?
```suggestion
  <refpurpose>Flush (send) the return value of the active output buffer handler and turns it off</refpurpose>
```

reference/outcontrol/functions/ob-end-flush.xml
```diff
@@ -82,13 +84,12 @@
   <para>
    <simplelist>
     <member><function>ob_start</function></member>
-    <member><function>ob_get_contents</function></member>
```
Girgias, line 85: Why remove the link to `ob_get_contents()`?

reference/outcontrol/functions/ob-flush.xml
```diff
@@ -3,25 +3,34 @@
 <refentry xml:id="function.ob-flush" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>ob_flush</refname>
-  <refpurpose>Flush (send) the output buffer</refpurpose>
+  <refpurpose>Flush (send) output handler return value</refpurpose>
```
Girgias, line 6:
```suggestion
  <refpurpose>Flush (send) the return value of the active output buffer handler</refpurpose>
```

reference/outcontrol/functions/ob-flush.xml
```diff
@@ -37,17 +46,24 @@
...
+  <para>
+   If the function fails it generates an <constant>E_NOTICE</constant>.
+  </para>
+ </refsect1>
+
  <refsect1 role="seealso">
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member><function>ob_get_contents</function></member>
```
Girgias, line 44: Ditto

reference/outcontrol/functions/ob-get-clean.xml
```diff
@@ -3,30 +3,51 @@
...
+   </link>
+   flag.
   </para>
+  <caution>
+   <simpara>
+    <function>ob_get_clean</function> will return false
+    but will not generate an <constant>E_NOTICE</constant>
+    if there is no active output buffer.
+   </simpara>
+  </caution>
```
Girgias, line 42: Should be moved to the return value section

Girgias (commented):
Looks mostly good, although frustrating that there is no way to have a "macro"/"function" in XML as this is a lot of duplicate wording :/

reference/outcontrol/functions/ob-end-clean.xml
```diff
@@ -3,29 +3,44 @@
 <refentry xmlns="http://docbook.org/ns/docbook" xml:id="function.ob-end-clean">
  <refnamediv>
   <refname>ob_end_clean</refname>
-  <refpurpose>Clean (erase) the output buffer and turn off output buffering</refpurpose>
+  <refpurpose>Clean (erase) active output buffer contents and turn off active output buffer</refpurpose>
```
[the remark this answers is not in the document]
Girgias: Yeah the documentation can be pretty inconsistent. I think indicative is better.

reference/outcontrol/functions/ob-clean.xml
Girgias: I think both are fine, but not sure, @Crell might know the difference?

reference/outcontrol/functions/ob-end-flush.xml
Girgias: Ah, that is indeed confusing behaviour and I expected this to turn off the handler... maybe change the last bit to `turns off the output buffering` ?

Girgias (approved):

reference/outcontrol/functions/ob-clean.xml
Crell: At least in common American English, I think either content or contents works in context.

PR #3078, opened 2024-01-07
Change PHP_DEBUG and PHP_ZTS to bool

haszi:
Change PHP_DEBUG and PHP_ZTS to bool and add that these are aliases of their ZEND equivalents. Follow up of #3054.


Girgias:
I've just added it to the 8.4 milestone.

PR #3084, opened 2024-01-08
Improve example code to read ini shorthand

PPazderski:
The existing example code cannot handle any later added shorthand formats and is simply outdated when a dedicated standard function exists.

### DIFF
--- a/reference/info/functions/ini-get.xml
+++ b/reference/info/functions/ini-get.xml
@@ -62,23 +62,7 @@ echo 'display_errors = ' . ini_get('display_errors') . "\n";
 echo 'register_globals = ' . ini_get('register_globals') . "\n";
 echo 'post_max_size = ' . ini_get('post_max_size') . "\n";
 echo 'post_max_size+1 = ' . (ini_get('post_max_size')+1) . "\n";
-echo 'post_max_size in bytes = ' . return_bytes(ini_get('post_max_size'));
-
-function return_bytes($val) {
-    $val = trim($val);
-    $last = strtolower($val[strlen($val)-1]);
-    switch($last) {
-        // The 'G' modifier is available
-        case 'g':
-            $val *= 1024;
-        case 'm':
-            $val *= 1024;
-        case 'k':
-            $val *= 1024;
-    }
-
-    return $val;
-}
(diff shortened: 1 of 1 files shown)


reference/info/functions/ini-get.xml
```diff
@@ -62,23 +62,7 @@ echo 'display_errors = ' . ini_get('display_errors') . "\n";
...
-            $val *= 1024;
-        case 'm':
-            $val *= 1024;
-        case 'k':
-            $val *= 1024;
-    }
-
-    return $val;
-}
+echo 'post_max_size in bytes = ' . ini_parse_quantity(ini_get('post_max_size'));
```
Girgias, line 65: This function is only available as of PHP 8.2, and the documentations covers PHP as of PHP 7.0.0.

So just replacing this line is not generally what we do. It might be better to rewrite the example completely to not try to compute the number of bytes.

Girgias (changes requested):

reference/info/functions/ini-get.xml
PPazderski: Ok I missed that. What brought me initial here was the kinda confusing code comment in the example which became this way from a change removing all references to PHP 5 and 7 (a0ae28d3bc85f927c22649ebd9a590b921534b7d).

Made me falsely believe PHP 7 isn't relevant here anymore. So it is to early for this change.

Girgias: No worries :) That commit only removed references to PHP ***7.0.0*** as this is the new "baseline", There are still plenty of PHP 7.1.0 mentions :)

philip: Update this PR to _add_ the new example with an "Available as of PHP 8.2.0" code comment?

Girgias: > Update this PR to _add_ the new example with an "Available as of PHP 8.2.0" code comment?

This would work yes.

lacatoire:
Superseded by #4372, merged in January 2025, which made the same change: the `return_bytes()` helper is gone and the example now calls `ini_parse_quantity(ini_get('post_max_size'))`. Nothing left to rebase here.

Thanks for the work @PPazderski, and sorry this was not flagged when the other PR landed.

PR #3085, opened 2024-01-09
Typo. NULL was just NUL

AndiKod:
At least I almost hope it's a typo so I won't waste the maintainers time :)


leonardolara:
In UTF-8 and ASCII for example, the byte represented by value 0 is called NUL.
So there is indeed a NUL byte.
It's not a typo, just an abbreviation for NULL. In my opinion this change will not affect comprehension.

Girgias:
Closing as `NUL` is the correct name for ASCII.

We probably are inconsistent in the docs about this.

PR #3088, opened 2024-01-10
Make code examples on enumeration documentation executable

mmalferov:
On the one hand, we are writing a closing tag ?> in examples, probably, such an agreement is in effect so that the user can copy the example code and paste it into a document that may contain, say, HTML markup, and immediately look at the result of this code. At the same time, there are functions in the documentation whose signatures contain three dots, which, when trying to execute the example, will result in a syntax error: Parse error: syntax error, unexpected token "..." Wouldn't it be a good practice to abandon the three dots in the definition of functions in the examples?

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -61,14 +61,16 @@ enum Suit
 <![CDATA[
 <?php
 
-function pick_a_card(Suit $suit) { ... }
+function pick_a_card(Suit $suit) {}
 
 $val = Suit::Diamonds;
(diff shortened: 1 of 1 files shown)


Girgias:
I think it makes sense to attempt to have all examples executable, however I would change the ``...`` to ``/* ... */`` instead of removing them.

Girgias (approved):

Girgias:
Thank you!

PR #3095, opened 2024-01-10
[PHP 8.3] Update `mt-srand` and `srand` functions

saundefined:


TimWolla (commented):
Thank you. Regarding phrasing it would probably make sense to steal my explanation of the `seed` parameter from https://www.php.net/manual/en/random-engine-mt19937.construct.php, because `mt_srand` is equivalent to Mt19937::__construct().

Girgias (approved):
Wording looks OK to me.

TimWolla (approved):

PR #3097, opened 2024-01-10
Update output_reset_rewrite_vars()

haszi:
Remove the session ID rewrite from the example as that is using a different functionality which is also not available with the default ini settings. Update example and its output to work with default settings.

Remove confusing first clause of the description as from the users perspective the only thing that happens is that the variables set are removed.


Girgias (approved):
TIL about this functionality

PR #3098, opened 2024-01-10
Update output control configuration page

haszi:
Updated the default value and clarified when a hidden field is added to forms.

### DIFF
--- a/appendices/ini.list.xml
+++ b/appendices/ini.list.xml
@@ -7,7 +7,7 @@
    PHP setup.
   </para>
   <para>
-   The "&Changeable;" column shows the modes determining when and where a directive may be set.  
+   The "&Changeable;" column shows the modes determining when and where a directive may be set.
    See the <link linkend="configuration.changes.modes">Changeable mode values</link> section for their definitions.
   </para>
   <para>
@@ -678,10 +678,16 @@
        <entry></entry>
       </row>
       <row>
-       <entry><link linkend="ini.url-rewriter.tags">url_rewriter.tags</link></entry>
-       <entry>"a=href,area=href,frame=src,form=,fieldset="</entry>
+       <entry><link linkend="ini.url-rewriter.hosts">url_rewriter.hosts</link></entry>
+       <entry>""</entry>
        <entry>PHP_INI_ALL</entry>
-       <entry></entry>
+       <entry>Available as of PHP 7.1.0.</entry>
+      </row>
+      <row>
+       <entry><link linkend="ini.url-rewriter.tags">url_rewriter.tags</link></entry>
+       <entry>"form="</entry>
+       <entry>PHP_INI_ALL</entry>
+       <entry>Prior to PHP 7.1.0, the default value was "a=href,area=href,frame=src,form=,fieldset=".</entry>
       </row>
       <row>
        <entry><link linkend="ini.user-agent">user_agent</link></entry>
--- a/reference/outcontrol/ini.xml
+++ b/reference/outcontrol/ini.xml
@@ -38,9 +38,12 @@
      <entry><link linkend="ini.url-rewriter.tags">url_rewriter.tags</link></entry>
      <entry>"form="</entry>
      <entry>PHP_INI_ALL</entry>
-     <entry>Before PHP 7.1.0, this was used to set
-     session's trans sid rewrite. From PHP 7.1.0, it is only used by
-     <function>output_add_rewrite_var</function>.</entry>
+     <entry>
+      As of PHP 7.1.0, this INI setting only affects
+      <function>output_add_rewrite_var</function>.
+      Prior to PHP 7.1.0, this INI setting enabled transparent session id support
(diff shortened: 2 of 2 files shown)


reference/outcontrol/ini.xml
```diff
@@ -36,7 +36,7 @@
     </row>
     <row>
      <entry><link linkend="ini.url-rewriter.tags">url_rewriter.tags</link></entry>
-     <entry>"a=href,area=href,frame=src,form=,fieldset="</entry>
+     <entry>"form="</entry>
      <entry>PHP_INI_ALL</entry>
      <entry>Before PHP 7.1.0, this was used to set
```
Girgias, line 41: Could you maybe reword the description to something similar to:
```
As of PHP 7.1.0, this INI setting only affects <function>output_add_rewrite_var</function>.
Prior to PHP 7.1.0, this INI setting enabled trans sid for the session extension.
```

Not that I really understand what that trans sid refers to.

appendices/ini.list.xml
```diff
@@ -677,11 +677,17 @@
...
+       <entry>PHP_INI_ALL</entry>
+       <entry>Available as of PHP 7.1.0.</entry>
+      </row>
       <row>
        <entry><link linkend="ini.url-rewriter.tags">url_rewriter.tags</link></entry>
-       <entry>"a=href,area=href,frame=src,form=,fieldset="</entry>
+       <entry>"form="</entry>
        <entry>PHP_INI_ALL</entry>
-       <entry></entry>
+       <entry>Prior to PHP 7.1.0, the default value was "a=href,area=href,frame=src,form=,fieldset=".</entry>
```
Girgias, line 690:
```suggestion
       <entry>Prior to PHP 7.1.0, the default value was "a=href,area=href,frame=src,form=,fieldset=".</entry>
```
Add a literal tag maybe? But it looks like this whole page is missing some so maybe do this as a follow-up QA PR

reference/outcontrol/ini.xml
```diff
@@ -137,20 +140,25 @@
...
+     <literal>url_rewriter.tags</literal> specifies HTML tags and attributes in which URLs
      are rewritten by <function>output_add_rewrite_var</function> values.
-     Defaults to
-     <literal>a=href,area=href,frame=src,input=src,form=</literal>
+     Defaults to <literal>"form="</literal>.
     </simpara>
     <simpara>
-     <literal>form</literal> is special tag. <literal>&lt;input hidden="session_id" name="session_name"&gt;</literal>
-     is added as form variable.
+     Adding <literal>form=</literal> or any form attribute
```
Girgias, line 148:
```suggestion
     Adding <literal>"form="</literal> or any form attribute
```
Is it meant to be a string?

Girgias (commented):
One final question but looks good to me thank you!

reference/outcontrol/ini.xml
```diff
@@ -66,9 +69,10 @@
...
     <para>
      You can enable output buffering for all files by setting this directive
-     to 'On'. If you wish to limit the size of the buffer to a certain size -
-     you can use a maximum number of bytes instead of 'On', as a value for
-     this directive (e.g., output_buffering=4096).
+     to <literal>'On'</literal>.
+     If you wish to limit the size of the buffer to a certain size -
+     you can use a maximum number of bytes instead of <literal>'On'</literal>,
+     as a value for this directive (e.g., <literal>output_buffering=4096</literal>).
      This directive is always Off in PHP-CLI.
```
Girgias, line 76: This is annoying, "you" shouldn't be used in the documentation but this usage already exists...

reference/outcontrol/ini.xml
```diff
@@ -66,9 +69,10 @@
    <listitem>
     <para>
      You can enable output buffering for all files by setting this directive
```
Girgias, line 71:
```suggestion
     Output buffering for all files can be enabled by setting this directive
```

reference/outcontrol/ini.xml
```diff
@@ -66,9 +69,10 @@
...
-     to 'On'. If you wish to limit the size of the buffer to a certain size -
-     you can use a maximum number of bytes instead of 'On', as a value for
-     this directive (e.g., output_buffering=4096).
+     to <literal>'On'</literal>.
+     If you wish to limit the size of the buffer to a certain size -
+     you can use a maximum number of bytes instead of <literal>'On'</literal>,
+     as a value for this directive (e.g., <literal>output_buffering=4096</literal>).
```
Girgias, line 75:
```suggestion
     To limit the size of the buffer, a number/quantity corresponding to the
     maximum numbers of bytes allowed can be used instead of
     <literal>'On'</literal> for the value of this directive.
     For example <literal>output_buffering=4096</literal>.
```
Do you know if using ``4M`` works as a quantity? I would expect _yes_ if it used the usual INI parsing semantics.

Girgias (commented):
Hopefully final remarks

Girgias (approved):

Girgias:
Thank you!

PR #3099, opened 2024-01-10
IntlChar::istitle and IntlChar::totitle examples corrections

leonardolara:
The current examples are incorrect and are identical to the examples of islower.xml and toupper.xml respectively.

The istitle() function determines wether the character is a titlecase code point ("Lt" general category of Unicode). Titlecase characters are digraphs (in one character) formed by a combination of two glyphs, one uppercase and the other lowercase.
Similarly, totitle() converts a character to titlecase, if available.

The examples included in these commits were retrieved from the Unicode specification, also presented in https://www.compart.com/en/unicode/category/Lt

### DIFF
--- a/reference/intl/intlchar/istitle.xml
+++ b/reference/intl/intlchar/istitle.xml
@@ -47,11 +47,19 @@
    <programlisting role="php">
     <![CDATA[
 <?php
-var_dump(IntlChar::istitle("A"));
-var_dump(IntlChar::istitle("a"));
-var_dump(IntlChar::istitle("Φ"));
-var_dump(IntlChar::istitle("φ"));
-var_dump(IntlChar::istitle("1"));
+//Latin Capital Letter Dz with Caron U+01C4
+var_dump(IntlChar::istitle("Ǆ")); 
+//Latin Capital Letter D with Small Letter Z with Caron U+01C5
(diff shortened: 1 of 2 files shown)


reference/intl/intlchar/istitle.xml
```diff
@@ -47,11 +47,19 @@
...
+var_dump(IntlChar::istitle("ǅ"));
+//Latin Small Letter Dz with Caron U+01C6
+var_dump(IntlChar::istitle("ǆ"));
+
+//Greek Capital Letter Alpha with Prosgegrammeni U+1FBC
+var_dump(IntlChar::istitle("ᾼ"));
+//Greek Small Letter Alpha with Ypogegrammeni U+1FB3
+var_dump(IntlChar::istitle("ᾳ"));
+//Greek Capital Letter Alpha U+0391
+var_dump(IntlChar::istitle("Α"));
```
Girgias, line 62: Please add a space between ``//`` and the comment.

Girgias (changes requested):
LGTM other than the nit

reference/intl/intlchar/istitle.xml
leonardolara: Spaces added.

Girgias (approved):

PR #3100, opened 2024-01-10
Update output_add_rewrite_var()

haszi:
Added some details on what exactly this function does (e.g. only one instance of 'URL-Rewriter' at any given time, when rewriting occurs, how duplicate name/value pairs are processed, etc.) and reworded most of it as it was very confusing to me.

Added an additional line to the example so that it will work with the current default settings.

Removed references to sessions as these are not directly related anymore.

### DIFF
--- a/reference/outcontrol/functions/output-add-rewrite-var.xml
+++ b/reference/outcontrol/functions/output-add-rewrite-var.xml
@@ -14,27 +14,30 @@
    <methodparam><type>string</type><parameter>value</parameter></methodparam>
   </methodsynopsis>
   <para>
-   This function adds another name/value pair to the URL rewrite mechanism.
-   The name and value will be added to URLs (as GET parameter) and forms
-   (as hidden input fields) the same way as the session ID when transparent
-   URL rewriting is enabled with <link
-    linkend="ini.session.use-trans-sid">session.use_trans_sid</link>.
+   This functions starts a <literal>'URL-Rewriter'</literal> output buffer handler
+   if it has not been started yet,
+   and passes the <parameter>name</parameter>
+   and <parameter>value</parameter> parameters to it.
+   Consequent calls to <function>output_add_rewrite_var</function>
+   will pass the parameters to the existing <literal>'URL-Rewriter'</literal>.
   </para>
   <para>
-   This function's behaviour is controlled by the <link
-    linkend="ini.url-rewriter.tags">url_rewriter.tags</link> and
-    <link linkend="ini.url-rewriter.hosts">url_rewriter.hosts</link> &php.ini;
-   parameters.
+   When the output buffer is flushed
+   (by calling <function>ob_flush</function>, <function>ob_end_flush</function>,
+   <function>ob_get_flush</function> or at the end of the script)
+   the <literal>'URL-Rewriter'</literal> adds the name/value pairs
+   as query parameters to <acronym>URL</acronym>s in attributes of HTML tags
+   and adds hidden fields to forms based on the values of the
+   <link linkend="ini.url-rewriter.tags">url_rewriter.tags</link> and
+   <link linkend="ini.url-rewriter.hosts">url_rewriter.hosts</link>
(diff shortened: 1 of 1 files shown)


reference/outcontrol/functions/output-add-rewrite-var.xml
```diff
@@ -14,27 +14,30 @@
...
   <para>
-   This function adds another name/value pair to the URL rewrite mechanism.
-   The name and value will be added to URLs (as GET parameter) and forms
-   (as hidden input fields) the same way as the session ID when transparent
-   URL rewriting is enabled with <link
-    linkend="ini.session.use-trans-sid">session.use_trans_sid</link>. 
+   This functions starts a <literal>'URL-Rewriter'</literal> output buffer handler
+   if it has not been started yet,
+   and passes the <parameter>name</parameter>
+   and <parameter>value</parameter> parameters to it.
```
Girgias, line 20: What do you mean by "pass" here? Does it write this to the (new) output buffer?

reference/outcontrol/functions/output-add-rewrite-var.xml
```diff
@@ -14,27 +14,30 @@
...
+   Consequent calls to <function>output_add_rewrite_var</function>
```
Girgias, line 21: Did you mean:
```suggestion
   Consecutive calls to <function>output_add_rewrite_var</function>
```
or

```suggestion
   Follow-up calls to <function>output_add_rewrite_var</function>
```
?

reference/outcontrol/functions/output-add-rewrite-var.xml
```diff
@@ -14,27 +14,30 @@
...
+   will pass the parameters to the existing <literal>'URL-Rewriter'</literal>.
```
Girgias, line 22:
```suggestion
   will pass the parameters to the existing <literal>'URL-Rewriter'</literal> handler.
```

reference/outcontrol/functions/output-add-rewrite-var.xml
```diff
@@ -14,27 +14,30 @@
...
+   the <literal>'URL-Rewriter'</literal> adds the name/value pairs
+   as query parameters to <acronym>URL</acronym>s in attributes of HTML tags
+   and adds hidden fields to forms based on the values of the
+   <link linkend="ini.url-rewriter.tags">url_rewriter.tags</link> and
+   <link linkend="ini.url-rewriter.hosts">url_rewriter.hosts</link>
+   configuration directives.
   </para>
   <para>
-   Note that this function can be successfully called at most once per request.
+   Each name/value pair added to the <literal>'URL-Rewriter'</literal>
```
Girgias, line 36:
```suggestion
   Each name/value pair added to the <literal>'URL-Rewriter'</literal> handler
```

reference/outcontrol/functions/output-add-rewrite-var.xml
```diff
@@ -86,7 +89,7 @@
         Before PHP 7.1.0, rewrite vars set by <function>output_add_rewrite_var</function>
         use the same Session module trans sid output buffer. Since PHP 7.1.0,
         dedicated output buffer is used, <link linkend="ini.url-rewriter.tags">
-        url_rewriter.tags</link> is used solely for output functions, <link 
+        url_rewriter.tags</link> is used solely for output functions, <link
         linkend="ini.url-rewriter.tags">url_rewriter.hosts</link> is added.
```
Girgias, line 93: Could you use ``As of`` and ``Prior to`` and also put the link tags on their own lines?

Girgias (commented):
Some questions, as this I don't know this feature

reference/outcontrol/functions/output-add-rewrite-var.xml
Girgias: Ah yes ``subsequent`` would be better :)

Girgias (approved):
Much clearer thank you!

PR #3101, opened 2024-01-10
Document approach to set cookie params not known to setcookie()

morsssss:
As per [discussion in the StackOverflow chat](https://chat.stackoverflow.com/transcript/message/56990396#56990396), I added a note to the `setcookie()` doc to say that one can use `header()` to create cookies with arbitrary params. I then added an example to the `header()` doc.

I also propose a minor language fix-up.

### DIFF
--- a/reference/network/functions/header.xml
+++ b/reference/network/functions/header.xml
@@ -223,9 +223,9 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
    <example>
     <title>Setting a cookie</title>
     <para>
-      <function>setcookie</function> provides a convenient way to set cookies. 
-      To set a cookie that includes attributes which <function>setcookie</function>
-      doesn't support, <function>header</function> can be used.
+     <function>setcookie</function> provides a convenient way to set cookies. 
+     To set a cookie that includes attributes which <function>setcookie</function>
+     doesn't support, <function>header</function> can be used.
     </para>
     <para>
      For example, the following sets a cookie that includes a 
--- a/reference/network/functions/header.xml
+++ b/reference/network/functions/header.xml
@@ -219,6 +219,24 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
     </para>
    </example>
   </para>
+  <para>
+   <example>
+    <title>Setting a cookie</title>
+    <para>
+      <function>setcookie</function> provides a convenient way to set cookies. However, <function>header</function> lets you include parameters that <function>setcookie</function> may not support.
+    </para>
+    <para>
+     For example, the following sets a cookie that includes a <literal>Partitioned</literal> attribute.
+    </para>
+    <programlisting role="php">
+<![CDATA[
+<?php
+header('Set-Cookie: name=value; Secure; Path=/; SameSite=None; Partitioned;');
+?>
+]]>
+    </programlisting>
+   </example>
+  </para>
  </refsect1>
 
  <refsect1 role="notes">
--- a/reference/network/functions/header.xml
+++ b/reference/network/functions/header.xml
@@ -224,8 +224,8 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
     <title>Setting a cookie</title>
     <para>
       <function>setcookie</function> provides a convenient way to set cookies. 
-      But <function>header</function> can be used to include attributes that 
-      <function>setcookie</function> doesn't support.
+      To set a cookie that includes attributes which <function>setcookie</function>
+      doesn't support, <function>header</function> can be used.
     </para>
     <para>
      For example, the following sets a cookie that includes a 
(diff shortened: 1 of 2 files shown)


reference/network/functions/header.xml
```diff
@@ -219,6 +219,24 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
     </para>
    </example>
   </para>
+  <para>
+   <example>
+    <title>Setting a cookie</title>
+    <para>
+      <function>setcookie</function> provides a convenient way to set cookies. However, <function>header</function> lets you include parameters that <function>setcookie</function> may not support.
```
Girgias, line 226: Please do not use personalization (i.e. ``you``) in the documentation.

Also please split lines at sensible break points to, ideally, be less than 80 chars.

reference/network/functions/setcookie.xml
```diff
@@ -170,6 +170,13 @@
        <literal>samesite</literal> element is omitted, no SameSite cookie
        attribute is set.
       </para>
+      <para>
+       <note>
+        <para>
+         To set a cookie that includes parameters that aren't among the keys listed, you can use <function>header</function>.
```
Girgias, line 176: ditto

Girgias (changes requested):
Overall looks good, just wording changes

reference/network/functions/header.xml
morsssss: Sorry about the breakpoints! Fixing that now.

As far as "you" goes - the existing two examples use "you" consistently. I thought I'd try to be consistent with them, and I don't like using passive voice.

But if you're saying that none of these should be using "you", I will try to set a better example with my own example!

reference/network/functions/setcookie.xml
morsssss: Ok! Here, removing the word "you" was trivial.

reference/network/functions/header.xml
Girgias: I know there are already some places that use "you" which shouldn't (most of the docs except the tutorial and FAQ sections) but let's not introduce more of them.

reference/network/functions/header.xml
```diff
@@ -219,6 +219,27 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
     </para>
    </example>
   </para>
+  <para>
+   <example>
+    <title>Setting a cookie</title>
+    <para>
+      <function>setcookie</function> provides a convenient way to set cookies. 
+      However, <function>header</function> lets you include parameters that 
+      <function>setcookie</function> may not support.
```
Girgias, line 228: One possible amendment
```suggestion
      However, to include parameters that are not supported by
      <function>setcookie</function> the <function>header</function>
      can be used to include them.
```

reference/network/functions/header.xml
morsssss: Ok! "You" removed.

reference/network/functions/header.xml
morsssss: I tried this, took a look, and was inspired to create a variant that looked good to me... wdyt?

reference/network/functions/header.xml
```diff
@@ -219,6 +219,27 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
     </para>
    </example>
   </para>
+  <para>
+   <example>
+    <title>Setting a cookie</title>
+    <para>
+      <function>setcookie</function> provides a convenient way to set cookies. 
+      But <function>header</function> can be used to include attributes that
```
Girgias, line 227: This "but" is just confusing
```suggestion
      <function>header</function> can be used to include attributes that 
```

reference/network/functions/setcookie.xml
```diff
@@ -170,6 +170,14 @@
        <literal>samesite</literal> element is omitted, no SameSite cookie
        attribute is set.
       </para>
+      <para>
+       <note>
+        <para>
+         To set a cookie that includes parameters that aren't among the keys listed,
```
Girgias, line 176: Use attribute here too? Or keep parameters

morsssss: Good catch. Making this change.

reference/network/functions/header.xml
morsssss: I don't love it either. It's hard for me to do this without using a subject... but I think this may be more what you're looking for.

reference/network/functions/header.xml
```diff
@@ -219,6 +219,27 @@ header("Expires: Sat, 26 Jul 1997 05:00:00 GMT"); // Date in the past
...
    </example>
   </para>
+  <para>
+   <example>
+    <title>Setting a cookie</title>
+    <para>
+      <function>setcookie</function> provides a convenient way to set cookies. 
+      To set a cookie that includes attributes which <function>setcookie</function>
+      doesn't support, <function>header</function> can be used.
+    </para>
```
Girgias, line 229:
```suggestion
    <para>
     <function>setcookie</function> provides a convenient way to set cookies. 
     To set a cookie that includes attributes which <function>setcookie</function>
     doesn't support, <function>header</function> can be used.
    </para>
```

Girgias (approved):
LGTM Thank you!

morsssss:
Thanks for all your help @Girgias !

PR #3102, opened 2024-01-10
ValueError on flags diferent to JSON_INVALID_UTF8_IGNORE

jeijei4:
If flags parameter is different than the constant JSON_INVALID_UTF8_IGNORE, a ValueError exception is thrown.

### DIFF
--- a/reference/json/functions/json-validate.xml
+++ b/reference/json/functions/json-validate.xml
@@ -100,6 +100,10 @@
    If <parameter>depth</parameter> is outside the allowed range,
    a <classname>ValueError</classname> is thrown.
   </para>
+  <para>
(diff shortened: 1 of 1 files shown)


reference/json/functions/json-validate.xml
```diff
@@ -100,6 +100,10 @@
    If <parameter>depth</parameter> is outside the allowed range,
    a <classname>ValueError</classname> is thrown.
   </para>
+  <para>
+   If <parameter>flags</parameter> is different than the constant <constant>JSON_INVALID_UTF8_IGNORE</constant>,
+   a <classname>ValueError</classname> is thrown.
```
Girgias, line 105:
```suggestion
   If <parameter>flags</parameter> is not a valid flag,
   a <classname>ValueError</classname> is thrown.
```

Lets keep it generic in case a new flag is supported.

Girgias (approved):

PR #3105, opened 2024-01-11
Update ob_implicit_flush()

haszi:
Add a note that this function doesn't work with user level output handlers as this seems to be the most common misunderstanding regarding this function.


Girgias (approved):

Girgias:
Thank you!

PR #3106, opened 2024-01-12
[PHP 8.3] Update functions

saundefined:


reference/array/functions/array-sum.xml
```diff
@@ -37,6 +37,34 @@
...
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.3.0</entry>
+       <entry>
+        Now emits <constant>E_WARNING</constant> when <parameter>array</parameter> values
+        cannot be converted to &integer; or &float;.
+        Previously &array;s and &object;s where ignored whilst every other value was cast to &integer;.
+        Moreover, objects that define a numeric cast (e.g. GMP) are now cast instead of ignored.
```
Girgias, line 59:
```suggestion
        Moreover, objects that define a numeric cast (e.g. <classname>GMP</classname>) are now cast instead of ignored.
```
Ditto for the other changelog and LGTM

Girgias (approved):
One nit and LGTM

DanielRuf:
I stumbledf upon this PR during a migration to PHP 8.3:

> array_sum() and array_product() now warn when values in the array **cannot be converted to int/float.**

This seems to be not the case. As we can see at https://3v4l.org/lKNfW#v8.3.20, the value (string) `3Hallo` is converted to `3`. So either the warning `Warning: A non-numeric value encountered in` or the documentation is not right.

Since **cannot be converted to int/float.** is not the case here, does this mean the documentation is inaccurate?

DanielRuf:
I stumbled upon this PR during a migration to PHP 8.3:

> array_sum() and array_product() now warn when values in the array **cannot be converted to int/float.**

This seems to be not the case. As we can see at https://3v4l.org/lKNfW#v8.3.20, the value (string) `3Hallo` is converted to `3`. So either the warning `Warning: A non-numeric value encountered in` or the documentation is not right.

Since **cannot be converted to int/float.** is not the case here, does this mean the documentation is inaccurate?

DanielRuf:
Relevant PR: https://github.com/php/php-src/pull/10161

DanielRuf:
I think we should reword the part in the documentation, at least we should fix some grammar issues (where => were):

```diff
+ array_sum() and array_product() now warn when values in the array cannot be converted to int/float. Previously arrays and objects were ignored whilst every other value was cast to int. Moreover, objects that define a numeric cast (e.g. GMP) are now casted instead of ignored.
- array_sum() and array_product() now warn when values in the array cannot be converted to int/float. Previously arrays and objects where ignored whilst every other value was cast to int. Moreover, objects that define a numeric cast (e.g. GMP) are now casted instead of ignored.

Girgias:
Please open a new issue. Old PRs are basically never read nor the proper place to discuss stuff.

For the grammar issues feel free to open a PR.

PR #3109, opened 2024-01-13
get_defined_constants - warning about magic constants

jorgsowa:
Added warning about not returning magic constants when using function `get_defined_constants()`.


reference/info/functions/get-defined-constants.xml
```diff
@@ -94,6 +94,12 @@ Array
    Returns an array of constant name => constant value array, optionally
    groupped by extension name registering the constant.
   </para>
+
+  <warning>
+  <para>
+   Function doesn't return <link linkend="language.constants.magic">magic constants</link>.
+  </para>
+  </warning>
```
Girgias, line 102: This seems pointless? Because obviously this is not possible as those "constants" are resolved at compile time.

jorgsowa: I have a big problem with the name magic constants. It's very misleading, as those are not constants in fact. The problem touches functions `define` and `defined` too.

It's fine for me to close this PR. Maybe I can come out with a better solution.

Girgias: Sure they are closer to "macros" but the documentation has used the term magic constants for years. Changing this should require discussion on the documentation mailing list.

philip: I agree that it's worth noting, but not as a warning. How about updating the description instead?

Current description:

```xml
Returns the names and values of all the constants currently defined. This includes those 
created by extensions as well as those created with the <function>define</function> function.
```

Potentially rewrite it to include the magic constants reference, say:

```xml
Returns the names and values of all currently defined constants. This includes constants 
defined by extensions and those created using the <function>define</function> function.
However, it excludes <link linkend="language.constants.magic">magic constants</link>, 
which resolve at compile time.
```

Girgias: This description is very nice :)

PR #3112, opened 2024-01-15
Update flush() documentation

haszi:
Removed sections of web server specific description (there is plenty of these in the comments of that page), add a sentence on how functionality differs on the CLI and in a web server, added a section on flush possibly interfering with output handlers and moved the notes user level handlers and web server/browser buffering into separate `<note>` tags.

Closes https://bugs.php.net/bug.php?id=65115.


Girgias:
Normally we try to integrate notes into the documentation and not have the notes explain some of the caveats.

But I am not sure how useful most of those notes are in the first place

Girgias (approved):

Girgias:
Thank you!

PR #3113, opened 2024-01-16
imagecropauto: Fixing typo

ThomasLandauer:
I guess :-)

### DIFF
--- a/reference/image/functions/imagecropauto.xml
+++ b/reference/image/functions/imagecropauto.xml
@@ -125,7 +125,7 @@
   &reftitle.returnvalues;
   <para>
    Returns a cropped image object on success&return.falseforfailure;.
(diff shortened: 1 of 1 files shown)


reference/image/functions/imagecropauto.xml
```diff
@@ -125,7 +125,7 @@
   &reftitle.returnvalues;
   <para>
    Returns a cropped image object on success&return.falseforfailure;.
-   If the complete image was cropped, <function>imagecrop</function> returns &false;.
+   If the complete image was cropped, <function>imagecropauto</function> returns &false;.
```
Girgias, line 128: I would probably reword this completely:
```suggestion
   &false; is also returned if the whole image was cropped.
```
What do you think?

Girgias (approved):

Girgias:
Thank you!

PR #3116, opened 2024-01-17
Update array.xml: make strtoupper an mb_strtoupper

mmalferov:
Can you accept this change? This is necessary in order for the array values translated into another language to be converted to uppercase


Girgias (approved):
Doesn't look harmful for me.

Although for doc-fr we don't usually translate examples nowadays.

PR #3117, opened 2024-01-18
Make the Operators list pure

mmalferov:
### Problem
Now the list of operators looks very noisy. Take a look:

operators-eng-dirty

It's hard for the eye to catch on to the desired section. The list only shows Operators, Operators, Operators…

### Proposal
Make the list of operators (on the [Operators](https://www.php.net/manual/en/language.operators.php) page and in the [menu](https://www.php.net/manual/en/language.operators.precedence.php), which is located at the top right) pure, but keep the full titles on the pages.

operators-eng-pure

Now it takes 5 times faster to select the desired section; the eye instantly catches on to the desired operator name.

P. S. I'm not 100 % sure that I chose the right way to add a short title for the sections via the `titleabbrev` tag, maybe someone will tell me how to do it better. I'm ready.

### DIFF
--- a/language/operators/arithmetic.xml
+++ b/language/operators/arithmetic.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <sect1 xml:id="language.operators.arithmetic">
  <title>Arithmetic Operators</title>
+ <titleabbrev>Arithmetic</titleabbrev>
  <simpara>
   Remember basic arithmetic from school? These work just
   like those.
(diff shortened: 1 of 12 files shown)


saundefined (approved):
The idea came up while working on the doc-ru, so LGTM.
But I'd be glad to hear others' opinions :)

sy-records (approved):
LGTM. Good job!

Girgias (commented):
The overall change makes sense to me, just some nits. :)

Girgias (approved):

PR #3120, opened 2024-01-18
Correction to IntlDateFormatter examples

leonardolara:
reference/intl/dateformatter/get-error-code.xml
reference/intl/dateformatter/get-error-message.xml

* Examples in get-error-code.xml and get-error-message.xml use the function IntlDateFormatter::format() with less parameters than expected, both in OO and procedural. Corrections made by this commit intend to generate the expected error code and message given in the example. Tested with PHP 8.3.1.
* Rephrased purpose of function in get-error-code.xml, there was redundant information.
* Correction to identation.
* Removed trailing ws.


Girgias (approved):

Girgias:
Thank you

PR #3121, opened 2024-01-18
Clarify output in ob_implicit_flush()

### LINKED ISSUE #3115: flush() and ob_implicit_flush() behaviour in 8.4
See https://github.com/php/php-src/issues/13164#issuecomment-1894398535 and discussion leading up to this

haszi:
Closes #3115 

Clarifies that empty strings or sending headers do not trigger implicit flushing.


Girgias (approved):
Thank you!

PR #3127, opened 2024-01-20
Update expm1.xml

mmalferov:
It was ambiguous:

1) ('e' to the power of num) minus one, or
2) 'e' to the power of (num minus one)

I mean: pow(2, 3) - 1 !== pow(2, 3 -1)

### DIFF
--- a/reference/math/functions/expm1.xml
+++ b/reference/math/functions/expm1.xml
@@ -16,9 +16,9 @@
   </methodsynopsis>
   <para>
    <function>expm1</function> returns the equivalent to
-   'exp(<parameter>num</parameter>) -  1' computed in a way that is
+   <code>exp(<parameter>num</parameter>) - 1</code> computed in a way that is
    accurate even if the value of <parameter>num</parameter> is near zero,
-   a case where 'exp (<parameter>num</parameter>) - 1' would be
+   a case where <code>exp(<parameter>num</parameter>) - 1</code> would be
    inaccurate due to subtraction of two numbers that are nearly equal.
   </para>
  </refsect1>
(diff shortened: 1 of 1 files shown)


reference/math/functions/expm1.xml
```diff
@@ -40,7 +40,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   'e' to the power of <parameter>num</parameter> minus one
+   'e' to the power of <parameter>num</parameter>, minus one.
```
Girgias, line 43: I think a lot of the "code" examples in this page should be rewritten to use the ``<code>`` tag.
Or maybe even the ``<inlineequation>`` docbook tag (although that one does not have any support in PhD :/)

Also this problem probably spans the whole math/ folder :/

reference/math/functions/expm1.xml
```diff
@@ -16,9 +16,9 @@
   </methodsynopsis>
   <para>
    <function>expm1</function> returns the equivalent to
-   'exp(<parameter>num</parameter>) -  1' computed in a way that is
+   <code>exp(<parameter>num</parameter>) - 1</code> computed in a way that is
    accurate even if the value of <parameter>num</parameter> is near zero,
-   a case where 'exp (<parameter>num</parameter>) - 1' would be
+   a case where <code>exp(<parameter>num</parameter>) - 1</code> would be
```
Girgias, line 21: Does this render correctly? I never checked, but I would imagine it needs to be
```suggestion
   a case where <code>exp($<parameter>num</parameter>) - 1</code> would be
```
At least

reference/math/functions/expm1.xml
Girgias, line 43:
```suggestion
   <literal>e</literal> to the power of <parameter>num</parameter>, minus one.
```

PR #3128, opened 2024-01-21
Update is-a.xml: paramter tag instead 'this'

mmalferov:
How about adding parameter tag instead 'this'?

### DIFF
--- a/reference/classobj/functions/is-a.xml
+++ b/reference/classobj/functions/is-a.xml
@@ -54,8 +54,9 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true; if the object <parameter>object_or_class</parameter> is of <parameter>class</parameter> object type or has this object type as one of
-   its supertypes, &false; otherwise.
+   Returns &true; if <parameter>object_or_class</parameter> is the
+   <parameter>class</parameter> object type,
+   or has <parameter>class</parameter> as one of its supertypes, &false; otherwise.
   </para>
  </refsect1>
  <refsect1 role="examples">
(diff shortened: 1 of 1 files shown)


reference/classobj/functions/is-a.xml
```diff
@@ -14,7 +14,7 @@
    <methodparam choice="opt"><type>bool</type><parameter>allow_string</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Checks if the given <parameter>object_or_class</parameter> is of this object type or has
+   Checks if the given <parameter>object_or_class</parameter> is of <parameter>class</parameter> object type or has
    this object type as one of its supertypes.
```
Girgias, line 18:
```suggestion
   Determines if the given <parameter>object_or_class</parameter> is the
   <parameter>class</parameter> object type,
   or has <parameter>class</parameter> as one of its supertypes.
```

Is probably better wording.

reference/classobj/functions/is-a.xml
```diff
@@ -53,7 +54,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true; if the object is of this object type or has this object type as one of
+   Returns &true; if the object <parameter>object_or_class</parameter> is of <parameter>class</parameter> object type or has this object type as one of
    its supertypes, &false; otherwise.
```
Girgias, line 58:
```suggestion
   Returns &true; if <parameter>object_or_class</parameter> is the
   <parameter>class</parameter> object type,
   or has <parameter>class</parameter> as one of its supertypes, &false; otherwise.
```

PR #3130, opened 2024-01-21
Convert curl-getinfo() constant list to table

haszi:
Converts the `curl-getinfo()` constant list to a table that matches the formatting of tables on the following functions' pages: `curl_multi_setopt()`, `curl_setopt()` and `curl_share_setopt()`.

### DIFF
--- a/reference/curl/functions/curl-getinfo.xml
+++ b/reference/curl/functions/curl-getinfo.xml
@@ -5,7 +5,7 @@
   <refname>curl_getinfo</refname>
   <refpurpose>Get information regarding a specific transfer</refpurpose>
  </refnamediv>
- 
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
@@ -14,7 +14,7 @@
    <methodparam choice="opt"><type class="union"><type>int</type><type>null</type></type><parameter>option</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Gets information about the last transfer. 
+   Gets information about the last transfer.
   </para>
  </refsect1>
 
@@ -93,7 +93,7 @@
         </listitem>
         <listitem>
          <simpara>
-          <constant>CURLINFO_REDIRECT_URL</constant> - With the <constant>CURLOPT_FOLLOWLOCATION</constant> option disabled: redirect URL found in the last transaction, that should be requested manually next. With the <constant>CURLOPT_FOLLOWLOCATION</constant> option enabled: this is empty. The redirect URL in this case is available in <constant>CURLINFO_EFFECTIVE_URL</constant> 
+          <constant>CURLINFO_REDIRECT_URL</constant> - With the <constant>CURLOPT_FOLLOWLOCATION</constant> option disabled: redirect URL found in the last transaction, that should be requested manually next. With the <constant>CURLOPT_FOLLOWLOCATION</constant> option enabled: this is empty. The redirect URL in this case is available in <constant>CURLINFO_EFFECTIVE_URL</constant>
          </simpara>
         </listitem>
         <listitem>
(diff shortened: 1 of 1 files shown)


reference/curl/functions/curl-getinfo.xml
```diff
@@ -28,346 +28,546 @@
...
-          <constant>CURLINFO_TOTAL_TIME_T</constant> - Total time in microseconds for the previous transfer, including name resolving, TCP connect etc.
-         </simpara>
-        </listitem>
-       </itemizedlist>
+       <informaltable>
+        <tgroup cols="3">
+         <thead>
+          <row>
+           <entry valign="top">Option</entry>
+           <entry valign="top">Description</entry>
```
Girgias, line 36:
```suggestion
           <entry valign="top">&Description;</entry>
```

Girgias (commented):
Does this make sense if it is going to be converted to something else again?

Girgias (approved):

Girgias:
I'll wait a bit before merging this

Girgias:
Okay, I just had a check via the following crappy bash script:
```bash
#! /bin/bash

FILE=reference/curl/functions/curl-getinfo.xml
DIR="./*/"
for d in $DIR
do
# FAILSAFE #
# Check if "$f" FILE exists and is a regular file and then only copy it #
  if [ -f "$d$FILE" ]
  then
    echo "$FILE exists in $d."
  else
    echo "Warning: Some problem with \""$d$FILE"\""
  fi
done
```

which results in:
```
Warning: Some problem with "./base/reference/curl/functions/curl-getinfo.xml"
reference/curl/functions/curl-getinfo.xml exists in ./de/.
reference/curl/functions/curl-getinfo.xml exists in ./en/.
reference/curl/functions/curl-getinfo.xml exists in ./es/.
reference/curl/functions/curl-getinfo.xml exists in ./fr/.
reference/curl/functions/curl-getinfo.xml exists in ./it/.
reference/curl/functions/curl-getinfo.xml exists in ./ja/.
Warning: Some problem with "./output/reference/curl/functions/curl-getinfo.xml"
Warning: Some problem with "./output_bightml/reference/curl/functions/curl-getinfo.xml"
Warning: Some problem with "./phd/reference/curl/functions/curl-getinfo.xml"
Warning: Some problem with "./pl/reference/curl/functions/curl-getinfo.xml"
reference/curl/functions/curl-getinfo.xml exists in ./pt_br/.
Warning: Some problem with "./ro/reference/curl/functions/curl-getinfo.xml"
reference/curl/functions/curl-getinfo.xml exists in ./ru/.
reference/curl/functions/curl-getinfo.xml exists in ./tr/.
Warning: Some problem with "./uk/reference/curl/functions/curl-getinfo.xml"
reference/curl/functions/curl-getinfo.xml exists in ./zh/.
```

This file is up-to date for all translations, so merging this and then replicating the change can be done easily and that would be great.

The `reference/curl/constants.xml` however is outdated in doc-es and doc-tr, now doc-tr should be easy to update: http://doc.php.net/revcheck.php?p=plain&lang=tr&hbp=5bc8ebe3c1e0bb133e058f1b8eb34a721149e893&f=reference/curl/constants.xml&c=on before doing any other changes, and doc-es is so outdated that I think I will just delete this file for simplicity sake.

Girgias:
> > This file is up-to date for all translations, so merging this and then replicating the change can be done easily and that would be great.
> > The `reference/curl/constants.xml` however is outdated in doc-es and doc-tr, now doc-tr should be easy to update: http://doc.php.net/revcheck.php?p=plain&lang=tr&hbp=5bc8ebe3c1e0bb133e058f1b8eb34a721149e893&f=reference/curl/constants.xml&c=on before doing any other changes, and doc-es is so outdated that I think I will just delete this file for simplicity sake.
> 
> Right. One issue is that it's not just the constant lists that are out of sync but some of the constant lists on the functions' pages too. Of course there are some constants on the constant lists that are not on the functions' lists and vice versa. On top of that, there are missing constants and probably some that should be removed. :-)
> 
> And it's because of the above that I suggested that I would convert the `curl_getinfo()` pages to a table for each translation first, and then move all tables to the constant lists and XInclude them on the functions' pages. The only page that would actually have any constants would be the constant list and once I'm at this point, I can move constant to their appropriate tables, eliminate duplicates, remove deprecated constants and add missing ones from the cURL website. Each of these steps would be one commit in one PR per language.
> 
> Once this (`doc-en` `curl_getinfo()`) PR is merged, I'll put together a PR for `doc-en` the way I described above and we can see whether that works. What do you think?

Looks good to me :) if you could provide a patch for translations that would be great too!

leonardolara:
@haszi, would you please provide a patch for translations?

PR #3133, opened 2024-01-22
Add details on PDORow

haszi:
Adds more details to `PDORow`'s description as suggested [here](https://github.com/php/doc-en/pull/3059#pullrequestreview-1802030301).

@kamil-tekiela could you also take a look to make sure this covers everything?

### DIFF
--- a/reference/pdo/pdorow.xml
+++ b/reference/pdo/pdorow.xml
@@ -17,20 +17,11 @@
     Objects of this class cannot be instantiated and are not serializable.
    </para>
    <para>
-    Properties of this class do not store a copy of the result set
-    but provide a link to it. This can result in significantly
-    lower memory consumption than using any other fetch mode.
-   </para>
-   <para>
-    In addition to its default <property>queryString</property> property,
-    the <classname>PDORow</classname> object creates properties
-    that correspond to the column names returned in the result set
-    as they are accessed.
-   </para>
-   <para>
-    Trying to <function>unset</function> or write to any property other than
-    <varname linkend="pdorow.props.querystring">queryString</varname>
-    will throw an <classname>Exception</classname>.
+    The <classname>PDORow</classname> object allows access
+    to the returned data as if both <constant>PDO::FETCH_OBJ</constant>
+    and <constant>PDO::FETCH_BOTH</constant> mode was used.
+    This means that the returned data can be accessed as object properties,
+    and as an array both indexed by the column name and a column offset number.
    </para>
    <caution>
     <simpara>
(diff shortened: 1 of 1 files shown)


reference/pdo/pdorow.xml
```diff
@@ -13,15 +13,37 @@
     by <methodname>PDOStatement::fetch</methodname>
     called with <constant>PDO_FETCH_LAZY</constant> fetch mode.
    </para>
+   <para>
+    Objects of this class cannot be instantiated and are not serializable.
+   </para>
+   <para>
+    Properties of this class do not store a copy of the result set
+    but provide a link to it. This can result in significantly
+    lower memory consumption than using any other fetch mode.
```
kamil-tekiela, line 22: I see what you are trying to say here, but it's not entirely accurate. 
I really like how Ulf Wendel described it:
> "PDO::FETCH_LAZY = PDO::FETCH_OBJ + PDO::FETCH_BOTH"
> PDO::FETCH_BOTH means that the returned data is both indexed by the column name and a column offset number.
https://bugs.php.net/bug.php?id=44327

The point of this mode is to allow the user to fetch the values in whichever way they please. Be it indexed array, associative array, or via property access. As such, there really isn't any memory saving when compared to any of the 3 fetch modes. Well, at least not the one users may be thinking of. Technically, there could be a few bytes difference due to how ZVALs are represented. Also, if you store the `PDORow`s in separate variables, the only memory consumed will be the few bytes needed to store the empty shell of that object. The actual data isn't part of the object. 

However, the result set's **row** isn't fetched from the result set until the first access on `PDORow`. With other fetch modes, the row would be fetched immediately to construct the returned array/object, but with `LAZY` it's only fetched when required. In practice, it doesn't change anything memory-wise as PHP still has to fetch the same data, it's just happening at a different point in time. 

Another issue I have with this sentence is that `PDORow`'s properties aren't linked to the data in a way some people may think. For example, if you do something like:
```
$lazy1 = $res->fetch(PDO::FETCH_LAZY);
$lazy2 = $res->fetch(PDO::FETCH_LAZY);
echo $lazy1->username;
echo $lazy2->username;
```
You don't end up with two rows in two distinct objects. Both will only enable you to fetch the data of the second row. That's because, as you said, the data isn't stored in the object. The closest analogy is the `__get` magic method. 

For this reason, I would suggest to also change the sentence below which says *"the <classname>PDORow</classname> object creates properties that correspond to the column names returned in the result set as they are accessed"*. The properties are never created. They do not exists at any point of time. Except `queryString`. 

So, I wouldn't recommend to change the description and not mention anything about potential memory savings that do not exist. Instead, say something along the lines of what Ulf said. 


@colshrapnel FYI

kamil-tekiela (changes requested):
Also, I don't know if the information about errors should be in a separate section, something like https://www.php.net/manual/en/pdostatement.execute.php#refsect1-pdostatement.execute-errors

kamil-tekiela (approved):
LGTM.

Girgias (approved):
LGTM thank you :)

PR #3134, opened 2024-01-22
Change note to normal para in get_parent_class

kamil-tekiela:


reference/classobj/functions/get-parent-class.xml
```diff
@@ -37,11 +37,9 @@
    Returns the name of the parent class of the class of which
    <parameter>object_or_class</parameter> is an instance or the name.
   </para>
-  <note>
-   <para>
-    If the object does not have a parent or the class given does not exist &false; will be returned.
-   </para>
-  </note>
+  <para>
+   If the object does not have a parent or the class given does not exist &false; will be returned.
```
Girgias, line 41:
```suggestion
   If the object does not have a parent or the given class does not exist &false; will be returned.
```
Maybe at the same time/follow-up?

Girgias (approved):

reference/classobj/functions/get-parent-class.xml
kamil-tekiela: Doesn't this also need a comma after "does not exist"?

Girgias: Huuum, possibly?

PR #3135, opened 2024-01-23
Update array key exists key parameter for PHP 8.0.0

schulerj89:
As of 8.0.0 [array_key_exists](https://github.com/php/php-src/blob/master/ext/standard/array.c#L6652) the `$key` parameter takes mixed arguments now (string, int, bool, null, and resource). So updated the docs to reflect that.

Screenshot (5)
Added screenshot for proof

### DIFF
--- a/reference/array/functions/array-key-exists.xml
+++ b/reference/array/functions/array-key-exists.xml
@@ -72,7 +72,7 @@
      <row>
       <entry>8.0.0</entry>
       <entry>
-       The <parameter>key</parameter> parameter now takes
+       The <parameter>key</parameter> parameter now accepts
        <parameter>bool</parameter>, <parameter>float</parameter>, <parameter>int</parameter>,
        <parameter>null</parameter>, <parameter>resource</parameter>, and
        <parameter>string</parameter> as arguments.
--- a/reference/array/functions/array-key-exists.xml
+++ b/reference/array/functions/array-key-exists.xml
@@ -58,30 +58,6 @@
   </note>
  </refsect1>
 
- <refsect1 role="changelog">
-  &reftitle.changelog;
-  <informaltable>
-   <tgroup cols="2">
-    <thead>
-     <row>
-      <entry>&Version;</entry>
-      <entry>&Description;</entry>
-     </row>
-    </thead>
-    <tbody>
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-key-exists.xml
```diff
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis>
    <type>bool</type><methodname>array_key_exists</methodname>
-   <methodparam><type class="union"><type>string</type><type>int</type></type><parameter>key</parameter></methodparam>
+   <methodparam><type>mixed</type><parameter>key</parameter></methodparam>
```
Girgias, line 13: It was always mixed in reality.

Please remove the changelog.

schulerj89: Good point! Changelog removed

Girgias: Actually... it was ``string|int|null`` in PHP 7.4, and it would do the silent cast for `null` to an empty string.

PHP 8.0 changed this to allow all offset types that get implicitly coerced (so booleans, floats, and resources now are also accepted and might emit the corresponding warnings). So it turns out it does need a changelog, and the type actually is ``string|int|float|bool|resource|null``.

Apologies for the mistake

schulerj89: No worries! I'll make the updates and push up.

schulerj89: Updated and attached a screenshot to show what the change will look like. Let me know.

Screenshot (6)

reference/array/functions/array-key-exists.xml
```diff
@@ -58,6 +58,31 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       The <parameter>key</parameter> parameter now takes
```
Girgias, line 75:
```suggestion
       The <parameter>key</parameter> parameter now accepts
```

Girgias (approved):
Thank you!

PR #3136, opened 2024-01-23
Add description to INI mode constants

haszi:
Adds descriptions to each `INI` mode constant, copied from their `PHP_INI` aliases.

These are also on the [missing constants list](https://github.com/php/doc-en/issues/2747) which I'm guessing is because these are in a `table` instead of a `variablelist`.

### DIFF
--- a/reference/info/constants.xml
+++ b/reference/info/constants.xml
@@ -279,7 +279,6 @@
   The following constants are only available if the host operating
   system is Windows, and can tell different versioning information
   so its possible to detect various features and make use of them.
-  They are all available as of PHP 5.3.0.
  </simpara>
  <table>
   <title>Windows specific constants</title>
--- a/reference/info/constants.xml
+++ b/reference/info/constants.xml
@@ -157,35 +157,56 @@
  </table>
 
  <table>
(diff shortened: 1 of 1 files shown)


Girgias:
Aaaaaaaaaaaaaaaaaaaaaaaaa

Well this is... unfortunate. However, as those are easily machine fixable I think it should be done, but the commit should have a ``[skip-revcheck]`` tag and the same script should be applied to all translations. Might as well also add the missing `<constant>` tags if possible.

But this should be a separate PR.

reference/info/constants.xml
```diff
@@ -259,9 +276,9 @@
  </table>
 
  <simpara>
-  The following constants are only available if the host operating 
-  system is Windows, and can tell different versioning information 
-  so its possible to detect various features and make use of them. 
+  The following constants are only available if the host operating
+  system is Windows, and can tell different versioning information
+  so its possible to detect various features and make use of them.
   They are all available as of PHP 5.3.0.
```
Girgias, line 282: Please remove this while at it, we assume the doc base level is PHP 7.0.0.

Girgias (approved):

PR #3138, opened 2024-01-24
Add DOM\DocumentType documentation

haszi:
Adds documentation for the DOM\DocumentType class. Part of #2162.

These classes under the `DOM` namespace seem to be the first ones that are aliases of other classes. As such, I didn't know what tag to use to indicate that this is an alias (`<refpurpose>` used for functions doesn't work here) so I added it as the first sentence of the intro. I've tried using `<subtitle>` but Phd gave me a `"No mapper found for 'subtitle'"` error.

I've `XInclude`d as much as could from `DOMDocumentType`. I've tried to `XInclude` the properties' `<modifier>`s and `<type>`s but ended up with no space between the words (e.g `"publicreadonlystring"`). I've tried `string-join()` but that didn't work so I'm guessing Phd doesn't support XPath 2.

The below is what I used which ended up being `"publicreadonlystring"` instead of `"public readonly string"`. Any help to achieve that would be appreciated.
```xml
<xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('domdocumenttype.synopsis')/db:classsynopsis/db:fieldsynopsis[descendant::db:varname[text()='name']]/*[not(name()='varname')])">
 <xi:fallback/>
</xi:include>
```

### DIFF
--- a/reference/dom/book.xml
+++ b/reference/dom/book.xml
@@ -19,6 +19,8 @@
  &reference.dom.constants;
  &reference.dom.examples;
 
+ &reference.dom.dom.documenttype;
+
  &reference.dom.domattr;
  &reference.dom.domcdatasection;
  &reference.dom.domcharacterdata;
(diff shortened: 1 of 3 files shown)


nielsdos (commented):
Thanks for tackling this!

Indeed, these are the first internal aliases, but they may not be for much longer...
Making them real classes was out of scope for the HTML5 RFC that introduced these class aliases in the first place. However, my new RFC expands upon this and makes them real classes: https://wiki.php.net/rfc/opt_in_dom_spec_compliance.

So I would hold off on trying to document these classes until that RFC gets accepted.

nielsdos:
Okay, no worries!
And thanks for your contributions to the docs so far :slightly_smiling_face:

PR #3143, opened 2024-01-25
Clarify support for caching_sha2_password

### LINKED ISSUE #3141: MySQLi support for caching_sha2_password
From manual page: https://php.net/mysqli.requirements

---

The manual page says conflicting things. 

"caching_sha2_password, a plugin that is not recognized by the older PHP (mysqlnd) releases"

"The caching_sha2_password plugin will be supported in a future PHP release"

Is it supported or not? The following seems to indicate that 7.4 supports caching_sha2_password.
https://stackoverflow.com/questions/59231143/does-mysqli-have-support-for-caching-sha2-password-in-php-7-4

That would be backed up by this bug report that seems to fix problems with caching_sha2_password in PHP 7.4
https://bugs.php.net/bug.php?id=79011

kamil-tekiela:
See https://bugs.php.net/bug.php?id=79275
Fixes #3141

### DIFF
--- a/reference/mysqli/setup.xml
+++ b/reference/mysqli/setup.xml
@@ -28,7 +28,7 @@
    not recognized by the older PHP (mysqlnd) releases. Instead, change it by
    setting <literal>default_authentication_plugin=mysql_native_password</literal>
    in <filename>my.cnf</filename>. The <emphasis>caching_sha2_password</emphasis>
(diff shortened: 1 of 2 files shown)


Girgias (approved):
This also fixes #3144 right?

kamil-tekiela:
Yes, it does.

PR #3146, opened 2024-01-25
Update functions - nullsafety argument.

SVGAnimate:
Where does PHP introduce typed arguments?

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -180,12 +180,14 @@ function recursion($a)
    </simpara>
 
    <para>
-    PHP supports passing arguments by value (the default), <link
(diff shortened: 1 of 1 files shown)


Girgias (changes requested):
Type declarations are already documented on the specific manual page, so redescripting nullable types here is really not needed.

I know this whole sections needs rewriting anyway, and was planning on doing that at one moment. Because here you are mixing thing about _declaring_ functions and _calling_ functions which really should not be described at the same point.

SVGAnimate:
@Girgias thank you for your reply.

> [Type declarations](https://www.php.net/manual/en/language.types.declarations.php) are already documented

I didn't find it...

Thanks for your help

PR #3148, opened 2024-01-26
Update ob_start() docs

haszi:
Updated `ob_start()` by removing inaccuracies and unnecessary details focused on buffering (instead of on the function) and linked to the appropriate sections of the new `user-level-output-buffers.xml` file.


Girgias (approved):
Thank you, looks good :)

PR #3152, opened 2024-01-28
Fix return value of property_exists

### LINKED ISSUE #3069: property-exists function should specify that it can return a null value in its declaration?
From manual page: https://php.net/function.property-exists

---

image

image

From what I understand the function can return **null** but the declaration doesn't say so. So in an IDE we don't have the info in the @return part: 

image

kamil-tekiela:
Fixes #3069


Girgias (approved):

Girgias:
Thank you!

PR #3164, opened 2024-02-02
Cleanup of GLOB constants

haszi:
Adds the missing GLOB_ERR constant, adds descriptions to each GLOB constant and `XInclude`s them on the `glob()` function's page.


Girgias (approved):
Looks reasonable

PR #3165, opened 2024-02-02
Add missing PASSWORD constants

haszi:
Adds two missing PASSWORD_ constants. 

These are the last two constants that were missing due to them being registered in a `c` instead of a `stub` file. I think a few of those ~30 constants could be registered through a `stub` file.


Girgias (approved):
Yes, those should be moved to the stub file, as we use those to check/update the documentation noadays.

PR #3168, opened 2024-02-07
Add DigitalOcean section in Cloud Hosting installation section

pronskiy:
As php.net and manual are hosted for years on DigitalOcean, it would be good to mention it in this section.


Girgias (commented):
Looks mostly fine other than the initial paragraph that reads weirdly to me

PR #3176, opened 2024-02-12
PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT is mysqlnd only

staabm:
note that `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` is [only available with mysqlnd](https://github.com/php/php-src/blob/87edeed3b9fcb9d066d8e18b10fb0ed8cb976873/ext/pdo_mysql/pdo_mysql.c#L143-L145).

Recently we ran into a problem that this constant requires a `defined` check, as its not always defined.

### DIFF
--- a/reference/pdo_mysql/constants.xml
+++ b/reference/pdo_mysql/constants.xml
@@ -239,7 +239,7 @@ foreach ($unbufferedResult as $row) {
    </term>
    <listitem>
     <para>
-     Provides a way to disable verification of the server SSL certificate.
+     Provides a way to disable verification of the server SSL certificate (requires mysqlnd).
     </para>
     <para>
      &version.exists.asof; 7.0.18 and PHP 7.1.4.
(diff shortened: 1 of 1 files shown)


reference/pdo_mysql/constants.xml
```diff
@@ -239,7 +239,7 @@ foreach ($unbufferedResult as $row) {
    </term>
    <listitem>
     <para>
-     Provides a way to disable verification of the server SSL certificate.
+     Provides a way to disable verification of the server SSL certificate (requires mysqlnd).
```
kamil-tekiela, line 242:
```suggestion
     Provides a way to disable verification of the server SSL certificate. This option is available only with mysqlnd.
```

staabm: 👍

kamil-tekiela: Sorry for the small nit, but could you add a line break between the sentences and put the new sentence on a new line?

staabm: sure - hopefully I did it correctly. or do I need a `<br>`?

kamil-tekiela (approved):
Thank you!

PR #3177, opened 2024-02-12
Improve description of defined

### LINKED ISSUE #3155: Documentaition - I think the description of defined in php.net is to check if the given string is defined as a constant.
I think the description of defined in php.net is to check if the given string is defined as a constant.

php.net has the following explanatory text,

> defined - Checks whether a given named constant exists

You may mistake it for the following.
```
if (defined(TEST)) {
    echo TEST; }
}
```

Of course, the following is written as a string, but it may take some people like me a while to realize it.

> defined(string $constant_name): bool

https://www.php.net/manual/en/function.defined.php

kamil-tekiela:
Fixes #3155


Girgias (approved):
LGTM

PR #3179, opened 2024-02-13
Replace deprecated `${var}` string interpolations with `{$var}`

takaram:
There have remained some code samples with `${var}` style interpolation which was deprecated since PHP 8.2.
This PR fixed them.

`language/types/string.xml` is going to be edited by #2000, so this PR does not change the file.


Girgias (approved):
Thank you!

PR #3181, opened 2024-02-13
Fix comment of mysqli_fetch_object

### LINKED ISSUE #3153: Incorrect comment "/* fetch associative array */"
From manual page: https://php.net/mysqli-result.fetch-object
---
There is code in example "Procedural style". The code has `/* fetch associative array */` comment. It is wrong comment. 
It will be correct like this `/* fetch object array */`

Fan2Shrek:
Fixes #3153 

Change the comment of procedural style example from `associative` to `object`


kamil-tekiela:
Nice catch! However, this comment and the other one are both useless. They don't say anything useful about the code. Could you please remove both of them?

kamil-tekiela:
Thank you.

PR #3182, opened 2024-02-14
Improve documentation about fnmatch wildcards

jawira:
I added a table with details about `fnmatch` wildcards.

### DIFF
--- a/reference/filesystem/functions/fnmatch.xml
+++ b/reference/filesystem/functions/fnmatch.xml
@@ -28,7 +28,75 @@
      <term><parameter>pattern</parameter></term>
      <listitem>
       <para>
-       The shell wildcard pattern.
+       The <parameter>pattern</parameter> to match against. Usually, the <parameter>pattern</parameter> will contain
+       wildcards like <literal>'?'</literal> and <literal>'*'</literal>.
+       <table>
+        <title>
+         Wildcards to be used in <parameter>pattern</parameter> parameter
+        </title>
+        <tgroup cols="2">
+         <thead>
+          <row>
+           <entry>Wildcard</entry>
+           <entry>Description</entry>
+          </row>
+         </thead>
+         <tbody>
+          <row>
+           <entry>
+            <literal>?</literal>
+           </entry>
+           <entry>
+            Question mark will match any single character.
+            For example, pattern "<literal>file?.txt</literal>" will match "<literal>file1.txt</literal>" and
+            "<literal>fileA.txt</literal>", but will not match "<literal>file10.txt</literal>".
+           </entry>
+          </row>
+          <row>
+           <entry>
+            <literal>*</literal>
+           </entry>
+           <entry>
+            Asterisk will match zero or more characters.
+            For example, pattern "<literal>foo*.xml</literal>" will match "<literal>foo.xml</literal>" and
+            "<literal>foobar.xml</literal>".
+           </entry>
+          </row>
+          <row>
+           <entry>
+            <literal>[ ]</literal>
+           </entry>
+           <entry>
+            Square brackets are used to create ranges or sets of characters.
+            For example, pattern "<literal>index.php[45]</literal>" will match "<literal>index.php4</literal>" and
+            "<literal>index.php5</literal>", but will not match "<literal>index.phpt</literal>".
+            Well known ranges are <literal>[0-9]</literal>,<literal>[a-z]</literal>, and <literal>[A-Z]</literal>.
+            You can use multiple sets and ranges at once, for example <literal>[0-9a-zABC]</literal>.
+           </entry>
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/fnmatch.xml
```diff
@@ -28,7 +28,75 @@
...
+       wildcards like <literal>'?'</literal> and <literal>'*'</literal>.
+       <table>
+        <title>
+         Wildcards to be used in <parameter>pattern</parameter> parameter
+        </title>
+        <tgroup cols="2">
+         <thead>
+          <row>
+           <entry>Wildcard</entry>
+           <entry>Description</entry>
```
Girgias, line 41:
```suggestion
           <entry>&Description;</entry>
```

Girgias (commented):
Doesn't this depend on the shell/OS?

jawira:
> Doesn't this depend on the shell/OS?

I don't think so, patterns seems to be the same in three test files:

* https://github.com/php/php-src/blob/master/ext/standard/tests/file/fnmatch_basic.phpt
* https://github.com/php/php-src/blob/master/ext/standard/tests/file/fnmatch_maxpathlen.phpt
* https://github.com/php/php-src/blob/master/ext/standard/tests/file/fnmatch_variation.phpt

reference/filesystem/functions/fnmatch.xml
```diff
@@ -28,7 +28,75 @@
...
+          </row>
+          <row>
+           <entry>
+            <literal>[ ]</literal>
+           </entry>
+           <entry>
+            Square brackets are used to create ranges or sets of characters.
+            For example, pattern "<literal>index.php[45]</literal>" will match "<literal>index.php4</literal>" and
+            "<literal>index.php5</literal>", but will not match "<literal>index.phpt</literal>".
+            Well known ranges are <literal>[0-9]</literal>,<literal>[a-z]</literal>, and <literal>[A-Z]</literal>.
```
Girgias, line 73: Question, what happens if I do a "non-sensical" range like `[+-/]` ? does it match a range of ASCII codepoints?

Also, I'm guessing this only works with ASCII and not UTF-8 and one cannot create a range like `[α-ω]`

Girgias (commented):
Some more questions but looks pretty good thank you!

reference/filesystem/functions/fnmatch.xml
jawira: Yes, "non-sensical" ranges will work, for example the following code will return `true` because digits (0-9) are between `+` and `=` in the ascii table.

```php
fnmatch('hell[+-=]', 'hell0'); // true
```

According to my tests ranges only work with ascii characters, which is logical since this is a pretty old function (it exists since php4). However star `*` will match anything, including unicode characters:

```php
fnmatch('foo*bar', 'foo🔥bar'); // true
```

reference/filesystem/functions/fnmatch.xml
```diff
@@ -28,7 +28,75 @@
...
+         </thead>
+         <tbody>
+          <row>
+           <entry>
+            <literal>?</literal>
+           </entry>
+           <entry>
+            Question mark will match any single character.
+            For example, pattern "<literal>file?.txt</literal>" will match "<literal>file1.txt</literal>" and
+            "<literal>fileA.txt</literal>", but will not match "<literal>file10.txt</literal>".
```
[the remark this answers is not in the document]
jawira: Got it, I will fix that.

reference/filesystem/functions/fnmatch.xml
```diff
@@ -28,7 +28,75 @@
...
+            For example, pattern <literal>"foo*.xml"</literal> will match <literal>"foo.xml"</literal> and
+            <literal>"foobar.xml"</literal>.
+           </entry>
+          </row>
+          <row>
+           <entry>
+            <literal>[ ]</literal>
+           </entry>
+           <entry>
+            Square brackets are used to create ranges or sets of characters.
```
Girgias, line 70:
```suggestion
            Square brackets are used to create ranges of ASCII codepoints or sets of characters.
```
Maybe to clarify the nonsensical ranges bit, not sure it makes sense to have the example for `fnmatch('hell[+-=]', 'hell0'); // true`

Girgias (approved):
Final comment about ASCII ranges and I think this is good to merge. Thank you!

reference/filesystem/functions/fnmatch.xml
jawira: I think the better is not to mention "nonsensical ranges", most of the people are used to "normal" ranges due to regex and talking about this will create confusion.

Girgias (approved):

Girgias:
Thank you!

PR #3185, opened 2024-02-16
[PHP 8.3] PDO_ODBC_TYPE

saundefined:


Girgias (commented):
@NattyNarwhal do you know what this new constant does?

NattyNarwhal:
IIRC it indicates the kind of ODBC library/driver manager; `PHP_ODBC_TYPE` just renames `ODBC_TYPE`, which is defined in `ext/odbc/php_odbc_includes.h`. Should be defined to i.e. `Win32` or `unixODBC`.

Adjacently, I am curious if anyone's using anything other than unixODBC/Win32/maybe iODBC though. (This includes SAP DB/Empress/Solid/Db2/Adabas/etc.) A lot are essentially linking against an ODBC driver (or SQL/CLI library) directly without any driver manager, which is a bit odd. At least for Db2, there's also a dedicated extension for that which I'm also maintaining, but I'm on the i side and not z/LUW, so maybe there's some difference there.

jhdxr (approved):
LGTM

PR #3186, opened 2024-02-17
[PHP 8.3] Update `get_class` and `get_parent_class`

saundefined:


reference/classobj/functions/get-class.xml
```diff
@@ -78,6 +74,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.3.0</entry>
+       <entry>
+        Now call without an argument emits an <constant>E_DEPRECATED</constant> warning;
+        previously, calling this function inside a class returned the name of that class.
```
Girgias, line 81:
```suggestion
        Calling <function>get_class</function> without an argument now emits an
        <constant>E_DEPRECATED</constant> warning;
        previously, calling this function inside a class returned the name of that class.
```

reference/classobj/functions/get-parent-class.xml
```diff
@@ -57,6 +53,13 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.3.0</entry>
+      <entry>
+       Now call without an argument emits an <constant>E_DEPRECATED</constant> warning;
+       previously, calling this function inside a class returned the parent class name of that class.
```
Girgias, line 60: Ditto

Girgias (changes requested):

Girgias (approved):

reference/classobj/functions/get-parent-class.xml
```diff
@@ -57,6 +53,14 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.3.0</entry>
+      <entry>
+       Calling <function>get_class</function> without an argument now emits an
```
leonardolara, line 59: Shouldn't it be "get-parent-class" ?

PR #3194, opened 2024-02-19
Fix return type of time()

JasperL-2157878:
Documentation displayed: time(): intvoid
Removed the unnecessary </void> tag.


Girgias:
That's a bug with the Render since: https://github.com/php/phd/commit/45cffd68d3afe5c8be101c66fbb03e5b5909970a

Girgias:
Related #3197

PR #3199, opened 2024-02-21
Sync some methodsynopses with stubs

kocsismate:
I intended to fix some of the methodsynopses which didn't need a changelog.

### DIFF
--- a/reference/errorfunc/functions/trigger-error.xml
+++ b/reference/errorfunc/functions/trigger-error.xml
@@ -83,7 +83,7 @@
      <row>
       <entry>8.0.0</entry>
       <entry>
-       The function now throws a <classname>ValueError</classname> if wrong
+       The function now throws a <classname>ValueError</classname> if an invalid
        <parameter>error_level</parameter> is specified. Previously, it returned &false;.
       </entry>
      </row>
--- a/reference/errorfunc/functions/trigger-error.xml
+++ b/reference/errorfunc/functions/trigger-error.xml
@@ -55,9 +55,41 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   This function returns &false; if wrong <parameter>error_level</parameter> is
-   specified, &true; otherwise.
+   &return.true.always;
   </para>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   This function throws a <classname>ValueError</classname> if
+   <parameter>error_level</parameter> is not one of
+   <constant>E_USER_ERROR</constant>, <constant>E_USER_WARNING</constant>,
+   <constant>E_USER_NOTICE</constant>, <constant>E_USER_DEPRECATED</constant>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       The function now throws a <classname>ValueError</classname> if wrong
+       <parameter>error_level</parameter> is specified. Previously, it returned &false;.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
  </refsect1>
 
  <refsect1 role="examples">
--- a/reference/errorfunc/functions/trigger-error.xml
+++ b/reference/errorfunc/functions/trigger-error.xml
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>bool</type><methodname>trigger_error</methodname>
+   <type>true</type><methodname>trigger_error</methodname>
    <methodparam><type>string</type><parameter>message</parameter></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>error_level</parameter><initializer><constant>E_USER_NOTICE</constant></initializer></methodparam>
   </methodsynopsis>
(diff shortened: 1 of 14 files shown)


reference/intl/intlgregoriancalendar/createfromdate.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis role="IntlGregorianCalendar">
-   <modifier>public</modifier> <type>static</type><methodname>IntlGregorianCalendar::createFromDate</methodname>
+   <modifier>public</modifier> <modifier>static</modifier> <type>static</type><methodname>IntlGregorianCalendar::createFromDate</methodname>
```
nielsdos, line 11:
```suggestion
   <modifier>public</modifier> <modifier>static</modifier> <methodname>IntlGregorianCalendar::createFromDate</methodname>
```

nielsdos (changes requested):
Looks mostly right, thx

reference/intl/intlgregoriancalendar/createfromdate.xml
kocsismate: Hm, It's a static method with a static return type ^^

nielsdos: Ah wow indeed. Sorry, then this is right.

nielsdos (approved):
My mistake, LGTM. Thanks.

reference/errorfunc/functions/trigger-error.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>bool</type><methodname>trigger_error</methodname>
+   <type>true</type><methodname>trigger_error</methodname>
```
Girgias, line 12: Only as of PHP 8.0, it used to return false for invalid error levels.

reference/network/functions/long2ip.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type class="union"><type>string</type><type>false</type></type><methodname>long2ip</methodname>
+   <type>string</type><methodname>long2ip</methodname>
```
Girgias, line 12: This is a change only on master.

reference/stream/functions/stream-context-set-options.xml
```diff
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>bool</type><methodname>stream_context_set_options</methodname>
+   <type>true</type><methodname>stream_context_set_options</methodname>
```
Girgias, line 11: This looks wrong, also doesn't seem to have a stub?

reference/stream/functions/stream-context-set-params.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>bool</type><methodname>stream_context_set_params</methodname>
+   <type>true</type><methodname>stream_context_set_params</methodname>
```
Girgias, line 12: This also seems wrong?

reference/outcontrol/functions/ob-start.xml
```diff
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis>
    <type>bool</type><methodname>ob_start</methodname>
-   <methodparam choice="opt"><type>callable</type><parameter>callback</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>callable</type><type>null</type></type><parameter>callback</parameter><initializer>&null;</initializer></methodparam>
```
Girgias, line 13: Welp... time to add to my list to refactor this function to use the ZPP callable check...

reference/stream/functions/stream-context-set-options.xml
kocsismate: there is a stub for it: https://github.com/php/php-src/blob/e630aacf795dfbfd875a624856674e29de32c42a/ext/standard/basic_functions.stub.php#L3398

reference/stream/functions/stream-context-set-params.xml
kocsismate: They have always (?) returned true. The return type is actually true since my refactoring: https://github.com/php/php-src/pull/12720

reference/network/functions/long2ip.xml
kocsismate: Yes, but it couldn't return false before as well according to https://github.com/php/php-src/pull/13395
So we can then document it, right?

reference/outcontrol/functions/ob-start.xml
kocsismate: Would be nice :)

reference/errorfunc/functions/trigger-error.xml
kocsismate: Your are right, I've just added an errors and a changelog section!

reference/stream/functions/stream-context-set-params.xml
Girgias: Right, it used to return false on invalid stream context, which is a ZPP failure effectively

reference/network/functions/long2ip.xml
Girgias: Ah yeah that's fine

reference/errorfunc/functions/trigger-error.xml
```diff
@@ -55,11 +55,43 @@
...
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       The function now throws a <classname>ValueError</classname> if wrong
+       <parameter>error_level</parameter> is specified. Previously, it returned &false;.
```
Girgias, line 87:
```suggestion
       The function now throws a <classname>ValueError</classname> if an invalid
       <parameter>error_level</parameter> is specified. Previously, it returned &false;.
```

Girgias (approved):
One nit, LGTM otherwise

PR #3206, opened 2024-02-27
Fix error constant IDs

haszi:
The script (which can be applied to all translations) will do the following:

 - remove IDs from error constants on the core constants page
 - standardize constant IDs on the error handling constant page


haszi:
I have just pushed a commit that has the changes generated by the script.

Girgias (commented):
Ran this on all translations, need to merge this with `[skip-revcheck]` commit without script

PR #3208, opened 2024-02-28
Update magic constant IDs

haszi:
Update magic constant IDs to "standard" global ID format so these can automatically be linked to by Phd.

I've also added a separate commit with the bash script that I made these changes with so that these can be applied to all translations without burdening the maintainers with manually doing these changes.


haszi:
The update script is as follows:

```bash
#! /bin/bash

# replace <link> with <constant>
\grep -lrE --include='*.xml' '<link linkend="constant\.trait">__TRAIT__<\/link>' | xargs -d '\n' sed -i 's/<link linkend="constant\.trait">__TRAIT__<\/link>/<constant>__TRAIT__<\/constant>/g'

# rename IDs to standard global constant ID format
\grep -lrE --include='*.xml' '"constant\.line"' | xargs -d '\n' sed -i 's/"constant\.line"/"constant\.--line--"/g'
\grep -lrE --include='*.xml' '"constant\.file"' | xargs -d '\n' sed -i 's/"constant\.file"/"constant\.--file--"/g'
\grep -lrE --include='*.xml' '"constant\.dir"' | xargs -d '\n' sed -i 's/"constant\.dir"/"constant\.--dir--"/g'
\grep -lrE --include='*.xml' '"constant\.function"' | xargs -d '\n' sed -i 's/"constant\.function"/"constant\.--function--"/g'
\grep -lrE --include='*.xml' '"constant\.class"' | xargs -d '\n' sed -i 's/"constant\.class"/"constant\.--class--"/g'
\grep -lrE --include='*.xml' '"constant\.trait"' | xargs -d '\n' sed -i 's/"constant\.trait"/"constant\.--trait--"/g'
\grep -lrE --include='*.xml' '"constant\.method"' | xargs -d '\n' sed -i 's/"constant\.method"/"constant\.--method--"/g'
\grep -lrE --include='*.xml' '"constant\.namespace"' | xargs -d '\n' sed -i 's/"constant\.namespace"/"constant\.--namespace--"/g'

```

language/constants.xml
```diff
@@ -292,13 +292,13 @@ echo ANIMALS[1]; // outputs "cat"
          does not have a trailing slash unless it is the root directory.
         </entry>
        </row>
-       <row xml:id="constant.function">
+       <row xml:id="constant.--function--">
         <entry><constant>__FUNCTION__</constant></entry>
```
Girgias, line 296: I think I would prefer if PhD would convert ``__`` (a double underscore) at the beginning and end of constants to empty string so the ID don't need to be this weird (probably `trim('_')` would just work.

haszi: Then it would be easier to just close this PR. I'll open a new one just for dropping that one link to `__TRAIT__`, I'll remove the dashes from the FFI PR and submit a new PR to remove them from `__COMPILER_HALT_OFFSET__`'s ID too. And of course I'll add the `trim('_')` to Phd, as you suggested.

Girgias:
Closed as discussed for a different solution

PR #3209, opened 2024-02-28
Update Parle constant IDs

haszi:
Update Parle constant IDs to the "standard" global and class constant ID formats.


Girgias (approved):
Ran script on translations

haszi:
The script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"parle\.constant\.utf32"' | xargs -d '\n' sed -i 's/"parle\.constant\.utf32"/"constant\.parle-internal-utf32"/g'
\grep -lrE --include='*.xml' '"parle(.)+\.constants\.flag-regex-' | xargs -d '\n' sed -i 's/\.constants\.flag-regex-/\.constants\./g'
```

PR #3210, opened 2024-02-28
Update expect constant IDs

haszi:
Update expect constant IDs to the "standard" global constant ID format.


haszi:
The script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"constants\.expect\.(.)+"' | xargs -d '\n' sed -i 's/"constants\.expect\./"constant\./g'
```

PR #3211, opened 2024-02-28
Update gender constant IDs

haszi:
Update gender constant IDs to the "standard" class constant ID format.


Girgias (approved):
Ran on translations

haszi:
The script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"gender\.constants\.(.)+"' | xargs -d '\n' sed -i 's/"gender\.constants\./"gender-gender\.constants\./g'
```

PR #3212, opened 2024-02-28
Update SQLite, Zookeeper and SNMP constant IDs

haszi:
Update SQLite, Zookeeper and SNMP constant IDs to the "standard" class constant ID format.


haszi:
The bash script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"(.)+\.class\.constants\.(.)+"' | xargs -d '\n' sed -i 's/\.class\.constants\./\.constants\./g'

```

Girgias (approved):
Ran on translations

PR #3213, opened 2024-02-28
Update APCu constant IDs

haszi:
Update Parle constant IDs to the "standard" global constant ID format.


Girgias (approved):
Ran on translations

haszi:
The script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"constant\.apcu-' | xargs -d '\n' sed -i 's/"constant\.apcu-/"constant\.apc-/g'
```

PR #3214, opened 2024-02-28
Update FFI constant ID

haszi:
Update FFI constant ID to the "standard" class constant ID format.

### DIFF
--- a/reference/ffi/ffi.xml
+++ b/reference/ffi/ffi.xml
@@ -49,7 +49,7 @@
      <modifier>public</modifier>
      <modifier>const</modifier>
      <type>int</type>
-     <varname linkend="ffi-ffi.constants.biggest-alignment">FFI::__BIGGEST_ALIGNMENT__</varname>
+     <varname linkend="ffi.constants.--biggest-alignment--">FFI::__BIGGEST_ALIGNMENT__</varname>
     </fieldsynopsis>
 
     <classsynopsisinfo role="comment">&Methods;</classsynopsisinfo>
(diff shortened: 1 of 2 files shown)


haszi:
The bash script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"ffi-ffi\.constants\.biggest-alignment"' | xargs -d '\n' sed -i 's/"ffi-ffi\.constants\.biggest-alignment"/"ffi\.constants\.--biggest-alignment--"/g'

```

reference/ffi/ffi.xml
```diff
@@ -64,7 +64,7 @@
   <section xml:id="ffi-ffi.constants">
    &reftitle.constants;
    <variablelist>
-    <varlistentry xml:id="ffi-ffi.constants.biggest-alignment">
+    <varlistentry xml:id="ffi.constants.--biggest-alignment--">
```
Girgias, line 67: Ditto same as magic constants, also the `ffi-ffi` is think is used to distinguish between global FFI constants and constants on the FFI class

haszi: This is a constant of the `FFI` class that's why it's in the `classname.constants.constant-name` format. Global (non-class) constant IDs are in the `constant.constant-name` format.

Looking at the `stub` and `arginfo` files, FFI doesn't seem to have any global (non-class) constants.

Girgias: Ah right, okay that makes sense

haszi: And this is the script for the latest ID:

```bash
\grep -lrE --include='*.xml' '"ffi-ffi\.constants\.biggest-alignment"' | xargs -d '\n' sed -i 's/"ffi-ffi\.constants\.biggest-alignment"/"ffi\.constants\.biggest-alignment"/g'
```

Girgias (approved):

PR #3215, opened 2024-02-28
Update Radius constant ID

haszi:
Update Radius constant ID to the "standard" global constant ID format.


Girgias (approved):
Ran on translations

haszi:
The script:

```bash
#! /bin/bash

\grep -lrE --include='*.xml' '"radius\.constants\.radius-mppe-key-len"' | xargs -d '\n' sed -i 's/"radius\.constants\.radius-mppe-key-len"/"constant.radius-mppe-key-len"/g'
```

Issue #3198, opened 2024-02-20
Strstr() Usage Note


Sunthief:
From manual page: https://php.net/function.strstr

---

> Note:
If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function [strpos()](https://www.php.net/manual/en/function.strpos.php) instead.

This should probably be updated to include mention of `str_contains()`.

PR #3235, opened 2024-03-03
Upgrade note on strstr functions

Fan2Shrek:
See #3198

### DIFF
--- a/reference/strings/functions/strstr.xml
+++ b/reference/strings/functions/strstr.xml
@@ -27,7 +27,7 @@
   </note>
   <note>
    <para>
-    If it is only requiered to determine if a particular <parameter>needle</parameter>
+    If it is only required to determine if a particular <parameter>needle</parameter>
     occurs within <parameter>haystack</parameter>, the faster and less memory
     intensive <function>str_contains</function> function should be used instead.
    </para>
--- a/reference/strings/functions/strstr.xml
+++ b/reference/strings/functions/strstr.xml
@@ -27,9 +27,9 @@
   </note>
   <note>
    <para>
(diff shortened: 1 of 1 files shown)


reference/strings/functions/strstr.xml
```diff
@@ -29,7 +29,7 @@
    <para>
     If you only want to determine if a particular <parameter>needle</parameter>
     occurs within <parameter>haystack</parameter>, use the faster and less memory
```
Girgias, line 31:
```suggestion
    If it is only requiered to determine if a particular <parameter>needle</parameter>
    occurs within <parameter>haystack</parameter>, the faster and less memory
```

reference/strings/functions/strstr.xml
```diff
@@ -29,7 +29,7 @@
...
-    intensive function <function>strpos</function> instead.
+    intensive function <function>str_contains</function> instead.
```
Girgias, line 32:
```suggestion
    intensive <function>str_contains</function> function should be used instead.
```

reference/strings/functions/strstr.xml
```diff
@@ -27,9 +27,9 @@
   </note>
   <note>
    <para>
-    If you only want to determine if a particular <parameter>needle</parameter>
-    occurs within <parameter>haystack</parameter>, use the faster and less memory
-    intensive function <function>strpos</function> instead.
+    If it is only requiered to determine if a particular <parameter>needle</parameter>
```
Girgias, line 30:
```suggestion
    If it is only required to determine if a particular <parameter>needle</parameter>
```

Girgias (approved):
Thank you!

PR #3237, opened 2024-03-05
Add Network LOG_LOCAL IDs

haszi:


reference/network/constants.xml
```diff
@@ -84,8 +84,36 @@
       <entry>kernel messages</entry>
      </row>
      <row xml:id="constant.log-local0">
-      <entry><constant>LOG_LOCAL0 ... LOG_LOCAL7</constant></entry>
-      <entry>reserved for local use, these are not available in Windows</entry>
+      <entry><constant>LOG_LOCAL0</constant></entry>
```
Girgias, line 87:
```suggestion
      <entry><constant>LOG_LOCAL<replaceable>[0-7]</replaceable></constant></entry>
```

Maybe? As I'm not sure those constants are referenced anywhere

haszi: While I agree that it makes sense and looks much better, I think that there should be at least one place in the docs where each constant is listed individually and each of them has a unique XML ID that can be linked to. 

You're right though: most of these constants are not referenced except `LOG_LOCAL0` and `LOG_LOCAL7` the latter of which currently doesn't have an XML ID. :-) If we don't add the IDs though these constants can never be referenced. :-)

Girgias: Ack

PR #3247, opened 2024-03-07
Change some constant and literal tags to the appropriate tags

haszi:
Add the `<constant>` tags to PHP constant wrapped in a `<literal>` tag or no tag at all, and swap the `<constant>` tag with a `<literal>` tag where it used on non-PHP constants.

### DIFF
--- a/reference/math/functions/is-finite.xml
+++ b/reference/math/functions/is-finite.xml
@@ -41,7 +41,7 @@
   &reftitle.returnvalues;
   <para>
    &true; if <parameter>num</parameter> is none of <constant>NAN</constant>,
-   <constant>INF</constant>, <literal>-INF</literal>, else &false;.
+   <constant>INF</constant>, -<constant>INF</constant>, else &false;.
   </para>
  </refsect1>
 
--- a/reference/math/functions/is-finite.xml
+++ b/reference/math/functions/is-finite.xml
@@ -16,7 +16,7 @@
    Returns whether the given <parameter>num</parameter> is a finite float.
   </para>
   <para>
-   A finite float is neither <literal>NAN</literal> (<function>is_nan</function>),
+   A finite float is neither <constant>NAN</constant> (<function>is_nan</function>),
    nor infinite (<function>is_infinite</function>).
   </para>
  </refsect1>
@@ -40,8 +40,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if <parameter>num</parameter> is none of <literal>NAN</literal>,
-   <literal>INF</literal>, <literal>-INF</literal>, else &false;.
+   &true; if <parameter>num</parameter> is none of <constant>NAN</constant>,
+   <constant>INF</constant>, <literal>-INF</literal>, else &false;.
   </para>
  </refsect1>
 
--- a/reference/strings/functions/nl-langinfo.xml
+++ b/reference/strings/functions/nl-langinfo.xml
@@ -46,7 +46,7 @@
          </thead>
          <tbody>
           <row>
-           <entry namest="c1" nameend="c2" align="center"><emphasis>LC_TIME Category Constants</emphasis></entry>
+           <entry namest="c1" nameend="c2" align="center"><emphasis><constant>LC_TIME</constant> Category Constants</emphasis></entry>
           </row>
           <row>
            <entry><constant>ABDAY_(1-7)</constant></entry>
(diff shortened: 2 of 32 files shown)


reference/strings/functions/nl-langinfo.xml
```diff
@@ -245,7 +245,7 @@
            <entry>Output string for <literal>"no"</literal>.</entry>
           </row>
           <row>
-           <entry namest="c1" nameend="c2" align="center"><emphasis>LC_CTYPE Category Constants</emphasis></entry>
```
kamil-tekiela, line 248: Why remove `<emphasis>` here? This change is inconsistent with all other headings in this table.

haszi: You're right, this was a mistake.

reference/mysql/ini.xml
```diff
@@ -173,8 +173,8 @@
      from the <envar>MYSQL_TCP_PORT</envar> environment
      variable, the <literal>mysql-tcp</literal> entry in
      <filename>/etc/services</filename> or the compile-time
-     <constant>MYSQL_PORT</constant> constant, in that order. Win32
-     will only use the <constant>MYSQL_PORT</constant> constant.
+     <literal>MYSQL_PORT</literal> constant, in that order. Win32
```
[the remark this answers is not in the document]
haszi: I was assuming that `<constant>` tags are used for runtime PHP constants only. 

Btw, this same paragraph is used in `mysqli/ini.xml` where `MYSQL_PORT` is wrapped in a `<literal>` tag.

https://github.com/php/doc-en/blob/affa37e16f562d9297e83b2e21ec416aadc8b72d/reference/mysqli/ini.xml#L177-L181

reference/strings/functions/nl-langinfo.xml
haszi: I've added this back and also found a few more missing `<constant>` tags in the same file.

reference/mysql/ini.xml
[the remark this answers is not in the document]
kamil-tekiela: Ok, I guess you could do that for consistency, but otherwise I don't see a reason to change this.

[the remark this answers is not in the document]
haszi: Every one of these `<constant>`s that I've changed to `<literal>` were `<literal>`s in at least one other part of the docs, so in that sense all these changes are about consistency. 
On the other hand, I'm not sure my understanding of what supposed to be a `<constant>` in the docs is correct.

haszi:
[This](https://gist.github.com/haszi/96a2313c0cfbc191a004002b202deb94) is the script I've used to make these changes, in case it's needed for the translations.

haszi:
> IMHO `<constant>` should be used for PHP and C constants, however we may want to add an annotation attribute `lang=c` (or similar) to those constants to not have them auto-link?

The link should not be an issue as that is only created if there is a corresponding ID indexed which should never be added for a C constant anyway. 

_My two cents_ :-)
I think that the majority of C constants that are in the documentation (I've mostly seen these in database related extensions) use the `<literal>` tag in a way something like `'the <constant>PHP_CONSTANT</constant> constant corresponds to the <literal>SOME_EXTERNAL_NON_PHP_DATABASE_CONSTANT</literal> option/attribute/constant'`. To me, the `<literal>` tag makes it obvious that the latter is not part of PHP so I cannot use it in my PHP code and if I want to know what that external constant is really doing, I'll have to look it up in the vendor's documentation. But this is obviously a matter of preference so... :-)

If the `<constant>` tag is only used for PHP constants then these can be cross-referenced with `php-src` and `xml:id`s, ensuring all constants registered in `php-src` are listed with an `xml:id` in the documentation and each of these can be linked to.
If it is used for any of the extensions' C constants, there can be no easy consistency checks done on whether these C constants are always using the `<constant>` tag and we'd lose the possiblity if cross-referencing PHP constants.

haszi:
I removed all the controversial `<constant>` to `<literal>` tag changes. Could you have another look?

reference/math/functions/is-finite.xml
```diff
@@ -40,8 +40,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if <parameter>num</parameter> is none of <literal>NAN</literal>,
-   <literal>INF</literal>, <literal>-INF</literal>, else &false;.
+   &true; if <parameter>num</parameter> is none of <constant>NAN</constant>,
+   <constant>INF</constant>, <literal>-INF</literal>, else &false;.
```
kamil-tekiela, line 44: You kept the negative here as it was and you changed it here https://github.com/php/doc-en/pull/3247/files#diff-d670f877fd79de280bc80d60436a7cfd730cc0b92efc292c4a196e41efa9615bR19 Can you keep it consistent? Which way should it be?

haszi: You're right (again). This was an oversight on my part (again). I just pushed a commit to change all `<literal>-INF</literal>`s to `-<constant>INF</constant>`.

kamil-tekiela (approved):

PR #3248, opened 2024-03-07
Fix mysqli constant IDs

haszi:
Fix mysqli constant IDs so these can be linked to.

This update was done with using following bash script:

```bash
#!/bin/bash
\grep -lrE --include="*.xml" '"constantmysqli-' | xargs -d '\n' sed -i 's/"constantmysqli-/"constant.mysqli-/g'
\grep -lrE --include="*.xml" '"constant.mysqli-type-fload"' | xargs -d '\n' sed -i 's/"constant.mysqli-type-fload"/"constant.mysqli-type-float"/g'

```

### DIFF
--- a/reference/mysqli/constants.xml
+++ b/reference/mysqli/constants.xml
@@ -4,7 +4,7 @@
  &reftitle.constants;
  &extension.constants;
   <variablelist>
-   <varlistentry xml:id="constantmysqli-read-default-group">
+   <varlistentry xml:id="constant.mysqli-read-default-group">
     <term>
      <constant>MYSQLI_READ_DEFAULT_GROUP</constant>
      (<type>int</type>)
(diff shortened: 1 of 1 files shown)


kamil-tekiela (approved):
I don't have the context of why we are doing this churn, but the change itself looks ok. Please check with translators to see if it's ok to do it.

haszi:
Thanks for the review.

> I don't have the context of why we are doing this churn, but the change itself looks ok. 

Phd now generates links to constants as long as their IDs are in one of two standard formats: `constant.namespace-global-constant-name` or `namespace-and-class.constants.constant-name`. My PRs are updating IDs to be in one of these two formats or adding IDs for constants that don't have one yet.


> Please check with translators to see if it's ok to do it.

I think @Girgias will take care of that.

Girgias:
> I don't have the context of why we are doing this churn, but the change itself looks ok. Please check with translators to see if it's ok to do it.

It's for constant tags to link properly, and I apply the changes to translations via the provided script

PR #3257, opened 2024-03-09
Fix: add flag true to json_decode function 

RivGames:
Uncaught exception: Error Cannot use object of type stdClass as array


nielsdos (approved):
Thanks!

PR #3263, opened 2024-03-12
Remove duplicate XMLReader IDs

haszi:
Remove duplicate XMLReader IDs by removing the constant page that was not linked to the documentation and remove the corresponding `FIXME` note from XMLReader's `book.xml`. The reason the constants page can be removed without any issues is that all of XMLReader's constants are class constants and listed on the class's synopsis page.


Girgias (approved):
LGTM

PR #3266, opened 2024-03-12
database.xml: add the caption text to the image in the caption tag

mmalferov:


security/database.xml
```diff
@@ -343,8 +343,12 @@ $result = mssql_query($query);
       <imageobject>
        <imagedata fileref="en/security/figures/xkcd-bobby-tables.png" format="PNG"/>
       </imageobject>
+      <caption>
+       <para>
```
Girgias, line 347: Maybe using `simpara` is better?

Girgias (approved):

PR #3267, opened 2024-03-13
construct.xml: fiexed the example's syntax and amend the function arg

mmalferov:
In strict mode: 

> Fatal error: Uncaught TypeError: var_export(): Argument #2 ($return) must be of type bool, int given


Girgias (approved):
LGTM

PR #3269, opened 2024-03-14
Update getmxrr.xml: Change argument name: `weight` -> `preference`

ThomasLandauer:
Page: https://www.php.net/manual/en/function.getmxrr.php

I'm suggesting to rename the argument to "preferences", since this is the term that the RFC uses: https://datatracker.ietf.org/doc/html/rfc5321#section-5.1

> MX records contain a preference indication that MUST be used in
   sorting if more than one such record appears (see below).  Lower
   numbers are more preferred than higher ones.

Waiting for feedback before finishing...


Girgias:
This is the argument name that is used in the php-src stubs, and the manual document those so that one can use the functions with named arguments.

If you want to change this please submit a PR to php-src.

PR #3275, opened 2024-03-16
pattern.syntax.xml: extra spaces have been removed and...

mmalferov:
...some literal-tags have been added


Girgias:
Can you split the whitespace only changes with the literal tag addition into 2 PRs?

mmalferov:
I removed the `literal` tags and left only those changes that relate to extra spaces

PR #3278, opened 2024-03-19
basic.xml: an indication of specific ways to create an instance of a …

mmalferov:
…class

It's hard to look for exactly which different ways of instantiation are discussed in the documentation

P. S. I probably specified through/by and a/the incorrectly, correct if necessary, please

### DIFF
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -333,7 +333,7 @@ object(SimpleClass)#1 (1) {
 
 class Test
 {
-    static public function getNew()
+    public static function getNew()
     {
         return new static;
     }
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -342,7 +342,7 @@ class Test
 class Child extends Test {}
 
 $obj1 = new Test(); // By the class name
-$obj2 = new $obj1; // Through the variable
+$obj2 = new $obj1; // Through the variable containing an object
 var_dump($obj1 !== $obj2);
 
 $obj3 = Test::getNew(); // By the class method
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -335,14 +335,14 @@ class Test
 {
     public static function getNew()
     {
-        return new static;
+        return new static();
     }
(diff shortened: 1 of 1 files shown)


mmalferov:
> Note: Prior to PHP 7.1, the arguments are not evaluated if there is no constructor function defined.

What arguments are we talking about? If we are talking about class properties, maybe we should write this? Like that:

> Note: Prior to PHP 7.1, the **properties** are not evaluated if there is no constructor **method** defined.

mmalferov:
> Note: Prior to PHP 7.1, the arguments are not evaluated if there is no constructor function defined.

An incomprehensible sentence. This is probably about the behavior of the code, an example of which I will give below. But the proposal does not disclose the behavior of the code prior to PHP 7.1.

```
<?php

class Foo
{
	// Prior to PHP 7.1, uncomment to get 1 in `echo $number` expression
	// public function __construct() {}
}

$number = 0;
new Foo(++$number); // The argument are not evaluated prior to PHP 7.1
echo $number; // 1, or 0 prior to PHP 7.1 without __construct() method or 1 with it

?>
```

Girgias:
Are the comments really that useful? It might make more sense to split the example into 4 smaller different examples?

mmalferov:
Mastodons like you don't need comments, experience allows you to understand everything the first time ;-) And beginners have to thoughtfully analyze every, line, of, code. The comment is at least visually separated by different ways of creating an instance of a class. I was asked: Where exactly in the example of 'a couple of ways'? And when I showed an example with comments, the student immediately understood what to pay attention to. I don't know why beginners have such difficulties. For example, they can reason like this: maybe `getNew()` is a built—in method of each PHP class and it is this method that creates an instance of the `Test` class? Stupid? Maybe. But if there were comments, there were no more questions. And it is unknown how quirky novice programmers can still reason. I'm sure the comments won't hurt.

These comments are just to say: here, here, here and here.

I wouldn't make four small examples out of one example, if only because then you won't have to describe the `Test` and `Child` classes four times :-)

PR #3287, opened 2024-03-22
Corrections to comments in PDO construct examples

leonardolara:
Correction to comments in the examples.

### DIFF
--- a/reference/pdo/pdo/construct.xml
+++ b/reference/pdo/pdo/construct.xml
@@ -157,7 +157,7 @@ odbc:DSN=SAMPLE;UID=john;PWD=mypass
     <programlisting role="php">
 <![CDATA[
 <?php
-/* Connect to an ODBC database using driver invocation */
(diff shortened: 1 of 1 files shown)


reference/pdo/pdo/construct.xml
```diff
@@ -157,7 +157,7 @@ odbc:DSN=SAMPLE;UID=john;PWD=mypass
     <programlisting role="php">
 <![CDATA[
 <?php
-/* Connect to an ODBC database using driver invocation */
+/* Connect to an ODBC database using URI invocation */
```
kamil-tekiela, line 160: Good find. 

ODBC is not a database, is it? What even is the purpose of these code comments? The example is properly described already. I'd remove all 3 comments.

leonardolara: I agree. These comments are redundant.

kamil-tekiela (approved):

PR #3288, opened 2024-03-24
variables.xml: amend examples code syntax

mmalferov:


kamil-tekiela:
I am going to close this PR but if you are still willing to pursue these changes, please create a new one. Though, I am not sure it's worth the effort right now. These are rather small changes and there areas that still need a much larger rework.

PR #3289, opened 2024-03-25
Copy constants from curl_getinfo() to constants page

haszi:
Copy constants from `curl_getinfo()` to constants page, merge descriptions of constants on both pages, remove existing constants from constants page and add all constants to their own constant list which will be rendered as a table.

Used the following script to make the first commit:
https://gist.github.com/haszi/fed1dca8fb3d88cd09d5b5ef64f4506a

After running the script, the whitespace and the missing period between some of the descriptions' sentences has been fixed manually. If this script is used to update translations, the same punctuation check should be done.


reference/curl/constants.xml
```diff
@@ -2227,774 +2182,774 @@
...
+  <varlistentry xml:id="constant.curl-version-altsvc">
    <term>
-    <constant>CURLINFO_CONNECT_TIME</constant>
+    <constant>CURL_VERSION_ALTSVC</constant>
     (<type>int</type>)
    </term>
    <listitem>
     <simpara>
-     
+     Available since PHP 7.3.6 and cURL 7.64.1
```
Girgias, line 2302: `since` should be `as of` if we are going to be mass fixing this.

haszi:
I have created a new file just for the `curl_getinfo()` constants and included it in the main cURL constants file.

I'll also update the script to do the same so it can be easier to apply this change to the translations and to make it easier to review the changes.

Girgias:
As we are creating a new file I don't think I need the script, but can you try and reorder the commits to be *first* the move of the relevant constants into the new file, and then have the `since` -> `as of` conversion commit happen?

haszi:
I just pushed the requested changes and also added some missing `<constant>` and `<replaceable>` tags in a separate commit.

reference/curl/constants_curl_getinfo.xml
```diff
@@ -0,0 +1,755 @@
+ <variablelist xml:id="constant.curl-getinfo.constants" role="constant_list">
```
Girgias, line 1: Does this need a:
```xml
```
Before the first tag?

Girgias (approved):
LGTM just one minor question

reference/curl/constants_curl_getinfo.xml
haszi: I really don't know but I just added it and it validates and renders without any issues so let's add it. :-)

Girgias:
Can you merge the last commit with the first one in a rebase, and force push so that I can merge the PR with a rebase?

haszi:
> Can you merge the last commit with the first one in a rebase, and force push so that I can merge the PR with a rebase?

Done. 

I've also updated the script I've made the first commit with (just in case anyone wants to use it to do the same in the translations).

reference/curl/constants_curl_getinfo.xml
```diff
@@ -0,0 +1,756 @@
+<?xml version="1.0" encoding="utf-8"?>
+ <variablelist xml:id="constant.curl-getinfo.constants" role="constant_list">
```
Girgias, line 2: Why is there an indentation now?

haszi: It was there last time too. :-) I'll remove it.

haszi: I've also found a duplicate sentence so I removed that in a new commit.

Girgias:
Thank you!

PR #3290, opened 2024-03-26
fix: document BC on count function in PHP 8.0

mlavrinenko:
[count](https://www.php.net/manual/en/function.count) provides information about its BC, but [migration documentation](https://www.php.net/manual/en/migration80.incompatible.php) does not.


Girgias (approved):

Girgias:
Thank you!

PR #3292, opened 2024-03-27
Clarify usage of Iterator::valid methods

andrewnicols:
This commit updates the documentation for the `valid` method of the following iterators:

- `EmptyIterator` - add a note that this will always return false
- `FilterIterator` - add a note that the value will be false until the pointer is advanced to the first valid element
- `IteratorIterator` - clarify that the method checks if the element is valid.

These now match the source.

I have checked other iterators and these seem to be correctly documented.

It's worth noting that the `ArrayIterator::valid` method behaves differently - it checks if the next element is valid (confirmed in source).

The `DirectoryIterator::valid` method may also be incorrectly documented but I am not 100% sure of this. The documentation states that the method checks if it is a valid _file_, but I believe that the method checks if the Iterator element is valid. However, given the DirectoryIterator should only Iterator over files in the first place, this is arguably correct.


nielsdos (approved):
Thanks a lot!

PR #3296, opened 2024-03-28
Add U_ error constants to intl constant list

haszi:
Add missing `U_` error constants to `intl` constant list. All of these are listed on the missing constant list (#2747).

_Edit:_
I used the following script which can also be used to update all translations:
https://gist.github.com/haszi/66f36bc412323e0365b8614aeabaa816
Please note that the scripts unconditionally adds the constants to the constants page, ie. without checking whether these are already present so running it more than once is not advised.

### DIFF
--- a/reference/intl/constants.xml
+++ b/reference/intl/constants.xml
@@ -306,6 +306,1560 @@
      </simpara>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="constant.u-ambiguous-alias-warning">
+    <term>
+     <constant>U_AMBIGUOUS_ALIAS_WARNING</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      This converter alias can go to different converter implementations.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-bad-variable-definition">
+    <term>
+     <constant>U_BAD_VARIABLE_DEFINITION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Missing <literal>'$'</literal> or duplicate variable name.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-assign-error">
+    <term>
+     <constant>U_BRK_ASSIGN_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Syntax error in RBBI rule assignment statement.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-error-limit">
+    <term>
+     <constant>U_BRK_ERROR_LIMIT</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      This must always be the last value to indicate the limit for Break Iterator failures.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-error-start">
+    <term>
+     <constant>U_BRK_ERROR_START</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Start of codes indicating Break Iterator failures.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-hex-digits-expected">
+    <term>
+     <constant>U_BRK_HEX_DIGITS_EXPECTED</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Hex digits expected as part of a escaped char in a rule.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-init-error">
+    <term>
+     <constant>U_BRK_INIT_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Initialization failure. Probable missing ICU Data.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-internal-error">
+    <term>
+     <constant>U_BRK_INTERNAL_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      An internal error (bug) was detected.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-malformed-rule-tag">
+    <term>
+     <constant>U_BRK_MALFORMED_RULE_TAG</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      The <literal>{nnn}</literal> tag on a rule is mal formed.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-mismatched-paren">
+    <term>
+     <constant>U_BRK_MISMATCHED_PAREN</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Mis-matched parentheses in an RBBI rule.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-new-line-in-quoted-string">
+    <term>
+     <constant>U_BRK_NEW_LINE_IN_QUOTED_STRING</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Missing closing quote in an RBBI rule.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-rule-empty-set">
+    <term>
+     <constant>U_BRK_RULE_EMPTY_SET</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Rule contains an empty Unicode Set.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-rule-syntax">
+    <term>
+     <constant>U_BRK_RULE_SYNTAX</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Syntax error in RBBI rule.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-semicolon-expected">
+    <term>
+     <constant>U_BRK_SEMICOLON_EXPECTED</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Missing <literal>';'</literal> at the end of a RBBI rule.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-unclosed-set">
+    <term>
+     <constant>U_BRK_UNCLOSED_SET</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      UnicodeSet witing an RBBI rule missing a closing <literal>']'</literal>.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-undefined-variable">
+    <term>
+     <constant>U_BRK_UNDEFINED_VARIABLE</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Use of an undefined <code>$Variable</code> in an RBBI rule.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-unrecognized-option">
+    <term>
+     <constant>U_BRK_UNRECOGNIZED_OPTION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Option in RBBI rules not recognized.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-variable-redfinition">
+    <term>
+     <constant>U_BRK_VARIABLE_REDFINITION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      RBBI rule variable redefined.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-buffer-overflow-error">
+    <term>
+     <constant>U_BUFFER_OVERFLOW_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      A result would not fit in the supplied buffer.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-ce-not-found-error">
+    <term>
+     <constant>U_CE_NOT_FOUND_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Currently used only while setting variable top, but can be used generally.
+     </simpara>
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-collator-version-mismatch">
+    <term>
+     <constant>U_COLLATOR_VERSION_MISMATCH</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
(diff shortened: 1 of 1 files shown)


reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-bad-variable-definition">
+    <term>
+     <constant>U_BAD_VARIABLE_DEFINITION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Missing '$' or duplicate variable name.
```
Girgias, line 294: Adding a literal tag here would be good I think

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    <term>
+     <constant>U_BRK_ERROR_START</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Start of codes indicating Break Iterator failures.
+     </simpara>
+    </listitem>
+   </varlistentry>
```
Girgias, line 330: I'm not sure I totally understand this?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+   <varlistentry xml:id="constant.u-brk-malformed-rule-tag">
+    <term>
+     <constant>U_BRK_MALFORMED_RULE_TAG</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      The {nnn} tag on a rule is mal formed.
```
Girgias, line 371: Literal tag

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-semicolon-expected">
+    <term>
+     <constant>U_BRK_SEMICOLON_EXPECTED</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Missing ';' at the end of a RBBI rule.
```
Girgias, line 426: Literal tag

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-undefined-variable">
+    <term>
+     <constant>U_BRK_UNDEFINED_VARIABLE</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Use of an undefined $Variable in an RBBI rule.
```
Girgias, line 448: Use a variable tag?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-brk-variable-redfinition">
+    <term>
+     <constant>U_BRK_VARIABLE_REDFINITION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      RBBI rule variable redefined./.
```
Girgias, line 470: Seem like something unneeded got added?
```suggestion
      RBBI rule variable redefined.
```

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-different-uca-version">
+    <term>
+     <constant>U_DIFFERENT_UCA_VERSION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      ucol_open encountered a mismatch between UCA version and collator image version, so the collator was constructed from rules. No impact to further function.
```
Girgias, line 514: Might need a better description?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-enum-out-of-sync-error">
+    <term>
+     <constant>U_ENUM_OUT_OF_SYNC_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      UEnumeration out of sync with underlying collection.
```
Girgias, line 525: Some tag for `UEnumeration` maybe?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-error-limit">
+    <term>
+     <constant>U_ERROR_LIMIT</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Alias of U_PLUGIN_ERROR_LIMIT.
```
Girgias, line 536: Constant tag missing

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    <term>
+     <constant>U_ERROR_WARNING_START</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Start of information results (semantically successful).
+     </simpara>
+    </listitem>
+   </varlistentry>
```
Girgias, line 561: Are these actually exposed?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+   <varlistentry xml:id="constant.u-illegal-argument-error">
+    <term>
+     <constant>U_ILLEGAL_ARGUMENT_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Start of codes indicating failure.
```
Girgias, line 723: Description looks wrong?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-invalid-function">
+    <term>
+     <constant>U_INVALID_FUNCTION</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      A '&amp;fn()' rule specifies an unknown transliterator.
```
Girgias, line 844: Literal tag?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-invalid-id">
+    <term>
+     <constant>U_INVALID_ID</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      A '::id' rule specifies an unknown transliterator.
```
Girgias, line 855: Literal tag?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1557 @@
...
+    <term>
+     <constant>U_INVARIANT_CONVERSION_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Unable to convert a UChar* string to char* with the invariant converter.
+     </simpara>
+    </listitem>
+   </varlistentry>
```
Girgias, line 924: Is this actually exposed?

Girgias (commented):
Partial review

reference/intl/constants.xml
haszi: So a short background to all these constants:

these constants are defined in a `C` `enum` in the `ICU` library's [`utypes.h`](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/utypes.h) header file (starting at line 415). I've copied all the descriptions from there and all the constants are directly exposed by `PHP`. A consequence of the latter is (IMO) that some of these constants do not make a lot of sense in `PHP`.

As for this constant, the `ICU`'s `C` `enum` has a lot of error codes and most of these only apply to a certain function. This constant (and a few more like this) actually has its values set in the `enum` which shows where a new set of error codes begins and since these values end in `0x00`, they can be used to detect certain types of error codes. At least that's what I think. :-) 
So this constant shows where the `BreakIterator` type error codes start, `U_REGEX_ERROR_START` shows where `regex` type error codes start, etc., and when used as flags these can all be used to check whether an error is of a certain 'type'.

reference/intl/constants.xml
haszi: Yes, they are. As I wrote above, all error codes of the underlying `C` `enum` are exposed.

reference/intl/constants.xml
haszi: Yep.

reference/intl/constants.xml
haszi: I just checked and it is the first error code in the `enum` set to the value of `1`, hence its description I guess. :-)

How about: 'Indicates an incorrect argument value.'?

reference/intl/constants.xml
haszi: This is a `C` `enum`. Maybe a `<code>` tag?

reference/intl/constants.xml
haszi: Not sure how to go about that, to be honest. The current description very accurately describes what is happening but it does sound very confusing when you are using `intl`/`ICU` from `PHP`...

reference/intl/constants.xml
haszi: The DTD doesn't allow for `<variable>`s inside `<simpara>`s. Would a `<code>` tag work?

reference/intl/constants.xml
haszi: > It might make sense to commit this script to the doc-base repo and add a comment at the top to explain how to generate the constant page?

As long as it's not being ran on `doc-en`, that would work. :-)

reference/intl/constants.xml
Girgias: Arg right, maybe code is good yes

reference/intl/constants.xml
Girgias: I think that is a better description :)

reference/intl/constants.xml
```diff
@@ -1259,7 +1260,7 @@
     </term>
     <listitem>
      <simpara>
-      Equivalent to Java ParseException.
+      Equivalent to Java <code>ParseException</code>.
```
Girgias, line 1263: Maybe reference PHP's ParseError instead?

reference/intl/constants.xml
Girgias: Huuum what's the issue with running it on doc-en? I suppose the idea would be to try and generate the whole page from the script with having a comment at the top indicating that this page is generated?

reference/intl/constants.xml
Girgias: Yah... I think a major paragraph at the start of the docs indicating that all ICU C++ constants are exposed and some don't make sense in PHP is required.

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-invalid-rbt-syntax">
+    <term>
+     <constant>U_INVALID_RBT_SYNTAX</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      A '::id' rule was passed to the RuleBasedTransliterator parser.
```
Girgias, line 877: Literal tag

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-malformed-pragma">
+    <term>
+     <constant>U_MALFORMED_PRAGMA</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      A 'use' pragma is invalid.
```
Girgias, line 944: Literal tag

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-regex-bad-interval">
+    <term>
+     <constant>U_REGEX_BAD_INTERVAL</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Error in {min,max} interval.
```
Girgias, line 1329: Literal tag maybe?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-regex-max-lt-min">
+    <term>
+     <constant>U_REGEX_MAX_LT_MIN</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      In {min,max}, max is less than min.
```
Girgias, line 1417: Literal tag if done for the previous case

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-resource-type-mismatch">
+    <term>
+     <constant>U_RESOURCE_TYPE_MISMATCH</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      an operation is requested over a resource that does not support it.
```
Girgias, line 1494:
```suggestion
      An operation is requested over a resource that does not support it.
```

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-safeclone-allocated-warning">
+    <term>
+     <constant>U_SAFECLONE_ALLOCATED_WARNING</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      A SafeClone operation required allocating memory (informational only).
```
Girgias, line 1516: Code tag?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-state-old-warning">
+    <term>
+     <constant>U_STATE_OLD_WARNING</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      ICU has to use compatibility layer to construct the service. Expect performance.
```
Girgias, line 1550: Performance degradation?

reference/intl/constants.xml
```diff
@@ -273,6 +273,1559 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-string-not-terminated-warning">
+    <term>
+     <constant>U_STRING_NOT_TERMINATED_WARNING</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      An output string could not be NUL-terminated because output length==destCapacity.
```
Girgias, line 1572: Code tags?

reference/intl/constants.xml
Girgias: Yeah that would work

reference/intl/constants.xml
haszi: If this PR is merged, the constants are already on the constants page and since the script unconditionally adds the constants to the constants page, there'd be a bunch duplicates. :-)

reference/intl/constants.xml
haszi: I agree, it's just hard to find a good place to add that to as `intl` has a bunch of class and global constants on probably 10+ different pages.

reference/intl/constants.xml
haszi: I just read up on Java's ParseException and it seems that it's used more like for methods parsing their input, e.g. date/times and throw on incorrect format or value. It doesn't seem similar enough to PHP's ParseError to me, but what do you think?

Girgias: Right, leave it then :)

Girgias:
Could you rebase the PR?

haszi:
> Could you rebase the PR?

Done.

reference/intl/constants.xml
```diff
@@ -306,6 +306,1560 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry xml:id="constant.u-too-many-aliases-error">
+    <term>
+     <constant>U_TOO_MANY_ALIASES_ERROR</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      There are too many aliases in the path to the requested resource. It is very possible that a circular alias definition has occured.
```
Girgias, line 1650:
```suggestion
      There are too many aliases in the path to the requested resource.
      It is very possible that a circular alias definition has occured.
```

Girgias (approved):
Minor nit, but LGTM considering what ext/intl does...

PR #3297, opened 2024-03-28
Fix DOMImplementation::hasfeature() example

KidFlo:
Fix broken code sample in DOMImplementation::hasfeature().

### DIFF
--- a/reference/dom/domimplementation/hasfeature.xml
+++ b/reference/dom/domimplementation/hasfeature.xml
@@ -105,9 +105,11 @@ $features = array(
   'Range'          => 'Range module',
   'Traversal'      => 'Traversal module'
 );
-               
+
+$implementation = new DOMImplementation;
+
 foreach ($features as $key => $name) {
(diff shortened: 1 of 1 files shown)


reference/dom/domimplementation/hasfeature.xml
```diff
@@ -105,9 +105,11 @@ $features = array(
   'Range'          => 'Range module',
   'Traversal'      => 'Traversal module'
 );
-               
+
+$implementation = new DOMImplementation;
+
 foreach ($features as $key => $name) {
-  if (DOMImplementation::hasFeature($key, '2.0')) {
+  if ($implementation::hasFeature($key, '2.0')) {
```
nielsdos, line 112: This should fix it completely, because you can't call `hasFeature` statically anymore:

```suggestion
  if ($implementation->hasFeature($key, '2.0')) {
```

nielsdos (commented):
Thanks, one minor detail that I'll fix in the merge.

PR #3299, opened 2024-03-29
Copy constants from nl_langinfo() to constants page

haszi:
Copy constants from `nl_langinfo()` to constants page.

The `ABDAY`, `DAY`, `ABMON` and `MON` constants need to be listed individually (as opposed to the current `(1-7)` and `(1-12)`) format) so they can be automatically linked to which will be done in a follow up PR.

I used the following script for this update which can also be used to update all translations having the necessary files in one go:
https://gist.github.com/haszi/be927f351ab7efe23f46a11cc92c2000
If used, please note that the script should only be ran once as it will just blindly add the new constant tables to the constants page.

### DIFF
--- a/reference/strings/constants.xml
+++ b/reference/strings/constants.xml
@@ -335,6 +335,498 @@
    </listitem>
   </varlistentry>
  </variablelist>
+ <variablelist role="constant_list">
+  <title><function>nl_langinfo</function> <constant>LC_TIME</constant> Category Constants</title>
+  <varlistentry>
+   <term>
+    <constant>ABDAY_(1-7)</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Abbreviated name of n-th day of the week.
+    </simpara>
+   </listitem>
+  </varlistentry>
+  <varlistentry>
+   <term>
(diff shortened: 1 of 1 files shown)


reference/strings/constants.xml
```diff
@@ -335,6 +335,498 @@
    </listitem>
   </varlistentry>
  </variablelist>
+ <variablelist role="constant_list">
+  <title><function>nl_langinfo</function> <constant>LC_TIME</constant> Category Constants</title>
+  <varlistentry>
+   <term>
+    <constant>ABDAY_(1-7)</constant>
```
Girgias, line 342: Should this use the `<replaceable>` tag or not?

haszi: Yes but I'd still like to leave this PR as it is due to following:

1. the script can be used on the translations to make the same exact change
2.  after this PR is merged, I will be opening another PR in which I'll delete these replaceable (non-) constants and will list each of the (actual) constants in their own `<varlistentry>`

So this PR in its current form wouldn't be too much work on the translations' side but the next one will have to be done manually anyway.

Girgias: I'll merge the PR in the current form and we can do the rest in a follow-up PR

PR #3301, opened 2024-03-31
get-class-vars.xml: amend the code of examples

mmalferov:


reference/classobj/functions/get-class-vars.xml
```diff
@@ -49,29 +49,28 @@
...
+{
+    public $var1; // This has no default value...
+    public $var2 = "xyz";
+    public $var3 = 100;
     private $var4;
 
-    // constructor
-    function __construct() {
-        // change some properties
+    function __construct()
```
TimWolla, line 59: Might as well:
```suggestion
    public function __construct()
```
To add the visibility everywhere.

reference/classobj/functions/get-class-vars.xml
```diff
@@ -49,29 +49,28 @@
...
 <?php
 
-class myclass {
-
-    var $var1; // this has no default value...
-    var $var2 = "xyz";
-    var $var3 = 100;
+class myclass
+{
+    public $var1; // This has no default value...
```
Girgias, line 54: Technically this has `null` as a default value.

kamil-tekiela: Yes, and please change the output too to match the changes. It should show NULL in the output.

PR #3303, opened 2024-04-02
migration83/incompatible.xml: remove the extra parentheses

mmalferov:


devnexen (approved):
true, it is not needed.

TimWolla:
Thanks.

PR #3306, opened 2024-04-07
Fix mysqli fetch field

schlndh:
This PR adds missing properties and unifies property descriptions for `mysqli_fetch_fields`, `mysqli_fetch_field` and `mysqli_fetch_field_direct`.

### DIFF
--- a/reference/mysqli/mysqli_result/fetch-field-direct.xml
+++ b/reference/mysqli/mysqli_result/fetch-field-direct.xml
@@ -83,6 +83,14 @@
       <entry>The default value for this field, represented as a string</entry>
      </row>
      <row>
+      <entry>db</entry>
+      <entry>The name of the database</entry>
+     </row>
+     <row>
+      <entry>catalog</entry>
+      <entry>The catalog name, always "def"</entry>
+     </row>
+     <row>
       <entry>max_length</entry>
       <entry>The maximum width of the field for the result set. As of PHP 8.1, this value is always <literal>0</literal>.</entry>
      </row>
--- a/reference/mysqli/mysqli_result/fetch-field-direct.xml
+++ b/reference/mysqli/mysqli_result/fetch-field-direct.xml
@@ -53,11 +53,11 @@
   </para>
   <para>
   <table xml:id="mysqli-result.fetch-field-direct.returnvalues">
-   <title>Object attributes</title>
+   <title>Object properties</title>
    <tgroup cols="2">
     <thead>
      <row>
-      <entry>Attribute</entry>
+      <entry>Property</entry>
       <entry>Description</entry>
      </row>
     </thead>
@@ -80,7 +80,7 @@
      </row>
      <row>
       <entry>def</entry>
-      <entry>The default value for this field, represented as a string</entry>
+      <entry>Reserved for default value, currently always ""</entry>
      </row>
      <row>
       <entry>db</entry>
@@ -96,11 +96,16 @@
      </row>
      <row>
       <entry>length</entry>
-      <entry>The width of the field, as specified in the table definition.</entry>
+      <entry>
+       The width of the field, in bytes, as specified in the table definition. Note that
+       this number (bytes) might differ from your table definition value (characters), depending on
+       the character set you use. For example, the character set utf8 has 3 bytes per character,
+       so varchar(10) will return a length of 30 for utf8 (10*3), but return 10 for latin1 (10*1).
+      </entry>
      </row>
      <row>
       <entry>charsetnr</entry>
-      <entry>The character set number for the field.</entry>
+      <entry>The character set number (id) for the field.</entry>
      </row>
      <row>
       <entry>flags</entry>
@@ -112,7 +117,7 @@
      </row>
      <row>
       <entry>decimals</entry>
-      <entry>The number of decimals used (for numeric fields)</entry>
+      <entry>The number of decimals for numeric fields, and the fractional seconds precision for temporal fields.</entry>
      </row>
     </tbody>
    </tgroup>
--- a/reference/mysqli/mysqli_result/fetch-field.xml
+++ b/reference/mysqli/mysqli_result/fetch-field.xml
@@ -82,15 +82,20 @@
      </row>
      <row>
       <entry>max_length</entry>
-      <entry>The maximum width of the field for the result set.</entry>
+      <entry>The maximum width of the field for the result set. As of PHP 8.1, this value is always <literal>0</literal>.</entry>
      </row>
      <row>
       <entry>length</entry>
-      <entry>The width of the field, as specified in the table definition.</entry>
+      <entry>
+       The width of the field, in bytes, as specified in the table definition. Note that
+       this number (bytes) might differ from your table definition value (characters), depending on
+       the character set you use. For example, the character set utf8 has 3 bytes per character,
+       so varchar(10) will return a length of 30 for utf8 (10*3), but return 10 for latin1 (10*1).
+      </entry>
      </row>
      <row>
       <entry>charsetnr</entry>
-      <entry>The character set number for the field.</entry>
+      <entry>The character set number (id) for the field.</entry>
      </row>
      <row>
       <entry>flags</entry>
@@ -102,7 +107,7 @@
      </row>
      <row>
       <entry>decimals</entry>
-      <entry>The number of decimals used (for integer fields)</entry>
+      <entry>The number of decimals for numeric fields, and the fractional seconds precision for temporal fields.</entry>
      </row>
     </tbody>
    </tgroup>
--- a/reference/mysqli/mysqli_result/fetch-fields.xml
+++ b/reference/mysqli/mysqli_result/fetch-fields.xml
@@ -70,6 +70,18 @@
       <entry>Original table name if an alias was specified</entry>
      </row>
      <row>
+      <entry>def</entry>
+      <entry>Reserved for default value, currently always ""</entry>
+     </row>
+     <row>
+      <entry>db</entry>
+      <entry>The name of the database</entry>
+     </row>
+     <row>
+      <entry>catalog</entry>
+      <entry>The catalog name, always "def"</entry>
+     </row>
+     <row>
       <entry>max_length</entry>
       <entry>The maximum width of the field for the result set. As of PHP 8.1, this value is always <literal>0</literal>.</entry>
      </row>
--- a/reference/mysqli/mysqli_result/fetch-fields.xml
+++ b/reference/mysqli/mysqli_result/fetch-fields.xml
@@ -108,7 +108,7 @@
      </row>
      <row>
       <entry>decimals</entry>
-      <entry>The number of decimals used (for integer fields)</entry>
+      <entry>The number of decimals for numeric fields, and the fractional seconds precision for temporal fields.</entry>
      </row>
     </tbody>
    </tgroup>


reference/mysqli/mysqli_result/fetch-field.xml
```diff
@@ -102,7 +107,7 @@
      </row>
      <row>
       <entry>decimals</entry>
-      <entry>The number of decimals used (for integer fields)</entry>
+      <entry>The number of decimals for numeric fields, and the fractional seconds precision for temporal fields.</entry>
```
schlndh, line 110: Taken from https://dev.mysql.com/doc/c-api/8.0/en/c-api-data-structures.html

reference/mysqli/mysqli_result/fetch-field-direct.xml
```diff
@@ -80,19 +80,32 @@
      </row>
      <row>
       <entry>def</entry>
-      <entry>The default value for this field, represented as a string</entry>
+      <entry>Reserved for default value, currently always ""</entry>
```
schlndh, line 83: See https://dev.mysql.com/doc/c-api/8.0/en/c-api-data-structures.html

> The default value of this field, as a null-terminated string. This is set only if you use mysql_list_fields().

And in my own testing the value has always been `""`.

reference/mysqli/mysqli_result/fetch-fields.xml
```diff
@@ -69,6 +69,18 @@
...
+     <row>
+      <entry>def</entry>
+      <entry>Reserved for default value, currently always ""</entry>
+     </row>
+     <row>
+      <entry>db</entry>
+      <entry>The name of the database</entry>
+     </row>
+     <row>
+      <entry>catalog</entry>
```
kamil-tekiela, line 81: Thanks for the PR. I have no idea what to do with this. `catalog` should not be documented. But as I can see it's there in 1/3 of these pages. Similarly with `def`. I think we just need to remove them either in 8.4 or 9.0. Maybe she should add them to the documentation just to make it complete.

reference/mysqli/mysqli_result/fetch-field.xml
```diff
@@ -82,15 +82,20 @@
...
+       The width of the field, in bytes, as specified in the table definition. Note that
+       this number (bytes) might differ from your table definition value (characters), depending on
+       the character set you use. For example, the character set utf8 has 3 bytes per character,
+       so varchar(10) will return a length of 30 for utf8 (10*3), but return 10 for latin1 (10*1).
+      </entry>
      </row>
      <row>
       <entry>charsetnr</entry>
-      <entry>The character set number for the field.</entry>
+      <entry>The character set number (id) for the field.</entry>
```
kamil-tekiela, line 98: Why do you add (id) here? It should be `ID` since it's an acronym for "Identification Document". But I would not add the ID here at all. I think it's unnecessary.

reference/mysqli/mysqli_result/fetch-field.xml
```diff
@@ -82,15 +82,20 @@
...
+      <entry>The maximum width of the field for the result set. As of PHP 8.1, this value is always <literal>0</literal>.</entry>
      </row>
      <row>
       <entry>length</entry>
-      <entry>The width of the field, as specified in the table definition.</entry>
+      <entry>
+       The width of the field, in bytes, as specified in the table definition. Note that
+       this number (bytes) might differ from your table definition value (characters), depending on
+       the character set you use. For example, the character set utf8 has 3 bytes per character,
+       so varchar(10) will return a length of 30 for utf8 (10*3), but return 10 for latin1 (10*1).
```
kamil-tekiela, line 93: Good but this is still not entirely accurate. This is true for string fields, but not for other fields. Also, I believe it's the connection character set, not the field character set, which could be misleading (I haven't verified it though). I would not be using the deprecated `utf8` charset as an example. Use `utf8mb4`.

reference/mysqli/mysqli_result/fetch-field-direct.xml
kamil-tekiela: If I am not mistaken this field is unused in PHP, so saying "currently" could be misleading. We have no plans on using it and I wonder why it was even added in the first place.

schlndh: I just copied the wording from one of the pages to the others, as I find this wording more accurate then the previous one.

Would you prefer something like "An empty string"?

reference/mysqli/mysqli_result/fetch-fields.xml
schlndh: I'm not sure I understand. Yes, the field is useless. But as long as it is returned as part of the object, it should be documented (at least to say that it's there and it's useless).

IMO if it gets removed in future PHP versions, it doesn't help from the perspective of the documentation, because the documentation should cover the old versions as well (i.e. the field should still be in the documentation and it should say that it was removed in version X).

reference/mysqli/mysqli_result/fetch-field.xml
schlndh: I just copied the description from one of the pages to the others to unify them. Either one is fine with me, I just wanted them to be the same on all 3 pages.

reference/mysqli/mysqli_result/fetch-field-direct.xml
kamil-tekiela: Huh, interesting. What about "Unused. Always an empty string"? And then keep it the same for "catalog" until it gets removed?

reference/mysqli/mysqli_result/fetch-field.xml
schlndh: It's a copy from one of the other pages for the purposes of unification.

reference/mysqli/mysqli_result/fetch-field.xml
kamil-tekiela: Right, I see. Can you do it the opposite way please?

reference/mysqli/mysqli_result/fetch-field.xml
schlndh: https://dev.mysql.com/doc/c-api/8.0/en/c-api-data-structures.html

>  The width of the field. This corresponds to the display length, in bytes.

> The server determines the length value before it generates the result set, so this is the minimum length required for a data type capable of holding the largest possible value from the result column, without knowing in advance the actual values that will be produced by the query for the result set.

> For string columns, the length value varies on the connection character set. For example, if the character set is latin1, a single-byte character set, the length value for a SELECT 'abc' query is 3. If the character set is utf8mb4, a multibyte character set in which characters take up to 4 bytes, the length value is 12.

reference/mysqli/mysqli_result/fetch-field.xml
```diff
@@ -70,23 +70,28 @@
...
```
kamil-tekiela, line 93:
```suggestion
       The width of the field, in bytes. For string columns,
       the length value varies on the connection character set. For example,
       if the character set is <literal>latin1</literal>, a single-byte character set,
       the length value for a <literal>SELECT 'abc'</literal> query is 3.
       If the character set is <literal>utf8mb4</literal>, a multibyte character
       set in which characters take up to 4 bytes, the length value is 12.
```

Copied from MySQL partially. It's better but not the best.

reference/mysqli/mysqli_result/fetch-field-direct.xml
```diff
@@ -80,15 +80,30 @@
...
      </row>
      <row>
       <entry>max_length</entry>
       <entry>The maximum width of the field for the result set. As of PHP 8.1, this value is always <literal>0</literal>.</entry>
      </row>
      <row>
       <entry>length</entry>
-      <entry>The width of the field, as specified in the table definition.</entry>
+      <entry>
+       The width of the field, in bytes. For string columns,
```
kamil-tekiela, line 100: Sorry for this but I just noticed I left an unnecessary comma. Could you please remove it?

reference/mysqli/mysqli_result/fetch-field-direct.xml
```diff
@@ -80,15 +80,30 @@
...
-      <entry>The default value for this field, represented as a string</entry>
+      <entry>Unused. Always an empty string</entry>
+     </row>
+     <row>
+      <entry>db</entry>
+      <entry>The name of the database</entry>
+     </row>
+     <row>
+      <entry>catalog</entry>
+      <entry>Unused. Always "def"</entry>
```
Girgias, line 91:
```suggestion
      <entry>Unused. Always <literal>"def"</literal></entry>
```

reference/mysqli/mysqli_result/fetch-field.xml
```diff
@@ -70,23 +70,30 @@
...
+      <entry>Unused. Always an empty string</entry>
      </row>
      <row>
       <entry>db</entry>
       <entry>The name of the database</entry>
      </row>
      <row>
       <entry>catalog</entry>
-      <entry>The catalog name, always "def"</entry>
+      <entry>Unused. Always "def"</entry>
```
Girgias, line 81: Ditto

reference/mysqli/mysqli_result/fetch-fields.xml
```diff
@@ -69,22 +69,36 @@
...
+      <entry>def</entry>
+      <entry>Unused. Always an empty string</entry>
+     </row>
+     <row>
+      <entry>db</entry>
+      <entry>The name of the database</entry>
+     </row>
+     <row>
+      <entry>catalog</entry>
+      <entry>Unused. Always "def"</entry>
```
Girgias, line 82: Ditto

Girgias (commented):
It seems like this table is copy pasted between 3 pages, is it possible to XInclude it instead?

kamil-tekiela (approved):

reference/mysqli/mysqli_result/fetch-field.xml
```diff
@@ -42,71 +42,9 @@
...
-      <entry>decimals</entry>
-      <entry>The number of decimals used (for integer fields)</entry>
-     </row>
-    </tbody>
-   </tgroup>
-  </table>
+   <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('mysqli-result.fetch-fields.returnvalues')/db:table)">
+    <xi:fallback/>
+   </xi:include>
   </para>
```
Girgias, line 48: Shouldn't the wrapping `<para>` tags be removed if the xml ID is on the `para` tag? (also I don't think we need those _anyway_)

reference/mysqli/mysqli_result/fetch-fields.xml
```diff
@@ -43,7 +43,6 @@
    Returns an array of objects containing field definition information.
 
   </para>
-  <para xml:id="mysqli-result.fetch-fields.returnvalues">
   <table>
```
Girgias, line 46: We still need an ID to XInclue, you shoule be able to move the ID on the `<table>` tag:

reference/mysqli/mysqli_result/fetch-fields.xml
```diff
@@ -41,9 +41,9 @@
   &reftitle.returnvalues;
   <para>
    Returns an array of objects containing field definition information.
+
   </para>
-  <para>
-  <table xml:id="mysqli-result.fetch-fields.returnvalues">
+  <table>
```
Girgias, line 46:
```suggestion
  <table xml:id="mysqli-result.fetch-fields.returnvalues">
```

reference/mysqli/mysqli_result/fetch-fields.xml
```diff
@@ -41,9 +41,9 @@
   &reftitle.returnvalues;
   <para>
    Returns an array of objects containing field definition information.
+
   </para>
```
Girgias, line 45:
```suggestion
  </para>

```

reference/mysqli/mysqli_result/fetch-fields.xml
```diff
@@ -42,7 +42,7 @@
   <para>
    Returns an array of objects containing field definition information.
   </para>
-  <para>
+
   <table xml:id="mysqli-result.fetch-fields.returnvalues">
```
Girgias, line 46:
```suggestion
  <table>
```

reference/mysqli/mysqli_result/fetch-fields.xml
Girgias: I'm dumb the XInclude is written in such a way to not refer to an ID

Girgias:
Thank you!

PR #3313, opened 2024-04-14
Update functions.arguments to include output of code samples

omahane:
# Description
On page https://www.php.net/manual/en/functions.arguments.php, examples 4 and 7 have both code and output, but examples 5 and 6 only have the code. This PR does a tiny bit of code change (to add line breaks in example 6)  but otherwise is about adding the output of examples 5 and 6.

# Screenshots
## Example 4
Screenshot 2024-04-14 at 10 08 53

## Examples 5 and 6
Screenshot 2024-04-14 at 10 12 07


Girgias (approved):
Thank you!

PR #3317, opened 2024-04-16
Document how to debug connection issues

RoSk0:
I've spent enormous amount of time trying to debug my connection issue. The main reason was lack of proper documentation on php.net which I'm trying to resolve here.

There is a TODO in the proposed change - the reason for it is that I'm not sure where that magical value `7` is coming from and what level it relates to. I've scanned the `php-src` code base, but found nothing useful. That TODO comment represents my findings.

### DIFF
--- a/reference/ldap/functions/ldap-set-option.xml
+++ b/reference/ldap/functions/ldap-set-option.xml
@@ -282,6 +282,23 @@ if (!ldap_set_option($ds, LDAP_OPT_SERVER_CONTROLS, array($ctrl1, $ctrl2))) {
 ]]>
     </programlisting>
    </example>
+   <example>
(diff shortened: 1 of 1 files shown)


MCMic (commented):
Sadly, seems true, we have the const for `LDAP_OPT_DEBUG_LEVEL` but not for each of the levels. Also sounds like this is specific to openldap and may behave differently depending on the versions.

@RoSk0 Could you clarify how the debug appears when this is used? Is it logged?

RoSk0:
Debug output goes to the STDOUT/STDERR. Not sure which as I was testing in the container environment- when script was called from CLI, the output was visible straight away,  when called via output was on on php-fpm container log.

Issue #3326, opened 2024-04-21
Migrate PHP Documentation schema to DocBook 5.2


Girgias:
DocBook 5.2 has finally been released: https://www.oasis-open.org/2024/02/13/the-docbook-schema-version-5-2-oasis-standard-published/

The primary motivation to migrate to this new version is to be able to use the new [`<enumsynopsis>` tag.](https://tdg.docbook.org/tdg/5.2/enumsynopsis)

However, there a some issues with the current documentation which makes this migration challenging.

## RelaxNG Schema

The new schema is only provided as a RelaxNG schema, whereas most of our tooling expects a DTD.
I have made progress in checking the documentation against it, but this required compiling [`jing-trang`](https://github.com/relaxng/jing-trang) locally to check what is causing issues. It is likely we will need to have this on CI somehow to improve the debugging experience for new contributors.

## Legacy custom markup

~~Classes, Interfaces, and Exceptions use a custom `phpdoc:` tag, and need to be converted to a structure that fits the DocBook schema, I have written a script that converts the relevant files to a new markup, but this does not validate the schema yet.~~

Turns out jumping directly to DocBook 5.2 simplifies some stuff as we just need to replace those tags with `<reference>`.

~~However, our "new" markup for extending classes/implementing interfaces doesn't comply :(~~
There seems to either be a bug in the DocBook 5.2 schema, or an undocumented BC break. Trying to contact the [relevant people](https://www.oasis-open.org/committees/comments/index.php?wg_abbrev=docbook) about this issue.

## Minor schema violation issues

Various files do not follow the DocBook schema properly and need to be fixed some bigger issues are:
 - [x] `features/commandline.xml`: Table on line 1705 has a warning in an entry (the previous PHP 5 changelog era table should be fully integrated and removed) (#3331)
 - [x] `reference/info/constants.xml`: Cannot use the warning entity in an `<entry>` tag in the `assert()` related constant table. Ideally, this constant table should be converted to a list. (#3342)
 - [x] `assert()` docs: ~~similar issue as above, cannot use `<methodsynopsis>` tags in a changelog entry~~ it is possible just need to wrap the markup properly. (#3343)
 - [x] `reference/pdo_cubrid/reference.xml`: (#3344)
 - [x] `reference/mysqlnd/stats.xml`: Cannot have a `<programlisting>` tag inside an `<entry>` tag. This whole page needs restructuring as using a table seems wrong. (#3357)
[...]

PR #3327, opened 2024-04-21
Fix minor markup conformance issue in DateTime extension

Girgias:
Part of https://github.com/php/doc-en/issues/3326, an `<example>` tag requires a `<title>`.

I couldn't come up with titles for the 3 examples, so I converted them to `<informalexample>` which don't have that requirement. @derickr if you have suggestions for titles to use instead, let me know.


derickr (approved):
This is fine, they don't really need a title.

PR #3328, opened 2024-04-22
Rewrite installation docs for Zip extension

Girgias:
This is part of #3326, a `<section>` tag needs to be immediately be followed by a `<title>` tag to conform to the DocBook schema.

At the same time, we take the opportunity to rewrite the documentation in such a way that it is easier to remove PHP 7 related docs when the time comes to remove them.

### DIFF
--- a/reference/zip/configure.xml
+++ b/reference/zip/configure.xml
@@ -6,7 +6,7 @@
  <section xml:id="zip.installation.linux">
   <title>Linux systems</title>
   <para>
-   In order to use these functions PHP must be compiled PHP with ZIP support
+   In order to use these functions PHP must be compiled with ZIP support
    by using the <option role="configure">--with-zip</option>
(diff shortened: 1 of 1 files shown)


reference/zip/configure.xml
```diff
@@ -3,42 +3,41 @@
...
-   <para>
-    A <option role="configure">--with-libzip=DIR</option>
-    configure option has been added to use a system libzip installation. libzip
-    version 0.11 is required, with 0.11.2 or later recommended.
-   </para>
-  </section>
+ <section xml:id="zip.installation.linux">
+  <title>Linux systems</title>
+  <para>
+   In order to use these functions PHP must be compiled PHP with ZIP support
```
Girgias, line 9:
```suggestion
   In order to use these functions PHP must be compiled with ZIP support
```

PR #3331, opened 2024-04-22
CLI docs: Remove PHP 5 era changelog and rework markup

Girgias:
To make it compliant with the DocBook schema.

Mark the introduction as not chunked so that the rendering of the manual doesn't just have a ToC: https://www.php.net/manual/en/features.commandline.php

### DIFF
--- a/features/commandline.xml
+++ b/features/commandline.xml
@@ -5,7 +5,7 @@
  <titleabbrev>Command line usage</titleabbrev>
  
  <!--Introduction: {{{-->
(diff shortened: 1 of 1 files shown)


kamil-tekiela (approved):

mmalferov:
Have you seen how `simplelist` is [rendered](https://www.php.net/manual/en/features.commandline.webserver.php)? I just wanted to make sure that the problem is known.

Screenshot 2024-04-24 152248

PR #3332, opened 2024-04-22
Add file modification history file

haszi:
Add file modification history file that contains the last modification date/time and the list of contributors for every file, and is used for [this feature](https://github.com/php/web-php/pull/977). The file also contains the commit hash of the last commit this file is based on.

This file was generated with [this script](https://gist.github.com/haszi/5284d2f2d61157dc65d5c47737f6fb28) but one contributor's name had to be changed as it contained a double quote which was manually replaced by `\"`.

*Edit:*
I replaced the escaped double-quotes (`\"`) with its HTML entity equivalent (`&qout;`) as this is easier to work with in the update script and needs no special handling later.


haszi:
I'm closing this since if these "history files" will ever be used in practice, they will not be part of the language repos so that the commit histories aren't polluted by their being updated constantly.

PR #3340, opened 2024-04-23
Document --with-ftp-ssl and --with-mysqlnd-ssl configure options

petk:
Added to php-src 8.4 via https://github.com/php/php-src/pull/14028

### DIFF
--- a/reference/ftp/configure.xml
+++ b/reference/ftp/configure.xml
@@ -8,9 +8,45 @@
   installing PHP.
  </para>
  <para>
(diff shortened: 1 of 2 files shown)


cmb69:
Maybe add that these are only for non Windows platforms.

kamil-tekiela (approved):

petk:
> Maybe add that these are only for non Windows platforms.

Added the word "Autotools" in there.

jimwins:
Thanks!

PR #3347, opened 2024-04-25
Update curl-exec.xml

php4fan:
Fixed an error in the description of the return value of curl_exec()

### DIFF
--- a/reference/curl/functions/curl-exec.xml
+++ b/reference/curl/functions/curl-exec.xml
@@ -34,7 +34,7 @@
   &reftitle.returnvalues;
   <para>
    &return.success; However, if the <constant>CURLOPT_RETURNTRANSFER</constant>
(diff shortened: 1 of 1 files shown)


mmalferov:
It seems to me that there is a reason why this commit has stalled. The constant [CURLOPT_FILE](https://www.php.net/manual/en/curl.constants.php#constant.curlopt-file) is deprecated as of cURL 7.9.7. We should probably use the [CURLOPT_WRITEFUNCTION](https://www.php.net/manual/en/curl.constants.php#constant.curlopt-writefunction) option instead ;-) Therefore, there is no need to mention the `CURLOPT_FILE` option in the documentation

jordikroon:
I am closing this issue as it's a minor change and needs a rebase regardless.. If you wish to pursue the suggested change, please open a new PR.

PR #3351, opened 2024-04-26
Fix incorrect reference to non-existent parameter in yaml_parse_file

davidburstrom:
It seems like a copy-paste from `yaml_parse`.


Girgias (approved):

davidburstrom:
NP!

PR #3360, opened 2024-04-30
pdo_cubrid/reference.xml: fix typo

mmalferov:


reference/pdo_cubrid/reference.xml
```diff
@@ -287,7 +287,7 @@ var_Dump($ret);
     &reftitle.notes;
     <note>
      <para>
-      When you estalish the connection to CUBRID, you should give username and
+      When you establish the connection to CUBRID, you should give username and
       password except DSN.
```
Girgias, line 291: I'm actually not sure what this note is attempting to say

cmb69: Frankly, I wouldn't worry about the wording, since that extension appears to be totally unmaintained (or whatever is going on there). https://pecl.php.net/package/CUBRID has it's latest release from 2017, but "browse source" refers to the old PHP SVN server, where the latest commit was "Tagging the 9.1.0.0001 release". I have no idea where the source code is developed now; the linked homepage http://www.cubrid.org/php_api_for_cubrid is 404.

Girgias: Is the "normal" CUBRID extension still maintained? If not we probably should remove it after the next archiving batch.

cmb69: Maybe @kadishmal can clarify?

kamil-tekiela: Should we drop the docs then?

cmb69: I found binaries and sources on https://www.cubrid.org/downloads (click PHP and/or PDO tab). These appear to be more recent than the PECL releases (although apparently they don't offer PHP 8 versions, yet). So I conclude that the PECL package has been abandoned (for whatever reason). As such, it may make sense to clarify that on PECL, and to also remove the documentation. But that should likely be discussed on the respective mailing list.

DanielEScherzer: While we can have that discussion separately, this is very clearly a typo that we can fix

DanielEScherzer (approved):

jordikroon:
@mmalferov Could you rebase this? Once done this can be merged. As Daniel mentioned. The discussion to drop this extension completely can be done separately and should not block these type of PRs imho.

PR #3362, opened 2024-05-01
Document overloads of XMLReader::open and ::XML

nielsdos:
Since these aren't deprecated or anything, I think we should document what exists now.


Girgias:
@kocsismate do we have a specific markup for such methods or not?

nielsdos:
FWIW, I just used the same markup that FFI::cdef used before the overload was yeeted out of that doc.

kocsismate:
I think this is fine syntax-wise, although the flip side is that we won't be able to synchronize this page with the stub from now on.

PR #3364, opened 2024-05-02
Add note for connecting to IBM i Db2 locally

jbh:
It took me over a week and contacting my good friend, @kadler, to finally figure out why my local connection kept failing. Sure, one could just pass the database name and not a DSN when using these functions locally, but frameworks like Doctrine force pass a DSN, and it's not optional. (See [IBMDB2/Driver.php](https://github.com/doctrine/dbal/blob/4.0.x/src/Driver/IBMDB2/Driver.php) and [IBMDB2/DataSourceName.php](https://github.com/doctrine/dbal/blob/4.0.x/src/Driver/IBMDB2/DataSourceName.php)).

After a lot of digging, it turns out that locally on IBM i, the connection string accepts only DSN, UID, and PWD. Using the string in the docs and the string that Doctrine builds from configurations does not work locally on IBM i. I plan to update their docs as well, but I believe this should be reflected in PHP's docs first.


Girgias:
@NattyNarwhal could you review this?

NattyNarwhal:
Is there IBM documentation we can cite here? Keep in mind with local connections can be a bit weird (i.e. running SQL/CLI in-job).

IIRC, the DSN should correspond to the relational database entries that you can modify with [`WRKRDBDIRE`](https://www.ibm.com/docs/en/i/7.4?topic=ssw_ibm_i_74/cl/wrkrdbdire.html).

jbh:
Sure, there is. When using `db2_connect` locally with a connection string instead of an explicit database name, it uses SQLDriverConnect. See [SQLDriverConnect's Usage](https://www.ibm.com/docs/en/i/7.4?topic=functions-sqldriverconnect-connect-data-source#rzadpfndvcon__title__5) docs, where it names the parameters as DSN (not DATABASE), UID, and PWD.

Furthermore, it's quite easy to test. On an IBM i:

```php
<?php

// Fails
db2_connect('DATABASE=dbname;HOSTNAME=localhost;UID=user;PWD=password', '', '');

// Succeeds
db2_connect('DSN=dbname;UID=user;PWD=password', '', '');
```

jbh:
Further details from Kevin, explaining how the two different parameter formats map and behave.

SQLConnect takes separate parameters for the DSN, user name, and password. This is what db2_connect('database', 'user', 'password') maps to.

The alternative is to use SQLDriverConnect which takes a single input parameter which is a connection string with semicolon-separated options. In this option there is no separate username/password fields so you must pass those (if applicable) in the UID/PWD options in the connection string. This is what you get when you use the db2_connect('DSN=dsn;UID=user;PWD=password;', '', '') syntax.

reference/ibm_db2/functions/db2-connect.xml
```diff
@@ -40,6 +40,14 @@
        <parameter>database</parameter> represents a complete connection
        string in the following format:
        <literallayout>DATABASE=<parameter>database</parameter>;HOSTNAME=<parameter>hostname</parameter>;PORT=<parameter>port</parameter>;PROTOCOL=TCPIP;UID=<parameter>username</parameter>;PWD=<parameter>password</parameter>;</literallayout>
+       <note>
+        <para>
+         The above format only works when connecting to IBM i Db2 remotely.
```
NattyNarwhal, line 45: I feel like there's a better way to phrase this. It should probably begin with the minimal set that's available on i, link documentation, then show the LUW(/z?) only options below those. Or mention about that those are only for LUW, and have separate paragraphs/bullets for each platform's supported options>

reference/ibm_db2/functions/db2-pconnect.xml
```diff
@@ -53,7 +53,67 @@
...
+       represents the database alias in the DB2 client catalog.
+      </para>
+      <para>
+       For an uncataloged connection to a database,
+       <parameter>database</parameter> represents a complete connection
+       string in the following format:
+       <literallayout>DATABASE=<parameter>database</parameter>;HOSTNAME=<parameter>hostname</parameter>;PORT=<parameter>port</parameter>;PROTOCOL=TCPIP;UID=<parameter>username</parameter>;PWD=<parameter>password</parameter>;</literallayout>
+       <note>
+        <para>
+         The above format only works when connecting to IBM i Db2 remotely.
```
NattyNarwhal, line 66: same for this block

NattyNarwhal (changes requested):
Sorry for late review, but I think there might be a better way to say this. See comments.

reference/ibm_db2/functions/db2-connect.xml
jbh: I agree about linking the docs, but do we really want to have the (what most would see as) edge case first? I feel like IBM i is the caveat here, not the other way around. Duplicating the paragraphs and bullets seems like overkill as well. Let me know what you think.

jbh: Friendly poke to keep this moving.

Could change the language to be more simple and not reference "the above format":

IBM i machines locally use [SQLDriverConnect](https://www.ibm.com/docs/en/i/7.4?topic=functions-sqldriverconnect-connect-data-source#rzadpfndvcon__title__5) which only accepts options DSN, UID, and PWD., like so:
         DSN=database;UID=username;PWD=password;

jbh:
Links added. @NattyNarwhal

NattyNarwhal (approved):
I think this is fine. I'll try to have someone else look at it, but I'm not sure if there's any better way to phrase this.

PR #3365, opened 2024-05-03
Add example on updating values with array_walk_recursive.

gitomato:


reference/array/functions/array-walk-recursive.xml
```diff
@@ -126,6 +106,41 @@ sour holds lemon
...
+  <para>
+   <example>
+    <title><function>array_walk_recursive</function> update example</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$sweet = array('a' => 'apple', 'b' => 'banana');
+$fruits = array('sweet' => $sweet, 'sour' => 'lemon');
+
+function test_update(&$item)
```
lacatoire:
```suggestion
function test_update(&$item, $key)
```

lacatoire: The callback signature includes $key for consistency with the documented callable contract of `array_walk_recursive`().

See implementation: https://github.com/php/php-src/blob/master/ext/standard/array.c

reference/array/functions/array-walk-recursive.xml
```diff
@@ -126,6 +106,41 @@ sour holds lemon
     </para>
    </example>
   </para>
+  <para>
+   <example>
+    <title><function>array_walk_recursive</function> update example</title>
```
cweiske: Please add a short explanatory text that awr can be used to update values. The title does not need to contain a_w_r as well, as we are on the awr page already.

cweiske (changes requested):

PR #3375, opened 2024-05-12
c14n.xml: amend a new-code

mmalferov:


nielsdos (approved):
If this is the style we normally use then fine by me :slightly_smiling_face:

PR #3385, opened 2024-05-15
Adding #[\SensitiveParameter] attribute

colshrapnel:
It seems that #[\SensitiveParameter] is already used for $password parameter so it should be reflected on the man page. Though I  was unable to find how to add attributes properly and also not sure it I added correct link. So this PR has to be supervised.

### DIFF
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -15,8 +15,7 @@
    <modifier>public</modifier> <methodname>mysqli::__construct</methodname>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>hostname</parameter><initializer>&null;</initializer></methodparam>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>username</parameter><initializer>&null;</initializer></methodparam>
-   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>#[\SensitiveParameter]</parameter><initializer>&null;</initializer></methodparam>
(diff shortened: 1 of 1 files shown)


reference/mysqli/mysqli/construct.xml
```diff
@@ -15,6 +15,7 @@
    <modifier>public</modifier> <methodname>mysqli::__construct</methodname>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>hostname</parameter><initializer>&null;</initializer></methodparam>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>username</parameter><initializer>&null;</initializer></methodparam>
+   <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>#[\SensitiveParameter]</parameter><initializer>&null;</initializer></methodparam>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>password</parameter><initializer>&null;</initializer></methodparam>
```
Girgias, line 19: SensitiveParameter is *not* a parameter of the function/method, it is an attribute *modifying* the behaviour of a parameter.

Using the `<modifier>` tag here is more appropriate, but this will need rendering support in PhD too.

One possible markup idea might be:
```suggestion
   <methodparam choice="opt"><modifier role="attribute">#[\SensitiveParameter]</modifier><type class="union"><type>string</type><type>null</type></type><parameter>password</parameter><initializer>&null;</initializer></methodparam>
```

Ideally it would also link to the relevant attribute automatically.

Girgias (changes requested):

kamil-tekiela:
Why was this PR closed?

TimexPeachtree (commented):
Great

PR #3387, opened 2024-05-17
Note that no-capture modifier (/n) support for preg_* was added in PHP 8.2

staabm:
as documented on https://php.watch/versions/8.2/preg-n-no-capture-modifier

### DIFF
--- a/reference/pcre/pattern.modifiers.xml
+++ b/reference/pcre/pattern.modifiers.xml
@@ -182,7 +182,7 @@
         Only named groups like <code>(?&lt;name&gt;xyz)</code> are capturing.
         This only affects which groups are capturing, it is still possible to
         use numbered subpattern references, and the matches array will still
-        contain numbered results.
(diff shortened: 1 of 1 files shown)


nielsdos (approved):
Thanks!

reference/pcre/pattern.modifiers.xml
```diff
@@ -182,7 +182,7 @@
         Only named groups like <code>(?&lt;name&gt;xyz)</code> are capturing.
         This only affects which groups are capturing, it is still possible to
         use numbered subpattern references, and the matches array will still
-        contain numbered results.
+        contain numbered results. Available since PHP 8.2.
```
Girgias, line 185: Should be ``Available as of PHP 8.2.0.``

staabm: Fix in https://github.com/php/doc-en/pull/3400

PR #3388, opened 2024-05-18
Add missing libxml related constants

haszi:
Add missing libxml related constants to ext/DOM and ext/libxml.

The references for these constants are:
- [XML_LOCAL_NAMESPACE](https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/include/libxml/tree.h?ref_type=heads)
- [XML_GLOBAL_NAMESPACE](https://github.com/GNOME/libxml2/commit/a4964b75003d138d4643ab03e3e116a8453f8308#diff-4d724327c74c8e7b349eb38258524a6228cc3e06b5d10f783557a8f38bcf737b) - Please note that this has been removed from libxml in October 2000 and should probably be dropped from PHP as well
- [LIBXML_LOADED_VERSION](https://github.com/GNOME/libxml2/blob/4fefba4cf63acbb17c9f9cee90b8a3b5213a2b8c/parser.c#L184-L189) - [the line in the arginfo file](https://github.com/php/php-src/blob/09d0e38ecffcae767505452a8ee853e4ed4ec6d3/ext/libxml/libxml_arginfo.h#L61) and [the internal PHP macro definition](https://github.com/php/php-src/blob/09d0e38ecffcae767505452a8ee853e4ed4ec6d3/ext/libxml/libxml.c#L47)

### DIFF
--- a/reference/dom/constants.xml
+++ b/reference/dom/constants.xml
@@ -271,7 +271,7 @@
       </entry>
       <entry>2</entry>
       <entry>
-       If the specified range of text does not fit into a 
+       If the specified range of text does not fit into a
        <classname>DOMString</classname>.
       </entry>
      </row>
@@ -386,7 +386,7 @@
       </entry>
       <entry>14</entry>
       <entry>
-       If an attempt is made to create or change an object in a way which is 
+       If an attempt is made to create or change an object in a way which is
        incorrect with regard to namespaces.
       </entry>
      </row>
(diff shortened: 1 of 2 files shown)


nielsdos:
I didn't even realize we still expose XML_GLOBAL_NAMESPACE... Should be deprecated and eventually removed as nothing can take on that value.

nielsdos (approved):

nielsdos:
Actually, I just looked into deprecating XML_GLOBAL_NAMESPACE, and it looks like the constant will never really be compiled in because we require libxml2 2.9.0 at the minimum, which doesn't have it anymore. So I'm going to drop it from the manual and drop it from php-src.

PR #3395, opened 2024-05-18
Add some missing intl constants

haszi:
References:

- [INTL_ICU_DATA_VERSION](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/uvernum.h#L151-L154)
- [ULOC_ACTUAL_LOCALE and ULOC_VALID_LOCALE](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/uloc.h#L339-L346)

### DIFF
--- a/reference/intl/constants.xml
+++ b/reference/intl/constants.xml
@@ -4,6 +4,17 @@
  &reftitle.constants;
  <para>
   <variablelist>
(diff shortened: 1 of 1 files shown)


devnexen (approved):

devnexen:
Thanks !

PR #3398, opened 2024-05-18
Add missing stream constants

haszi:


kocsismate:
OMG these are awesome! Thank you!

PR #3400, opened 2024-05-20
More precise version

staabm:


reference/pcre/pattern.modifiers.xml
```diff
@@ -182,7 +182,7 @@
         Only named groups like <code>(?&lt;name&gt;xyz)</code> are capturing.
         This only affects which groups are capturing, it is still possible to
         use numbered subpattern references, and the matches array will still
-        contain numbered results. Available since PHP 8.2.
+        contain numbered results. Available since PHP 8.2.0.
```
Girgias, line 185: Our style guide says to use `as of` instead of `since`.
```suggestion
        contain numbered results. Available as of PHP 8.2.0.
```

Girgias (approved):

PR #3402, opened 2024-05-20
Document SensitiveParameter attributes - part 1

kocsismate:


kocsismate:
cc @haszi

kamil-tekiela:
I didn't look thoroughly but both PRs look fine.

haszi:
I've noticed that while `ldap_connect()`'s [stub file](https://github.com/php/php-src/blob/8094bd1b58755d1d29877262e0ef92e2ba6dfc35/ext/ldap/ldap.stub.php#L626-L633) is showing a signature with 5 parameters with a fallback signature with 2 parameters, `doc-en` lists a signature with 1 parameter with the two parameter signature being deprecated as of PHP 8.3. This doesn't seem to be correct, is it?

```php
#ifdef HAVE_ORALDAP
    function ldap_connect(?string $uri = null, int $port = 389, string $wallet = UNKNOWN, #[\SensitiveParameter] string $password = UNKNOWN, int $auth_mode = GSLC_SSL_NO_AUTH): LDAP\Connection|false {}
#ifdef LDAP_API_FEATURE_X_OPENLDAP
    function ldap_connect_wallet(?string $uri = null, string $wallet, #[\SensitiveParameter] string $password, int $auth_mode = GSLC_SSL_NO_AUTH): LDAP\Connection|false {}
#endif
#else
    function ldap_connect(?string $uri = null, int $port = 389): LDAP\Connection|false {}
#endif
```
image

kocsismate:
> I've noticed that while ldap_connect()'s [stub file](https://github.com/php/php-src/blob/8094bd1b58755d1d29877262e0ef92e2ba6dfc35/ext/ldap/ldap.stub.php#L626-L633) is showing a signature with 5 parameters with a fallback signature with 2 parameters, doc-en lists a signature with 1 parameter with the two parameter signature being deprecated as of PHP 8.3. This doesn't seem to be correct, is it?

The `function ldap_connect(?string $uri = null, int $port = 389, string $wallet = UNKNOWN, #[\SensitiveParameter] string $password = UNKNOWN, int $auth_mode = GSLC_SSL_NO_AUTH): LDAP\Connection|false {}` signature is only available for Oracle, but this signature has never been documented.

So both signatures described in the manual page is correct, albeit the latter one is deprecated.

Girgias (approved):

PR #3407, opened 2024-05-22
Add new example to match using comparison operators

Marqitos:
Validation of operation of the example using Comparison Operators
https://onlinephp.io/c/a6825

Related to Spanish translation:
https://github.com/php/doc-es/pull/86

### DIFF
--- a/language/control-structures/match.xml
+++ b/language/control-structures/match.xml
@@ -76,7 +76,7 @@ var_dump($output);
   &example.outputs;
    <screen>
 <![CDATA[
-string(17) "You're a teenager"
+string(8) "Teenager"
 ]]>
    </screen>
   </example>
--- a/language/control-structures/match.xml
+++ b/language/control-structures/match.xml
@@ -29,7 +29,7 @@ $return_value = match (subject_expression) {
 ]]>
   </programlisting>
   
-   <example>
+  <example>
    <title>Basic <literal>match</literal> usage</title>
     <programlisting role="php">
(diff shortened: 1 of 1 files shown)


language/control-structures/match.xml
```diff
@@ -54,6 +54,33 @@ string(19) "This food is a cake"
...
+
+ $output = match (true) {
+     $age < 2 => "You're a baby",
+     $age < 13 => "You're a child",
+     $age <= 19 => "You're a teenager",
+     $age > 19 => "You're a young adult",
+     $age >= 40 => "You're an old adult"
+ };
+
+ var_dump($output);
```
Girgias, line 72: The indent is weird here.

Also we try to not use personalization in the docs (i.e. using "you") so could you change the strings to just be something like `"Baby"`, `"Child"`, etc.

Girgias (commented):
Makes sense overall, thank you!

language/control-structures/match.xml
```diff
@@ -54,6 +54,33 @@ string(19) "This food is a cake"
...
+
+var_dump($output);
+?>
+]]>
+  </programlisting>
+  &example.outputs;
+   <screen>
+<![CDATA[
+string(17) "You're a teenager"
+]]>
```
Girgias, line 80: Result needs fixing, but otherwise LGTM

Marqitos: Corrected, sorry for the work I have given you.

PR #3410, opened 2024-05-22
Fixed formats and example for localized notations under compound formats

genius257:


reference/datetime/formats.xml
```diff
@@ -784,7 +784,7 @@ object(DateTimeImmutable)#1 (3) {
     <tbody>
      <row>
       <entry>Common Log Format</entry>
-      <entry><literal>dd</literal> "/" <literal>M</literal> "/" <literal>YY</literal> : <literal>HH</literal> ":" <literal>II</literal> ":" <literal>SS</literal> <literal>space</literal> <literal>tzcorrection</literal></entry>
+      <entry><literal>DD</literal> "/" <literal>M</literal> "/" <literal>YY</literal> ":" <literal>HH</literal> ":" <literal>II</literal> ":" <literal>SS</literal> <literal>space</literal> <literal>tzcorrection</literal></entry>
```
derickr, line 787: Adding `"` around `:` is correct, but changing `DD` to `dd` is not. The parsing rule has:

```
/* Common Log Format: 10/Oct/2000:13:55:36 -0700 */
clf              = day "/" monthabbr "/" year4 ":" hour24lz ":" minutelz ":" secondlz space tzcorrection;
```

With `day` defined as:

```
day   = (([0-2]?[0-9]) | ("3"[01])) daysuf?;
```

the `dd` and `mm` rules **are** however missing in the table here (whereas `DD` and `MM` are included). These rows needs to be copied from the table for "Date Formats (https://www.php.net/manual/en/datetime.formats.php#datetime.formats.date)".

reference/datetime/formats.xml
```diff
@@ -839,8 +839,8 @@ object(DateTimeImmutable)#1 (3) {
      </row>
      <row>
       <entry>WDDX</entry>
-      <entry><literal>YY</literal> "-" <literal>mm</literal> "-" <literal>dd</literal> "T" <literal>hh</literal> ":" <literal>ii</literal> ":" <literal>ss</literal></entry>
-      <entry>"2008-7-1T9:3:37"</entry>
+      <entry><literal>YY</literal> "-" <literal>MM</literal> "-" <literal>DD</literal> "T" <literal>hh</literal> ":" <literal>ii</literal> ":" <literal>ss</literal></entry>
+      <entry>"2008-07-01T09:03:37"</entry>
```
derickr, line 843: This is not a correct change. `mm` and `dd` mean either one or two digits, where `MM` and `DD` format as leading-0 strings. 

The parsing rule in `parse_date.re` is:

```
wddx             = year4 "-" month "-" day "T" hour24 ":" minute ":" second;
```

with `month` and `day` specified as:

```
month = "0"? [0-9] | "1"[0-2];
day   = (([0-2]?[0-9]) | ("3"[01])) daysuf?;
```

Please remember that these are rules to **parse** these formats, and not output formats (there are no WDDX and CLF constants).

derickr (changes requested):

PR #3413, opened 2024-05-26
Add missing cURL error constants and clean up cURL error constants

haszi:
This PR moves the existing `CURLE_` error constans to their own `<variablelist>` and file, adds the six missing constants and adds descriptions to all constants that are available at the [libcurl error codes](https://curl.se/libcurl/c/libcurl-errors.html) page. 

Some constants have already been removed from libcurl so I couldn't find their descriptions. If anyone knows where I could find this information, please let me know.

Also, I didn't add any information on when constants were [added, deprecated and removed](https://curl.se/libcurl/c/symbols-in-versions.html) from libcurl but let me know if this would be useful and I'll add all that too.


reference/curl/constants_curl_error.xml
```diff
@@ -0,0 +1,850 @@
...
+ </varlistentry>
+ <varlistentry xml:id="constant.curle-http-returned-error">
+  <term>
+   <constant>CURLE_HTTP_RETURNED_ERROR</constant>
+   (<type>int</type>)
+  </term>
+  <listitem>
+   <simpara>
+    This is returned if <constant>CURLOPT_FAILONERROR</constant> is set &true;
+    and the HTTP server returns an error code that is &gt;= 400.
```
Girgias, line 454: Probably best to just write this out
```suggestion
    and the HTTP server returns an error code that is greater or equal than 400.
```

reference/curl/constants_curl_error.xml
```diff
@@ -0,0 +1,850 @@
...
+ <varlistentry xml:id="constant.curle-ssl-connect-error">
+  <term>
+   <constant>CURLE_SSL_CONNECT_ERROR</constant>
+   (<type>int</type>)
+  </term>
+  <listitem>
+   <simpara>
+    A problem occurred somewhere in the SSL/TLS handshake.
+    You really want the error buffer and read the message there
+    as it pinpoints the problem slightly more.
```
Girgias, line 705: Personalization should not be used in the docs.

reference/curl/constants_curl_error.xml
```diff
@@ -0,0 +1,850 @@
...
+ </varlistentry>
+ <varlistentry xml:id="constant.curle-too-many-redirects">
+  <term>
+   <constant>CURLE_TOO_MANY_REDIRECTS</constant>
+   (<type>int</type>)
+  </term>
+  <listitem>
+   <simpara>
+    Too many redirects. When following redirects, libcurl hit the maximum amount.
+    Set your limit with <constant>CURLOPT_MAXREDIRS</constant>.
```
Girgias, line 774: Ditto personalization

reference/curl/constants_curl_error.xml
```diff
@@ -0,0 +1,850 @@
...
+ <varlistentry xml:id="constant.curle-unsupported-protocol">
+  <term>
+   <constant>CURLE_UNSUPPORTED_PROTOCOL</constant>
+   (<type>int</type>)
+  </term>
+  <listitem>
+   <simpara>
+    The URL you passed to libcurl used a protocol that this libcurl does not support.
+    The support might be a compile-time option that you did not use,
+    it can be a misspelled protocol string or just a protocol libcurl has no code for.
```
Girgias, line 798: Ditto

Girgias (commented):
I'm guessing the issues with the descriptions already exist, but better to fix them now.

haszi:
> I'm guessing the issues with the descriptions already exist, but better to fix them now.

I've copied the descriptions from libcurl's website... :-)

Girgias (approved):

reference/curl/constants_curl_error.xml
```diff
@@ -0,0 +1,849 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kocsismate, line 1: This is a fantastic improvement, but the only problem with this approach is that gen_stub.php cannot detect that the constant list is included into the global constants page via an entity. :( Would it be possible to move them back?

Girgias: I would rather keep them split, would an XInclude solve the issue?

kocsismate: No, it wouldn't. gen_stub only has access to the source file, but it doesn't "run" it. I'm not sure if this limitation would be easy to be solved, or not though.

Girgias: Can't we then manually add the split out files for the script to check then?

kocsismate: I guess we could, but then it means that those separate files have to be provided for gen_stub one way or another (e.g. configuration or hard-coding the file name, just like I did for tokens: https://github.com/php/php-src/blob/0ad2e4d997549c85aac61120a94ea3da0b709339/build/gen_stub.php#L5329C54-L5329C60). BTW I'm wondering what the motivation is behind moving these constants away?

Girgias: Maintaining the cURL constants in translations is _very_ painful as there are so many, and it is hard to sync them up and to find what needs to update. It also means we group them together properly instead of adding them at random places in the list.

kocsismate: OK, then I think I'll just hard-code these file names into the regex ¯\_(ツ)_/¯ Another way to solve the issue which I can imagine is to add support for a file level PHPDoc tag which would make the provided file names included in the search. But I think the latter solution is a bit cumbersome.

PR #3414, opened 2024-05-26
Add missing PCNTL error constants

haszi:
Descriptions are from [this man page](https://www.man7.org/linux/man-pages/man3/errno.3.html) with the exception of the `ECAPMODE` constant which is from [some](https://man.freebsd.org/cgi/man.cgi?query=open&sektion=2&n=1) [BSD](https://man.freebsd.org/cgi/man.cgi?query=rename) [pages](https://man.freebsd.org/cgi/man.cgi?memfd_create).

### DIFF
--- a/reference/pcntl/constants.xml
+++ b/reference/pcntl/constants.xml
@@ -1297,8 +1297,10 @@
    </term>
    <listitem>
     <simpara>
-     Too many open files. Commonly caused by exceeding the RLIMIT_NOFILE resource limit.
-     Can also be caused by exceeding the limit specified in /proc/sys/fs/nr_open.
+     Too many open files. Commonly caused by exceeding
+     the <literal>RLIMIT_NOFILE</literal> resource limit.
+     Can also be caused by exceeding the limit specified in
+     <filename>/proc/sys/fs/nr_open</filename>.
     </simpara>
    </listitem>
   </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/pcntl/constants.xml
```diff
@@ -1155,6 +1155,280 @@
...
+  </varlistentry>
+  <varlistentry xml:id="constant.pcntl-emfile">
+   <term>
+    <constant>PCNTL_EMFILE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Too many open files. Commonly caused by exceeding the RLIMIT_NOFILE resource limit.
+     Can also be caused by exceeding the limit specified in /proc/sys/fs/nr_open.
```
Girgias, line 1301: The RMLIMIT_NOFILE should possibly be in a literal or constant tag, and the file path should be in a `filename` tag possibly with a relevant class attribute: https://tdg.docbook.org/tdg/5.2/filename

reference/pcntl/constants.xml
```diff
@@ -1155,6 +1155,280 @@
...
+  </varlistentry>
+  <varlistentry xml:id="constant.pcntl-enfile">
+   <term>
+    <constant>PCNTL_ENFILE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Too many open files in system.
+     On Linux, this is probably a result of encountering the /proc/sys/fs/file-max limit.
```
Girgias, line 1324: Ditto

Girgias (changes requested):

Girgias (approved):

PR #3416, opened 2024-05-27
dorequest.xml: format the sample code

mmalferov:


reference/soap/soapclient/dorequest.xml
```diff
@@ -111,31 +111,36 @@
...
-function Add($x, $y) {
-  return $x + $y;
-}
 
-class LocalSoapClient extends SoapClient {
-  private $server;
+function Add($x, $y)
+{
+    return $x + $y;
+}
```
Girgias, line 118: Why put the `{` on a new line?

mmalferov: I was based on an example of Coding Style of PSR-12:

> A function declaration looks like the following. Note the placement of parentheses, commas, spaces, and braces:

```
<?php

function fooBarBaz($arg1, &$arg2, $arg3 = [])
{
    // function body
}
```

;-)

Girgias: blergh... I think the main issue here is that doing ad-hoc changes is not very useful, we don't have a consistent style of examples anyway.

mmalferov: In a quarter of the cases, the examples contain a line break before the opening curly brace:

https://github.com/search?q=repo%3Aphp%2Fdoc-en+%2F%5Efunction%28.*%3F%29%5Cn%5C%7B%2F&type=code

3/4 of examples doesn't contain:

https://github.com/search?q=repo%3Aphp%2Fdoc-en+%2F%5Efunction%28.*%3F%29%5C%7B%2F&type=code

You're right, we don't have a consistent style of examples. Why not start with this example? In any case, we won't find the truth. You have one style, I have another, and that guy has a third. This is a good reason to start following the PSR-12 in the examples ;-)

P. S. If the authority of the PSR-12 is not enough, I think I can move the opening curly brace to the same line as the function name. Just make a final decision and I'll obey it

Girgias: I'm not against using PSR-12 (/the CS PER), but what I would like to see if some tooling that can update the examples across the whole documentation, and that we can run in CI to enforce the coding style except if the `<programinglisting>` tag has a special `annotation` attribute because it is sensitive to whitespace or some other reason.

Because without that, I can see example flip-flopping and stop being conformant with a style.

mmalferov:
Are we merging these changes or closing the PR? :)

Issue #3074, opened 2024-01-06
Random\IntervalBoundary enum is not documented


Girgias:
The current markup we use for classes (and exceptions) uses a custom tag and thus is not standard DocBook.

This prevents us from upgrading the version of DocBook, namely using the prerelease version of DocBook 5.2 which introduces new enum tags which are similar to the `classsynopsis` tags.

Issue #3405, opened 2024-05-21
Support more than parameter attributes attributes


TimWolla:
since #3402 / php/phd#127 parameter attributes are supported, but other types of attributes are not yet.

Support should be added for all types of attribute. Class attributes are already in use (the `#[\Attribute]` attribute) and function / method attributes might be added with php/php-src#11293.

haszi:
@Girgias Based on the current class, function and method markup, and the new markup for parameter attributes, would the below markup work for class and function/method attributes?

```xml
<ooclass>
  <modifier role="attribute">AttributeName</modifier>
  <classname>DateTime</classname>
</ooclass>
```

```xml
<methodsynopsis>
  <modifier role="attribute">AttributeName</modifier><modifier>public</modifier> <type>string</type><methodname>CURLFile::getFilename</methodname>
  <void/>
</methodsynopsis>
```

Girgias:
I think the proposed markup makes sense, and is compliant with DocBook 5.2

haszi:
The class attribute on the class synopsis and function/method attribute on the function/method pages could look like this:

![class_attribute](https://github.com/php/doc-en/assets/13745441/fe6f7d55-59bb-48ce-8436-3e9e1a5e594d)

![function_method_attribute](https://github.com/php/doc-en/assets/13745441/9940e067-14c3-403a-8745-f1c23c6909f6)

haszi:
How should the method listings on class synopsis pages be handled? Attributes on above the method signature or on the same line?

![method_list_attribute_1](https://github.com/php/doc-en/assets/13745441/e1a0aab7-dfa9-4c35-b22d-c5f3b86f00f2)
![method_list_attribute_2](https://github.com/php/doc-en/assets/13745441/c95e892c-f78c-4c04-afa6-0cc2a4e9f693)

Girgias:
I think for methods it probably should be above the method signature, @TimWolla @kocsismate what do you think?

TimWolla:
I agree. Putting it above the method would match the common userland style guides and it would better scale with a larger number of attributes (e.g. both `#[\Deprecated]` and `#[\Override]`).

Please also add the leading backslash for copy-and-paste safety with regard to namespaces.

kocsismate:
Yes, above please! :)

PR #3419, opened 2024-05-27
8.3: Document Random\IntervalBoundary

Girgias:
Part of: https://github.com/php/doc-en/issues/2796

Fixes:
 - https://github.com/php/doc-en/issues/3074

Requires:
 - https://github.com/php/doc-en/issues/3326
 - PhD support to render it like an enum (https://github.com/php/phd/issues/89)

I kinda wish I could use a `<refentry>` tag instead, but we need to insert this into a `<book>` element which requires an intermediate tag anyway, so steal the suboptimal markup from classes.

@haszi do you think this makes sense, and is this not too complicated to add support to PhD for it?

### DIFF
--- a/reference/random/random.intervalboundary.xml
+++ b/reference/random/random.intervalboundary.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8"?>
+<reference xmlns="http://docbook.org/ns/docbook" xml:id="enum.intervalboundary" role="enum">
+ <title>IntervalBoundary</title>
+ <titleabbrev>IntervalBoundary</titleabbrev>
+
+ <partintro>
+  <section xml:id="enum.intervalboundary.intro">
+   &reftitle.intro;
+   <simpara>
+    Intro
+   </simpara>
+  </section>
+
+  <section xml:id="enum.intervalboundary.synopsis">
+   <!-- TODO New Reftitle entity -->
+   &reftitle.classsynopsis;
+
+   <enumsynopsis>
+    <enumname>IntervalBoundary</enumname>
+    <!-- <modifier role="enum_backing_type">string|int</modifier> -->
+    
+    <enumitem>
+     <enumidentifier>ClosedOpen</enumidentifier>
+     <!--
+     <enumvalue>
+      <literal>"string value"</literal>
+      <literal>25</literal>
+     </enumvalue>
+     -->
+     <enumitemdescription>
+      Description of enum case.
+     </enumitemdescription>
+    </enumitem>
+
+    <enumitem>
+     <enumidentifier>ClosedClosed</enumidentifier>
+     <!--
+     <enumvalue>
+      <literal>"string value"</literal>
+      <literal>25</literal>
+     </enumvalue>
+     -->
+     <enumitemdescription>
+      Description of enum case.
+     </enumitemdescription>
+    </enumitem>
+
+    <enumitem>
+     <enumidentifier>OpenClosed</enumidentifier>
+     <!--
+     <enumvalue>
+      <literal>"string value"</literal>
+      <literal>25</literal>
+     </enumvalue>
+     -->
+     <enumitemdescription>
+      Description of enum case.
+     </enumitemdescription>
+    </enumitem>
+
+    <enumitem>
+     <enumidentifier>OpenOpen</enumidentifier>
+     <!--
+     <enumvalue>
+      <literal>"string value"</literal>
+      <literal>25</literal>
+     </enumvalue>
+     -->
+     <enumitemdescription>
+      Description of enum case.
+     </enumitemdescription>
+    </enumitem>
+
+   </enumsynopsis>
+  </section>
+ </partintro>
+</reference>
+<!-- Keep this comment at the end of the file
+Local variables:
+mode: sgml
+sgml-omittag:t
+sgml-shorttag:t
+sgml-minimize-attributes:nil
+sgml-always-quote-attributes:t
+sgml-indent-step:1
+sgml-indent-data:t
+indent-tabs-mode:nil
+sgml-parent-document:nil
+sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+sgml-exposed-tags:nil
+sgml-local-catalogs:nil
+sgml-local-ecat-files:nil
+End:
+vim600: syn=xml fen fdm=syntax fdl=2 si
+vim: et tw=78 syn=sgml
+vi: ts=1 sw=1
+-->
--- a/reference/random/random.intervalboundary.xml
+++ b/reference/random/random.intervalboundary.xml
@@ -7,9 +7,9 @@
   <section xml:id="enum.random.intervalboundary.intro">
    &reftitle.intro;
    <simpara>
-    The <enumname>Random\IntervalBoundary</enumname> cases determines if the
-    boundary values can be returned or not by
-    <methodname>Random\Randomizer::getFloat</methodname>.
+    The <enumname>Random\IntervalBoundary</enumname> enum specifies
+    whether an interval includes the boundary values within the set of values
+    lying in the interval.
    </simpara>
   </section>
 
@@ -22,30 +22,34 @@
     <enumitem>
      <enumidentifier>ClosedOpen</enumidentifier>
      <enumitemdescription>
-      The <parameter>min</parameter> value may be returned.
-      However, the <parameter>max</parameter> value will never be returned.
+      A left-open interval.
+      The lower boundary is included in the interval,
+      the upper boundary is not.
      </enumitemdescription>
     </enumitem>
 
     <enumitem>
      <enumidentifier>ClosedClosed</enumidentifier>
      <enumitemdescription>
-      Both the <parameter>min</parameter> and <parameter>max</parameter> value may be returned.
+      A closed interval.
+      Both boundary values are included in the interval.
      </enumitemdescription>
     </enumitem>
 
     <enumitem>
      <enumidentifier>OpenClosed</enumidentifier>
      <enumitemdescription>
-      The <parameter>max</parameter> value may be returned.
-      However, the <parameter>min</parameter> value will never be returned.
+      A right-open interval.
+      The upper boundary is included in the interval,
+      the lower boundary is not.
      </enumitemdescription>
     </enumitem>
 
     <enumitem>
      <enumidentifier>OpenOpen</enumidentifier>
      <enumitemdescription>
-      Neither the <parameter>min</parameter> or <parameter>max</parameter> will ever be returned.
+      An open interval.
+      Neither boundary value is included in the interval.
      </enumitemdescription>
     </enumitem>
 
--- a/reference/random/random.intervalboundary.xml
+++ b/reference/random/random.intervalboundary.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <reference xmlns="http://docbook.org/ns/docbook" xml:id="enum.random.intervalboundary" role="enum">
- <title>Random\IntervalBoundary</title>
+ <title>The Random\IntervalBoundary Enum</title>
  <titleabbrev>Random\IntervalBoundary</titleabbrev>
 
  <partintro>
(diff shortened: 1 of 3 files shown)


reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8"?>
+<reference xmlns="http://docbook.org/ns/docbook" xml:id="enum.intervalboundary" role="enum">
```
TimWolla, line 2:
```suggestion
<reference xmlns="http://docbook.org/ns/docbook" xml:id="enum.random-intervalboundary" role="enum">
```

haszi:
> @haszi do you think this makes sense, and is this not too complicated to add support to PhD for it?

This is very much in line with the new class markup you used which is great. PhD support should not be an issue and most of the work will be done when the new class markup is converted.

I assume the rendered enum synopsis would look something like this:

enum_synopsis

TimWolla:
@haszi Note that related to #3405, enum cases can also have attributes (they are class constants internally). So it's probably to take that into account right away.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,97 @@
...
+      <literal>"string value"</literal>
+      <literal>25</literal>
+     </enumvalue>
+     -->
+     <enumitemdescription>
+      Description of enum case.
+     </enumitemdescription>
+    </enumitem>
+
+   </enumsynopsis>
```
kocsismate, line 74: Hmm, so far `classsynopsis` elements only contained the signature itself, but not the descriptions. Is it intentional that it's different for `enumsynopsis` elements?

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,97 @@
...
+   <simpara>
+    Intro
+   </simpara>
+  </section>
+
+  <section xml:id="enum.intervalboundary.synopsis">
+   <!-- TODO New Reftitle entity -->
+   &reftitle.classsynopsis;
+
+   <enumsynopsis>
```
kocsismate, line 18: Just a FYI: we should also think about the markup for implemented interfaces, enum constants, methods

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,97 @@
...
+    <enumname>IntervalBoundary</enumname>
+    <!-- <modifier role="enum_backing_type">string|int</modifier> -->
+    
+    <enumitem>
```
kocsismate, line 22: Other types of symbols have a dedicated section title for classes and interfaces. Would it be a good idea to add them for enums as well?

reference/random/random.intervalboundary.xml
Girgias: The DocBook 5.2 markup describes the markup in that manner: https://tdg.docbook.org/tdg/5.2/enumsynopsis (via the `<enumitem>` tag) so yes.

For reference, the `<classsynopsis>` description is located here: https://tdg.docbook.org/tdg/5.2/classsynopsis

reference/random/random.intervalboundary.xml
Girgias: Right... I forgot that PHP enums can do this sort of stuff...

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,97 @@
...
+  </section>
+
+  <section xml:id="enum.intervalboundary.synopsis">
+   <!-- TODO New Reftitle entity -->
+   &reftitle.classsynopsis;
+
+   <enumsynopsis>
+    <enumname>Random\IntervalBoundary</enumname>
+    <!-- <modifier role="enum_backing_type">string|int</modifier> -->
+
```
Girgias, line 21:
```suggestion

```

reference/random/random.intervalboundary.xml
kocsismate: That's a pity, because this way automatic replacement is going to be much more difficult, since there are free form values in the generated enumsynopsis item, and we should somehow try to preserve them.

Girgias: > That's a pity, because this way automatic replacement is going to be much more difficult, since there are free form values in the generated enumsynopsis item, and we should somehow try to preserve them.

I'm not sure whether that's difficult? One just needs to ignore the `<enumitemdescription>` tag, but also, I don't see how an enum would change after it has been introduced.

kocsismate: classsynopsis elements are generated based on the current stub definitions for each class/interface/enum signature which are encountered on class synopses pages (https://github.com/php/php-src/blob/0ad2e4d997549c85aac61120a94ea3da0b709339/build/gen_stub.php#L5560) and then the respective elements are replaced altogether one by one (the `replaceAndCompareXmls()` call a few lines below).

This means that if there is anything inside  classsynopsis elements which are not generatable based on the definitions in stubs, then those things will be removed. A workaround for enums could be that the enum case descriptions are retrieved from the manual pages and they are inserted into the generated elements... This complicates the process, but maybe it's not that difficult if this is going to be the preferred approach.

Also, It has just came to my mind that these descriptions would result in enum definitions which are completely not valid PHP code 🤔  while for the rest of the symbols, we generate at least a remotely valid PHP code since a few years.

reference/random/random.intervalboundary.xml
kocsismate: Yeah, and I've just realized that e.g. `fieldsynopsis` elements inside `enumsynopsis` are not even allowed in DocBook. :/

kocsismate:
I've just drafted a PR which adds support for this feature in stubs: https://github.com/php/php-src/pull/14369 I intentionally modified Random\IntervalBoundary and also checked in the generated result for testing purposes.

reference/random/random.intervalboundary.xml
Girgias: Yeah, they are not, because DocBook (rightfully) assumes that enums are just that, and not our weird thing based on objects.

I think we might need to use the `<classsynopsis>` tags when an enum has constants, properties, or methods. However, I would imagine this to be rare for internal enums.

haszi:
Since all enums implement either the [UnitEnum](https://www.php.net/manual/en/class.unitenum.php) or the [BackedEnum](https://www.php.net/manual/en/class.backedenum.php) interface and inherit their default method implementation, all enum markup will have to include both an `<enumsynopsis>` and a `<classsynopsis>` element. Neither of these can have the other as its child but we could have one after the other in the synopsis `<section>` like this (I think this is valid DocBook 5.2 markup):

```php
 <section xml:id="enum.intervalboundary.synopsis">
   &reftitle.classsynopsis;

  <enumsynopsis>
   ... // enum (cases) markup
  </enumsynopsis>

  <classsynopsis>
   ... // regular class markup
  </classsynopsis>
 </section>
```

`<enumname>` in the `<enumsynopsis>` element would define the enum's name and any class/interface name in the `<classynopsis>` element would be ignored.

What do you think?

kocsismate:
> <enumname> in the <enumsynopsis> element would define the enum's name and any class/interface name in the <classynopsis> element would be ignored.

Hmm, interesting option... I wouldn't have thought about it. But I think I'd prefer to just have one element which has to be `classynopsis` because of the restrictions we face.

TimWolla:
This will now also be required for `RoundingMode` (and some other enums) in 8.4.

cmb69:
Hmm, this is ugly. We can't use `<enumsynopsis>` due to its limitation, and we can't use `<classsynopsis>` because its `class` attribute only supports "class" and "interface". I'm afraid we need to redesign enums altogether. ;)

Okay, let's be pragmatic and use `<classsynopsis>` (and `role="enum"` if necessary).

Girgias:
> Hmm, this is ugly. We can't use `<enumsynopsis>` due to its limitation, and we can't use `<classsynopsis>` because its `class` attribute only supports "class" and "interface". I'm afraid we need to redesign enums altogether. ;)
> 
> Okay, let's be pragmatic and use `<classsynopsis>` (and `role="enum"` if necessary).

We will need the `role="enum"` as otherwise the generated markup by PhD will make it look like classes.

I *still* don't think any internal enum is going to actually implement interfaces or have any sort of properties, even if this is allowed in userland. So the "problem" seems rather theoretical, and even then it would be possible to add the above "hack fix" with the `role="enum"`.

Girgias:
> > I _still_ don't think any internal enum is going to actually implement interfaces or have any sort of properties, even if this is allowed in userland. So the "problem" seems rather theoretical, and even then it would be possible to add the above "hack fix" with the `role="enum"`.
> 
> I'm a bit worried about translations. If we have a number of enum documented with `<enumsynopsis>`, and later need to change everything …
> 
> On the other hand, maybe we will never need more, and I'm overthinking.

What I meant, is that it would be perfectly fine for both markups to coexist, where the `<classsynopsis class="class" role="enum">` usage would be reserved for "complex" enums that actually behave like classes.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
+<reference xmlns="http://docbook.org/ns/docbook" xml:id="enum.random.intervalboundary" role="enum">
+ <title>Random\IntervalBoundary</title>
```
TimWolla, line 3:
```suggestion
 <title>The Random\IntervalBoundary Enum</title>
```

For consistency with #4080.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,75 @@
...
+ <titleabbrev>Random\IntervalBoundary</titleabbrev>
+
+ <partintro>
+  <section xml:id="enum.random.intervalboundary.intro">
+   &reftitle.intro;
+   <simpara>
+    The <enumname>Random\IntervalBoundary</enumname> cases determines if the
+    boundary values can be returned or not by
+    <methodname>Random\Randomizer::getFloat</methodname>.
+   </simpara>
```
TimWolla, line 13: I'm not sure if I would make it specific to `Randomizer::getFloat()`. It's the only current use case and it will likely remain the only use case, but technically the enum would be reusable for other things. The detail that it will determine whether the boundaries can be returned is already explained in the `$boundary` parameter of `getFloat()`.

Perhaps something like:

> The Random\IntervalBoundary enum specifies whether or not an interval includes the boundary values within the set of values lying in the interval.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,75 @@
...
+
+   <enumsynopsis>
+    <enumname>Random\IntervalBoundary</enumname>
+
+    <enumitem>
+     <enumidentifier>ClosedOpen</enumidentifier>
+     <enumitemdescription>
+      The <parameter>min</parameter> value may be returned.
+      However, the <parameter>max</parameter> value will never be returned.
+     </enumitemdescription>
```
TimWolla, line 27: A left-open interval. The lower boundary is included in the interval, the upper boundary is not.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,75 @@
...
+    </enumitem>
+
+    <enumitem>
+     <enumidentifier>ClosedClosed</enumidentifier>
+     <enumitemdescription>
+      Both the <parameter>min</parameter> and <parameter>max</parameter> value may be returned.
```
TimWolla, line 33: A closed interval. Both boundary values are included in the interval.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,75 @@
...
+     </enumitemdescription>
+    </enumitem>
+
+    <enumitem>
+     <enumidentifier>OpenClosed</enumidentifier>
+     <enumitemdescription>
+      The <parameter>max</parameter> value may be returned.
```
TimWolla, line 40: A right-open interval. The upper boundary is included in the interval, the lower boundary is not.

reference/random/random.intervalboundary.xml
```diff
@@ -0,0 +1,75 @@
...
+      However, the <parameter>min</parameter> value will never be returned.
+     </enumitemdescription>
+    </enumitem>
+
+    <enumitem>
+     <enumidentifier>OpenOpen</enumidentifier>
+     <enumitemdescription>
+      Neither the <parameter>min</parameter> or <parameter>max</parameter> will ever be returned.
```
TimWolla, line 48: An open interval. Neither boundary value is included in the interval.

TimWolla:
Ah and: Should the PR be renamed? It's not just a proposal at this point, no?

TimWolla (approved):

PR #3422, opened 2024-05-28
Remove stream wrappers and context options from function method listing

haszi:
Add a role attribute with an appropriate value to all stream wrappers and context options. PhD is patched also to index these as non-functions/methods and with that these entries are removed from the function method listing.

*Edit:*
The PhD side of this fix is [this one](https://github.com/php/phd/pull/133).

### DIFF
--- a/language/context/ftp.xml
+++ b/language/context/ftp.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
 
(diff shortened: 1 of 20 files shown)


Girgias:
Can you move the whitespace commit into a separate PR so that I can `[skip-revcheck]` the second commit and apply it on translations?

Girgias:
Merged the other PR, can you rebase this one? :)

haszi:
Rebase done.

PR #3427, opened 2024-05-30
Variable Scope Globals Example #2 8.1.0 disclaimer

sn-dustin:
This PR adds a disclaimer as writing to `$GLOBALS` is no longer supported in PHP 8.1.0


sn-dustin:
Erroneous - only writing to the entire `$GLOBALS` array is not supported.

PR #3436, opened 2024-06-04
may also return "Unknown" if no description was found

divinity76:
see

php -r 'const XML_NS_ERR_UNDEFINED_NAMESPACE = 201; var_dump(xml_error_string(XML_NS_ERR_UNDEFINED_NAMESPACE));' string(7) "Unknown"

### DIFF
--- a/reference/xml/functions/xml-error-string.xml
+++ b/reference/xml/functions/xml-error-string.xml
@@ -38,7 +38,7 @@
   &reftitle.returnvalues;
   <para>
    Returns a string with a textual description of the error
-   <parameter>error_code</parameter>, or &null; if no description was found.
+   <parameter>error_code</parameter>, or &null; or string(7) "Unknown" if no description was found.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


divinity76:
update: per https://github.com/php/php-src/blob/bbbe56eecc6ff6b3971ba4f924d51c9aa11955c5/ext/xml/xml.c#L1390 it always returns "Unknown" when not found; never null

(imo that shit should be fixed in the next major php version but oh well)

nielsdos (changes requested):
This change isn't right.
If it's compiled with libexpat you can get a NULL value if there's no description: https://github.com/libexpat/libexpat/blob/bfd178c6350bbb46dbc71dae299a1d531468bf8a/expat/lib/xmlparse.c#L2415
Either our compat layer should be adapted, or we should add a NULL check for libexpat which translates the string to "Unknown". I think the former is the best one.

divinity76:
adapting the compat layer sounds best to me too, but that would be a (small) breaking change 🤔

nielsdos:
Yeah, although the compat layer is actually the default one so the impact may be larger...
As for wording: we don't use var_dump-style strings in the docs. You should use `<literal>Unknown</literal>`

nielsdos (approved):
Thanks!

PR #3437, opened 2024-06-04
fdiv: improve note about comparisons

alexislefebvre:
I used name of operators from the doc: https://www.php.net/manual/en/language.operators.comparison.php


Girgias (approved):

Girgias:
Thank you!

PR #3439, opened 2024-06-05
mongodb/architecture.xml: The code of the examples seems invalid

mmalferov:
Lines: 422-456.

In particular:

```
class ContainerClass implements MongoDB\BSON\Serializable {
  public $things = AnotherClass4 implements MongoDB\BSON\Serializable {
    private $elements = [ 0 => 'foo', 2 => 'bar' ];
    function bsonSerialize(): array {
      return $this->elements;
    }
  }
  function bsonSerialize(): array {
      return [ 'things' => $this->things ];
  }
} // => { "things" : { "0" : "foo", "2" : "bar" } }
```

Wrong `$things` definition and two `bsonSerialize()` method instead one. And `$elements` doesn't seem to be supposed to be in this example.

Etc.

I suggest replacing the code of the examples like this.

Or replace `AnotherClass4 — AnotherClass6` class names to `ContainerClass1 — ContainerClass3`

### DIFF
--- a/reference/mongodb/architecture.xml
+++ b/reference/mongodb/architecture.xml
@@ -424,39 +424,90 @@ class AnotherClass3 implements MongoDB\BSON\Serializable
 {
     private $elements = ['foo', 'bar'];
 
-    public function bsonSerialize(): array;
+    public function bsonSerialize(): array
     {
         return $this->elements;
     }
 } // => {"0": "foo", "1": "bar"}
 
+/**
+ * Containered classes
+ */
+
 class AnotherClass4 implements MongoDB\BSON\Serializable
 {
-    private $things = [0 => 'foo', 2 => 'bar'];
+    private $elements = [0 => 'foo', 2 => 'bar'];
 
     public function bsonSerialize(): array
+    {
+        return $this->elements;
+    }
+}
+
+class ContainerClass1 implements MongoDB\BSON\Serializable
+{
+    public $things;
+
+    public function __construct()
+    {
+        $this->things = new AnotherClass4();
+    }
+
+    function bsonSerialize(): array
     {
         return ['things' => $this->things];
     }
 } // => {"things": {"0": "foo", "2": "bar"}}
 
+
 class AnotherClass5 implements MongoDB\BSON\Serializable
 {
-    private $things = [0 => 'foo', 2 => 'bar'];
+    private $elements = [0 => 'foo', 2 => 'bar'];
 
     public function bsonSerialize(): array
     {
-        return ['things' => array_values($this->elements)];
+        return array_values($this->elements);
+    }
+}
+
+class ContainerClass2 implements MongoDB\BSON\Serializable
+{
+    public $things;
+
+    public function __construct()
+    {
+        $this->things = new AnotherClass5();
+    }
+
+    public function bsonSerialize(): array
+    {
+        return ['things' => $this->things];
     }
 } // => {"things": ["foo", "bar"]}
 
+
 class AnotherClass6 implements MongoDB\BSON\Serializable
 {
-    private $things = ['foo', 'bar'];
+    private $elements = ['foo', 'bar'];
+
+    function bsonSerialize(): object
+    {
+        return (object) $this->elements;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


reference/mongodb/architecture.xml
```diff
@@ -386,83 +386,93 @@ foreach ($managers as $manager) {
...
-    private $elements = [ 'foo', 'bar' ];
-    function bsonSerialize(): array {
-      return (object) $this->elements;
+} // => {"things": {"0": "foo", "1": "bar"}}
+
+class UpperClass implements MongoDB\BSON\Persistable
+{
+    public $foo = 42;
+    protected $prot = "wine";
+    private $fpr = "cheese";
```
jmikola, line 467:
```suggestion
    protected $prot = 'wine';
    private $fpr = 'cheese';
```

We can use single-quoted strings here for consistency with the earlier examples.

reference/mongodb/architecture.xml
```diff
@@ -386,83 +386,93 @@ foreach ($managers as $manager) {
...
+class stdClass
+{
+    public $foo = 42;
+} // => {"foo": 42}
+
+class MyClass
+{
+    public $foo = 42;
+    protected $prot = "wine";
+    private $fpr = "cheese";
```
jmikola, line 398:
```suggestion
    protected $prot = 'wine';
    private $fpr = 'cheese';
```

reference/mongodb/architecture.xml
```diff
@@ -386,83 +386,93 @@ foreach ($managers as $manager) {
...
+} // => {"foo": 42}
+
+class AnotherClass1 implements MongoDB\BSON\Serializable
+{
+    public $foo = 42;
+    protected $prot = "wine";
+    private $fpr = "cheese";
```
jmikola, line 405:
```suggestion
    protected $prot = 'wine';
    private $fpr = 'cheese';
```

reference/mongodb/architecture.xml
```diff
@@ -386,83 +386,93 @@ foreach ($managers as $manager) {
...
 
-class AnotherClass3 implements MongoDB\BSON\Serializable {
-  private $elements = [ 'foo', 'bar' ];
-  function bsonSerialize(): array {
-      return $this->elements;
-  }
-} // => { "0" : "foo", "1" : "bar" }
-
-class ContainerClass implements MongoDB\BSON\Serializable {
-  public $things = AnotherClass4 implements MongoDB\BSON\Serializable {
```
jmikola, line 423: This dates back to https://github.com/php/doc-en/commit/47d9aeda69eb31dfc5592844087caa069bf70228 (:wave: @derickr). I _think_ the original motivation was to demonstrate how nested Serializable instances are handled during BSON encoding. If so, I think the valid syntax should have been:

```php
AnotherClass4 implements MongoDB\BSON\Serializable {
    private $elements = [0 => 'foo', 2 => 'bar'];
  
    function bsonSerialize(): array {
        return $this->elements;
    }
} // => {"0": "foo", "2": "bar"}

class ContainerClass1 implements MongoDB\BSON\Serializable {
    public $things;

    public function __construct() {
        $this->things = new AnotherClass4();
    }

    function bsonSerialize(): array {
        return ['things' => $this->things];
    }
} // => { "things" : { "0" : "foo", "2" : "bar" } }
```

It looks like the old code examples were using some pseudo-code similar to [anonymous classes](https://www.php.net/manual/en/language.oop5.anonymous.php), which wouldn't have been fully supported at the time this was written. And even then, an anonymous class wouldn't have been usable to initialize a property in its declaration.

jmikola (commented):
If my comment on AnotherClass4 and ContainerClass1 is helpful, you can use that as a basis to derive AnotherClass5 and ContainerClass2, followed by AnotherClass6 and ContainerClass3.

But if that's too much bother, we can let this sit and I'll revisit this next week. Just let me know what you prefer.

I'd rather not merge as-is and lose the ContainerClass examples entirely, though.

jmikola (commented):
LGTM with some additional suggestions.

jmikola (approved):
Thanks @mmalferov.

PR #3442, opened 2024-06-09
Add replaceable tag to the appropriate nl_langinfo constants

haszi:
Add `<replaceable>` tag to the appropriate constants. 

Maybe adding some formatting to the `replaceable` class in the css file in `web-php` would be nice (e.g. making these italic **ABDAY_***(1-7)* instead of **ABDAY_(1-7)**).


Girgias (approved):

Girgias:
It would be nice to have some CSS support indeed

PR #3444, opened 2024-06-11
Update filectime.xml

KenChenRyan:
Info on what to do with the return variable. Inspired by the page for the filemtime function.

### DIFF
--- a/reference/filesystem/functions/filectime.xml
+++ b/reference/filesystem/functions/filectime.xml
@@ -37,7 +37,8 @@
   &reftitle.returnvalues;
   <para>
    Returns the time the file was last changed, &return.falseforfailure;.
-   The time is returned as a Unix timestamp.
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
I'm against this. I see no reason to point anyone to `date()` here.

cmb69:
@kamil-tekiela, I generally agree, but consistency also makes sense. As is, `date()` is referred to by `filemtime()` and `getlastmod()`, but not by `fileatime()` and `filectime()`.

kamil-tekiela:
> @kamil-tekiela, I generally agree, but consistency also makes sense. As is, `date()` is referred to by `filemtime()` and `getlastmod()`, but not by `fileatime()` and `filectime()`.

Ok, I didn't know that. In that, my opinion is that it would be better to do the opposite and remove `date()` from the other pages.

PR #3445, opened 2024-06-11
Update fileatime.xml

KenChenRyan:
Info on what to do with the return variable. Inspired by the page for the filemtime function.


Girgias:
It might make more sense to combine this and #3444 into one PR where you create a new XML entity in `language-snippets.ent` to have consistent wording.

PR #3447, opened 2024-06-12
Added CVEs to crypt

Sc00bz:
Also there should maybe be a note on `CRYPT_STD_DES` to state it's worse than `CRYPT_MD5` and `CRYPT_MD5` was EOLed because it was not good enough.

`CRYPT_STD_DES` has a fixed 25 iterations of DES.
`CRYPT_MD5` has a fixed 1000 iterations of MD5.

`CRYPT_STD_DES` is 102x faster to crack than `CRYPT_MD5` on a [RTX 4090](https://gist.github.com/Chick3nman/32e662a5bb63bc4f51b847bb422222fd).

### DIFF
--- a/reference/strings/functions/crypt.xml
+++ b/reference/strings/functions/crypt.xml
@@ -69,7 +69,7 @@
    <listitem>
     <simpara>
      <constant>CRYPT_MD5</constant> - MD5 hashing with a twelve character salt starting with
-     $1$ (CVE-2012-3287)
+     $1$
     </simpara>
    </listitem>
    <listitem>
@@ -92,7 +92,7 @@
      is used to indicate how many times the hashing loop should be executed, much like the cost
      parameter on Blowfish. The default number of rounds is 5000, there is a minimum of
      1000 and a maximum of 999,999,999. Any selection of N outside this range will be truncated to
-     the nearest limit. (CVE-2016-20013)
+     the nearest limit.
     </simpara>
    </listitem>
    <listitem>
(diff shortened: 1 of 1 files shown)


Girgias:
I have no idea why mentioning those CVEs in the way you did improves the documentations. The point of the `crypt()` function is to be compatible with the OS function and other programming languages the use it.

The security issues with the function are well known and if *that* is the issue with this page it would make more sense to add a warning block somewhere in this page to warn about this.

Sc00bz:
I was trying for minimal changes but you're right. I turned it into a warning.

reference/strings/functions/crypt.xml
```diff
@@ -106,6 +106,14 @@
     </simpara>
    </listitem>
   </itemizedlist>
+  <warning>
+   <simpara>
+    <constant>CRYPT_MD5</constant> reached end of life because it is too fast for password hashing
+    (CVE-2012-3287). <constant>CRYPT_STD_DES</constant> is about 100 times faster for an attacker and
+    should also not be used. <constant>CRYPT_SHA256</constant> and <constant>CRYPT_SHA512</constant>
+    have a DoS attack and should not be used (CVE-2016-20013).
```
cmb69, line 114: Maybe better:
```suggestion
    are vulnerable to DoS attacks and should not be used (CVE-2016-20013).
```

cmb69 (commented):
This looks like a reasonable addition to the docs now (barring my suggestion maybe).

PR #3456, opened 2024-06-17
Move curl_multi_setopt() constants to a dedicated variablelist

haszi:
…and add descriptions. Descriptions were copied from the page of `curl_multi_setopt()`.


Girgias:
This now conflicts with another merge

PR #3460, opened 2024-06-18
Move CURLPAUSE constants to their own variablelist

haszi:


Girgias:
This now conflicts with another merge

PR #3462, opened 2024-06-19
Remove empty installation pages

haszi:
This PR removes the installation pages for every extension that doesn't need to be installed (ie. is part of PHP's core extensions).


TimWolla (commented):
ack for random

Girgias (approved):

PR #3463, opened 2024-06-20
WeakMap::offsetGet() does not return null

staabm:
see https://3v4l.org/QjCMI

see RFC https://wiki.php.net/rfc/weak_maps

> Reading a non-existent key results in an Error exception.

refs https://github.com/phpstan/phpstan/issues/10655

### DIFF
--- a/language/predefined/weakmap/offsetget.xml
+++ b/language/predefined/weakmap/offsetget.xml
@@ -34,8 +34,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns the value associated to the object passed as argument, &null;
-   otherwise.
+   Returns the value associated to the object passed as argument, throws a `Error` otherwise.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


language/predefined/weakmap/offsetget.xml
```diff
@@ -34,8 +34,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns the value associated to the object passed as argument, &null;
-   otherwise.
+   Returns the value associated to the object passed as argument, throws a <type>Error</type> otherwise.
```
Girgias, line 37: Errors should be in a new and dedicated Error/Exception section.

Girgias (changes requested):

language/predefined/weakmap/offsetget.xml
staabm: committed a fix. hopefully thats what you had in mind

Girgias: Yes that's what I had in mind, but now the mention of the Error should be removed from the return value.

And sorry for the delay.

staabm: Fixed

Girgias (approved):
Thank you

PR #3466, opened 2024-06-21
Added info about invert property

woytam:
Added info about `absolute` parameter and returned property `invert` of `DateInterval` object.

### DIFF
--- a/reference/datetime/datetimeinterface/diff.xml
+++ b/reference/datetime/datetimeinterface/diff.xml
@@ -69,6 +69,9 @@
    difference between the two dates.
   </para>
   <para>
+   The <parameter>absolute</parameter> parameter affects only <parameter>invert</parameter> property of <classname>DateInterval</classname> object.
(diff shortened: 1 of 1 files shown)


derickr (changes requested):
LGTM, after it's properly line-wrapped.

reference/datetime/datetimeinterface/diff.xml
```diff
@@ -68,6 +68,9 @@
    The <classname>DateInterval</classname> object represents the
    difference between the two dates.
   </para>
+  <para>
+   The <parameter>absolute</parameter> parameter affects only <parameter>invert</parameter> property of <classname>DateInterval</classname> object.
```
Girgias, line 72:
```suggestion
   The <parameter>absolute</parameter> parameter only affects the
   <parameter>invert</parameter> property of a
   <classname>DateInterval</classname> object.
```

PR #3468, opened 2024-06-21
Remove frontpage authors and editors

haszi:


cmb69:
In my opinion, that is a good idea, since that list is outdated anyway (to my knowledge). However, that might be something to discuss on the documentation discussion mailing list.

Potential problem with that: the ML no longer seems to work; at least [via NNTP](https://news-web.php.net/php.doc) there are no messages for almost half a year, and when I try to connect from my local NNTP client, it says "no such article found". Can anybody confirm that the mailing list is still active and this is only a NNTP issue? Maybe @jimwins knows more about that?

jimwins:
> Potential problem with that: the ML no longer seems to work; at least [via NNTP](https://news-web.php.net/php.doc) there are no messages for almost half a year, and when I try to connect from my local NNTP client, it says "no such article found". Can anybody confirm that the mailing list is still active and this is only a NNTP issue? Maybe @jimwins knows more about that?

I opened issue php/web-news#17 about that a couple of weeks ago, but don’t (yet) have access to the systems to debug or fix it myself. Should be easy to fix once I do. Waiting on @derickr on that.

haszi:
> In my opinion, that is a good idea, since that list is outdated anyway (to my knowledge). 

Also, all those listed on the front page are also listed in the Authors/Editors/User Note Maintainers sections of the documentations' [Preface page](https://www.php.net/manual/en/preface.php).

> However, that might be something to discuss on the documentation discussion mailing list.

I just checked the mailing lists and I'm not sure there is one that is actively used for discussing documentation. :-(

cmb69:
The documentation discussion mailing list used to be used for discussions. I don't know about the current state, since https://news-web.php.net/php.doc is no longer up-to-date (see the comments above).

derickr:
The mailing list works, but for some odd reason it has forgotten all the emails from before the Feb 14th mailing list server swap. Likely because the list always had odd message numbering. The new list is confused, and has deleted all the ones I converted (I have a backup though). This also confuses colobus because the new messages since Feb 14th are starting with ID 1 again.

derickr:
I fixed the `phpdoc` list now — the reason for this is that during the conversion between old and new server, we never actually migrated the archive of old messages, because they had a really odd numbering scheme. And in the hub hub of trying to deal with gmail non-sense, I had forgotten about this.

However, I spent some time this morning to address this, and it should now all work according to plan.

Let me know if you encounter something odd (at systems@php.net please).

jimwins:
The preface should also be updated to no longer say that most active people are mentioned on the front page.

haszi:
> The preface should also be updated to no longer say that most active people are mentioned on the front page.

I just pushed a commit with this change.

I also just wanted to send an email to the mailing list but noticed that you already did that yesterday, so thanks for that. :-)

Wirone:
This addresses my [concern](https://x.com/_Codito_/status/1587037353327067140) and I am all for it!

> (...) but maybe most importantly because that info just doesn't belong there.  Users want to see the TOC, and not a list of contributors.

💯!

cmb69 (approved):
Thank you! The patch looks fine. I suggest to wait a couple of days to give people time to raise concerns.

vrana:
The credits of 7 authors and the editor were simply removed without any replacement. The preface lists them only as past User Note Maintainers. I would be super-careful when removing someone's credit because that's pretty much the only reward.
#4605 is the fix.

PR #3469, opened 2024-06-22
appendices/about.xml: There are exactly 10 languages

mmalferov:


Girgias (commented):
Is the Polish language currently counted in this? As it is having a lot of activity and I'm guessing will be available soon not sure it makes sense to change this.

mmalferov:
It looks like you're right. I will close the PR

PR #3470, opened 2024-06-22
Fix typo in wrappers/file.xml

msamesch:
Is that even considered a typo?

### DIFF
--- a/language/wrappers/file.xml
+++ b/language/wrappers/file.xml
@@ -10,15 +10,19 @@
  <refsect1 role="description"><!-- {{{ -->
   &reftitle.description;
   <para>
-   <emphasis>Filesystem</emphasis> is the default wrapper used with PHP and represents the local filesystem.
-   When a relative path is specified (a path which does not begin with /, \, \\, or a Windows drive letter)
-   the path provided will be applied against the current working directory.  In many cases this is the
-   directory in which the script resides unless it has been changed.  Using the CLI sapi, this defaults
-   to the directory from which the script was called.
+   <emphasis>file://</emphasis> is the default wrapper used with PHP and
+   represents the local filesystem.
+   When a relative path is specified (a path which does not begin with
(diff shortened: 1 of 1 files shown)


language/wrappers/file.xml
```diff
@@ -10,15 +10,19 @@
  <refsect1 role="description"><!-- {{{ -->
   &reftitle.description;
   <para>
-   <emphasis>Filesystem</emphasis> is the default wrapper used with PHP and represents the local filesystem.
-   When a relative path is specified (a path which does not begin with /, \, \\, or a Windows drive letter)
-   the path provided will be applied against the current working directory.  In many cases this is the
-   directory in which the script resides unless it has been changed.  Using the CLI sapi, this defaults
-   to the directory from which the script was called.
+   <emphasis>file://</emphasis> is the default wrapper used with PHP and
```
Girgias, line 13: This should be a `literal` tag too I think.

language/wrappers/file.xml
```diff
@@ -10,15 +10,19 @@
...
+   represents the local filesystem.
+   When a relative path is specified (a path which does not begin with
+   <literal>/</literal>, <literal>\</literal>, <literal>\\</literal>, or a
+   Windows drive letter) the path provided will be applied against the current
+   working directory. In many cases this is the directory in which the script
+   resides unless it has been changed. Using the <acronym>CLI</acronym> sapi,
```
Girgias, line 19: SAPI should also use the acronym tag I think, as I'm pretty sure we have this defined.

Girgias (approved):
Minor nits, but lgtm and tahnk you!

language/wrappers/file.xml
msamesch: What about `filename`?

Girgias: That does also make sense

Girgias (approved):

msamesch:
Thanks for the review.

PR #3476, opened 2024-06-22
Add links to INI_* constants

haszi:
Replace XInclude'd variablelist with a duplicate which does not have any IDs for the constants. Add IDs to the variablelist on the constant page.

### DIFF
--- a/install/ini.xml
+++ b/install/ini.xml
@@ -262,6 +262,14 @@ $ PHP_INI_SCAN_DIR=/usr/local/etc/php.d: php
 
   <para>
    <variablelist>
+<!--
+Replace everything inside the <variablelist> element with an <xi:include>
+once libxml2 gets XInclude 1.1 attribute copying support.
+The below <xi:include> will include the appropriate elements
+but needs all top-level xml:id's removed (see XInclude 1.1 set-xml-id).
+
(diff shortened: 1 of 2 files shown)


Girgias:
The lack of support for XInclude 1.1 is really annoying....

haszi:
> The lack of support for XInclude 1.1 is really annoying....

If only we had attribute copying or `set-xml-id` attribute support from 1.1... :-)

haszi:
Even though this PR duplicates the table for the `INI_*` constants, I still think this makes sense as these are probably the most frequently used constants in the documentation. If there are no objections, I'd like to merge this around the end of next week.

jimwins:
Maybe leave the xi:include in a comment and mention that once whatever tool in the process that needs to add XInclude support does, this section can be fixed?

jimwins:
A breadcrumb for anyone interested in the future: it's `libxml2` that lacks XInclude 1.1 support, the spec is 8 years old, and the open issue for adding support in `libxml2` doesn't show any sign of interest: https://gitlab.gnome.org/GNOME/libxml2/-/issues/322

haszi:
> Maybe leave the xi:include in a comment and mention that once whatever tool in the process that needs to add XInclude support does, this section can be fixed?

@jimwins Would the comment in my last commit work?

jimwins:
@haszi Looks great, thanks!

Girgias (approved):

PR #3487, opened 2024-06-25
fopen.xml: add link to the function page

mmalferov:


mmalferov:
Or the meaning of the word `fopen` is a term for generic functions that enable to interact with streams like:

- file
- open
- fwrite
- fclose
- file_get_contents
- file_put_contents

And adding the link is not correct?

Girgias (approved):

PR #3497, opened 2024-06-27
socket-getpeername.xml: remove excessive length of refpurpose

mmalferov:
Or at least no longer than that, as example: `Queries the remote side of the given socket for a host/port or a file system path`


mmalferov:
Maybe or: `Polls the remote side of the given socket` (I don't have enough sense of English to shorten the title but keep the meaning)

Girgias (approved):
This sounds sensible to me, maybe @devnexen wants to have another look at it?

devnexen (approved):

PR #3499, opened 2024-06-28
constants.xml: add constant description

mmalferov:
Can we add a few words about these constants? The [socket_getpeername()](https://www.php.net/manual/en/function.socket-getpeername.php) function page mentions these constants, but there is not a word about them in the documentation


Girgias:
Can you rebase the PR?

mmalferov:
It's done

devnexen (approved):
nice

PR #3500, opened 2024-06-28
trim.xml: additionally describe the characters `parameter`

mmalferov:
```<?php

$string = "foo123bar";

var_dump(
    trim($string, "a..z"), // 123
    trim($string, "a..Z"), // Warning: trim(): Invalid '..'-range, '..'-range needs to be incrementing
);```

### DIFF
--- a/reference/strings/functions/trim.xml
+++ b/reference/strings/functions/trim.xml
@@ -80,7 +80,7 @@
        Optionally, the stripped characters can also be specified using
        the <parameter>characters</parameter> parameter.
        Simply list all characters that you want to be stripped. With
-       <literal>..</literal> you can specify a range of characters.
+       <literal>..</literal> you can specify an incrementing range of characters.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/strings/functions/trim.xml
+++ b/reference/strings/functions/trim.xml
@@ -80,7 +80,7 @@
        Optionally, the stripped characters can also be specified using
        the <parameter>characters</parameter> parameter.
        Simply list all characters that need to be stripped. With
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: reference/strings/functions/trim.xml [hunk 1]
```diff
@@ -80,7 +80,7 @@
        Optionally, the stripped characters can also be specified using
        the <parameter>characters</parameter> parameter.
        Simply list all characters that you want to be stripped. With
-       <literal>..</literal> you can specify a range of characters.
+       <literal>..</literal> you can specify an incrementing range of characters.
```
Girgias, line 83: Ideally, this shouldn't use personalization (i.e. usage of "you") to match the manual's style guide.

mmalferov: I have made changes and removed personalization ;-)

P. S. I'm sure you feel English much better than I do. It would be cool in cases where a change of wording in English is required, you suggested a change option. All I wanted in this commit was to add the word `incrementing` to the text ;-) At this stage,  I am not ready to create sentences in English that programmers all over the world will be read :-D

reference/strings/functions/trim.xml
```diff
@@ -79,8 +79,8 @@
       <para>
        Optionally, the stripped characters can also be specified using
        the <parameter>characters</parameter> parameter.
-       Simply list all characters that you want to be stripped. With
-       <literal>..</literal> you can specify a range of characters.
+       Simply list all characters that need to be stripped. With
+       <literal>..</literal> it is possible specify an incrementing range of characters.
```
Girgias, line 83:
```suggestion
       <literal>..</literal> it is possible to specify an incrementing range of characters.
```

PR #3502, opened 2024-06-28
Add missing OCI8 constants

haszi:


devnexen:
hmm considering that oci8 had been unbundled (recently), not sure it s worth the effort. cc @Girgias

haszi:
> hmm considering that oci8 had been unbundled (recently), 

True but the documentations include a lot of unbundled extension too

> not sure it s worth the effort. cc @Girgias

and the effort has been made already. :-)

Girgias (approved):
LGTM

We document everything from PHP 7.0 upwards, so this should be documented.

PR #3511, opened 2024-06-29
Add missing mysqli constants

haszi:
Add missing mysqli constants.
Rename mistyped `MYSQLI_TRANS_START_CONSISTENT_SNAPSHOT` constant to `MYSQLI_TRANS_START_WITH_CONSISTENT_SNAPSHOT`.

### DIFF
--- a/reference/mysqli/constants.xml
+++ b/reference/mysqli/constants.xml
@@ -36,6 +36,7 @@
      <simpara>
       Indicates to the server that the client can handle sandbox mode
       for expired passwords.
+      Can be used with <function>mysqli_options</function>.
      </simpara>
     </listitem>
    </varlistentry>
@@ -150,6 +151,7 @@
      <simpara>
       Indicates to the server that the client can handle sandbox mode
       for expired passwords.
+      Can be used with <function>mysqli_real_connect</function>.
      </simpara>
     </listitem>
    </varlistentry>
@@ -261,7 +263,9 @@
     </term>
     <listitem>
      <simpara>
-      Copy results from the internal <literal>mysqlnd</literal> buffer
+      As of PHP 8.1, this constants no longer has any effect.
+      Before PHP 8.1, this constant is used to copy results
+      from the internal <literal>mysqlnd</literal> buffer
       into the PHP variables fetched.
       By default, <literal>mysqlnd</literal> will use a reference logic
       to avoid copying and duplicating results held in memory.
@@ -804,28 +808,6 @@
      </para>
     </listitem>
    </varlistentry>
-   <varlistentry xml:id="constant.mysqli-no-data">
-    <term>
-     <constant>MYSQLI_NO_DATA</constant>
-     (<type>int</type>)
-    </term>
-    <listitem>
-     <para>
-      No more data available for bind variable
-     </para>
-    </listitem>
-   </varlistentry>
-   <varlistentry xml:id="constant.mysqli-data-truncated">
-    <term>
-     <constant>MYSQLI_DATA_TRUNCATED</constant>
-     (<type>int</type>)
-    </term>
-    <listitem>
-     <para>
-      Data truncation occurred. Available since MySQL 5.0.5.
-     </para>
-    </listitem>
-   </varlistentry>
    <varlistentry xml:id="constant.mysqli-enum-flag">
     <term>
      <constant>MYSQLI_ENUM_FLAG</constant>
@@ -928,17 +910,6 @@
      </para>
     </listitem>
    </varlistentry>
-   <varlistentry xml:id="constant.mysqli-set-charset-dir">
-    <term>
-     <constant>MYSQLI_SET_CHARSET_DIR</constant>
-     (<type>int</type>)
-    </term>
-    <listitem>
-     <simpara>
-      The path name to the directory that contains character set definition files.
-     </simpara>
-    </listitem>
-   </varlistentry>
    <varlistentry xml:id="constant.mysqli-report-index">
     <term>
      <constant>MYSQLI_REPORT_INDEX</constant>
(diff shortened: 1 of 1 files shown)


reference/mysqli/constants.xml
```diff
@@ -116,6 +141,40 @@
      </para>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="constant.mysqli-client-can-handle_expired_passwords">
+    <term>
+     <constant>MYSQLI_CLIENT_CAN_HANDLE_EXPIRED_PASSWORDS</constant>
```
kamil-tekiela, line 146: I wonder why we have two constants for this. If you know, it would be a good idea to add to the documentation.

reference/mysqli/constants.xml
```diff
@@ -195,6 +254,24 @@
      </para>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="constant.mysqli-store-result-copy_data">
+    <term>
+     <constant>MYSQLI_STORE_RESULT_COPY_DATA</constant>
```
kamil-tekiela, line 259: Huh, I thought this was already deprecated and due for removal. It looks like that's not the case. We should probably add it to the deprecations for PHP 8.4 as soon as possible. Also, this constant is already documented. https://www.php.net/manual/en/mysqli.store-result.php

reference/mysqli/constants.xml
```diff
@@ -836,6 +928,17 @@
      </para>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="constant.mysqli-set-charset-dir">
+    <term>
+     <constant>MYSQLI_SET_CHARSET_DIR</constant>
```
kamil-tekiela, line 933: That's not true. PHP doesn't support this feature and this constant has no effect in PHP. I don't know know how it got exposed to the userland, but documenting this seems pointless. Let's just remove the constant in PHP 8.4

reference/mysqli/constants.xml
```diff
@@ -932,6 +1035,29 @@
...
+     <constant>MYSQLI_SERVER_QUERY_WAS_SLOW</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+     Last statement took more than the time value specified
+     in server variable <literal>long_query_time</literal>.
+     </simpara>
+    </listitem>
+   </varlistentry>
```
kamil-tekiela, line 1060: I would recommend that we do not add this to the documentation. These constants are deprecated anyway and due for removal soon. I don't even know what is their use or if they do anything at all. If I remember correctly, I deprecated them because they have no use in PHP. These descriptions would only confuse users.

reference/mysqli/constants.xml
```diff
@@ -1040,6 +1166,17 @@
...
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="constant.mysqli-refresh-backup-log">
+    <term>
+     <constant>MYSQLI_REFRESH_BACKUP_LOG</constant>
+     (<type>int</type>)
+    </term>
+    <listitem>
+     <simpara>
+      Closes and reopens the backup log files.
```
kamil-tekiela, line 1176: Where did you get this from? What does it even mean? How is this constant used and what exactly does it do? 
I am not sure, but I feel like it's one of the dead constants that never got implemented in PHP. We probably should not add it to the documentation.

reference/mysqli/constants.xml
```diff
@@ -1144,6 +1280,29 @@
      </para>
     </listitem>
    </varlistentry>
+   <varlistentry xml:id="constant.mysqli-async">
+    <term>
+     <constant>MYSQLI_ASYNC</constant>
```
kamil-tekiela, line 1285: This is already documented https://www.php.net/manual/en/mysqli.query.php

kamil-tekiela (changes requested):

reference/mysqli/constants.xml
haszi: As of PHP 8.1, this constant has no effect so I'm adding that to the description.

haszi: > Also, this constant is already documented. https://www.php.net/manual/en/mysqli.store-result.php

The constants in this PR are part of the push to get all global constants onto their respective constants pages (and enable constant linking for all of them).

reference/mysqli/constants.xml
haszi: I'll leave the constant in the docs but remove the description. I'll remove the constant when this gets deprecated.

reference/mysqli/constants.xml
haszi: You're right, none of the six constants you deprecated in https://github.com/php/php-src/pull/6850 do anything (these two included). 

For the sake of consistency, I think we should either remove or leave all six in the docs until they are removed. What do you think?

reference/mysqli/constants.xml
haszi: From the mysql documentation. 

I just checked `php-src` and this is just like the six constants you deprecated, it doesn't do anything. However the other six are handled, this should be handled the same way at the same time I think.

reference/mysqli/constants.xml
haszi: This is due to the same reason as [above](https://github.com/php/doc-en/pull/3511#discussion_r1659966368).

reference/mysqli/constants.xml
kamil-tekiela: https://github.com/php/php-src/pull/14724

kamil-tekiela: There is no point in adding this to the documentation since it should not even exist in the first place. It's a bug.

reference/mysqli/constants.xml
kamil-tekiela: If any of these constants are linked to from anywhere else in the docs then leave them in the documentation but with a blank description. If they would only be listed on this single page, then just remove them altogether.

reference/mysqli/constants.xml
kamil-tekiela: Yeah, this would used in `mysqli_refresh()` but I can't find this in the MySQL docs. I assume it's part of some extension or add-on.

reference/mysqli/constants.xml
kamil-tekiela: ok, but I don't know if it's a good idea to duplicate this information in both places.

reference/mysqli/constants.xml
haszi: I found it on [this page](https://docs.oracle.com/cd/E19957-01/mysql-refman-6.0/news.html#news-6-0-8) of the Oracle website.

kamil-tekiela: Ok, leave like that.

reference/mysqli/constants.xml
haszi: @Girgias What is the oldest version of PHP we need to document? If it's 8.1, this constant could be removed (from `php-src` too).

Girgias: @haszi we currently document everything from PHP 7.0 onwards, so please keep it.

reference/mysqli/constants.xml
haszi: None of these are referenced anywhere in the documentation so I'm removing them.

kamil-tekiela (approved):
I think it's good enough now

PR #3514, opened 2024-06-29
Simplify Stringable example with PHP 8 syntax

Krinkle:
Stringable is introduced in PHP 8.0, so constructor promotion can be assumed.


Girgias (approved):
Makes sense to me.

PR #3520, opened 2024-07-03
Documents "No such file or directory" warning when calling `disk_*_space` functions.

timacdonald:
Documenting that both `disk_total_space` and `disk_free_space` will trigger a warning when specifying a disk that does not exist.

Screenshot 2024-07-03 at 10 20 27

### DIFF
--- a/reference/filesystem/functions/disk-free-space.xml
+++ b/reference/filesystem/functions/disk-free-space.xml
@@ -48,6 +48,14 @@
    Returns the number of available bytes as a float
    &return.falseforfailure;.
   </para>
+
+  <note>
+   <para>
+    When attempting to read a file system or disk partition that does not exist,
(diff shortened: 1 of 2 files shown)


reference/filesystem/functions/disk-free-space.xml
```diff
@@ -48,6 +48,14 @@
...
   </para>
+
+  <note>
+   <para>
+    When attempting to read a file system or disk partition that does not exist,
+    an <constant>E_WARNING</constant>-level error message will be issued and
+    the result will be &false;.
+   </para>
+  </note>
  </refsect1>
```
Girgias, line 59: This should be in a new Error/Exceptions sections to follow the usual manual structure.

Girgias (changes requested):

PR #3521, opened 2024-07-05
Update fwrite.xml

andreicir:
Fix parameter name ('string' to 'data')


cmb69:
Thank you, that looks correct.

PR #3528, opened 2024-07-07
Correct the type of $tidy->errorBuffer

nielsdos:
This is a nullable string: https://github.com/php/php-src/blob/e9254494ffde88af0160a30cee53c3dfce226edf/ext/tidy/tidy.stub.php#L866

### DIFF
--- a/reference/tidy/tidy/errorbuffer.xml
+++ b/reference/tidy/tidy/errorbuffer.xml
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <para>&style.oop; (property):</para>
-  <fieldsynopsis><modifier>public</modifier> <type>string</type><varname linkend="tidy.props.errorbuffer">tidy-&gt;errorBuffer</varname></fieldsynopsis>
+  <fieldsynopsis><modifier>public</modifier> <type class="union"><type>string</type><type>null</type></type><varname linkend="tidy.props.errorbuffer">tidy-&gt;errorBuffer</varname></fieldsynopsis>
   <para>&style.procedural;:</para>
   <methodsynopsis>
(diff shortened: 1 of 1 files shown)


reference/tidy/tidy/errorbuffer.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <para>&style.oop; (property):</para>
-  <fieldsynopsis><modifier>public</modifier> <type>string</type><varname linkend="tidy.props.errorbuffer">tidy-&gt;errorBuffer</varname></fieldsynopsis>
+  <fieldsynopsis><modifier>public</modifier> <type class="union"><type>string</type><type>null</type></type><varname linkend="tidy.props.errorbuffer">tidy-&gt;errorBuffer</varname></fieldsynopsis>
   <para>&style.procedural;:</para>
```
cmb69, line 13: RIght. It looks like the `tidy` properties are not documented according to the stubs (`tidy::$value` isn't documented at all). /cc @kocsismate

nielsdos: They actually are documented on the class overview page: https://www.php.net/manual/en/class.tidy
This page is about the error buffer itself.

cmb69: Oh, indeed! I was looking at https://www.php.net/manual/en/book.tidy.php, where `::$value` is missing. And also all properties of `tidyNode`. Might be a PhD issue; I'll see to it.

cmb69: Actually, there is no issue; it's just that `::$errorBuffer` is documented on its own page, while `::$value` is documented on the class page. Sorry for the noise.

Girgias (approved):

PR #3535, opened 2024-07-09
Update incompatible.xml: Move number_format() notice of out range() notice

ictgtvt:


Girgias (approved):
Oh yeah, that was strange, thanks for catching this!

PR #3536, opened 2024-07-09
Do not throw for deprecations in ErrorException example

TimWolla:
Incidentally this also fixes a deprecation in PHP 8.4 (nullable parameter).


Girgias (approved):

Girgias:
Thank you!

PR #3540, opened 2024-07-09
ev/book.xml: add the missing dot character

mmalferov:


reference/ev/book.xml
```diff
@@ -32,9 +32,7 @@
   </para>
   <para>
    For details refer to the
-   <link
- xlink:href="http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod">documentation
- of libev</link>
+   <link xlink:href="http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod">documentation of libev</link>.
```
Girgias, line 35: Is this link still working?

cmb69: It appears so, but it might be better to move the link to global.ent for easier maintainance of translations.

Girgias: Also a good point

philip: @Girgias and @cmb69 there are ~140 external non-entity links in `doc-en/`, do you think it's worth moving all to `global.ent` in one go? In terms of translation love. I was about to but decided to value my kneecaps first :)

Girgias: I think you can do it in one go. We do have quite active translations at the moment. :)

philip: fwiw, here was the commit to address this although it appears it missed the one for this issue. Maybe it was intentional (so it'd be handled separately by this PR/issue) or maybe not, I don't remember! https://github.com/php/doc-en/commit/e8ac70bf549a723cb36465667a6109d9933b8619

PR #3542, opened 2024-07-10
gmagick/setup.xml: add the missing dot character

mmalferov:


reference/gmagick/setup.xml
```diff
@@ -28,7 +28,7 @@
   </note>
   <para>
     For Windows, the build of the Gmagick extension is available
-    <link xlink:href="&url.gmagick.winbuilds;">here</link>
+    <link xlink:href="&url.gmagick.winbuilds;">here</link>.
```
Girgias, line 31: Maybe changing the "here" would also be a good idea, as I don't think this follows industry best practises

cmb69: Right. And that site appears to be down, anyway. Might be best to support gmagick builds via https://github.com/php/php-windows-builder, and to refer to PECL as usual.

cmb69: Ugh, easier said than done. Would need something like https://github.com/vitoc/gmagick/pull/60, and apparently, php-windows-builder doesn't run pre-releases, but gmagick apparently never had a stable release, and the project looks somewhat unmaintained.

Maybe just drop that paragraph?

kamil-tekiela:
Resolved via 330dda45b4875377dfae3586f005f57f6fcd8199

PR #3544, opened 2024-07-10
setup.xml: change the case of the letter `c` in the extention name: `…

mmalferov:
…libc`

### DIFF
--- a/reference/inotify/setup.xml
+++ b/reference/inotify/setup.xml
@@ -8,7 +8,7 @@
  <section xml:id="inotify.requirements">
   &reftitle.required;
   <para>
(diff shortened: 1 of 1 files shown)


reference/inotify/setup.xml
```diff
@@ -8,7 +8,7 @@
  <section xml:id="inotify.requirements">
   &reftitle.required;
   <para>
-   This extension requires Linux 2.6.13 or newer, and a recent libC.
+   This extension requires Linux 2.6.13 or newer, and a recent libc.
```
Girgias, line 11: Is it even possible to not satisfy these requirements?

Maybe the whole requirements section should be dropped?

Girgias (approved):

PR #3546, opened 2024-07-11
Add earlier hint about array_column supporting objects

Seldaek:
I know it says below in the parameter description, but tbh I didn't read that far before I assumed this only worked for arrays :D


cmb69:
As a user, reading the current documentation, I might wonder what a "column of an array" is (after all, there are no real multidimensional arrays in PHP). While the description of the `array` parameter clarifies that, the function description just repeats that "column of an array". Maybe that section should be updated to clarify what is meant by that, instead of changing the refpurpose (which is supposed to be as short as possible).

Seldaek:
Yeah I am not sure how to best describe all these things in a consistent way per the docs, but this reads fairly confusing I agree.

PR #3549, opened 2024-07-11
Update use cases for the mode parameter in PDOStatement::fetchAll()

colshrapnel:
Currently, case 3 makes no sense (PDO::FETCH_COLUMN|PDO::FETCH_UNIQUE would result in error) and case 4 is misleading (describing rather vanilla PDO::FETCH_GROUP's behavior). I added vanilla PDO::FETCH_UNIQUE and PDO::FETCH_GROUP cases and fixed PDO::FETCH_GROUP|PDO::FETCH_COLUMN.

Also, how does one remove a comment? 
- [88699](https://www.php.net/manual/en/pdostatement.fetchall.php#88699) is bogus, they need PDO::FETCH_UNIQUE but using PDO::FETCH_GROUP
- [122299](https://www.php.net/manual/en/pdostatement.fetchall.php#122299) same as above
- [115175](https://www.php.net/manual/en/pdostatement.fetchall.php#115175) is bogus, PDO::FETCH_UNIQUE|PDO::FETCH_GROUP makes no sense and effectively just PDO::FETCH_UNIQUE

### DIFF
--- a/reference/pdo/pdostatement/fetchall.xml
+++ b/reference/pdo/pdostatement/fetchall.xml
@@ -54,14 +54,22 @@
        <parameter>column</parameter> parameter.
       </para>
       <para>
-       To fetch only the unique values of a single column from the result set,
-       bitwise-OR <constant>PDO::FETCH_COLUMN</constant> with
-       <constant>PDO::FETCH_UNIQUE</constant>.
+       To index the resulting array by certain column's value (instead of
+       consecutive numbers), put this column's name first in the column
+       list in SQL, and use <constant>PDO::FETCH_UNIQUE</constant>.
+       This column must contain only unique values or some data will be lost.
       </para>
       <para>
-       To return an associative array grouped by the values of a specified
-       column, bitwise-OR <constant>PDO::FETCH_COLUMN</constant> with
-       <constant>PDO::FETCH_GROUP</constant>.
+       To group results in the form of 3-dimensional array indexed by values
+       of a specified column, put this column's name first in
+       the column list in SQL and use <constant>PDO::FETCH_GROUP</constant>.
+      </para>
+      <para>
+       To group results in the form of 2-dimensional array (which indices will
+       be values of one column and values will be list arrays of values
+       from another column, list only these 2 columns in the column list in SQL
+       and use bitwise-OR <constant>PDO::FETCH_GROUP</constant> with
+       <constant>PDO::FETCH_COLUMN</constant>.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/pdo/pdostatement/fetchall.xml
+++ b/reference/pdo/pdostatement/fetchall.xml
@@ -54,13 +54,13 @@
        <parameter>column</parameter> parameter.
       </para>
       <para>
-       To index the resulting array by certain column's value (instead of
+       To index the resulting array by a certain column's value (instead of
        consecutive numbers), put this column's name first in the column
        list in SQL, and use <constant>PDO::FETCH_UNIQUE</constant>.
        This column must contain only unique values or some data will be lost.
       </para>
       <para>
-       To group results in the form of 3-dimensional array indexed by values
+       To group results in the form of a 3-dimensional array indexed by values
        of a specified column, put this column's name first in
        the column list in SQL and use <constant>PDO::FETCH_GROUP</constant>.
       </para>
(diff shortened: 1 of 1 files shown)


reference/pdo/pdostatement/fetchall.xml
```diff
@@ -54,14 +54,22 @@
        <parameter>column</parameter> parameter.
       </para>
       <para>
-       To fetch only the unique values of a single column from the result set,
-       bitwise-OR <constant>PDO::FETCH_COLUMN</constant> with
-       <constant>PDO::FETCH_UNIQUE</constant>.
+       To index the resulting array by certain column's value (instead of
+       consecutive numbers), put this column's name first in the column
+       list in SQL, and use <constant>PDO::FETCH_UNIQUE</constant>.
+       This column must contain only unique values or some data will be lost.
```
Girgias, line 60: I think this reads better?
```suggestion
       To index the resulting array by a certain column's value (instead of
       consecutive numbers), put this column's name first in the column
       list in SQL, and use <constant>PDO::FETCH_UNIQUE</constant>.
       This column must contain only unique values or some data will be lost.
```

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -54,14 +54,22 @@
...
       </para>
       <para>
-       To return an associative array grouped by the values of a specified
-       column, bitwise-OR <constant>PDO::FETCH_COLUMN</constant> with
-       <constant>PDO::FETCH_GROUP</constant>.
+       To group results in the form of 3-dimensional array indexed by values
+       of a specified column, put this column's name first in
+       the column list in SQL and use <constant>PDO::FETCH_GROUP</constant>.
```
Girgias, line 65:
```suggestion
       To group results in the form of a 3-dimensional array indexed by values
       of a specified column, put this column's name first in
       the column list in SQL and use <constant>PDO::FETCH_GROUP</constant>.
```

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -54,14 +54,22 @@
...
+      </para>
+      <para>
+       To group results in the form of 2-dimensional array (which indices will
+       be values of one column and values will be list arrays of values
+       from another column, list only these 2 columns in the column list in SQL
+       and use bitwise-OR <constant>PDO::FETCH_GROUP</constant> with
+       <constant>PDO::FETCH_COLUMN</constant>.
```
Girgias, line 72:
```suggestion
       To group results in the form of a 2-dimensional array, whose indices will
       be the values of a column, and the associates values will be an array of values
       from a second column.
       Those 2 columns should be the only ones in the column list of the SQL,
       and <parameter>mode</parameter> must be a bitwise-OR of
       <constant>PDO::FETCH_GROUP</constant> with
       <constant>PDO::FETCH_COLUMN</constant>.
```

Girgias (commented):
I deleted the notes, and suggested some phrasing improvements. Please let me know if they are correct or not.

reference/pdo/pdostatement/fetchall.xml
kamil-tekiela: I like neither. How about this?

```suggestion
       To group results in the form of a 2-dimensional array
       use bitwise-OR <constant>PDO::FETCH_GROUP</constant> with
       <constant>PDO::FETCH_COLUMN</constant>. 
       The results will be grouped by the first column, with the array element's value
       being a list array of the corresponding entries from the second column. 
```

reference/pdo/pdostatement/fetchall.xml
```diff
@@ -54,14 +54,22 @@
...
+       To group results in the form of a 3-dimensional array indexed by values
+       of a specified column, put this column's name first in
+       the column list in SQL and use <constant>PDO::FETCH_GROUP</constant>.
+      </para>
+      <para>
+       To group results in the form of a 2-dimensional array
+       use bitwise-OR <constant>PDO::FETCH_GROUP</constant> with
+       <constant>PDO::FETCH_COLUMN</constant>. 
+       The results will be grouped by the first column, with the array element's value
+       being a list array of the corresponding entries from the second column.
```
kamil-tekiela, line 72:
```suggestion
       <constant>PDO::FETCH_COLUMN</constant>.
       The results will be grouped by the first column, with the array element's value
       being a list array of the corresponding entries from the second column.
```

kamil-tekiela (approved):
Thank you

PR #3553, opened 2024-07-11
strpos.xml: add the `parameter` tag to the offset

mmalferov:


reference/strings/functions/strpos.xml
```diff
@@ -59,7 +59,7 @@
   &reftitle.returnvalues;
   <para>
    Returns the position of where the needle exists relative to the beginning of
-   the <parameter>haystack</parameter> string (independent of offset).
+   the <parameter>haystack</parameter> string (independent of <parameter>offset</parameter>).
    Also note that string positions start at 0, and not 1.
```
cmb69, line 63: Maybe also mark up these numbers as `<literal>` right away, Other than that, this looks good to me.

mmalferov: Done: https://github.com/php/doc-en/pull/3553/commits/98f7028b1f3a9c14a4e8083d096da025b96a1632! ;)

cmb69 (approved):
Thank you!

PR #3554, opened 2024-07-11
strpos.xml: add the curly braces for the string interpolation

mmalferov:


kamil-tekiela:
Why are we doing this?

mmalferov:
It is not necessary to do this ;-)

There are two reasons for me.

1. It just seemed to me that it was a useful practice allways to enclose references in double-quoted strings in curly brackets to indicate the end of the variable name. Until the recent modification of the page that describes the string type, it contained a good example that the Internet still remembers:

```
<?php
$juice = "apple";

echo "He drank some $juice juice." . PHP_EOL;

// Unintended. "s" is a valid character for a variable name, so this refers to $juices, not $juice.
echo "He drank some juice made of $juices." . PHP_EOL;

// Explicitly specify the end of the variable name by enclosing the reference in braces.
echo "He drank some juice made of {$juice}s.";
```

Anyone who always writes variables in strings in curly brackets will never become a {$juice}s-victim.

2. Since as of PHP 8.2 using of `${var}` in strings is deprecated, examples with correct enclosing of refrences looking good for me. In other words, we remind readers once again of the syntax used to write variables wrapped in curly brackets in strings: `{$varname}` but not `${varname}` :-)

kamil-tekiela:
That example is still there https://www.php.net/manual/it/language.types.string.php#language.types.string.parsing

kamil-tekiela:
I don't know about this change. It seems to me like unnecessary churn. I'd leave it as it is now. We can encourage it in new additions.

mmalferov:
> That example is still there https://www.php.net/manual/it/language.types.string.php#language.types.string.parsing

You're right, the example is still in place, I was looking for it badly.

cmb69:
> That example is still there https://www.php.net/manual/it/language.types.string.php#language.types.string.parsing

It seems to me the Italian translation is not up to date. English looks quite different now: https://www.php.net/manual/en/language.types.string.php

kamil-tekiela:
Ohh you're right. I didn't notice it's Italian, because it's in English.

cmb69:
> I didn't notice it's Italian, because it's in English.

Oh, copying the original is the worst thing translators can do.

Girgias:
> > I didn't notice it's Italian, because it's in English.
> 
> Oh, copying the original is the worst thing translators can do.

If it is a partial translation with a WIP status then it is "fine" as revcheck will still track, but if none of the file is translated then it should be removed.

PR #3556, opened 2024-07-11
Create reference/gettext/functions/-.xml

MauricioFauth:
The `_()` function is an alias of the `gettext()` function.

Adding a dedicated manual page for this function will remove the need for the special cases created for it when generating the manual.

See:
- https://bugs.php.net/bug.php?id=63490
- https://github.com/php/systems/commit/b67fa163ba95deb68e3d51c4b4a3b6b74fafca71
- https://github.com/php/web-php/commit/fdec771e1f20751f1eb97d94b4cd4c1fb937632b
- https://github.com/php/web-php/commit/8abf4eee4028b14c50fe61578fd0085a1436f536
- https://github.com/php/php-src/blob/PHP-4.0/ext/gettext/gettext.c#L39

### DIFF
--- a/reference/gettext/functions/-.xml
+++ b/reference/gettext/functions/-.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.-" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>_</refname>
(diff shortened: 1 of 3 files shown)


MauricioFauth:
Update: I renamed the file from `_.xml` to `-.xml` and the refentry id from `function._` to `function.-`.

This follows the manual standard of replacing the `_` with `-` and also allows to remove the special cases from the `php/web-php` repository.

PR #3560, opened 2024-07-13
constants.xml: add dots at the end of sentences

mmalferov:


devnexen (approved):

Girgias:
This needs a rebase to fix the conflicts

cmb69:
I'm not quite sure about this. Few if any of these are proper sentences, and as such terminating with a full-stop wouldn't be quite right, although there is at least one occurrence where this is already done. I suggest to either convert the bullet items to full sentences, or to remove the full-stops altogether (maybe replacing them with semicolons).

cc @kamil-tekiela

PR #3562, opened 2024-07-14
eio/constants.xml: add dots at the ends of sentences…

mmalferov:
…and formatting nits


reference/eio/constants.xml
```diff
@@ -95,16 +95,16 @@
...
-     unspecified;
+      <function>eio_readdir</function> flag. If specified, the result argument of the callback
+      becomes an array with the following keys:
+      <literal>'names'</literal> - array of directory names
+      <literal>'dents'</literal> - array of <literal>struct eio_dirent</literal>-like
+      arrays having the following keys each:
+      <literal>'name'</literal> - the directory name;
+      <literal>'type'</literal> - one of <emphasis>EIO_DT_*</emphasis>
+      constants;
+      <literal>'inode'</literal> - the inode number, if available, otherwise
```
Girgias, line 106: Maybe this should be a list?

cmb69: Or maybe just refer to `eio_readdir()` here; on [that page there is the same information](https://www.php.net/manual/en/function.eio-readdir.php#refsect1-function.eio-readdir-returnvalues) (which likely should be formatted in a more readable way; a (nested) list might be a good idea).

jordikroon:
Given that this PR has been open and the suggested change goes beyond the scope of this. I don't see a reason to keep block this much longer.

If anyone wants to pick up the remaining comments, they are free to do so. 

 > Or maybe just refer to eio_readdir() here; on [that page there is the same information](https://www.php.net/manual/en/function.eio-readdir.php#refsect1-function.eio-readdir-returnvalues) (which likely should be formatted in a more readable way; a (nested) list might be a good idea).

PR #3568, opened 2024-07-15
Document types for apcu_exists()

BenMorel:


Girgias (approved):
I don't know APCU, so trusting this is correct.

PR #3570, opened 2024-07-15
trigger-error.xml: add the `literal` tag for the E_USER_*

mmalferov:


reference/errorfunc/functions/trigger-error.xml
```diff
@@ -43,7 +43,7 @@
      <term><parameter>error_level</parameter></term>
      <listitem>
       <para>
-       The designated error type for this error. It only works with the E_USER
+       The designated error type for this error. It only works with the <literal>E_USER_*</literal>
```
cmb69, line 46: Not sure what the current stance on marking up such constants is, but maybe better
```suggestion
       The designated error type for this error. It only works with the <constant><replaceable>E_USER_*</replaceable></constant>
```

That *may* break PhD rendering, though.

salathe: We already use similar to the following in several places, which seems preferable.

```suggestion
       The designated error type for this error. It only works with the <constant>E_USER_<replaceable>*</replaceable></constant>
```

Girgias (approved):

PR #3571, opened 2024-07-15
Use "unique" less in describing `uniqid()`

jimwins:
The identifier generated is time-based either in its entirety (with default options) or with some random data (with `more_entropy` enabled) so avoid describing it as unique without immediate caveats.

It's not much more than a fancy timestamp when called without any parameters, so let's not make "unique" such a central part of its description.

### DIFF
--- a/reference/misc/functions/uniqid.xml
+++ b/reference/misc/functions/uniqid.xml
@@ -14,8 +14,9 @@
    <methodparam choice="opt"><type>bool</type><parameter>more_entropy</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Gets an identifier based on the current time in microseconds with an
-   optional prefix and/or a random value appended.
+   Gets an identifier based on the current time with microsecond precision,
+   prefixed with the given <parameter>prefix</parameter> and optionally
+   appending a randomly generated value.
   </para>
   &caution.cryptographically-insecure;
   <warning>
@@ -38,7 +39,9 @@
       <para>
        Can be useful, for instance, if you generate identifiers
        simultaneously on several hosts that could generate the
-       same identifier at the same microsecond.
+       same identifier at the same microsecond. (This can happen
+       even on a single host if the system clock is moved backwards,
+       such as by an NTP adjustment.)
       </para>
       <para>
        With an empty <parameter>prefix</parameter>, the returned string will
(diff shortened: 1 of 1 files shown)


reference/misc/functions/uniqid.xml
```diff
@@ -14,18 +14,16 @@
    <methodparam choice="opt"><type>bool</type><parameter>more_entropy</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Gets a prefixed unique identifier based on the current time in
-   microseconds.
+   Gets an identifier based on the current time in microseconds with an
+   optional prefix and/or a random value appended.
```
TimWolla, line 18: This might be not being a native speaker, but it reads as if the prefix would be appended. Perhaps:

```
Gets an identifier based on the current time with microsecond precision, prefixed with the given <parameter>prefix</parameter> and optionally appending a randomly generated value.
```

reference/misc/functions/uniqid.xml
```diff
@@ -14,18 +14,16 @@
...
-     This function does not guarantee uniqueness of return
-     value. Since most systems adjust system clock by NTP or like,
-     system time is changed constantly. Therefore, it is possible that
-     this function does not return unique ID for the
-     process/thread. Use <parameter>more_entropy</parameter> to
-     increase likelihood of uniqueness.
+     This function does not guarantee the uniqueness of the return
+     value because the value is based on the current time in microseconds
+     or the current time with a small amount of random data appended
+     if <parameter>more_entropy</parameter> is &true;.
```
TimWolla, line 26: I found the bit that the time does not move monotonically useful in the original version (i.e. NTP may move the clock backwards).

TimWolla (commented):
Thank you, this is a great change. I have some minor remarks, but overall this is looking good.

reference/misc/functions/uniqid.xml
jimwins: Yes, much better.

reference/misc/functions/uniqid.xml
jimwins: Good idea, since that can cause collisions even on a single host.

TimWolla (approved):

Girgias (approved):

PR #3574, opened 2024-07-17
First class callable syntax

divinity76:
modernize the example a bit


Girgias:
The documentation covers PHP 7 and PHP 8.

Therefore we cannot accept this change for the moment.

PR #3575, opened 2024-07-17
Fix pow() return value section

staabm:
I don't think `pow()` can return an `object`?

### DIFF
--- a/reference/math/functions/pow.xml
+++ b/reference/math/functions/pow.xml
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
    <methodsynopsis>
-   <type class="union"><type>int</type><type>float</type></type><methodname>pow</methodname>
+   <type class="union"><type>int</type><type>float</type><type>object</type></type><methodname>pow</methodname>
    <methodparam><type>mixed</type><parameter>num</parameter></methodparam>
    <methodparam><type>mixed</type><parameter>exponent</parameter></methodparam>
   </methodsynopsis>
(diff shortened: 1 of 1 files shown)


devnexen:
Seems you re correct, i do not see any other possibility than float/int.

devnexen (approved):

cmb69:
Well, https://github.com/php/php-src/blob/d55ef3f339a6d8206ee293d5cc7d9ab6044d3e6a/ext/standard/basic_functions.stub.php#L3212C3-L3212C63 says it can; if that's wrong, the stubs would need to be fixed, and later synced with the docs. If it's not wrong, there should be some clarification in the docs (I think it's about internal objects).

devnexen:
I admit I did not look the stub but just the implementation(s).

devnexen:
Finally I retract, not only GMP overrides but also libraries php decimal do override the do_operation handler.

staabm:
added a paragraph and a example

cmb69 (approved):
Thank you, @staabm! This looks good to me.

reference/math/functions/pow.xml
```diff
@@ -54,7 +54,7 @@
    otherwise it will be returned as a <type>float</type>.
   </para>
   <para>
-   PHP-Extensions may override the behaviour of this operation and make it return objects.
+   PHP-Extensions may override the behaviour of this operation and make it return a object.
```
devnexen, line 57: nit: `an object` I think

devnexen:
Thanks !

Issue #3358, opened 2024-04-29
Add history for PHP 7 and 8


mbomb007:
From manual page: https://php.net/history.php

---

The list of updates/changes to PHP on the history page stops at PHP 5. The page should be updated to add information for the more recent versions.

PR #3576, opened 2024-07-18
Add history for PHP 6, 7, and 8 (#3358)

jimwins:
Could be fleshed out more, but I think this hits the high points.


cmb69 (approved):
Thank you for filling in the missing bits and pieces! And sure, more can be added, but this appears to be a good first step to present to readers of the manual.

sy-records (approved):

PR #3584, opened 2024-07-19
install/windows/index.xml: add the `option` tag

mmalferov:


install/windows/index.xml
```diff
@@ -57,7 +57,8 @@
   </para>
 
   <para>
-   The CRT installer supports the /quiet and /norestart command-line switches, so you can script running it.
+   The CRT installer supports the <command>/quiet</command> and <command>/norestart</command>
```
cmb69, line 60: I think that should better be:
```suggestion
   The CRT installer supports the <option>/quiet</option> and <option>/norestart</option>
```

However, I'm not even sure whether these are still supported by the CRT installers.

mmalferov: These are still supported ;-)

https://learn.microsoft.com/en-us/cpp/windows/redistributing-visual-cpp-files?view=msvc-170#command-line-options-for-the-redistributable-packages

I changed the tags from `command` to 'options'

cmb69: Thank you for checking this!

cmb69:
Thank you!

PR #3586, opened 2024-07-20
Move curl-setopt constants to their own variablelist

haszi:
This PR copies all the `CURLOPT_*` constants from the `curl-setopt()` page to a dedicated `<variablelist>` on the cURL constants page. The changes in the first commit were made using [this script](https://gist.github.com/haszi/fc7fa4e19975b2884fc78cbe8c3303a9) and the second (DocBook compliant markup) commit was made by going through the individual issues one by one in order for the docs to compile.

The constant descriptions removed in third commit (and probably a lot more) will be added in a follow-up PR once all the `CURL*` constants are documented.

There is more  that could have been updated in this new file (e.g. adding tags, referencing groups of constants (e.g. `CURLSSLOPT_*`) instead of listing them individually, etc.) but this PR has lots of changes as it is. I will probably open follow-up PRs once all the `CURL*` constants are documented.


Girgias (approved):
Thank you for the work!

I see a couple of existing issues, but those should be addressed later.

Issue #3384, opened 2024-05-14
PHP_OUTPUT_HANDLER_PROCESSED is only available as of PHP 8.4


adazmy:
### Description

While listing php constants, PHP_OUTPUT_HANDLER_PROCESSED is not defined.

Version: 
PHP 8.3.6 (cli) (built: Apr 11 2024 20:23:19) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.3.6, Copyright (c) Zend Technologies
    with Zend OPcache v8.3.6, Copyright (c), by Zend Technologies

The following code:

```php
<?php

echo PHP_OUTPUT_HANDLER_PROCESSED;

```

Resulted in this output:
```
Fatal error: Uncaught Error: Undefined constant "PHP_OUTPUT_HANDLER_PROCESSED" in /in/fa1o8:3
Stack trace:
#0 {main}
  thrown in /in/fa1o8 on line 3

Process exited with code 255.
```

But I expected this output instead:
```
// Constant value
```

### PHP Version

Output for 8.1.0 - 8.1.28, 8.2.0 - 8.2.19, 8.3.0 - 8.3.7

### Operating System

Ubuntu 20.04

Girgias:
Correct, this is a doc-bug.

PR #3587, opened 2024-07-20
Fix GH-3384

haszi:
Add a sentence from which PHP version PHP_OUTPUT_HANDLER_PROCESSED is available.

Closes GH-3384

### DIFF
--- a/reference/outcontrol/constants.xml
+++ b/reference/outcontrol/constants.xml
@@ -198,6 +198,7 @@
    <listitem>
     <para>
      Indicates that the output handler successfully processed the buffer.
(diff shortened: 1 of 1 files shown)


devnexen (approved):

cmb69:
If it is only available as of PHP 8.4.0, it probably should not have been documented yet, but to save work, maybe it's good to merge this PR. @Girgias, what do you think?

Girgias (approved):

Girgias:
I think we should start to merge PHP 8.4 stuff as alpha versions are out, yes it is not set in stone, but a bunch of stuff is.

PR #3594, opened 2024-07-22
chr.xml: use more relevant a `method` instead a `function`

mmalferov:


cmb69 (approved):
Thank you!

Girgias (approved):

reference/intl/intlchar/chr.xml
```diff
@@ -16,7 +16,7 @@
    Returns a string containing the character specified by the Unicode code point value.
   </para>
   <para>
-   This function complements <function>IntlChar::ord</function>.
+   This method complements <function>IntlChar::ord</function>.
```
Girgias, line 19: Maybe at the same time?
```suggestion
   This method complements <methodname>IntlChar::ord</methodname>.
```

mmalferov: It so happens that the elephant goes unnoticed )

PR #3595, opened 2024-07-23
digit.xml: add tags, amend example

mmalferov:
In the description of the method,

> In this case the value is c-'A'+10 for latin lettera A-Z

and

> In this case the value is ch-'a'+10

— it looks like this text was copied aimlessly from the comment of the ICU source code file:

[source/common/unicode/uchar.h](https://github.com/unicode-org/icu/blob/40b2ec3c3727bca975824fdb9d1f084207d535ff/icu4c/source/common/unicode/uchar.h#L4252)

And these sentences, IMHO, do not make sense in the context of the PHP method description, but only `u_digit` definition:

[source/common/uchar.cpp](https://github.com/unicode-org/icu/blob/40b2ec3c3727bca975824fdb9d1f084207d535ff/icu4c/source/common/uchar.cpp#L452)

### DIFF
--- a/reference/intl/intlchar/digit.xml
+++ b/reference/intl/intlchar/digit.xml
@@ -28,10 +28,14 @@
      In this case the value is the character's decimal digit value.
     </member>
     <member>
-     The character is one of the uppercase Latin letters <literal>'A'</literal> through <literal>'Z'</literal>.
+     The character is one of the uppercase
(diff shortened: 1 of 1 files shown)


lacatoire (approved):

jordikroon:
Thanks!

PR #3600, opened 2024-07-24
Added example for void return type in documentation

TImoteoVieira:
This added code improves the documentation by providing a practical example of how to use PHP's void.

### DIFF
--- a/language/types/void.xml
+++ b/language/types/void.xml
@@ -17,6 +17,37 @@
   </simpara>
  </note>
 
+ <sect2 xml:id="language.types.void.example">
+  <title>Example: Using void Return Type</title>
+
+  <para>
+   Here is an example of a function that sends an email. The function has a
+   <type>void</type> return type because it does not return any value.
+  </para>
+
+  <informalexample>
+   <programlisting role="php">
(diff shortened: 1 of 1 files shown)


language/types/void.xml
```diff
@@ -17,6 +17,37 @@
   </simpara>
  </note>
 
+ <sect2 xml:id="language.types.void.example">
+  <title>Example: Using void Return Type</title>
+
+  <para>
+   Here is an example of a function that sends an email. The function has a
+   <type>void</type> return type because it does not return any value.
```
Girgias, line 25: I'm not sure about this phrasing.

In PHP all functions (yes even  `void` functions) return a value, which is mentioned above. So this phrasing isn't accurate.

Girgias (changes requested):
I am not sure if there is much point in having an example on when to use a type.

Most of the type system pages are there to be succinct and explain what the type is/means, not provide examples (string and array are old and not really up to what I wish they were)

kamil-tekiela:
I agree with Girgias. I'm closing this PR, but if you have another idea on how to improve this page, please feel free to open a new PR.

PR #3602, opened 2024-07-24
fordigit.xml: amend return type

mmalferov:


reference/intl/intlchar/fordigit.xml
```diff
@@ -54,7 +54,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   The character representation (as a <type>string</type>) of the specified digit in the specified radix.
+   The character representation (as a <type>int</type>) of the specified digit in the specified radix.
```
cmb69, line 57: This may use `&integer;` instead of the more verbose `<type>int</type>`, but I'm not sure about the current best practice.

Girgias: I don't think there is a preference? It feels to me that when the entity is used the "type" tag is not that important, as some languages will edit it (which is slightly cursed)

cmb69: Thanks! Then this should be fine as is.

PR #3346, opened 2024-04-25
Update minimum versions of development tools

petk:
This updates the development tools versions required to configure and build PHP across various versions:

* autoconf:
    * PHP 7.3 and later: 2.68+
    * PHP 7.2: 2.64+
    * From PHP 5.4 to 7.1: 2.59+
* re2c:
    * From PHP 8.3 and later 1.0.3+
    * From PHP 5.2 to 8.2: 0.13.4+
    * PHP 5.1 required 0.9.11+ (not noting in the docs due to obsoletion)
* bison:
    * From PHP 7.4 and later: 3.0.0+
    * From PHP 7.0 to 7.3: 2.4+ (including Bison 3.x)
    * From PHP 5.5 to 5.6: 2.4 to 2.7
    * PHP 5.4: 1.28, 1.35, 1.75, 2.0 to 2.6.4

PHP versions are sorted in reverse order from newest to oldest PHP.

This is also a sync with https://github.com/php/web-php/pull/983 because these are listed on two places. In this docs and in here https://www.php.net/git.php

Maybe we could simplify the https://php.net/git.php and link this documentation page there.

Also, from what I see is that PHP 5.x is somehow trying to become removed from the manual. If that's so then maybe these coulb be noted like this `PHP 7.1 and older`  :thinking:

PR #3603, opened 2024-07-24
Update Unix requirements section

cmb69:
C99 is required as of https://github.com/php/php-src/commit/b51a99ae358b3295de272bb3c3edb6913eeb0fd4.

There is likely more to update on [that page](https://www.php.net/manual/en/install.unix.php). If anybody happens to know which versions exactly, please tell me. Otherwise I'll check that out.

PS: The info about bison and r2c versions is apparently up-to-date.

### DIFF
--- a/install/unix/index.xml
+++ b/install/unix/index.xml
@@ -38,7 +38,7 @@
      </listitem> 
      <listitem>
       <simpara>
-       An ANSI C compiler
+       An ANSI C compiler (as of PHP 8.0.0, C99 compatibility is assumed)
       </simpara>
      </listitem> 
      <listitem>


Girgias (approved):
We have updated the re2c and bison requirements a few times

cmb69:
I was wondering about autoconf, automake and libtool requirements, but couldn't find any version in the sources so far. Not even sure if these are needed; I think at least autoconf is not necessary when building a proper release.

And we should note that re2c and bison are only required to rebuilt the parsers, what is usually not what users want to do. Maybe remove that part altogether and link to some developer docs (maybe just https://github.com/php/php-src?tab=readme-ov-file#building-php-source-code).

I'll have a closer look at this as soon as possible.

cmb69:
> I was wondering about autoconf, automake and libtool requirements, but couldn't find any version in the sources so far. Not even sure if these are needed; I think at least autoconf is not necessary when building a proper release.

@petk, do you happen to know about autoconf, automake and libtool version requirements, or where they can be looked up the source code?

petk:
> > I was wondering about autoconf, automake and libtool requirements, but couldn't find any version in the sources so far. Not even sure if these are needed; I think at least autoconf is not necessary when building a proper release.
> 
> @petk, do you happen to know about autoconf, automake and libtool version requirements, or where they can be looked up the source code?

Autoconf, re2c and Bison were already rechecked and updated via #3346. They are ok. Minimum versions are defined in the configure.ac. And that Automake and Libtool should be removed as they are not required. But I'll re-check why are they even listed there. The bundled Libtool unfortunately couldn't be updated to the latest version for PHP-8.4 yet (#13476). Will be done in the next version.

petk:
The Libtool and Automake can be removed from this page. Automake was once needed in PHP and some parts needed the aclocal tool which is part of the Automake package. The Libtool was also used a bit inconsistently and there were glitches with versions and the bundled libtool (probably) so therefore it was needed.

https://www.php.net/git.php#buildconf_fail

About updating this bundled Libtool, I would keep the bundled way only. Because there is also an option to use the system Libtool (latest version) and some PHP installations do that already. But it is much easier to have a shorter list of requirements in the docs. 

Also this php.net/git.php page can be probably updated at the bottom (we've left this out back then a bit). But I'll still keep these historical notes on my own somewhere to understand the changes across the versions. https://github.com/petk/php-build-system/blob/master/docs/evolution.md :D

PR #3604, opened 2024-07-24
json: added json_last_error_msg to see-also

staabm:
I think `json_last_error_msg()` is more helpful then the older `json_last_error()` variant, therefore adding it to the see-also list


cmb69 (approved):
Makes sense to me. Thank you!

Girgias (approved):

PR #3605, opened 2024-07-24
getbidipairedbracket.xml: add missing spaces

mmalferov:


reference/intl/intlchar/getbidipairedbracket.xml
```diff
@@ -16,7 +16,7 @@
    Maps the specified character to its paired bracket character.
   </para>
   <para>
-   For <literal>Bidi_Paired_Bracket_Type!=None</literal>, this is the same as <function>IntlChar::charMirror</function>.
+   For <literal>Bidi_Paired_Bracket_Type != None</literal>, this is the same as <function>IntlChar::charMirror</function>.
```
cmb69, line 19: I guess that rather should be something like
```suggestion
   For <code>IntlChar::PROPERTY_BIDI_PAIRED_BRACKET_TYPE !== IntlChar::BPT_NONE</code>, this is the same as <function>IntlChar::charMirror</function>.
```

Although I've never just that part of the code base, and the existing documentation seems to be poor.

devnexen (approved):
sorry did not see it sooner !

cmb69:
Thank you!

PR #3608, opened 2024-07-24
phpversion.xml: move replaceable part of constant name

mmalferov:
Since the versioning constants names is `PHP_*_VERSION` rather than `PHP_VERSION_*`:

* PHP_MAJOR_VERSION
* PHP_MINOR_VERSION
* PHP_RELEASE_VERSION

### DIFF
--- a/reference/info/functions/phpversion.xml
+++ b/reference/info/functions/phpversion.xml
@@ -109,7 +109,7 @@ if (!defined('PHP_VERSION_ID')) {
 // may have, this doesn't require to use version_compare() everytime 
 // you check if the current PHP version may not support a feature.
 //
-// For example, we may here define the PHP_VERSION_* constants thats 
+// For example, we may here define the PHP_*_VERSION constants thats 
 // not available in versions prior to 5.2.7
 
 if (PHP_VERSION_ID < 50207) {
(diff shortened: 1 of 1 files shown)


reference/info/functions/phpversion.xml
```diff
@@ -91,26 +93,29 @@ echo phpversion('tidy');
...
 <?php
-// PHP_VERSION_ID is available as of PHP 5.2.7, if our 
+
+// PHP_VERSION_ID is available as of PHP 5.2.7, if our
 // version is lower than that, then emulate it
 if (!defined('PHP_VERSION_ID')) {
     $version = explode('.', PHP_VERSION);
 
     define('PHP_VERSION_ID', ($version[0] * 10000 + $version[1] * 100 + $version[2]));
 }
```
Girgias, line 103: This can be removed, we assume PHP 7.0 in the docs.

reference/info/functions/phpversion.xml
```diff
@@ -75,11 +75,13 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-// prints e.g. 'Current PHP version: 4.1.1'
+
+// Prints e.g. 'Current PHP version: 4.1.1'
```
Girgias, line 79: Maybe we could update to something more modern than PHP 4? :)

Girgias (changes requested):
Thank you, but while at it we might modernize the example

Girgias (approved):

PR #3610, opened 2024-07-24
Extensions/Membership: fix placement HRTime

jrfnl:
Updates: https://www.php.net/manual/en/extensions.membership.php

Please correct me if I'm wrong, but AFAIK, HRTime is a PECL extension and not a bundled extension.

Refs:
https://www.php.net/manual/en/hrtime.installation.php

### DIFF
--- a/appendices/extensions.xml
+++ b/appendices/extensions.xml
@@ -246,7 +246,6 @@
     <listitem><para><xref linkend="book.fileinfo"/></para></listitem>
     <listitem><para><xref linkend="book.filter"/></para></listitem>
     <listitem><para><xref linkend="book.ftp"/></para></listitem>
-    <listitem><para><xref linkend="book.hrtime"/></para></listitem>
     <listitem><para><xref linkend="book.iconv"/></para></listitem>
     <listitem><para><xref linkend="book.image"/></para></listitem>
     <listitem><para><xref linkend="book.intl"/></para></listitem>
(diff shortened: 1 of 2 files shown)


jrfnl:
Looks like the build is failing as the QA script does not handle the URL for the MongoDB extension correctly ?

jrfnl:
Let me split this PR into two so the two changes do not have to block each other.

jrfnl:
I've moved the second commit to PR #3611 and updated the PR title and description for this PR after the split.

appendices/extensions.xml
```diff
@@ -330,6 +329,7 @@
     <listitem><para><xref linkend="book.geoip"/></para></listitem>
     <listitem><para><xref linkend="book.gmagick"/></para></listitem>
     <listitem><para><xref linkend="book.gnupg"/></para></listitem>
+    <listitem><para><xref linkend="book.hrtime"/></para></listitem>
```
Girgias, line 332: Was this regenerated via the `doc-base/scripts/qa/extensions.xml.php` script?

Girgias (commented):
This is confusing as we have an `hrtime` function >-< but yes the extension is in PECL AFAIK

appendices/extensions.xml
jrfnl: No, I updated the file manually, but AFAICS, the QA check runs that script anyhow and has validated the change as correct. Or am I misinterpreting the action ?
https://github.com/php/doc-en/blob/1422d543bf028003a543fd1ce920c6a301a93a51/.github/workflows/integrate.yaml#L47

Girgias: The point is to use the script to update the file, which you can do by executing the file without ``--check`` but it must be fine if the QA script validates it

jrfnl: @Girgias Good to know for the next time. Bit awkward though with the script being in another repo than the one you're contributing to...  (a repo I haven't got locally, so I apparently never needed it before)

jrfnl: Good to know. I've never needed it so far (40+ commits in).

PR #3611, opened 2024-07-25
Extensions/Membership: add MongoDB

jrfnl:
Updates:
* https://www.php.net/manual/en/extensions.alphabetical.php
* https://www.php.net/manual/en/extensions.membership.php

The MongoDB extension appears to be missing from the extension pages.

The build for this will fail due to the extension not having a `book.xml` page, but a `set.xml` page instead and the QA script not handling this. Should I rename the XML file ? Or submit an update for the QA script ?

Ref:
* https://www.php.net/manual/en/set.mongodb.php

### DIFF
--- a/appendices/extensions.xml
+++ b/appendices/extensions.xml
@@ -82,6 +82,7 @@
    <listitem><simpara><xref linkend="book.memcached"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mhash"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.misc"/></simpara></listitem>
+   <listitem><simpara><xref linkend="set.mongodb"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mqseries"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mysql"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mysql-xdevapi"/></simpara></listitem>
@@ -342,6 +343,7 @@
     <listitem><para><xref linkend="book.mcrypt"/></para></listitem>
     <listitem><para><xref linkend="book.memcache"/></para></listitem>
     <listitem><para><xref linkend="book.memcached"/></para></listitem>
+    <listitem><para><xref linkend="set.mongodb"/></para></listitem>
     <listitem><para><xref linkend="book.mqseries"/></para></listitem>
     <listitem><para><xref linkend="book.mysql"/></para></listitem>
     <listitem><para><xref linkend="book.mysql-xdevapi"/></para></listitem>


cmb69:
Maybe @alcaeus or @jmikola can have a look at this?

alcaeus:
I honestly don't know the difference between a book and a set, but if it's a simple case of renaming the XML file I think it's better to rename the file than to adapt the QA script.

jrfnl:
> I honestly don't know the difference between a book and a set

Maybe @bjori or @derickr can clarify ? They look to have been involved in the introduction of this extension handbook.

cmb69:
Maybe @Girgias and @haszi are more into that nowadays.

haszi:
> I honestly don't know the difference between a book and a set

A book is a "complete book" and a set is a collection of books and is the top of the DocBook structural hierarchy (according to the official docs for [book](https://tdg.docbook.org/tdg/5.2/book.html) and [set](https://tdg.docbook.org/tdg/5.2/set)).
In practical terms, I think books are more useful for extension references as they can have multiple types of children elements (e.g appendices, chapters, references) while sets can only have articles, books and sets.

> The build for this will fail due to the extension not having a `book.xml` page, but a `set.xml` page instead and the QA script not handling this. Should I rename the XML file ? 

To add MongoDB to the list of extensions, changing `set.xml` to `book.xml` won't be enough unfortunately. The top `<set>` element needs to be converted to a `<book>`, its children `<book>` elements to `<chapter>`s, `<article>`s and `<procedure>`s to `<section>`s and probably some more deeper down the hierarchy. Also, at least some of the `xml:id`s need to be updated too.

I just checked and other than MongoDB, only MySQL is using a `<set>` instead of a `<book>` element as its top level element.

> Should I rename the XML file ? Or submit an update for the QA script ?

My preference would be to update these two extensions' sets to books. If nobody has any objections, I can take a look at this this weekend.

What do you think @Girgias?

php-pulls:
It was done because it was documenting two extensions, bson, and mongodb. This shouldn't be needed now, as these have been merged into one extension ages ago.

php-pulls:
I think you should just leave fixing this to the mongodb folk, they keep these docs up to date.

cmb69:
Hey @php-pulls, you're still showing up as that, and not as @derickr. :)

derickr:
I think I fixed that now with this email?

Girgias:
I am in favour of moving away from `<set>` and use `<book>`.

alcaeus:
> I think you should just leave fixing this to the mongodb folk, they keep these docs up to date.

That will take a while, as we currently have no bandwidth for larger changes to docs. If somebody else can and wants to handle this, it would be greatly appreciated!

haszi:
I've opened a PR in `doc-base` (https://github.com/php/doc-base/pull/138) and `doc-en` (https://github.com/php/doc-en/pull/3627) to address this. 

Please note that the changes made in the `doc-en` PR will have to be made in all translations too or they won't build once the `doc-base` PR is merged. 

If needed, I can help with opening a PR in some of the translations. Since all of these changes are to the markup and to the `xml:id`s, this should not be a problem (hopefully).

haszi:
@jrfnl  The `doc-base` and documentation repository PRs have been merged.

jrfnl:
Closing as fixed via the transformation of the MongoDB docs from a `set` to a `book`.

PR #3613, opened 2024-07-25
Update index.xml

sikanderiqbal:
direct link to Visual C Redistributable files

### DIFF
--- a/install/windows/index.xml
+++ b/install/windows/index.xml
@@ -43,9 +43,9 @@
   </para>
 
   <para>
-   The Microsoft Visual C++ Redistributable for Visual Studio 2019 is suitable
+   The Microsoft Visual C++ Redistributable for Visual Studio 2022 is suitable
    for all these PHP versions, see
-   <link xlink:href="https://visualstudio.microsoft.com/downloads/">https://visualstudio.microsoft.com/downloads/</link>.
+   <link xlink:href="https://visualstudio.microsoft.com/downloads/#microsoft-visual-c-redistributable-for-visual-studio-2022">https://visualstudio.microsoft.com/downloads/</link>.
   </para>
 
   <para>
--- a/install/windows/index.xml
+++ b/install/windows/index.xml
@@ -45,7 +45,7 @@
   <para>
    The Microsoft Visual C++ Redistributable for Visual Studio 2022 is suitable
    for all these PHP versions, see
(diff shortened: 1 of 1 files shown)


install/windows/index.xml
```diff
@@ -43,9 +43,9 @@
   </para>
 
   <para>
-   The Microsoft Visual C++ Redistributable for Visual Studio 2019 is suitable
+   The Microsoft Visual C++ Redistributable for Visual Studio 2022 is suitable
```
cmb69, line 46: This makes certainly sense.

install/windows/index.xml
```diff
@@ -43,9 +43,9 @@
...
    for all these PHP versions, see
-   <link xlink:href="https://visualstudio.microsoft.com/downloads/">https://visualstudio.microsoft.com/downloads/</link>.
+   <link xlink:href="https://visualstudio.microsoft.com/downloads/#microsoft-visual-c-redistributable-for-visual-studio-2022">https://visualstudio.microsoft.com/downloads/</link>.
```
cmb69, line 48: I'm not quite sure about. Either link redirects me (German user) to https://visualstudio.microsoft.com/de/downloads/, so the additional URI fragment likely doesn't make a difference for many users. But I'm not against adding it.

In a perfect world, the actual URI should be added to https://github.com/php/doc-base/blob/master/entities/global.ent, and we would use only the respective entity here. If you want to do this, please go ahead; otherwise this change would be fine as well.

cmb69 (commented):
Thank you!

install/windows/index.xml
sikanderiqbal: @cmb69 Ah, the redirect to /de doesn't retain the URL fragment. What a shame and so unfortunate that websites use ``#`` to alter page behaviour. Remember twitter's ``#!`` days?

I can certainly add an entry to ``global.ent`` but since that exists in a different repository so we would have to wait for a PR merge there before this one can be updated? If you prefer that route, please let me know and I'll do the adjustment.

cmb69: > Remember twitter's #! days?

Well, I remember Twitter's days. ;)

> I can certainly add an entry to global.ent but since that exists in a different repository so we would have to wait for a PR merge there before this one can be updated?

That shouldn't be a problem. So please go ahead (this way translators don't have anything to do if the URI changes again).

PR #3614, opened 2024-07-26
hash: Clean up the listed algorithms for the `hash_*()` algorithm parameter

TimWolla:


reference/hash/functions/hash-file.xml
```diff
@@ -25,8 +25,7 @@
      <term><parameter>algo</parameter></term>
      <listitem>
       <para>
-       Name of selected hashing algorithm (i.e. <literal>md5</literal>, <literal>sha256</literal>,
-       <literal>haval160,4</literal>, etc..).
+       Name of selected hashing algorithm (e.g. <literal>sha256</literal>).
```
cmb69, line 28: I'm not exactly whether this should be
```suggestion
       Name of selected hashing algorithm (e.g. <literal>"sha256"</literal>).
```

reference/hash/functions/hash-hkdf.xml
```diff
@@ -25,8 +25,8 @@
      <term><parameter>algo</parameter></term>
      <listitem>
       <para>
-       Name of selected hashing algorithm (i.e. "sha256", "sha512", "haval160,4", etc..)
-       See <function>hash_algos</function> for a list of supported algorithms.
+       Name of selected hashing algorithm (e.g. <literal>sha256</literal>).
+       For a list of supported algorithms see <function>hash_algos</function>.
```
cmb69, line 29: Given that they repeat a couple of times, it might be a good idea to define an entity in https://github.com/php/doc-en/blob/master/language-snippets.ent.

cmb69 (commented):
Thank you!

reference/hash/functions/hash-hkdf.xml
Girgias: Having an entity or XIncluding the parameter from a "main" copy is also an option, but I agree that would be better than manually having it copied everywhere

jimwins: Should this reference hash_hmac_algos instead of hash_algos?

reference/hash/functions/hash-hmac-file.xml
```diff
@@ -24,7 +24,8 @@
      <term><parameter>algo</parameter></term>
      <listitem>
       <para>
-       Name of selected hashing algorithm (i.e. "md5", "sha256", "haval160,4", etc..) See <function>hash_hmac_algos</function> for a list of supported algorithms.
+       Name of selected hashing algorithm (e.g. <literal>sha256</literal>).
+       For a list of supported algorithms see <function>hash_algos</function>.
```
jimwins, line 28: I think this should also be hash_hmac_algos.

reference/hash/functions/hash-hmac.xml
```diff
@@ -24,7 +24,8 @@
      <term><parameter>algo</parameter></term>
      <listitem>
       <para>
-       Name of selected hashing algorithm (i.e. "md5", "sha256", "haval160,4", etc..) See <function>hash_hmac_algos</function> for a list of supported algorithms.
+       Name of selected hashing algorithm (e.g. <literal>sha256</literal>).
+       For a list of supported algorithms see <function>hash_algos</function>.
```
jimwins, line 28: Another one that should be hash_hmac_algos.

reference/hash/functions/hash-pbkdf2.xml
```diff
@@ -28,9 +28,8 @@
      <term><parameter>algo</parameter></term>
      <listitem>
       <para>
-       Name of selected hashing algorithm (i.e. <literal>md5</literal>,
-       <literal>sha256</literal>, <literal>haval160,4</literal>, etc..) See
-       <function>hash_algos</function> for a list of supported algorithms.
+       Name of selected hashing algorithm (e.g. <literal>sha256</literal>).
+       For a list of supported algorithms see <function>hash_algos</function>.
```
jimwins, line 32: Another one that should be hash_hmac_algos.

reference/hash/functions/hash-hkdf.xml
cmb69: > Should this reference hash_hmac_algos instead of hash_algos?

Indeed.

reference/hash/functions/hash-file.xml
TimWolla: Changed.

reference/hash/functions/hash-hkdf.xml
TimWolla: > Should this reference hash_hmac_algos instead of hash_algos?

Fixed for all of them, including a proper note for `hash_init()`. Given that there are three different variants now, I've opted to not make this an entity.

jimwins:
LGTM!

jimwins (approved):

cmb69 (approved):
Thank you!

PR #3618, opened 2024-07-27
Remove documentation/example of HTTP Digest Authorization

jimwins:
It was declared obsolete as a standard in 2011. Some of the supporting code is still technically part of the SAPI layer, but there's no reason to document it here.


cmb69 (approved):
Yeah, I think this info can be removed from the manual. Thank you!

(reading a bit further down gives me a blast from the past: "Internet Explorer", "Lynx" and even "Netscape Navigator" are mentioned there!)

PR #3619, opened 2024-07-27
Update examples using md5(...) to use hash('md5', ...)

jimwins:
`hash()` has existed since 5.1.2, should be fine to update these old examples now.

### DIFF
--- a/reference/mcrypt/functions/mcrypt-module-open.xml
+++ b/reference/mcrypt/functions/mcrypt-module-open.xml
@@ -117,8 +117,8 @@
     $iv = mcrypt_create_iv(mcrypt_enc_get_iv_size($td), MCRYPT_DEV_RANDOM);
     $ks = mcrypt_enc_get_key_size($td);
 
-    /* Create key */
-    $key = substr(md5('very secret key'), 0, $ks);
+    /* Create key (example only: MD5 is not a good hash algorithm for this) */
+    $key = substr(hash('md5', 'very secret key'), 0, $ks);
 
     /* Intialize encryption */
     mcrypt_generic_init($td, $key, $iv);
--- a/reference/pdo_sqlite/PDO/sqliteCreateFunction.xml
+++ b/reference/pdo_sqlite/PDO/sqliteCreateFunction.xml
@@ -146,7 +146,7 @@
 <?php
 function md5_and_reverse($string) 
 {
-    return strrev(md5($string));
+    return strrev(hash('md5', $string));
 }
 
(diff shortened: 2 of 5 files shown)


reference/mcrypt/functions/mcrypt-module-open.xml
```diff
@@ -117,8 +117,8 @@
     $iv = mcrypt_create_iv(mcrypt_enc_get_iv_size($td), MCRYPT_DEV_RANDOM);
     $ks = mcrypt_enc_get_key_size($td);
 
-    /* Create key */
-    $key = substr(md5('very secret key'), 0, $ks);
+    /* Create key (example only: MD5 is not a good hash algorithm for this) */
+    $key = substr(hash('md5', 'very secret key'), 0, $ks);
```
cmb69, line 121: It's probably a good idea to use another hash function (instead of mentioning that it is not a good algo in this case).

Maybe also for the radius examples below.

/cc @TimWolla

jimwins: Sure, I just didn't want to get into the weeds on what it should be replaced with.

The Radius examples need to be MD5 because that's part of the underlying protocol.

Why anyone thought that leaving that detail in the PHP part of using the extension is a head-scratcher, but not a problem we need to solve. I think in a more perfect world there would be something like a `radius_put_chap_password()` function.

cmb69: Well, maybe there's no real issue here, since neither extension looks really maintained, and the documentation might be removed from the manual in a couple of years.

I just though that we should try to be examplary when replacing `md5()` with `hash()`, and should really consider the usage of the hashing algorithm. I've never worked with mcrypt, so I don't really know which algo to use. The `mcrypt_generic_init()` documentation states about `$key`:

> The maximum length of the key should be the one obtained by calling mcrypt_enc_get_key_size() and every value smaller than this is legal.

Assuming the actually meant actually "shorter" instead of "smaller", long hashes may not even work. I have no idea what `mcrypt_enc_get_key_size()` might return.

jimwins: Okay, this example is for a function that was removed in PHP 7.2, so I don't think it's worth putting any energy into improving the example beyond what I already did.

cmb69: Fair enough.

cmb69 (approved):
Looks good to me.

PR #3622, opened 2024-07-28
curl-close.xml: Add a hint on how to destroy the cURL handle

mmalferov:
I know that the cURL handle is now an instance of the `CurlHandle` class. I also know that we can free up resources by deleting an object through the `unset()` language construct. Maybe the mention of unset() is redundant. But something tells me that the `curl_close()` page violates the principle: Rejecting — offer! The page warned that the curl_close() function is a no-op as of PHP 8.0.0, but did not offer an alternative (at least in the code example).

What do dear colleagues think about this?

### DIFF
--- a/reference/curl/functions/curl-close.xml
+++ b/reference/curl/functions/curl-close.xml
@@ -72,7 +72,7 @@ curl_setopt($ch, CURLOPT_HEADER, 0);
 curl_exec($ch);
 
 // Close cURL resource, and free up system resources
-curl_close($ch); // As of PHP 8.0.0 use unset($ch)
+curl_close($ch); // As of PHP 8.0.0 use unset($ch) instead
 
 ?>
 ]]>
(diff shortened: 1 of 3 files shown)


cmb69:
Well, it's not really about unsetting the variable, but rather about the refcount of the object; when the latter decreases to zero, the object is released. For the given code examples, `unset($ch)` works the same like `$ch = null` or even `$ch = "bye"` would. And if these examples are used stand-alone, there is even no need for this (at the end of the script, global variables get their refcount decreased, anyway). Same if the code snippets were the body of a function. But just adding something like `$ch1 = $ch` somewhere in the middle of these examples, will not release the object when `unset($ch)` is called.

I think that users must understand the refcount principle, and also what happens when a certain object is released (i.e. what's happening in `__destruct()`, basically). Whether a hint at `unset()` helps them to be reminded of that – I don't know.

PR #3623, opened 2024-07-28
curl-init.xml: replace 'resource' to 'session'

mmalferov:
I think that in the context of the curl_init() function, it is appropriate to replace the term "resource" with the term "session", since the resource is no longer relevant since PHP 8.0.0, and the session is relevant for both newer versions of PHP and before PHP 8.0.0

### DIFF
--- a/reference/curl/functions/curl-init.xml
+++ b/reference/curl/functions/curl-init.xml
@@ -89,18 +89,20 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-// create a new cURL resource
+
+// Create a new cURL session
 $ch = curl_init();
 
-// set URL and other appropriate options
+// Set URL and other appropriate options
 curl_setopt($ch, CURLOPT_URL, "http://www.example.com/");
 curl_setopt($ch, CURLOPT_HEADER, 0);
 
-// grab URL and pass it to the browser
+// Grab URL and pass it to the browser
 curl_exec($ch);
 
-// close cURL resource, and free up system resources
+// Close cURL session, and free up system resources
 curl_close($ch);
+
 ?>
 ]]>
     </programlisting>
--- a/reference/curl/functions/curl-init.xml
+++ b/reference/curl/functions/curl-init.xml
@@ -90,7 +90,7 @@
 <![CDATA[
 <?php
 
-// Create a new cURL session
+// Initializes a new cURL session
 $ch = curl_init();
 
 // Set URL and other appropriate options
--- a/reference/curl/functions/curl-init.xml
+++ b/reference/curl/functions/curl-init.xml
@@ -5,7 +5,7 @@
   <refname>curl_init</refname>
   <refpurpose>Initialize a cURL session</refpurpose>
  </refnamediv>
- 
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
@@ -14,8 +14,8 @@
   </methodsynopsis>
   <para>
    Initializes a new session and return a cURL handle for use with the
-   <function>curl_setopt</function>, <function>curl_exec</function>,
-   and <function>curl_close</function> functions.  
+   <function>curl_setopt</function> and <function>curl_exec</function>
+   functions.
   </para>
  </refsect1>
 
@@ -28,7 +28,7 @@
      <listitem>
       <para>
        If provided, the <constant>CURLOPT_URL</constant> option will be set
-       to its value. You can manually set this using the 
+       to its value. You can manually set this using the
        <function>curl_setopt</function> function.
       </para>
       <note>
@@ -49,7 +49,7 @@
    Returns a cURL handle on success, &false; on errors.
   </para>
  </refsect1>
- 
+
  <refsect1 role="changelog">
   &reftitle.changelog;
   <informaltable>
@@ -98,10 +98,7 @@ curl_setopt($ch, CURLOPT_URL, "http://www.example.com/");
 curl_setopt($ch, CURLOPT_HEADER, 0);
 
 // Grab URL and pass it to the browser
-curl_exec($ch);
-
-// Close cURL session, and free up system resources
-curl_close($ch);
+echo curl_exec($ch);
 
 ?>
 ]]>
@@ -114,7 +111,6 @@ curl_close($ch);
   &reftitle.seealso;
   <para>
    <simplelist>
-    <member><function>curl_close</function></member>
     <member><function>curl_multi_init</function></member>
    </simplelist>
   </para>
--- a/reference/curl/functions/curl-init.xml
+++ b/reference/curl/functions/curl-init.xml
@@ -14,8 +14,6 @@
   </methodsynopsis>
   <para>
    Initializes a new session and returns a cURL handle.
-   <function>curl_setopt</function> and <function>curl_exec</function>
-   functions.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/curl/functions/curl-init.xml
```diff
@@ -89,18 +89,20 @@
...
 curl_setopt($ch, CURLOPT_URL, "http://www.example.com/");
 curl_setopt($ch, CURLOPT_HEADER, 0);
 
-// grab URL and pass it to the browser
+// Grab URL and pass it to the browser
 curl_exec($ch);
 
-// close cURL resource, and free up system resources
+// Close cURL session, and free up system resources
 curl_close($ch);
```
cmb69, line 104: The new comment doesn't seem to be right. The old comment would at least be correct prior to PHP 8.0.0. Not sure how to improve the comment; see also the related https://github.com/php/doc-en/pull/3622#issuecomment-2254435350.

kamil-tekiela:
Doesn't it still use a resource under the hood? I think cUrl resource still makes sense.

reference/curl/functions/curl-init.xml
mmalferov: I seem to be confused about the times before PHP 8.0.0 and after)

I don't understand why we can't change the current comment:

> close cURL resource, and free up system resources

with a new one:

> Close cURL session, and free up system resources

for the call `curl_close($ch)`.

I'll leave it here so I don't forget what the fuctions are doing:

* curl_init — Initialize a cURL session
* curl_close — Close a cURL session

— Prior to PHP 8.0.0, the part "close the cURL resource" was relevant, but since version 8.0.0 it has not been, since `resource` was replaced by `CurlHandle` class.
— The phrase "close the cURL session" is relevant for both version.

===

I think it's better to move on to the second part of the sentence after we've dealt with the first one. Because the wording

> and free up system resources

worked only prior to PHP 8.0.0 and lost its relevance as of PHP 8.0.0. 

And my main question in [#3622](https://github.com/php/doc-en/pull/3622#issuecomment-2254435350)  was how to replace the outdated `curl_close` call as of PHP 8.0.0? 

I remember about the refcounter and the garbage collector, which destroys symbols that have no references left. That's not the main question. The main question is that once the `curl_close` call was specified in the examples, it was probably necessary for some reason. Calling `curl_close` no longer has an effect, but remains in the example. The documentation mentions the lack of effect as of PHP 8.0, but does not say what to do for those who, as of PHP 8.0.0, for some reason, still need to free up system resources occupied by easy handles (e.g. in multi handle contained 5000 easy handles) manually (e.g. after reading of easy handle data).

But help me first understand the first part of the sentence: why we can't replace the word `resource` with the word `session`?

mmalferov:
I wish I could agree with you, but I can't. cUrl resource still makes sense... for what? I think, it make sense only prior to PHP 8.0.0.

In the meantime, I'm trying to realize **a)** why the `curl_close()` calls were specified in the examples (probably to free up system resources manually when the need for a easy handle disappeared) and if these calls were actualy needed prior to PHP 8.0.0, then **b)** what exactly should replace them as of PHP 8.0.0, taking into account the fact that the `curl_close()` call no longer has an effect?..

reference/curl/functions/curl-init.xml
cmb69: You are right that this change would already be an improvement, and as such might be good as is. But maybe we can even do better:
````
// Close cURL session, and free up system resources (prior to PHP 8.0.0)
````

Comment (kamil-tekiela) [reply reference/curl/functions/curl-init.xml:104]: > You are right that this change would already be an improvement, and as such might be good as is. But maybe we can even do better:
> 
> ```
> // Close cURL session, and free up system resources (prior to PHP 8.0.0)
> ```

That would still be misleading. If I understand correctly, this function does nothing as of PHP 8.0. It doesn't close a session and it doesn't free up resources. Prior to PHP 8.0, it closed the resource. 

I propose a simpler solution, just remove `curl_close` and its comment from this example. It's useless anyway.

File: reference/curl/functions/curl-init.xml [hunk 2]
```diff
@@ -89,18 +89,20 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-// create a new cURL resource
+
+// Create a new cURL session
```
kamil-tekiela, line 93:
```suggestion
// Initializes a new cURL session
```

Since this is how the page describes it, let's keep it aligned. And if you remove the `curl_close` then there's no issue with this example.

reference/curl/functions/curl-init.xml
cmb69: Yeah, maybe we should just remove it.

mmalferov: I have removed every mention of `curl_close` from the page in https://github.com/php/doc-en/pull/3623/commits/f33d219e9d92c45ef767f9ddb7c7f085eae2a42d commit. At the same time, I removed the trailing whitespace characters and added the `echo` construction to `curl_exec($ch)` so that the output would actually be passed to the browser

reference/curl/functions/curl-init.xml
```diff
@@ -14,8 +14,8 @@
   </methodsynopsis>
   <para>
    Initializes a new session and return a cURL handle for use with the
```
kamil-tekiela, line 16:
```suggestion
   Initializes a new session and returns a cURL handle.
```
And then remove the text after this. The list isn't exhaustive anyway.

reference/curl/functions/curl-init.xml
```diff
@@ -89,18 +89,17 @@
...
 
-// set URL and other appropriate options
+// Set URL and other appropriate options
 curl_setopt($ch, CURLOPT_URL, "http://www.example.com/");
 curl_setopt($ch, CURLOPT_HEADER, 0);
 
-// grab URL and pass it to the browser
-curl_exec($ch);
+// Grab URL and pass it to the browser
+echo curl_exec($ch);
```
kamil-tekiela, line 101: Correct me if I'm wrong, but I thought that `curl_exec()` already does `echo` internally. You'd need to explicitly disable it with [`CURLOPT_RETURNTRANSFER`](https://www.php.net/manual/en/curl.constants.php#constant.curlopt-returntransfer)

mmalferov: You are right, `curl_exec` display result without `echo`, if CURLOPT_RETURNTRANSFER not provided; I was just playing with cURL and used this option))

I looked at the `curl_exec` page and found no mention of the function having an internal output. I think we should go back to `curl_exec` a little later and add one sentence about it.

Yes, I almost forgot. Maybe it makes sense to add a mention of `easy` to the wording for a simple descriptor? For example, like this:

> Initializes a new session and returns a cURL easy handle.

kamil-tekiela: No, just leave it as I suggested and remove the rest of the sentence.

kamil-tekiela (approved):
Thanks! Look ok now. @cmb69 What do you think?

reference/curl/functions/curl-init.xml
```diff
@@ -28,7 +26,7 @@
      <listitem>
       <para>
        If provided, the <constant>CURLOPT_URL</constant> option will be set
-       to its value. You can manually set this using the 
+       to its value. You can manually set this using the
        <function>curl_setopt</function> function.
```
Girgias, line 30: Remove personalization while we are at it.
```suggestion
       to its value. This can be set manually using the
       <function>curl_setopt</function> function.
```

PR #3625, opened 2024-07-28
http-auth.xml: remove mention of digest

mmalferov:
Digest HTTP Authentication example has been removed, but the mention of the content of the deleted paragraph remains


features/http-auth.xml
```diff
@@ -51,7 +51,7 @@ if (!isset($_SERVER['PHP_AUTH_USER'])) {
     uppercase "B", the realm string must be enclosed in double (not single) quotes,
     and exactly one space should precede the <emphasis>401</emphasis> code in the 
     <emphasis>HTTP/1.0 401</emphasis> header line. Authentication parameters have
-    to be comma-separated as seen in the digest example above.
+    to be comma-separated.
```
cmb69, line 54: Good catch, thank you! I'm not sure if the whole sentence should be removed; I believe basic authentication has no further parameters than realm. Mabye @jimwins knows.

jimwins: I'm not sure, but I think it would be safe to remove the whole sentence.

PR #3627, opened 2024-07-28
Convert MongoDB documentation from a top element of set to book

haszi:
Convert all underlying elements as needed:
 - `<books>` to `<part>`s or `<chapter>`s
 - `<article>`s to `<section>`s

Move the Predefined Constants page from the Installing/Configuring section to its own

Change the following `xml:id`s:
 - `book.mongodb` to `mongodb.mongodb`
 - `set.mongodb` to `book.mongodb`
 - `book.bson` to `mongodb.bson`

Please note that these changes only work when the accompanying `doc-base` changes are merged too (https://github.com/php/doc-base/pull/138).

*Note:*
Please note that because of the difference between `root chunks` and `container chunks` in `PhD` (ie. a `<book>` and a `<part>`), the four pages linked to from the main [MongoDB page](https://www.php.net/manual/en/set.mongodb) ([MongoDB](https://www.php.net/manual/en/book.mongodb.php), [BSON](https://www.php.net/manual/en/book.bson.php), [Monitoring](https://www.php.net/manual/en/mongodb.monitoring.php) and [Exceptions](https://www.php.net/manual/en/mongodb.exceptions.php)) are only going to list their underlying classes without listing each method of each class.
E.g. the [MongoDB Extension Classes page](https://www.php.net/manual/en/book.mongodb.php) will only list the 19 classes belonging to that group of classes instead of listing the 19 classes plus all their 144 methods. As before, the methods are still listed on the pages of their respective classes.

### DIFF
--- a/appendices/extensions.xml
+++ b/appendices/extensions.xml
@@ -82,6 +82,7 @@
    <listitem><simpara><xref linkend="book.memcached"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mhash"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.misc"/></simpara></listitem>
+   <listitem><simpara><xref linkend="book.mongodb"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mqseries"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mysql"/></simpara></listitem>
    <listitem><simpara><xref linkend="book.mysql-xdevapi"/></simpara></listitem>
@@ -342,6 +343,7 @@
     <listitem><para><xref linkend="book.mcrypt"/></para></listitem>
     <listitem><para><xref linkend="book.memcache"/></para></listitem>
     <listitem><para><xref linkend="book.memcached"/></para></listitem>
+    <listitem><para><xref linkend="book.mongodb"/></para></listitem>
     <listitem><para><xref linkend="book.mqseries"/></para></listitem>
     <listitem><para><xref linkend="book.mysql"/></para></listitem>
     <listitem><para><xref linkend="book.mysql-xdevapi"/></para></listitem>
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -3585,7 +3585,7 @@ local: {
      </para>
      <note>
       <simpara>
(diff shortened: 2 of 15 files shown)


haszi:
This PR is now ready for review. The build failure is due to this PR only working in tandem with the accompanying `doc-base` PR (https://github.com/php/doc-base/pull/138) as the overall hierarchy of the manual is defined in `doc-base`.

cmb69:
Thank you!

This appears to be a very reasonable change to me, and I think it's okay for translations to catch up with this soon (given that the required changes are kept within bounds).

So, in my opinion, this is good to go.

Girgias (approved):
This looks reasonable to me and not that hard to update.

haszi:
Please advise whether we can move forward with this and the `doc-base` PR or if I need to make any changes to this PR.

If this gets merged, I can open a PR with the same changes in some/all of the translation repos.

cmb69 (approved):
Fine for me!

Maybe @alcaeus and @jmikola want to have a look, though.

alcaeus:
Current set of changes LGTM!

cmb69:
I think this just needs https://github.com/php/doc-base/pull/138 to work.

jmikola:
No objections from my end. Thanks for taking this on @haszi.

> Please note that because of the difference between root chunks and container chunks in PhD (ie. a `<book>` and a `<part>`), the four pages linked to from the main [MongoDB page](https://www.php.net/manual/en/set.mongodb) ([MongoDB](https://www.php.net/manual/en/book.mongodb.php), [BSON](https://www.php.net/manual/en/book.bson.php), [Monitoring](https://www.php.net/manual/en/mongodb.monitoring.php) and [Exceptions](https://www.php.net/manual/en/mongodb.exceptions.php)) are only going to list their underlying classes without listing each method of each class.

That will be a welcome improvement.

jmikola:
@cmb69: Thanks for merging.

Would it be possible to setup a redirect from https://www.php.net/manual/en/set.mongodb.php to https://www.php.net/manual/en/book.mongodb.php? I expect there are many existing links out there pointing to the original URL, which is now a 404.

In the meantime I'll look through MongoDB's sites to ensure we update our own links.

jmikola:
It's not clear to me exactly how many URLs changed as a result of this PR, but we're finding additional 404s since this was merged.

For example, changing `<article>` to `<section>` in `configure.xml` removed specific pages for each install section such that https://www.php.net/manual/en/mongodb.installation.pecl.php is now a 404 and that content is now accessible via an anchor within a page for all of the install docs: https://www.php.net/manual/en/mongodb.installation.php#mongodb.installation.pecl

I'm going to attempt a local build before and after this PR to compare URLs and try to add those as additional redirects in the vein of https://github.com/php/web-php/pull/1100.

PR #3632, opened 2024-08-01
fix windows instructions for parallel

realFlowControl:


reference/parallel/setup.xml
```diff
@@ -30,15 +30,14 @@
    <link xlink:href="&url.pecl.package;parallel">&url.pecl.package;parallel</link>.
   </para>
   <para>
- Windows users can download prebuilt release binaries from the <link xlink:href="&url.pecl.windows.releases;parallel">PECL</link> website.
+ Windows users can download prebuilt release binaries from the <link xlink:href="&url.pecl.package;parallel">PECL</link> website.
   </para>
   <caution>
    <para>
- Windows users need to take the additional step of adding pthreadVC2.dll (distributed with Windows releases) to their <envar>PATH</envar>.
+ Windows users need to take the additional step of adding pthreadVC3.dll (distributed with Windows releases) to their <envar>PATH</envar>.
```
cmb69, line 37: Since older parallel version (prior to 1.2.0) are still available and ship pthreadVC2.dll, maybe make that more generic, and state pthreadVC?.dll. And probably some markup could be applied, possibly:
```suggestion
 Windows users need to take the additional step of adding <filename>pthreadVC<replaceable>?</replaceable>.dll</filename> (distributed with Windows releases) to their <envar>PATH</envar>.
```

reference/parallel/setup.xml
```diff
@@ -30,15 +30,14 @@
    <link xlink:href="&url.pecl.package;parallel">&url.pecl.package;parallel</link>.
   </para>
   <para>
- Windows users can download prebuilt release binaries from the <link xlink:href="&url.pecl.windows.releases;parallel">PECL</link> website.
+ Windows users can download prebuilt release binaries from the <link xlink:href="&url.pecl.package;parallel">PECL</link> website.
```
cmb69, line 33: Indeed, it's probably best to refer Windows users to the PECL page. There are a couple of other extension which still use this entity; I'll follow up on that.

cmb69 (commented):
Thank you!

reference/parallel/setup.xml
realFlowControl: Good Idea!

PR #3633, opened 2024-08-01
Don't link to the direct downloads, but rather to the PECL page

cmb69:
Besides that new builds are no longer placed on that server, users better go to the PECL page, where they can get more information anyway.

See also #3632.


jimwins (approved):
LGTM

sy-records (approved):

PR #3635, opened 2024-08-01
Moved early algorithm description to warning block

mumumu:
image

### DIFF
--- a/reference/hash/book.xml
+++ b/reference/hash/book.xml
@@ -51,12 +51,14 @@
       currently approved by United States Federal Information Processing
       Standards</link>.
      </simpara>
-     <simpara>
-      Some of the early cryptographic algorithms, such as <literal>"md4"</literal>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! I agree that this information should be more emphasized, but a `<warning>` might be too strong. Maybe use `<caution>` instead. @jimwins, @TimWolla, thoughts?

TimWolla:
<caution> makes sense to me and matches the caution note for `rand()`: https://www.php.net/manual/en/function.rand

jimwins:
Works for me, no strong opinion on warning vs. caution.

mumumu:
OK. changed to `<caution>` on https://github.com/php/doc-en/pull/3635/commits/b4c7903b8bd3c0328f91d8bb6f0c779e9f59ede5.

cmb69 (approved):

PR #3636, opened 2024-08-01
Add missing CURLOPT constants

haszi:
Add the 52 missing `CURLOPT_*` constants and their descriptions to the cURL predefined constant list. 

Reference used: https://curl.se/libcurl/c/symbols-in-versions.html

### DIFF
--- a/reference/curl/constants_curl_setopt.xml
+++ b/reference/curl/constants_curl_setopt.xml
@@ -377,7 +377,8 @@
    </term>
    <listitem>
     <para>
-     Pass a <type>string</type> naming a file with the concatenation of CRL (in PEM format)
+     Pass a <type>string</type> naming a file with the concatenation of
+     CRL (Certificate Revocation List) (in PEM format)
      to use in the certificate validation that occurs during the SSL exchange.
      When cURL is built to use GnuTLS,
      there is no way to influence the use of CRL passed
--- a/reference/curl/constants_curl_setopt.xml
+++ b/reference/curl/constants_curl_setopt.xml
@@ -17,6 +17,44 @@
     </para>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-accept-encoding">
+   <term>
+    <constant>CURLOPT_ACCEPT_ENCODING</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Sets the contents of the <literal>Accept-Encoding:</literal> header
+     sent in an HTTP request.
+     Available as of cURL 7.21.6
+    </para>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-accepttimeout-ms">
+   <term>
+    <constant>CURLOPT_ACCEPTTIMEOUT_MS</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     The maximum number of milliseconds to wait for a server
+     to connect back to libcurl when an active FTP connection is used.
+     Available as of cURL 7.24.0
+    </para>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-address-scope">
+   <term>
+    <constant>CURLOPT_ADDRESS_SCOPE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     The scope id value to use when connecting to IPv6 addresses.
+     Available as of cURL 7.19.0
+    </para>
+   </listitem>
+  </varlistentry>
   <varlistentry xml:id="constant.curlopt-altsvc">
    <term>
     <constant>CURLOPT_ALTSVC</constant>
@@ -47,6 +85,19 @@
      <constant>CURLALTSVC_H3</constant>, and
      <constant>CURLALTSVC_READONLYFILE</constant>.
      Available as of PHP 8.2.0 and cURL 7.64.1.
+    </para>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-append">
+   <term>
+    <constant>CURLOPT_APPEND</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Setting this option to <literal>1</literal> will cause FTP uploads
+     to append to the remote file instead of overwriting it.
+     Available as of cURL 7.17.0
     </para>
    </listitem>
   </varlistentry>
@@ -312,6 +363,30 @@
     </para>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-crlfile">
+   <term>
+    <constant>CURLOPT_CRLFILE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a string naming a file with the concatenation of CRL (in PEM format)
+     to use in the certificate validation that occurs during the SSL exchange.
+     When curl is built to use GnuTLS,
+     there is no way to influence the use of CRL passed
+     to help in the verification process.
+     When libcurl is built with OpenSSL support,
+     <literal>X509_V_FLAG_CRL_CHECK</literal>
+     and <literal>X509_V_FLAG_CRL_CHECK_ALL</literal> are both set,
+     requiring CRL check against all the elements of the certificate chain
+     if a CRL file is passed.
+     Also note that <constant>CURLOPT_CRLFILE</constant> implies
+     <constant>CURLSSLOPT_NO_PARTIALCHAIN</constant>
+     as of cURL 7.71.0 due to an OpenSSL bug.
+     Available as of cURL 7.19.0
+    </para>
+   </listitem>
+  </varlistentry>
   <varlistentry xml:id="constant.curlopt-customrequest">
    <term>
     <constant>CURLOPT_CUSTOMREQUEST</constant>
@@ -346,6 +421,25 @@
     <para>
      The default protocol to use if the URL is missing a scheme name.
      Available as of PHP 7.0.7 and cURL 7.45.0
+    </para>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-dirlistonly">
+   <term>
+    <constant>CURLOPT_DIRLISTONLY</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Setting this option to <literal>1</literal> will have different effects
+     based on the protocol it is used with.
+     FTP and SFTP based URLs will list only the names of files in a directory.
+     POP3 will list the email message or messages on the POP3 server.
+     For FILE, this option has no effect
+     as directories are always listed in this mode.
+     Using this option with <constant>CURLOPT_WILDCARDMATCH</constant>
+     will prevent the latter from having any effect.
+     Available as of cURL 7.17.0.
     </para>
    </listitem>
   </varlistentry>
@@ -409,6 +503,20 @@
      Set the local IPv6 address that the resolver should bind to. The argument
      should contain a single numerical IPv6 address as a string.
      Available as of PHP 7.0.7 and cURL 7.33.0
+    </para>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-dns-servers">
+   <term>
+    <constant>CURLOPT_DNS_SERVERS</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a string with a comma-separated list of DNS servers to be used
+     instead of the system default
+     (e.g.: <literal>192.168.1.100,192.168.1.101:8080</literal>).
+     Available as of cURL 7.24.0
     </para>
    </listitem>
   </varlistentry>
@@ -568,6 +676,55 @@
     </para>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-fnmatch-function">
+   <term>
+    <constant>CURLOPT_FNMATCH_FUNCTION</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a callback that will be used for wildcard matching.
+     The signature of the callback should be:
+     <methodsynopsis>
+      <type>int</type><methodname><replaceable>callback</replaceable></methodname>
+      <methodparam><type>resource</type><parameter>curlHandle</parameter></methodparam>
+      <methodparam><type>string</type><parameter>pattern</parameter></methodparam>
+      <methodparam><type>string</type><parameter>string</parameter></methodparam>
+     </methodsynopsis>
+     <variablelist role="function_parameters">
+      <varlistentry>
+       <term><parameter>curlHandle</parameter></term>
+       <listitem>
+        <simpara>
+         The cURL handle.
+        </simpara>
+       </listitem>
+      </varlistentry>
+      <varlistentry>
+       <term><parameter>pattern</parameter></term>
+       <listitem>
+        <simpara>
+         The wildcard pattern.
+        </simpara>
+       </listitem>
+      </varlistentry>
+      <varlistentry>
+       <term><parameter>string</parameter></term>
+       <listitem>
+        <simpara>
+         The string to run the wildcard pattern matching on.
+        </simpara>
+       </listitem>
+      </varlistentry>
+     </variablelist>
+     The callback should return
+     <constant>CURL_FNMATCHFUNC_MATCH</constant> if pattern matches the string,
+     <constant>CURL_FNMATCHFUNC_NOMATCH</constant> if not
+     or <constant>CURL_FNMATCHFUNC_FAIL</constant> if an error occurred.
+     Available as of cURL 7.21.0.
+    </para>
+   </listitem>
+  </varlistentry>
   <varlistentry xml:id="constant.curlopt-followlocation">
    <term>
     <constant>CURLOPT_FOLLOWLOCATION</constant>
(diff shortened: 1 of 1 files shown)


reference/curl/constants_curl_setopt.xml
```diff
@@ -568,6 +676,55 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-fnmatch-function">
+   <term>
+    <constant>CURLOPT_FNMATCH_FUNCTION</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a callback that will be used for wildcard matching.
```
Girgias, line 686: Probably?
```suggestion
     Pass a <type>callable</type> that will be used for wildcard matching.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -412,6 +506,20 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-dns-servers">
+   <term>
+    <constant>CURLOPT_DNS_SERVERS</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a string with a comma-separated list of DNS servers to be used
```
Girgias, line 516: Type tag around string?

reference/curl/constants_curl_setopt.xml
```diff
@@ -312,6 +363,30 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-crlfile">
+   <term>
+    <constant>CURLOPT_CRLFILE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a string naming a file with the concatenation of CRL (in PEM format)
```
Girgias, line 373: Type tag around string?

reference/curl/constants_curl_setopt.xml
```diff
@@ -675,6 +832,33 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-ftp-account">
+   <term>
+    <constant>CURLOPT_FTP_ACCOUNT</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a string that will be sent as account information over FTP
```
Girgias, line 842: Type tag?

reference/curl/constants_curl_setopt.xml
```diff
@@ -675,6 +832,33 @@
...
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-ftp-alternative-to-user">
+   <term>
+    <constant>CURLOPT_FTP_ALTERNATIVE_TO_USER</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a string that will be used to try to authenticate over FTP
```
Girgias, line 856: Ditto

reference/curl/constants_curl_setopt.xml
```diff
@@ -714,6 +927,28 @@
...
+     which shuts down the SSL/TLS layer after authenticating
+     making the rest of the control channel communication unencrypted.
+     Use one of the following values:
+     <constant>CURLFTPSSL_CCC_NONE</constant>
+      - do not attempt to use CCC.
+     <constant>CURLFTPSSL_CCC_PASSIVE</constant>
+      - Do not initiate the shutdown, but wait for the server to do it.
+      Do not send a reply.
+     <constant>CURLFTPSSL_CCC_ACTIVE</constant>
+      - Initiate the shutdown and wait for a reply.
```
Girgias, line 947: Looks like this should be a simplelist probably?

reference/curl/constants_curl_setopt.xml
```diff
@@ -936,6 +1201,21 @@
...
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-http-transfer-decoding">
+   <term>
+    <constant>CURLOPT_HTTP_TRANSFER_DECODING</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     If set to <literal>0</literal>, transfer decoding is disabled.
+     If set to <literal>1</literal>, transfer decoding is enabled (default).
```
Girgias, line 1212: Does this check for a boolean value in the extension or just doesn't care?

Girgias (commented):
Cursory glance, main question is if the extension handles boolean options with a bool or not.

Also I don't remember if @cmb69 knows a lot about cURL or not.

haszi:
> Cursory glance, main question is if the extension handles boolean options with a bool or not.

If I understand the [source](https://github.com/php/php-src/blob/0f8259e896dd609aa9f5a8274ead17d81185a45c/ext/curl/interface.c#L1803C33-L1815C51) correctly, the extension converts all values to long before passing them to the appropriate C function. I tested it locally and bool CURLOPT options accepted bool, int and float values too.

reference/curl/constants_curl_setopt.xml
haszi: The extension just doesn't care and converts all int and bool options' values to long internally.

reference/curl/constants_curl_setopt.xml
haszi: You're right but for now, I'd like to keep this and the rest of the similar lists as they are and address this in a follow-up PR where I'd convert these to `<simplelist>`s and remove the descriptions of the listed constants (as these are either in my [other PR](https://github.com/php/doc-en/pull/3671) or I'll add them if they're missing).

reference/curl/constants_curl_setopt.xml
cmb69: [CURLOPT_HTTP_TRANSFER_DECODING;](https://curl.se/libcurl/c/CURLOPT_HTTP_TRANSFER_DECODING.html):

> Pass a long to tell libcurl how to act on transfer decoding. If set to zero, transfer decoding is disabled, if set to 1 it is enabled (default).

I wonder how that is actually implemented in cURL.

haszi:
@Girgias I've made some updates to the constant list based on your review comments. Could you check my last commit and let me know whether the general direction of the changes will work or not? If it does, I'll update the entire file. If it doesn't, I didn't waste too much time on it yet. :-)

Girgias:
> @Girgias I've made some updates to the constant list based on your review comments. Could you check my last commit and let me know whether the general direction of the changes will work or not? If it does, I'll update the entire file. If it doesn't, I didn't waste too much time on it yet. :-)

I quite like what you did in the last commit :)

haszi:
A note in case someone would like to tackle this in `php-src`:  values passed to some of the `CURLOPT_*` constants are not type checked and these constants should probably be added to the appropriate parts of the huge `switch` statement of `_php_curl_setopt()` in `ext/curl`'s `interface.c`.

cmb69:
`CURLOPT_FNMATCH_FUNCTION` and `CURLOPT_HEADERFUNCTION` are already properly handled there (see the "callables" section at the top of that `switch` (macro magic). The others are aliases for already handled cases, and there must be no duplicates (adding a comment might make sense, though).

It would be more important, I think, to declare the missing `@alias`es in curl.stub.php.

haszi:
> `CURLOPT_FNMATCH_FUNCTION` and `CURLOPT_HEADERFUNCTION` are already properly handled there (see the "callables" section at the top of that `switch` (macro magic). 

@cmb69 Sorry that I wasn't specific enough but I meant that these constants are not being type checked on `master` anymore (https://github.com/php/php-src/pull/13291 removed the two you have mentioned). But you're right, these were there before https://github.com/php/php-src/pull/13291 so I'm really suggesting to take a look at these on `master` only.

cmb69:
@haszi, I've actually debugged `CURLOPT_FNMATCH_FUNCTION` and `CURLOPT_HEADERFUNCTION` on master, and both are properly type checked. As of https://github.com/php/php-src/pull/13291, this happens due to `HANDLE_CURL_OPTION_CALLABLE(ch, CURLOPT_FNMATCH_, handlers.fnmatch, curl_fnmatch);` and `HANDLE_CURL_OPTION_CALLABLE_PHP_CURL_USER(ch, CURLOPT_HEADER, write_header);`. You just won't be able to seek for these constants, due to the macro magic.

haszi:
> @haszi, I've actually debugged `CURLOPT_FNMATCH_FUNCTION` and `CURLOPT_HEADERFUNCTION` on master, and both are properly type checked. As of [php/php-src#13291](https://github.com/php/php-src/pull/13291), this happens due to `HANDLE_CURL_OPTION_CALLABLE(ch, CURLOPT_FNMATCH_, handlers.fnmatch, curl_fnmatch);` and `HANDLE_CURL_OPTION_CALLABLE_PHP_CURL_USER(ch, CURLOPT_HEADER, write_header);`. You just won't be able to seek for these constants, due to the macro magic.

You're right. How did I even miss that part? Sorry for the noise and thanks for checking on this. :-)

haszi:
@Girgias I've added some applicable tags, information on what type of value each option accepts and (where known) what the default value is.

reference/curl/constants_curl_setopt.xml
```diff
@@ -50,6 +94,20 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-append">
+   <term>
+    <constant>CURLOPT_APPEND</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Setting this option to <literal>1</literal> will cause FTP uploads
```
Girgias, line 104: I think having an `<acronym>` tag around FTP would be good here.

reference/curl/constants_curl_setopt.xml
```diff
@@ -312,6 +370,30 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-crlfile">
+   <term>
+    <constant>CURLOPT_CRLFILE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a <type>string</type> naming a file with the concatenation of CRL (in PEM format)
```
Girgias, line 380: What does CRL mean? Might make sense to add a definition of it in `doc-base/entities/acronyms.xml`

reference/curl/constants_curl_setopt.xml
```diff
@@ -349,6 +431,26 @@
...
+    <constant>CURLOPT_DIRLISTONLY</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Setting this option to <literal>1</literal> will have different effects
+     based on the protocol it is used with.
+     FTP and SFTP based URLs will list only the names of files in a directory.
+     POP3 will list the email message or messages on the POP3 server.
+     For FILE, this option has no effect
```
Girgias, line 445: I'm not sure what

> For FILE [...]

refers to

reference/curl/constants_curl_setopt.xml
```diff
@@ -349,6 +431,26 @@
...
+   <term>
+    <constant>CURLOPT_DIRLISTONLY</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Setting this option to <literal>1</literal> will have different effects
+     based on the protocol it is used with.
+     FTP and SFTP based URLs will list only the names of files in a directory.
+     POP3 will list the email message or messages on the POP3 server.
```
Girgias, line 444: FTP and POP3 have acronyms defined but not SFTP

reference/curl/constants_curl_setopt.xml
```diff
@@ -675,6 +840,35 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-ftp-account">
+   <term>
+    <constant>CURLOPT_FTP_ACCOUNT</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a <type>string</type> that will be sent as account information over FTP
```
Girgias, line 850: Acronym tag for FTP

reference/curl/constants_curl_setopt.xml
```diff
@@ -675,6 +840,35 @@
...
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.curlopt-ftp-alternative-to-user">
+   <term>
+    <constant>CURLOPT_FTP_ALTERNATIVE_TO_USER</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Pass a <type>string</type> that will be used to try to authenticate over FTP
```
Girgias, line 866: Ditto

reference/curl/constants_curl_setopt.xml
```diff
@@ -703,6 +897,38 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-ftp-response-timeout">
+   <term>
+    <constant>CURLOPT_FTP_RESPONSE_TIMEOUT</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     A timeout in seconds cURL will wait for a response from an FTP server.
```
Girgias, line 907: Ditto

reference/curl/constants_curl_setopt.xml
```diff
@@ -714,6 +940,22 @@
...
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-ftp-ssl-ccc">
+   <term>
+    <constant>CURLOPT_FTP_SSL_CCC</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     This option makes cURL use CCC (Clear Command Channel)
+     which shuts down the SSL/TLS layer after authenticating
```
Girgias, line 951: Possibly add `CCC` to the acronyms file and refer to it here?

reference/curl/constants_curl_setopt.xml
```diff
@@ -740,6 +982,38 @@
...
+  <varlistentry xml:id="constant.curlopt-ftp-use-pret">
+   <term>
+    <constant>CURLOPT_FTP_USE_PRET</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Set to <literal>1</literal> to send a <literal>PRET</literal> command
+     before <literal>PASV</literal> (and <literal>EPSV</literal>).
+     Has no effect when using the active FTP transfers mode.
```
Girgias, line 994: Acronym tag for FTP

reference/curl/constants_curl_setopt.xml
```diff
@@ -1009,6 +1314,24 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-issuercert">
+   <term>
+    <constant>CURLOPT_ISSUERCERT</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     If set to a <type>string</type> naming a file holding a CA certificate in PEM format,
```
Girgias, line 1324: Acronym tag around `PEM`

reference/curl/constants_curl_setopt.xml
```diff
@@ -1289,6 +1790,27 @@
...
+     Set a <type>string</type> consisting of a comma separated list of hostnames
+     that do not require a proxy to get reached.
+     Each name in this list is matched as either a domain
+     which contains the hostname or the hostname itself.
+     The only wildcard available in the <type>string</type>
+     is a single <literal>*</literal> character which matches all hosts,
+     effectively disabling the proxy.
+     Setting this option to an empty <type>string</type> enables the proxy for all hostnames.
+     Since cURL 7.86.0, IP addresses set with this option
+     can be provided using CIDR notation.
```
Girgias, line 1809: CIDR maybe should have an acronym defined

reference/curl/constants_curl_setopt.xml
```diff
@@ -2983,6 +3817,26 @@
...
   </varlistentry>
+  <varlistentry xml:id="constant.curlopt-use-ssl">
+   <term>
+    <constant>CURLOPT_USE_SSL</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <para>
+     Sets the desired level of SSL/TLS for the transfer
+     when using FTP, SMTP, POP3, IMAP etc.
```
Girgias, line 3828: I think a comma is required here
```suggestion
     when using FTP, SMTP, POP3, IMAP, etc.
```

Girgias (commented):
Not sure if the acronym tag comments are that useful as it would be adding a lot of them

cmb69:
> Not sure if the acronym tag comments are that useful as it would be adding a lot of them

At least adding some would be great! And others might be added later; in my opinion, our current list of acronyms is really meager.

haszi:
> > Not sure if the acronym tag comments are that useful as it would be adding a lot of them
> 
> At least adding some would be great! And others might be added later; in my opinion, our current list of acronyms is really meager.

I took a look at `acronyms.xml` in `doc-base` and we only have 70 acronyms listed there. While I agree that more need to be added there and used on the `CURLOPT_*` constants page, I think I'd like to tackle that as a separate set of PRs (ie. adding them in `doc-base` and probably a large number of PRs adding the `<acronym>` tags for a large part of the `doc-en` files).

Another (selfish) reason is that I have a PR ready with the same type of tags/value type/default value changes made to all `CURLOPT_*` constants that I've added to ones in this PR and that PR is already humongous. :-)

reference/curl/constants_curl_setopt.xml
haszi: Basically `file://`.

reference/curl/constants_curl_setopt.xml
```diff
@@ -1069,6 +1393,59 @@
...
+    <para>
+     Set the kerberos security level for FTP and also enables kerberos awareness.
+     This should be set to one of the following <type>string</type>s:
+     <simplelist type="inline">
+      <member><literal>clear</literal></member>
+      <member><literal>safe</literal></member>
+      <member><literal>confidential</literal></member>
+      <member><literal>private</literal></member>
+     </simplelist>
+     .
```
cmb69, line 1411: I guess that stray period looks strange when rendered; I'd probably remove it.

cmb69 (approved):
On a cursory look, this looks good to me (a single nit below). And I'm fine with postponing the acronyms.

Maybe wait on @Girgias's review before merging.

reference/curl/constants_curl_setopt.xml
haszi: The period is needed at the end of the sentence because `<simplelist>` rendering only adds the `<member>` separating commas.
That `<simplelist>` actually renders like this:

`This should be set to one of the following strings: clear, safe, confidential, private .`

haszi:
One more thing in regards to acronyms: we should probably add an `ENTITY` for at least the most frequently used ones (eg.: HTTP, FTP, SSL/TLS, etc.) so that we can use something like `&HTTP;` instead of the unnecessarily long `<acronym>HTTP</acronym>`.

reference/curl/constants_curl_setopt.xml
cmb69: D'oh! I've missed the `type="inline"`. Sorry for the noise!

Girgias:
> One more thing in regards to acronyms: we should probably add an `ENTITY` for at least the most frequently used ones (eg.: HTTP, FTP, SSL/TLS, etc.) so that we can use something like `&HTTP;` instead of the unnecessarily long `<acronym>HTTP</acronym>`.

That is a good point, could added in `doc-base/entities/global.ent`

Girgias (approved):
LGTM :)

PR #3637, opened 2024-08-02
Update integer overflow example

wpdevelopment11:
I think this way it's more obvious what you're trying to show.

### DIFF
--- a/language/types/integer.xml
+++ b/language/types/integer.xml
@@ -107,42 +107,23 @@ integer     : decimal
   </para>
 
   <example>
-   <title>Integer overflow on a 32-bit system</title>
+   <title>Integer overflow</title>
    <programlisting role="php">
 <![CDATA[
 <?php
-$large_number = 2147483647;
-var_dump($large_number);                     // int(2147483647)
+$large_number = 50000000000000000000;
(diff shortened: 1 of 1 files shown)


cmb69:
It would say it's debatable what's more obvious. Another option would be to write `var_dump($large_number + 1);`.

wpdevelopment11:
For me it's not immediately obvious that the second number differs only in one here:
```
$large_number = 2147483647;
var_dump($large_number);                     // int(2147483647)
$large_number = 2147483648;
var_dump($large_number);                     // float(2147483648)
$million = 1000000;
```

cmb69:
I had a closer look: now we have two different examples (for 32bit and 64bit); it might make sense to have only one example, using `PHP_INT_MAX` and stating the results for 32bit/64bit in the comments. If we go this way, we can't specify any literal numbers, but would need to go with `+=` or `+ 1` anyway.

wpdevelopment11:
Updated according to your comments.
I think that large literal is still useful to show that not only expressions are interpreted as floats, but literals too.
Moved the integer division example, I don't think that it's related to overflow.

cmb69:
Thank you!

PR #3642, opened 2024-08-06
session-name.xml: use the correct directive name and add the link to

mmalferov:
There is no directive `session.transid` in PHP. It's probably about the `session.use_trans_sid` directive instead `session.transid`, if I got the context right

### DIFF
--- a/reference/session/functions/session-name.xml
+++ b/reference/session/functions/session-name.xml
@@ -22,8 +22,8 @@
    If a new session <parameter>name</parameter> is
    supplied, <function>session_name</function> modifies the HTTP cookie
    (and output content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
-   enabled). Once the HTTP cookie is
-   sent, <function>session_name</function> raises error.
+   enabled). Once the HTTP cookie has been
+   sent, calling <function>session_name</function> raises an <constant>E_WARNING</constant>.
    <function>session_name</function> must be called
    before <function>session_start</function> for the session to work
    properly.
--- a/reference/session/functions/session-name.xml
+++ b/reference/session/functions/session-name.xml
@@ -21,7 +21,7 @@
   <para>
    If a new session <parameter>name</parameter> is
    supplied, <function>session_name</function> modifies the HTTP cookie
-   (and output content when <literal>session.transid</literal> is
+   (and output content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
    enabled). Once the HTTP cookie is
    sent, <function>session_name</function> raises error.
    <function>session_name</function> must be called
--- a/reference/session/functions/session-name.xml
+++ b/reference/session/functions/session-name.xml
@@ -21,7 +21,7 @@
   <para>
    If a new session <parameter>name</parameter> is
    supplied, <function>session_name</function> modifies the HTTP cookie
-   (and output content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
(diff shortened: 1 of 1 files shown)


reference/session/functions/session-name.xml
```diff
@@ -21,7 +21,7 @@
   <para>
    If a new session <parameter>name</parameter> is
    supplied, <function>session_name</function> modifies the HTTP cookie
-   (and output content when <literal>session.transid</literal> is
+   (and output content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
    enabled). Once the HTTP cookie is
    sent, <function>session_name</function> raises error.
```
cmb69, line 26: Frankly, the whole part doesn't make much sense to me. The last sentence is particularly confusing.

Maybe @Girgias can have a look at this.

mmalferov: In russian docs, i translated the last sentence like this:

> Функция session_name() выдаёт ошибку, если функцию вызвали после отправки cookie HTTP-протокола (The session_name() function returns an error if the function was called after sending the HTTP cookie).

```
<?php

/* OK */
session_name('customCookieName');
session_start(); // It adds cookie 'customCookieName=snsor4a6cjbq0cfa0ujbb6s8fk'
```

```
<?php

/* Bad */
session_start();
session_name('customCookieName');
// Warning: session_name(): Session name cannot be changed when a session is active
```
P. S. It seems to me that the whole paragraph makes sense, since it explains the behavior of the session mechanism a) when the 'session.use-trans-sid' option is enabled (which changes the name of the cookie data block and URLs in the page content for <a>, <input> of form and others fields, depending on the settings of the `session.trans_sid_tags` option) and b) when calling the `session_name` function _ after sending cookies ;-)

cmb69, line 26: Okay, this should probably be something like:
```suggestion
   enabled). Once the HTTP cookie has been
   sent, calling <function>session_name</function> raises an <constant>E_WARNING</constant>.
```

reference/session/functions/session-name.xml
```diff
@@ -21,7 +21,7 @@
   <para>
    If a new session <parameter>name</parameter> is
    supplied, <function>session_name</function> modifies the HTTP cookie
-   (and output content when <literal>session.transid</literal> is
+   (and output content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
```
cmb69, line 24: I still don't understand this part. Does calling `session_name("whatever")` already output content (assuming `session.use_trans_sid=1`) and sends a `set-cookie` header? I think this only happens during further script execution.

Just looked at the German translation, and according to this it should better be: "(and the content of the output when `session.use_trans_sid` is enabled)"

reference/session/functions/session-name.xml
```diff
@@ -21,9 +21,9 @@
   <para>
    If a new session <parameter>name</parameter> is
    supplied, <function>session_name</function> modifies the HTTP cookie
-   (and output content when <literal>session.transid</literal> is
-   enabled). Once the HTTP cookie is
-   sent, <function>session_name</function> raises error.
+   (and output content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
```
kamil-tekiela, line 24:
```suggestion
   (and outputs the content when <link linkend="ini.session.use-trans-sid">session.use_trans_sid</link> is
```

Girgias (approved):
Looks fine to me

cmb69:
Thank you!

Issue #3643, opened 2024-08-07
Potential deprecated list of "Experimental Extensions"


iszmais:
From manual page: https://php.net/extensions.state

---

Hello,

I stumbled upon this manual page and was quite irritated about the list of experimental extensions (also that there are no deprecated extensions at all).
Since I could not find any other source that states these extensions to be experimental, im not sure if this data is still valid.

Could it be that this documentation itself is deprecated?

Greetings,
@iszmais

cmb69:
Indeed, this page is grossly out-dated. There is https://github.com/php/doc-base/blob/master/scripts/qa/extensions.xml.php which is able to update https://www.php.net/manual/en/extensions.alphabetical.php and https://www.php.net/manual/en/extensions.membership.php, but https://www.php.net/manual/en/extensions.state.php is apparently not yet supported. A PR would be welcome.

PR #3644, opened 2024-08-07
Remove extension state tags which do no longer apply

cmb69:
Note that I just grepped for the state comments, and removed those which do no longer apply. Almost certainly there are extensions which would need to be tagged, but maybe this PR is a good start to tackle https://github.com/php/doc-en/issues/3643.

### DIFF
--- a/appendices/extensions.xml
+++ b/appendices/extensions.xml
@@ -423,12 +423,7 @@
    &extcat.state.experimental;
 
    <itemizedlist>
-    <listitem><para><xref linkend="book.gnupg"/></para></listitem>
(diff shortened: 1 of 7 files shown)


jimwins (approved):
LGTM

jimwins (commented):
Shouldn't xmlrpc end up in the deprecated list since it was also marked that way?

jimwins:
Maybe the `doc-en` CI should be running the QA scripts like `doc-base` does so this would have been caught, too.

jimwins:
Don't mind me, I got myself confused. xmlrpc does still get listed as deprecated, the QA script is run in CI, and the second State being removed was being ignored, anyways.

PR #3645, opened 2024-08-07
Update extensions lists by status (fixes #3643)

### LINKED ISSUE #3643: Potential deprecated list of "Experimental Extensions"

jimwins:
Includes fixing invalid states for gnupg and svn

### DIFF
--- a/appendices/extensions.xml
+++ b/appendices/extensions.xml
@@ -413,7 +413,9 @@
    &extcat.state.deprecated;
 
    <itemizedlist>
-    <listitem><para/></listitem>
+    <listitem><para><xref linkend="book.mysql"/></para></listitem>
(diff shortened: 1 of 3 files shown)


cmb69:
See also #3644; I think for svn and gnupg we can drop the beta status (while the former is currently beta, it had stable releases in the past).

cmb69:
But hey, get this in, so we can see whether php/doc-base#145 gets green.

PR #3647, opened 2024-08-10
Clarify glob() issues raised by user notes and add another example

jimwins:


reference/filesystem/functions/glob.xml
```diff
@@ -80,14 +87,9 @@
   &reftitle.returnvalues;
   <para>
    Returns an array containing the matched files/directories, an empty array
-   if no file matched or &false; on error.
+   if no file matched or &false; on error. Unless the <literal>GLOB_NOSORT</literal>
+   was used, the names will be sorted alphanumerically.
```
Girgias, line 91:
```suggestion
   if no file matched or &false; on error.
   Unless <constant>GLOB_NOSORT</constant> was used,
   the names will be sorted alphanumerically.
```

Girgias (approved):
Minor nit, but LGTM

PR #3650, opened 2024-08-11
Maximum length of prefix in session_create_id

jorgsowa:
Maximum length of prefix is 256 characters.

Reference:
https://github.com/php/php-src/blob/master/ext/session/session.c#L379
https://github.com/php/php-src/blob/master/ext/session/session.c#L75

### DIFF
--- a/reference/session/functions/session-create-id.xml
+++ b/reference/session/functions/session-create-id.xml
@@ -41,7 +41,7 @@
         is prefixed by <parameter>prefix</parameter>. Not all
         characters are allowed within the session id.  Characters in
         the range <literal>a-z A-Z 0-9 , (comma) and -
-        (minus)</literal> are allowed.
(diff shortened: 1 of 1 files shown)


cmb69:
> Maximum length of prefix is 256 characters.

Right. Thanks for the PR!

Unfortunately, the code is not clear that it is about the *prefix* here (and elsewhere about the length of the actual SID without prefix), but we can't easily change this, because `php_session_valid_key()` is public.

And the error message can be confusing:

````php
var_dump(session_create_id(str_repeat("a", 257)));
````

````
Warning: session_create_id(): Prefix cannot contain special characters. Only aphanumeric, ',', '-' are allowed in …
bool(false)
````

jorgsowa:
Thank you. I created related PR for php-src also: https://github.com/php/php-src/pull/15338

PR #3651, opened 2024-08-11
session-create-id.xml: preg_* functions style for range

mmalferov:
Can we specify a range in the pcre-style, without introductory constructions in the sentence?


kamil-tekiela (approved):
I agree. PHP developers should be familiar with this syntax and it's more clear than before.

PR #3655, opened 2024-08-13
session-save-path.xml: add the `literal` tag to the file systems

mmalferov:


reference/session/functions/session-save-path.xml
```diff
@@ -35,7 +35,7 @@
         <para>
          On some operating systems, you may want to specify a path on a
          filesystem that handles lots of small files efficiently. For example,
-         on Linux, reiserfs may provide better performance than ext2fs.
+         on Linux, <literal>reiserfs</literal> may provide better performance than <literal>ext2fs</literal>.
```
cmb69, line 38: Both reiserfs and ext2fs have been deprecated; until we can come up with contemporary examples, I suggest to remove this.

mmalferov: Obsolete wording removed

cmb69:
Thank you!

PR #3659, opened 2024-08-16
Remove the top-level <copyright>, redundant with the <legalnotice>

jimwins:
I think this was mentioned in the discussion around #3468, the `<copyright>` block is redundant with the information in the `<legalnotice>` which is the first thing in the ToC, so we can just remove it. Will just leave a lonely `pubdate` before the ToC, but we could look at styling that better via CSS on the website, at least. Or maybe just turn it into a `<date>` within `manual.xml.in` so it ends up after the ToC along with a language snippet that actually identifies it as the date the manual was generated.

### DIFF
--- a/bookinfo.xml
+++ b/bookinfo.xml
@@ -6,11 +6,6 @@
   <pubdate><?dbtimestamp format="Y-m-d"?></pubdate>
   &frontpage.editors;
 
-  <copyright>
-   <year>1997-<?dbtimestamp format="Y"?></year>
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you!

> Or maybe just turn it into a `<date>` within `manual.xml.in` so it ends up after the ToC along with a language snippet that actually identifies it as the date the manual was generated.

That might be the best option; I don't have a strong opinion on what to do with the generation date, though.

PR #3660, opened 2024-08-18
gc.xml objects are passed by references by default

mmalferov:
This is not completely true:

https://www.php.net/manual/en/language.oop5.references.php


kamil-tekiela (approved):
Thanks. Although, it would be nice to have a different name for this.

PR #3661, opened 2024-08-18
New `&Construtor;` entity to avoid translatable text in parts of XML source

alfsb:
This make the word "Constructor" automatically translated in `<classsynopsisinfo>`, as is already done in most other `<classsynopsisinfo>` texts.

### DIFF
--- a/language-defs.ent
+++ b/language-defs.ent
@@ -78,6 +78,7 @@
 <!ENTITY Properties        "Properties">
 <!ENTITY InheritedConstants "Inherited constants">
 <!ENTITY Constants         "Constants">
+<!ENTITY Constructor       "Constructor">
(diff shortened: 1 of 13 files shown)


cmb69:
I'm not sure if that makes sense. If the classsynopsis would be generated by gen_stub.php, the comment would go away. So why not remove it?

alfsb:
This can be pushed as `[skip-revcheck]`, to not disturb already translated files.

cmb69 (approved):
The cmark classsynopses are not generated by gen-stub.php, because the extension has no stub files. If it had, we wouldn't need to have this discussion, but anyway, I'm fine with this PR.

Girgias (approved):
I agree with the logic here.

PR #3664, opened 2024-08-19
mysql_xdevapi/Collection::remove() - Match comment to code intention

leonardolara:
'age desc' with limit(1) will return the record with the highest age, not the lowest


cmb69:
Thank you!

PR #3668, opened 2024-08-21
Change description of MYSQLI_TYPE_INTERVAL

kamil-tekiela:


reference/mysqli/constants.xml
```diff
@@ -673,7 +673,7 @@
     </term>
     <listitem>
      <para>
-      Field is defined as <literal>INTERVAL</literal>
+      Alias of <constant>MYSQLI_TYPE_ENUM</constant>
```
cmb69, line 676: Thank you! Maybe add "(removed as of PHP 8.4.0)" right away?

reference/mysqli/constants.xml
```diff
@@ -673,7 +673,7 @@
     </term>
     <listitem>
      <para>
-      Alias of <constant>MYSQLI_TYPE_ENUM</constant>
+      Alias of <constant>MYSQLI_TYPE_ENUM</constant> (removed as of PHP 8.4.0)
```
cmb69, line 676: Oh, should have checked this earlier, but according to some other constants on this page this should probably be:
```suggestion
      Alias of <constant>MYSQLI_TYPE_ENUM</constant>. Removed as of PHP 8.4.0.
```

Unrelated to this PR, but still worth checking: there are two constants on this page which say "MySQL 5.0.3 and up". Do we still support MySQL 5.0, or even older versions?

cmb69 (approved):
Thank you!

PR #3671, opened 2024-08-21
Add missing cURL constants

haszi:
This PR adds the last few missing constants from https://github.com/php/doc-en/issues/2747. All constant descriptions and cURL version availability is based on the [cURL Symbols page](https://curl.se/libcurl/c/symbols-in-versions.html).

As the last commit sorts the constants which throws off the diff, this PR is definitely best reviewed commit by commit.

As a side note, I used [this script](https://gist.github.com/haszi/bca07054a7d4504d773d8d7fdfcffe2f) to sort the constants which may or may not be useful for doing the same in the translations.

### DIFF
--- a/reference/curl/constants.xml
+++ b/reference/curl/constants.xml
@@ -1828,6 +1828,460 @@
     </simpara>
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.curl-fnmatchfunc-fail">
+   <term>
+    <constant>CURL_FNMATCHFUNC_FAIL</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Returned by the wildcard match callback function if an error occured.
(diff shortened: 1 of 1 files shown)


haszi:
@Girgias Could you take a look at this PR as well?

cmb69 (approved):
Thank you very much! In my opinion this is good to be merged (should probably not be squashed-merged). While I see some potential for futher improvements (a few descriptions are missing, lots of unmarked acronyms, and other minor details), let's not delay adding useful information to manual. We can still improve later.

haszi:
> ...While I see some potential for futher improvements (a few descriptions are missing, lots of unmarked acronyms, and other minor details)...

I'm planning on doing this in a separate PR as there are a whole lot of constants in this file that need their descriptions entered and/or updated.

Girgias:
I had forgotten about this sorry

PR #3673, opened 2024-08-22
Use Makefile instead of docker-compose.yml

jimwins:
This allows us to build different formats more easily, and use `phd` and `doc-base` from parallel checkouts, which is much easier for issues that span the repositories.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -13,5 +13,7 @@ RUN git clone --depth 1 https://github.com/php/phd.git &&  \
 
 RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
 
+ENV FORMAT=xhtml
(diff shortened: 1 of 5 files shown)


Girgias:
The Docker file was introduced to make it easier for people to contribute.

I don't think removing it is sensible, but maybe @afilina doesn't mind as the original person adding it IIRC.

jimwins:
Sure, we can leave `docker-compose.yml` around. Maybe it's useful for someone doing docs development on Windows?

afilina:
I don't mind using `make` and ditch the docker-compose file. I wouldn't want the team to maintain several similar solutions.

If someone asks for a Windows option, we can always point them to the commands within the Makefile.

I vote in favor of this PR as-is.

Crell:
No objection here.

PR #3674, opened 2024-08-22
Fixed incomplete information in realpath() docs

php4fan:
Not only an empty string, but also the boolean value FALSE, when supplied as argument to `realpath()`, is interpreted as the current directory.

Feel free to rephrase.


cmb69:
Yeah, since `false` is coerced to an empty string in coercive typing mode (i.e. the default). This is already explained on other page, see e.g. https://www.php.net/manual/en/language.types.string.php#language.types.string.casting.

PR #3677, opened 2024-08-23
Add information about Composer and example of using its autoloader

jimwins:
Long past time that we included information about Composer in the documentation. Related PR for `doc-base` incoming.

### DIFF
--- a/install/composer.xml
+++ b/install/composer.xml
@@ -30,12 +30,11 @@
    to the code. The <link xlink:href="&url.composer;/doc/01-basic-usage.md">Composer
    "Basic Usage" documentation</link> goes into this in more depth.
   </simpara>
-  <para>
-   <example>
-    <title>
-     <literal>composer.json</literal> that requires a single package
-    </title>
-    <programlisting role="javascript">
+  <example>
+   <title>
+    <literal>composer.json</literal> that requires a single package
+   </title>
+   <programlisting role="javascript">
 <![CDATA[
 {
     "require": {
@@ -43,9 +42,8 @@
     }
 }
 ]]>
-    </programlisting>
-   </example>
-  </para>
+   </programlisting>
+  </example>
 
  </sect1>
 </chapter>
--- a/install/composer.xml
+++ b/install/composer.xml
@@ -17,8 +17,9 @@
   </simpara>
   <simpara>
    For example, if you are writing a PHP application or website that needs
-   to work with <abbrev>UUID</abbrev> values, you might want to use Ben
-   Ramsey's <literal>ramsey/uuid</literal> package that implements the
+   to work with <abbrev>UUID</abbrev> values, you might want to use
+   <link xlink:href="&url.packagist.package;ramsey/uuid">Ben Ramsey's
+   <literal>ramsey/uuid</literal> package</link> that implements the
    widely known and used types of UUIDs that are defined by
    <link xlink:href="&url.rfc;4122">RFC 4122</link>.
   </simpara>
--- a/install/composer.xml
+++ b/install/composer.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+
+<chapter xml:id="install.composer" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <title>Installation of Composer and third-party packages</title>
+
+ <sect1 xml:id="install.composer.intro">
+  <title>Introduction to Composer</title>
+  <simpara>
+   &link.composer; is a dependency manager for PHP that makes it possible
+   to define third-party packages of code that your project uses that can
+   then be easily installed and updated. It leverages the the built-in
+   <link linkend="language.oop5.autoload">class autoloading features</link>
+   of PHP, repositories of PHP packages such as
+   <link xlink:href="&url.packagist;">Packagist</link>, and common project
+   layout and coding conventions.
+  </simpara>
+  <simpara>
+   For example, if you are writing a PHP application or website that needs
+   to work with <abbrev>UUID</abbrev> values, you might want to use Ben
+   Ramsey's <literal>ramsey/uuid</literal> package that implements the
+   widely known and used types of UUIDs that are defined by
+   <link xlink:href="&url.rfc;4122">RFC 4122</link>.
+  </simpara>
+  <simpara>
+   Briefly, you would do this by creating a <literal>composer.json</literal>
+   in your project, using Composer to install the latest version of the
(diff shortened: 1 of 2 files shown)


cmb69:
A bold move, Jim!

I'm somewhat concerned about opening a can of worms when merging this; I consider it quite likely that there quickly will be lots of PRs suggesting to add examples or even documentation about some packages, and then tooling, frameworks, IDEs, etc. Maybe we should consider defining the scope first; maybe we should handle this on a case by case basis; maybe we should instead link to some external sites (e.g. https://phptherightway.com/); maybe we should try to join forces?

In any way, we shouldn't introduce personalization into the PHP manual (e.g. no "you", but instead using passive voice).

jimwins:
I think possibly having the problem of fending off too many contributions to the manual is vastly preferable to the existing problem of not acknowledging major chunks of the PHP ecosystem in the name of some sort of neutrality.

More here: https://trainedmonkey.com/2024/08/23/writing_documentation_in_anger

The PHP project is where we gather to join forces, I don't think these fundamentals should be left to external sites.

cmb69:
I agree (also with your blog post), but we should properly discuss this, since there has been [big drama a couple of years ago](https://github.com/php/web-bugs/pull/77), and I would not like to see this happening again. Maybe there should even be an RFC about "endorsing third-party tools" (or so), to settle this topic once and for all.

cc @Crell

Crell:
I completely agree with adding Composer references to the docs, for all the reasons @jimwins mentions.  However, I also agree that we should probably not call out `ramsey/uuid` in particular as an example.  Nothing against Ben, but that part is a much more slippery slope than composer (which has basically zero competition).  A `foo/bar` package or similar is probably better.

As far as proper discussion, the most straightforward way would be an RFC "is it OK to talk about Composer in the docs? Y/N".  (We did a similar RFC recently for linking to the Foundation site, which passed easily.)  It should be a pretty easy RFC, and would quash any squeamishness about if it's OK or not.

jimwins:
Thanks for linking to that other issue, that fills in a historical event I knew was out there but hadn't run across. As it happens, I have another related blog post:

https://trainedmonkey.com/2024/07/27/the_rules_can_matter

Here's a draft RFC that is broader than just asking for permission to mention Composer:

https://wiki.php.net/rfc/web-and-doc-use-not-endorsement

cmb69:
> RFC "is it OK to talk about Composer in the docs? Y/N"

Possible, but rather limited in scope, as follow-ups will surely come; from Jim's blog post:

> I would love to add a chapter on static analysis tools. And another one about linting and refactoring tools. Maybe a chapter on frameworks.

And of course, there is the issue about actually using it. I think we're using Composer and a few packages already in different Web repos, but still not in some repos which could benefit from some tooling (see e.g. https://github.com/php/php-sdk-binary-tools/pull/22 or https://github.com/php/php-sdk-binary-tools/commit/78c9f7b6d71b51aa8174da1470f654bd5c612232, the latter being an issue detected with static analysis; or even https://externals.io/message/124843#124866).

jrfnl:
> ... much more slippery slope than composer (which has basically zero competition). 

Nothing against Composer, but in that case, what about PHIVE ? And shouldn't the history with PEAR and its installer be mentioned ?

I do see the slippery slope here, even in the mention of Composer.

derickr:
I had not even heard of PHIVE. And PEAR is probably going to be mothballed into a static site.

I've written a more general thought on all of this: https://news-web.php.net/php.internals/125300

> I completely agree with adding Composer references to the docs, for all the reasons @jimwins mentions. However, I also agree that we should probably not call out `ramsey/uuid` in particular as an example. Nothing against Ben, but that part is a much more slippery slope than composer (which has basically zero competition). A `foo/bar` package or similar is probably better.

Using foo/bar doesn't have any meaning for users. *real* packages do. Picking an example from a community member as opposed to a large opinionated frameworks makes perfect sense ti me.

cmb69:
> I had not even heard of PHIVE.

Its main developer even has a php.net account (apparently for as long as you do). :)

> I do see the slippery slope here, even in the mention of Composer.

I agree. It starts with Composer. Then someone asks "what does a composer do?" Right, he's writing a Symfony. And who would sing the overture? Lara Vel, of course! And it ain't over till the slim lady sings. ;)

oladoyinbov:
It would be better to create a section on the PHP website homepage (at the footer section) to list out the most popular PHP frameworks and packages, you can check python.org for a good sample of what I am talking about.

Frameworks like Laravel, Symfony, CodeIgniter, CakePHP and Package Managers like Composer should be added to help beginners and the less informed access the PHP ecosystem more quickly.

Girgias:
> It would be better to create a section on the PHP website homepage (at the footer section) to list out the most popular PHP frameworks and packages, you can check python.org for a good sample of what I am talking about.
> 
> Frameworks like Laravel, Symfony, CodeIgniter, CakePHP and Package Managers like Composer should be added to help beginners and the less informed access the PHP ecosystem more quickly.

This is already being discussed on the Internals ML.

install/composer.xml
```diff
@@ -0,0 +1,51 @@
...
+<![CDATA[
+{
+    "require": {
+        "ramsey/uuid": "^4.7"
+    }
+}
+]]>
+    </programlisting>
+   </example>
+  </para>
```
Girgias, line 48: This wrapping para tag is unnecessary

language/oop5/autoload.xml
```diff
@@ -79,6 +79,26 @@ string(5) "ITest"
...
+    </programlisting>
+   </example>
+   <example>
+    <title>Using Composer's autoloader</title>
+    <para>
+     &link.composer; generates a <literal>vendor/autoload.php</literal>
+     with is set up to automatically load packages that are being managed
+     by Composer. By including this file, those packages can be used without
+     any additional work.
+    </para>
```
Girgias, line 92:
```suggestion
    <simpara>
     &link.composer; generates a <literal>vendor/autoload.php</literal>
     with is set up to automatically load packages that are being managed
     by Composer. By including this file, those packages can be used without
     any additional work.
    </simpara>
```

Girgias (changes requested):
Final XML remarks

Girgias (approved):
LGTM

jimwins:
I'll hold off on pushing this and the accompanying `doc-base` change until at least Monday in case anyone else wants to chime in.

theodorejb (commented):
Fix duplicate "the" and improve wording.

Crell:
I'm fine with @theodorejb's wording change.  Otherwise, shipit.  (Once tests pass.)

jimwins:
Integration test won't pass in CI because it requires simultaneous changes from php/doc-base#147

PR #3678, opened 2024-08-23
declare.xml: add the missing dot char and amend case of sentences

mmalferov:


language/control-structures/declare.xml
```diff
@@ -24,16 +24,16 @@ declare (directive)
...
+  directive), the <literal>encoding</literal> directive (see below for more
   information on the
   <link linkend="control-structures.declare.encoding">encoding</link>
-  directive) and the <literal>strict_types</literal> directive (See for more
+  directive) and the <literal>strict_types</literal> directive (see for more
   information the
   <link linkend="language.types.declarations.strict">strict typing</link>
-  section on the type declarations page)
+  section on the type declarations page).
  </para>
```
Girgias, line 37: Maybe it would be better to use a `<simplelist>` naming the directives with every item having a link, instead of repeating the same text multiple times?

mmalferov: Done!

Girgias:
Thank you!

PR #3680, opened 2024-08-24
Fix type juggling notes for list() and array destructuring

### LINKED ISSUE #3640: Destructuring assignment from null does not raise warning
https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.destructuring says 

> Attempting to access an array key which has not been defined is the same as accessing any other undefined variable: an [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning)-level error message ([E_NOTICE](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-notice)-level prior to PHP 8.0.0) will be issued, and the result will be [null](https://www.php.net/manual/en/reserved.constants.php#constant.null).

but this isn't actually true when assigning null; `[$x, $y] = null;` does not raise a notice or warning ([eval](https://3v4l.org/QbKjjO)). Same goes for destructuring with an associative array ([eval](https://3v4l.org/7YVY6)). This seems to be specific to null (e.g. `[$x, $y] = [];` raises a warning: [eval](https://3v4l.org/I9q9X)) and to destructuring assignment (e.g. `null[0]` raises a warning: [eval](https://3v4l.org/JkHnO)).

Not sure if this is a bug or a documentation issue, but either there should be a warning or this special case should be documented.

tgr:
Update notes about the handling of missing keys and non-arrays during array destructuring to match the behavior seen on 3v4l.org, which doesn't issue a warning or notice in some cases where the documentation claims it does. Also update list() docs to provide the same information.

list():
* missing key: https://3v4l.org/bYBt6
* destructuring an integer: https://3v4l.org/7lNhK
* a boolean: https://3v4l.org/p5LDd
* a string: https://3v4l.org/KHSAH
* null: https://3v4l.org/lk1iu

array notation:
* missing key: https://3v4l.org/KCiTC
* destructuring an integer: https://3v4l.org/louVG
* a boolean: https://3v4l.org/SWL5L
* a string: https://3v4l.org/aK90L
* null: https://3v4l.org/Y8cOY

Fixes #3640.

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -734,6 +734,11 @@ echo $b, PHP_EOL;    // prints 1
       issued, and the result will be &null;.
     </para>
    </note>
+   <note>
+    <para>
+      Destructuring a scalar value assigns &null; to all variables.
+    </para>
+   </note>
   </sect3>
 
  </sect2><!-- end syntax -->
--- a/reference/array/functions/list.xml
+++ b/reference/array/functions/list.xml
@@ -16,8 +16,8 @@
    Like <function>array</function>, this is not really a function,
    but a language construct. <function>list</function> is used to
    assign a list of variables in one operation.
-   Strings can not be unpacked and <function>list</function> expressions 
-   can not be completely empty.
+   Only arrays and objects that implement <link linkend="class.arrayaccess">ArrayAccess</link> can be unpacked.
+   <function>list</function> expressions can not be completely empty.
   </para>
   <note>
    <para>
@@ -30,6 +30,19 @@
    destructuring of arrays with non-integer or non-sequential keys. For more details on
    array destructuring, see the <link linkend="language.types.array.syntax.destructuring">array destructuring section</link>.
   </para>
+  <note>
+   <para>
+    Attempting to access an array key which has not been defined is
+    the same as accessing any other undefined variable:
+    an <constant>E_WARNING</constant>-level error message
+    (<constant>E_NOTICE</constant>-level prior to PHP 8.0.0) will be
+    issued, and the result will be &null;.
+   </para>
+   <para>
+    Attempting to unpack a scalar assigns &null; to all variables.
+    Attempting to unpack an object that does not implement ArrayAccess is a fatal error.
+   </para>
+  </note>
  </refsect1>
  <refsect1 role="parameters">
   &reftitle.parameters;


language/types/array.xml
```diff
@@ -391,10 +391,7 @@ $secondElement = getArray()[1];
    </note>
    <note>
     <para>
-     Array dereferencing a scalar value which is not a <type>string</type>
-     yields &null;. Prior to PHP 7.4.0, that did not issue an error message.
-     As of PHP 7.4.0, this issues <constant>E_NOTICE</constant>;
-     as of PHP 8.0.0, this issues <constant>E_WARNING</constant>.
```
MatmaRex, line 397: This was about array dereferencing (`$a = 42; $a[0]`) not destructuring (`[$a] = 42;`), and I think it was correct. Only destructuring weirdly yields null with no warning.

reference/array/functions/list.xml
```diff
@@ -25,6 +25,20 @@
...
+    Attempting to access an array key which has not been defined is
+    the same as accessing any other undefined variable:
+    an <constant>E_WARNING</constant>-level error message
+    (<constant>E_NOTICE</constant>-level prior to PHP 8.0.0) will be
+    issued, and the result will be &null;.
+   </para>
+  </note>
+  <note>
+   <para>
+    Attempting to access element of a scalar value yields &null;.
```
MatmaRex, line 39: I would say: "Attempting to unpack a scalar value assigns null to all variables." to be consistent with the existing wording.

There's also already a note above: "Strings can not be unpacked" which should probably say "Only arrays and objects that implement ArrayAccess can be unpacked".

And an example below, where the comment could be clarified:
```
// list() doesn't work with strings
list($bar) = "abcde";
var_dump($bar); // NULL
```

language/types/array.xml
```diff
@@ -391,10 +391,7 @@ $secondElement = getArray()[1];
...
+     Dereferencing a scalar value ields &null;.
```
DanielEScherzer:
```suggestion
     Dereferencing a scalar value yields &null;.
```

DanielEScherzer (changes requested):
Sorry it has taken us so long to review this; thanks for the patch, one small issue

MatmaRex:
This probably needs an update now for changes in PHP 8.5: https://wiki.php.net/rfc/warnings-php-8-5#destructuring_non-array_values

language/types/array.xml
tgr: Moved the note to the right place.

reference/array/functions/list.xml
tgr: Updated the wording and the other note. IMO documenting behavior when misused is not worth the space for adding an example.

language/types/array.xml
tgr: Fixed.

tgr:
Sorry, I forgot about this PR completely. Thanks for the feedback, rebased and updated.

> This probably needs an update now for changes in PHP 8.5: https://wiki.php.net/rfc/warnings-php-8-5#destructuring_non-array_values

I think the docs generally don't include details about not-yet-released PHP versions?

DanielEScherzer:
> Sorry, I forgot about this PR completely. Thanks for the feedback, rebased and updated.
> 
> > This probably needs an update now for changes in PHP 8.5: https://wiki.php.net/rfc/warnings-php-8-5#destructuring_non-array_values
> 
> I think the docs generally don't include details about not-yet-released PHP versions?

Yeah, we don't need to start worrying about that yet

PR #3681, opened 2024-08-24
Remove incorrect migration entry

nielsdos:


cmb69:
We should probably fix https://github.com/php/php-src/blob/8da94cfc41df84c44703c1ef725336efb15d14bb/UPGRADING#L398, too.

nielsdos:
Done via https://github.com/php/php-src/commit/cd6bac7f9f592798ffe0b940d8b7245f0ac25c96

PR #3684, opened 2024-08-26
Update array_merge() documentation: clarify behavior with non-array arguments in PHP 7.4 and 8.x

daniellz:
This pull request updates the documentation for the array_merge() function to include details about how the function behaves when non-array arguments are passed. It clarifies that:
•	In PHP 7.4, passing non-array arguments triggers an E_WARNING and returns null.
•	In PHP 8.0 and later, passing non-array arguments results in a Fatal error.

### DIFF
--- a/reference/array/functions/array-diff.xml
+++ b/reference/array/functions/array-diff.xml
@@ -156,6 +156,10 @@ $result = array_diff($source, $filter);
     array. Of course you can check deeper dimensions by using
     <literal>array_diff($array1[0], $array2[0]);</literal>.
    </para>
+   <para>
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you for the PR!

However, this is the normal behavior of all internal functions, and as such is not supposed to be documented for each single function (there are thousands of them). From the [documentation](https://www.php.net/manual/en/functions.internal.php):

> If the parameters given to a function are not what it expects, such as passing an [array](https://www.php.net/manual/en/language.types.array.php) where a [string](https://www.php.net/manual/en/language.types.string.php) is expected, the return value of the function is undefined. In this case it will likely return [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) but this is just a convention, and cannot be relied upon. As of PHP 8.0.0, a [TypeError](https://www.php.net/manual/en/class.typeerror.php) exception is supposed to be thrown in this case.

PR #3685, opened 2024-08-26
constants.xml: add the missing plus character, + → | and the `replaceable` tag

mmalferov:


reference/info/constants.xml
```diff
@@ -95,7 +95,7 @@
     <simpara>
      All the credits, equivalent to using: <literal xmlns="http://docbook.org/ns/docbook">CREDITS_DOCS +
```
cmb69, line 96: There is no need for the namespace definition here:
```suggestion
     All the credits, equivalent to using: <literal>CREDITS_DOCS +
```

cmb69: And actually I think that should better be `|` instead of `+` signs.

reference/info/constants.xml
```diff
@@ -298,7 +298,7 @@
    </term>
    <listitem>
     <para>
-     Issues a PHP warning for each failed assertion
+     Throws a PHP exception for each failed assertion
```
cmb69, line 301:
```suggestion
     Throws an <classname>AssertionError</classname> for each failed assertion
```

reference/info/constants.xml
```diff
@@ -428,8 +428,10 @@
    </term>
    <listitem>
     <simpara>
-     This contains the value used to determine the <literal xmlns="http://docbook.org/ns/docbook">PHP_WINDOWS_NT_*</literal>
-     constants. This value may be one of the <literal xmlns="http://docbook.org/ns/docbook">PHP_WINDOWS_NT_*</literal> constants
+     This contains the value used to determine the
+     <literal xmlns="http://docbook.org/ns/docbook">PHP_WINDOWS_NT_<replaceable>*</replaceable></literal>
+     constants. This value may be one of the
+     <literal xmlns="http://docbook.org/ns/docbook">PHP_WINDOWS_NT_<replaceable>*</replaceable></literal> constants
```
cmb69, line 434: Sorry, overlooked this so far:
```suggestion
     This contains the value used to determine the
     <literal>PHP_WINDOWS_NT_<replaceable>*</replaceable></literal>
     constants. This value may be one of the
     <literal>PHP_WINDOWS_NT_<replaceable>*</replaceable></literal> constants
```

Feel free to reformat that paragraph, if you like.

PR #3672, opened 2024-08-22
Normalize version strings in changelog entries

jimwins:
Now they're always `N.X.Y` for a PHP version, or `PECL extension N.X.Y` for PECL extensions.

PR #3687, opened 2024-08-26
Clarify second param of Memcache::delete and how it was changed

jimwins:
It never should have been a parameter, basically, and was never a timeout. Just a wonky edge of the underlying interface.

Was originally noticed as part of #3672, split out to be reviewed/merged on its own.


Girgias (approved):
LGTM

PR #3695, opened 2024-08-28
COM variant_fix() and variant_int(): differences are now clear in MSDN

leonardolara:
There is a difference between variant_fix() and variant_int() in MSDN docs.
See [VarInt function](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-varint) and  [VarFix function](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-varfix).
The note was removed from variant_fix() and the logic was fixed in variant_int().

### DIFF
--- a/reference/com/functions/variant-fix.xml
+++ b/reference/com/functions/variant-fix.xml
@@ -48,17 +48,6 @@
   </para>
  </refsect1>
 
- <refsect1 role="notes">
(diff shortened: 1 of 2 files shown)


cmb69:
Nice catch. Thank you!

PR #3697, opened 2024-08-29
cubrid_affected_rows(): fix example output

leonardolara:


cmb69:
Thank you! That looks reasonable.

PR #3699, opened 2024-08-29
cubrid-data-seek.xml: remove parameter tag

leonardolara:
The "row number" in the sentence is not a direct reference to the function parameter row_number.


cmb69:
Thank you!

Issue #3702, opened 2024-08-30
get_debug_type missing documentation of behavior for anonymous classes extending another class or implementing an interface


patrickkusebauch:
From manual page: https://php.net/function.get-debug-type

This omission caused a bug (by me) in PHPStan dynamic return type function implementation: https://github.com/phpstan/phpstan/issues/11562

patrickkusebauch:
According to the commit implementing the function (https://github.com/php/php-src/commit/ef0e4478c51540510b67f7781ad240f5e0592ee4) the extending class or implementing interface should be replace the part before `@anonymous`.

PR #3703, opened 2024-08-30
Update get-debug-type.xml

patrickkusebauch:
Closes: https://github.com/php/doc-en/issues/3702

Adds more examples to `get_debug_type` output in cases of anonymous classes that extend another class and/or implement an interface.

### DIFF
--- a/reference/var/functions/get-debug-type.xml
+++ b/reference/var/functions/get-debug-type.xml
@@ -121,7 +121,7 @@
         <row>
          <entry>Objects from Anonymous Classes</entry>
          <entry>
-          <literal>"class@anonymous"</literal>
+          <literal>"class@anonymous"</literal> or parent class name/interface name if the class extends another class or implements an interface e.g. <literal>"Foo\Bar@anonymous"</literal>
          </entry>
          <entry>
           Anonymous classes are those created through the $x = new class { ... } syntax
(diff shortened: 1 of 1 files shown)


cmb69 (approved):
Thank you!

reference/var/functions/get-debug-type.xml
```diff
@@ -121,7 +121,7 @@
         <row>
          <entry>Objects from Anonymous Classes</entry>
          <entry>
-          <literal>"class@anonymous"</literal>
+          <literal>"class@anonymous"</literal> or parent class name/interface name if the class extends another class or implements an interface e.g. <literal>"Foo\Bar@anonymous"</literal>
```
Girgias, line 124: This is only a recent change and would require a changelog somewhere.

I think this might have been in 8.1?

patrickkusebauch: According to https://onlinephp.io/c/a0758 it was always that way.

cmb69: https://3v4l.org/hg9FA confirms that.

patrickkusebauch: Also, the test did not change since the beginning: https://github.com/php/php-src/blame/master/ext/standard/tests/general_functions/get_debug_type_basic.phpt

Girgias: Then I'm confusing with something else that was changed recently.

PR #3704, opened 2024-08-31
bugfix docs example

abdelrahman-gado:
change docs example from creating an object of Ds\Sequence which is an interface to Ds\Vector
[Docs Example link ](https://www.php.net/manual/en/ds-sequence.apply.php#refsect1-ds-sequence.apply-examples)


cmb69:
Thank you!

PR #3705, opened 2024-09-01
date.xml replace the `DateTime::format` method by interface method

mmalferov:
The same as: https://github.com/php/doc-en/pull/3693


cmb69:
Thank you!

PR #3716, opened 2024-09-05
integer.xml: wrap NaN, Inf and -Inf with the `literal` tag

mmalferov:
There are two infinities: positive and negative. I think we should mention both, not just the positive one ;-)


cmb69:
Thank you!

PR #3720, opened 2024-09-06
array-map.xml: add an explanation of the 'zip operation' term

mmalferov:
Why not add a few words about what a 'zip operation' is? Yes, the `array_map` page contains a sample code, but it does not contain a formal explanation, as far as all other doc pages. I suppose this information will be useful :-)

### DIFF
--- a/reference/array/functions/array-map.xml
+++ b/reference/array/functions/array-map.xml
@@ -40,7 +40,8 @@
       </para>
       <para>
        &null; can be passed as a value to <parameter>callback</parameter>
-       to perform a zip operation on multiple arrays.
+       to perform a zip operation on multiple arrays and return an array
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-map.xml
```diff
@@ -40,7 +40,8 @@
       </para>
       <para>
        &null; can be passed as a value to <parameter>callback</parameter>
-       to perform a zip operation on multiple arrays.
+       to perform a zip operation on multiple arrays and return an array
+       whose elements are each an array holding the elements of the input arrays of the same index.
```
cmb69, line 44: Nice! Maybe also point to the example:
```suggestion
       whose elements are each an array holding the elements of the input arrays of the same index (see example below).
```

PR #3721, opened 2024-09-06
Documentation updates for ext-mongodb 1.20

jmikola:
This PR will collect various changes for ext-mongodb 1.20 (see: [PHPC tickets with `documentation-needed` label](https://jira.mongodb.org/issues/?jql=project%20%3D%20PHPC%20AND%20labels%20%3D%20documentation-needed)).

 * [PHPC-1489](https://jira.mongodb.org/browse/PHPC-1489): Deprecate using ReadPreference integer constants
 * [PHPC-2347](https://jira.mongodb.org/browse/PHPC-2347): Deprecate BSON functions in favor of Document methods
 * [PHPC-2349](https://jira.mongodb.org/browse/PHPC-2349): Deprecate WriteException class
 * [PHPC-2411](https://jira.mongodb.org/browse/PHPC-2411): Deprecate SSLConnectionException class

### DIFF
--- a/reference/mongodb/functions/bson/fromjson.xml
+++ b/reference/mongodb/functions/bson/fromjson.xml
@@ -7,6 +7,16 @@
   <refpurpose>Returns the BSON representation of a JSON value</refpurpose>
  </refnamediv>
 
+ <refsynopsisdiv>
+  <warning>
+   <para>
+    This function has been <emphasis>DEPRECATED</emphasis> as of extension
+    version 1.20.0 and will be removed in 2.0. Applications should use
+    <methodname>MongoDB\BSON\Document::fromJSON</methodname> instead.
+   </para>
+  </warning>
+ </refsynopsisdiv>
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
--- a/reference/mongodb/mongodb/driver/exception/writeexception.xml
+++ b/reference/mongodb/mongodb/driver/exception/writeexception.xml
@@ -8,6 +8,15 @@
 
  <partintro>
 
+  <warning>
+   <simpara>
+    This exception class is <emphasis>DEPRECATED</emphasis> as of extension
+    version 1.20.0 and will be removed in 2.0. This exception was never directly
+    thrown by the extension. Applications should use
+    <classname>MongoDB\Driver\Exception\BulkWriteException</classname> instead.
+   </simpara>
+  </warning>
+
 <!-- {{{ MongoDB\Driver\Exception\WriteException intro -->
   <section xml:id="mongodb-driver-exception-writeexception.intro">
    &reftitle.intro;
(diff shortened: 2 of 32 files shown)


reference/mongodb/mongodb/driver/exception/writeexception.xml
```diff
@@ -3,11 +3,20 @@
 
 <reference xml:id="class.mongodb-driver-exception-writeexception" role="class" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
 
- <title>The MongoDB\Driver\Exception\WriteException class</title>
+ <title>The MongoDB\Driver\Exception\WriteException class (deprecated)</title>
```
cmb69, line 6: You can mark deprecated entities in https://github.com/php/doc-en/blob/master/reference/mongodb/versions.xml; just add a `deprecated` attribute (see e.g. https://github.com/php/doc-en/blob/master/reference/uopz/versions.xml), and the entity will be shown in the sidebar as deprecated.

jmikola: Thanks, I actually just discovered that independently while working on the most recent commit for deprecated BSON functions. I've removed "(deprecated)" from existing page titles and will be using the `deprecated` attribute in `versions.xml` going forward.

While I have your attention: does the `removed` attribute in `versions.xml` have any effect? I see it used several times (albeit less frequently than `deprecated`), but I wasn't able to discern any impact to the rendered documentation.

cmb69: I'm not sure what's up with the `removed` attribute. As far as I know, it is completely ignored by tooling. Maybe it was supposed to have any functionality, but nobody implemented that so far, or maybe it is supposed to be informational to clarify that it had not been overlooked to update the `from` attribute (what appears to be quite sensible to me).

reference/mongodb/functions/bson/fromjson.xml
```diff
@@ -7,6 +7,16 @@
   <refpurpose>Returns the BSON representation of a JSON value</refpurpose>
  </refnamediv>
 
+ <refsynopsisdiv>
+  <warning>
+   <para>
+    This function has been <emphasis>DEPRECATED</emphasis> as of extension
+    version 1.20.0 and will be removed in 2.0. Applications should use
+    <methodname>MongoDB\BSON\Document::fromJSON</methodname> instead.
```
cmb69, line 15: It *may* make sense to use an XML entity reference here to avoid the repetition (see languages-snippets.ent). Have you considered something like
````xml
<!ENTITY mongodb.warn.deprecated.function-1-20-0.removed-2-0 'This function has been <emphasis xmlns="http://docbook.org/ns/docbook">DEPRECATED</emphasis> as of extension
    version 1.20.0 and will be removed in 2.0. Instead, applications should use '>
````

cmb69: Indeed, removing entities in language-snippets.ent can break translations (if they are using the entities, but have not updated there own language-snippets.ent yet). Checking this would likely be a manual task (about 10 active translations).

Removing entities from global.ent, on the other hand, is easier, since doc-base PRs run builds for all active languages in CI, so issues will be detected.

reference/mongodb/mongodb/driver/exception/writeexception.xml
Girgias: The long term objective is to use more information from the `versions.xml` file to generate the standard warning blocks instead of needing to manually edit source files.

PR #3728, opened 2024-09-08
apache2.xml: amend the XML markup

mmalferov:
I did not change the text, I only changed:

* Amend an XML markup (for the most part, the indentation were corrected)
* Removed the extra space characters (double spaces)
* Added tags: `command`, `literal`, `filename`

### DIFF
--- a/install/unix/apache2.xml
+++ b/install/unix/apache2.xml
@@ -356,7 +356,7 @@ service httpd restart
    To build a multithreaded version of Apache, the target system must support threads.
    In this case, PHP should also be built with
    Zend Thread Safety (ZTS). Under this configuration, not all extensions will be available.
-   The recommended setup is to build Apache with the default 
+   The recommended setup is to build Apache with the default
    <filename>prefork</filename> MPM-Module.
   </para>
  </note>
--- a/install/unix/apache2.xml
+++ b/install/unix/apache2.xml
@@ -115,17 +115,17 @@ make install
 
   <listitem>
    <para>
-    Now, configure and build PHP. This is where you customize PHP
+    Now, configure and build PHP.  This is where you customize PHP
     with various options, like which extensions will be enabled. Run
-    <command>./configure --help</command> for a list of available options. In our example
-    we'll do a simple configure with Apache 2 and MySQL support.
+    ./configure --help for a list of available options.  In our example
+    we'll do a simple configure with Apache 2 and MySQL support. 
    </para>
 
    <para>
     If you built Apache from source, as described above, the below example will
-    match your path for <command>apxs</command>, but if you installed Apache some other way, you'll
+    match your path for apxs, but if you installed Apache some other way, you'll
     need to adjust the path to apxs accordingly. Note that some distros may rename
-    <command>apxs</command> to <command>apxs2</command>.
+    apxs to apxs2.
    </para>
 
    <informalexample>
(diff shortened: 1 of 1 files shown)


Girgias:
Can you split the commits out so that the whitespace changes can be comitted as a `[skip-revhceck]` ?

mmalferov:
Don't get too mad at me, I seem to understand some of the words, but I don't fully understand what I have to do, and what `[skip-revcheck]` is. Obviously it's about skipping the revision check, but I don't understand where this check is performed, and why it depends on me [confusion emoji here]

It's easier for me to create a new patch and transfer the changes I made in the file there. install/unix/apache2.xml . Maybe I should do that?

cmb69:
> Obviously it's about skipping the revision check, but I don't understand where this check is performed, and why it depends on me

There is [revcheck.php](https://github.com/php/doc-base/blob/master/scripts/revcheck.php) which is used by some translators. If a commit is marked with `[skip-revcheck]`, the script ignores this commit, since it is not supposed to be relevant for translations.

Having a separate commit with only whitespace changes is unlikely to be relevant for translations, so can be marked with `[skip-revchec]`.

> It's easier for me to create a new patch and transfer the changes I made in the file there. install/unix/apache2.xml . Maybe I should do that?

Whatever is easier for you. The idea is just to separate commits which are relevant for translations from those which are not.

mmalferov:
Oh, that's it! Thanks for the explanation.

I hope I understood the purpose of the `[skip-revcheck]` tag correctly. Changes to the doc-en with such a tag will not affect translations in other languages, that is, they will not force other translators to make efforts to bring the language files into line with the files of the English version, in which indentation corrections were made.

Before I make changes, please answer one more question: is it not necessary to clean up the indentation in translation files in other languages? After all, it's one thing to break a sentence in another language not in the same place where the break appears in the English version, and another thing is indentation, which changes the structure of the document

cmb69:
> Before I make changes, please answer one more question: is it not necessary to clean up the indentation in translation files in other languages? After all, it's one thing to break a sentence in another language not in the same place where the break appears in the English version, and another thing is indentation, which changes the structure of the document

Well, it depends. Some translations may already use some other indentation deliberately, in which case they likely want to ignore respective doc-en changes. Some other translations may stick closely to the indentation in doc-en, and as such would not want to ignore such changes to doc-en.

In any way, it might be best to separate changes with different purpose into separate commits (not necessarily PRs, although when they are squash-merged, it's better to have separate PRs). Regardless of `[skip-revcheck]`, this can be useful for translators.

mmalferov:
Now the file has only whitespaces changes ;-) I removed added tags

cmb69:
Thank you!

mmalferov:
In Russian there is proverb: «За что боролись, на то и напоролись» (What we fought for and ran into it), about actions that led to a result directly opposite to the one that these actions were aimed at achieving :-)

The corrected indents went unnoticed by other translators because of the `[skip-revchek]` tag, but after adding `literal`, `command`, etc tags, in the translation tool, translators in other languages still see every change: both spaces and tags, because the addition of tags that occurred after the spaces became part of the last change ))

For Russian:

http://doc.php.net/revcheck.php?p=plain&lang=ru&hbp=b0dd627899cf5c57d3d64fac07e9b6579dfca082&f=install/unix/apache2.xml&c=on

or French:

http://doc.php.net/revcheck.php?p=plain&lang=fr&hbp=b0dd627899cf5c57d3d64fac07e9b6579dfca082&f=install/unix/apache2.xml&c=on

as example 👍

Girgias:
This seems like an issue in how we produce diffs on the revcheck script :/

PR #3731, opened 2024-09-09
reference/win32service: remove the extra space and add the missing dot cahrs

mmalferov:


reference/win32service/functions/win32-add-right-access-service.xml
```diff
@@ -38,15 +38,16 @@
...
       </para>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><parameter>right</parameter></term>
      <listitem>
       <para>
-       The rights allowed for <parameter>username</parameter>. Use <link linkend="win32service.constants.rights">constants</link> for define this value.
+       The rights allowed for <parameter>username</parameter>.
+       Use <link linkend="win32service.constants.rights">constants</link> for define this value.
```
cmb69, line 50: This sentence should be improved. Besides the minor grammar issue, read it without the `<link>` markup: "Use constants to define this value." So all in all it might better be something like:
```suggestion
       The <link linkend="win32service.constants.rights">service rights constants</link> can be used to define this value.
```

cmb69:
Thank you!

PR #3737, opened 2024-09-10
Remove an extra space characters

mmalferov:


sy-records (approved):

reference/ldap/functions/ldap-get-values.xml
```diff
@@ -105,8 +105,9 @@
   &reftitle.examples;
   <para>
    <example>
-    <title>List all values of the "mail" attribute for a
-     directory entry </title>
+    <title>
+     List all values of the "mail" attribute for a directory entry
+    </title>
```
cmb69, line 110: Maybe put this is one line:
```suggestion
    <title>List all values of the "mail" attribute for a directory entry</title>
```

reference/mysqlnd/plugin.xml
```diff
@@ -1393,8 +1390,7 @@ mysqlnd_plugin_set_conn_proxy(new proxy());
    methods and you can choose to have "pre" or "post" hooks.
   </para>
   <para>
-   <emphasis role="bold"> Buildin class:
-   mysqlnd_plugin_connection::connect() </emphasis>
+   <emphasis role="bold">Buildin class: mysqlnd_plugin_connection::connect()</emphasis>
```
cmb69, line 1393: Should be
```suggestion
   <emphasis role="bold">Built-in class: mysqlnd_plugin_connection::connect()</emphasis>
```

reference/image/functions/imagecopy.xml
```diff
@@ -21,9 +21,9 @@
   <para>
    Copy a part of <parameter>src_image</parameter> onto
    <parameter>dst_image</parameter> starting at the x,y coordinates
-   <parameter>src_x</parameter>, <parameter>src_y </parameter> with
+   <parameter>src_x</parameter>, <parameter>src_y</parameter> with
    a width of <parameter>src_width</parameter> and a height of
-   <parameter>src_height</parameter>.  The portion defined will be copied
+   <parameter>src_height</parameter>. The portion defined will be copied
```
cmb69, line 26: I'm fine with removing this extra space, but there is an (old) convention in English plain text to put two spaces after a full-stop. Unless there is already a coding style rule for this regarding the docs, we should not do more such changes.

cc @Crell

cmb69 (commented):
Thank you! This looks good besides the couple of nits below.

I think that finally can merged with `[skip-revcheck]`.

reference/image/functions/imagecopy.xml
Crell: I am not aware of a firm requirement in the docs so far.  However, I am a firm supporter of two-space, so absent a formal decision on a formatting style (which is out of scope for this thread), I would not make this change.

mmalferov: Wow! Two spaces after the dot at the end of the sentence? I didn't know about such a delicate moment. But what function does the double space perform? I opened a lot of XML documents in the @php/doc-en repository, but saw so few double spaces after the period that I was sure it was a typo ))

I don't know if I was able to find sufficiently authoritative sources, but a cursory search led to the answer that the double space was getting old today, unless "the teacher does require putting a double space" :-)

It seems that monospaced fonts are to blame

- https://www.thesaurus.com/e/writing/is-there-1-space-or-2-after-a-period/
- https://www.grammarly.com/blog/spaces-after-period/#:~:text=It%20is%20generally%20agreed%20that,space%20is%20correct%2C%20not%20two.

I see that the "teacher requires" to leave two spaces, but logic says that only one space is required. I feel confused and cognitive dissonance while I go to return two spaces…

P. S.

`A  t e s t  m e s s a g e  w i t h  d o u b l e  s p a c e s  b e e t w e e n  w o r d s, a n d w i t h o u t o n e` 

Obviously, it will be hard to read the text here without two spaces =)

cmb69: > but a cursory search led to the answer that the double space was getting old today
> It seems that monospaced fonts are to blame

Yeah, I think that stems from typewriter times, but actually, I like it too (and do it when writing emails in English; and would if writing in German, but that would probably be regarded as being totally wrong). One point where the double space is helpful for readers, is to distinguish between abbreviations and full-stops.

cmb69:
Thank you!

cmb69:
> I think that finally can merged with `[skip-revcheck]`.

And I forgot that. D'oh!

PR #3739, opened 2024-09-11
Deprecate CursorId, document Cursor::getId() returning Int64

jmikola:
https://jira.mongodb.org/browse/PHPC-2412

### DIFF
--- a/reference/mongodb/mongodb/driver/cursor/getid.xml
+++ b/reference/mongodb/mongodb/driver/cursor/getid.xml
@@ -10,13 +10,22 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>final</modifier> <modifier>public</modifier> <type>MongoDB\Driver\CursorId</type><methodname>MongoDB\Driver\Cursor::getId</methodname>
-   <void />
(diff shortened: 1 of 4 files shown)


reference/mongodb/mongodb/driver/cursor/getid.xml
```diff
@@ -60,17 +99,17 @@ $bulk->insert(['x' => 3]);
 $manager->executeBulkWrite('db.collection', $bulk);
 
 $cursor = $manager->executeQuery('db.collection', $query);
-var_dump($cursor->getId());
+var_dump($cursor->getId(true));
```
alcaeus, line 102: WDYT of leaving both behaviours in for the time being?

```suggestion
var_dump($cursor->getId());
var_dump($cursor->getId(true));
```

jmikola: In that case would you want the deprecation notice included in the expected output?

I intentionally removed the CursorId example knowing it would be temporary anyway after 2.0 removes `$asInt64`. At that point, the only reference to the original behavior would be in a changelog entry.

Considering the method itself is seldom used, I didn't see much benefit.

alcaeus: Yeah, that's a good point. Fine to leave as is 👍

PR #3743, opened 2024-09-13
constants.xml Unify titles for families of constants

mmalferov:


reference/imagick/constants.xml
```diff
@@ -121,7 +121,7 @@
  </variablelist>
 
  <variablelist xml:id="imagick.constants.disposetype">
-  <title>DISPOSE constants</title>
+  <title>DISPOSE_* constants</title>
```
cmb69, line 124: I'm not sure if we should go "all in", i.e.

```suggestion
  <title><constant>DISPOSE_<replaceable>*</replaceable></constant> constants</title>
```

What do you think?

sy-records: 🤔️ Others in this PR should also add `<constant> <replaceable>*</replaceable></constant>`

mmalferov: You're right, it's my mistake. It's done! 👍

sy-records: missing constant

mmalferov: @sy-records Could you please show me where the constants are missing?

reference/imagick/constants.xml
```diff
@@ -4,7 +4,7 @@
  &reftitle.constants;
  &extension.constants;
  <variablelist xml:id="imagick.constants.colortype">
-  <title>COLOR_* constants</title>
+  <title>COLOR_<replaceable>*</replaceable> constants</title>
```
sy-records, line 7:
```suggestion
  <title><constant>COLOR_<replaceable>*</replaceable></constant> constants</title>
  
```

mmalferov: I don't think I got enough sleep today )) Done!

sy-records (approved):
Thank you!

PR #3744, opened 2024-09-14
Recommend obtaining libedit instead of GNU Readline

petk:
The GNU Readline library is GPL 3 licensed and incompatible with PHP license.


cmb69 (approved):
Thank you!

cc @remicollet

PR #3745, opened 2024-09-14
Remove GNU Readline in favor of libedit

petk:
The libedit library is a non-GPL replacement for the GNU Readline library.

### DIFF
--- a/reference/readline/configure.xml
+++ b/reference/readline/configure.xml
@@ -5,13 +5,39 @@
  <para>
   To use these functions you must compile the CGI or CLI version of PHP
   with readline support. You need to configure PHP
-  <option role="configure">--with-readline[=DIR]</option>.
-  If you want to use the libedit readline replacement, configure PHP
-  <option role="configure">--with-libedit[=DIR]</option>.
+  <option role="configure">--with-readline</option>.
  </para>
  <para>
   On Windows this extension is available by default as of PHP 7.1.0.
  </para>
+ <simplesect role="changelog">
+  &reftitle.changelog;
+  <para>
+   <informaltable>
+    <tgroup cols="2">
+     <thead>
+      <row>
+       <entry>&Version;</entry>
+       <entry>&Description;</entry>
+      </row>
+     </thead>
+     <tbody>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Configure option <option role="configure">--with-libedit</option>
+        has been removed in favor of the
+        <option role="configure">--with-readline</option> which now doesn't
+        accept a DIR argument anymore in favor of the pkg-config and now links
+        to the libedit library, a non-GPL replacement for the GNU Readline
+        library.
+       </entry>
+      </row>
+     </tbody>
+    </tgroup>
+   </informaltable>
+  </para>
+ </simplesect>
 </section>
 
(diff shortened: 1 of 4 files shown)


reference/readline/configure.xml
```diff
@@ -5,13 +5,39 @@
  <para>
   To use these functions you must compile the CGI or CLI version of PHP
   with readline support. You need to configure PHP
-  <option role="configure">--with-readline[=DIR]</option>.
-  If you want to use the libedit readline replacement, configure PHP
-  <option role="configure">--with-libedit[=DIR]</option>.
+  <option role="configure">--with-readline</option>.
```
Girgias, line 8: Note the documentation shouldn't use personalization (i.e. using "you") maybe you could fix the text at the same time?

This feels like this should probably use some standard wording/an existing XML entity

reference/readline/setup.xml
```diff
@@ -8,14 +8,8 @@
...
-   To use the readline functions, you need to install libreadline. You can
-   find libreadline on the home page of the GNU Readline project, at
-   <link xlink:href="&url.readline;">&url.readline;</link>.
-   It's maintained by Chet Ramey, who's also the author of Bash.
-  </para>
-  <para>
-   You can also use these functions with the libedit library, a non-GPL
-   replacement for the readline library. The libedit library is BSD
+   To use the readline functions, you need to install the libedit library, a
+   non-GPL replacement for the GNU Readline library. The libedit library is BSD
```
Girgias, line 12: Ditto personalization

reference/readline/configure.xml
petk: Fixed in the next commit. I'm not sure I know what "standard wording/an existing XML entity" means here.

reference/readline/setup.xml
petk: Fixed in the next commit.

Girgias (approved):

cmb69:
What is the actual status here? Apparently, the constant `READLINE_LIB` has not been removed, and there are still `--with-libedit` *and* `--with-readline[=DIR]`. Ah, https://github.com/php/php-src/issues/15882 is still open, and https://github.com/php/php-src/pull/13184 has not been merged. So the documentation changes need to be reverted for now.

petk:
Yes, this was merged a bit too quickly. The code related PR wasn't merged to PHP 8.4 yet. Then let's merge it in PHP 8.5 or something.

cmb69:
Reverted via 51779f7c42f39934403805a42b83cef7126a5fc1.

I think we should go ahead with the removal of libreadline support in php-src.

petk:
I also think it's about time. Just for the info, it will be 6 years since the initial attempt to fix this and almost 1 year since the 2nd attempt. I'm actually quite amazed how long such issues take. And there is at least one more such issue - the gdbm library. At least from my current checks.

cmb69:
> I'm actually quite amazed how long such issues take.

Someone needs to push. :)

> And there is at least one more such issue - the gdbm library.

I have no problems with dropping this one. Should probably also drop support for ancient Berkeley DB 1 - 3 (can't imagine anybody using these for years).

petk:
This is really difficult to keep track of.

PR #3746, opened 2024-09-14
openssl-pkey-new.xml add output of current behavior + nits

mmalferov:


reference/openssl/functions/openssl-pkey-new.xml
```diff
@@ -67,7 +67,10 @@
      </row>
      <row>
       <entry>7.1.0</entry>
-      <entry>The <parameter>curve_name</parameter> option was added to make it possible to create EC keys.</entry>
+      <entry>
+       The <parameter>curve_name</parameter> key of the <parameter>options</parameter> parameter was
```
cmb69, line 71:
```suggestion
       The <literal>curve_name</literal> key of the <parameter>options</parameter> parameter was
```

reference/openssl/functions/openssl-pkey-new.xml
```diff
@@ -81,27 +84,47 @@
...
 $public_key = openssl_pkey_get_public($public_key_pem);
 var_dump($public_key);
+
 ?>
 ]]>
    </programlisting>
    &example.outputs.similar;
    <screen>
 <![CDATA[
+// Prior to PHP 8.0.0
```
cmb69, line 102: There is no real change in behavior; you just get different results each time you run the example.

cmb69 (commented):
Thank you!

reference/openssl/functions/openssl-pkey-new.xml
mmalferov: Oh! I guess I should have been more careful. I commented on the entire output for versions before and after 8, not on the actual key value (see, I even left the same keys for the results of different versions, although the keys were obviously different in the actual output). In older versions, the function returned a resource, and since version 8, it returned an object. It looks like you're right, and the version comment should have been left either a) just before `resource` (for versions before 8) and before `object` (since version 8), or b) changed the comments to: // Output prior to / as of PHP x.x.x, to make it clear that we're talking about the entire output, not the actual key value, which supposedly differs depending on the version ;-)

PR #3748, opened 2024-09-15
book.xml Add the `titleabbrev` tag for correct translation to other lang

mmalferov:


cmb69:
Do you mean that it can make sense for translations to have *different* `<title>` and `<titleabbrev>` for these cases?

mmalferov:
Yes, it make sense, see https://www.php.net/manual/ru/book.random.php for example.

In Russian, the word "Random" carries too vague a meaning, so it is better to give a detailed explanation on the page itself, and this requires both tags — `title` and `titleabbrev`, so that the short title remains in the table of contents, and the detailed one — on the page itself ;-)

cmb69:
Thanks for the clarification!

PR #3750, opened 2024-09-15
book.xml Capitalize the first letter of the sentense + tags

mmalferov:


reference/yaconf/book.xml
```diff
@@ -11,20 +11,20 @@
...
+   it parses <literal>INI</literal> files, and store the result
    in PHP when PHP is started, the result lives with the
    whole PHP lifecycle.
   </para>
   <para>
    Yaconf stores all configurations as
    interned string or immutable array, which means they are
    not refcounted-able, thus when you retrieving configurations
-   from yaconf, it could be considered as zero-copy, very fast.
+   from Yaconf, it could be considered as zero-copy, very fast.
```
Girgias, line 22: Maybe?
```suggestion
   from <acronym>Yaconf</acronym>, it could be considered as zero-copy, very fast.
```
I also don't know if we have an acronym for `INI`

PR #3751, opened 2024-09-15
book.xml Add missing space and tags

mmalferov:


reference/taint/book.xml
```diff
@@ -9,20 +9,21 @@
...
-   $_POST or $_COOKIE) to some functions, taint will warn you about that.
+   When taint is enabled, if you pass a tainted string (comes from <varname>$_GET</varname>,
+   <varname>$_POST</varname> or <varname>$_COOKIE</varname>) to some functions, taint will warn you about that.
   </para>
   <example>
    <title><function>Taint</function>example</title>
    <programlisting role="php">
 <![CDATA[
 <?php
+
```
cmb69, line 26: I would not add these newlines here and below. Ideally, we should agree on a (minimal) coding standard for PHP examples in the manual, and enforce that CS via some QA script. Otherwise we may go back and forth, over time.

mmalferov: A blank lines after `<?php` and before `?>` has been removes ;)

cmb69:
Thank you!

PR #3753, opened 2024-09-15
socket-recv.xml Replace the interned text with a variable

mmalferov:


reference/sockets/functions/socket-recv.xml
```diff
@@ -212,9 +212,7 @@ echo "OK.\n\n";
 ?>
 ]]>
     </programlisting>
-    <para>
-     The above example will produce something like:
-    </para>
+    &example.outputs;
```
cmb69, line 215:
```suggestion
    &example.outputs.similar;
```

devnexen (approved):

PR #3755, opened 2024-09-16
get-debug-type.xml Make the example executable + tags

mmalferov:


reference/var/functions/get-debug-type.xml
```diff
@@ -155,15 +158,18 @@ echo get_debug_type($fp) . PHP_EOL;
...
 
-echo get_debug_type(new class implements A {}) . PHP_EOL;
-echo get_debug_type(new class implements A,B {}) . PHP_EOL;
-echo get_debug_type(new class extends C {}) . PHP_EOL;
-echo get_debug_type(new class extends C implements A {}) . PHP_EOL;
+echo get_debug_type(new class implements A {}) . PHP_EOL
+    , get_debug_type(new class implements A,B {}) . PHP_EOL
+	   , get_debug_type(new class extends C {}) . PHP_EOL
+	   , get_debug_type(new class extends C implements A {}) . PHP_EOL
+;
```
Girgias, line 172: The indent here seems to be slightly off

cmb69: I think I would leave that as is; the leading commas are somwhat uncommon, and someone reading it, might suggest to change it again (we really should have a CS for the examples in the manual).

Maybe change the dots to commas, though.

Girgias: No there was an actual issue with the indent which was mixing tabs and spaces

reference/var/functions/get-debug-type.xml
```diff
@@ -142,6 +143,9 @@
     <programlisting role="php">
 <![CDATA[
 <?php
+
+namespace Foo;
+
 echo get_debug_type(null) . PHP_EOL;
```
kamil-tekiela, line 149: Small nit, but while you are changing something here, I would replace `PHP_EOL` with `"\n"`. That constant should be used only when parsing something.

cmb69: > That constant should be used only when parsing something.

And possibly not even then. :)

If touching this line, I would suggest to replace the dot with a comma (no need to concatenate the string before echoing).

cmb69:
I'm fine with the example changes, but it seems to me we should establish a coding standard, and automatically check that (so we don't go back and forth). A while ago I found https://github.com/php/doc-base/blob/master/scripts/qa/check-examples.php which currently applies PHP's linting to the examples; maybe that could be expanded (or changed) to run some tool (maybe PHP_CodeSniffer or PHP CS Fixer).

jimwins:
Examples are supposed to conform to the PEAR coding standard. http://doc.php.net/tutorial/style.php

Automatically checking it is a great idea.

Girgias:
> Examples are supposed to conform to the PEAR coding standard. http://doc.php.net/tutorial/style.php

We really should change this to be the PSR as I don't know who still uses PEAR, but that is another discussion.

> Automatically checking it is a great idea.

I concur.

cmb69:
At least one issue with PEAR CS would be the required [header comment blocks](https://pear.php.net/manual/en/standards.header.php), which don't make sense for our examples.

And I think we should add some additional style guidelines (e.g. always use `,` for `echo`; I don't think this is covered by any of the widespread style guidelines).

Girgias:
I just asked in the FIG discord if the CS PER mentions anything about `echo` with `,`.

kamil-tekiela:
PEAR CS is also very ambiguous in many cases. It's severely dated and PHP should not follow it. But we are hijacking this PR for unrelated discussion. I think we should deal with this PR and then raise a discussion about changing CS.

reference/var/functions/get-debug-type.xml
```diff
@@ -121,10 +121,11 @@
         <row>
          <entry>Objects from Anonymous Classes</entry>
          <entry>
-          <literal>"class@anonymous"</literal> or parent class name/interface name if the class extends another class or implements an interface e.g. <literal>"Foo\Bar@anonymous"</literal>
+          <literal>"class@anonymous"</literal> or parent class name/interface name if
+          the class extends another class or implements an interface e.g. <literal>"Foo\Bar@anonymous"</literal>
```
jimwins, line 125: Should undo this re-wrapping just to avoid confusing translators.

jimwins (approved):
LGTM with one small nit.

Discussion about CS taken to mailing list: https://news-web.php.net/php.doc/969388442

I'll also add a note about using `,` instead of `.` with `echo` to the style guide.

Girgias (approved):
LGTM, thank you!

PR #3756, opened 2024-09-16
socket-bind.xml: Remove notes related to Windows 9x/Me

leonardolara:
Remove notes related to Windows 9x/Me which is not supported by the PHP versions currently documented in the manual.


cmb69:
Those where the days … ;)

Thank you!

PR #3759, opened 2024-09-17
Use output tags

timacdonald:
This is computer output.

From what I can see, we usually wrap computer output in `computeroutput` tags without quotes.

See `a is greater than b` on: https://www.php.net/manual/en/control-structures.else.php

I would say the docs site should probably style this tag in monospace font any highlight somehow.

### DIFF
--- a/chapters/intro.xml
+++ b/chapters/intro.xml
@@ -39,7 +39,7 @@
    <para>
     Instead of lots of commands to output HTML (as seen in C or Perl),
     PHP pages contain HTML with embedded code that does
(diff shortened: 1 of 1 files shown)


cmb69:
> I would say the docs site should probably style this tag in monospace font any highlight somehow.

Indeed. Maybe for a start we could add a `span.computeroutput` selector to https://github.com/php/web-php/blob/f252981e5bf16659b2bdadd1d9d68743e41f9d15/styles/theme-medium.css#L350-L360.

cmb69:
Anyhow, thanks for the PR!

PR #3760, opened 2024-09-17
Change onboarding duration

timacdonald:
I think the sentence `writing simple scripts in a few hours` is misleading and detrimental to our message that starting with PHP is easy. 

I can copy and paste the example on the page and I already have a PHP script.

I'd like to propose we change it to something a little more motivational for a new comer.

> With PHP, you can get up and running and be writing simple scripts in no time at all.

### Before

Screenshot 2024-09-17 at 11 12 49

### After

Screenshot 2024-09-17 at 11 15 45


Girgias (approved):
This wording makes more sense to me

PR #3762, opened 2024-09-18
apache2.xml Wrap commands, literals and filenames with tags

mmalferov:


cmb69:
Thank you! While this file is in need of futher updates (there should be no personalization, i.e. use of "you", and some stuff is ridiculously outdated, e.g. .php2 files), this is a good step.

PR #3773, opened 2024-09-23
Document `FFI\CType::getFuncParameterCount()` purpose and return

DanielEScherzer:


reference/ffi/ctype/getfuncparametercount.xml
```diff
@@ -27,7 +25,9 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   
+   Returns the number of parameters for the underlying function type.
+	  If the underlying type is not a function, a
+	  <exceptionname>FFI\Exception</exceptionname> is thrown.
```
Girgias, line 30: The indent is off here, we use 1 space for indentation for XML files.

Girgias (changes requested):
Note the `<refpurpose>` tag still just says "Description"

reference/ffi/ctype/getfuncparametercount.xml
DanielEScherzer: This entire block is indented 1 space more than than the `<para>` tag, is that not correct?

DanielEScherzer:
> Note the `<refpurpose>` tag still just says "Description"

Done

reference/ffi/ctype/getfuncparametercount.xml
cmb69: There is a tab at the beginning of line 29 and 30.

DanielEScherzer: Ah, thanks

reference/ffi/ctype/getfuncparametercount.xml
```diff
@@ -27,7 +25,9 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   
+   Returns the number of parameters for the underlying function type.
+   If the underlying type is not a function, a
+   <exceptionname>FFI\Exception</exceptionname> is thrown.
```
cmb69, line 30: That should probably be
```suggestion
   If the underlying type is not a function, an
   <exceptionname>FFI\Exception</exceptionname> is thrown.
```
(at least I read FFI as EFF-EFF-I)

DanielEScherzer: done

cmb69 (approved):
Thank you! This looks fine to me.

Girgias (approved):
Thank you!

PR #3774, opened 2024-09-23
Clarify passphrase for openssl_encrypt and openssl_decrypt

bukka:
Extend info for passphrase behaviour and note that it doesn't use KDF.

### DIFF
--- a/reference/openssl/functions/openssl-decrypt.xml
+++ b/reference/openssl/functions/openssl-decrypt.xml
@@ -19,7 +19,7 @@
    <methodparam choice="opt"><type>string</type><parameter>aad</parameter><initializer>""</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Takes a raw or base64 encoded string and decrypts it using a given method and key.
+   Takes a raw or base64 encoded string and decrypts it using a given method and passphrase.
   </para>
 
  </refsect1>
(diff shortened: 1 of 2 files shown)


reference/openssl/functions/openssl-decrypt.xml
```diff
@@ -68,7 +77,9 @@
      <term><parameter>iv</parameter></term>
      <listitem>
       <para>
-       A non-NULL Initialization Vector.
+       A non-NULL Initialization Vector. If the IV is shorter than expected, it is padded with
```
cmb69, line 80: Maybe:
```suggestion
       A non-&null; Initialization Vector. If the IV is shorter than expected, it is padded with
```

reference/openssl/functions/openssl-encrypt.xml
```diff
@@ -70,8 +70,17 @@
...
+       <literal>NUL</literal> characters; if the passphrase is longer than expected, it is
+       silently truncated.
       </para>
+      <caution>
+       <simpara>
+        There is no key derivation function used for <parameter>passphrase</parameter> as its name
+        might suggest. The only operation used is padding with <literal>NUL</literal> characters
+        or truncation if the length is different than expected.
+       </simpara>
+      </caution>
```
cmb69, line 83: This looks wrong, since this is about the `$iv` parameter; `$passphrase` is documented above.

cmb69 (commented):
Thank you!

reference/openssl/functions/openssl-encrypt.xml
bukka: ah yeah, it was late... :)

PR #3775, opened 2024-09-23
Remove very old (4.x) changelog information on a few functions

jimwins:


kamil-tekiela (approved):

reference/mysql/functions/mysql-list-tables.xml
```diff
@@ -61,30 +61,6 @@
...
-      <row>
-       <entry>&Version;</entry>
-       <entry>&Description;</entry>
-      </row>
-     </thead>
-     <tbody>
-      <row>
-       <entry>4.3.7</entry>
-       <entry>
-        This function became deprecated.
```
Girgias, line 79: Does it make sense to just remove the documentation page then, as I'm guessing it was removed.

Girgias (approved):
One question about the `ext/mysql` doc page but LGTM otherwise.

reference/mysql/functions/mysql-list-tables.xml
jimwins: I considered that, but all of the ext/mysql documentation is sort of an historical artifact because that extension was removed in 7.0.0, so I figured I wouldn't disturb it too much. Maybe I'll propose removing all of it in another PR.

Girgias: It was moved to PECL and had some recently-ish releases.

jimwins: Okay, going to just leave the chapter alone, then, aside from removing the changelog entry. Whoever is maintaining the extension in PECL can decide how they want to deal with the documentation.

cmb69: > It was moved to PECL and had some recently-ish releases.

Doesn't look like it had releases, but only a few commits (https://pecl.php.net/package/mysql and https://github.com/php/pecl-database-mysql).

PR #3776, opened 2024-09-24
Add titleabbrev+title to preface and reformat info box

jimwins:
Kind of a proof of concept for using titleabbrev+title for more sections so the front page of the manual isn't quite so sparse.


Girgias (approved):
MSTM

PR #3779, opened 2024-09-24
Update and fix docs for openssl_open and openssl_seal

bukka:
The current docs for `openssl_seal` and `openssl_open` are outdated and incorrect in many places. This fixes the issues and makes them up to date with more info.

### DIFF
--- a/reference/openssl/functions/openssl-open.xml
+++ b/reference/openssl/functions/openssl-open.xml
@@ -18,14 +18,13 @@
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>iv</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   <function>openssl_open</function> opens (decrypts)
-   <parameter>data</parameter> using the private key associated with
-   the key identifier <parameter>private_key</parameter> and the envelope key
-   <parameter>encrypted_key</parameter>, and fills
-   <parameter>output</parameter> with the decrypted data.
-   The envelope key is generated when the
-   data are sealed and can only be used by one specific private key. See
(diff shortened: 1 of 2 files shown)


reference/openssl/functions/openssl-seal.xml
```diff
@@ -73,8 +72,9 @@
        The cipher method.
        <caution>
         <simpara>
-         The default value (<literal>'RC4'</literal>) is considered insecure.
-         It is strongly recommended to explicitly specify a secure cipher method.
+         The default value for PHP versions before 8.0 is (<literal>'RC4'</literal>) which is
```
cmb69, line 75: Nit:
```suggestion
         The default value for PHP versions prior to 8.0.0 is (<literal>'RC4'</literal>) which is
```

reference/openssl/functions/openssl-open.xml
```diff
@@ -81,7 +82,8 @@
      <term><parameter>iv</parameter></term>
      <listitem>
       <para>
-       The initialization vector.
+       The initialization vector used for decryption of <parameter>data</parameter>. It is required
+       if the cipher method requires IV.
```
cmb69, line 86: If we have a list of supported ciphers (and whether they require an IV), we should link to it. I don't think we have such a list, though.

cmb69 (commented):
Thank you very much! Just a couple of comments, but otherwise looks good.

reference/openssl/functions/openssl-open.xml
bukka: We cannot that get list but it can be found out from `openssl_cipher_iv_length` so added a note about it.

cmb69 (approved):
Thank you!

PR #3785, opened 2024-09-24
Change reference from installation.problems to faq.installation

jimwins:
A chapter with three tiny sections is annoying.

### DIFF
--- a/install/unix/source.xml
+++ b/install/unix/source.xml
@@ -84,9 +84,9 @@
  <simpara>
   After the configuration script was been run, PHP can be built using
   the <command>make</command> command.
-  The <link linkend="install.problems">Problems section of this
-  manual</link> has further information about how to handle build
-  problems.
+  The <link linkend="faq.installation">Installation section of the
+  frequently asked questions</link> has further information about how
+  to handle build problems.
  </simpara>
 
  <note>


cmb69:
Makes sense to collapse this, but I'm not sure whether we should keep the info about the install mailing list. Apparently, nobody is responding there for a couple of years: https://news-web.php.net/php.install.

jimwins:
Maybe it would be better to just remove this section entirely and link directly to faq.installation instead.

Girgias:
If the section is removed, remember to first delete all the notes on these pages.

Girgias:
I do think removing this and linking to the FAQ is better, as this section seems kinda pointless?

Not sure what I think about the "if you found a bug" thing, as it makes some sense to have this somewhere, but not in the installation section.

jimwins:
Okay, removal it is. The stuff about bug reporting is already in the general FAQ. Requires a doc-base change because the section is in `manual.xml.in`.

Girgias:
Could you split it in two commits? First change where the link points, merge this, update translations, then remove the page and update doc-base, seems like a more reliable course of action

jimwins:
Sorry for the flapping, fixed this PR to just update the reference. Will submit another PR to remove.

Girgias (approved):

PR #3786, opened 2024-09-25
Stop chunking this chapter, add titleabbrev+title, write-through to remove 'you'

jimwins:


chapters/intro.xml
```diff
@@ -43,38 +46,38 @@
     The PHP code is enclosed in special <link
     linkend="language.basic-syntax.phpmode">start and end processing
     instructions <code>&lt;?php</code> and <code>?&gt;</code></link>
-    that allow you to jump into and out of "PHP mode."
+    that allow jumping into and out of "PHP mode."
```
Girgias, line 49: Seem better, no?
```suggestion
    that allow jumping in and out of "PHP mode."
```

chapters/intro.xml
```diff
@@ -83,26 +86,23 @@
...
-       and main target field for PHP. You need three things
+       Server-side scripting. This is the most widely used
+       and main target field for PHP. Three things are needed
        to make this work: the PHP parser (CGI or server
-       module), a web server and a web browser. You need to
-       run the web server, with a connected PHP installation.
-       You can access the PHP program output with a web browser,
-       viewing the PHP page through the server. All these can
-       run on your home machine if you are just experimenting
+       module), a web server and a web browser. All these can
```
Girgias, line 92: Oxford comma
```suggestion
       module), a web server, and a web browser. All these can
```

Girgias (commented):
Minor nits.

I needed to double-check the [style guide](http://doc.php.net/tutorial/style.php) if the introduction was an accepted place for personalization, but it doesn't seem like.

It would probably be a good idea to have a GH Action to warn on this.

Girgias (approved):

chapters/intro.xml
Girgias: Actually, should the full stop not be outside the quotes ?

Girgias: And possibly even use the DocBook `<quote>` tag

jimwins:
> I needed to double-check the [style guide](http://doc.php.net/tutorial/style.php) if the introduction was an accepted place for personalization, but it doesn't seem like.
> 
> It would probably be a good idea to have a GH Action to warn on this.

I think the case could be made for not allowing personalization anywhere, going forward. I'm pretty sure that tutorials and FAQs could be written without needing to lean on "you", too.

Girgias (approved):

PR #3788, opened 2024-09-25
Add caution about using the same DH params in openssl_dh_compute_key

bukka:
Currently there is no warning that it requires using the same DH parameters.


cmb69 (approved):
Thank you!

PR #3789, opened 2024-09-25
Add caution about openssl_pkey_derive key_length parameter

bukka:
We should probably deprecate that `key_length` in 8.5...

### DIFF
--- a/reference/openssl/functions/openssl-pkey-derive.xml
+++ b/reference/openssl/functions/openssl-pkey-derive.xml
@@ -52,7 +52,7 @@
       <simpara>
        This parameter should not be used, as it does not work as expected. It never returns a secret
        longer than the size of the prime. If the desired length is smaller than the size of the
-       prime, it truncate the length only for ECDH keys but fails for DH keys.
+       prime, it truncates the length only for ECDH keys but fails for DH keys.
       </simpara>
      </caution>
     </listitem>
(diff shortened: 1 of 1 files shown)


reference/openssl/functions/openssl-pkey-derive.xml
```diff
@@ -46,8 +46,15 @@
...
     <listitem>
      <para>
-      If not zero, will set the desired length of the derived secret.
+      If not zero, will attempt to set the desired length of the derived secret.
      </para>
+     <caution>
+      <simpara>
+       This parameter should not be used, as it does not work as expected. It never returns a secret
+       longer than the size of the prime. If the desired length is smaller than the size of the
+       prime, it truncate the length only for ECDH keys but fails for DH keys.
```
cmb69, line 55:
```suggestion
       prime, it truncates the length only for ECDH keys but fails for DH keys.
```

cmb69 (commented):
Thank you!

I agree that it's probably good to deprecated this parameter.

I also think that we should add [acronym definitions](https://github.com/php/doc-base/blob/master/entities/acronyms.xml) for `ECDH` and `DH`, but that can be done in a separate PR.

Girgias:
I created the 8.5 mass deprecation RFC with a stub for the deprecation: https://wiki.php.net/rfc/deprecations_php_8_5

Feel free to fill it out. :)

bukka:
> I created the 8.5 mass deprecation RFC with a stub for the deprecation: https://wiki.php.net/rfc/deprecations_php_8_5

Thanks

> Feel free to fill it out. :)

Add some description

PR #3796, opened 2024-09-26
Tweak the title and depersonalize the password hashing FAQ section

jimwins:


faq/passwords.xml
```diff
@@ -86,9 +86,9 @@
      <para>
       Another option is the <function>crypt</function> function, which
       supports several hashing algorithms. When using
-      this function, you are guaranteed that the algorithm you select is
+      this function, it is guaranteed that the algorithm selected is
       available, as PHP contains native implementations of each supported
-      algorithm, in case one or more are not supported by your system.
+      algorithm, in case one or more are not supported by the system.
```
Girgias, line 91: I don't think I would recommend `crypt()`, the recommendation should really just be `password_hash()`

jimwins: Okay, tackled the TODO that was already there. Just mention `crypt()` in passing now and leave the text discouraging its use.

cmb69 (approved):
Thank you! This looks good to me.

faq/passwords.xml
```diff
@@ -4,161 +4,150 @@
...
     <answer>
-     <para>
+     <simpara>
       When using <function>password_hash</function> or
       <function>crypt</function>, the return value includes the salt as part
       of the generated hash. This value should be stored verbatim in the
       database, as it includes information about the hash function that was
       used and can then be given directly to
       <function>password_verify</function> or <function>crypt</function> when
       verifying passwords.
```
Girgias, line 143: `password_verify()` can be used with outputs from `crypt()` so maybe drop the implicit suggestion that one can compare two crypt hashes? (something something timing attacks)

Girgias (approved):
One final comment but LGTM otherwise

faq/passwords.xml
jimwins: Good point, and because I cut out the other (longer) reference to timing attacks, I've added in a new warning.

Girgias (approved):
Thank you!

PR #3797, opened 2024-09-26
FILTER_VALIDATE_BOOL: make it clearer than `null` is treated as an empty string

### LINKED ISSUE #3801: FILTER_VALIDATE_BOOLEAN returns wrong result for null
### Description

The following code:

```php
<?php
var_dump(filter_var(null, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE));
```

Resulted in this output:
```
bool(false)
```

```
null
```
Documentation states

>  Returns true for "1", "true", "on" and "yes". Returns false otherwise.
If FILTER_NULL_ON_FAILURE is set, false is returned only for "0", "false", "off", "no", and "", and null is returned for all non-boolean values. 

So the expected result should return `null` as `null` is not '0', 'false', 'off' or ''.

### PHP Version

php 8.0.28

### Operating System

Debian

DanielEScherzer:
Ref php/doc-en#3801

### DIFF
--- a/reference/filter/filters.xml
+++ b/reference/filter/filters.xml
@@ -43,6 +43,10 @@
           &null; is returned for all non-boolean values.
          </para>
          <para>
+          Note that, since values are <link linkend="language.types.string.casting">converted
+          to string</link> before processing, &null; is treated as "".
+         </para>
+         <para>
           String values are trimmed using <function>trim</function> before comparison.
          </para>
         </entry>


cmb69:
Thank you for the PR! However, I'm not sure how to document this. Note that `null` is just a special case; a stringable object would also be accepted (https://3v4l.org/jFgDD). But even `FILTER_VALIDATE_BOOL` is only a special case; it seems to me that `FILTER_REQUIRE_SCALAR` is generally a mis-nomer, and badly documented: "Requires the value to be scalar.", while in practice it is fine with most non scalar values, too (e.g. https://3v4l.org/8ZVMA).

kamil-tekiela:
Perhaps it should say that all values are converted to string

Girgias:
This PR needs to be reworked, I agree that saying any value being checked is going to be cast to string is better.

lacatoire:
Superseded by #4158, which removed reference/filter/filters.xml and moved the note into reference/filter/functions/filter-var.xml as "Scalar values are converted to string internally before they are filtered."

The remaining inaccuracy is in reference/filter/constants.xml, not here: it says non-boolean values return null under FILTER_NULL_ON_FAILURE, but null input returns false, since it casts to "" and matches the false list.

PR #3799, opened 2024-09-26
add Zip::ER_TRUNCATED_ZIP

remicollet:
Introduced in libzip 1.11
Available in zip extension version 1.22.4 (and php 8.4)


cmb69:
Thank you!

PR #3803, opened 2024-09-26
Clarify extension status

jimwins:
This tries to sync up the status of some experimental/unmaintained extensions.


cmb69 (approved):
For some extensions, it's a bit hard to categorize them (e.g. zmq never had a stable release, so experimental, but it's also unmaintained, so might be considered deprecated). But generally, this looks good to me.

Girgias:
I think we should really start considering creating an archive and removing abandoned extensions from the manual.

cmb69:
Perhaps even trying documenting (all) PECL extensions is too ambitious? Maybe analytics will help to sort out which extensions are more important than others, but I wonder what happens when PIE takes off. I mean it's great to have everything in one place, and conforming to a certain style, but the few volunteers catering to doc-en are fighting a losing battle for years, unfortunately.

Girgias:
Some extension have their documentation outside php.net already, I don't think it is reasonable for us to document these extensions.

jimwins:
Looking at the [extensions by membership](https://www.php.net/manual/en/extensions.membership.php) in the manual now, I would not be sad to excise the documentation for almost all of the "PECL Extensions". It would be great if the PHP project could provide some sort of infrastructure for building documentation for extensions and supporting translations, but I'm not sure we have the organizational bandwidth to do so. As we move towards thinking of extensions more like we do third-party PHP code (Composer packages), not providing that infrastructure would be in line with that.

PR #3804, opened 2024-09-27
Specify C11 requirement since 8.4

iluuu1994:


install/unix/source.xml
```diff
@@ -9,7 +9,7 @@
     <link xlink:href="&url.gnu.make;">GNU <command>make</command></link>
    </member>
    <member>
-    A C compiler (as of PHP 8.0.0, C99 compatibility is assumed)
+    A C compiler (C11 compatibility is required since PHP 8.4, C99 since 8.0)
```
cmb69, line 12: Please stick with the "as of" wording and the full version number (see https://github.com/php/web-doc/blob/master/tutorial/style.md#chronology)

```suggestion
    A C compiler (as of PHP 8.0.0, C99 compatibility is required; as of PHP 8.4.0, C11 compatibility is required)
```

cmb69:
Thank you!

PR #3805, opened 2024-09-27
Add missing version info for Document::toPHP and Iterator ctor

jmikola:
Noticed both of the following pages were displaying "(No version information available, might only be in Git)":

 * https://www.php.net/manual/en/mongodb-bson-document.tophp.php
 * https://www.php.net/manual/en/mongodb-bson-iterator.construct.php


GromNaN (approved):

alcaeus (approved):
Whoops, good catch. Thank you!

PR #3809, opened 2024-09-29
Simplify array_replace description and add nested array example

jimwins:
Based on discussion at PR #1470

### DIFF
--- a/reference/array/functions/array-replace.xml
+++ b/reference/array/functions/array-replace.xml
@@ -65,21 +65,25 @@ $replacements = array(0 => "pineapple", 4 => "cherry");
 $replacements2 = array(0 => "grape");
 
 $basket = array_replace($base, $replacements, $replacements2);
-print_r($basket);
+var_dump($basket);
 ?>
 ]]>
     </programlisting>
     &example.outputs;
     <screen role="php">
 <![CDATA[
-Array
-(
-    [0] => grape
-    [1] => banana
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-replace.xml
```diff
@@ -85,6 +80,40 @@ Array
...
+    <title>Example of how nested arrays are handled</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$base = [ 'citrus' => [ 'orange', 'lemon' ], 'pome' => [ 'apple' ] ];
+$replacements = [ 'citrus' => [ 'grapefruit' ] ];
+$replacements2 = [ 'citrus' => [ 'kumquat', 'citron' ], 'pome' => [ 'loquat' ] ];
+
+$basket = array_replace($base, $replacements, $replacements2);
+print_r($basket);
```
Girgias, line 96: I usually prefer `var_dump()` then `print_r()` as it is whitespace sensitive in very awkward places :/

jimwins: I just copied what the first example was doing. I've changed them both to `var_dump()`.

jimwins: I did leave alone the `array(...)` syntax on the first example even though I used `[ ... ]` in the second.

Girgias (approved):
Thank you!

PR #3817, opened 2024-10-01
wordwrap.xml Add errors/changelog sections, specify U+0020 instead U+20

mmalferov:
https://stackoverflow.com/questions/1273693/why-is-u-used-to-designate-a-unicode-code-point

### DIFF
--- a/reference/strings/functions/wordwrap.xml
+++ b/reference/strings/functions/wordwrap.xml
@@ -48,7 +48,7 @@
      <listitem>
       <para>
        The line is broken using the optional <parameter>break</parameter> parameter.
-       It cannot be an empty string.
+       It must not be an empty string.
       </para>
      </listitem>
     </varlistentry>
--- a/reference/strings/functions/wordwrap.xml
+++ b/reference/strings/functions/wordwrap.xml
@@ -74,13 +74,37 @@
    Returns the given string wrapped at the specified length.
   </para>
  </refsect1>
- 
+
  <refsect1 role="errors">
   &reftitle.errors;
   <simpara>
    If <parameter>break</parameter> is empty string,
-   a <classname>ValueError</classname> is thrown.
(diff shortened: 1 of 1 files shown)


reference/strings/functions/wordwrap.xml
```diff
@@ -75,6 +75,38 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   If <parameter>break</parameter> is empty string,
+   emitted an <constant>E_WARNING</constant> and returned &false;.
+  </simpara>
```
Girgias, line 83: This should say that a ValueError is thrown.

reference/strings/functions/wordwrap.xml
```diff
@@ -47,8 +47,8 @@
      <term><parameter>break</parameter></term>
      <listitem>
       <para>
-       The line is broken using the optional
-       <parameter>break</parameter> parameter.
+       The line is broken using the optional <parameter>break</parameter> parameter.
+       It cannot be an empty string.
```
Girgias, line 51: We recently changed the wording of the error to align with other errors that use `must`
```suggestion
       It must not be an empty string.
```

Girgias (changes requested):

reference/strings/functions/wordwrap.xml
```diff
@@ -75,6 +75,38 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   If <parameter>break</parameter> is empty string,
```
Girgias, line 81:
```suggestion
   If <parameter>break</parameter> is an empty string,
```

Girgias (approved):
Minor nit and LGTM

PR #3818, opened 2024-10-01
Update CURLOPT constant descriptions

haszi:
Updated descriptions of `CURLOPT_*` constants with the following:

- added some missing descriptions
- added type information on the associated accepted values
- added default values
- added cURL version information on availability, deprecation, removal 
- added applicable DocBook tags
- replaced lists of accepted value constants with their generalized version (eg.  `CURLALTSVC_*` instead of listing all `CURLALTSVC_` constants individually)
- standardized minor style details (e.g. use cURL instead of curl, libcurl, CURL; sentences to end in a period; remove double quotes from within `<literal>` tags, etc.)

Unfortunately since `CURLOPT` constants are the biggest subset of all cURL constants, the diff is pretty big.

### DIFF
--- a/reference/curl/constants_curl_setopt.xml
+++ b/reference/curl/constants_curl_setopt.xml
@@ -13,7 +13,7 @@
      as <constant>CURLOPT_UNIX_SOCKET_PATH</constant>. These two options
      share the same storage and therefore only one of them can be set
      per handle.
-     Available as of PHP 7.3.0 and cURL 7.53.0
+     Available as of PHP 7.3.0 and cURL 7.53.0.
     </para>
    </listitem>
   </varlistentry>
@@ -68,8 +68,10 @@
    </term>
    <listitem>
     <para>
-     Pass the filename for cURL to use as the Alt-Svc cache file to read existing cache contents from and
-     possibly also write it back to a after a transfer, unless <constant>CURLALTSVC_READONLYFILE</constant>
+     Pass a <type>string</type> with the filename for cURL to use
+     as the Alt-Svc cache file to read existing cache contents from
+     and possibly also write it back to a after a transfer,
+     unless <constant>CURLALTSVC_READONLYFILE</constant>
      is set via <constant>CURLOPT_ALTSVC_CTRL</constant>.
      Available as of PHP 8.2.0 and cURL 7.64.1.
     </para>
@@ -85,11 +87,10 @@
      Populate the bitmask with the correct set of features to instruct cURL how to handle Alt-Svc for the
      transfers using this handle. cURL only accepts Alt-Svc headers over HTTPS. It will also only complete
      a request to an alternative origin if that origin is properly hosted over HTTPS.
-     Setting any bit will enable the alt-svc engine. The options are:
-     <constant>CURLALTSVC_H1</constant>,
-     <constant>CURLALTSVC_H2</constant>,
-     <constant>CURLALTSVC_H3</constant>, and
-     <constant>CURLALTSVC_READONLYFILE</constant>.
+     Setting any bit will enable the alt-svc engine.
+     Set to any of the
+     <constant>CURLALTSVC_<replaceable>*</replaceable></constant> constants.
+     Defaults to Alt-Svc handling being disabled.
      Available as of PHP 8.2.0 and cURL 7.64.1.
     </para>
    </listitem>
@@ -117,6 +118,8 @@
     <para>
      &true; to automatically set the <literal>Referer:</literal> field in
      requests where it follows a <literal>Location:</literal> redirect.
+     Defaults to <literal>0</literal>.
+     Available as of cURL 7.1.
     </para>
    </listitem>
   </varlistentry>
@@ -127,10 +130,10 @@
    </term>
    <listitem>
     <para>
-     Provides AWS V4 signature authentication on HTTP(S) header.
+     Provides AWS V4 signature authentication on HTTP(S) header as a <type>string</type>.
      This option overrides any other authentication types that have been set in
      <constant>CURLOPT_HTTPAUTH</constant>. This method cannot be combined with other authentication types.
-     Available as of PHP 8.2.0 and cURL 7.75.0
+     Available as of PHP 8.2.0 and cURL 7.75.0.
     </para>
    </listitem>
   </varlistentry>
@@ -141,7 +144,8 @@
    </term>
    <listitem>
     <para>
-
+     This constant is no longer used as of PHP 5.5.0.
+     Deprecated as of PHP 8.4.0.
     </para>
    </listitem>
   </varlistentry>
@@ -154,6 +158,8 @@
     <para>
      The size of the buffer to use for each read. There is no guarantee
      this request will be fulfilled, however.
+     This option accepts any value that can be cast to a valid <type>int</type>.
+     Defaults to <constant>CURL_MAX_WRITE_SIZE</constant> (16kB).
      Available as of cURL 7.10.
     </para>
    </listitem>
@@ -165,9 +171,10 @@
    </term>
    <listitem>
     <para>
-     The name of a file holding one or more certificates to verify the
+     A <type>string</type> with the name of a file holding one or more certificates to verify the
      peer with. This only makes sense when used in combination with
      <constant>CURLOPT_SSL_VERIFYPEER</constant>. Might require an absolute path.
+     Available as of cURL 7.4.2.
     </para>
    </listitem>
   </varlistentry>
@@ -178,9 +185,9 @@
    </term>
    <listitem>
     <para>
-     The name of a PEM file holding one or more certificates to verify the
+     A <type>string</type> with the name of a PEM file holding one or more certificates to verify the
      peer with. This option overrides <constant>CURLOPT_CAINFO</constant>.
-     Available as of PHP 8.2.0 and cURL 7.77.0
+     Available as of PHP 8.2.0 and cURL 7.77.0.
     </para>
    </listitem>
   </varlistentry>
@@ -191,8 +198,9 @@
    </term>
    <listitem>
     <para>
-     A directory that holds multiple CA certificates. Use this option
-     alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
+     A <type>string</type> with a directory that holds multiple CA certificates.
+     Use this option alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
+     Available as of cURL 7.9.8.
     </para>
    </listitem>
   </varlistentry>
@@ -203,6 +211,10 @@
    </term>
    <listitem>
     <para>
+     Sets the maximum time in seconds any in memory cached CA certificate store
+     may be kept and reused for new connections.
+     This option accepts any value that can be cast to a valid <type>int</type>.
+     Defaults to <literal>86400</literal> (24 hours).
      Available as of PHP 8.3.0 and cURL 7.87.0
     </para>
    </listitem>
@@ -215,8 +227,10 @@
    <listitem>
     <para>
      &true; to output SSL certification information to <constant>STDERR</constant>
-     on secure transfers. Added in cURL 7.19.1.
+     on secure transfers.
      Requires <constant>CURLOPT_VERBOSE</constant> to be on to have an effect.
+     Defaults to &false;.
+     Available as of cURL 7.19.1.
     </para>
    </listitem>
   </varlistentry>
@@ -229,6 +243,9 @@
     <para>
      The number of seconds to wait while trying to connect. Use 0 to
      wait indefinitely.
+     This option accepts any value that can be cast to a valid <type>int</type>.
+     Defaults to <literal>300</literal>.
+     Available as of cURL 7.7.
     </para>
    </listitem>
   </varlistentry>
@@ -239,12 +256,14 @@
    </term>
    <listitem>
     <para>
-     The number of milliseconds to wait while trying to connect. Use 0 to
-     wait indefinitely.
-     <!-- http://curl.haxx.se/libcurl/c/curl_easy_setopt.html -->
-     If libcurl is built to use the standard system name resolver, that
+     The number of milliseconds to wait while trying to connect.
+     Use <literal>0</literal> to wait indefinitely.
+     If cURL is built to use the standard system name resolver, that
      portion of the connect will still use full-second resolution for
-     timeouts with a minimum timeout allowed of one second. Available as of cURL 7.16.2.
+     timeouts with a minimum timeout allowed of one second.
+     This option accepts any value that can be cast to a valid <type>int</type>.
+     Defaults to <literal>300000</literal>.
+     Available as of cURL 7.16.2.
     </para>
    </listitem>
   </varlistentry>
@@ -258,6 +277,7 @@
      &true; tells the library to perform all the required proxy authentication
      and connection setup, but no data transfer. This option is implemented for
      HTTP, SMTP and POP3.
+     Defaults to &false;.
      Available as of cURL 7.15.2.
     </para>
    </listitem>
@@ -270,9 +290,9 @@
    <listitem>
     <para>
      Connect to a specific host and port instead of the URL's host and port.
-     Accepts an array of strings with the format
+     Accepts an <type>array</type> of <type>string</type>s with the format
      <literal>HOST:PORT:CONNECT-TO-HOST:CONNECT-TO-PORT</literal>.
-     Available as of PHP 7.0.7 and cURL 7.49.0
+     Available as of PHP 7.0.7 and cURL 7.49.0.
     </para>
    </listitem>
   </varlistentry>
@@ -283,10 +303,11 @@
    </term>
    <listitem>
     <para>
-     The contents of the <literal>"Cookie: "</literal> header to be
-     used in the HTTP request.
+     A <type>string</type> with the contents
+     of the <literal>Cookie: </literal> header to be used in the HTTP request.
      Note that multiple cookies are separated with a semicolon followed
-     by a space (e.g., "<literal>fruit=apple; colour=red</literal>")
+     by a space (e.g., <literal>fruit=apple; colour=red</literal>).
+     Available as of cURL 7.1.
     </para>
    </listitem>
   </varlistentry>
@@ -297,11 +318,12 @@
    </term>
    <listitem>
     <para>
-     The name of the file containing the cookie data. The cookie file can
-     be in Netscape format, or just plain HTTP-style headers dumped into
-     a file.
-     If the name is an empty string, no cookies are loaded, but cookie
+     A <type>string</type> with the name of the file containing the cookie data.
+     The cookie file can be in Netscape format,
+     or just plain HTTP-style headers dumped into a file.
+     If the name is an empty <type>string</type>, no cookies are loaded, but cookie
      handling is still enabled.
+     Available as of cURL 7.1.
     </para>
    </listitem>
   </varlistentry>
@@ -312,8 +334,9 @@
    </term>
    <listitem>
     <para>
-     The name of a file to save all internal cookies to when the
-     handle's destructor is called.
+     A <type>string</type> with the name of a file to save all internal cookies to
+     when the handle's destructor is called.
+     Available as of cURL 7.9.0.
      <warning>
       <simpara>
        As of PHP 8.0.0, <function>curl_close</function> is a no-op
@@ -332,13 +355,27 @@
    </term>
    <listitem>
     <para>
-     A cookie string (i.e. a single line in Netscape/Mozilla format, or a regular
+     A cookie <type>string</type> (i.e. a single line in Netscape/Mozilla format, or a regular
      HTTP-style Set-Cookie header) adds that single cookie to the internal cookie store.
-     <literal>"ALL"</literal> erases all cookies held in memory.
-     <literal>"SESS"</literal> erases all session cookies held in memory.
-     <literal>"FLUSH"</literal> writes all known cookies to the file specified by <constant>CURLOPT_COOKIEJAR</constant>.
-     <literal>"RELOAD"</literal> loads all cookies from the files specified by <constant>CURLOPT_COOKIEFILE</constant>.
-     Available as of cURL 7.14.1
+     <simplelist type="inline">
+      <member>
+       <literal>ALL</literal>
+       erases all cookies held in memory
+      </member>
+      <member>
+       <literal>SESS</literal>
+       erases all session cookies held in memory
+      </member>
+      <member>
+       <literal>FLUSH</literal>
+       writes all known cookies to the file specified by <constant>CURLOPT_COOKIEJAR</constant>
+      </member>
+      <member>
+       <literal>RELOAD</literal>
+       loads all cookies from the files specified by <constant>CURLOPT_COOKIEFILE</constant>
+      </member>
+     </simplelist>
+     Available as of cURL 7.14.1.
     </para>
    </listitem>
   </varlistentry>
@@ -349,12 +386,13 @@
    </term>
    <listitem>
     <para>
-     &true; to mark this as a new cookie "session". It will force libcurl
+     &true; to mark this as a new cookie "session". It will force cURL
      to ignore all cookies it is about to load that are "session cookies"
-     from the previous session. By default, libcurl always stores and
+     from the previous session. By default, cURL always stores and
      loads all cookies, independent if they are session cookies or not.
      Session cookies are cookies without expiry date and they are meant
      to be alive and existing for this "session" only.
+     Available as of cURL 7.9.7.
     </para>
    </listitem>
   </varlistentry>
@@ -367,6 +405,7 @@
     <para>
      &true; to convert Unix newlines to CRLF newlines
      on transfers.
+     Available as of cURL 7.1.
     </para>
    </listitem>
   </varlistentry>
@@ -378,7 +417,7 @@
    <listitem>
     <para>
      Pass a <type>string</type> naming a file with the concatenation of
-     CRL (Certificate Revocation List) (in PEM format)
+	 CRL (Certificate Revocation List) (in PEM format)
      to use in the certificate validation that occurs during the SSL exchange.
      When cURL is built to use GnuTLS,
      there is no way to influence the use of CRL passed
@@ -391,7 +430,7 @@
      Also note that <constant>CURLOPT_CRLFILE</constant> implies
      <constant>CURLSSLOPT_NO_PARTIALCHAIN</constant>
      as of cURL 7.71.0 due to an OpenSSL bug.
-     Available as of cURL 7.19.0
+     Available as of cURL 7.19.0.
     </para>
    </listitem>
   </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/curl/constants_curl_setopt.xml
```diff
@@ -68,8 +68,10 @@
    </term>
    <listitem>
     <para>
-     Pass the filename for cURL to use as the Alt-Svc cache file to read existing cache contents from and
-     possibly also write it back to a after a transfer, unless <constant>CURLALTSVC_READONLYFILE</constant>
+     Pass a <type>string</type> with the filename for cURL to use
+     as the Alt-Svc cache file to read existing cache contents from
+     and possibly also write it back to a after a transfer,
+     unless <constant>CURLALTSVC_READONLYFILE</constant>
```
cmb69, line 74: It might be better not to mix line wrap changes with content changes (harder for translators).

reference/curl/constants_curl_setopt.xml
```diff
@@ -127,10 +130,10 @@
    </term>
    <listitem>
     <para>
-     Provides AWS V4 signature authentication on HTTP(S) header.
+     Provides AWS V4 signature authentication on HTTP(S) header as a <type>string</type>.
```
cmb69, line 133: Could also use `&string;`, but `<type>string</type>` might be preferred nowadays.

reference/curl/constants_curl_setopt.xml
```diff
@@ -154,6 +158,8 @@
     <para>
      The size of the buffer to use for each read. There is no guarantee
      this request will be fulfilled, however.
+     This option accepts any value that can be cast to a valid <type>int</type>.
+     Defaults to <constant>CURL_MAX_WRITE_SIZE</constant> (16kB).
```
cmb69, line 162: Might be "dangerous" to include the actual size (16kB) here, since this might change upstream. At least maybe say
```suggestion
     Defaults to <constant>CURL_MAX_WRITE_SIZE</constant> (currently, 16kB).
```

cmb69 (commented):
Wow, a lot of work. Thank you!

Just left a couple of notes, before `CURLE_OPERATION_TIMEOUTED`. :)

reference/curl/constants_curl_setopt.xml
```diff
@@ -378,7 +417,7 @@
    <listitem>
     <para>
      Pass a <type>string</type> naming a file with the concatenation of
-     CRL (Certificate Revocation List) (in PEM format)
+	 CRL (Certificate Revocation List) (in PEM format)
```
Girgias, line 420: Seems to have an unintended indent change from spaces to tab here

reference/curl/constants_curl_setopt.xml
```diff
@@ -614,6 +666,7 @@
     <para>
      Like <constant>CURLOPT_RANDOM_FILE</constant>, except a filename
      to an Entropy Gathering Daemon socket.
+     Available as of cURL 7.7 and deprecated as of cURL 7.84.0.
```
Girgias, line 669: Maybe?
```suggestion
     Available as of cURL 7.7.0 and deprecated as of cURL 7.84.0.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -624,12 +677,17 @@
...
-     is set, a header containing all supported encoding types is sent.
-     Available as of cURL 7.10.
+     The contents of the <literal>Accept-Encoding: </literal> header as a <type>string</type>.
+     This enables decoding of the response. Supported encodings are:
+     <simplelist type="inline">
+      <member><literal>identity</literal></member>
+      <member><literal>deflate</literal></member>
+      <member><literal>gzip</literal></member>
+     </simplelist>
+     .
```
Girgias, line 687: Doesn't this cause whitespace?

reference/curl/constants_curl_setopt.xml
```diff
@@ -654,8 +713,9 @@
    <listitem>
     <para>
      &true; to fail verbosely if the HTTP code returned
-     is greater than or equal to 400. The default behavior is to return
+     is greater than or equal to <literal>400</literal>. The default behavior is to return
      the page normally, ignoring the code.
+     Available as of cURL 7.1.
```
Girgias, line 718: Should this have a patch version?

reference/curl/constants_curl_setopt.xml
```diff
@@ -795,6 +862,9 @@
     <para>
      An alias of
      <constant>CURLOPT_TRANSFERTEXT</constant>. Use that instead.
+     Available as of cURL 7.1, deprecated as of cURL 7.11.1
+     and last available in cURL 7.15.5.
+     Removed as of PHP 7.3.
```
Girgias, line 867:
```suggestion
     Removed as of PHP 7.3.0.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -977,9 +1051,10 @@
...
    <listitem>
     <para>
-     &true; to first try an EPSV command for FTP
-     transfers before reverting back to PASV. Set to &false;
-     to disable EPSV.
+     Set to &true; to first try an <literal>EPSV</literal> command for FTP
+     transfers before reverting back to <literal>PASV</literal>.
+     Set to &false; to disable <literal>EPSV</literal>.
+    Available as of cURL 7.9.2.
     </para>
```
Girgias, line 1058: Indent slightly off here

reference/curl/constants_curl_setopt.xml
```diff
@@ -1143,21 +1224,24 @@
...
+      <member><constant>CURLAUTH_BASIC</constant></member>
+      <member><constant>CURLAUTH_DIGEST</constant></member>
+      <member><constant>CURLAUTH_GSSNEGOTIATE</constant></member>
+      <member><constant>CURLAUTH_NTLM</constant></member>
+      <member><constant>CURLAUTH_AWS_SIGV4</constant></member>
+      <member><constant>CURLAUTH_ANY</constant></member>
+      <member><constant>CURLAUTH_ANYSAFE</constant></member>
+     </simplelist>
+     .
+     If more than one method is used, cURL will poll the server to see
```
Girgias, line 1238: Ditto whitespace concern?

reference/curl/constants_curl_setopt.xml
```diff
@@ -1380,16 +1471,20 @@
...
-     <literal>"private".</literal>.
-     If the string does not match one of these,
-     <literal>"private"</literal> is used. Setting this option to &null;
+     <simplelist type="inline">
+      <member><literal>clear</literal></member>
+      <member><literal>safe</literal></member>
+      <member><literal>confidential</literal></member>
+      <member><literal>private</literal></member>
+     </simplelist>
+     .
```
Girgias, line 1482: Ditto

reference/curl/constants_curl_setopt.xml
```diff
@@ -1695,9 +1818,11 @@
...
    <listitem>
     <para>
-     &true; to be completely silent with regards to
-     the cURL functions.
-     Removed as of cURL 7.15.5; use <constant xmlns="http://docbook.org/ns/docbook">CURLOPT_RETURNTRANSFER</constant> instead.
+     Set to &true; to be completely silent with regards to the cURL functions.
+     Use <constant>CURLOPT_RETURNTRANSFER</constant> instead.
+     Available as of cURL 7.1, deprecated as of cURL 7.8
+     and last available in cURL 7.15.5.
+     Removed as of PHP 7.3.
```
Girgias, line 1825:
```suggestion
     Removed as of PHP 7.3.0.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -2025,8 +2168,8 @@
    </term>
    <listitem>
     <para>
-     A callback accepting five parameters.
-     The first is the cURL resource, the second is the total number of
+     A <type>callable</type> accepting five parameters and returning an <type>int</type>.
+     The first parameter is the cURL resource, the second is the total number of
      bytes expected to be downloaded in this transfer, the third is
      the number of bytes downloaded so far, the fourth is the total
```
Girgias, line 2174: Maybe this should use method synopsis?

reference/curl/constants_curl_setopt.xml
```diff
@@ -2627,7 +2813,9 @@
    </term>
    <listitem>
     <para>
-     A filename to be used to seed the random number generator for SSL.
+     A <type>string</type> with a filename to be used
+     to seed the random number generator for SSL.
+     Available as of cURL 7.7.0 and deprecated as of 7.84.0.
```
Girgias, line 2818:
```suggestion
     Available as of cURL 7.7.0 and deprecated as of cURL 7.84.0.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -2665,14 +2855,15 @@
...
-     A callback accepting three parameters.
-     The first is the cURL resource, the second is a
-     stream resource provided to cURL through the option
-     <constant>CURLOPT_INFILE</constant>, and the third is the maximum
-     amount of data to be read. The callback must return a string
+     A <type>callable</type> accepting three parameters and returning a <type>string</type>.
+     The first parameter is the cURL resource, the second is a
+     stream <type>resource</type> provided to cURL through the option
+     <constant>CURLOPT_READDATA</constant>, and the third is the maximum
+     amount of data to be read. The callback must return a <type>string</type>
```
Girgias, line 2862: Ditto methodsynopsis remark

reference/curl/constants_curl_setopt.xml
```diff
@@ -3912,11 +4186,12 @@
...
     <para>
-     A callback accepting two parameters.
-     The first is the cURL resource, and the second is a
-     string with the data to be written. The data must be saved by
-     this callback. It must return the exact number of bytes written
+     A <type>callable</type> accepting two parameters and returning an <type>int</type>.
+     The first parameter is the cURL resource, and the second is a
+     <type>string</type> with the data to be written. The data must be saved by
+     this callback. The callback must return the exact number of bytes written
      or the transfer will be aborted with an error.
```
Girgias, line 4193: Ditto method synopsys remark

reference/curl/constants_curl_setopt.xml
```diff
@@ -3949,9 +4229,14 @@
...
-     Has a similar purpose as <constant>CURLOPT_PROGRESSFUNCTION</constant> but is more modern
-     and the preferred option from cURL.
+     A <type>callable</type> accepting five parameters and returning an <type>int</type>.
+     The first parameter is the cURL resource, the second is the total number of
+     bytes expected to be downloaded in this transfer, the third is
+     the number of bytes downloaded so far, the fourth is the total
+     number of bytes expected to be uploaded in this transfer, and the
+     fifth is the number of bytes uploaded so far.
+     Return <literal>1</literal> to abort the transfer
+     and set a <constant>CURLE_ABORTED_BY_CALLBACK</constant> error.
```
Girgias, line 4239: Ditto methodsynopsis remark

Girgias (changes requested):
There also seems to be missing a bunch of patch versions for cURL availability (mainly `7.1`)

reference/curl/constants_curl_setopt.xml
haszi: Hmm. You're right, there is a space between the last item on the list and the period in the rendered file. I'll move the periods right after the `</simplelist>` tags.

reference/curl/constants_curl_setopt.xml
haszi: Good question. I've got all the version info from the [curl website](https://curl.se/libcurl/c/symbols-in-versions.html) and it seems that up until cURL 7.11.0, none of the major versions were using a patch version on the ".0" versions. Even the [release notes](https://curl.se/ch/7.10.html) refer to these versions with a patch version.

I'm OK with either leaving it as is or adding the ".0"s to these versions. Which one do you prefer?

reference/curl/constants_curl_setopt.xml
cmb69: If I remember correctly, we deliberately decided to not use the entities in the function signatures (where `<type>string</type>` is supposed to occur often).

Maybe @Girgias knows whether it's preferable to use type elements or the respective entities elsewhere.

reference/curl/constants_curl_setopt.xml
Girgias: I would prefer to add the `.0` to be consistent in our docs :)

reference/curl/constants_curl_setopt.xml
Girgias: It is better to use the `<type>` element, as some translations translate the entity (don't ask me why doc-fr does it, but it does).

cmb69: With the same reasoning, they can translate the element as well. And soon we're back to
````
beschneide(zeichenkette $zeichenkette, zeichenkette $zeichen = " \n\r\t\v\x00"): zeichenkette
````

Comment (Girgias) [reply reference/curl/constants_curl_setopt.xml:133]: I'm pretty sure some of the QA scripts that can be used for translation check this. Usually the entity is better used when we want to talk about the "concept" of a string (or other type) rather then directly referring to it as a PHP type.

But the distinction is not useful most of the time. So maybe we should just fix translations that do that.

File: reference/curl/constants_curl_setopt.xml (hunk 1 above)
Comment (haszi) [reply reference/curl/constants_curl_setopt.xml:74]: Should I rebase and remove all the line wrap changes or just do it in a follow-up commit on this PR?

Comment (cmb69) [reply reference/curl/constants_curl_setopt.xml:74]: If the PR finally gets squashed merged, it doesn't matter; do whatever is easier for you. :)

**Comment by haszi** [comment]
I think I have addressed all review comments such as:
- removed line wrap changes (all that I could find)
- added cURL .0 patch version numbers
- replaced callback descriptions with `<methodsynopsis>` elements 
- clarified that `CURLOPT_BUFFERSIZE` is *currently* 16KB (ie. can change)
- replaced tab with spaces
- added missing whitespace
- removed line break between `</simplelist>` tags and periods
- clarified that deprecation for `CURLOPT_RANDOM_FILE` is the cURL version number

File: reference/curl/constants_curl_setopt.xml [hunk 18]
```diff
@@ -198,8 +196,7 @@
    </term>
    <listitem>
     <para>
-     A <type>string</type> with a directory that holds multiple CA certificates.
-     Use this option alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
+     A <type>string</type> with a directory that holds multiple CA certificates. Use this option alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
```
Girgias, line 199: Minor nit, line break at the full stop before "Use this option [...]".

reference/curl/constants_curl_setopt.xml
```diff
@@ -256,8 +253,7 @@
    </term>
    <listitem>
     <para>
-     The number of milliseconds to wait while trying to connect.
-     Use <literal>0</literal> to wait indefinitely.
+     The number of milliseconds to wait while trying to connect. Use <literal>0</literal> to wait indefinitely.
```
Girgias, line 256: Ditto line break

reference/curl/constants_curl_setopt.xml
```diff
@@ -318,12 +313,10 @@
    </term>
    <listitem>
     <para>
-     A <type>string</type> with the name of the file containing the cookie data.
-     The cookie file can be in Netscape format,
-     or just plain HTTP-style headers dumped into a file.
+     A <type>string</type> with the name of the file containing the cookie data. The cookie file can be in Netscape format, or just plain HTTP-style headers dumped into a file.
```
Girgias, line 316: Line break

reference/curl/constants_curl_setopt.xml
```diff
@@ -1549,8 +1537,7 @@
    <listitem>
     <para>
      Can be used to set protocol specific login options, such as the
-     preferred authentication mechanism via <literal>AUTH=NTLM</literal>
-     or <literal>AUTH=*</literal>, and should be used in conjunction with the
+     preferred authentication mechanism via <literal>AUTH=NTLM</literal> or <literal>AUTH=*</literal>, and should be used in conjunction with the
```
Girgias, line 1540: Is the `*` in AUTH a literal or a placeholder?

Girgias (commented):
It looks like we got rid of some semantic breaks :/

reference/curl/constants_curl_setopt.xml
Girgias: Ah was this a revert of a semantic line break?

Looking just from the previous review state was unclear ('thanks GitHub)

haszi: Yeah, all these line break changes all reverts.

reference/curl/constants_curl_setopt.xml
haszi: It is a literal and refers to any authorization method being acceptable.

reference/curl/constants_curl_setopt.xml
```diff
@@ -191,8 +196,8 @@
    </term>
    <listitem>
     <para>
-     A directory that holds multiple CA certificates. Use this option
-     alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
+     A <type>string</type> with a directory that holds multiple CA certificates. Use this option alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
```
Girgias, line 199:
```suggestion
     A <type>string</type> with a directory that holds multiple CA certificates.
     Use this option alongside <constant>CURLOPT_SSL_VERIFYPEER</constant>.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -239,12 +253,13 @@
    </term>
    <listitem>
     <para>
-     The number of milliseconds to wait while trying to connect. Use 0 to
-     wait indefinitely.
-     <!-- http://curl.haxx.se/libcurl/c/curl_easy_setopt.html -->
-     If libcurl is built to use the standard system name resolver, that
+     The number of milliseconds to wait while trying to connect. Use <literal>0</literal> to wait indefinitely.
```
Girgias, line 256:
```suggestion
     The number of milliseconds to wait while trying to connect.
     Use <literal>0</literal> to wait indefinitely.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -297,11 +313,10 @@
    </term>
    <listitem>
     <para>
-     The name of the file containing the cookie data. The cookie file can
-     be in Netscape format, or just plain HTTP-style headers dumped into
-     a file.
-     If the name is an empty string, no cookies are loaded, but cookie
+     A <type>string</type> with the name of the file containing the cookie data. The cookie file can be in Netscape format, or just plain HTTP-style headers dumped into a file.
```
Girgias, line 316:
```suggestion
     A <type>string</type> with the name of the file containing the cookie data.
     The cookie file can be in Netscape format, or just plain HTTP-style headers dumped into a file.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -312,8 +327,8 @@
    </term>
    <listitem>
     <para>
-     The name of a file to save all internal cookies to when the
-     handle's destructor is called.
+     A <type>string</type> with the name of a file to save all internal cookies to when the handle's destructor is called.
```
Girgias, line 330:
```suggestion
     A <type>string</type> with the name of a file to save all internal cookies to when
     the handle's destructor is called.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -459,8 +492,8 @@
    </term>
    <listitem>
     <para>
-     &true; to not allow URLs that include a username. Usernames are allowed by default (0).
-     Available as of PHP 7.3.0 and cURL 7.61.0
+     &true; to not allow URLs that include a username. Usernames are allowed by default.
+     Available as of PHP 7.3.0 and cURL 7.61.0.
```
Girgias, line 496:
```suggestion
     &true; to not allow URLs that include a username.
     Usernames are allowed by default.
     Available as of PHP 7.3.0 and cURL 7.61.0.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -496,9 +532,9 @@
    </term>
    <listitem>
     <para>
-     Set the local IPv4 address that the resolver should bind to. The argument
-     should contain a single numerical IPv4 address as a string.
-     Available as of PHP 7.0.7 and cURL 7.33.0
+     Set the local IPv4 address that the resolver should bind to. The argument should contain a single numerical IPv4 address.
+     This option accepts a <type>string</type> or &null;.
```
Girgias, line 536:
```suggestion
     Set the local IPv4 address that the resolver should bind to.
     The argument should contain a single numerical IPv4 address.
     This option accepts a <type>string</type> or &null;.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -509,9 +545,9 @@
    </term>
    <listitem>
     <para>
-     Set the local IPv6 address that the resolver should bind to. The argument
-     should contain a single numerical IPv6 address as a string.
-     Available as of PHP 7.0.7 and cURL 7.33.0
+     Set the local IPv6 address that the resolver should bind to. The argument should contain a single numerical IPv6 address.
+     This option accepts a <type>string</type> or &null;.
```
Girgias, line 549:
```suggestion
     Set the local IPv6 address that the resolver should bind to.
     The argument should contain a single numerical IPv6 address.
     This option accepts a <type>string</type> or &null;.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -666,8 +713,9 @@
    </term>
    <listitem>
     <para>
-     The file that the transfer should be written to. The default
-     is <constant>STDOUT</constant> (the browser window).
+     Accepts a file handle <type>resource</type>
+     to the file that the transfer should be written to. The default is <constant>STDOUT</constant> (the browser window).
```
Girgias, line 717:
```suggestion
     to the file that the transfer should be written to.
     The default is <constant>STDOUT</constant> (the browser window).
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -817,12 +872,11 @@
...
    <listitem>
     <para>
-     The value which will be used to get the IP address to use
-     for the FTP "PORT" instruction. The "PORT" instruction tells
+     A <type>string</type> which will be used to get the IP address to use for the FTP <literal>PORT</literal> instruction. The <literal>PORT</literal> instruction tells
      the remote server to connect to our specified IP address.  The
-     string may be a plain IP address, a hostname, a network
-     interface name (under Unix), or just a plain '-' to use the
-     systems default IP address.
+     <type>string</type> may be a plain IP address, a hostname, a network interface name (under Unix), or just a plain <literal>-</literal> to use the system's default IP address.
```
Girgias, line 877:
```suggestion
     <type>string</type> may be a plain IP address, a hostname,
     a network interface name (under Unix),
     or just a plain <literal>-</literal> to use the system's default IP address.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -964,9 +1016,8 @@
    </term>
    <listitem>
     <para>
-     &true; to use EPRT (and LPRT) when doing active
-     FTP downloads. Use &false; to disable EPRT and LPRT and use PORT
-     only.
+     Set to &true; to use <literal>EPRT</literal> (and <literal>LPRT</literal>) when doing active FTP downloads. Set to &false; to disable <literal>EPRT</literal> and <literal>LPRT</literal> and use <literal>PORT</literal> only.
```
Girgias, line 1019:
```suggestion
     Set to &true; to use <literal>EPRT</literal> (and <literal>LPRT</literal>) when doing active FTP downloads.
     Set to &false; to disable <literal>EPRT</literal> and <literal>LPRT</literal> and use <literal>PORT</literal> only.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -977,9 +1028,8 @@
    </term>
    <listitem>
     <para>
-     &true; to first try an EPSV command for FTP
-     transfers before reverting back to PASV. Set to &false;
-     to disable EPSV.
+     Set to &true; to first try an <literal>EPSV</literal> command for FTP transfers before reverting back to <literal>PASV</literal>. Set to &false; to disable <literal>EPSV</literal>.
```
Girgias, line 1031:
```suggestion
     Set to &true; to first try an <literal>EPSV</literal> command for FTP transfers before reverting back to <literal>PASV</literal>.
     Set to &false; to disable <literal>EPSV</literal>.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -1292,8 +1372,8 @@
    </term>
    <listitem>
     <para>
-     The name of the outgoing network interface to use. This can be an
-     interface name, an IP address or a host name.
+     Set to a <type>string</type> with the name of the outgoing network interface to use. This can be an interface name, an IP address or a host name.
```
Girgias, line 1375:
```suggestion
     Set to a <type>string</type> with the name of the outgoing network interface to use.
     This can be an interface name, an IP address or a host name.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -1352,11 +1433,10 @@
    </term>
    <listitem>
     <para>
-     &true; to keep sending the request body if the HTTP code returned is
-     equal to or larger than 300. The default action would be to stop sending
+     Set to &true; to keep sending the request body if the HTTP code returned is equal to or larger than <literal>300</literal>. The default action would be to stop sending
```
Girgias, line 1436:
```suggestion
     Set to &true; to keep sending the request body if the HTTP code returned is equal to or larger than <literal>300</literal>.
     The default action would be to stop sending
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -1695,9 +1798,12 @@
    </term>
    <listitem>
     <para>
-     &true; to be completely silent with regards to
+     Set to &true; to be completely silent with regards to
      the cURL functions.
-     Removed as of cURL 7.15.5; use <constant xmlns="http://docbook.org/ns/docbook">CURLOPT_RETURNTRANSFER</constant> instead.
+     Use <constant>CURLOPT_RETURNTRANSFER</constant> instead.
+     Available as of cURL 7.1, deprecated as of cURL 7.8
```
Girgias, line 1804:
```suggestion
     Available as of cURL 7.1.0, deprecated as of cURL 7.8.0
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -2713,8 +3025,8 @@
    </term>
    <listitem>
     <para>
-     The contents of the <literal>"Referer: "</literal> header to be used
-     in a HTTP request.
+     A <type>string</type> with the contents of the <literal>Referer: </literal> header to be used in a HTTP request.
```
Girgias, line 3028:
```suggestion
     A <type>string</type> with the contents of the <literal>Referer: </literal>
     header to be used in a HTTP request.
```

Girgias (approved):
Readding semantic breaks where we are already editing it so it affects translations anyway

PR #3819, opened 2024-10-01
Fix Example section, Notes section before changelog

sy-records:


cmb69 (approved):
Thank you!

cmb69:
> Do we have the preferred order of these sections documented anywhere?

See https://github.com/php/doc-base/blob/master/RFC/skeletons/function.xml (and maybe elsewhere).

> Maybe it should, commented out?

+1

Girgias:
We have a QA script that should check this, but it is brittle as it doesn't follow XIncludes

PR #3820, opened 2024-10-01
Re-enable section-order QA script

jimwins:


cmb69:
I assume it had a reason why that script had been disabled via d9a3445921474a99c8defd392b506745f51f3dbc. Maybe @Girgias remembers?

Girgias:
This feels like something I was encountering because of the migration to DocBook 5.2

PR #3824, opened 2024-10-03
Update stream-context-get-params.xml: Adding possible `false` return value

ThomasLandauer:
Page: https://www.php.net/manual/en/function.stream-context-get-params.php

Info is taken from user comment https://www.php.net/manual/en/function.stream-context-get-params.php#118399 - I didn't doublecheck this.

Wording is taken from https://www.php.net/manual/en/function.stream-filter-prepend.php#refsect1-function.stream-filter-prepend-returnvalues

### DIFF
--- a/reference/stream/functions/stream-context-get-params.xml
+++ b/reference/stream/functions/stream-context-get-params.xml
@@ -39,6 +39,9 @@
   <para>
    Returns an associate array containing all context options and parameters.
   </para>
+  <para>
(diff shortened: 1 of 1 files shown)


Girgias (changes requested):
This is a TypeError failure, which is something that we do not document as it had undefined behaviour and since 8.0 consistently throws a TypeError.

PR #3825, opened 2024-10-03
Update string.xml Fix the code example

mmalferov:
There is no index 4 in the `$arr` variable, which has an index 3, maybe it was meant $arr[3][2]:

```
Array
(
    [0] => 1
    [1] => 2
    [2] => 3
    [3] => Array
        (
            [0] => 41
            [1] => 42
            [2] => 43
        )

    [key] => Indexed value
    [const-key] => Key with minus sign
    [foo] => Array
        (
            [0] => foo1
            [1] => foo2
            [2] => foo3
        )

)
```


Girgias (approved):
Thank you!

PR #3827, opened 2024-10-03
Updated paragraph on initialization of static variables

agroholism:
Updated information about restrictions on initialization of static variables to comply with PHP 8.3

### DIFF
--- a/language/variables.xml
+++ b/language/variables.xml
@@ -469,8 +469,8 @@ function test()
    </para>
 
    <para>
-    Before PHP 8.3, you could only initialize a static variable using 
-    a constant expression. As of PHP 8.3, dynamic expressions
+    Prior to PHP 8.3.0, you could only initialize a static variable using
+    a constant expression. As of PHP 8.3.0, dynamic expressions
     (e.g., function calls) are also allowed:
    </para>
    <para>
@@ -482,7 +482,7 @@ function test()
 function foo(){
     static $int = 0;          // correct 
     static $int = 1+2;        // correct
-    static $int = sqrt(121);  // correct in PHP >= 8.3
+    static $int = sqrt(121);  // correct as of PHP 8.3.0
 
     $int++;
     echo $int;
--- a/language/variables.xml
+++ b/language/variables.xml
@@ -469,9 +469,9 @@ function test()
    </para>
 
    <para>
-    Static variables can be assigned values which are the
-    result of constant expressions, but dynamic expressions, such as function
-    calls, will cause a parse error.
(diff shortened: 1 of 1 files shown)


language/variables.xml
```diff
@@ -469,9 +469,9 @@ function test()
    </para>
 
    <para>
-    Static variables can be assigned values which are the
-    result of constant expressions, but dynamic expressions, such as function
-    calls, will cause a parse error.
+    Before PHP 8.3, you could only initialize a static variable using 
+    a constant expression. As of PHP 8.3, dynamic expressions
```
cmb69, line 473: Per the style guidelines, this should be
```suggestion
    Prior to PHP 8.3.0, you could only initialize a static variable using 
    a constant expression. As of PHP 8.3.0, dynamic expressions
```

language/variables.xml
```diff
@@ -469,9 +469,9 @@ function test()
...
+    (e.g., function calls) are also allowed:
```
cmb69, line 474: Not sure if that comma makes sense.

language/variables.xml
```diff
@@ -482,7 +482,7 @@ function test()
 function foo(){
     static $int = 0;          // correct 
     static $int = 1+2;        // correct
-    static $int = sqrt(121);  // wrong  (as it is a function)
+    static $int = sqrt(121);  // correct in PHP >= 8.3
```
cmb69, line 485:
```suggestion
    static $int = sqrt(121);  // correct as of PHP 8.3.0
```

language/variables.xml
```diff
@@ -520,12 +520,6 @@ var_dump(Bar::counter()); // int(4), prior to PHP 8.1.0 int(2)
 ]]>
     </programlisting>
    </example>
-
-   <note>
-    <para>
-     Static declarations are resolved in compile-time.
```
cmb69, line 526: It's okay to remove this; alternatively we could have a (lengthy) explanation about what is resolved during compile time, and what is resolved during run time).

cmb69 (commented):
Thank you for the PR!

language/variables.xml
agroholism: Thank you for your edits!

language/variables.xml
agroholism: Yes, I am also a bit confused. In American English, it is recommended to use "e.g." followed by a comma. However, in existing documents, there are a lot of examples of both the usage with and without a comma. But I agree with you on this.

language/variables.xml
agroholism: I'm not sure if a long explanation will provide much value for the reader.

PR #3828, opened 2024-10-04
Asymmetric Visiblity and Final properties

Crell:
First stab at documenting asymmetric visibility and final properties.  (Technically final properties came in with hooks, but that's a much bigger task and it does impact aviz, so I included it here.)

### DIFF
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -59,10 +59,9 @@ class ChildClass extends BaseClass {
     </programlisting>
    </example>
   </para>
-  <para>
-   <example>
-    <title>Final property example as of PHP 8.4.0</title>
-    <programlisting role="php">
+ <example>
+  <title>Final property example as of PHP 8.4.0</title>
+  <programlisting role="php">
 <![CDATA[
 <?php
 class BaseClass {
@@ -75,13 +74,11 @@ class ChildClass extends BaseClass {
 // Results in Fatal error: Cannot override final property BaseClass::$test
 ?>
 ]]>
-    </programlisting>
-   </example>
-  </para>
-  <para>
-   <example xml:id="language.oop5.final.example.php81">
-    <title>Final constants example as of PHP 8.1.0</title>
-    <programlisting role="php">
+  </programlisting>
+ </example>
+ <example xml:id="language.oop5.final.example.php81">
+  <title>Final constants example as of PHP 8.1.0</title>
+  <programlisting role="php">
 <![CDATA[
 <?php
 class Foo
@@ -97,9 +94,8 @@ class Bar extends Foo
 // Fatal error: Bar::X cannot override final constant Foo::X
 ?>
 ]]>
-    </programlisting>
-   </example>
-  </para>
+  </programlisting>
+ </example>
 
   <note>
    <simpara>
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -108,7 +108,7 @@ class Bar extends Foo
   </note>
   <note>
    <simpara>
-    A property that is declared <literal>private(set)</literal> is implicitly <literal>final</literal>.
+    A property that is declared <link linkend="language.oop5.visibility-members-aviz"><literal>private(set)</literal></link> is implicitly <literal>final</literal>.
    </simpara>
   </note>
  </sect1>
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -3,7 +3,7 @@
  <sect1 xml:id="language.oop5.final" xmlns="http://docbook.org/ns/docbook">
   <title>Final Keyword</title>
   <para>
-   The final keyword prevents child classes from overriding a method or constant by
+   The final keyword prevents child classes from overriding a method, property, or constant by
    prefixing the definition with <literal>final</literal>. If the class
    itself is being defined final then it cannot be extended.
   </para>
@@ -60,6 +60,25 @@ class ChildClass extends BaseClass {
    </example>
   </para>
   <para>
+   <example>
+    <title>Final property example as of PHP 8.4.0</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+class BaseClass {
+   final protected string $test;
+}
+
+class ChildClass extends BaseClass {
+    public string $test;
+}
+// Results in Fatal error: Cannot override final property BaseClass::$test
+?>
+]]>
+    </programlisting>
+   </example>
+  </para>
+  <para>
    <example xml:id="language.oop5.final.example.php81">
     <title>Final constants example as of PHP 8.1.0</title>
     <programlisting role="php">
@@ -84,10 +103,12 @@ class Bar extends Foo
 
   <note>
    <simpara>
-    Properties cannot be declared final: only classes, methods, and constants (as of PHP 8.1.0) may be declared as final.
-   </simpara>
-   <simpara>
     As of PHP 8.0.0, private methods may not be declared final except for the constructor.
+   </simpara>
+  </note>
+  <note>
+   <simpara>
+    A property that is declared <literal>private(set)</literal> is implicitly <literal>final</literal>.
    </simpara>
   </note>
  </sect1>
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -103,7 +103,7 @@ class Bar extends Foo
 
   <note>
    <simpara>
-    As of PHP 8.0.0, private methods may not be declared final except for the constructor.
+    As of PHP 8.0.0, private methods may not be declared final except for the <link linkend="#language.oop5.decon.constructor">constructor</link>.
    </simpara>
   </note>
   <note>
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -99,7 +99,7 @@ class Bar extends Foo
 
   <note>
    <simpara>
-    As of PHP 8.0.0, private methods may not be declared final except for the <link linkend="#language.oop5.decon.constructor">constructor</link>.
+    As of PHP 8.0.0, private methods may not be declared final except for the <link linkend="language.oop5.decon.constructor">constructor</link>.
    </simpara>
   </note>
   <note>
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -192,7 +192,12 @@ Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be ac
   <sect2 xml:id="language.oop5.properties.readonly-properties">
    <title>Readonly properties</title>
    <para>
-    As of PHP 8.1.0, a property can be declared with the <code>readonly</code> modifier, which prevents modification of the property after initialization.
+    As of PHP 8.1.0, a property can be declared with the <literal>readonly</literal> modifier,
+    which prevents modification of the property after initialization.  Prior to PHP 8.4.0
+    a <literal>readonly</literal> property is implicitly private-set, and may only be written to
+    from the same class.  As of PHP 8.4.0, <literal>readonly</literal> properties are implicitly
+    <literal>protected(set)</literal>, so may be set from child classes.  That may be overridden
+    explicitly if desired.
     <example>
      <title>Example of readonly properties</title>
      <programlisting role="php">
--- a/language/oop5/properties.xml
+++ b/language/oop5/properties.xml
@@ -196,7 +196,8 @@ Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be ac
     which prevents modification of the property after initialization.  Prior to PHP 8.4.0
     a <literal>readonly</literal> property is implicitly private-set, and may only be written to
     from the same class.  As of PHP 8.4.0, <literal>readonly</literal> properties are implicitly
-    <literal>protected(set)</literal>, so may be set from child classes.  That may be overridden
+    <link linkend="language.oop5.visibility-members-aviz"><literal>protected(set)</literal></link>,
+    so may be set from child classes.  That may be overridden
     explicitly if desired.
     <example>
      <title>Example of readonly properties</title>
--- a/language/oop5/visibility.xml
+++ b/language/oop5/visibility.xml
@@ -78,6 +78,117 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
      </programlisting>
     </example>
    </para>
+   <para>
+    As of PHP 8.4, properties may also have their
+    visibility set asymmetrically, with a different scope for
+    reading (<emphasis>get</emphasis>) and writing (<emphasis>set</emphasis>).
+    Specifically, the <emphasis>set</emphasis> visibility may be
+    specified separately, provided it is not more permissive than the
+    default visibility.
+   </para>
+   <para>
+    <example>
+     <title>Asymmetric Property visibility</title>
+     <programlisting role="php">
+      <![CDATA[
+<?php
+class Book
+{
+    public function __construct(
+        public private(set) string $title,
+        public protected(set) string $author,
+        protected private(set) int $pubYear,
+    ) {}
+}
+
+class SpecialBook extends Book
+{
+    public function update(string $author, int $year): void
+    {
+        $this->author = $author; // OK
+        $this->pubYear = $year; // Fatal Error
+    }
+}
+
+$b = new Book('How to PHP', 'Peter H. Peterson', 2024);
+
+echo $b->title; // Works
+echo $b->author; // Works
+echo $b->pubYear; // Fatal Error
+
+$b->title = 'How not to PHP'; // Fatal Error
+$b->author = 'Pedro H. Peterson'; // Fatal Error
+$b->pubYear = 2023; // Fatal Error
+?>
+]]>
+     </programlisting>
+    </example>
+   </para>
+   <para>There are a few caveats with asymmetric visibility to be aware of.</para>
+   <itemizedlist>
+    <listitem>
+     <simpara>
+      Only typed properties may have a separate <emphasis>set</emphasis> visibility.
+     </simpara>
+     <simpara>
+      The <emphasis>set</emphasis> visibility must be the same
+      as <emphasis>get</emphasis> or more restrictive. That is,
+      <code>public protected(set)</code> and <code>protected protected(set)</code>
+      are allowed, but <code>protected public(set) will cause a syntax error.</code>
+     </simpara>
+     <simpara>
+      If a property is <literal>public</literal> writeable, then the main visibility may be
+      omitted.  That is, <code>public private(set)</code> and <code>private(set)</code>
+      will have the same result.
+     </simpara>
+     <simpara>
+      A property with <literal>private(set)</literal> visibility
+      is automatically <literal>final</literal>, and may not be redeclared in a child class.
+     </simpara>
+     <simpara>
+      Obtaining a reference to a property follows <emphasis>set</emphasis> visibility, not <emphasis>get</emphasis>.
+      That is because a reference may be used to modify the property value.
+     </simpara>
+     <simpara>
+      Similarly, trying to write to an array property involves both a <emphasis>get</emphasis> and
+      <emphasis>set</emphasis> operation internally, and therefore will follow the <emphasis>set</emphasis>
+      visibility, as that is always the more restrictive.
+     </simpara>
+    </listitem>
+   </itemizedlist>
+   <para>
+    When a class extends another, the child class may redefine
+    any property that is not <literal>final</literal>.  When doing so,
+    it may widen either the main visibility or the <emphasis>set</emphasis>
+    visibility, provided that the new visibility is the same or wider
+    than the parent class.  However, be aware that if a <literal>private</literal>
+    property is overridden, it does not actually change the parent's property
+    but creates a new property with a different internal name.
+   </para>
+   <para>
+    <example>
+     <title>Asymmetric Property inheritance</title>
+     <programlisting role="php">
+      <![CDATA[
+<?php
+class Book
+{
+    protected string $title;
+    public protected(set) string $author;
+    protected private(set) int $pubYear;
+}
+
+class SpecialBook extends Book
+{
+    public protected(set) $title; // OK, as reading is wider and writing the same.
+    public string $author; // OK, as reading is the same and writing is wider.
+    public protected(set) int $pubYear; // Fatal Error. private(set) properties are final.
+}
+?>
+]]>
+     </programlisting>
+    </example>
+   </para>
   </sect2>
 
   <sect2 xml:id="language.oop5.visiblity-methods">
--- a/language/oop5/visibility.xml
+++ b/language/oop5/visibility.xml
@@ -126,8 +126,8 @@ $b->pubYear = 2023; // Fatal Error
       </programlisting>
      </example>
     </para>
+    <para>There are a few caveats regarding asymmetric visibility:</para>
     <itemizedlist>
-     <title>Caveats regarding asymmetric visibility</title>
      <listitem>
       <simpara>
        Only typed properties may have a separate <emphasis>set</emphasis> visibility.
--- a/language/oop5/visibility.xml
+++ b/language/oop5/visibility.xml
@@ -90,7 +90,7 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
     <example>
      <title>Asymmetric Property visibility</title>
      <programlisting role="php">
-      <![CDATA[
+<![CDATA[
 <?php
 class Book
 {
@@ -169,7 +169,7 @@ $b->pubYear = 2023; // Fatal Error
     <example>
      <title>Asymmetric Property inheritance</title>
      <programlisting role="php">
-      <![CDATA[
+<![CDATA[
 <?php
 class Book
 {
--- a/language/oop5/visibility.xml
+++ b/language/oop5/visibility.xml
@@ -83,8 +83,8 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
     <para>
      As of PHP 8.4, properties may also have their
      visibility set asymmetrically, with a different scope for
-     reading (<emphasis>get</emphasis>) and writing (<emphasis>set</emphasis>).
-     Specifically, the <emphasis>set</emphasis> visibility may be
+     reading (<literal>get</literal>) and writing (<literal>set</literal>).
+     Specifically, the <literal>set</literal> visibility may be
      specified separately, provided it is not more permissive than the
      default visibility.
     </para>
@@ -130,20 +130,20 @@ $b->pubYear = 2023; // Fatal Error
     <itemizedlist>
      <listitem>
       <simpara>
-       Only typed properties may have a separate <emphasis>set</emphasis> visibility.
+       Only typed properties may have a separate <literal>set</literal> visibility.
       </simpara>
      </listitem>
      <listitem>
       <simpara>
-       The <emphasis>set</emphasis> visibility must be the same
-       as <emphasis>get</emphasis> or more restrictive. That is,
+       The <literal>set</literal> visibility must be the same
+       as <literal>get</literal> or more restrictive. That is,
        <code>public protected(set)</code> and <code>protected protected(set)</code>
-       are allowed, but <code>protected public(set) will cause a syntax error.</code>
+       are allowed, but <code>protected public(set)</code> will cause a syntax error.
       </simpara>
      </listitem>
      <listitem>
       <simpara>
-       If a property is <literal>public</literal> writeable, then the main visibility may be
+       If a property is <literal>public</literal>, then the main visibility may be
        omitted.  That is, <code>public private(set)</code> and <code>private(set)</code>
        will have the same result.
       </simpara>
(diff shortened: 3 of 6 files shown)


language/oop5/final.xml
```diff
@@ -84,10 +103,12 @@ class Bar extends Foo
...
   <note>
    <simpara>
-    Properties cannot be declared final: only classes, methods, and constants (as of PHP 8.1.0) may be declared as final.
+    As of PHP 8.0.0, private methods may not be declared final except for the constructor.
    </simpara>
+  </note>
+  <note>
    <simpara>
-    As of PHP 8.0.0, private methods may not be declared final except for the constructor.
+    A property that is declared <literal>private(set)</literal> is implicitly <literal>final</literal>.
```
cmb69, line 111: As a long time PHP user, the reader might wonder what this `private(set)` is, and where to find out more about it. I suggest to link that to the new "section" in visibility. I think the first `para` could have an `xml:id` which can be used for `xlink`ng.

language/oop5/properties.xml
```diff
@@ -192,7 +192,12 @@ Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be ac
   <sect2 xml:id="language.oop5.properties.readonly-properties">
    <title>Readonly properties</title>
    <para>
-    As of PHP 8.1.0, a property can be declared with the <code>readonly</code> modifier, which prevents modification of the property after initialization.
+    As of PHP 8.1.0, a property can be declared with the <literal>readonly</literal> modifier,
+    which prevents modification of the property after initialization.  Prior to PHP 8.4.0
+    a <literal>readonly</literal> property is implicitly private-set, and may only be written to
+    from the same class.  As of PHP 8.4.0, <literal>readonly</literal> properties are implicitly
+    <literal>protected(set)</literal>, so may be set from child classes.  That may be overridden
```
cmb69, line 199: ditto (i.e. link)

language/oop5/visibility.xml
```diff
@@ -74,6 +74,117 @@ echo $obj2->private; // Undefined
...
+
+$b->title = 'How not to PHP'; // Fatal Error
+$b->author = 'Pedro H. Peterson'; // Fatal Error
+$b->pubYear = 2023; // Fatal Error
+?>
+]]>
+     </programlisting>
+    </example>
+   </para>
+   <para>There are a few caveats with asymmetric visibility to be aware of.</para>
```
cmb69, line 127: I think this should either be a `<simpara>` or the `<para>` should include the `<itemizedlist>`. (I wonder whether PhD supports `<itemizedlist><title>…</title>…</itemizedlist>`, which might semantically be more appropriate.)

cmb69 (commented):
Thank you! (can't really comment on the actual contents)

language/oop5/visibility.xml
Crell: I have no meaningful preference here.  I have a hard time telling which list style to use where, so whatever you think PhD will accept as valid, I can go with.

language/oop5/visibility.xml
```diff
@@ -74,6 +74,117 @@ echo $obj2->private; // Undefined
...
+    reading (<emphasis>get</emphasis>) and writing (<emphasis>set</emphasis>).
+    Specifically, the <emphasis>set</emphasis> visibility may be
+    specified separately, provided it is not more permissive than the
+    default visibility.
+   </para>
+   <para>
+    <example>
+     <title>Asymmetric Property visibility</title>
+     <programlisting role="php">
+      <![CDATA[
```
cmb69, line 93: Nit: it's customary to not indent `<![CDATA[`:
```suggestion
<![CDATA[
```

language/oop5/visibility.xml
```diff
@@ -74,6 +74,117 @@ echo $obj2->private; // Undefined
...
+    visibility, provided that the new visibility is the same or wider
+    than the parent class.  However, be aware that if a <literal>private</literal>
+    property is overridden, it does not actually change the parent's property
+    but creates a new property with a different internal name.
+   </para>
+   <para>
+    <example>
+     <title>Asymmetric Property inheritance</title>
+     <programlisting role="php">
+      <![CDATA[
```
cmb69, line 172:
```suggestion
<![CDATA[
```

language/oop5/visibility.xml
cmb69: Okay, on having a closer look, I'm not even sure any list makes sense, because currently there is only *one* item. If you intented to have an item for every `<simpara>`, that needs additional markup.

<details>
<summary>Multiple list items with title</summary>

````xml
   <itemizedlist>
    <title>Caveats with asymmetric visibility</title>
    <listitem>
     <simpara>
      Only typed properties may have a separate <emphasis>set</emphasis> visibility.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      The <emphasis>set</emphasis> visibility must be the same
      as <emphasis>get</emphasis> or more restrictive. That is,
      <code>public protected(set)</code> and <code>protected protected(set)</code>
      are allowed, but <code>protected public(set) will cause a syntax error.</code>
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      If a property is <literal>public</literal> writeable, then the main visibility may be
      omitted.  That is, <code>public private(set)</code> and <code>private(set)</code>
      will have the same result.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      A property with <literal>private(set)</literal> visibility
      is automatically <literal>final</literal>, and may not be redeclared in a child class.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      Obtaining a reference to a property follows <emphasis>set</emphasis> visibility, not <emphasis>get</emphasis>.
      That is because a reference may be used to modify the property value.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      Similarly, trying to write to an array property involves both a <emphasis>get</emphasis> and
      <emphasis>set</emphasis> operation internally, and therefore will follow the <emphasis>set</emphasis>
      visibility, as that is always the more restrictive.
     </simpara>
    </listitem>
   </itemizedlist>
````

caveats
</details>

This seems to be the best solution semantically, but doesn't look like it should.

<details>
<summary>Multiple items with `simpara` as introduction</summary>

````xml
   <simpara>There are a few caveats with asymmetric visibility to be aware of:</simpara>
   <itemizedlist>
    <listitem>
     <simpara>
      Only typed properties may have a separate <emphasis>set</emphasis> visibility.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      The <emphasis>set</emphasis> visibility must be the same
      as <emphasis>get</emphasis> or more restrictive. That is,
      <code>public protected(set)</code> and <code>protected protected(set)</code>
      are allowed, but <code>protected public(set) will cause a syntax error.</code>
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      If a property is <literal>public</literal> writeable, then the main visibility may be
      omitted.  That is, <code>public private(set)</code> and <code>private(set)</code>
      will have the same result.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      A property with <literal>private(set)</literal> visibility
      is automatically <literal>final</literal>, and may not be redeclared in a child class.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      Obtaining a reference to a property follows <emphasis>set</emphasis> visibility, not <emphasis>get</emphasis>.
      That is because a reference may be used to modify the property value.
     </simpara>
    </listitem>
    <listitem>
     <simpara>
      Similarly, trying to write to an array property involves both a <emphasis>get</emphasis> and
      <emphasis>set</emphasis> operation internally, and therefore will follow the <emphasis>set</emphasis>
      visibility, as that is always the more restrictive.
     </simpara>
    </listitem>
   </itemizedlist>
````

caveats2
</details>

Semantic nonsense, but looks quite okay.

<details>
<summary>Note without list</summary>

````xml
   <note>
    <title>Caveats with asymmetric visibility</title>
    <simpara>
     Only typed properties may have a separate <emphasis>set</emphasis> visibility.
    </simpara>
    <simpara>
     The <emphasis>set</emphasis> visibility must be the same
     as <emphasis>get</emphasis> or more restrictive. That is,
     <code>public protected(set)</code> and <code>protected protected(set)</code>
     are allowed, but <code>protected public(set) will cause a syntax error.</code>
    </simpara>
    <simpara>
     If a property is <literal>public</literal> writeable, then the main visibility may be
     omitted.  That is, <code>public private(set)</code> and <code>private(set)</code>
     will have the same result.
    </simpara>
    <simpara>
     A property with <literal>private(set)</literal> visibility
     is automatically <literal>final</literal>, and may not be redeclared in a child class.
    </simpara>
    <simpara>
     Obtaining a reference to a property follows <emphasis>set</emphasis> visibility, not <emphasis>get</emphasis>.
     That is because a reference may be used to modify the property value.
    </simpara>
    <simpara>
     Similarly, trying to write to an array property involves both a <emphasis>get</emphasis> and
     <emphasis>set</emphasis> operation internally, and therefore will follow the <emphasis>set</emphasis>
     visibility, as that is always the more restrictive.
    </simpara>
   </note>
````

caveats2
</details>

Oh, the `<simpara>`s don't have any effect.

<details>
<summary>Note without list but paras</summary>

````xml
   <note>
    <title>Caveats with asymmetric visibility</title>
    <para>
     Only typed properties may have a separate <emphasis>set</emphasis> visibility.
    </para>
    <para>
     The <emphasis>set</emphasis> visibility must be the same
     as <emphasis>get</emphasis> or more restrictive. That is,
     <code>public protected(set)</code> and <code>protected protected(set)</code>
     are allowed, but <code>protected public(set) will cause a syntax error.</code>
    </para>
    <para>
     If a property is <literal>public</literal> writeable, then the main visibility may be
     omitted.  That is, <code>public private(set)</code> and <code>private(set)</code>
     will have the same result.
    </para>
    <para>
     A property with <literal>private(set)</literal> visibility
     is automatically <literal>final</literal>, and may not be redeclared in a child class.
    </para>
    <para>
     Obtaining a reference to a property follows <emphasis>set</emphasis> visibility, not <emphasis>get</emphasis>.
     That is because a reference may be used to modify the property value.
    </para>
    <para>
     Similarly, trying to write to an array property involves both a <emphasis>get</emphasis> and
     <emphasis>set</emphasis> operation internally, and therefore will follow the <emphasis>set</emphasis>
     visibility, as that is always the more restrictive.
    </para>
   </note>
````

caveats3
</details>

Not great, but visually okay. Although the generated HTML of notes is semantic garbage.

So maybe go with "Multiple items with `simpara` as introduction" for now. Might be the best compromise regarding semantics and visual presentation.

File: language/oop5/final.xml (hunk 1 above)
Comment (cmb69) [inline language/oop5/final.xml:111]: So, I suggest to wrap the new contents in visibility.xml in a `<sect3>` element:
````xml
   <sect3 xml:id="language.oop5.visibility-members-aviz">
    <title>Asymmetric Property Visibility</title>
````
and then change this line to something like:
```suggestion
    A property that is declared
    <link linkend="language.oop5.visibility-members-aviz"><literal>private(set)</literal></link>
    is implicitly <literal>final</literal>.
```

language/oop5/visibility.xml
Crell: Ugh.  My IDE always does that when I copy-paste code samples...

language/oop5/final.xml
Crell: Done

language/oop5/visibility.xml
Crell: Er, that would be a dumb typo on my part; it was supposed to be a `<listitem><simpara></simpara></listitem>` chain all the way down, which is what I saw used elsewhere.

language/oop5/properties.xml
```diff
@@ -192,7 +192,13 @@ Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be ac
   <sect2 xml:id="language.oop5.properties.readonly-properties">
    <title>Readonly properties</title>
    <para>
-    As of PHP 8.1.0, a property can be declared with the <code>readonly</code> modifier, which prevents modification of the property after initialization.
+    As of PHP 8.1.0, a property can be declared with the <literal>readonly</literal> modifier,
```
cmb69, line 195: Nit for thought: is `<literal>` preferable to `<code>` here? I'm not sure.

language/oop5/visibility.xml
```diff
@@ -78,6 +78,130 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
      </programlisting>
     </example>
    </para>
+   <sect3 xml:id="language.oop5.visibility-members-aviz">
+    <title>Asymmetric Property Visibility</title>
+    <para>
+     As of PHP 8.4, properties may also have their
+     visibility set asymmetrically, with a different scope for
+     reading (<emphasis>get</emphasis>) and writing (<emphasis>set</emphasis>).
```
cmb69, line 86: I don't like `<emphasis>` here (and below) for semantic reasons, although `<literal>` is also not quite appropriate.

language/oop5/visibility.xml
```diff
@@ -78,6 +78,130 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
...
+$b->title = 'How not to PHP'; // Fatal Error
+$b->author = 'Pedro H. Peterson'; // Fatal Error
+$b->pubYear = 2023; // Fatal Error
+?>
+]]>
+      </programlisting>
+     </example>
+    </para>
+    <itemizedlist>
+     <title>Caveats regarding asymmetric visibility</title>
```
cmb69, line 130: While this is the proper way[TM] to mark it up, the HTML rendering is off, since it is converted to `<h1>` (while the `<sect3>` title above is converted to `<h4>`). Either we fix PhD in this regard, or we should work around with having the list's title as a para/simpara above the list. I'm not happy with either.

cmb69 (commented):
Thank you! This looks quite good now, but a few things to consider.

@Girgias, since this is mostly about aviz (i.e. PHP 8.4), should we track this PR somewhere?

Also, what to do about the `<itemizedlist><title>`?

language/oop5/properties.xml
Crell: `<literal>` seems more common from what I saw, so I tried to standardize on that.  I have no particular preference, but this seemed to be the convention.

language/oop5/visibility.xml
Crell: Yeah, I wasn't a fan of that either but don't know what's preferable.  I'm happy to change it to whatever we decide is least wrong.

language/oop5/visibility.xml
Crell: I'll move it back to an intro paragraph to avoid blocking this PR on PhD updates.

language/oop5/visibility.xml
cmb69: It's probably okay to use `<literal>`; after all, it's a literal part of the aviz syntax.

Crell: Done.

cmb69 (approved):
Thank you!

language/oop5/final.xml
```diff
@@ -55,6 +55,25 @@ class ChildClass extends BaseClass {
 }
 // Results in Fatal error: Class ChildClass may not inherit from final class (BaseClass)
 ?> 
+]]>
+    </programlisting>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title>Final property example as of PHP 8.4.0</title>
```
Girgias, line 64: The example tag does not need to be nested in a `<para>` tag. I see that the other ones do it, but it is just useless.

language/oop5/final.xml
```diff
@@ -55,6 +55,25 @@ class ChildClass extends BaseClass {
...
+class BaseClass {
+   final protected string $test;
+}
+
+class ChildClass extends BaseClass {
+    public string $test;
+}
+// Results in Fatal error: Cannot override final property BaseClass::$test
+?>
 ]]>
```
Girgias, line 77: I would prefer you have a `<screen>` tag that shoes the output than a comment, as comments in examples are somewhat annoying for translators.

language/oop5/final.xml
```diff
@@ -84,10 +103,12 @@ class Bar extends Foo
 
   <note>
    <simpara>
-    Properties cannot be declared final: only classes, methods, and constants (as of PHP 8.1.0) may be declared as final.
+    As of PHP 8.0.0, private methods may not be declared final except for the constructor.
```
Girgias, line 106: Drive by improvement, add a link to the page explaining constructors?

language/oop5/properties.xml
```diff
@@ -192,7 +192,13 @@ Fatal error: Uncaught Error: Typed property Shape::$numberOfSides must not be ac
...
+    which prevents modification of the property after initialization.  Prior to PHP 8.4.0
+    a <literal>readonly</literal> property is implicitly private-set, and may only be written to
+    from the same class.  As of PHP 8.4.0, <literal>readonly</literal> properties are implicitly
+    <link linkend="language.oop5.visibility-members-aviz"><literal>protected(set)</literal></link>,
+    so may be set from child classes.  That may be overridden
+    explicitly if desired.
```
Girgias, line 201: I prefer to add line breaks at the end of sentences.
Because if those need to be amended in the future it is easier for translators to see what needs fixing.
```suggestion
    which prevents modification of the property after initialization.*
    Prior to PHP 8.4.0 a <literal>readonly</literal> property is implicitly private-set,
    and may only be written to from the same class.
    As of PHP 8.4.0, <literal>readonly</literal> properties are implicitly
    <link linkend="language.oop5.visibility-members-aviz"><literal>protected(set)</literal></link>,
    so may be set from child classes. 
    That may be overridden explicitly if desired.
```

language/oop5/visibility.xml
```diff
@@ -78,6 +78,130 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
...
+    <para>
+     As of PHP 8.4, properties may also have their
+     visibility set asymmetrically, with a different scope for
+     reading (<literal>get</literal>) and writing (<literal>set</literal>).
+     Specifically, the <literal>set</literal> visibility may be
+     specified separately, provided it is not more permissive than the
+     default visibility.
+    </para>
+    <para>
+     <example>
```
Girgias, line 92: No need to have a para tag around the example tag.

language/oop5/visibility.xml
```diff
@@ -78,6 +78,130 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
...
+      </simpara>
+     </listitem>
+     <listitem>
+      <simpara>
+       Similarly, trying to write to an array property involves both a <literal>get</literal> and
+       <literal>set</literal> operation internally, and therefore will follow the <literal>set</literal>
+       visibility, as that is always the more restrictive.
+      </simpara>
+     </listitem>
+    </itemizedlist>
```
Girgias, line 170: Move the itemizedlist tag inside the para tag as they form "one block".

language/oop5/visibility.xml
```diff
@@ -78,6 +78,130 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
...
+     When a class extends another, the child class may redefine
+     any property that is not <literal>final</literal>.  When doing so,
+     it may widen either the main visibility or the <literal>set</literal>
+     visibility, provided that the new visibility is the same or wider
+     than the parent class.  However, be aware that if a <literal>private</literal>
+     property is overridden, it does not actually change the parent's property
+     but creates a new property with a different internal name.
+    </para>
+    <para>
+     <example>
```
Girgias, line 181: No need to have a para tag wrapping the example tag.

language/oop5/visibility.xml
```diff
@@ -78,6 +78,130 @@ $obj2->printHello(); // Shows Public2, Protected2, Undefined
...
+     </listitem>
+    </itemizedlist>
+    <para>
+     When a class extends another, the child class may redefine
+     any property that is not <literal>final</literal>.  When doing so,
+     it may widen either the main visibility or the <literal>set</literal>
+     visibility, provided that the new visibility is the same or wider
+     than the parent class.  However, be aware that if a <literal>private</literal>
+     property is overridden, it does not actually change the parent's property
+     but creates a new property with a different internal name.
```
Girgias, line 178: Ditto remark about using line breaks at the end of a sentence.

reference/reflection/reflectionproperty/isprivate.xml
```diff
@@ -25,7 +25,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if the property is private, &false; otherwise.
+   &true; if the property is private, &false; otherwise.  Note this refers only to the
+   main visibility, and not to a <link linkend="language.oop5.visibility-members-aviz">set-visibility</link>, if specified.
```
Girgias, line 29:
```suggestion
   &true; if the property is private, &false; otherwise.
   Note this refers only to the  main visibility, and not to a
   <link linkend="language.oop5.visibility-members-aviz">set-visibility</link>, if specified.
```
Or maybe this should even be put inside a `<note>` tag.

reference/reflection/reflectionproperty/isprotected.xml
```diff
@@ -25,7 +25,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if the property is protected, &false; otherwise.
+   &true; if the property is protected, &false; otherwise.  Note this refers only to the
+   main visibility, and not to a <link linkend="language.oop5.visibility-members-aviz">set-visibility</link>, if specified.
```
Girgias, line 29:
```suggestion
   &true; if the property is protected, &false; otherwise.
   Note this refers only to the main visibility, and not to a
   <link linkend="language.oop5.visibility-members-aviz">set-visibility</link>, if specified.
```

reference/reflection/reflectionproperty/ispublic.xml
```diff
@@ -25,7 +25,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   &true; if the property is public, &false; otherwise.
+   &true; if the property is marked public, &false; otherwise.  Note this refers only to the
+   main visibility, and not to a <link linkend="language.oop5.visibility-members-aviz">set-visibility</link>, if specified.
```
Girgias, line 29:
```suggestion
   &true; if the property is marked public, &false; otherwise.
   Note this refers only to the main visibility, and not to a
   <link linkend="language.oop5.visibility-members-aviz">set-visibility</link>, if specified.
```

Girgias (changes requested):
Mainly markup remarks.

Girgias:
> Thank you! This looks quite good now, but a few things to consider.
> 
> @Girgias, since this is mostly about aviz (i.e. PHP 8.4), should we track this PR somewhere?

Please add it to the milestone, and I will create an issue the moment I finish the migration guide.

> 
> Also, what to do about the `<itemizedlist><title>`?

What is the issue here?

cmb69:
> > Also, what to do about the `<itemizedlist><title>`?
> 
> What is the issue here?

PhD issue. The [`<title>` is always rendered as `<h1>`](https://private-user-images.githubusercontent.com/2306138/373865232-900ce6cf-67fe-4d0c-b3b7-870383900c75.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mjg4MzE4MzksIm5iZiI6MTcyODgzMTUzOSwicGF0aCI6Ii8yMzA2MTM4LzM3Mzg2NTIzMi05MDBjZTZjZi02N2ZlLTRkMGMtYjNiNy04NzAzODM5MDBjNzUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTAxMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEwMTNUMTQ1ODU5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OWU4YzM4OGYxY2IyODEwZGVmM2UyMGUzZjQ4ZjU2NTE3ZWNmMGYxYjc3ZDViY2Q0NjgzZTJjMTRhMmQ0NDZkOCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.orWMAcvaum1NCVBSwR6G9YFmtBVlGAhH5jR3hxK6kHU), so is practically useless.

Girgias:
> > > Also, what to do about the `<itemizedlist><title>`?
> > 
> > 
> > What is the issue here?
> 
> PhD issue. The [`<title>` is always rendered as `<h1>`](https://private-user-images.githubusercontent.com/2306138/373865232-900ce6cf-67fe-4d0c-b3b7-870383900c75.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mjg4MzE4MzksIm5iZiI6MTcyODgzMTUzOSwicGF0aCI6Ii8yMzA2MTM4LzM3Mzg2NTIzMi05MDBjZTZjZi02N2ZlLTRkMGMtYjNiNy04NzAzODM5MDBjNzUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTAxMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEwMTNUMTQ1ODU5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OWU4YzM4OGYxY2IyODEwZGVmM2UyMGUzZjQ4ZjU2NTE3ZWNmMGYxYjc3ZDViY2Q0NjgzZTJjMTRhMmQ0NDZkOCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.orWMAcvaum1NCVBSwR6G9YFmtBVlGAhH5jR3hxK6kHU), so is practically useless.

I suppose for the moment just leave it an open a PhD issue, as it should track nesting (because I imagine this being an issue outside of `<itemizedlist>`).

language/oop5/properties.xml
Crell: Is there any standard for this?  @cmb69?  I have not seen that approach used anywhere else, and it feels kinda clumsy to me.

cmb69: No, there are now hard rules about this. Still, it can make sense to take this approach (unless overdone); "good" diffs may be more important than readability of the doc sources; after all, you can't have very good readability at all (e.g. the long `<link>` element). A rule of thumb *might* be to use relatively short lines (say 72 columns, or maybe even less), and only go above that limit where required (e.g. to not split a long "inline" element), and prefer to break lines after a full stop.

Anyway, if you take @Girgias' suggestion, don't forget to remove the spurious `*` at the end of the first line.

language/oop5/final.xml
Girgias: Probably because people didn't look at the DocBook spec, or PhD used to render this weirdly.

language/oop5/properties.xml
Girgias: I fixed the suggestion.

But this is a rule that is used by other large orgs, e.g. [GitLab](https://docs.gitlab.com/ee/development/documentation/styleguide/#line-length) and is called [Semantic Line Breaks](https://sembr.org/)

cmb69: We should also consider https://docs.gitlab.com/ee/development/documentation/styleguide/#documentation-is-the-single-source-of-truth-ssot ;)

language/oop5/final.xml
Crell: OK, I took out the para on all examples in the file.

Crell:
I addressed everything above aside from the sentence-per-line bits.  That sounds like it's not a formal convention yet, and it's non-trivial to do, so I'd rather hold off until/unless there's some policy decision on it to follow.

Anything else to address before we merge this?

cmb69 (approved):
Yeah, let's not be too picky here. :)

Crell:
(Rebased to try and fix the build.)

language/oop5/final.xml
```diff
@@ -78,16 +94,17 @@ class Bar extends Foo
...
-    </programlisting>
-   </example>
-  </para>
+  </programlisting>
+ </example>
 
   <note>
    <simpara>
-    Properties cannot be declared final: only classes, methods, and constants (as of PHP 8.1.0) may be declared as final.
+    As of PHP 8.0.0, private methods may not be declared final except for the <link linkend="#language.oop5.decon.constructor">constructor</link>.
```
cmb69, line 102:
```suggestion
    As of PHP 8.0.0, private methods may not be declared final except for the <link linkend="language.oop5.decon.constructor">constructor</link>.
```

language/oop5/visibility.xml
```diff
@@ -74,10 +73,129 @@ echo $obj2->private; // Undefined
...
+      <simpara>
+       Similarly, trying to write to an array property involves both a <literal>get</literal> and
+       <literal>set</literal> operation internally, and therefore will follow the <literal>set</literal>
+       visibility, as that is always the more restrictive.
+      </simpara>
+     </listitem>
+    </itemizedlist>
+    <para>
+     When a class extends another, the child class may redefine
+     any property that is not <literal>final</literal>.  When doing so,
```
Girgias, line 169: Maybe there should be a link to `final` keyword here.

Feels like we should add more keywords entities to `doc-base/entities/global.ent`

Girgias (approved):

PR #3830, opened 2024-10-04
Socket is a reserved classname

afilina:
The fact that this classname can no longer be used for custom classes (making legacy code crash) is not explicitly stated anywhere.

### DIFF
--- a/appendices/migration80/incompatible.xml
+++ b/appendices/migration80/incompatible.xml
@@ -1305,6 +1305,11 @@ $array["key"];
      <function>socket_addrinfo_lookup</function> have been removed.
     </para>
    </listitem>
+   <listitem>
+    <para>
+     The <classname>Socket</classname> classname can no longer be used for custom classes.
+    </para>
+   </listitem>
   </itemizedlist>
  </sect2>
 


afilina:
I feel like the list above serves a different purpose: it matters to those who interact with these classes directly. When upgrading PHP versions, the code blows up in a completely different area, where I don't actually care about checking the returned value. The real problem is that the arrival of new classes in the global namespace starts interfering with existing classes in the application code, causing a fatal error.

Ideally, I would add a whole section or even page on classnames that became reserved, as that is most useful when upgrading legacy code.

cmb69:
Well, actually all new classes and interfaces should be listed on an own page (like for [migration81](https://www.php.net/manual/en/migration81.new-classes.php) and [migration 83](https://www.php.net/manual/en/migration83.new-classes.php)), regardless whether they stem from resource to object conversion, or are new features. Of course, such new classes and interfaces may cause backward incompatibilites, but it's the same with new functions and constants which are also listed on their dedicated pages (but not in the "Backward Incompatible Changes" section).

Would that be okay for you?

afilina:
Yes, that would be perfect. I'm not sure how to create a new page in the docs. Perhaps if you could create one, I would be able to fill it with a whole bunch of class names.

cmb69:
Please see and double-check #3832; maybe I've missed one or three.

PR #3831, opened 2024-10-04
MongoDB committransaction: Typo correction

leonardolara:
Correction of typo in value string.


Girgias:
Pinging @alcaeus as I don't know if the typo is bug in Mongo or just a mistake here.

alcaeus (approved):
Just a typo in the docs, thanks for fixing it!

PR #3832, opened 2024-10-04
Add missing new classes to migration guides

cmb69:
While these are already documented in the incompatible changes section regarding the resource to object conversions, they should also be listed explicitly as new classes.

---

Note that I used to verbose `<itemizedlist>` markup for consistency, although a `<simplelist>` seems more appropriate.

### DIFF
--- a/appendices/migration80.xml
+++ b/appendices/migration80.xml
@@ -22,6 +22,7 @@
  </para>
 
  &appendices.migration80.new-features;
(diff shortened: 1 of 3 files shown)


Girgias (approved):
Yeah, I feel those might also need to be added for 8.2/8.3

cmb69:
I didn't find any for 8.2 or 8.3 (checked the incompatible sections of the migration guide).

PR #3834, opened 2024-10-05
Move the parse_url() caution up to description section

cmb69:
This information is apparently not widely known, so we put it in a more prominent place.

---

In my opinion, the notes section should not contain any notes, cautions, and warnings at all. These should be put where they make most sense; the notes section should only contain "normal" text.


Girgias (approved):
Yeah... the notes section is kinda useless.

PR #3839, opened 2024-10-06
Update descriptions of disabled output handler

haszi:
Documents the changes made in https://github.com/php/php-src/pull/13087 and https://github.com/php/php-src/pull/15183. Also related to https://github.com/php/php-src/issues/16133.

### DIFF
--- a/reference/outcontrol/constants.xml
+++ b/reference/outcontrol/constants.xml
@@ -186,7 +186,7 @@
      Indicates that the output handler is disabled.
      This flag is set when the output handler returns &false;
      or fails while processing the buffer.
-     Prior to PHP 8.4.0, this flag can be set when starting an output buffer.
+     Prior to PHP 8.4.0, this flag could be set when starting an output buffer.
     </para>
    </listitem>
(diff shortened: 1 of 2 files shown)


reference/outcontrol/constants.xml
```diff
@@ -185,8 +185,8 @@
     <para>
      Indicates that the output handler is disabled.
      This flag is set when the output handler returns &false;
-     or fails while processing the buffer,
-     or it was set prior to calling the output handler.
+     or fails while processing the buffer.
+     Prior to PHP 8.4.0, this flag can be set when starting an output buffer.
```
cmb69, line 189: Present tense or past tense? I'd think the latter is more appropriate (although PHP 8.4.0 has not yet been released):
```suggestion
     Prior to PHP 8.4.0, this flag could be set when starting an output buffer.
```
Analogous below.

cmb69 (commented):
Thank you very much!

cmb69 (approved):
Thank you!

Girgias (approved):

PR #3845, opened 2024-10-07
Gmagick compositeimage.xml: set parameter name to lowercase

leonardolara:
According to the standards (and to the source code), the parameter name should be in lowercase.

### DIFF
--- a/reference/gmagick/gmagick/compositeimage.xml
+++ b/reference/gmagick/gmagick/compositeimage.xml
@@ -12,7 +12,7 @@
   <methodsynopsis>
    <modifier>public</modifier> <type>Gmagick</type><methodname>Gmagick::compositeimage</methodname>
    <methodparam><type>Gmagick</type><parameter>source</parameter></methodparam>
-   <methodparam><type>int</type><parameter>COMPOSE</parameter></methodparam>
+   <methodparam><type>int</type><parameter>compose</parameter></methodparam>
    <methodparam><type>int</type><parameter>x</parameter></methodparam>
    <methodparam><type>int</type><parameter>y</parameter></methodparam>
   </methodsynopsis>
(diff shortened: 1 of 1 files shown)


cmb69:
Hm, it's still uppercase in the stubs:

https://github.com/vitoc/gmagick/blob/2936e731e964719d84995a1e64722de5c5c0ee48/gmagick.stub.php#L43

and in the arginfo

https://github.com/vitoc/gmagick/blob/2936e731e964719d84995a1e64722de5c5c0ee48/gmagick_arginfo.h#L110

So I would argue that the docs are correct.

leonardolara:
> Hm, it's still uppercase in the stubs:
> 
> https://github.com/vitoc/gmagick/blob/2936e731e964719d84995a1e64722de5c5c0ee48/gmagick.stub.php#L43
> 
> and in the arginfo
> 
> https://github.com/vitoc/gmagick/blob/2936e731e964719d84995a1e64722de5c5c0ee48/gmagick_arginfo.h#L110
> 
> So I would argue that the docs are correct.

Sorry about that, you're right.
I think I checked in the wrong place:
https://github.com/vitoc/gmagick/blob/2936e731e964719d84995a1e64722de5c5c0ee48/gmagick_methods.c#L387

leonardolara:
So, the parameters section should also have "COMPOSE" in uppercase, right?
File updated.

cmb69:
> I think I checked in the wrong place:

Ah, the old protos! These could have been used to generate documentation, but I'm not sure if there is still support for that. Anyway, you always need to check the arginfo (and if this has been generated by gen_stub.php, the respective stub files).

Anyhow, thanks for the patch!

cmb69:
> Anyway, you always need to check the arginfo (and if this has been generated by gen_stub.php, the respective stub files).

Or you can actually try with named arguments. :)

PR #3849, opened 2024-10-07
Deprecated mysqli_kill()

kamil-tekiela:


Girgias (approved):

reference/mysqli/mysqli/kill.xml
```diff
@@ -52,6 +52,30 @@
...
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       Both <methodname>mysqli::kill</methodname> and
+       <function>mysqli_kill</function> are deprecated now. Use the
```
jimwins, line 70:
```suggestion
       <function>mysqli_kill</function> are now deprecated. Use the
```

Overthinking it, do we even want to use "now" in changelog entries? PR #3850 had the same wording issue.

cmb69: I think we use "now" all the time.

jimwins (approved):

PR #3852, opened 2024-10-08
Refactor Language variables section variable scope

kamil-tekiela:


language/variables.xml
```diff
@@ -254,51 +259,49 @@ $unset_array[] = 'value'; // Does not generate a warning.
...
    <simpara>
-    Here the <varname>$a</varname> variable will be available within
-    the included <filename>b.inc</filename> script.  However, within
-    user-defined functions a local function scope is introduced.  Any
-    variable used inside a function is by default limited to the local
-    function scope.  For example:
+    Any variable created inside a function is limited to the local function
+    scope. Every function, except
+    <link linkend="functions.arrow">arrow functions</link>,
+    has its own local scope.
```
kamil-tekiela, line 283: I am not sure about this. If someone can suggest a better version, I would appreciate.

language/predefined/variables.xml
```diff
@@ -7,7 +7,7 @@
  <partintro>
 
   <para>
-   PHP provides a large number of predefined variables to all scripts. The
+   PHP provides a number of predefined variables to all scripts. The
```
Girgias, line 10: This reads better to me, but maybe @jimwins or @Crell want to chime in
```suggestion
   PHP provides some predefined variables to all scripts. The
```

language/variables.xml
```diff
@@ -210,8 +215,8 @@ $unset_array[] = 'value'; // Does not generate a warning.
    <title>Predefined Variables</title>
    
    <para>
-    PHP provides a large number of predefined variables to any script
-    which it runs. Many of these variables, however, cannot be fully
+    PHP provides a number of predefined variables to any script
+    it runs. Many of these variables, however, cannot be fully
```
Girgias, line 219: Ditto the previous remark

language/variables.xml
```diff
@@ -254,51 +259,49 @@ $unset_array[] = 'value'; // Does not generate a warning.
...
    <simpara>
-    Here the <varname>$a</varname> variable will be available within
-    the included <filename>b.inc</filename> script.  However, within
-    user-defined functions a local function scope is introduced.  Any
-    variable used inside a function is by default limited to the local
-    function scope.  For example:
+    Any variable created inside a function is limited to the local function
+    scope. Every function, except
+    <link linkend="functions.arrow">arrow functions</link>,
+    has its own local scope.
```
Girgias, line 283: This might be better, but I'm not sure I totally understood the point you were trying to say with the arrow functions, as they cannot really define a variable inside of their body which "leaks": https://3v4l.org/1iZL62
```suggestion
    Any variable created inside a named function or an
    <link linkend="functions.anonymous">anonymous</link> function
    is limited to the scope of the function body.
    However, <link linkend="functions.arrow">arrow functions</link>,
    bind variables from the parent scope to make them available inside the body.
```

language/variables.xml
```diff
@@ -314,7 +317,12 @@ test();
    <sect2 xml:id="language.variables.scope.global">
     <title>The <literal>global</literal> keyword</title>
     <simpara>
-     First, an example use of <literal>global</literal>:
+     The <literal>global</literal> keyword is used to bring a variable
+     from a global scope into a local scope. The keyword can be used with
+     a list of variables or a single variable. A local variable will be created
+     referencing the global variable of the same name. If the global
+     variable does not exist, the variable will be created in global scope and
+     assigned &null;.
```
Girgias, line 325: Maybe instead of saying "import" we could say "bind" as this seems to be the more standard Comp. Sci. terminology. :)

language/predefined/variables.xml
jimwins: I'd rewrite this: "PHP provides predefined variables that represent [external variables], built-in environment variables, and other information about the execution environment such as the number and values of the arguments passed to the script in the CLI environment."

I don't think this intro needs to mention 'last error messages' because it's deprecated and 'last retrieved headers' is just not that interesting in most cases (and `$http_response_header` should probably be deprecated, anyways).

language/variables.xml
```diff
@@ -224,7 +229,7 @@ $unset_array[] = 'value'; // Does not generate a warning.
     PHP also provides an additional set of predefined arrays
     containing variables from the web server (if applicable), the
     environment, and user input. These arrays are rather special
-    in that they are automatically global - i.e., automatically
+    in that they are automatically global, i.e., automatically
```
jimwins, line 232: I'd simplify to "These arrays are automatically available in every scope."

language/variables.xml
```diff
@@ -254,51 +259,49 @@ $unset_array[] = 'value'; // Does not generate a warning.
 
    <simpara>
     The scope of a variable is the context within which it is defined.
-    For the most part all PHP variables only have a single scope.
-    This single scope spans <function>include</function>d and
-    <function>require</function>d files as well. For example:
+    PHP has a function scope and a global scope.
+    Any variable defined outside a function is limited to the global scope.
+    The global scope spans <function>include</function>d and
+    <function>require</function>d files as well.
```
jimwins, line 265: This is wrong, files included by `include` and `require` inherit the scope from where they are included but any functions and classes end up in the global scope. There's a paragraph in the `include` docs that puts it well, I think, maybe we copy that here or reference it.

Not wild about constructions like `<function>include</function>d`, too, that might be awkward for translations.

kamil-tekiela: Thanks. Please check the new changes

language/variables.xml
kamil-tekiela: I was thinking about this and I am not sure what even the purpose of this subsection is. It duplicates the information from the other pages. Maybe we should drop this section?

language/variables.xml
kamil-tekiela: These are the words I was looking for. Thanks!

language/variables.xml
```diff
@@ -254,51 +249,53 @@ $unset_array[] = 'value'; // Does not generate a warning.
...
-    Here the <varname>$a</varname> variable will be available within
-    the included <filename>b.inc</filename> script.  However, within
-    user-defined functions a local function scope is introduced.  Any
-    variable used inside a function is by default limited to the local
-    function scope.  For example:
+    Any variable created inside a named function or an
+    <link linkend="functions.anonymous">anonymous</link> function
+    is limited to the scope of the function body.
+    However, <link linkend="functions.arrow">arrow functions</link>,
+    bind variables from the parent scope to make them available inside the body.
```
Girgias, line 274: Nit: the trailing comma is not needed.
```suggestion
    However, <link linkend="functions.arrow">arrow functions</link>
    bind variables from the parent scope to make them available inside the body.
```

language/variables.xml
```diff
@@ -254,51 +249,53 @@ $unset_array[] = 'value'; // Does not generate a warning.
...
-    This script will generate an undefined variable <constant>E_WARNING</constant>
-    (or a <constant>E_NOTICE</constant> prior to PHP 8.0.0)
-    diagnostic. However, if the
-    <link linkend="ini.display-errors">display_errors</link> INI setting is set to hide
-    such diagnostics then nothing at all will be outputted.
+    The example above will generate an undefined variable <constant>E_WARNING</constant>
+    (or an <constant>E_NOTICE</constant> prior to PHP 8.0.0).
     This is because the echo statement
     refers to a local version of the <varname>$a</varname> variable,
     and it has not been assigned a value within this scope.
```
Girgias, line 301: Is the note below this line about the behaviour of C that relevant?

language/variables.xml
```diff
@@ -314,7 +311,12 @@ test();
...
-     First, an example use of <literal>global</literal>:
+     The <literal>global</literal> keyword is used to bind a variable
+     from a global scope into a local scope. The keyword can be used with
+     a list of variables or a single variable. A local variable will be created
+     referencing the global variable of the same name. If the global
+     variable does not exist, the variable will be created in global scope and
+     assigned &null;.
     </simpara>
     <para>
      <example>
```
Girgias, line 322: Maybe while at it, fix the useless para tag wrapping example? But maybe it is better to just leave it be.

language/variables.xml
kamil-tekiela: I don't and I was afraid of touching it.

language/variables.xml
kamil-tekiela: Let's do it maybe next PR

language/variables.xml
Girgias: Understandable, we can always think about it later.

Girgias (approved):

Girgias:
Thank you!

PR #3853, opened 2024-10-08
curl: Clean up examples for `curl_multi_*()` usage

TimWolla:
The various `curl_multi_*()` pages each contained a (slightly different) example and based from a look at the curl reference documentation they were all wrong: They called `curl_multi_select` even if `curl_multi_exec` failed, which according to my understanding of the documentation results in “undefined behavior”.

Replace all these examples by a single fully-fledged and opinionated example on the `curl_multi_exec()` page. This example showcases how to combine all the `curl_multi_*()` functions, including correct error handling.

The example leaves out the handling of `CURLM_CALL_MULTI_PERFORM`, because it will not be returned for curl 7.20+. The minimally supported curl version as of PHP 8.0 is curl 7.29. A lower curl version is only acceptable for PHP 7.x (and earlier).


cmb69:
I haven't checked your example, but see https://bugs.php.net/bug.php?id=77682, which had been reported by the lead developer of cURL.

TimWolla:
Thanks for the pointer. Perhaps `curl_multi_select()` *is* legal to call and only `curl_multi_exec()` is not (https://curl.se/libcurl/c/curl_multi_perform.html):

> When this function returns error, the state of all transfers are uncertain and they cannot be continued. [curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html) should not be called again on the same multi handle after an error has been returned, unless first removing all the handles and adding new ones.

The example on that manual page however checks the return value of `curl_multi_perform()` before calling `curl_multi_wait()`, though.

In any case the error checking right after the `curl_multi_exec()` call makes sense to me: If an error occurred, we should bail out and not do anything else.

Girgias (commented):
I think the overall logic makes sense, but it might be better to move this example to `reference/curl/examples.xml` and possibly link this page more often?

TimWolla:
> and possibly link this page more often?

Is there some precedent for function pages to link back to the extension examples?

Girgias:
> > and possibly link this page more often?
> 
> Is there some precedent for function pages to link back to the extension examples?

I don't think so, but I don't see why we can't start doing this :)

Girgias (approved):
Looks sensible to me, maybe @adoy knows? Or @Ayesh ?

Ayesh (approved):
About the `$status !== CURLM_OK` check right after the `curl_multi_exec`, I think it makes total sense. Libcurl `curl_multi_perform` says it returns `CURLM_OK` even if individual transfers return errors, and I see the new example handles this situation.

I don't have a strong opinion on fixing existing examples with the `CURLM_OK` checks vs removing them altogether and adding a proper example to `curl_multi_exec`.

A couple very minor nits:
 - For the `curl_multi_exec($mh, $unfinishedHandles)` calls, could we use `curl_multi_exec($mh, $still_running)` instead? Existing examples, function parameter name, and libcurl docs use `still_running`.
 - IMO, using `curl_multi_info_read` makes the example longer? We can maybe use the full examples (with `curl_multi_info_read`) in the `curl_multi_info_read` example, but use a minimal example in other pages?

TimWolla:
> I don't have a strong opinion on fixing existing examples with the `CURLM_OK` checks vs removing them altogether and adding a proper example to `curl_multi_exec`.

Using curl multi correctly and safely is non-trivial. I recently spent several hours in the code of php-src and libcurl to find out how to do it correctly, handling all possible edge cases (and even I failed at my initial attempt, as can be seen by https://github.com/php/doc-en/pull/3853/commits/cd45aad0a79f0be3eb1d7d4f585adbedbf6e6001). This PR is the result of that. Basically making any change will result in buggy or inefficient behavior one way or another. Copying the same example onto multiple doc pages does not help anyone and will just lead to the examples diverging in case of future adjustments.

> IMO, using `curl_multi_info_read` makes the example longer? We can maybe use the full examples (with `curl_multi_info_read`) in the `curl_multi_info_read` example, but use a minimal example in other pages?

See above. I don't believe that you can safely use curl multi without ever using `curl_multi_info_read()`: You are generally interested in whether your requests succeed or not and using `curl_multi_info_read()` is the reliable way to find that out. The example is long, because curl multi is an expert API.

> For the `curl_multi_exec($mh, $unfinishedHandles)` calls, could we use `curl_multi_exec($mh, $still_running)` instead? Existing examples, function parameter name, and libcurl docs use `still_running`.

I'd rather not, because I want to keep the symmetry with `$updatedHandles`.

Girgias:
Let merge then as is, and we can always improve later.

PR #3855, opened 2024-10-11
Document exception chaining

greg0ire:
This is already documented as [a user note][The note], but the note does not explain what happens if both exceptions already have a previous exception.

Here is what I believe is the related piece of code:

```c
/* Chain potential exception from wrapping finally block */
if (Z_OBJ_P(fast_call)) {
	if (ex) {
		if (zend_is_unwind_exit(ex) || zend_is_graceful_exit(ex)) {
			/* discard the previously thrown exception */
			OBJ_RELEASE(Z_OBJ_P(fast_call));
		} else {
			zend_exception_set_previous(ex, Z_OBJ_P(fast_call));
		}
	} else {
		ex = EG(exception) = Z_OBJ_P(fast_call);
	}
}
```

[The note]: https://www.php.net/manual/en/language.exceptions.php#129177

### DIFF
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -79,6 +79,12 @@
     is executed. Additionally, if the &finally; block also contains a &return; statement,
     the value from the &finally; block is returned.
    </para>
+   <para>
+    Another notable interaction is between an exception thrown from within a &try; block,
+    and an exception thrown from within a &finally; block. If both blocks throw an exception,
+    then the exception thrown from the &finally; block will be the one that is propagated,
+    and the exception thrown from the &try; block will used as its previous exception.
+   </para>
   </sect1>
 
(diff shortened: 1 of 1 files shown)


language/exceptions.xml
```diff
@@ -79,6 +79,12 @@
     is executed. Additionally, if the &finally; block also contains a &return; statement,
     the value from the &finally; block is returned.
    </para>
+   <para>
+    Another notable interaction is between an exception thrown from with a &try; block,
```
stof, line 83:
```suggestion
    Another notable interaction is between an exception thrown from within a &try; block,
```

greg0ire: Thanks!

jimwins (approved):
LGTM.

DanielEScherzer:
Sorry it took us so long to merge this; thank you for your contribution!

PR #3868, opened 2024-10-14
apache2.xml Add the `literal` tag for .php ext

mmalferov:


install/windows/apache2.xml
```diff
@@ -139,7 +139,7 @@ FcgidWrapper "c:/php/php-cgi.exe" .php
     </programlisting>
    </example>
 
-   Files with a .php extension will now be executed by the PHP FastCGI
+   Files with a <literal>.php</literal> extension will now be executed by the PHP FastCGI
```
cmb69, line 142: I think this is okay, or would `<filename>` be more appropriate?

mmalferov: IMO, this is not a filepath, it is a file extension.  [6](https://github.com/search?q=repo%3Aphp%2Fdoc-en+%22literal%3E.php%3C%22&type=code) vs [4](https://github.com/search?q=repo%3Aphp%2Fdoc-en+%22filename%3E.php%3C%22&type=code) :-)

But mostly the `literal` tag is used for file extensions; similarly, I looked at a dozen other file extensions.

Let's decide if it is necessary `.php` is wrapped in the `filename` tag, while the rest of the extensions are wrapped in the `literal` tag. And when the decision is made, I will add another change to this commit, in which I will lead the rest `.php` to a single view

cmb69: Indeed, we can still change that "everywhere", if we want to. For now this is good. Thank you!

PR #3869, opened 2024-10-14
odbc-num-rows.xml: Replace literal with emphasis

leonardolara:
Replace `<literal>` with the more suitable `<emphasis>`.


cmb69:
Thank you!

PR #3873, opened 2024-10-15
Add missing constants and function for Memcached

cracksalad:
This PR...
- adds 30 previously undocumented class constants (2 of which are new in v3.3.0)
- adds 1 previously undocumented method
- fixes some inconsistencies in constant documentation

References:
- [PECL extension source code](https://github.com/php-memcached-dev/php-memcached/blob/master/php_memcached.c)
- [libmemcached-awesome documentation](https://awesomized.github.io/libmemcached/libmemcached/memcached_behavior.html)


reference/memcached/constants.xml
```diff
@@ -9,26 +9,77 @@
...
+  <varlistentry xml:id="memcached.constants.opt-compression-type">
+   <term><constant>Memcached::OPT_COMPRESSION_TYPE</constant></term>
+   <listitem>
+    <para>
+     Specifies the compression algorithm to use, if compression is enabled.
+     The valid algorithms are <constant>Memcached::COMPRESSION_FASTLZ</constant>,
+     <constant>Memcached::COMPRESSION_ZLIB</constant> and
+     <constant>Memcached::COMPRESSION_ZSTD</constant>.
+    </para>
+    <para>Type: &integer;, default: <constant>Memcached::COMPRESSION_FASTLZ</constant>.</para>
```
cmb69, line 31: As I understand it, `&integer;` and friends are discouraged nowadays.

reference/memcached/constants.xml
```diff
@@ -9,26 +9,77 @@
...
+     transparently.
+    </para>
     <para>Type: &boolean;, default: &true;.</para>
    </listitem>
   </varlistentry>
 
+  <varlistentry xml:id="memcached.constants.opt-compression-type">
+   <term><constant>Memcached::OPT_COMPRESSION_TYPE</constant></term>
+   <listitem>
+    <para>
```
cmb69, line 25: Unless that is not properly supported by PhD, I would use `<simpara>` inside of `<listitem>`.

cmb69 (commented):
Thank you! The markup looks okay (see comments below, though), but I don't know about the actual contents. Maybe @m6w6 can have a look?

reference/memcached/constants.xml
cracksalad: Oh okay, what to use instead?

cmb69: You can resolve that manually: `<type>integer</type>`. However, I still don't understand why translators apparently translate `&integer;`. See https://github.com/php/doc-en/pull/3818#discussion_r1786135924 for some background.

Girgias: It should be `<type>int</type>` however. As `integer` is not a proper PHP type :)

reference/memcached/memcached/setencodingkey.xml
```diff
@@ -0,0 +1,75 @@
...
+  <para>
+   This method sets the AES encryption/decryption key for data written to and read from Memcached.
+  </para>
+  <para>Available as of Memcached 3.1.0.</para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
```
Girgias, line 24: The para tag wrapping the variablelist is useless here

Girgias: Yeah I don't really know why some extension docs do that. I'm not sure the inconsistency is that much of a problem compared to annoying translations with markup changes.

But I need to go back to write some doc tooling for our XML.

reference/memcached/constants.xml
Girgias: `&true;`, `&false;` and `&null;` should not be changed.

reference/memcached/memcached/setencodingkey.xml
cracksalad: Removed `<para>` tags wrapping a `<variablelist>` tag for all Memcached methods. Stats: all methods accepting arguments but one were affected.

m6w6:
Awesome work, @cracksalad, thank you so much!

reference/memcached/constants.xml
```diff
@@ -42,39 +93,69 @@
...
+
+  <varlistentry xml:id="memcached.constants.serializer-msgpack">
+   <term><constant>Memcached::SERIALIZER_MSGPACK</constant></term>
+   <listitem>
+    <simpara>The <link xlink:href="&url.pecl.package;msgpack">MessagePack</link> serializer.</simpara>
+    <para>
+     This serializer is supported only if memcached is configured with
+     <literal>--enable-memcached-msgpack</literal> option and the
+     <literal>msgpack</literal> extension is loaded.
+    </para>
```
cmb69, line 131: Just noticed that you sometimes use `<simpara>`, sometimes `<para>` and sometimes both in a  `<listitem>` element. Please use always `<simpara>` there (unless `<para>` would be required, but I don't think this is the case here).

cracksalad: @cmb69 When using 
```xml
<simpara>foo</simpara>
<simpara>bar</simpara>
```
there is no line break in between.

In this specific example it would look like this (which in this case would not be a problem, but other cases would look worse):
grafik

cmb69: Oh, indeed, for whatever reason a `simpara` inside a `listitem` is rendered as `span`:

https://github.com/php/phd/blob/8428f31cfd767cff36812108c79ff4b9b518edd4/phpdotnet/phd/Package/Generic/XHTML.php#L278-L284

Although [DocBook says](https://tdg.docbook.org/tdg/5.2/simpara):

> Processing expectations
Formatted as a displayed block.

Not sure what to do. Fixing PhD seems to be trivial, but might break some intentional rendering. @Girgias, @haszi, @jimwins, thoughts?

jimwins: I think `<simpara>` should be rendered as `<p>` and if that causes HTML layout problems when it is inside a `<note>`, `<listitem>`, or `<entry>`, we need to change the CSS styling so it looks how we want. Maybe we should make it `<p class="simpara">`.

(I would also be happy to see `<simpara>` just go away in favor of always using `<para>` because it seems rather pointless and probably some artifact of optimizing for old XML processing toolchains. But I haven't looked deeply at DocBook to see if there are places where you can only use `<simpara>` and not `<para>` that justify its existence.)

jimwins: Yeah, mapping `<para>` to `<p>` is problematic given HTML's handling of block elements within those, but maybe we can leave that as a problem for another day. Mapping `<simpara>` to `<p>` seems rather obviously correct to me, so I'd support changing that.

haszi: I agree with `<simpara>` being mapped to `<p>`.

I've checked `doc-en` and it seems that `<simpara>`s in `<listitem>`s are also used for simpler lists (see [here](https://www.php.net/manual/en/extensions.alphabetical.php) and [here](https://www.php.net/manual/en/function.odbc-gettypeinfo.php#refsect1-function.odbc-gettypeinfo-returnvalues)). If `<simpara>` is switched to a `<p>` element the vertical spacing will increase a lot. Which is not to say that we shouldn't make this change just that we need an alternative way of making these less vertical space-consuming lists.

Girgias: I can see 3 solutions to the problem here.

Mapping `<simpara>` to `<p>` except if it is in a `<itemizedlist><listitem>` or `<orderedlist><listitem>` sequence.
(this would cover the issue here where a `<simpara>` is used in a `<listitem>` in a `<varlistentry>`.

We convert those compact lists to `<simplelist>` tags.

We add  a `role="compact"` on `<itemizedlist>`/`<orderedlist>` to change the processing inline (this might be annoying considering how PhD is implemented).

haszi: I think `<simplelist>` wouldn't make sense semantically as it is supposed to be `"An undecorated list of single words or short phrases"`.

The implementation of the other two options would take about the same amount of effort but I'm in favor of the `role="compact"` on `<itemizedlist>`/`<orderedlist>` elements as this shows intent in the documentation and can directly be controlled from there instead of the special casing hack that would only be known to those few of us who were part of this conversation. :-)

Girgias: Then let's do the `role="compact"` option. :)

reference/memcached/memcached/setoption.xml
```diff
@@ -44,6 +44,8 @@
        <simplelist>
         <member><literal>Memcached::OPT_HASH</literal> requires <literal>Memcached::HASH_*</literal> values.</member>
         <member><literal>Memcached::OPT_DISTRIBUTION</literal> requires <literal>Memcached::DISTRIBUTION_*</literal> values.</member>
+        <member><literal>Memcached::OPT_SERIALIZER</literal> requires <literal>Memcached::SERIALIZER_*</literal> values.</member>
+        <member><literal>Memcached::OPT_COMPRESSION_TYPE</literal> requires <literal>Memcached::COMPRESSION_*</literal> values.</member>
```
Girgias, line 48: Should be constant tags instead of literal no?

reference/memcached/constants.xml
```diff
@@ -60,6 +60,11 @@
     <simpara>
      Specifies the compression level to use, if compression is enabled.
     </simpara>
+    <simplelist>
+     <member><literal>Memcached::COMPRESSION_FASTLZ</literal> does not support compression levels.</member>
+     <member><literal>Memcached::COMPRESSION_ZSTD</literal>: minimum level is <literal>-22</literal> and maximum level is <literal>22</literal>.</member>
+     <member><literal>Memcached::COMPRESSION_ZLIB</literal>: minimum level is <literal>0</literal> and maximum level is <literal>9</literal>.</member>
+    </simplelist>
     <para>Type: <type>int</type>, default: <literal>3</literal>.</para>
```
Girgias, line 68: Just realizing that normally we put the type info in the `<term>` tag, but this is a follow-up.

reference/memcached/constants.xml
cracksalad: How to proceed with this PR: Should I wait for the `role="compact"` change to be in place or should I change the `para`s to `simpara`s right now?

reference/memcached/constants.xml
cracksalad: What about the default values? 
(1) Should they be moved to the `<term>` tag as well or 
(2) would you describe them in a `simpara` kind of as is or 
(3) should we move them to the `setOption`-method doc, cause constants do not really have defaults, or 
(4) should we restructure the "Predefined Constants" site to have different sections (one for the `OPT_*` constants, one for option values, one for `RES_*` constants etc.) and maybe use a table for the `OPT_*` constants (as is for [DateTime](https://www.php.net/manual/en/datetime.constants.php)) where we could add a column for the defaults
(5) more differently...

cmb69: I don't quite understand how a constant would have a default value. However, I agree with @Girgias that we can do that in a follow-up PR; this one is already very comprehensive.

reference/memcached/constants.xml
cmb69: I suggest to stick with you have for now. Changing to `simpara` would break the rendering, and waiting on `role="compact"` to be implemented might leave this important PR hanging around.

reference/memcached/constants.xml
cracksalad: > I don't quite understand how a constant would have a default value.

It doesn't! But the constant is used as the "name" of an option which has a default. I think it is really important to document those defaults, but I am not sure where. Since you can think of the documentation of those constants as the documentation of the configuration options, it seems to be the right place.

Girgias: The standard way to markup constant lists is with `<variablelist>`, but you can have multiple ones and even add a `<title>` tag to describe the grouping, look at `streams/constants.xml` for a rewrite that I recently did.

Tables are not great as they might not fit the width of the screen (e.g. mobile) and at present nearly all constant lists don't use tables.

reference/memcached/memcached/construct.xml
```diff
@@ -23,38 +23,36 @@
...
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term><parameter>connection_str</parameter></term>
+    <listitem>
+     <para>
+      <!-- TODO Document constructor params -->
+     </para>
+    </listitem>
+   </varlistentry>
```
Girgias, line 54: Create a new issue for this, as this is going to be forgotten...

reference/memcached/memcached/setencodingkey.xml
```diff
@@ -0,0 +1,72 @@
...
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><methodname>Memcached::get</methodname></member>
+    <member><methodname>Memcached::add</methodname></member>
+    <member><methodname>Memcached::set</methodname></member>
+   </simplelist>
+  </para>
+ </refsect1>
```
Girgias, line 50: No need for a para tag wrapping the simplelist

reference/memcached/memcached/setencodingkey.xml
```diff
@@ -0,0 +1,72 @@
...
+      The AES key.
+     </para>
+    </listitem>
+   </varlistentry>
+  </variablelist>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
```
Girgias, line 36: To follow the new style guide
```suggestion
  <simpara>
```

reference/memcached/memcached/setencodingkey.xml
```diff
@@ -0,0 +1,72 @@
...
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>key</parameter></term>
+    <listitem>
+     <para>
+      The AES key.
+     </para>
```
Girgias, line 28: To follow the new style guide
```suggestion
     <simpara>
      The AES key.
     </simpara>
```

reference/memcached/memcached/setencodingkey.xml
```diff
@@ -0,0 +1,72 @@
...
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis role="Memcached">
+   <modifier>public</modifier> <type>bool</type><methodname>Memcached::setEncodingKey</methodname>
+   <methodparam><type>string</type><parameter>key</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   This method sets the AES encryption/decryption key for data written to and read from Memcached.
+  </para>
```
Girgias, line 17: To follow the new style guide
```suggestion
  <simpara>
   This method sets the AES encryption/decryption key for data written to and read from Memcached.
  </simpara>
```

reference/memcached/memcached/setencodingkey.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="memcached.setencodingkey" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
```
Girgias, line 3: There is no use of this as far as I can see?
```suggestion
<refentry xml:id="memcached.setencodingkey" xmlns="http://docbook.org/ns/docbook">
```

Girgias (commented):
Please revert the markup changes for functions where the documentation did not change (effectively revert ` Remove unnecessary para tag wrapping variablelist tag for Memcached methods`). Yes, the markup is not what I consider ideal, but no need to give unnecessary work to translations.

cracksalad:
> Please revert the markup changes for functions where the documentation did not change (effectively revert ` Remove unnecessary para tag wrapping variablelist tag for Memcached methods`). Yes, the markup is not what I consider ideal, but no need to give unnecessary work to translations.

But there were exactly **42 :tada:** changed files, now there are only 5 left 😢

cmb69:
@cracksalad, we *try* to distinguish between fixes/improvements to the documentation which affect all translations (beside EN, there are about 10 of these active), and changes which are likely only relevant for EN (such as whitespace changes, typo fixes etc.) The latter can be marked as `skip-revcheck`, so translators do not need to care about these.

The (now reverted) markup changes would have affected all translations; much work for a small gain. Still, we may consider to do these changes for EN in a separate PR; that is then still something which translators may skip at their own decision.

cracksalad:
@cmb69 I was just joking. 42 simply felt right :joy:

So this `skip-revcheck` is a PR level marking, such that we have to at least do those changes in a different PR. Did I get that right?

cmb69:
> 42 simply felt right

Ah, got it. :)

`skip-revcheck` is applied to individual commits, but it's simpler to have separate PRs (often we want to squash-merge commits).

reference/memcached/memcached/addserver.xml
```diff
@@ -42,7 +42,7 @@
      <listitem>
       <para>
        The hostname of the memcache server. If the hostname is invalid,
-       data-related operations will set
```
Girgias, line 45: It seems the only changes in this file are trailing whitespace changes now, which should also be reverted to make the PR/squashed merged commit more focused and not bother translations.

reference/memcached/memcached/setoption.xml
```diff
@@ -28,7 +28,7 @@
      <term><parameter>option</parameter></term>
      <listitem>
       <para>
-       One of the <literal>Memcached::OPT_*</literal> constant.
+       One of the <constant>Memcached::OPT_*</constant> constant.
```
Girgias, line 31: For automatic linking during render.
```suggestion
       One of the <constant>Memcached::OPT_<replaceable>*</replaceable></constant> constant.
```

reference/memcached/memcached/setoption.xml
```diff
@@ -43,8 +43,10 @@
        <para>
         The options listed below require values specified via constants.
         <simplelist>
-         <member><literal>Memcached::OPT_HASH</literal> requires <literal>Memcached::HASH_*</literal> values.</member>
-         <member><literal>Memcached::OPT_DISTRIBUTION</literal> requires <literal>Memcached::DISTRIBUTION_*</literal> values.</member>
+         <member><constant>Memcached::OPT_HASH</constant> requires <constant>Memcached::HASH_*</constant> values.</member>
+         <member><constant>Memcached::OPT_DISTRIBUTION</constant> requires <constant>Memcached::DISTRIBUTION_*</constant> values.</member>
+         <member><constant>Memcached::OPT_SERIALIZER</constant> requires <constant>Memcached::SERIALIZER_*</constant> values.</member>
+         <member><constant>Memcached::OPT_COMPRESSION_TYPE</constant> requires <constant>Memcached::COMPRESSION_*</constant> values.</member>
```
Girgias, line 49: Ditto please add `<replaceable>` tags around the `*` in constant tags.
Also maybe add some line breaks so that the lines are less than 80 chars (not mandatory but recommended)

Girgias (changes requested):
Some minor comments now, hopefully last round of reviews.

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
+
+  <varlistentry xml:id="memcached.constants.serializer-msgpack">
+   <term><constant>Memcached::SERIALIZER_MSGPACK</constant></term>
+   <listitem>
+    <simpara>
+     The <link xlink:href="&url.pecl.package;msgpack">MessagePack</link> serializer.
+    </simpara>
+    <para>
+     This serializer is supported only if memcached is configured with
+     <literal>--enable-memcached-msgpack</literal> option and the
```
Girgias, line 138:
```suggestion
     <option>--enable-memcached-msgpack</option> option and the
```

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
+     <literal>msgpack</literal> extension is loaded.
```
Girgias, line 139: This probably should have a link, but considering it is external to the php docs it needs a global entity. So I would recommend doing this in a follow-up PR as this requires changes to doc-base.

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
    <term><constant>Memcached::OPT_HASH</constant></term>
    <listitem>
-    <para>Specifies the hashing algorithm used for the item keys. The valid
+    <simpara>
+     Specifies the hashing algorithm used for the item keys. The valid
      values are supplied via <constant>Memcached::HASH_*</constant> constants.
      Each hash algorithm has its advantages and its disadvantages. Go with the
-     default if you don't know or don't care.</para>
-    <para>Type: &integer;, default: <constant>Memcached::HASH_DEFAULT</constant></para>
+     default if you don't know or don't care.
```
Girgias, line 165: Please remove personalization (i.e. `you`) the manual style indicates to not use it (yes we know there is a lot of existing usages).

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
+    </para>
    </listitem>
   </varlistentry>
 
   <varlistentry xml:id="memcached.constants.opt-prefix-key">
    <term><constant>Memcached::OPT_PREFIX_KEY</constant></term>
    <listitem>
-    <para>This can be used to create a "domain" for your item keys. The value
+    <simpara>
+     This can be used to create a "domain" for your item keys. The value
```
Girgias, line 148: Ditto remove personalization while at it please.

Edit: seeing there is a lot on this page already... well this is upsetting but might be better to done in a follow-up PR.

reference/memcached/constants.xml
```diff
@@ -166,10 +258,18 @@
    </listitem>
   </varlistentry>
 
+  <varlistentry xml:id="memcached.constants.distribution-virtual-bucket">
+   <term><constant>Memcached::DISTRIBUTION_VIRTUAL_BUCKET</constant></term>
+   <listitem>
+    <para>Virtual Bucket hashing key distribution algorithm (also called vbucket).</para>
```
Girgias, line 264: This is new, so might as well use the new manual style.
```suggestion
    <simpara>Virtual Bucket hashing key distribution algorithm (also called vbucket).</simpara>
```

reference/memcached/constants.xml
```diff
@@ -178,206 +278,366 @@
...
-     (set, add, replace, append, prepend, delete, increment, decrement, etc.). 
-     Storage commands will be sent without spending time waiting for a reply 
+    <simpara>
+     Enables or disables ignoring the result of storage commands
+     (set, add, replace, append, prepend, delete, increment, decrement, etc.).
+     Storage commands will be sent without spending time waiting for a reply
      (there would be no reply).
-     Retrieval commands such as <function>Memcached::get</function> are unaffected by this setting.</para>
-    <para>Type: &boolean;, default: &false;.</para>
+     Retrieval commands such as <function>Memcached::get</function> are unaffected by this setting.
```
Girgias, line 353: Aside: as this is a pre-existing issue, but those should be in a `<methodname>` tags.
```suggestion
     Retrieval commands such as <methodname>Memcached::get</methodname> are unaffected by this setting.
```

reference/memcached/constants.xml
```diff
@@ -178,206 +278,366 @@
...
   </varlistentry>
 
   <varlistentry xml:id="memcached.constants.opt-server-failure-limit">
    <term><constant>Memcached::OPT_SERVER_FAILURE_LIMIT</constant></term>
    <listitem>
-    <para>Specifies the failure limit for server connection attempts. The
+    <simpara>
+     Specifies the failure limit for server connection attempts. The
+     server will be removed after this many continuous connection
+     failures.
```
Girgias, line 471: Considering this is being amended regardless, add some semantic breaks?
```suggestion
     Specifies the failure limit for server connection attempts.
     The server will be removed after this many continuous connection failures.
```

reference/memcached/constants.xml
```diff
@@ -178,206 +278,366 @@
...
    </listitem>
   </varlistentry>
 
   <varlistentry xml:id="memcached.constants.get-extended">
    <term><constant>Memcached::GET_EXTENDED</constant></term>
    <listitem>
-    <para>A flag for <function>Memcached::get</function>, <function>Memcached::getMulti</function> and
+    <para>
+     A flag for <function>Memcached::get</function>, <function>Memcached::getMulti</function> and
      <function>Memcached::getMultiByKey</function> to ensure that the CAS token values
```
Girgias, line 625: Ditto same remark about preexiting issues in regard to the use of `<function>` tags.

reference/memcached/constants.xml
```diff
@@ -178,206 +278,366 @@
...
   <varlistentry xml:id="memcached.constants.get-preserve-order">
    <term><constant>Memcached::GET_PRESERVE_ORDER</constant></term>
    <listitem>
-    <para>A flag for <function>Memcached::getMulti</function> and
+    <para>
+     A flag for <function>Memcached::getMulti</function> and
      <function>Memcached::getMultiByKey</function> to ensure that the keys are
      returned in the same order as they were requested in. Non-existing keys
-     get a default value of NULL.</para>
+     get a default value of NULL.
```
Girgias, line 639: This feels like it is talking about the PHP `null` which should use an entity.
```suggestion
     get a default value of &null;.
```

reference/memcached/constants.xml
```diff
@@ -518,6 +782,76 @@
    </listitem>
   </varlistentry>
 
+  <varlistentry xml:id="memcached.constants.res-stored">
+   <term><constant>Memcached::RES_STORED</constant></term>
+   <listitem>
+    <para>Item stored.</para>
```
Girgias, line 788: As this (and below) constants are new, could you please use:
```suggestion
    <simpara>Item stored.</simpara>
```

Girgias (changes requested):
Okay, I clearly didn't properly go through the constant list.
Hopefully this is really the last one.

Girgias:
Also please rebase, as it now conflicts because of #4028

reference/memcached/constants.xml
cracksalad: I am currently working on a msgpack doc as well, PR coming soon. I will add a proper link in there.

reference/memcached/constants.xml
cracksalad: There are 10 occurences of `you` in this file. 
If I understand your reasoning correctly, I ignore your personalization related comments for now and will come back to them in a future PR.

reference/memcached/constants.xml
cracksalad: Ignoring this for now as well...

reference/memcached/constants.xml
cracksalad: Sorry, I somehow missed those.

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
+   <term><constant>Memcached::SERIALIZER_MSGPACK</constant></term>
+   <listitem>
+    <simpara>
+     The <link xlink:href="&url.pecl.package;msgpack">MessagePack</link> serializer.
+    </simpara>
+    <para>
+     This serializer is supported only if memcached is configured with
+     <option>--enable-memcached-msgpack</option> option and the
+     <literal>msgpack</literal> extension is loaded.
+    </para>
```
Girgias, line 140:
```suggestion
    <simpara>
     This serializer is supported only if memcached is configured with
     <option>--enable-memcached-msgpack</option> option and the
     <literal>msgpack</literal> extension is loaded.
    </simpara>
```

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
    <listitem>
-    <para>The JSON serializer.</para>
+    <para>The JSON serializer. This serializer deserializes JSON to an object.</para>
+   </listitem>
+  </varlistentry>
+
+  <varlistentry xml:id="memcached.constants.serializer-json-array">
+   <term><constant>Memcached::SERIALIZER_JSON_ARRAY</constant></term>
+   <listitem>
+    <para>The JSON array serializer. This serializer deserializes JSON to an associative array.</para>
```
Girgias, line 126:
```suggestion
    <simpara>
     The JSON array serializer.
     This serializer deserializes JSON to an associative array
    </simpara>
```

reference/memcached/constants.xml
```diff
@@ -42,39 +100,71 @@
...
+     The <link linkend="book.igbinary">igbinary</link> serializer.
      Instead of textual representation it stores PHP data structures in a
-     compact binary form, resulting in space and time gains.</para>
+     compact binary form, resulting in space and time gains.
+    </simpara>
+    <para>
+     This serializer is supported only if memcached is configured with
+     <option>--enable-memcached-igbinary</option> option and the
+     <link linkend="intro.igbinary">igbinary</link> extension is loaded.
+    </para>
```
Girgias, line 112:
```suggestion
    <simpara>
     This serializer is supported only if memcached is configured with
     <option>--enable-memcached-igbinary</option> option and the
     <link linkend="intro.igbinary">igbinary</link> extension is loaded.
    </simpara>
```

reference/memcached/constants.xml
```diff
@@ -9,26 +9,84 @@
...
+   <listitem>
+    <simpara>The ZLIB compression algorithm.</simpara>
+   </listitem>
+  </varlistentry>
+
+  <varlistentry xml:id="memcached.constants.compression-zstd">
+   <term><constant>Memcached::COMPRESSION_ZSTD</constant></term>
+   <listitem>
+    <para>The ZSTD compression algorithm.</para>
+    <para>Available as of Memcached 3.3.0.</para>
```
Girgias, line 55:
```suggestion
    <simpara>
     The ZSTD compression algorithm.
    </simpara>
    <simpara>
     Available as of Memcached 3.3.0.
    </simpara>
```

reference/memcached/constants.xml
```diff
@@ -9,26 +9,84 @@
...
+     <constant>Memcached::COMPRESSION_ZSTD</constant>.
+    </simpara>
+    <para>Type: <type>int</type>, default: <constant>Memcached::COMPRESSION_FASTLZ</constant>.</para>
+   </listitem>
+  </varlistentry>
+
+  <varlistentry xml:id="memcached.constants.compression-fastlz">
+   <term><constant>Memcached::COMPRESSION_FASTLZ</constant></term>
+   <listitem>
+    <para>The FASTLZ compression algorithm.</para>
```
Girgias, line 40:
```suggestion
    <simpara>
     The FASTLZ compression algorithm.
    </simpara>
```

Girgias (approved):
Thank you!

PR #3877, opened 2024-10-16
Document SoapServer::__getLastResponse

nielsdos:


reference/soap/soapserver/getlastresponse.xml
```diff
@@ -0,0 +1,74 @@
...
+   <modifier>public</modifier> <type class="union"><type>string</type><type>null</type></type><methodname>SoapServer::__getLastResponse</methodname>
+   <void/>
+  </methodsynopsis>
+  <simpara>
+   Returns the XML sent in the last SOAP response.
+  </simpara>
+  <note>
+   <simpara>
+    This method works only if the <classname>SoapServer</classname> object
+    was created with the <literal>trace</literal> option set to &true;.
```
jimwins, line 21: Is it worth documenting what is returned when trace was not set to true? (Or does it throw an exception?)

MauricioFauth (approved):

reference/soap/soapserver/getlastresponse.xml
jimwins: Fair enough. No objections from me.

jimwins (approved):

PR #3882, opened 2024-10-17
Fix dom_import_simplexml() return type

cmb69:
See <https://github.com/php/php-src/pull/16489>.

### DIFF
--- a/reference/dom/functions/dom-import-simplexml.xml
+++ b/reference/dom/functions/dom-import-simplexml.xml
@@ -15,10 +15,9 @@
    <methodparam><type>object</type><parameter>node</parameter></methodparam>
   </methodsynopsis>
   <para>
-   This function takes the node <parameter>node</parameter> of class
-   <link linkend="ref.simplexml">SimpleXML</link> and makes it into a
-   <classname>DOMElement</classname> node. This new object can then be used
-   as a native <classname>DOMElement</classname> node. 
+   This function takes the given attribute or element <parameter>node</parameter> (a
+   <classname>SimpleXMLElement</classname> instance) and converts it to a
+   <classname>DOMAttr</classname> or <classname>DOMElement</classname> node, repectively.
   </para>
  </refsect1>
  <refsect1 role="parameters">
@@ -29,7 +28,7 @@
      <term><parameter>node</parameter></term>
      <listitem>
       <para>
-       The <classname>SimpleXMLElement</classname> node.
+       The attribute or element node to convert (a <classname>SimpleXMLElement</classname> instance).
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/dom/functions/dom-import-simplexml.xml
```diff
@@ -11,7 +11,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>DOMElement</type><methodname>dom_import_simplexml</methodname>
+   <type class="union"><type>DomAttr</type><type>DOMElement</type></type><methodname>dom_import_simplexml</methodname>
```
nielsdos, line 14:
```suggestion
   <type class="union"><type>DOMAttr</type><type>DOMElement</type></type><methodname>dom_import_simplexml</methodname>
```

nielsdos (changes requested):

nielsdos:
The documentation of this function is also a bit whack...

Description:
> This function takes the node node of class [SimpleXML](https://www.php.net/manual/en/ref.simplexml.php) and makes it into a [DOMElement](https://www.php.net/manual/en/class.domelement.php) node. This new object can then be used as a native [DOMElement](https://www.php.net/manual/en/class.domelement.php) node.

This is inaccurate now


Return values:
> The [DOMElement](https://www.php.net/manual/en/class.domelement.php) node added.

What ???

reference/dom/functions/dom-import-simplexml.xml
cmb69: Oh, that was wishful thinking, Thank you for checking!

cmb69:
> What ??? What does added even mean here?

Indeed, makes no sense, so updated. (a bit sad, though, since it would have had its 20th birthday in a couple of weeks ;)

Before updating/improving/fixing the description, could you please have a look at the implementation of dom_import_simplexml_common():

https://github.com/php/php-src/blob/42e179ef9dc7f5b8cf6f1135c6be8e7ffe4aa5cb/ext/dom/php_dom.c#L566-L570

Is it really okay to cast an arbitrary object to a `php_libxml_node_object`?

nielsdos:
Note that `php_libxml_import_node` will fail if the object is not actually an xml object.
I suppose it would be cleaner to check failure of `php_libxml_import_node` first and only then do the cast, but I don't think the current ordering hurts.

cmb69:
Ah, I've overlooked `php_libxml_import_node`!

I've pushed a further commit regarding the function and the parameter descriptions. However, it seems to me that not only `SimpleXMLElement`s can be imported, but rather arbitrary objects which have been registered with libxml. Might be worth to document that at some time.

nielsdos:
Indeed it's broader than simplexml. However, in practice only DOM & simplexml register handlers to be able to do it.
Also a small nit on wording: you used "convert", but it's not actually converting in the sense that it's a clone; they're working on the same xml nodes so it's more like creating a "view". i.e. changes you make via dom are gonna be visible in simplexml and vice versa.

cmb69:
Thanks for the clarification! I tried to improve the wording, and added a respective example.

nielsdos (approved):

nielsdos:
Thanks a lot!

PR #3883, opened 2024-10-17
Document `FFI\CType::getFuncParameterType()`

DanielEScherzer:


jimwins (changes requested):
The `index` parameter should get described, too. Maybe: "Index of the parameter (zero-based)."

DanielEScherzer:
> The `index` parameter should get described, too. Maybe: "Index of the parameter (zero-based)."

Done

reference/ffi/ctype/getfuncparametertype.xml
```diff
@@ -26,7 +24,7 @@
     <term><parameter>index</parameter></term>
     <listitem>
      <para>
-      
+      Index of function parameter, zero-indexed.
```
jimwins, line 27:
```suggestion
      Index of the function parameter, zero-based.
```

We use zero-based (and "zero based") in quite a few places in the docs, but not "zero indexed".

jimwins (changes requested):

reference/ffi/ctype/getfuncparametertype.xml
DanielEScherzer: I've always heard it as "indexed" rather than "based", but if there is an existing standard in this repo, then no worries, changed

reference/ffi/ctype/getfuncparametertype.xml
```diff
@@ -15,8 +15,6 @@
```
cmb69, line 15: There should be at least a short description; maybe just duplicate the first sentence of the return type section:
```suggestion
Returns the type of a parameter for the underlying function type.
```

Girgias (approved):

PR #3884, opened 2024-10-18
enablesslchecks.xml: updated sslChecks link in changelog

leonardolara:
To be consistent with recent update in method description.

### DIFF
--- a/reference/oauth/oauth/disablesslchecks.xml
+++ b/reference/oauth/oauth/disablesslchecks.xml
@@ -16,7 +16,7 @@
   <para>
    Turns off the usual SSL peer certificate and host checks,
    this is not for production environments. Alternatively, the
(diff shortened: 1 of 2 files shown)


Girgias:
I don't know how we could reasonably implement automatic linking for `<parameter>` as it would require somehow to specify the function/method of it?

However, maybe what we could do is `<parameter><xref /></parameter>` or `<paremeter><olink></olink></paramter>` for the linking?

cmb69:
That is not a parameter, but rather a property of OAuth.

Girgias:
> That is not a parameter, but rather a property of OAuth.

aaaaahhhh well that makes it simpler, but I also just edited my comment as I forgot common linking attributes exist

cmb69:
Right. See https://github.com/php/phd/issues/164#issuecomment-2422560394 (includes a POC).

PR #3887, opened 2024-10-20
Improve description of the iv parameter

MedAhamada:


cmb69:
Indeed, that apparently slipped through when PR #3774 had been merged. @bukka, can you confirm that this change is correct?

cmb69:
Thank you!

bukka:
Ah yeah sorry, it was wrong copy and past on my side.

PR #3891, opened 2024-10-22
Do not call attributes a “class” in the page title

TimWolla:


cmb69:
Maybe "attribute class"?

TimWolla:
The constructor already calls it just attribute: https://www.php.net/manual/en/sensitiveparameter.construct.php

> SensitiveParameter::__construct — Construct a new SensitiveParameter attribute instance

Girgias (approved):

PR #3892, opened 2024-10-22
Clarify `strcspn()` example

wouterj:
Using human readable words for the `$characters` parameter creates confusion, as each character is used separately in the function.

I propose to simplify the example by using one character when focusing on the standard behavior or the offset/length parameters. And modifying the example to use multiple consecutive characters instead of a real word (with duplicate characters).

### DIFF
--- a/reference/strings/functions/strcspn.xml
+++ b/reference/strings/functions/strcspn.xml
@@ -154,29 +154,26 @@
    <programlisting role="php">
 <![CDATA[
 <?php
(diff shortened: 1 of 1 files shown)


jimwins (approved):
LGTM.

TimWolla (commented):
Should we also have an example where the `$string` does not contain any of the `$characters`?

wouterj:
Example added, as requested.

TimWolla (approved):
Thanks!

PR #3898, opened 2024-10-22
Add migration example of property hooks.

Crell:
Just the stock standard example, really.

### DIFF
--- a/appendices/migration84/new-features.xml
+++ b/appendices/migration84/new-features.xml
@@ -11,8 +11,10 @@
    <title>Property Hooks</title>
 
    <simpara>
-    Object properties may now have additional logic associated with their <literal>get</literal> and <literal>set</literal>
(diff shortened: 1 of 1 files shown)


appendices/migration84/new-features.xml
```diff
@@ -11,17 +11,45 @@
    <title>Property Hooks</title>
 
    <simpara>
-    TODO
+    Object properties may now have additional logic associated with their <literal>get</literal> and <literal>set</literal>
+    operations.  Depending on the usage, that may or may not make the property virtual, that is, it has no backing value at all.
```
cmb69, line 15: These lines a rather long; I suggest to add some line breaks. Possibly:
```suggestion
    Object properties may now have additional logic associated with their
    <literal>get</literal> and <literal>set</literal> operations.
    Depending on the usage, that may or may not make the property virtual,
    that is, it has no backing value at all.
```

Girgias (approved):
Thank you!

PR #3899, opened 2024-10-22
Document XML_OPTION_PARSE_HUGE

nielsdos:


reference/xml/functions/xml-parser-set-option.xml
```diff
@@ -59,6 +59,15 @@
             XML parser.  Enabled by default.
            </entry>
           </row>
+          <row>
+           <entry><constant>XML_OPTION_PARSE_HUGE</constant></entry>
+           <entry>bool</entry>
+           <entry>
+            Allows parsing documents larger than 10 MB.
+            This option should only be enabled when the document size is
+            bounded.
```
cmb69, line 68: Probably clear enough, but maybe hint at DoS?

reference/xml/constants.xml
```diff
@@ -257,6 +257,17 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.xml-option-parse-huge">
+   <term>
+    <constant>XML_OPTION_PARSE_HUGE</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Available as of PHP 8.4.0.
```
cmb69, line 267: I have no idea whether libexpat is still properly supported, but if it is, we should clarify here (or maybe below) that the constant is only relevant for libxml2. I'm also not sure whether PHP 7.0 still supported libxml2 < 2.7.0, and if so, what would happen there (would basically `XML_OPTION_PARSE_HUGE` be enabled by default, and could not be disabled?)

cmb69 (commented):
Thank you! Maybe add some clarifications, but generally looks good.

reference/xml/constants.xml
nielsdos: you are right; why does it have to be so complex :(

nielsdos: Just checked, PHP 7 supports libxml2 2.6.11

cmb69 (approved):
Thank you!

PR #3901, opened 2024-10-22
Document DOMImplementation::createDocument() change for PHP  8.4

nielsdos:
Technically this can throw INVALID_STATE_ERR on allocation failure, just like other DOM node classes, but people shouldn't be handling that anyway as that happens when the system allocator is OOM...


Girgias (approved):
Thank you!

PR #3904, opened 2024-10-23
Use <keycap> and <keycombo> DocBook tags

Girgias:


jimwins (approved):
A few errors (mostly '0'/'O' mix-ups) but LGTM otherwise.

PR #3910, opened 2024-10-23
Some fixes to the 8.4 migration guide

Girgias:


Girgias:
@jimwins do you know why it is not picking up the DocBook 5.2 DTD?

Girgias:
I remember ... there is no DTD for 5.2, we need to set-up some sort of RelaxNG check rather than DTD...

appendices/migration84/incompatible.xml
```diff
@@ -459,9 +459,9 @@
 
   <simpara>
    Some DOM methods previously returned &false; or a
-   <constant>PHP_ERR</constant> <exceptionname>DOMException</exceptionname>
+   <constant>DOM_PHP_ERR</constant> <exceptionname>DOMException</exceptionname>
```
TimWolla, line 462:
```suggestion
   <exceptionname>DOMException</exceptionname> with code <constant>DOM_PHP_ERR</constant>
```

And likewise below? This avoids two identifiers after each other which renders badly.

TimWolla (commented):
While at it, please add the missing word "function" here:

https://github.com/php/doc-en/blob/4309102ef25eba1a9aa5a87d24e679286b7927e9/appendices/migration84/deprecated.xml#L94

---------------

LGTM other than these two remarks.

appendices/migration84/other-changes.xml
```diff
@@ -716,7 +727,7 @@
     Improved the performance of <classname>\Random\Randomizer</classname>,
     with a specific focus on the
     <methodname>\Random\Randomizer::getBytes</methodname>,
-    and <methodname>getBytesFromString()</methodname> methods.
+    and <methodname>\Random\Randomizer::getBytesFromString</methodname> methods.
```
TimWolla, line 730: These should not have the leading backslash. Also affects:

https://github.com/php/doc-en/blob/4309102ef25eba1a9aa5a87d24e679286b7927e9/appendices/migration84/deprecated.xml#L340

appendices/migration84/other-changes.xml
```diff
@@ -716,7 +727,7 @@
     Improved the performance of <classname>\Random\Randomizer</classname>,
```
TimWolla, line 727: Same here.

TimWolla (approved):

PR #3913, opened 2024-10-24
deprecated.xml Explain rising to the power of a negative number

mmalferov:


cmb69 (approved):
Thank you!

Girgias (approved):
Improvement for latter, is that we could use MathML for some nice rendering. However I don't know if PhD currently supports this.

appendices/migration84/deprecated.xml
```diff
@@ -84,8 +84,10 @@ function foo(T1 $a, ?T2 $b, T3 $c) {}
    <simpara>
     Raising a number to the power of a negative number is equivalent to taking
     the reciprocal of the number raised to the positive opposite of the power.
+    That is, <literal>10 ^ -2</literal> is the same as <literal>1 / (10 ^ 2)</literal>.
```
TimWolla, line 87: `^` is misleading, because in PHP this is bitwise-XOR. This should probably use `**` or the `<superscript>` tag as already used for Xoshiro256**'s jump method:

https://github.com/php/doc-en/blob/9a976335826507a6acae96c3df00713e8e91c81c/reference/random/random/engine/xoshiro256starstar/jump.xml#L17

mmalferov: I replace the `^` with the `<superscript>` tag

appendices/migration84/deprecated.xml
```diff
@@ -84,9 +84,13 @@ function foo(T1 $a, ?T2 $b, T3 $c) {}
...
     the reciprocal of the number raised to the positive opposite of the power.
+    That is, <literal>10<superscript>-2</superscript></literal> is the same as
+    <literal>1 / (10<superscript>2</superscript>)</literal>.
     Therefore raising <literal>0</literal> to the power of a negative number
-    corresponds to dividing by <literal>0</literal>.
-    Thus, this behaviour has been deprecated.
+    corresponds to dividing by <literal>0</literal>, i.e.
+    <literal>0<superscript>-2</superscript></literal> is the same as
+    <literal>1 / (0<superscript>2</superscript>)</literal>, or
+    <literal>1 / 0</literal>. Thus, this behaviour has been deprecated.
```
cmb69, line 93: American English nit:
```suggestion
    <literal>1 / 0</literal>. Thus, this behavior has been deprecated.
```

PR #3919, opened 2024-10-24
Remove developer books

Girgias:
Are these even useful nowadays?


cmb69 (approved):
> Are these even useful nowadays?

Have they ever been? I have no idea what their purpose would have been.

Anyway, we should probably adapt doc_gen to no longer generate these files.

Issue #3902, opened 2024-10-23
MHASH Constants no longer exist in PHP 8


Girgias:
From manual page: https://php.net/mhash.constants

---

The mhash functions have been deprecated since 8.1, but no idea what the status of the constants are and when they were removed or deprecated.

cmb69:
The constants are generally still there. See

https://github.com/php/php-src/blob/551a9ef5e66bbcd421a7a08ff869bb210e9299d8/ext/hash/hash.c#L57-L100

and their definition:

https://github.com/php/php-src/blob/551a9ef5e66bbcd421a7a08ff869bb210e9299d8/ext/hash/hash.c#L1185-L1193

And nope, they have not been deprecated so far. Wouldn't be hard to [do that](https://github.com/php/php-src/pull/16569), though. I don't think it would be useful to put them into the hash.stub.php, for usual treatment, given that mhash will be removed for PHP 9 anyway.

Girgias:
Okay, so I do need to fix the esotheric XML markup then. Also confused why 3v4l.org doesn't have them.

cmb69:
> so I do need to fix the esotheric XML markup then

Well, maybe. But more importantly, `MHASH_SHA192` and `MHASH_SNEFRU128` are listed there but are not supported (master; probably older branches, too). And `RIPEMD160`, `FNV132`, `FNV1A32`, `FNV164`, `FNV1A64`, `JOAAT`, `CRC32C`, `MURMUR3A`, `MURMUR3C`, `MURMUR3F`, `XXH32`, `XXH64`, `XXH3` and `XXH128` are supported, but are not listed there (need the `MHASH_` prefix). I can figure out the details of when introduced, and when removed, but I would like us to not step on our toes. So if you want to convert the markup, I'll wait.

> Also confused why 3v4l.org doesn't have them.

Needs `configure --with-mhash`.

Girgias:
> > so I do need to fix the esotheric XML markup then
> 
[...]

PR #3923, opened 2024-10-24
Update mhash constants page

cmb69:
We fix the uncommon markup, remove constants which are not available, and add constant which are available (including the PHP version they have been introduced). We do not, however, document the purpose of these legacy constants, which should be self-explaining anyway.

Closes GH-3902.

---

Might best be reviewed commit by commit.


Girgias (approved):
Thank you!

PR #3924, opened 2024-10-24
edit description for is_file (php#3719)

vera-ju:


reference/filesystem/functions/is-file.xml
```diff
@@ -13,7 +13,8 @@
    <methodparam><type>string</type><parameter>filename</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Tells whether the given file is a regular file.
+   Tells whether the given file is a regular file. It will also return true if <parameter>filename</parameter> is a
+   symbolic link.
```
cmb69, line 17: I'm not sure about that; doesn't it check the symlink's target instead?

vera-ju: Oh yes, thank you. You are right. I will change the description again :-)

cmb69:
This is now correct (I think), but I'm not sure how to handle this. Symlinks are usually resolved by all functions (there are some deliberate exceptions, though), but documenting that everywhere feels clumsy. Now, we have that already documented for `is_dir()`, but not for `file_exists()`. Maybe there should be some general info available somewhere?

@Girgias, thoughts?

Girgias (approved):
I think having a generic description of how file paths are resolved might be a good idea. But I will merge this PR as it is to get this moving forward.

PR #3925, opened 2024-10-24
Document clone behaviour of DOMXPath

nielsdos:
Is this useful to document? No idea.


Girgias (approved):
I think, it is just for completeness.

PR #3926, opened 2024-10-24
Add Jing to the Dockerfile so it can be used for validation

jimwins:
Along with php/doc-base#170 this runs Jing when the initial RelaxNG validation fails to get more useful error messages.


jordikroon:
As https://github.com/php/doc-en/pull/4399 added Java JRE to the docker image this MR became obsolete.
Thus closing this as completed. Thank you for your efforts.

PR #3932, opened 2024-10-25
language-snippets.ent Explain the meaning of nums and add tags

mmalferov:
Maybe we can explain what these numbers mean and get rid of the excessive number of parentheses? :)

### DIFF
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -3778,38 +3778,38 @@ local: {
  <itemizedlist xmlns="http://docbook.org/ns/docbook">
     <listitem>
      <simpara>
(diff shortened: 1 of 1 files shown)


Girgias (commented):
Probably just keeping the hexadecimal codepoint is fine?

cmb69:
We could also use the mnemonics (LF, CR, HT, etc.); see https://en.wikipedia.org/wiki/ASCII#Character_set. These could possibly be `<acronyms>`.

mmalferov:
I changed the list and added mnemonics to it. Please see if everything is correct now

Girgias (approved):

PR #3933, opened 2024-10-26
arrayiterator.xml Add tags

mmalferov:


reference/spl/arrayiterator.xml
```diff
@@ -11,13 +11,15 @@
    &reftitle.intro;
    <para>
     Allows the removal of elements, and the modification of
-    keys or values while iterating over Arrays or Objects.
+    keys or values while iterating over <link linkend="language.types.array">Arrays</link>
+    or <link linkend="language.types.object">Objects</link>.
```
cmb69, line 15: What about the following?
```suggestion
    keys or values while iterating over <type>array</type>s or <type>object</type>s.
```

reference/spl/arrayiterator.xml
```diff
@@ -11,13 +11,14 @@
...
    </para>
    <para>
     When you want to iterate over the same array multiple times you need to
-    instantiate ArrayObject and let it create ArrayIterator instances that
-    refer to it either by using &foreach; or by calling its getIterator()
-    method manually. 
+    instantiate <classname>ArrayObject</classname> and let it
+    create <classname>ArrayIterator</classname> instances that
+    refer to it either by using &foreach; or by calling the
+    <methodname>ArrayObject::getIterator</methodname> method manually.
```
Girgias, line 21: To remove the personalization, rephrase as:
```
To iterate the same array more than once, it is recommended to instantiate <classname>ArrayObject</classname> and use the <classname>ArrayIterator</classname> instances implicitly created when using &foreach;, or create one by calling the <methodname>ArrayObject::getIterator</methodname> method manually

mmalferov: Can we write ArrayIterator in the singular so that it doesn't seem like ArrayObject is creating multiple iterator instances in foreach, and add too more clarify about the array stored internally?

> To iterate the same array more than once, it is recommended to instantiate <classname>ArrayObject</classname> and use the <classname>ArrayIterator</classname> instance either implicitly created when using &foreach; to iterate over the array stored internally, or create one by calling the <methodname>ArrayObject::getIterator</methodname> method manually.

Girgias: Sure that works too

PR #3938, opened 2024-10-27
Remove cURL constants from function pages

haszi:
Remove cURL constants from function pages as these constants have already been moved to the cURL extension's constant page.


cmb69:
If linking works (sorry, I haven't closely followed), this seems to be a very good idea!

haszi:
> If linking works (sorry, I haven't closely followed), this seems to be a very good idea!

It's not working yet but it [could any time now](https://github.com/php/phd/pull/175). :-)

PR #3940, opened 2024-10-27
curl-multi-select.xml Replace `curl_multi` to `cURL multi handle`

mmalferov:
...and clarify number of which descriptors are returned


cmb69:
Thank you!

PR #3943, opened 2024-10-27
curl-multi-remove-handle.xml Correct the inaccuracy of the description

mmalferov:
The function removes a normal handle from the cURL multi handle, rather than multi handle


cmb69:
Thank you!

PR #3944, opened 2024-10-27
curl-multi-exec.xml Change the function purpose

mmalferov:
I'm not sure I'm offering a better description, perhaps a formulation like `Run the connections of handles contained in the given cURL multi handle [asynchronously]` would be better, or some similar. I was just a little confused by the words `sub-connections` (the explanation of which is not given, although it is clear that we are talking about the connections of descriptors contained in the multi-descriptor), and the `current` word... current in relation to what, and multiple `connections` for a `the current cURL handle` (it sounds like a normal handle contains a lot of connections).

And I also don't know if we should add a paragraph saying that the multi handle starts many connections at once, which can improve the performance of application

### DIFF
--- a/reference/curl/functions/curl-multi-exec.xml
+++ b/reference/curl/functions/curl-multi-exec.xml
@@ -14,7 +14,7 @@
    <methodparam><type>int</type><parameter role="reference">still_running</parameter></methodparam>
   </methodsynopsis>
   <para>
-   Processes each of the handles in the stack.  This method can be called whether or not a handle 
+   Processes each of the handles in the stack.  This method can be called whether or not a handle
    needs to read or write data.
   </para>
  </refsect1>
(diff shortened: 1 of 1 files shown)


reference/curl/functions/curl-multi-exec.xml
```diff
@@ -3,9 +3,9 @@
 <refentry xml:id="function.curl-multi-exec" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>curl_multi_exec</refname>
-  <refpurpose>Run the sub-connections of the current cURL handle</refpurpose>
+  <refpurpose>Run the connections of handles of a set of cURL handles</refpurpose>
```
kamil-tekiela, line 6: What does "running a connection" mean? Shouldn't it say more or less the same as the description below?

mmalferov: You're right, the word "connections" is confusing. Probably, it should have been formulated like this:

> Performs transfers on all easy handles added in cURL multi handle

English is not my native language, and again I'm not sure if I'm offering the best wording. I only know that the current function assignment is even worse ;)

> Run the sub-connections of the current cURL handle

kamil-tekiela: Yes, you could do that but I feel it would be better to keep it the same as below. 
```suggestion
  <refpurpose>Processes each of the handles in the stack</refpurpose>
```

PR #3946, opened 2024-10-27
getiterator.xml Specify the exact classname for the new iterator

mmalferov:
Can we specify which iterator is created using this method? This will also link the method and the specific iterator

### DIFF
--- a/reference/spl/arrayobject/getiterator.xml
+++ b/reference/spl/arrayobject/getiterator.xml
@@ -13,7 +13,8 @@
    <void/>
   </methodsynopsis>
   <para>
-   Create a new <classname>ArrayIterator</classname> from an <classname>ArrayObject</classname> instance.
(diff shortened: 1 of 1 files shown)


cmb69:
[ArrayObject::__construct](https://www.php.net/manual/en/arrayobject.construct.php) allows to specify the iterator class (third parameter), so this is not necessarily `ArrayIterator` (unfortunately).

mmalferov:
Yes, you're right, I've been thinking about it. Maybe like this?

`Create a new iterator (default is <classname>ArrayIterator</classname>) from an <classname>ArrayObject</classname> instance`.

cmb69:
Maybe also mark up `<interfacename>Iterator</interfacename>`?

mmalferov:
These links were missing. The ArrayObject and ArrayIterator class (along with the Iterator interface) lived separately 👍

cmb69:
Thank you!

Issue #3841, opened 2024-10-06
Link to short array syntax (via a See Also section?) from `array()` page.


Girgias:
From manual page: https://php.net/function.array

---

This would integrate note https://www.php.net/manual/en/function.array.php#115181

Girgias:
Fixed by #3947

PR #3947, opened 2024-10-28
explain short syntax in arrays

driade:
see https://github.com/php/doc-en/issues/3841

### DIFF
--- a/reference/array/functions/array.xml
+++ b/reference/array/functions/array.xml
@@ -14,7 +14,7 @@
   <para>
    Creates an array. Read the section on the
    <link linkend="language.types.array">array type</link> for more information
-   on what an array is.
+   on what an array is, including details about the alternative bracket syntax (<literal>[]</literal>).
   </para>
  </refsect1>
  <refsect1 role="parameters">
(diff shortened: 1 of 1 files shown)


cmb69:
Analogous to PR #3957, I think it's most important to link to the [arrays section](https://www.php.net/manual/en/language.types.array.php).

reference/array/functions/array.xml
```diff
@@ -50,6 +50,19 @@
    information on what an array is.
   </para>
  </refsect1>
+ <refsect1 role="short-syntax">
+  <title>Short Array Syntax (PHP 5.4+)</title>
+  <para>From PHP 5.4, arrays can use bracket syntax (`[]`) as a concise alternative to `array()`.</para>
```
cmb69, line 55: Note that the PHP manual currently is supposed to document PHP 7 and 8; as such referring to PHP 5.4 is superfluous.

reference/array/functions/array.xml
```diff
@@ -50,6 +50,19 @@
    information on what an array is.
   </para>
  </refsect1>
+ <refsect1 role="short-syntax">
```
cmb69, line 53: At least for `refsect1` (and related) only a couple of pre-defined `role`s are supported. Instead of a `refsect1`, this *could* be an `example` (although I'm not sure that we want to put an example in the description section).

Girgias:
This probably should be documented in the array type description and rework this file to refer back to the "short" syntax.

Girgias (approved):
LGTM

Issue #2405, opened 2023-04-05
htmlentities Example #1 is incorrect as of 8.1.0


basscl:
From manual page: https://php.net/function.htmlentities

---

Example 1:
<?php
$str = "A 'quote' is <b>bold</b>";

// Outputs: A 'quote' is &lt;b&gt;bold&lt;/b&gt;
echo htmlentities($str);

// Outputs: A &#039;quote&#039; is &lt;b&gt;bold&lt;/b&gt;
echo htmlentities($str, ENT_QUOTES);
?>

this was the output previous to 8.1.0, but as the default $flags parameter is now ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401, the two lines given will echo the same output.

PR #3948, opened 2024-10-28
fix example as it's not valid as of php 8.1

driade:
see https://github.com/php/doc-en/issues/2405


Girgias (commented):
Thank you!

While you are at it, would you mind converting the comments to a proper `<screen>` tag for output?

driade:
@Girgias there you have

Girgias (approved):
Thank you!

PR #3950, opened 2024-10-28
Add type information for UPLOAD_* constants

Girgias:


Girgias:
Actually realised it was at the wrong place, could you re-review @cmb69 ?

cmb69 (commented):
Maybe we should keep the error constant values; these can be useful for development, and especially some quick experiments.

Otherwise, this is good.

PR #3951, opened 2024-10-28
curl-multi-select.xml Describe in more detail

mmalferov:
The term `activity` is too general, and does not explain exactly what the function is waiting for. Besides, at least for me, it's not obvious what exactly the function is blocking: the PHP script, the current connection, all of the cURL multi handle connections?..

I propose to consider a more detailed description of the function

### DIFF
--- a/reference/curl/functions/curl-multi-select.xml
+++ b/reference/curl/functions/curl-multi-select.xml
@@ -3,10 +3,7 @@
 <refentry xml:id="function.curl-multi-select" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>curl_multi_select</refname>
-  <refpurpose>
-   Wait until reading or writing is possible
-   for any cURL multi handle connection
-  </refpurpose>
+  <refpurpose>Wait until reading or writing is possible for any cURL multi handle connection</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
--- a/reference/curl/functions/curl-multi-select.xml
+++ b/reference/curl/functions/curl-multi-select.xml
@@ -3,7 +3,10 @@
 <refentry xml:id="function.curl-multi-select" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>curl_multi_select</refname>
-  <refpurpose>Wait for activity on any cURL multi handle connection</refpurpose>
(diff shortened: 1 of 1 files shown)


reference/curl/functions/curl-multi-select.xml
```diff
@@ -14,7 +17,10 @@
    <methodparam choice="opt"><type>float</type><parameter>timeout</parameter><initializer>1.0</initializer></methodparam>
   </methodsynopsis>
   <para>
-    Blocks until there is activity on any of the cURL multi handle connections.
+   Blocks the execution of the script while there are
+   any the cURL multi handle connections reading data from or
+   writing data to easy cURL handles, or until a timeout (if it is
+   not equal to <literal>0</literal>) occurs, whichever comes first.
```
TimWolla, line 23: This is not quite correct. It does not block *while* reading/writing. It blocks *until* reading or writing is possible.

- Reading becomes possible if the remote end has sent some data that is sitting in the kernel buffers.
- Writing becomes possible if the kernel sent some of the data sitting in the kernel buffers and they are no longer full.

So an accurate description would be something like the following:

> Blocks the execution of the script until a curl handle attached to the curl multi handle would be able to make progress on the next call to `curl_multi_exec` or until the timeout strikes (whichever comes first).

The `refpurpose` would need to be adjusted similarly.

reference/curl/functions/curl-multi-select.xml
```diff
@@ -3,7 +3,10 @@
 <refentry xml:id="function.curl-multi-select" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>curl_multi_select</refname>
-  <refpurpose>Wait for activity on any cURL multi handle connection</refpurpose>
+  <refpurpose>
+   Wait until reading or writing is possible
+   for any cURL multi handle connection
+  </refpurpose>
```
cmb69, line 9: I would keep a single line; if it is too long, the refpurpose is likely too long as well. In this case, the length may be just okay.

cmb69 (approved):
Thank you! This looks good to me. Maybe @TimWolla wants to have another look?

TimWolla (approved):

cmb69:
Thank you!

PR #3954, opened 2024-10-28
Document 'process' resource type

jimwins:
Based on comment: https://www.php.net/manual/en/resource.php#128764

### DIFF
--- a/appendices/resources.xml
+++ b/appendices/resources.xml
@@ -1303,6 +1303,20 @@
      <entry> </entry>
     </row>
    <row>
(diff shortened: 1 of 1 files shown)


cmb69 (approved):

Girgias:
This page should ideally be XIncludes and/or auto-generated. But I don't know if XInclude is currently possible as it might have IDs and libxml does not support XInclude 1.1

cmb69:
I wouldn't worry to much about the resource types page; in a couple of years we may be able to drop it. :)

PR #3956, opened 2024-10-29
[PHP 8.4] Add manual for grapheme_str_split function

youkidearitai:
Add grapheme_str_split function.

RFC:
- https://wiki.php.net/rfc/grapheme_str_split

ref: #3872

### DIFF
--- a/reference/intl/grapheme/grapheme-str-split.xml
+++ b/reference/intl/grapheme/grapheme-str-split.xml
@@ -1,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.grapheme-str-split" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>grapheme_str_split</refname>
+  <refpurpose>Given a grapheme cluster string, which must be encoded in UTF-8</refpurpose>
+ </refnamediv>
+ <refsect1 role="description">
+  &reftitle.description;
+  <para>&style.procedural;</para>
+  <methodsynopsis>
+   <type class="union"><type>array</type><type>false</type></type><methodname>grapheme_str_split</methodname>
+   <methodparam><type>string</type><parameter>string</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>length</parameter><initializer>1</initializer></methodparam>
+  </methodsynopsis>
+  <para>
+   This function will return an array of strings, it is a version of <function>str_split</function> with support for grapheme cluster byte characters.
+   If the <parameter>length</parameter> parameter is specified, the string is broken down into chunks of the specified length in grapheme clusters (not bytes).
+  </para>
+  </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>string</parameter></term>
+     <listitem>
+      <para>
+       The &string; to split into grapheme clusters or chunks. Must be valid UTF-8.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>length</parameter></term>
+     <listitem>
+      <para>
+        If specified, each element of the returned array will be composed of grapheme clusters instead of a single grapheme cluster.
+      </para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   <function>grapheme_str_split</function> returns an array of strings.
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>str_split</function></member>
+    <member><function>mb_str_split</function></member>
+     <member>
+      <link xlink:href="&uri.unicode.graphemes;">
+       Unicode Text Segmentation: Grapheme Cluster Boundaries
+      </link>
+     </member>
+   </simplelist>
+  </para>
+ </refsect1>
+</refentry>
+<!-- Keep this comment at the end of the file
+Local variables:
+mode: sgml
+sgml-omittag:t
+sgml-shorttag:t
+sgml-minimize-attributes:nil
+sgml-always-quote-attributes:t
+sgml-indent-step:1
+sgml-indent-data:t
+indent-tabs-mode:nil
+sgml-parent-document:nil
+sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+sgml-exposed-tags:nil
+sgml-local-catalogs:nil
+sgml-local-ecat-files:nil
+End:
+vim600: syn=xml fen fdm=syntax fdl=2 si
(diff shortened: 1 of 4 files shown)


reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.grapheme-str-split" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <refnamediv>
+  <refname>grapheme_str_split</refname>
+  <refpurpose>Given a grapheme cluster string, which must be encoded in UTF-8</refpurpose>
```
cmb69, line 6: This doesn't look like a good description of the purpose of this function. `str_split()` uses "Convert a string to an array", which is better, but also somewhat unclear. Maybe:
```suggestion
  <refpurpose>Split a string into an array</refpurpose>
```

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,87 @@
...
+ </refnamediv>
+ <refsect1 role="description">
+  &reftitle.description;
+  <para>&style.procedural;</para>
```
cmb69, line 10: I think this `para` is superfluous.

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,87 @@
...
+  <methodsynopsis>
+   <type class="union"><type>array</type><type>false</type></type><methodname>grapheme_str_split</methodname>
+   <methodparam><type>string</type><parameter>string</parameter></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>length</parameter><initializer>1</initializer></methodparam>
+  </methodsynopsis>
+  <para>
+   This function will return an array of strings, it is a version of <function>str_split</function> with support for grapheme cluster byte characters.
+   If the <parameter>length</parameter> parameter is specified, the string is broken down into chunks of the specified length in grapheme clusters (not bytes).
```
cmb69, line 18: This lines are too long.

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,87 @@
...
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>string</parameter></term>
+     <listitem>
+      <para>
+       The &string; to split into grapheme clusters or chunks. Must be valid UTF-8.
```
cmb69, line 30: It is now common to avoid `&string;`, but to use `<type>string</type>` instead.

I also wonder what happens if invalid UTF-8 is given; would the function detect this, or would it return an inproper split?

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,87 @@
...
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><parameter>length</parameter></term>
+     <listitem>
+      <para>
+        If specified, each element of the returned array will be composed of grapheme clusters instead of a single grapheme cluster.
```
cmb69, line 38: While the default is `1`, that does not imply that the user cannot explicitly specify a `$length` of `1`. And I don't think there is anything special about `1`. Maybe:
```suggestion
        Each element of the returned array will be composed of <parameter>length</parameter> grapheme clusters.
```

Also, what happens if a non-positive `$length` is given.

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,87 @@
...
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>str_split</function></member>
+    <member><function>mb_str_split</function></member>
+     <member>
+      <link xlink:href="&uri.unicode.graphemes;">
+       Unicode Text Segmentation: Grapheme Cluster Boundaries
+      </link>
```
cmb69, line 62: Ah, very nice!

cmb69 (commented):
Thank you for the PR!

reference/intl/grapheme/grapheme-str-split.xml
youkidearitai: Removed.

reference/intl/grapheme/grapheme-str-split.xml
youkidearitai: This lines are separated.

reference/intl/grapheme/grapheme-str-split.xml
youkidearitai: Use `<type>string</type>` instead.

> I also wonder what happens if invalid UTF-8 is given; would the function detect this, or would it return an inproper split?

Don't detect invalid UTF-8, returns improper split array. So I added condition that returns false.

reference/intl/grapheme/grapheme-str-split.xml
youkidearitai: Thanks!

reference/strings/functions/str-split.xml
```diff
@@ -151,6 +151,7 @@ Array
    <para>
     <function>str_split</function> will split into bytes, rather than characters when dealing with a multi-byte encoded string.
     Use <function>mb_str_split</function> to split the string into code points. 
+    Use <function>grapheme_str_split</function> to split the string into grapheme clusetrs.
```
cmb69, line 154: I think we should reword this, while we're at it:
```suggestion
    <function>mb_str_split</function> can be used to split the string into code points. 
    <function>grapheme_str_split</function> can be used to split the string into grapheme clusters.
```

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,98 @@
...
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>string</parameter></term>
+     <listitem>
+      <para>
+       The <type>string</type> to split into grapheme clusters or chunks. Support only UTF-8.
```
cmb69, line 33:
```suggestion
       The <type>string</type> to split into grapheme clusters or chunks.
       <parameter>string</parameter> must be valid UTF-8.
```

reference/intl/grapheme/grapheme-str-split.xml
```diff
@@ -0,0 +1,98 @@
...
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <para>
+   <function>grapheme_str_split</function> returns an array of strings. Returns false when opening text fails.
```
cmb69, line 52: While "opening" is technically correct, and also corresponds to the intl error message, I think we should just use:
```suggestion
   <function>grapheme_str_split</function> returns an array of strings, &return.falseforfailure;.
```
No need to state again that the string needs to be valid UTF-8.

cmb69 (commented):
Thank you! Besides some minor issue, this looks good to me.

cmb69:
Thank you!

Issue #1753, opened 2022-08-17
Array destructuring documentation and list()


ollieread:
There doesn't seem to be a page on array destructuring in the docs, but when google for "php array destructuring" the only php.net link that appears, is for `list()`.

![CleanShot 2022-08-17 at 11 47 37](https://user-images.githubusercontent.com/469515/185100494-76c7ad98-aa15-49d0-b15c-89efbb157e3f.png)

Interestingly, this displays the following text, that doesn't appear on the page, that I can fine.

> This enables destructuring of arrays with non-integer or non-sequential keys.

If I am understanding this correctly, the shorthand syntax of 

```php
[$one, $two] = $array;
```

Is identical to

```php
list($one, $two) = $array;
```

If this is the case, the `list()` page should be updated.

(RFC for feature: https://wiki.php.net/rfc/short_list_syntax)

Girgias:
Fixed by #3957

PR #3957, opened 2024-10-29
explain shorthand syntax

driade:
see https://github.com/php/doc-en/issues/1753


cmb69:
Thank you for the PR! I'm not sure, though, that we should add an example on the `list()` page. It seems to more sensible to link to the [array destructuring section](https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.destructuring), likely from the description section.

driade:
Let's see if this works better.

Girgias (approved):

PR #3963, opened 2024-10-29
Rework stream constants page

Girgias:
This organizes the constants into dedicated lists relevant to their usage.

There are some to-dos about things I noticed while I reworked this page.


jimwins (approved):
I didn't look for/at any content changes, but comparing the before/after layout, it looks good. We might want to consider adding more vertical space between the `<title>` of a `<variablelist>` and the entries, but that's not an issue unique to this.

PR #3964, opened 2024-10-29
[PHP 8.4] Add errors for GetText functions

saundefined:


reference/gettext/functions/dgettext.xml
```diff
@@ -49,7 +49,38 @@
    A <type>string</type> on success.
   </para>
  </refsect1>
- 
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   Throws a <classname>ValueError</classname> if <parameter>domain</parameter>
```
Girgias, line 56: Should be `<exceptionname>`

Girgias (changes requested):
Tag should be amended

reference/gettext/functions/bind-textdomain-codeset.xml
```diff
@@ -52,6 +52,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   Throws a <exceptionname>ValueError</exceptionname> if <parameter>domain</parameter>
+   is the empty &string;.
```
cmb69, line 59: As I understand it, nowaday the following is preferred:
```suggestion
   is the empty <type>string</type>.
```

saundefined: I missed that, thanks! I'll keep that in mind

Girgias:
Yes, except if we want to wait for the PR that uses PI to land and use those instead?

PR #3965, opened 2024-10-29
[PHP 8.4] Add errors for curl_multi_select

saundefined:


reference/curl/functions/curl-multi-select.xml
```diff
@@ -50,6 +50,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   Throws a <classname>ValueError</classname> if <parameter>timeout</parameter>
```
Girgias, line 56: `<exceptionname>` for ValueError.

Girgias (commented):
I wonder if it would make more sense to create a `<?phpdoc` instructor to generate these sorts of messages.

These seem to only be used by PhD, but surely one could "resolve" them in the configure.php script

PR #3966, opened 2024-10-29
Add PIs to render value errors

Girgias:
Instead of #3965 prototype for https://github.com/php/doc-base/pull/173

### DIFF
--- a/reference/curl/functions/curl-multi-select.xml
+++ b/reference/curl/functions/curl-multi-select.xml
@@ -50,6 +50,11 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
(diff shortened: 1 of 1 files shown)


lacatoire:
Hey @Girgias Worth keeping open for doc-base#173, or close it?

jordikroon:
Perhaps @alfsb might have an opinion on this as well.

alfsb:
This is something that I would like to tackle, yes. It is annotated on doc-base 199 as `Make <?phpdoc more general`, but is something I cannot foresee working on before the R&D for userland XML assembler is fully completed or abandoned. I think that the PR on doc-base sufices to track this improvement.

Girgias:
Eh probably close as this is stale and while I would like to work on something like this in the future, but validating XML sources via RelaxNG with PIs will be a mess so :/

PR #3970, opened 2024-10-30
Update curl-exec.xml Mention the direct output

mmalferov:
It seems to me an omission that in the description of the `curl_exec()` there is no mention of the transfer output to the browser (if https://www.php.net/manual/en/curl.constants.php#constant.curlopt-returntransfer option is not set)


cmb69:
Thank you!

PR #3972, opened 2024-10-30
[PHP 8.4] Add manual for `pg_result_memory_size()`

KentarouTakeda:
Ref: php/php-src#14214

### DIFF
--- a/reference/pgsql/functions/pg-result-memory-size.xml
+++ b/reference/pgsql/functions/pg-result-memory-size.xml
@@ -42,10 +42,9 @@
 
  <refsect1 role="examples">
   &reftitle.examples;
-  <para>
-   <example>
-    <title><function>pg_result_memory_size</function> example</title>
-    <programlisting role="php">
+  <example>
+   <title><function>pg_result_memory_size</function> example</title>
+   <programlisting role="php">
 <![CDATA[
 <?php
 $db = pg_connect("dbname=users user=me");
@@ -57,15 +56,14 @@ $size = pg_result_memory_size($res);
 var_dump($size);
 ?>
 ]]>
-    </programlisting>
-    &example.outputs.similar;
-    <screen>
+   </programlisting>
+   &example.outputs.similar;
+   <screen>
 <![CDATA[
 int(3288)
 ]]>
-    </screen>
-   </example>
-  </para>
+   </screen>
+  </example>
  </refsect1>
 
  <refsect1 role="seealso">
--- a/reference/pgsql/functions/pg-result-memory-size.xml
+++ b/reference/pgsql/functions/pg-result-memory-size.xml
@@ -14,9 +14,8 @@
    <methodparam><type>PgSql\Result</type><parameter>result</parameter></methodparam>
   </methodsynopsis>
   <simpara>
-   Returns the amount of memory, in bytes, allocated to the specified query result <classname>PgSql\Result</classname> instance.
-  </simpara>
-  <simpara>
+   Returns the amount of memory, in bytes, allocated to the specified query result
+   <classname>PgSql\Result</classname> instance.
    This value is the same amount that would be freed by <function>pg_free_result</function>.
   </simpara>
  </refsect1>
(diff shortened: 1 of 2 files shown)


reference/pgsql/functions/pg-result-memory-size.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<!-- splitted from ./en/functions/pgsql.xml, last change in rev 1.2 -->
```
cmb69, line 3: Superfluous line; should be removed.

reference/pgsql/functions/pg-result-memory-size.xml
```diff
@@ -0,0 +1,102 @@
...
+  <methodsynopsis>
+   <type>int</type><methodname>pg_result_memory_size</methodname>
+   <methodparam><type>PgSql\Result</type><parameter>result</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   Returns the amount of memory, in bytes, allocated to the specified query result <classname>PgSql\Result</classname> instance.
+  </para>
+  <para>
+   This value is the same amount that would be freed by <function>pg_free_result</function>.
+  </para>
```
cmb69, line 21: Wem prefer to use `simpara` instead of `para` nowdays, where possible:
```suggestion
  <simpara>
   Returns the amount of memory, in bytes, allocated to the specified query result <classname>PgSql\Result</classname> instance.
  </simpara>
  <simpara>
   This value is the same amount that would be freed by <function>pg_free_result</function>.
  </simpara>
```
And maybe the two paragraphs should be collapsed, and the long line should be split:
```suggestion
  <simpara>
   Returns the amount of memory, in bytes, allocated to the specified query result
   <classname>PgSql\Result</classname> instance.
   This value is the same amount that would be freed by <function>pg_free_result</function>.
  </simpara>
```

reference/pgsql/functions/pg-result-memory-size.xml
```diff
@@ -0,0 +1,102 @@
...
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>pg_result_memory_size</function> example</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+$db = pg_connect("dbname=users user=me") || die();
```
cmb69, line 53: At least this should use `or` instead of `||` (different operator precedence), but I would just drop it since proper error handling is not necessary for the example at hand:
```suggestion
$db = pg_connect("dbname=users user=me");
```

cmb69 (commented):
Thank you for the PR!

reference/pgsql/functions/pg-result-memory-size.xml
KentarouTakeda: Fixed with 1c0d2f4

reference/pgsql/functions/pg-result-memory-size.xml
KentarouTakeda: Fixed with 36ede60

reference/pgsql/functions/pg-result-memory-size.xml
KentarouTakeda: Fixed 8266b84 and be8b31a. In addition, `simpara` can be used in 64bb230 as well.

reference/pgsql/functions/pg-result-memory-size.xml
```diff
@@ -0,0 +1,100 @@
...
+]]>
+    </programlisting>
+    &example.outputs.similar;
+    <screen>
+<![CDATA[
+int(3288)
+]]>
+    </screen>
+   </example>
+  </para>
```
cmb69, line 68: Ah, overlooked this on the first review. `<para>` around `<example>` is useless, and can be removed:
```suggestion
  <example>
   <title><function>pg_result_memory_size</function> example</title>
   <programlisting role="php">
<![CDATA[
<?php
$db = pg_connect("dbname=users user=me");

$res = pg_query($db, 'SELECT 1');

$size = pg_result_memory_size($res);

var_dump($size);
?>
]]>
   </programlisting>
   &example.outputs.similar;
   <screen>
<![CDATA[
int(3288)
]]>
   </screen>
  </example>
```

cmb69 (commented):
Thank you! Almost good (one minor issue below).

reference/pgsql/functions/pg-result-memory-size.xml
KentarouTakeda: Exactly! Thanks!
Fixed with 31ddb1f

cmb69 (approved):
Thank you very much!

PR #3976, opened 2024-10-30
weakreference.xml Give more details about the WeakReferences

mmalferov:
Maybe we should describe in more detail the features of weak references?

I am not sure about the correctness of the wording, but the scant information about WeakReference needs to be supplemented

### DIFF
--- a/language/predefined/weakreference.xml
+++ b/language/predefined/weakreference.xml
@@ -13,6 +13,13 @@
    <para>
     Weak references allow the programmer to retain a reference to an object which does not prevent
     the object from being destroyed. They are useful for implementing cache like structures.
+    The garbage collector will destroy the original object, and a weak reference will return &null;
+    when calling the <methodname>WeakReference::get<methodname> method,
+    only if there are no variables left that store the
+    <link linkend="language.oop5.references">ID of the original object</link>,
+    more precisely, the object is destroed only when the
+    <link linkend="features.gc.refcounting-basics">refcount</link> for the object is reset to zero;
+    at the same time, creating a weak references do not increase the reference counter.
    </para>
    <para>
     <classname>WeakReference</classname>s cannot be serialized.
--- a/language/predefined/weakreference.xml
+++ b/language/predefined/weakreference.xml
@@ -13,13 +13,11 @@
    <para>
     Weak references allow the programmer to retain a reference to an object which does not prevent
     the object from being destroyed. They are useful for implementing cache like structures.
-    The garbage collector will destroy the original object, and a weak reference will return &null;
-    when calling the <methodname>WeakReference::get</methodname> method,
-    only if there are no variables left that store the
-    <link linkend="language.oop5.references">ID of the original object</link>,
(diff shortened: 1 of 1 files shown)


language/predefined/weakreference.xml
```diff
@@ -13,6 +13,13 @@
    <para>
     Weak references allow the programmer to retain a reference to an object which does not prevent
     the object from being destroyed. They are useful for implementing cache like structures.
+    The garbage collector will destroy the original object, and a weak reference will return &null;
+    when calling the <methodname>WeakReference::get</methodname> method,
+    only if there are no variables left that store the
+    <link linkend="language.oop5.references">ID of the original object</link>,
+    more precisely, the object is destroed only when the
+    <link linkend="features.gc.refcounting-basics">refcount</link> for the object is reset to zero;
+    at the same time, creating a weak references do not increase the reference counter.
```
jimwins, line 22:
```suggestion
    If the original object has been destroyed, &null; will be returned 
    when calling the <methodname>WeakReference::get</methodname> method.
    The original object will be destroyed when the
    <link linkend="features.gc.refcounting-basics">refcount</link> for it becomes zero;
    creating weak references do not increase the <literal>refcount</literal>.
```

jimwins (commented):
Maybe we should also add something on `WeakReference` and an example to `language/references.xml`

mmalferov:
It's a good idea. Probably, the best section for WR it is a Unsetting References, because the peculiarity of WR is precisely how they behave when destroying variables that store a pointer to an object

I am not the best author, so if you help me write a description, I can add another commit to this PR (or, if it would be better, in different PR) with changes in the file 'language/references.xml' ;)

Girgias:
I think opening an issue about mentioning WeakReference/WeakMap in the `language/reference.xml` page might be better to track this.

language/predefined/weakreference.xml
```diff
@@ -13,6 +13,11 @@
    <para>
     Weak references allow the programmer to retain a reference to an object which does not prevent
     the object from being destroyed. They are useful for implementing cache like structures.
+    If the original object has been destroyed, &null; will be returned
+    when calling the <methodname>WeakReference::get</methodname> method.
+    The original object will be destroyed when the
+    <link linkend="features.gc.refcounting-basics">refcount</link> for it becomes zero;
```
Girgias, line 19: Might be better
```suggestion
    <link linkend="features.gc.refcounting-basics">refcount</link> for it drops to zero;
```

language/predefined/weakreference.xml
```diff
@@ -13,6 +13,11 @@
...
+    creating weak references do not increase the <literal>refcount</literal>.
```
Girgias, line 20:
```suggestion
    creating weak references does not increase the <literal>refcount</literal> of the object being referenced.
```

Girgias (commented):
I don't know the English grammar rules around semicolons, so will need someone else to judge if this is correct.

Girgias (approved):
Thank you!

PR #3977, opened 2024-10-31
Group math constants and use standard <variablelist> tag

Girgias:
This also adds a note to recommend using the new RoundingMode enum.

### DIFF
--- a/reference/math/constants.xml
+++ b/reference/math/constants.xml
@@ -3,163 +3,297 @@
 <appendix xml:id="math.constants" xmlns="http://docbook.org/ns/docbook">
  &reftitle.constants;
  &extension.constants.core;
- <para>
-  <table>
-   <title>Math constants</title>
-   <tgroup cols="4">
-    <thead>
-     <row>
-      <entry>Constant</entry>
-      <entry>Value</entry>
-      <entry>Description</entry>
-      <entry>Availability</entry>
-     </row>
-    </thead>
-    <tbody>
-     <row xml:id="constant.m-pi">
-      <entry><constant>M_PI</constant></entry>
-      <entry>3.14159265358979323846</entry>
-      <entry>Pi</entry>
-      <entry></entry>
-     </row>
-     <row xml:id="constant.m-e">
-      <entry><constant>M_E</constant></entry>
-      <entry>2.7182818284590452354</entry>
-      <entry>e</entry>
-      <entry></entry>
-     </row>
-     <row xml:id="constant.m-log2e">
-      <entry><constant>M_LOG2E</constant></entry> 
-      <entry>1.4426950408889634074</entry>
-      <entry>log_2 e</entry>
-      <entry></entry>
-     </row>
-     <row xml:id="constant.m-log10e">
-      <entry><constant>M_LOG10E</constant></entry>
-      <entry>0.43429448190325182765</entry>
-      <entry>log_10 e</entry>
-      <entry></entry>
-     </row>
-     <row xml:id="constant.m-ln2">
-      <entry><constant>M_LN2</constant></entry>   
-      <entry>0.69314718055994530942</entry>
-      <entry>log_e 2</entry>
-      <entry></entry>
-     </row>
-     <row xml:id="constant.m-ln10">
-      <entry><constant>M_LN10</constant></entry>  
-      <entry>2.30258509299404568402</entry>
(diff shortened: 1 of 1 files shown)


reference/math/constants.xml
```diff
@@ -3,163 +3,297 @@
...
+   <listitem>
+    <simpara>
+     Infinity
+    </simpara>
+   </listitem>
+  </varlistentry>
+ </variablelist>
 
+ <variablelist>
+  <title>Rounding constants</title>
```
Girgias, line 242: Maybe this should mention `<function>round</function>`

reference/math/constants.xml
```diff
@@ -3,163 +3,297 @@
...
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.m-e">
+   <term>
+    <constant>M_E</constant>
+    (<type>float</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Approximation of Euler's number <emphasis>e</emphasis>
```
cmb69, line 28: While I understand the intention to get *e*, I don't think this is appropriate markup. Unfortunately, [Unicode](https://www.compart.com/en/unicode/U+2107) screwed up, so we can't use it. Probably best to stick with `<literal>e</literal>` for now.

reference/math/constants.xml
```diff
@@ -3,163 +3,297 @@
...
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.m-log2e">
+   <term>
+    <constant>M_LOG2E</constant>
+    (<type>float</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Approximation of <literal>log_2(e)</literal>
```
cmb69, line 40: What about
```suggestion
     Approximation of <literal>log<subscript>2</subscript>e</literal>
```
Looks quite good for me:

log

I don't mind to have parens around the argument or not. Anyhow, the other `log`s below could use the same.

reference/math/constants.xml
```diff
@@ -3,163 +3,297 @@
...
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.m-ln2">
+   <term>
+    <constant>M_LN2</constant>
+    (<type>float</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Approximation of <literal>log_e(2)</literal>
```
cmb69, line 64: This could use `ln`:
```suggestion
     Approximation of <literal>ln(2)</literal>
```
Again, with or without parens.

And I just learned that ISO introduced `lb` (for log base 2) and `lg` (for log base 10). Not sure if we should use these.

reference/math/constants.xml
```diff
@@ -3,163 +3,297 @@
...
-      <entry>INF (as a float)</entry>
-      <entry>The infinite</entry>
-      <entry></entry>
-     </row>
-    </tbody>
-   </tgroup>
-  </table>
- </para>
-</appendix>
+ <!-- TODO: Use MathML the moment PhD supports it -->
```
cmb69, line 6: I'm still not convinced that MathML is a good idea. The uses are probably limited, some stuff might be confusing for programmers not having a solid mathematical background, and it likely requires to deliver a bunch of JS for older browsers. However, a different discussion.

cmb69 (commented):
Thank you! Generally very sensible, but maybe we can improve a bit (see comments below).

reference/math/constants.xml
cmb69: Well, as soon as we support `<enumname>`, I don't think there's the need to link to `round()`.

reference/math/constants.xml
Girgias: I never heard of `lb` and `lg` so I'd rather not.

reference/math/constants.xml
Girgias: There exists a basic CSS profile to render MathML for simple elements: https://github.com/fred-wang/mathml.css but yes different discussion.

reference/math/constants.xml
Girgias: Euler's constant is not the same as Euler's number (yes not at all confusing) but even then, it seems like it is usually a lowercase gamma rather than an epsilon so Unicode messed up anyway

cmb69 (approved):

PR #3978, opened 2024-10-31
Add documentation for new MyClass()->method() without parentheses

cookieguru:
New feature in core mentioned in #3872

https://wiki.php.net/rfc/new_without_parentheses

### DIFF
--- a/language/oop5/basic.xml
+++ b/language/oop5/basic.xml
@@ -374,7 +374,7 @@ bool(true)
 <![CDATA[
 <?php
 echo (new DateTime())->format('Y');
-// surrounding parentheses are optional since 8.4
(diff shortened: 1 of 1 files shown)


language/oop5/basic.xml
```diff
@@ -374,6 +374,8 @@ bool(true)
 <![CDATA[
 <?php
 echo (new DateTime())->format('Y');
+// surrounding parentheses are optional since 8.4
```
cmb69, line 377: Nit:
```suggestion
// surrounding parentheses are optional as of PHP 8.4.0
```

cmb69 (commented):
Thank you! See comment below; otherwise this is good (it seems that this syntax is not mentioned anywhere else).

cmb69:
Thank you!

PR #3981, opened 2024-10-31
PDO Drivers: remove the visual noice

mmalferov:
Take a look at these screenshots instead of a thousand words:

### Before

visual_garbage

menu_visual_garbage

### After

clear_view

menu_clear_view

The names of the drivers in the list follow the general heading: **PDO Drivers,** so it is clear that each element of the list is a separate driver name, and the suffix ` (PDO)` only creates visual noise, so suffixes should be removed

### DIFF
--- a/reference/pdo_cubrid/reference.xml
+++ b/reference/pdo_cubrid/reference.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <reference xml:id="ref.pdo-cubrid" xmlns="http://docbook.org/ns/docbook"><?phpdoc extension-membership="pecl" ?>
  <title>CUBRID Functions (PDO_CUBRID)</title>
- <titleabbrev>CUBRID (PDO)</titleabbrev>
+ <titleabbrev>CUBRID</titleabbrev>
  <partintro>
 
   <section xml:id="pdo-cubrid.intro">&reftitle.intro;
--- a/reference/pdo_dblib/reference.xml
+++ b/reference/pdo_dblib/reference.xml
@@ -4,7 +4,7 @@
  <reference xml:id="ref.pdo-dblib" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
   <?phpdoc extension-membership="bundledexternal" ?>
   <title>Microsoft SQL Server and Sybase Functions (PDO_DBLIB)</title>
-  <titleabbrev>MS SQL Server (PDO)</titleabbrev>
+  <titleabbrev>MS SQL Server</titleabbrev>
   <partintro>
 
    <section xml:id="pdo-dblib.intro">
--- a/reference/pdo_firebird/reference.xml
+++ b/reference/pdo_firebird/reference.xml
@@ -10,8 +10,9 @@
    <section xml:id="pdo-firebird.intro">
    &reftitle.intro;
     <para>
-     PDO_FIREBIRD is a driver that implements the PHP Data Objects (PDO)
-     interface to enable access from PHP to Firebird database.
+     PDO_FIREBIRD is a driver that implements the <link linkend="intro.pdo">PHP
+     Data Objects (PDO) interface</link>
+     to enable access from PHP to Firebird database.
     </para>
    </section>
    &reference.pdo-firebird.configure;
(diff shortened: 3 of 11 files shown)


kamil-tekiela:
Makes sense, but I don't know how I feel about this. I believe the idea was to make it obvious in the title of the page that it's for PDO driver as compared to the dedicated drivers. See this for example:

image

or in the browser tab:
image

Having PDO in the title is actually useful in distinguishing it from the other pages. That's why I think we should not remove this.

cmb69:
Indeed, the `titleabbrev` may be used elsewhere where having the "PDO" makes sense.

mmalferov:
### 0

There is a saying in the place where I live: _Transfer from a sore head to a healthy one_, which means: to shift from the guilty (the abbr-titles instead full titles, the hidden breadcrums on mobile devices instead horizontall scrolled, the search engine snippets instead of convenience for users etc.) to the innocent (the abbr-titles of a PDO Drivers pages :-)

### 1

> Having PDO in the title is actually useful in distinguishing it from the other pages.

The current headlines interfere with users, create visual noise and nothing else. In the example you gave, you can clearly see which page you are talking about, if you imagine that the garbage ` (PDO)` was removed from the header.

Otherwise, we need to add the names of the parent levels to each documentation header like predefined constants to see exactly which predefined constants we are talking about: [https://www.php.net/manual/en/uodbc.constants.php](https://www.php.net/manual/en/uodbc.constants.php)

— It's a terse headline, isn't it? :)

And also do the same with all the other sections that violate the logic you suggested. Otherwise, it is not clear why it applies only to the page with the list of PDO drivers.

Maybe the tab headings are unclear because they should be replaced with full headings, not abbreviated ones. And show a breadcrumbs with horizontal scroll on mobile devices.

In any case, users don't deserve to see `_(PDO)` 11 times in a row ))

For example, [this page](https://www.php.net/manual/en/book.pgsql.php) should look like this, otherwise it's not clear which section we're in right now:

Screenshot 2024-11-01 032941


### 2

> Indeed, the titleabbrev may be used elsewhere where having the "PDO" makes sense.

A specific example of such a page would be more convincing ;-)

### 3

Is it really true that nothing jumps inside you when you look at a list that says `_(PDO)`, `_(PDO)`, `_(PDO)`,... — 11 times in a row?!. I don't believe in it.

The headings on the page are the headings of the page, they are primarily for people. Tab headings are also for people, but I dare say that to a much lesser extent and they are looked at less often; and the responsibility of page headings does not include responsibility for tab headings. Search engines without `_(PDO)` they show enough information to understand where the link leads. Very briefly: I don't see any good reason to leave ` (PDO)`, and the reason to delete them is people

Why breadcrumbs are hidden on small—screen devices is a big mystery...

### P.S.

My tabs look like this. Therefore, only hovering the mouse gives me the full name of the page, so I (like, apparently, you) don't care, they wrote in the title of the tab ` (PDO)` or not

tabs


### P.P.S.

We've done this trick with headlines before, and then there were no objections: https://github.com/php/doc-en/pull/3117

kamil-tekiela:
> Otherwise, we need to add the names of the parent levels to each documentation header like predefined constants to see exactly which predefined constants we are talking about: https://www.php.net/manual/en/uodbc.constants.php

That would actually be nice. 

---

I am not completely against this change. I completely see where you are coming from with it, and it makes sense. But in this particular case, it's not much of a gain. It would be better if we had a solution so we could have the cake and eat it, too. I haven't looked at how the titles are generated. I am unsure what the difference is between short and full titles. 

There are other issues I can see with these pages. For example, why do we have a table and a list with the same information in https://www.php.net/manual/en/pdo.drivers.php I would say that this page needs a more holistic approach than what you are proposing.

kamil-tekiela:
@cmb69  Shouldn't the full title be used in the page's title and the abbreviated title only in the table of contents?

mmalferov:
> That would actually be nice.

It seems to me that breadcrumbs are enough for the user to have the opportunity to understand which part of the site he is currently in. And to do this, we just need to make changes to a few CSS rules

> I would say that this page needs a more holistic approach than what you are proposing

You are absolutely right. But even if everything stays the same and the noisy headlines are removed, the page will get better, not worse. And if there is an opportunity to improve the page, why not do it ;) Moreover, the changes have already been made, and it remains to press a couple of buttons for the changes to take effect))

Issue #3939, opened 2024-10-27
Document what the manual considers as a "scalar" type


Girgias:
This comes up quite often but the manual does not have any place where it defines what a scalar type is.

Would integrate note https://www.php.net/manual/en/function.is-scalar.php#55813 and could simplify the `is_scalar()` docs.

ekinhbayar:
Would the [types.intro](https://www.php.net/manual/en/language.types.intro.php) page be a good candidate for mentioning this?

cmb69:
Yes. Maybe we should have two lists of types (scalar vs. non-scalar) there. Could then simplify the `is_scalar()` page to link to types.intro instead of mentioning there which types are scalar, and which are not.

ekinhbayar:
If nobody has started on this, I can take some time to prep a first draft over the weekend.

ekinhbayar:
Browsing around a bit, I also noticed that there's actually a mention of scalar types as part of [types.type-system](https://www.php.net/manual/en/language.types.type-system.php) 🤔

PR #3986, opened 2024-11-03
Document what's considered as a scalar type

ekinhbayar:
Initial take for issue #3939.

It made me go through all of the types documented on the type-system page, and this may be more than what the issue originally asks for. In which case, I can do a simpler attempt. Parts of this may be confusing or simply wrong too, if so, please correct me or let me know. This was a sort of a recap of theory and a categorization exercise which I may have failed : )

### DIFF
--- a/language/types/type-system.xml
+++ b/language/types/type-system.xml
@@ -106,6 +106,28 @@
     </listitem>
    </itemizedlist>
   </para>
+
+  <sect3 xml:id="language.types.type-system.atomic.scalar">
+   <title>Scalar types</title>
+   <simpara>
+    A value is considered scalar if it is of type <type>int</type>,
+    <type>float</type>, <type>string</type> or <type>bool</type>.
+   </simpara>
+  </sect3>
+
+  <sect3 xml:id="language.types.type-system.atomic.user-defined">
+   <title>User-defined types</title>
+   <simpara>
+    It is possible to define custom types with
+    <link linkend="language.oop5.interfaces">interfaces</link>,
+    <link linkend="language.oop5.basic.class">classes</link> and
+    <link linkend="language.types.enumerations">enumerations</link>.
+    These are considered as user-defined types, or class-types.
+    For example, a class called <literal>Elephant</literal> can be defined,
+    then objects of type <literal>Elephant</literal> can be instantiated,
+    and a function can request a parameter of type <literal>Elephant</literal>.
+   </simpara>
+  </sect3>
  </sect2>
 
  <sect2 xml:id="language.types.type-system.composite">
--- a/language/types.xml
+++ b/language/types.xml
@@ -28,6 +28,8 @@
    runtime. However, it is possible to statically type some aspect of the
    language via the use of
    <link linkend="language.types.declarations">type declarations</link>.
+   Different types that are supported by PHP's type system can be found at the
+   <link linkend="language.types.type-system">type system</link> page.
   </para>
 
   <para>
--- a/reference/var/functions/is-scalar.xml
+++ b/reference/var/functions/is-scalar.xml
@@ -18,10 +18,7 @@
    Finds whether the given variable is a scalar.
   </para>
   <para>
-   Scalar variables are those containing an <type>int</type>, 
-   <type>float</type>, <type>string</type> or <type>bool</type>.  
-   Types <type>array</type>, <type>object</type>, <type>resource</type> 
-   and <type>null</type> are not scalar.
+   See <link linkend="language.types.type-system.atomic.scalar">scalar types</link> for more information.
   </para>
   <note>
    <para>


language/types/type-system.xml
```diff
@@ -21,139 +21,192 @@
...
     </simpara>
+    <itemizedlist>
+     <listitem>
+      <simpara><type>false</type></simpara>
+     </listitem>
+     <listitem>
+      <simpara><type>true</type></simpara>
+     </listitem>
+    </itemizedlist>
    </listitem>
```
Girgias, line 88: I'm back to wondering if we should call those `singleton` types, but that is orthogonal to this change.

language/types/type-system.xml
```diff
@@ -21,139 +21,192 @@
...
+   <listitem>
+    <simpara><type>callable</type> type</simpara>
    </listitem>
   </itemizedlist>
 
-  <sect3 xml:id="language.types.type-system.composite.intersection">
-   <title>Intersection types</title>
+  <sect3 xml:id="language.types.type-system.atomic.scalar">
+   <title>Scalar types</title>
    <para>
```
Girgias, line 112: Nit: to follow the new style guide
```suggestion
   <simpara>
```

language/types/type-system.xml
```diff
@@ -21,139 +21,192 @@
...
-    <literal>V</literal> will be written as <literal>T&amp;U&amp;V</literal>.
+    A value is considered scalar if it is of type <type>int</type>,
+    <type>float</type>, <type>string</type> or <type>bool</type>.
+   </para>
+   <para>
+    While <literal>null</literal> value can only be of type <type>null</type>
+    and therefore is scalar, built-in function <function>is_scalar</function>
+    does not consider values of type <type>null</type> to be scalar.
+    To test for <literal>null</literal>, use the <function>is_null</function> function.
    </para>
```
Girgias, line 121: I'm not sure I would mention `null` here, this might still make more sense as a caution on the `is_scalar()` docs.

language/types/type-system.xml
```diff
@@ -21,139 +21,192 @@
...
-    <literal>V</literal> will be written as <literal>T|U|V</literal>.
-    If one of the types is an intersection type, it needs to be bracketed
-    with parenthesis for it to written in <acronym>DNF</acronym>:
-    <literal>T|(X&amp;Y)</literal>.
+    Types <type>never</type> and <type>void</type> can be declared as the
+    return type of a function to indicate whether the function does not
+    terminate or does not return a value. Their usage as a parameter
+    or a property type is not permitted.
    </para>
   </sect3>
```
Girgias, line 133: Does this section really make that much sense? This is described on the respective pages already.

language/types/type-system.xml
```diff
@@ -21,139 +21,192 @@
...
+      <simpara>Return-only types</simpara>
+      <itemizedlist>
+       <listitem>
+        <simpara><type>never</type> type</simpara>
+       </listitem>
+       <listitem>
+        <simpara><type>void</type> type</simpara>
+       </listitem>
+      </itemizedlist>
+     </listitem>
```
Girgias, line 67: I don't know if it is that useful to indicate that those types are return only in the list of types?

language/types/type-system.xml
ekinhbayar: Just to fully confirm, is it on purpose that these "value types" are considered outside of the built-in types?

language/types/type-system.xml
```diff
@@ -2,170 +2,201 @@
...
+ </simpara>
 
  <sect2 xml:id="language.types.type-system.atomic">
   <title>Atomic types</title>
-  <para>
+  <simpara>
    Some atomic types are built-in types which are tightly integrated with the
    language and cannot be reproduced with user defined types.
-  </para>
+  </simpara>
```
Girgias, line 22: I meant to only use simpara for new text, to prevent work and large diff for translations, yes this is no ideal. Sorry for being unclear, if you could partially revert the change.

language/types/type-system.xml
```diff
@@ -2,170 +2,201 @@
...
+   <title>User-defined types</title>
+   <simpara>
+    It is possible to define custom types with
+    <link linkend="language.oop5.interfaces">interfaces</link>,
+    <link linkend="language.oop5.basic.class">classes</link> and
+    <link linkend="language.types.enumerations">enumerations</link>.
+    These are considered as user-defined types, or class-types.
+    For example, a class called <literal>Elephant</literal> can be defined, then objects
+    of type <literal>Elephant</literal> can be instantiated, and a function can request
+    a parameter of type <literal>Elephant</literal>.
```
Girgias, line 123: Semantic line breaks
```suggestion
    For example, a class called <literal>Elephant</literal> can be defined,
    then objects of type <literal>Elephant</literal> can be instantiated,
    and a function can request a parameter of type <literal>Elephant</literal>.
```

language/types/type-system.xml
```diff
@@ -2,170 +2,201 @@
...
+   Values of type <type>array</type>, and <type>object</type> can be composed of multiple
+   atomic and composite elements.
+  </simpara>
+
+  <note>
+   <simpara>
+    The <type>resource</type> type is an opaque type, meaning it's internal structure
+    depends on the implementation. Therefore, it is considered neither scalar nor composite.
+   </simpara>
+  </note>
```
Girgias, line 141: A resource is "like an object, just worse" so not exactly sure why it is mentioned here.

language/types/type-system.xml
```diff
@@ -2,170 +2,201 @@
...
+   </simpara>
+  </sect3>
+ </sect2>
+
+ <sect2 xml:id="language.types.type-system.composite">
+  <title>Composite types</title>
+  <simpara>
+   It is possible to combine multiple atomic types into composite types.
+   Values of type <type>array</type>, and <type>object</type> can be composed of multiple
+   atomic and composite elements.
```
Girgias, line 133: How can an array be composed of multiple atomic and composite types? It seems you are mixing elements/properties of array/objects with types which is incorrect.

A class is a type, an "array-shape" would be a type, what is inside an array/object is "irrelevant" from the type system PoV, and this is not what I was going for when writing this section about composite types, which is about how to combine different types together (e.g. `T|U`, `X&Y` and a type in DNF `array|(Traversable&Countable)`).

language/types/type-system.xml
```diff
@@ -2,170 +2,201 @@
...
+    The <type>resource</type> type is an opaque type, meaning it's internal structure
+    depends on the implementation. Therefore, it is considered neither scalar nor composite.
+   </simpara>
+  </note>
+
+  <sect3 xml:id="language.types.type-system.composite.type-combination">
+   <title>Combining types</title>
+   <simpara>
+    PHP allows types to be combined in the following ways:
+   </simpara>
```
Girgias, line 147: I don't understand why you are adding a sect3 here?

Girgias (commented):
The diff GitHub is providing is... kinda garbage tried my best to comment on the content.

language/types/type-system.xml
Girgias: They are built-in, but they also represent a single *value* of a larger type.

For example, if we had singleton types for integers and a function only accepts a range of integers, we could write it like `function foo(1|2|3|4 $range1to4inclusive)`.

I did put them outside because they felt weird to be with the other built-in types, but they could be moved with them.

language/types/type-system.xml
ekinhbayar: I should've thought about translations, will indeed partially revert this, no worries.

language/types/type-system.xml
ekinhbayar: I am indeed mistaken on couple things here, one being element/property types vs containing more than its value, this is partly because I was reading https://phplang.org/spec/05-types.html as well. It is clear now though, knowing what's referred to as composite is really about unions and intersections, thank you. I'll revert changes to this section entirely.

language/types/type-system.xml
ekinhbayar: I think I was initially coming from the perspective of... tell the type of each listed item here so it's clear what's non-scalar, but it's definitely not necessary.

language/types/type-system.xml
ekinhbayar: I think I've addressed all but this one now. I could leave it as it is, though it may be nice to define it perhaps as simple as you just did : )

language/types/type-system.xml
```diff
@@ -21,91 +21,108 @@
...
+     </listitem>
+     <listitem>
+      <simpara><link linkend="language.types.enumerations">Enumerations</link></simpara>
+     </listitem>
+    </itemizedlist>
+   </listitem>
+   <listitem>
+    <simpara><type>callable</type> type</simpara>
+   </listitem>
+  </itemizedlist>
```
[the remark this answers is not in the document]
ekinhbayar: It sure did, I think the diff looks a lot better now.

ekinhbayar:
hopefully the last ping for you on this @Girgias : ) I was wondering whether the new style guides documented somewhere, the contributing docs link on readme [can't be found](https://github.com/php/doc-base/blob/master/CONTRIBUTING_DOCS.md).

Also, at this stage of the PR, should I be requesting a review again?

Girgias:
> hopefully the last ping for you on this @Girgias : ) I was wondering whether the new style guides documented somewhere, the contributing docs link on readme [can't be found](https://github.com/php/doc-base/blob/master/CONTRIBUTING_DOCS.md).
> 
> Also, at this stage of the PR, should I be requesting a review again?

Style guide is on the [doc.php.net](https://doc.php.net/guide/style.md) and has always been, but might be good to cross-link better.

No need to ask for rereviews as I get the email, just do it if I forget for a while. :)

ekinhbayar:
Thanks for your time!

PR #3987, opened 2024-11-03
Add IDs to GLOB constants

haszi:
The second commit alphabetically sorts the constants so this PR is easier to review commit by commit.

### DIFF
--- a/reference/filesystem/constants.xml
+++ b/reference/filesystem/constants.xml
@@ -82,8 +82,8 @@
    </listitem>
   </varlistentry>
  </variablelist>
- <variablelist xml:id="constant.glob-constant-variablelist">
-  <varlistentry>
+ <variablelist role="constant_list">
+  <varlistentry xml:id="constant.glob-brace">
    <term>
     <constant>GLOB_BRACE</constant>
      (<type>int</type>)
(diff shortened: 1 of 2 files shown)


reference/filesystem/constants.xml
```diff
@@ -83,6 +83,21 @@
...
+    <constant>GLOB_AVAILABLE_FLAGS</constant>
+     (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     All <constant>GLOB_<replaceable>*</replaceable></constant> flags combined.
+     Equivalent to <literal>0</literal> | <constant>GLOB_BRACE</constant> |
+     <constant>GLOB_MARK</constant> | <constant>GLOB_NOSORT</constant> |
+     <constant>GLOB_NOCHECK</constant> | <constant>GLOB_NOESCAPE</constant> |
+     <constant>GLOB_ERR</constant> | <constant>GLOB_ONLYDIR</constant>
```
Girgias, line 97: Follow-up/PhD quest: Maybe this should be a `<code>` tag? But I don't think it supports rendering tags within of it?

Girgias (approved):
Thank you!

reference/filesystem/constants.xml
haszi: Makes sense.

What doesn't make sense is that `<constant>` tags cannot be children of `<code>` tags in [DocBook 5.2](https://tdg.docbook.org/tdg/5.2/code#element.db.code_refsection1_refsection4). :-(

Girgias: I've opened [an issue on the DocBook repo](https://github.com/docbook/docbook/issues/265) (until their mailing lists are fixed)

PR #3988, opened 2024-11-03
new-features.xml Change `ucfirst` to `mb_ucfirst`

mmalferov:
Can we replace `ucfirst` with `mb_ucfirst` for the local first name when translation?

### DIFF
--- a/appendices/migration84/new-features.xml
+++ b/appendices/migration84/new-features.xml
@@ -31,7 +31,7 @@ class Person
     // All write operations go though this hook, and the result is what is written.
     // Read access happens normally.
     public string $firstName {
-        set => ucfirst(strtolower($value));
+        set => mb_ucfirst(strtolower($value));
     }
 
     // All write operations go through this hook, which has to write to the backing value itself.


TimWolla (commented):
Perhaps this example should be entirely replaced by the one in https://github.com/php/web-php/pull/1079#pullrequestreview-2389016000, because of the reasons given there?

Girgias:
Yeah, using the new example might be a good idea.

kamil-tekiela:
Let's use the other example. If you'd like, please create a new PR that copies the example over to this page. 

As an aside, using `mb_ucfirst` on a byte string from `strtolower` is incorrect.

mmalferov:
> As an aside, using `mb_ucfirst` on a byte string from `strtolower` is incorrect.

Could you explain why and give an example, please

kamil-tekiela:
> > As an aside, using `mb_ucfirst` on a byte string from `strtolower` is incorrect.
> 
> Could you explain why and give an example, please

Well, it simply doesn't make much sense. If the string is multibyte then use `mb_strtolower`, else use `ucfirst`

mmalferov:
In the Russian documentation, it was agreed to translate the text in the code examples into the native language that is to be translated (if the translation does not break the code example). Therefore, we wanted to translate the name `Peter` as  `«Пётр»`

```php
<?php

$firstName = 'peter';
$cyrillicName = 'пётр';

// Converts all ASCII chars to lowercase, then capitalized the first ASCII char
echo ucfirst(strtolower($firstName)), "\n"; //  Prints 'Peter'

// The same, but the first character 'п' (cyrillic 'pe') is 2 bytes length, therefore  ucfirst() will ignore it
echo ucfirst(strtolower($cyrillicName)), "\n"; // Prints 'пётр' (note, the first char is in lowercase) - it is bad


// Whereas mb_ucfirst() capitalizes the first 2-byte alphabetic character 'п' (pe)
echo mb_ucfirst(strtolower($cyrillicName)), "\n"; // Prints 'Пётр' - well good!
```

The mb_ucfirst() did not break the example with the English name, and it made sense to convert the first letter of the Cyrillic name even when working with the strtolower(). For the rest, I agree—either we work with multibyte characters, or with ASCII ;)

PR #3994, opened 2024-11-03
session: Remove session ID-format related INIs from session/security.xml

TimWolla:
These are deprecated with PHP 8.4, because the defaults are fine.

see #3993


Girgias (approved):
MSTM, thank you!

PR #3983, opened 2024-11-01
[PHP 8.4] Add manual for `PDO::PGSQL_ATTR_RESULT_MEMORY_SIZE`

KentarouTakeda:
Ref: php/php-src#14260

PR #3997, opened 2024-11-04
fix(pdo_pgsql/constants): Correct use of `null` values

KentarouTakeda:
fixes: https://github.com/php/doc-en/pull/3983#discussion_r1827141548


Girgias (approved):
Thank you!

PR #4000, opened 2024-11-04
document potential Windows deadlock

divinity76:
on Windows php's require() and "php file.php" reacts differently to the LOCK_EX situation, require() will fail immediately (like a LOCK_NB) while cli "php file.php" will wait/hang until unlocked. Make that clear in the documentation.

### DIFF
--- a/reference/filesystem/functions/flock.xml
+++ b/reference/filesystem/functions/flock.xml
@@ -189,11 +189,26 @@ fclose($fp);
     <literal>FAT</literal> and its derivates and will therefore always
     return &false; under these environments.
    </para>
+   <para>
+    On Windows, if php tries to execute a LOCK_EX'ed php source code file, it will hang until the file is unlocked, meaning
+<![CDATA[
+<?php
+if($argv[1] ?? null === 'child') {die();}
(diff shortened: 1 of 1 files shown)


lacatoire (approved):

lacatoire:
LGTM - Documents a real Windows-specific gotcha. The scenario (script locking itself then spawning child) is realistic and the deadlock behavior is important to warn developers about.

reference/filesystem/functions/flock.xml
```diff
@@ -188,12 +188,27 @@ fclose($fp);
...
+   </para>
   </warning>
   <note>
    <para>
     On Windows, if the locking process opens the file a second time, it cannot
     access the file through this second handle until it unlocks the file.
    </para>
+   <para>
+    On Windows, if you try to require a LOCK_EX'ed file, it will fail with ErrorException: require(): Read of X bytes failed with errno=13 Permission denied
+   </para>
```
lacatoire:
```suggestion
   <simpara>
    On Windows, if you try to require a LOCK_EX'ed file, it will fail with ErrorException: require(): Read of X bytes failed with errno=13 Permission denied
   </simpara>
```

PR #4005, opened 2024-11-05
Group related filesystems constants together

Girgias:
And add title to indicate with what functions they are used.

Also stop using tables, as I don't think it brings much value in these cases.

### DIFF
--- a/reference/filesystem/constants.xml
+++ b/reference/filesystem/constants.xml
@@ -82,167 +82,13 @@
    </listitem>
   </varlistentry>
  </variablelist>
- <variablelist role="constant_list">
-  <varlistentry xml:id="constant.glob-available-flags">
-   <term>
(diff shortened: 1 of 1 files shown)


reference/filesystem/constants.xml
```diff
@@ -82,11 +82,114 @@
    </listitem>
   </varlistentry>
  </variablelist>
- <variablelist role="constant_list">
```
cmb69, line 85: Maybe we should special PhD support for `role="contant_list"` alltogether?

cmb69 (approved):
Thank you!

reference/filesystem/constants.xml
Girgias: I think you've dropped a word while writing the sentence!

PR #4010, opened 2024-11-05
Add MySQLi deprecations to `versions.xml`

TimWolla:
see php/doc-en#3958

### DIFF
--- a/reference/mysqli/versions.xml
+++ b/reference/mysqli/versions.xml
@@ -55,14 +55,14 @@
  <function name="mysqli_info" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_init" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_insert_id" from="PHP 5, PHP 7, PHP 8"/>
- <function name="mysqli_kill" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_kill" from="PHP 5, PHP 7, PHP 8" deprecated="PHP 8.4.0"/>
  <function name="mysqli_more_results" from="PHP 5, PHP 7, PHP 8"/>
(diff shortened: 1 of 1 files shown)


reference/mysqli/versions.xml
```diff
@@ -55,22 +55,22 @@
  <function name="mysqli_info" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_init" from="PHP 5, PHP 7, PHP 8"/>
  <function name="mysqli_insert_id" from="PHP 5, PHP 7, PHP 8"/>
- <function name="mysqli_kill" from="PHP 5, PHP 7, PHP 8"/>
+ <function name="mysqli_kill" from="PHP 5, PHP 7, PHP 8" deprecated="PHP 8.4"/>
```
kamil-tekiela, line 58:
```suggestion
 <function name="mysqli_kill" from="PHP 5, PHP 7, PHP 8" deprecated="PHP 8.4.0"/>
```
I think we specify the full version everywhere.

kamil-tekiela (changes requested):
Thanks, I forgot this exists. Could you please mark all 6 of them as deprecated?

reference/mysqli/versions.xml
TimWolla: utf8_(de|en)code use the short variant: https://github.com/php/doc-en/blob/0f1d7fba407bd1385616508a1e42e31fe009bff2/reference/strings/versions.xml#L106-L107, but given that the introductory version is named in full this makes sense to me to adjust.

TimWolla:
> Could you please mark all 6 of them as deprecated?

Bah. This duplication between procedural and OO API is terrible. Now adjusted.

kamil-tekiela (approved):

PR #4013, opened 2024-11-06
is-scalar.xml Give a more precise formulation

mmalferov:
The function 'is_scalar` can check the scalarity of not only variables, as the manual claims, but also other expressions

```
<?php

var_dump(
	is_scalar(1),
	is_scalar([]),
	is_scalar(new stdClass()),
	is_scalar(list(0 => $name) = ['Bob']),
	is_scalar(1 > 2),
        // etc.
);
```


Girgias (approved):
Thank you!

PR #4017, opened 2024-11-06
Add a note about time zone changes in PHP 8.2

BackEndTea:
This MR documents the change made in: https://github.com/php/php-src/commit/6770158d474fa442ce55bba4ec32dd2703828b33

As mentioned in the issue: https://github.com/php/php-src/issues/11496

Since PHP 8.2, the fallback of the timezone is now UTC, rather than the timezone supplied by the system.

### DIFF
--- a/appendices/migration82/other-changes.xml
+++ b/appendices/migration82/other-changes.xml
@@ -125,10 +125,10 @@
     The properties of <classname>DatePeriod</classname> are now properly declared.
    </para>
 
-   <para>
-    Default timezones will fall back to UTC, rather than system timezone if 
+   <simpara>
+    Default timezones will fall back to UTC, rather than system timezone if
     none are defined in the <link linkend="ini.date.timezone">date.timezone</link> ini setting.
-   </para>
+   </simpara>
   </sect3>
 
   <sect3 xml:id="migration82.other-changes.extensions.intl">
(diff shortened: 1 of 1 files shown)


appendices/migration82/other-changes.xml
```diff
@@ -124,6 +124,11 @@
    <para>
     The properties of <classname>DatePeriod</classname> are now properly declared.
    </para>
+
+   <para>
+    Default timezones will fall back to UTC, rather than system timezone if 
+    none are defined in the <link linkend="ini.date.timezone">date.timezone</link> ini setting.
+   </para>
```
Girgias, line 131:
```suggestion
   <simpara>
    Default timezones will fall back to UTC, rather than system timezone if
    none are defined in the <link linkend="ini.date.timezone">date.timezone</link> ini setting.
   </simpara>
```

appendices/migration82/other-changes.xml
```diff
@@ -124,6 +124,11 @@
    <para>
     The properties of <classname>DatePeriod</classname> are now properly declared.
    </para>
+
+   <simpara>
+    Default timezones will fall back to UTC, rather than system timezone if
+    none are defined in the <link linkend="ini.date.timezone">date.timezone</link> ini setting.
+   </simpara>
```
derickr, line 131: That is not what happened in the code linked to. It was only a change in warning. PHP never really used direct system timezone. In the far past, it tried to guess which timezone the system was by looking at UTC offset, the DST flag, and the abbreviation. The warning for that was finally removed in PHP 7.0.

derickr (changes requested):

PR #4024, opened 2024-11-09
fix(language/oop5/final): indent level

KentarouTakeda:
Sorry for the small change.

Would you mind fixing the inconsistent tag nesting and indentation levels?

### DIFF
--- a/language/oop5/final.xml
+++ b/language/oop5/final.xml
@@ -97,17 +97,17 @@ class Bar extends Foo
   </programlisting>
  </example>
 
-  <note>
-   <simpara>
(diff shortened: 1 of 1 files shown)


Girgias:
This is a rather recent change so shouldn't be too bad on translations, but I'd rather not in general. Maybe if rev-check would automatically skip and ignore whitespace changes those sorts would be more reasonable (also see https://github.com/php/web-doc/issues/57)

KentarouTakeda:
Thank you for your detailed explanation. I think it makes sense. I was a bit hesitant too.

Next time, I'll take the update date and other factors into consideration when making suggestions.

alfsb:
> Sorry for the small change.

That what `[skip-revcheck]` is for. But I see it is not documented at all on https://github.com/php/doc-base/tree/master/docs. I will take this later.

PR #4016, opened 2024-11-06
newlazy*.xml Fix the typos

mmalferov:

PR #4026, opened 2024-11-09
[PHP 8.4] fix(reflectionclass/newlazyghost): Confusion between proxy and ghost

KentarouTakeda:
It seems that the description of `newLazyGhost()` is the same as `newLazyProxy()`.


arnaud-lb:
@KentarouTakeda thank you for the patch, but I believe that this is a duplicate of https://github.com/php/doc-en/pull/4016.

KentarouTakeda:
I have confirmed that it has been fixed by #4016 .

PR #4035, opened 2024-11-11
PHP-8.4: Document fpow()

Girgias:
~~Waiting on https://github.com/php/php-src/pull/16751~~

### DIFF
--- a/reference/math/functions/fpow.xml
+++ b/reference/math/functions/fpow.xml
@@ -14,8 +14,8 @@
   </methodsynopsis>
   <simpara>
    Returns the floating point result of raising
-   <parameter>num1</parameter> to the power of <parameter>exponent</parameter>.
-   If <parameter>num1</parameter> is zero and <parameter>exponent</parameter>
+   <parameter>num</parameter> to the power of <parameter>exponent</parameter>.
+   If <parameter>num</parameter> is zero and <parameter>exponent</parameter>
    is less than zero, then <constant>INF</constant> is returned.
   </simpara>
  </refsect1>
@@ -46,7 +46,7 @@
   &reftitle.returnvalues;
   <simpara>
    Returns a <type>float</type> corresponding to
-   <code><parameter>num</parameter><superscript>$exponent</superscript></code>.
+   <code>$num<superscript>$exponent</superscript></code>.
   </simpara>
  </refsect1>
 
--- a/reference/math/functions/fpow.xml
+++ b/reference/math/functions/fpow.xml
@@ -27,7 +27,7 @@
     <term><parameter>num</parameter></term>
     <listitem>
      <simpara>
-      The base to number.
+      The base to use.
      </simpara>
     </listitem>
    </varlistentry>
--- a/reference/math/functions/fpow.xml
+++ b/reference/math/functions/fpow.xml
@@ -57,6 +57,7 @@
    <programlisting role="php">
 <![CDATA[
 <?php
+var_dump(fpow(10, 2));
 var_dump(fpow(0, -3));
 var_dump(fpow(-1, 5.5));
 ?>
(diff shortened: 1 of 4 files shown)


reference/math/functions/fpow.xml
```diff
@@ -0,0 +1,108 @@
...
+  &reftitle.description;
+  <methodsynopsis>
+   <type>float</type><methodname>fpow</methodname>
+   <methodparam><type>float</type><parameter>num</parameter></methodparam>
+   <methodparam><type>float</type><parameter>exponent</parameter></methodparam>
+  </methodsynopsis>
+  <simpara>
+   Returns the floating point result of raising
+   <parameter>num1</parameter> to the power of <parameter>exponent</parameter>.
+   If <parameter>num1</parameter> is zero and <parameter>exponent</parameter>
```
cmb69, line 18:
```suggestion
   <parameter>num</parameter> to the power of <parameter>exponent</parameter>.
   If <parameter>num</parameter> is zero and <parameter>exponent</parameter>
```

reference/math/functions/fpow.xml
```diff
@@ -0,0 +1,108 @@
...
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>num</parameter></term>
+    <listitem>
+     <simpara>
+      The base to number.
```
cmb69, line 30: Is that proper English? Do you mean "The base number."?

reference/math/functions/fpow.xml
```diff
@@ -0,0 +1,108 @@
...
+    </listitem>
+   </varlistentry>
+  </variablelist>
+ </refsect1>
+
+ <refsect1 role="returnvalues">
+  &reftitle.returnvalues;
+  <simpara>
+   Returns a <type>float</type> corresponding to
+   <code><parameter>num</parameter><superscript>$exponent</superscript></code>.
```
cmb69, line 49: Hmm, that looks inconsistent. Either both should be `parameter`, or neither. I think the latter makes more sense, so:
```suggestion
   <code>$num<superscript>$exponent</superscript></code>.
```

reference/math/functions/fpow.xml
```diff
@@ -0,0 +1,108 @@
...
+ </refsect1>
+
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <example xml:id="function.fpow.example.basic">
+   <title><function>fpow</function> example</title>
+   <programlisting role="php">
+<![CDATA[
+<?php
+var_dump(fpow(0, -3));
```
cmb69, line 60: Maybe add a "normal" example (something like `fpow(10, 2)`).

reference/math/functions/fpow.xml
```diff
@@ -0,0 +1,108 @@
...
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <simplelist>
+   <member><link linkend="language.operators.arithmetic"><literal>/</literal></link> - Floating-point division</member>
+   <member><function>pow</function></member>
+   <member><function>fdiv</function></member>
+   <member><function>fmod</function></member>
+   <!-- TODO ** operator link?
```
cmb69, line 82: Maybe. On the other hand, we already link to `pow` what might be good enough.

cmb69 (commented):
A couple of nits, but basically looks good. Thank you!

reference/math/functions/fpow.xml
Girgias: Yes I don't know what I did this here

reference/math/functions/fpow.xml
Girgias: Parameter is not allowed within `<superscript>` so that's why I did that, but happy to change.

cmb69 (approved):
Thanks!

PR #4037, opened 2024-11-11
PHP 8.4: Document reflection changes

Girgias:
Drive-by fix for 8.1 ommision

### DIFF
--- a/reference/reflection/reflectionclass/resetaslazyghost.xml
+++ b/reference/reflection/reflectionclass/resetaslazyghost.xml
@@ -10,7 +10,7 @@
   <methodsynopsis role="ReflectionClass">
    <modifier>public</modifier> <type>void</type><methodname>ReflectionClass::resetAsLazyGhost</methodname>
    <methodparam><type>object</type><parameter>object</parameter></methodparam>
-   <methodparam><type>callable</type><parameter>initializer</parameter></methodparam>
+   <methodparam><type>callable</type><parameter>factory</parameter></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>options</parameter><initializer>0</initializer></methodparam>
   </methodsynopsis>
   <simpara>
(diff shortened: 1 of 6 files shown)


reference/reflection/reflectionclass/resetaslazyghost.xml
```diff
@@ -10,7 +10,7 @@
   <methodsynopsis role="ReflectionClass">
    <modifier>public</modifier> <type>void</type><methodname>ReflectionClass::resetAsLazyGhost</methodname>
    <methodparam><type>object</type><parameter>object</parameter></methodparam>
-   <methodparam><type>callable</type><parameter>initializer</parameter></methodparam>
```
arnaud-lb, line 13: This one is a bug in the stub, unfortunately :( The parameter should have been called "initializer".

arnaud-lb (approved):
Thank you!

reference/reflection/reflectionclass/resetaslazyghost.xml
nielsdos: 8.4.0 is not released yet, why wait for 8.4.1?

Crell (approved):

reference/reflection/reflectionclass/resetaslazyghost.xml
arnaud-lb: I would like to, but can we still make changes to 8.4.0? As I understand, the current RC will be released as 8.4.0? Or am I mistaken?

arnaud-lb: Oh, that's a good news. Thank you

reference/reflection/reflectionproperty.xml
```diff
@@ -176,6 +206,55 @@
...
+     <varlistentry xml:id="reflectionproperty.constants.is-final">
+      <term><constant>ReflectionProperty::IS_FINAL</constant></term>
+      <listitem>
+       <simpara>
+        Indicates the property is
+        <link linkend="language.oop5.final">final</link>.
+        Available as of PHP 8.1.0.
+       </simpara>
+      </listitem>
+     </varlistentry>
```
Girgias, line 256: Actually I think I misunderstood this, as this is part of property hooks right @Crell @iluuu1994 ?

iluuu1994: Yes. Properties may not be final before 8.4.

PR #4047, opened 2024-11-12
Document some SPL changes in PHP-8.4

Girgias:


reference/spl/splobjectstorage.xml
```diff
@@ -33,7 +33,7 @@
     </oointerface>
 
     <oointerface>
-     <interfacename>Iterator</interfacename>
+     <interfacename>SeekableIterator</interfacename>
```
cmb69, line 36: Don't we need a changelog entry for this?

cmb69 (approved):
Looks good (one possibly issue below). Thank you!

reference/spl/splobjectstorage.xml
Girgias: Ah yes we do indeed.

cmb69 (approved):
Thank you!

PR #4048, opened 2024-11-12
PHP 8.4: Document SplObjectStorage::seek()

Girgias:
@nielsdos could you document this as, IIRC, you added this?

Related (#4047)

### DIFF
--- a/reference/spl/splobjectstorage/seek.xml
+++ b/reference/spl/splobjectstorage/seek.xml
@@ -19,18 +19,16 @@
 
  <refsect1 role="parameters">
   &reftitle.parameters;
-  <para>
-   <variablelist>
-    <varlistentry>
-     <term><parameter>offset</parameter></term>
-     <listitem>
-      <simpara>
-       The position to seek to.
-      </simpara>
-     </listitem>
-    </varlistentry>
-   </variablelist>
-  </para>
+  <variablelist>
+   <varlistentry>
+    <term><parameter>offset</parameter></term>
+    <listitem>
+     <simpara>
+      The position to seek to.
+     </simpara>
+    </listitem>
+   </varlistentry>
+  </variablelist>
  </refsect1>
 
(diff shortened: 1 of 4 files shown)


nielsdos:
Pushed commits, including 2 drive-by fixes:
1) one simpara fix because I copied that section from that file and noticed it
2) Documenting ArrayIterator

reference/spl/splobjectstorage/seek.xml
```diff
@@ -11,26 +11,26 @@
...
      <listitem>
       <simpara>
-       Description.
+       The position to seek to.
       </simpara>
      </listitem>
     </varlistentry>
    </variablelist>
-  </simpara>
+  </para>
```
Girgias: Ideally this wrapping para should be removed, but I'm fine to leave the `<para>` tag if it is pre-exiting

reference/spl/splobjectstorage/seek.xml
```diff
@@ -0,0 +1,120 @@
...
+    <varlistentry>
+     <term><parameter>offset</parameter></term>
+     <listitem>
+      <simpara>
+       The position to seek to.
+      </simpara>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </para>
```
Girgias, line 33:
```suggestion
  <variablelist>
   <varlistentry>
    <term><parameter>offset</parameter></term>
    <listitem>
     <simpara>
      The position to seek to.
     </simpara>
    </listitem>
   </varlistentry>
  </variablelist>
```

reference/spl/splobjectstorage/seek.xml
nielsdos: The simpara came out of the stub generator, but that was wrong so I changed it back to para: https://github.com/php/php-src/pull/16764

nielsdos:
I think this could be merged, no?

PR #4056, opened 2024-11-12
Use modern markup for E_* constants

Girgias:
- Indicate that E_STRICT is deprecated as of PHP 8.4.0
- explain the case where E_RECOVERABLE_ERROR can happen
- E_USER_ERROR should not be used any more.
- Rewording which removes personalization


reference/errorfunc/constants.xml
```diff
@@ -3,258 +3,338 @@
...
+    </simpara>
+    <simpara>
+     Value of the constant: <literal>4096</literal>
+    </simpara>
+    <note>
+     <simpara>
+      This error level is effectively unused,
+      the only case where this can happen is when interpreting an
+      &object; as a <type>bool</type> fails.
+      Which can only happen for internal objects.
```
cmb69, line 311: Not sure about the full stop in the first line. Maybe
```suggestion
      &object; as a <type>bool</type> fails,
      which can only happen for internal objects.
```
or
```suggestion
      &object; as a <type>bool</type> fails.
      This can only happen for internal objects.
```

cmb69 (approved):
Thank you! Looks good (two nits below).

PR #4062, opened 2024-11-13
functions.xml Eliminate confusion in fn's parameters and arguments

mmalferov:
If we agree that a) parameters are part of the function signature, i.e. an integral part of the function declaration, and b) arguments are information that comes to the function from the outside, then we must eliminate confusion in the text. To simplify it: the arguments are passed, and the parameters take values from the arguments and, after argument's expression are evaluated, set values for their parameters. If so, there seem to be inconsistent places in the text where parameters are passed and functions take parameters rather than arguments or something like that

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -488,7 +488,7 @@ function bar(?A $a, $b) {}       // Recommended
     </note>
     <note>
      <simpara>
-      Parameters that are expect the argument by reference may have a default value.
+      Parameters that expect the argument by reference may have a default value.
      </simpara>
     </note>
    </sect2>
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -211,7 +211,7 @@ function takes_array($input)
     long or contains long variable names, making it convenient to list parameters vertically.
    </para>
    <example>
-    <title>Function Parameter List with trailing Comma</title>
+    <title>Function parameter list with trailing comma</title>
     <programlisting role="php">
 <![CDATA[
 <?php
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -173,7 +173,7 @@ function recursion($a)
    <title>Function parameters and arguments</title>
  
    <simpara>
-    The function parameters are declared in the function signature. 
+    The function parameters are declared in the function signature.
     Information may be passed to functions via the argument list,
     which is a comma-delimited list of expressions. The arguments are
     evaluated from left to right and the result is assigned to the parameters of
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -271,7 +271,7 @@ echo $str;    // outputs 'This is a string, and something extra.'
     <para>
      A function may define default values for parameters using syntax similar
      to assigning a variable. The default is used only when the parameter's argument is
-     not passed; in particular, note that passing &null; does <emphasis>not</emphasis>
+     not passed. Note that passing &null; does <emphasis>not</emphasis>
      assign the default value.
     </para>
     <para>
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -5,7 +5,7 @@
 
   <sect1 xml:id="functions.user-defined">
    <title>User-defined functions</title>
- 
+
    <para>
     A function may be defined using syntax such as the following:
    </para>
@@ -168,10 +168,10 @@ function recursion($a)
    </para>
 
   </sect1>
- 
+
   <sect1 xml:id="functions.arguments">
    <title>Function parameters and arguments</title>
- 
+
    <simpara>
     The function parameters are declared in the function signature.
     Information may be passed to functions via the argument list,
@@ -232,7 +232,7 @@ function takes_many_args(
 
    <sect2 xml:id="functions.arguments.by-reference">
     <title>Passing arguments by reference</title>
- 
+
     <simpara>
      By default, function arguments are passed by value (so that if
      the value of the argument within the function is changed, it does
@@ -267,7 +267,7 @@ echo $str;    // outputs 'This is a string, and something extra.'
    </sect2>
    <sect2 xml:id="functions.arguments.default">
     <title>Default parameter values</title>
- 
+
     <para>
      A function may define default values for parameters using syntax similar
      to assigning a variable. The default is used only when the parameter's argument is
@@ -383,7 +383,7 @@ function makeyogurt($container = "bowl", $flavour)
 {
     return "Making a $container of $flavour yogurt.\n";
 }
- 
+
 echo makeyogurt("raspberry"); // "raspberry" is $container, not $flavour
 ?>
 ]]>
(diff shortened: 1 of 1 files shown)


mmalferov:
If the current proposal does not contain errors, can we at least accept them and make global changes later? I have already made changes to the Russian version and received criticism from colleagues for the fact that the translation text deviates from the original. But I can't leave the literal translation either, because the current original confuses the reader by calling arguments parameters and vice versa. I would like the function parameters to remain parameters, and the arguments that the function receives from the outside to remain arguments.

language/functions.xml
```diff
@@ -204,12 +206,12 @@ function takes_array($input)
...
-    As of PHP 8.0.0, the list of function arguments may include a trailing comma, which
-    will be ignored.  That is particularly useful in cases where the list of arguments is
-    long or contains long variable names, making it convenient to list arguments vertically.
+    As of PHP 8.0.0, the list of function parameters may include a trailing comma, which
+    will be ignored.  That is particularly useful in cases where the list of parameters is
+    long or contains long variable names, making it convenient to list parameters vertically.
    </para>
    <example>
-    <title>Function Argument List with trailing Comma</title>
+    <title>Function Parameter List with trailing Comma</title>
```
kamil-tekiela, line 214:
```suggestion
    <title>Function parameter list with trailing comma</title>
```
The title case is very inconsistent here. The example above doesn't use title case, so please make it consistent with it.

language/functions.xml
```diff
@@ -204,12 +206,12 @@ function takes_array($input)
     </example>
    </para>
    <para>
-    As of PHP 8.0.0, the list of function arguments may include a trailing comma, which
-    will be ignored.  That is particularly useful in cases where the list of arguments is
-    long or contains long variable names, making it convenient to list arguments vertically.
+    As of PHP 8.0.0, the list of function parameters may include a trailing comma, which
```
kamil-tekiela, line 209: Doesn't this apply to argument list also?

language/functions.xml
```diff
@@ -260,16 +262,16 @@ echo $str;    // outputs 'This is a string, and something extra.'
...
- 
+    <title>Default parameter values</title>
+
     <para>
-     A function may define default values for arguments using syntax similar
-     to assigning a variable. The default is used only when the parameter is
-     not specified; in particular, note that passing &null; does <emphasis>not</emphasis>
+     A function may define default values for parameters using syntax similar
+     to assigning a variable. The default is used only when the parameter's argument is
+     not passed; in particular, note that passing &null; does <emphasis>not</emphasis>
```
kamil-tekiela, line 274:
```suggestion
     not passed. Note that passing &null; does <emphasis>not</emphasis>
```

language/functions.xml
```diff
@@ -486,7 +488,7 @@ function bar(?A $a, $b) {}       // Recommended
     </note>
     <note>
      <simpara>
-      Arguments that are passed by reference may have a default value.
+      Parameters that are expect the argument by reference may have a default value.
```
kamil-tekiela, line 491:
```suggestion
      Parameters that expect the argument by reference may have a default value.
```

language/functions.xml
```diff
@@ -497,11 +499,11 @@ function bar(?A $a, $b) {}       // Recommended
     <simpara>
      PHP has support for variable-length argument lists in
      user-defined functions by using the
-     <literal>...</literal> token.
+     <literal>...</literal> token also called a splat or spread operator in other languages.
```
kamil-tekiela, line 502: I'd leave it out of this PR.

language/functions.xml
```diff
@@ -564,7 +566,7 @@ echo add(...$a);
     </para>
 
     <para>
-     You may specify normal positional arguments before the
+     You may specify normal positional parameters before the
      <literal>...</literal> token. In this case, only the trailing arguments
      that don't match a positional argument will be added to the array
```
kamil-tekiela, line 571: Is this correct? I am actually confused about this whole sentence.

language/functions.xml
```diff
@@ -701,11 +703,11 @@ htmlspecialchars($string, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401, 'UTF-8', fa
     </example>
 
     <para>
-     Passing the same parameter multiple times results in an Error exception.
+     Passing the same argument multiple times results in an Error exception.
```
kamil-tekiela, line 706:
```suggestion
     Passing an argument to the same named parameter multiple times results in an
     <classname>Error</classname> exception.
```

language/functions.xml
```diff
@@ -701,11 +703,11 @@ htmlspecialchars($string, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401, 'UTF-8', fa
...
     </para>
 
     <example>
-     <title>Error thrown when passing the same parameter multiple times</title>
+     <title>Error thrown when passing the same argument multiple times</title>
```
kamil-tekiela, line 710: Same as above

kamil-tekiela (commented):
In general, this is a good change. I think we should merge it after few adjustments. 

These terms matter mostly when there's ambiguity about what is discussed. In a lot of cases, you could use either term and it would still make sense.

language/functions.xml
mmalferov: This apply to argument list also, but in the example we see only list of parameters, but not the function call with aruments and trailing comma ;)

And from the context of the following sentence,

> That is particularly useful in cases where the list of arguments is
    long or contains long variable names, making it convenient to list arguments vertically.

it is clear that we are talking about parameters — more precisely, about long parameter names in the function signature.

Perhaps the permissibility of passing arguments with an trailing comma should also be mentioned and an example of calling a function with a list of arguments and an trailing comma should be given.

kamil-tekiela: > Perhaps the permissibility of passing arguments with an trailing comma should also be mentioned and an example of calling a function with a list of arguments and an trailing comma should be given.

Perhaps you can do it in the next PR.

language/functions.xml
mmalferov: I understand it this way:

function someFn($one, $two, ...$other) {} // OK
function someFn(...$other, $one, $two) {} // Wrong

For example, after such a call `someFn(1)` the `$other` will stay an empty. But when the calling `someFn(1, 2, 3, 4, 5)`, the trailing arguments (i.e. 3, 4, and 5) that don't match a positional argument (i.e. 1 and 2) will be added to the array ($other).

The paragraph looks like the correct text

kamil-tekiela: Right, this paragraph needs to be rewritten for clarity, but not in this PR. Maybe I will submit one PR later.

kamil-tekiela (approved):
Thanks. It looks good to me and a definite improvement.

PR #4063, opened 2024-11-13
PHP 8.4: trigger_error() updates

Girgias:
Turns out the return type is already correct.

I have no idea for a "good" example, I was going to go for something to use `E_USER_DEPRECATED` but realised the new `Deprecated` attribute is better suited for this.

Maybe @TimWolla has opinions on how to format it for a See Also section?

### DIFF
--- a/reference/errorfunc/functions/trigger-error.xml
+++ b/reference/errorfunc/functions/trigger-error.xml
@@ -46,6 +46,12 @@
        The designated error type for this error. It only works with the <constant>E_USER_<replaceable>*</replaceable></constant>
        family of constants, and will default to <constant>E_USER_NOTICE</constant>.
       </para>
+      <warning>
+       Passing <constant>E_USER_ERROR</constant> as the
+       <parameter>error_level</parameter> is now deprecated.
+       Throw an <exceptionname>Exception</exceptionname> or
+       call <function>exit</function> instead.
+      </warning>
      </listitem>
     </varlistentry>
    </variablelist>
@@ -81,6 +87,22 @@
     </thead>
     <tbody>
      <row>
+      <entry>8.4.0</entry>
+      <entry>
+       Passing <constant>E_USER_ERROR</constant> as the
+       <parameter>error_level</parameter> is now deprecated.
+       Throw an <exceptionname>Exception</exceptionname> or
+       call <function>exit</function> instead.
+      </entry>
+     </row>
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       The function now has a return type of <type>true</type>
+       instead of <type>bool</type>.
+      </entry>
+     </row>
+     <row>
       <entry>8.0.0</entry>
       <entry>
        The function now throws a <classname>ValueError</classname> if an invalid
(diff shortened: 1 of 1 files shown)


reference/errorfunc/functions/trigger-error.xml
```diff
@@ -103,8 +125,8 @@
     <programlisting role="php">
 <![CDATA[
 <?php
-if ($divisor == 0) {
-    trigger_error("Cannot divide by zero", E_USER_ERROR);
+if (is_nan($divisor)) {
+    trigger_error("Cannot divide by NAN", E_USER_WARNING);
```
cmb69, line 129: The message makes not much sense, given that `E_USER_WARNING` would not abort, so the division will be carried out. Maybe:
```suggestion
    trigger_error("Division by NAN", E_USER_WARNING);
```

reference/errorfunc/functions/trigger-error.xml
```diff
@@ -132,6 +154,7 @@ if ($divisor == 0) {
     <member><function>set_error_handler</function></member>
     <member><function>restore_error_handler</function></member>
     <member>The <link linkend="errorfunc.constants">error level constants</link></member>
+    <member>The <classname>Deprecated</classname> attribute</member>
```
cmb69, line 157: IMO, okay as is.

cmb69 (commented):
Thank you!

reference/errorfunc/functions/trigger-error.xml
```diff
@@ -46,6 +46,12 @@
        The designated error type for this error. It only works with the <constant>E_USER_<replaceable>*</replaceable></constant>
        family of constants, and will default to <constant>E_USER_NOTICE</constant>.
       </para>
+      <warning>
+       Passing <constant>E_USER_ERROR</constant> as the
+       <parameter>error_level</parameter> is now deprecated.
+       Throw an <exceptionname>Exception</exceptionname> or
+       call <function>exit</function> instead.
+      </warning>
```
TimWolla, line 54: Not sure if this is necessary to mention: It's already in the changelog and users will notice when they attempt to do this.

reference/errorfunc/functions/trigger-error.xml
TimWolla: Agreed.

reference/errorfunc/functions/trigger-error.xml
TimWolla: Yeah, finding a good example is hard, because generally speaking there is no good reason to use `trigger_error()` nowadays. Either use an Exception or use [`error_log()`](https://www.php.net/manual/en/function.error-log.php) or another logging facility. But PHP's internal error reporting mechanism which allows for side-effects is not a good idea.

That said, how about this:

```
$password = $_POST['password'] ?? '';
if ($password === '') {
  trigger_error("Using an empty password is unsafe", E_USER_WARNING);
}
$hash = password_hash($password, PASSWORD_DEFAULT);
```

cmb69: That would require that the error is shown to the user, though.

reference/errorfunc/functions/trigger-error.xml
Girgias: I think documenting it here is good, not sure how often changelogs are read

PR #4070, opened 2024-11-14
PHP 8.4: exit() is now a function

Girgias:


reference/misc/functions/exit.xml
```diff
@@ -3,105 +3,154 @@
...
+   <title>Basic <function>exit</function> example</title>
+   <programlisting role="php">
 <![CDATA[
 <?php
 
-$filename = '/path/to/data-file';
-$file = fopen($filename, 'r')
-    or exit("unable to open file ($filename)");
+// exit program normally
+exit;
```
TimWolla, line 126: Do we want to list the parenthesis-less version first or at all? I would not want to *promote* its usage.

Girgias: Reasonable point, I'll change this.

reference/misc/functions/exit.xml
```diff
@@ -3,105 +3,153 @@
 <refentry xml:id="function.exit" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>exit</refname>
-  <refpurpose>Output a message and terminate the current script</refpurpose>
+  <refpurpose>Terminate the current script with a given status code or message</refpurpose>
```
TimWolla, line 6:
```suggestion
  <refpurpose>Terminate the current script with a status code or message</refpurpose>
```

Perhaps to shorten this. It's obvious that the status code needs to be given.

TimWolla (approved):

PR #4073, opened 2024-11-15
Pcntl php84

devnexen:


devnexen:
for #3872

reference/pcntl/constants.xml
```diff
@@ -1502,6 +1526,57 @@
...
+    (<type>enum</type>)
+   </term>
+   <listitem>
+    <simpara>
+     Runs the process at high priority level but below Pcntl::UserInteractive ones.
+     Available as of PHP 8.4.0 (macOs only).
+    </simpara>
+   </listitem>
+  </varlistentry>
+ </variablelist>
```
Girgias, line 1579: This is an enum, so it should not be documented here actually see #4080 for how such a page should look (albeit we don't have rendering support yet).

Girgias (changes requested):

reference/pcntl/constants.xml
devnexen: Yes I expected you would ask me changes, kept it as a separate commit. Ok will have a look.

Girgias: Feel free to merge the first commit then :)

PR #4074, opened 2024-11-15
getclosurecalledclass.xml Provide more detailed info + CS

mmalferov:


reference/reflection/reflectionfunctionabstract/getclosurecalledclass.xml
```diff
@@ -13,7 +13,7 @@
   </methodsynopsis>
   <simpara>
    Returns the class as a <classname>ReflectionClass</classname> that
-   corresponds to <literal>static::</literal> inside the
+   corresponds to resolving the class name by calling <literal>static::</literal> inside the
```
TimWolla, line 16: I don't think calling is not the correct word, because this is not a function:
```suggestion
   corresponds to resolving the class name corresponding to <literal>static::</literal> inside the
```

TimWolla (approved):
LGTM other than the one remark.

PR #4075, opened 2024-11-15
override.xml Write a few words about `#[\Override]`

mmalferov:


language/predefined/attributes/override.xml
```diff
@@ -8,6 +8,12 @@
   <section xml:id="override.intro">
    &reftitle.intro;
    <para>
+    This attribute is used to check that a method exists in
+    a parent class or implemented interface.
+    The attribute <code>#["\Override"]</code> cannot target a class property, class constant,
+    or other class members except for class methods.
+    The method, marked this attribute but no matching parent method exists,
+    will cause an fatal error.
    </para>
```
TimWolla, line 17:
```suggestion
   <para>
    This attribute is used to indicate that a method is intended
    to override a method of a parent class or that it implements
    a method defined in an interface.
   </para>
   <para>
    If no method with the same name exists in a parent class or
    in an implemented interface a compile-time error will be
    emitted.
   </para>
```

I've rewritten this to make the description less technical and to explain the target use case: Communicating intent.

I've dropped the bit about the allowed targets, because I believe it would be better solved with #3897.

TimWolla (changes requested):
Thank you. While I've rewritten the entire text, it gave me a reason to finally do this.

language/predefined/attributes/override.xml
```diff
@@ -8,6 +8,14 @@
...
    <para>
+    This attribute is used to indicate that a method is intended
+    to override a method of a parent class or that it implements
+    a method defined in an interface.
+   </para>
+   <para>
+    If no method with the same name exists in a parent class or
+    in an implemented interface a compile-time error will be
+    emitted.
    </para>
```
Girgias, line 19:
```suggestion
   <simpara>
    This attribute is used to indicate that a method is intended
    to override a method of a parent class or that it implements
    a method defined in an interface.
   </simpara>
   <simpara>
    If no method with the same name exists in a parent class or
    in an implemented interface a compile-time error will be
    emitted.
   </simpara>
```

PR #4076, opened 2024-11-15
Fix Apache functions return types to reflect the stubs

scoopandrun:
The stubs for Apache functions `apache_request_headers`, `getallheaders` and `apache_response_headers` are as follows:

```php
<?php

// https://github.com/php/php-src/blob/PHP-8.3.12/sapi/apache2handler/php_functions.stub.php
// https://github.com/php/php-src/blob/PHP-8.3.12/sapi/cgi/cgi_main.stub.php

function apache_request_headers(): array {}

/** @alias apache_request_headers */
function getallheaders(): array {}

function apache_response_headers(): array {}
```

However, the return types in the docs wrongly include `false` as a return type for those functions.  
There even are some discrepancies within the same page between the **Description** content and the **Return Values** section.

This PR aims to fix these inconsistencies.

### DIFF
--- a/reference/apache/functions/apache-request-headers.xml
+++ b/reference/apache/functions/apache-request-headers.xml
@@ -27,8 +27,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   An associative array of all the HTTP headers in the current request, or
-   &false; on failure.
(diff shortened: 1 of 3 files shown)


Girgias (approved):
It seems like it used to be able to return false in PHP 5, thanks for spotting this!

PR #4078, opened 2024-11-16
[PHP 8.4] PCRE - PCRE2-v10.44 updates

Ayesh:
- Added note about `/r` modifier.
 - Regex sub-pattern quantifier changes.

More changes noted [here](https://github.com/PCRE2Project/pcre2/blob/pcre2-10.44/ChangeLog#L7-L385) and [here](https://php.watch/versions/8.4/pcre2-regexp-syntax-changes). 

Unicode script list [here](https://www.php.net/manual/en/regexp.reference.unicode.php) was not updated for quite a while. As of PHP 8.4.0, there are close to 230 scripts and 160 shorthand script names. The table would need a bigger update, and not part of the PR, but will be added in a future PR.


reference/pcre/pattern.syntax.xml
```diff
@@ -1573,11 +1573,20 @@
...
-   but a literal string of four characters.
+   as a literal character. For example, <literal>{,6}</literal>
+   is not a quantifier, but a literal string of four characters.
+
+   As of PHP 8.4.0, the PCRE extension is bundled with PCRE2 version 10.44,
+   which allows patterns such as <literal>\d{,8}</literal> and they are
+   interpreted as <literal>\d{0,8}</literal>.
+
+   Further, as of PHP 8.4.0, space characters around quantifiers such as
+   <literal>\d{0 , 8}</literal> and <literal>\d{ 0 , 8 }</literal> are allowed.
```
Girgias, line 1589: Possibly those should be 3 different `<simpara>` ?

Girgias (commented):
Thanks, one remark.

reference/pcre/pattern.syntax.xml
Ayesh: Thank you, it makes sense that we split them. 

|`<simpara>`|`<para>`|
|--|--|
|image|image|

I tried with three `<simpara>` tags, and a separate `<para>`, I liked the second one because it puts all PHP 8.4 stuff together, and looks more compact. What do you think? I pushed with the `<para>` tag now, but if you think `<simpara>` makes more sense, I can quickly change to that too 🙏.

Girgias: `<simpara>` is better than a `<para>` tag, it is just the grouping that matters here.

reference/pcre/pattern.syntax.xml
```diff
@@ -1573,11 +1573,22 @@
...
   </para>
```
Girgias, line 1592:
```suggestion
  </simpara>
```

reference/pcre/pattern.syntax.xml
```diff
@@ -1573,11 +1573,22 @@
 
    <literal>\d{8}</literal>
 
-   matches exactly 8 digits. An opening curly bracket that
+   matches exactly 8 digits.
+
+  </para>
+  <para>
```
Girgias, line 1579:
```suggestion
  <simpara>
```

PR #4083, opened 2024-11-17
[PHP 8.4] Function return type changes

Ayesh:


Girgias (approved):
Doesn't seem they could return false in any case previously, so I don't think another changelog is needed.

PR #4084, opened 2024-11-17
Remove duplicate entry for E_* constants

Girgias:
They are better described in reference/errorfunc/constants.xml

Not exactly sure we should do this, but I found this by working on a new Documentation tool I'm working on.


cmb69 (commented):
Hmm, unsure about this either. One may argue that this page should list all core constants, not only those that are declared elsewhere. xinclude to the rescue? Not really sure about that either. Another option might be to group the constants, and have a single link to error reporting constants. If PhD properly supports multiple `terms` for a single `varlistentry`, that might be the best option.

Girgias:
We cannot use XInclude because libxml2 only supports XInclude 1.0 (and not 1.1) which will copy the IDs.

I fixed it in my tool by parsing the appendix first, and letting `reference/*` files override it.

PR #4086, opened 2024-11-17
MySQLi: Fix incorrect XML IDs for constant linking

Girgias:
Found via a new documentation tool I am working on.

### DIFF
--- a/reference/curl/constants_curl_setopt.xml
+++ b/reference/curl/constants_curl_setopt.xml
@@ -1,5 +1,5 @@
  <variablelist role="constant_list">
-  <title><function>curl-setopt</function></title>
+  <title><function>curl_setopt</function></title>
   <varlistentry xml:id="constant.curlopt-abstract-unix-socket">
    <term>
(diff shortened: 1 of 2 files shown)


kamil-tekiela (approved):

haszi:
I'm wondering whether we should include something in either `doc-base/configure.php` or PhD that would check existing IDs against the contents of `<function>`, `<methodname>`, `<constant>`, etc. tags which could give us an idea of the extent of similar issues.

Girgias:
> I'm wondering whether we should include something in either `doc-base/configure.php` or PhD that would check existing IDs against the contents of `<function>`, `<methodname>`, `<constant>`, etc. tags which could give us an idea of the extent of similar issues.

Maybe, but it might be a bit tricky? I have added this to check in https://github.com/Girgias/stub-to-docbook, and it should be possible to use it to write some sort of report about the current status of constants. But a lot of them don't even seem to have IDs in external extensions.

PR #4087, opened 2024-11-17
Group E_* constants in appendix

cmb69:
Listing all `E_*` constants in the appendix appears to be reasonable, but having a dedicated `varlistentry` for each seems to be overkill.

---

e-consts


Girgias (approved):
MSTM

PR #4088, opened 2024-11-17
enumerations.xml Amend the CS of the code examples

mmalferov:


language/enumerations.xml
```diff
@@ -952,8 +980,13 @@ enum UserStatus: string
 <?php
 
 foreach (UserStatus::cases() as $case) {
-    printf('<option value="%s">%s</option>\n', $case->value, $case->label());
+    printf(
+        "<option value=\"%s\">%s</option>\n",
+        $case->value,
+        $case->label()
+    );
```
tiffany-taylor, line 987: I'm not sure I agree with breaking this out into separate lines. If I recall correctly, this isn't done elsewhere in the manual in similar situations, but do correct me if I'm wrong.

tiffany-taylor:
Changes in this PR are made to the pages in this section: https://www.php.net/manual/en/language.enumerations.php

language/enumerations.xml
mmalferov: Rolled back

vrana:
I see no point in adding newlines before `?>`.

Crell:
Please remove the newlines before `?>`, as we don't need those, and rebase the PR.  We should be able to merge it then.  Thanks.

jordikroon:
The conflicts in this PR (https://github.com/php/doc-en/pull/5479) are caused by a recent revision that enabled WASM for examples. So it would need another rebase..

I don't think changing code styles for examples on it's own gives any benefits the docs. The docs don't follow any code style guidelines other than it should be understandable for a developer. 

Adding whitespaces or newlines don't really give any reasonable benefits to this. For that reason I don't see any valid reason to make this change.

@tiffany-taylor / @Crell any thoughts on this?

Crell:
Ideally code blocks should follow PER-CS, though the degree to which it's worth it to apply those to pre-existing code blocks is debatable.  (The whole docs is very inconsistent on just about everything, as many people reject the idea of updating it to current conventions for fear of making more noise for translators.)

jordikroon:
> Ideally code blocks should follow PER-CS, though the degree to which it's worth it to apply those to pre-existing code blocks is debatable.

Ah I see, it is in fact [explicitly written](https://doc.php.net/guide/cs-for-examples.md) in the style docs. Though I have seen differences in opinion. I prefer to do the `simpara` principle. So only updating examples when these blocks are touched. Or do it file based if an alteration of these documentation files is required.

Anyway, since it's a relatively small change I have no objections to merge this.

Still need to do a rebase I'm afraid.

PR #4089, opened 2024-11-17
Refactor network constant page, and associated functions

Girgias:
Remove duplication of information
Use standard markup
Remove personalization

I am wondering if part of the return value section for `dns_get_record()` should not be moved to the constant page, but maybe not.


reference/network/functions/dns-get-record.xml
```diff
@@ -52,31 +52,10 @@
...
-       parameter. May be any one of the following:
-       <constant>DNS_A</constant>, <constant>DNS_CNAME</constant>,
-       <constant>DNS_HINFO</constant>, <constant>DNS_CAA</constant>,
-       <constant>DNS_MX</constant>, <constant>DNS_NS</constant>,
-       <constant>DNS_PTR</constant>, <constant>DNS_SOA</constant>,
-       <constant>DNS_TXT</constant>, <constant>DNS_AAAA</constant>,
-       <constant>DNS_SRV</constant>, <constant>DNS_NAPTR</constant>,
-       <constant>DNS_A6</constant>, <constant>DNS_ALL</constant>
-       or <constant>DNS_ANY</constant>.
+       To limit the query, use on of the
```
cmb69, line 55:
```suggestion
       To limit the query, use one of the
```

reference/network/functions/openlog.xml
```diff
@@ -42,136 +42,45 @@
...
       </para>
-      <note>
-       <para>
-        <constant>LOG_USER</constant> is the only valid log type under Windows
-        operating systems
-       </para>
-      </note>
+      <simpara>
+       This argument indicates the nature of the programming which is
+       logging the message.
```
cmb69, line 80: Not sure about the wording; the old one was copied almost literally from [man syslog](https://www.man7.org/linux/man-pages/man3/syslog.3.html).

reference/network/constants.xml
```diff
@@ -3,272 +3,695 @@
...
+   <listitem>
+    <simpara>
+     Defined as part of early IPv6, but downgraded to historic by
+     <link xlink:href="&url.rfc;6563">RFC 6563</link>.
+    </simpara>
+    <warning>
+     <simpara>
+      Not available on Windows.
+     </simpara>
+    </warning>
```
cmb69, line 561: DNS_A6 is available on Windows for ages (PHP 5.3 or 5.4).

reference/network/constants.xml
```diff
@@ -3,272 +3,695 @@
...
+    <simpara>
+     Reserved for local use.
+    </simpara>
+    <warning>
+     <simpara>
+      Not available on Windows.
+     </simpara>
+    </warning>
+   </listitem>
+  </varlistentry>
```
cmb69, line 293: Could group these; not sure about the IDs, though.

reference/network/constants.xml
```diff
@@ -3,272 +3,695 @@
...
+    <constant>LOG_USER</constant>
+    (<type>int</type>)
+   </term>
+   <listitem>
+    <simpara>
+     For generic user-level messages.
+    </simpara>
+    <warning>
+     <simpara>
+      On Windows, this is the only valid log type.
```
cmb69, line 354: No longer true as of PHP 4.0; however, the `facility` is completely ignored on Windows. ;)

cmb69 (commented):
> I am wondering if part of the return value section for dns_get_record() should not be moved to the constant page, but maybe not.

I'd leave that as is.

reference/network/constants.xml
Girgias: We split this explicitly for constant linking.

reference/network/constants.xml
Girgias: So should I drop all of the `Not available on Windows` comments for LOG_LOCAL? (or would those need to be defined manually to 0 on Windows?)

cmb69: `LOG_LOCAL?` is indeed not available on Windows (but the others, including `LOG_USER` are). However, it doesn't matter what you pass as `facitiliy` to `openlog()`; in the [Windows implementation](https://heap.space/xref/php-src/win32/wsyslog.c?r=17997a90#146), the argument is unused. So besides removing this warning, see the suggestion below.

reference/network/functions/openlog.xml
```diff
@@ -42,136 +42,45 @@
...
+       This is useful as the syslog configuration can handle messages
+       logged by different facilities differently.
+      </simpara>
```
cmb69, line 83:
```suggestion
      </simpara>
      <note>
       <simpara>
        This parameter has no meaning on Windows.
       </simpara>
      </note>
```
Or maybe is ignored/unused.

reference/network/constants.xml
Girgias: Well the docs were wrong then :D

reference/network/functions/openlog.xml
Girgias: I was looking at the [IBM docs](https://www.ibm.com/docs/en/zos/3.1.0?topic=functions-syslog-send-message-control-log) for wording, but happy to change it.

reference/network/constants.xml
Girgias: Is `DNS_CAA` available on Windows too?

cmb69: No, `DNS_CAA` is not available (apparently nothing like this is supported; quickly tried).

cmb69 (approved):
Looks good. Thank you!

PR #4090, opened 2024-11-17
POSIX: Fix typo in constant

Girgias:
Seems like a typo?

Also, the `POSIX_PC_SYMLINK_MAX` from PHP 8.3 seems to be undocumented, and the `POSIX_SC_OPEN_MAX` which has existed since I don't know when is also not documented. Could you document those @devnexen?

### DIFF
--- a/reference/posix/constants.xml
+++ b/reference/posix/constants.xml
@@ -573,9 +573,9 @@
      </simpara>
     </listitem>
    </varlistentry>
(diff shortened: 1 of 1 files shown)


devnexen (approved):

Girgias:
I found out from when `POSIX_SC_OPEN_MAX` is... it is just in master... so ignore this.

PR #4093, opened 2024-11-18
PHP 8.4: CSV related changes and deprecations

Girgias:
I *might* have pushed it a bit with the XIncludes and XPath queries.

### DIFF
--- a/reference/filesystem/functions/fgetcsv.xml
+++ b/reference/filesystem/functions/fgetcsv.xml
@@ -102,7 +102,7 @@
         As of PHP 8.4.0, depending on the default value of
         <parameter>escape</parameter> is deprecated.
         It needs to be provided explicitly either positionally or by the use
-        of <link linkend="functions.named-arguments">Named Arguments</link>.
+        of <link linkend="functions.named-arguments">named arguments</link>.
        </simpara>
       </warning>
      </listitem>
--- a/reference/filesystem/functions/fgetcsv.xml
+++ b/reference/filesystem/functions/fgetcsv.xml
@@ -217,7 +217,6 @@ if (($handle = fopen("test.csv", "r")) !== FALSE) {
    <member><function>explode</function></member>
    <member><function>file</function></member>
    <member><function>pack</function></member>
-   <member><function>fputcsv</function></member>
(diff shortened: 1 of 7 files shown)


reference/filesystem/functions/fgetcsv.xml
```diff
@@ -96,6 +97,14 @@
         even meant to escape itself.
        </simpara>
       </note>
+      <warning>
+       <simpara>
+        As of PHP 8.4.0, depending on the default value of
+        <parameter>escape</parameter> is deprecated.
+        It needs to be provided explicitly either positionally or by the use
+        of <link linkend="functions.named-arguments">Named Arguments</link>.
```
cmb69, line 105: Not sure if we should use title case here.

cmb69 (commented):
> I _might_ have pushed it a bit with the XIncludes and XPath queries.

Yeah, that looks somewhat fragile (e.g. relying on the parameter numbers). Might be better to use some entities here.

reference/strings/functions/str-getcsv.xml
```diff
@@ -104,6 +75,19 @@
...
+      <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('function.fgetcsv')/db:refsect1[@role='changelog']//db:row[db:entry[text()='8.4.0']]/.)">
+       <xi:fallback/>
+      </xi:include>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Now throws a <exceptionname>ValueError</exceptionname> if
+        <parameter>separator</parameter>, <parameter>enclosure</parameter>,
+        or <parameter>escape</parameter> is invalid.
+        This mimics the behaviour of <function>fgetcsv</function> and
```
cmb69, line 87:
```suggestion
        This mimics the behavior of <function>fgetcsv</function> and
```

cmb69 (approved):
I'm not a fan of xincludes, but this looks quite reasonable. Minor nit below.

Girgias:
I prefer XIncludes to Entities as `language-snippets` is a pain as a translator IMHO.

PR #4099, opened 2024-11-20
PHP 8.4: Single underscore as class name is deprecated

Girgias:


cmb69 (approved):
I think I'd put the sentence about the deprecation at the bottom of this paragraph, but it is okay as is.

PR #4101, opened 2024-11-20
Add tags for arithmetic operator page

Girgias:
Slightly change wording

### DIFF
--- a/language/operators/arithmetic.xml
+++ b/language/operators/arithmetic.xml
@@ -18,7 +18,7 @@
    </thead>
    <tbody>
     <row>
-     <entry>+$a</entry>
+     <entry><code>+$a</code></entry>
      <entry>Identity</entry>
      <entry>
       Conversion of <varname>$a</varname> to <type>int</type> or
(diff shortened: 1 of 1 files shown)


language/operators/arithmetic.xml
```diff
@@ -18,56 +18,58 @@
...
  </table>
  <simpara>
-  The division operator ("/") returns a float value unless the two operands
-  are integers (or strings that get converted to integers) and the numbers
-  are evenly divisible, in which case an integer value will be returned. For
-  integer division, see <function>intdiv</function>.
+  The division operator <literal>/</literal> returns a <type>float</type>
+  value unless the two operands are <type>int</type> (or
+  <link linkend="language.types.numeric-strings">numeric strings</link>
+  which are type juggled to <type>int</type>) and the numerator is a multiple
```
kamil-tekiela, line 70:
```suggestion
  The division operator <literal>/</literal> returns a &float;
  value unless the two operands are &integer; (or
  <link linkend="language.types.numeric-strings">numeric strings</link>
  which are type juggled to &integer;) and the numerator is a multiple
```

kamil-tekiela (approved):

language/operators/arithmetic.xml
Girgias: Some translations change how the XML entities act, so I would rather not use them.

PR #4102, opened 2024-11-20
PHP 8.4: Raising 0 to a negative exponent is deprecated

Girgias:


cmb69 (approved):
Thank you!

Should we also mention that in the artihmetic operator section? Or maybe amend the table there to tell the `**` is the same as `pow()`?

Girgias:
> Thank you!
> 
> Should we also mention that in the artihmetic operator section? Or maybe amend the table there to tell the `**` is the same as `pow()`?

I thought about this, because I thought that page would already indicate throwing an Error for division by zero but it doesn't soooo not exactly sure

PR #4103, opened 2024-11-20
PHP 8.4: Indicate extensions have been unbundled

Girgias:


reference/oci8/configure.xml
```diff
@@ -242,6 +242,12 @@ extension=oci8.so
 </section>
 <section>
  <title>Installing OCI8 as a Shared Extension when Building PHP</title>
+ <warning>
+  <simpara>
+   This extension is <emphasis>UNBUNDLED</emphasis> as of PHP 8.4.0.
+   Thus this is no longer applicable as of PHP 8.4.0.
+  </simpara>
+ </warning>
```
cmb69, line 250: I think this warning (and the warning below) can be dropped. After all, users can still copy the extension into the php-src tree and build like before.

cmb69 (commented):
Thank you!

reference/imap/book.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
- 
+<!-- State: deprecated -->
 <book xml:id="book.imap" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <?phpdoc extension-membership="bundledexternal" ?>
```
Girgias, line 4: Actually I probably should fix these

reference/imap/configure.xml
```diff
@@ -1,36 +1,49 @@
...
+   This location depends on where you created this directory according to the
+   description above. <productname>Windows</productname> users may include the
+   <filename>php_imap.dll</filename> DLL in &php.ini;.
+  </para>
+  <note>
+   <simpara>
+    Depending on how the c-client was configured, you might also need
+    to add <option
+    role="configure">--with-imap-ssl=/path/to/openssl/</option> and/or
+    <option role="configure">--with-kerberos=/path/to/kerberos</option>
```
cmb69, line 32: Let's ignore that `--with-imap-ssl` and `--with-kerberos` no longer accept paths as of PHP 7.4.0 (using pkg-config instead). :)

cmb69 (approved):
Thank you!

PR #4107, opened 2024-11-20
PHP 8.4: Document type widening of mode param of round()

Girgias:
@haszi not sure if I should use a `<constant>` tag here for linking, but `<enumidentifier>` is not allowed... I could use `<code>`, however.

### DIFF
--- a/reference/math/functions/max.xml
+++ b/reference/math/functions/max.xml
@@ -61,7 +61,7 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>value_array</parameter></term>
+     <term><parameter>value</parameter></term>
      <listitem>
       <para>
        An array containing the values.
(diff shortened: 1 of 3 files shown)


haszi:
> @haszi not sure if I should use a `<constant>` tag here for linking, but `<enumidentifier>` is not allowed... I could use `<code>`, however.

Semantically, `<constant>` seems to be much closer to what we want here.

PHP isn't the only language to allow enums as default values so the `<enumidentifier>` not being allowed seems like an oversight in the DocBook spec. :-)

haszi:
`<constant>` won't link to an enum. I'm not sure how to handle the linking though.

Girgias:
> `<constant>` won't link to an enum. I'm not sure how to handle the linking though.

I suppose one could add a `role` attribute to the constant tag, e.g. `<constant role="enum">` to link enum cases?

Girgias:
Superseded by #4425

PR #4110, opened 2024-11-20
Hooks reflection

Crell:
New PR, includes #4039, which generates the stubs.

Resolves #4039.

I'm not certain if the enum is in the right place, but I wasn't sure where else to put it.  Open to suggestions.

### DIFF
--- a/reference/reflection/reflectionproperty/gethook.xml
+++ b/reference/reflection/reflectionproperty/gethook.xml
@@ -2,7 +2,7 @@
 <refentry xml:id="reflectionproperty.gethook" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>ReflectionProperty::getHook</refname>
-  <refpurpose>Description</refpurpose>
+  <refpurpose>Returns a reflection object for a specified hook.</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
@@ -13,7 +13,7 @@
   </methodsynopsis>
   &warn.undocumented.func;
   <simpara>
-   Description.
+   Gets the reflection of the property's hook, if any.
   </simpara>
  </refsect1>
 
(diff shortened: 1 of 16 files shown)


Girgias (approved):

Girgias:
I didn't realize, but @Crell you didn't address some of the review comments.

reference/reflection/reflectionproperty/gethook.xml
```diff
@@ -0,0 +1,91 @@
...
+<?php
+class Example
+{
+    public string $name { get => "Name here"; }
+}
+
+$rClass = new \ReflectionClass(Example::class);
+$rProp = $rClass->getProperty('name');
+$rProp->getHook(PropertyHookType::Get); // Returns an instance of \ReflectionMethod.
+$rProp->getHook(PropertyHookType::Set); // Returns null.
```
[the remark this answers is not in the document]
Crell: Hm.  In that case, do we even need the screen tag anymore?  Just have it show output and leave it to the new wasm runner.

Crell:
Switched all the examples to generate output.  I did not include the output, because it's a hassle to do and we now have the Wasm runner that will produce it for us on-demand.

Crell:
@Girgias Which comments haven't been addressed?

Girgias (changes requested):
Output of examples is missing.

Crell:
I disagree that manually pre-generating the output is worthwhile now that we have the Wasm runner.  If someone else would like to modify the PR to include that, go ahead, but I have limited time with a baby now, so won't be doing so.

reference/filesystem/functions/touch.xml
```diff
@@ -135,6 +135,27 @@ if (!touch('some_file.txt', $time)) {
...
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>clearstatcache</function></member>
+    <member><function>fileatime</function></member>
+    <member><function>filemtime</function></member>
+   </simplelist>
+  </para>
```
Girgias, line 158: Seems this snuck in the PR. when it should be in another one.

Crell: Oops indeed.  Rebased out that commit, and also rebased on master to avoid conflicts.

philip:
+1 to committing this.

PR #4112, opened 2024-11-20
Document PHP 8.3 proc_get_status() change

nielsdos:


reference/exec/functions/proc-get-status.xml
```diff
@@ -90,8 +90,17 @@
       <entry>
        The exit code returned by the process (which is only
        meaningful if <literal>running</literal> is &false;).
-       Only first call of this function return real value, next calls return
-       <literal>-1</literal>.
+       Prior to PHP 8.3.0, only first call of this function returned the real
+       value, the following calls returned <literal>-1</literal>.
```
Girgias, line 94:
```suggestion
       Prior to PHP 8.3.0, only the first call of this function returned the real
       value, the following calls returned <literal>-1</literal>.
```

Girgias (approved):
Minor nit, but LGTM :)

PR #4114, opened 2024-11-20
PHP 8.4: Document registerPhpFunctions() changes

nielsdos:


reference/dom/domxpath/registerphpfunctions.xml
```diff
@@ -29,8 +29,9 @@
        Use this parameter to only allow certain functions to be called from XPath.
       </para>
       <para>
-       This parameter can be either a <type>string</type> (a function name) or 
-       an <type>array</type> of function names.
+       This parameter can be either a <type>string</type> (a function name), or
+       an <type>array</type> of function names or
+       <type>string</type> => <type>callable</type> entries.
```
Girgias, line 34: Possibly? As the current wording was leaving me confused.

```suggestion
       This parameter can be one of the following:
       a <type>string</type> (a function name),
       an indexed <type>array</type> of function names,
       or an associative <type>array</type> with keys being the function name
       and the associated value being the <type>callable</type>.
```

Girgias (approved):
Thank you!

PR #4116, opened 2024-11-21
property-hooks.xml Replace 'all' with 'non-static', as not all + typos

mmalferov:
The word `all` seems to deceive the reader a little :)

How about adding some specifics about which properties are allowed to declare hooks for?

### DIFF
--- a/language/oop5/property-hooks.xml
+++ b/language/oop5/property-hooks.xml
@@ -25,8 +25,9 @@
   </listitem>
  </orderedlist>
  <simpara>
-  There are two hooks available on typed or untyped non-static properties: <literal>get</literal> and <literal>set</literal>.
+  There are two hooks available on non-static properties: <literal>get</literal> and <literal>set</literal>.
   They allow overriding the read and write behavior of a property, respectively.
+  Hooks are available for both typed and untyped properties.
  </simpara>
  <simpara>
   A property may be "backed" or "virtual".
--- a/language/oop5/property-hooks.xml
+++ b/language/oop5/property-hooks.xml
@@ -228,7 +228,7 @@ class Example
   </example>
   <simpara>
    For virtual properties, if a hook is omitted then that operation does
-   not exist and trying to use it wil produce an error.
+   not exist and trying to use it will produce an error.
    Virtual properties take up no memory space in an object.
    Virtual properties are suited for "derived" properties,
    such as those that are the combination of two other properties.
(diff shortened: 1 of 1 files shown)


language/oop5/property-hooks.xml
```diff
@@ -95,7 +95,7 @@ print $example->foo;
   </simpara>
   <simpara>
    At least one of the hooks references <code>$this->foo</code>, the property itself.
-   That means the property wll be "backed."
+   That means the property wll be "backed".
```
Crell, line 98: This one is a little tricky.  Standard English (especially in the US) says the period goes inside the quote.  It's informally common to put it outside when the period is not part of the quote, which makes logical sense; I do that often myself, in informal writing.  However, formal writing with every editor I've worked with puts the period inside the quote.  (Same for commas.)  Absent a formal decision to do this site wide, we should stick with standard English convention.

language/oop5/property-hooks.xml
```diff
@@ -25,7 +25,7 @@
   </listitem>
  </orderedlist>
  <simpara>
-  There are two hooks available on all properties: <literal>get</literal> and <literal>set</literal>.
+  There are two hooks available on typed or untyped non-static properties: <literal>get</literal> and <literal>set</literal>.
```
Crell, line 28: Agreed on clarifying non-static.  But "typed or untyped" reduces to all, so is redundant.

Crell:
Looks good aside from those comments, and the need to rebase.  Thanks.

language/oop5/property-hooks.xml
mmalferov: By saying "typed or untyped", I meant that hooks can work with both typed and untyped properties, or rather, that it is not necessary to type a property ;) As opposed to, for example, read-only properties that work only with typed properties

language/oop5/property-hooks.xml
mmalferov: Thanks for the explanation. By "quote" the rules mean quote (quotation marks) like "foo", or a quote like: Did he say, "Hello, world!"? Or both?

ekinhbayar: Just in case it wasn't noticed at all, I'd like to point out:  `wll` -> `will`

language/oop5/property-hooks.xml
mmalferov: I have indicated the availability of hooks for both typed and untyped properties in a separate sentence. It seems that now it doesn't look redundant and informs the reader that hooks will work with both typed properties and untyped ones.

Crell (approved):

Crell:
Looks good now.  Thanks!

PR #4122, opened 2024-11-21
PHP 8.4: Flushing headers without a body will now succeed

nielsdos:


reference/outcontrol/functions/flush.xml
```diff
@@ -54,6 +54,28 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       Flushing headers without a body will now succeed in FastCGI.
```
Girgias, line 71: Is FastCGI documented somewhere in the manual? If yes adding a link would be good.

Girgias (approved):
Minor remark but LGTM.

reference/outcontrol/functions/flush.xml
nielsdos: Not that I know.

PR #4143, opened 2024-11-24
Filter: Document FILTER_CALLBACK

Girgias:
Remove the "Other filter" section

### DIFF
--- a/reference/filter/constants.xml
+++ b/reference/filter/constants.xml
@@ -1047,7 +1047,7 @@
     </simpara>
     <simpara>
      The value returned by the callback will be the value returned by
-     the filter function that is called.
+     the invoked filter function.
     </simpara>
     <example>
      <title>
@@ -1085,7 +1085,7 @@ NULL
     </example>
     <warning>
      <simpara>
-      This filter cannot be used with any other filter flags, e.g;
+      This filter cannot be used with any other filter flags, e.g.
       <constant>FILTER_NULL_ON_FAILURE</constant>.
      </simpara>
     </warning>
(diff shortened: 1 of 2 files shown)


reference/filter/constants.xml
```diff
@@ -1039,11 +1039,56 @@
...
     <simpara>
-     ID of "callback" filter.
+     This filter delegates the filtering to a user defined function.
+     The <type>callable</type> is passed via the
+     <parameter>options</parameter> parameter as the value associated to
+     the <literal>'options'</literal> key.
     </simpara>
+    <simpara>
+     The value returned by the callback will be the value returned by
+     the filter function that is called.
```
kamil-tekiela, line 1050:
```suggestion
     the invoked filter function.
```

reference/filter/constants.xml
```diff
@@ -1039,11 +1039,56 @@
...
+     <screen>
+<![CDATA[
+string(10) "val1dL0gin"
+NULL
+]]>
+     </screen>
+    </example>
+    <warning>
+     <simpara>
+      This filter cannot be used with any other filter flags, e.g;
```
kamil-tekiela, line 1088:
```suggestion
      This filter cannot be used with any other filter flags, e.g.
```

kamil-tekiela (approved):
Good, but what I think this page still lacks is the information about the callable signature. The example doesn't specify parameter type but makes assumption that it's always string. Is that true?

kamil-tekiela:
> > Good, but what I think this page still lacks is the information about the callable signature. The example doesn't specify parameter type but makes assumption that it's always string. Is that true?
> 
> Yeah it is always a string: https://3v4l.org/J6k7r

Good. Interesting that the last two didn't even trigger the callback.

PR #4144, opened 2024-11-24
Filter: Describe sanitization flags in constant page

Girgias:


reference/filter/book.xml
```diff
@@ -1,42 +1,53 @@
...
-   sanitize the data, so it may alter it by removing undesired characters.
-   For example, passing in <constant>FILTER_SANITIZE_EMAIL</constant> will
+  </simpara>
+  <simpara>
+   A validation filter is used to check if the data meets certain criteria.
+   These filters are identified by the
+   <constant>FILTER_VALIDATE_<replaceable>*</replaceable></constant>
+   constants.
+   For example, the <constant>FILTER_VALIDATE_EMAIL</constant> filter
+   determine if the data is a valid email address.
```
kamil-tekiela, line 26:
```suggestion
   can be used to determine if the data is a valid email address.
```

reference/filter/book.xml
```diff
@@ -1,42 +1,53 @@
...
+   However, it will never alter the input data.
+  </simpara>
+  <simpara>
+   Sanitization on the other hand will "clean-up" the data,
```
kamil-tekiela, line 30:
```suggestion
   Sanitization on the other hand will "clean up" the data,
```
It should be a verb.

reference/filter/book.xml
```diff
@@ -1,42 +1,53 @@
...
+   therefore it may alter the input data by removing undesired characters.
```
kamil-tekiela, line 31:
```suggestion
   therefore it may alter the input data by adding or removing characters.
```
Sigh... Did I ever mention how I hate the sanitization filters?

reference/filter/book.xml
```diff
@@ -1,42 +1,53 @@
...
+  <simpara>
+   Most filters support optional <emphasis>flags</emphasis> that can tweak
+   the behavior of the filter.
+   These flags are identified by the
+   <constant>FILTER_FLAG_<replaceable>*</replaceable></constant>
+   constants.
+   For example, using the <constant>FILTER_FLAG_PATH_REQUIRED</constant> with
+   the <constant>FILTER_VALIDATE_URL</constant> validation filter
+   requires that the <acronym>URL</acronym> has a path.
+   (E.g. <literal>/foo</literal> in <literal>https://example.org/foo</literal>).
```
kamil-tekiela, line 49:
```suggestion
   requires that the <acronym>URL</acronym> has a path
   (e.g. <literal>/foo</literal> in <literal>https://example.org/foo</literal>).
```
The parenthesised part is not a sentence on its own.

reference/filter/constants.xml
```diff
@@ -853,10 +859,24 @@
    </term>
    <listitem>
     <simpara>
-     ID of "string" filter.
-     (<emphasis>Deprecated</emphasis> as of PHP 8.1.0,
-     use <function>htmlspecialchars</function> instead.)
+     This filter strip tags and HTML-encodes double and single quotes.
```
kamil-tekiela, line 862:
```suggestion
     This filter strips tags and HTML-encodes double and single quotes.
```

reference/filter/constants.xml
```diff
@@ -853,10 +859,24 @@
...
+    </simpara>
+    <simpara>
+     Optionally it can strip or encode special characters if used together with
```
kamil-tekiela, line 865:
```suggestion
     Optionally it can strip or encode specified characters if used together with
```
I can't think of a better word, but we should stop calling them special.

reference/filter/constants.xml
```diff
@@ -879,7 +899,13 @@
    </term>
    <listitem>
     <simpara>
-     ID of "encoded" filter.
+     This filter URL-encode a string.
```
kamil-tekiela, line 902:
```suggestion
     This filter URL-encodes a string.
```

reference/filter/constants.xml
```diff
@@ -879,7 +899,13 @@
...
+    </simpara>
+    <simpara>
+     Optionally it can strip or encode special characters if used together with
```
kamil-tekiela, line 905:
```suggestion
     Optionally it can strip or encode specified characters if used together with
```

reference/filter/constants.xml
```diff
@@ -889,8 +915,22 @@
     (<type>int</type>)
    </term>
    <listitem>
+    <para>
+     This filter HTML-encode
```
kamil-tekiela, line 919:
```suggestion
     This filter HTML-encodes
```

reference/filter/constants.xml
```diff
@@ -889,8 +915,22 @@
...
+      <member><literal>"</literal></member>
+      <member><literal>&lt;</literal></member>
+      <member><literal>&gt;</literal></member>
+      <member><literal>&amp;</literal></member>
+     </simplelist>
+     and characters with an ASCII value less than 32.
+    </para>
     <simpara>
-     ID of "special_chars" filter.
+     Optionally it can strip special characters if used together with
```
kamil-tekiela, line 930:
```suggestion
     Optionally it can strip specified characters if used together with
```

reference/filter/book.xml
Girgias: They are a mess I agree...

reference/filter/constants.xml
```diff
@@ -853,10 +873,53 @@
...
+      <emphasis>Deprecated</emphasis> as of PHP 8.1.0,
+      use <function>htmlspecialchars</function> instead.
+     </simpara>
+    </warning>
+    <warning>
+     <simpara>
+      The way this filter strips tags is not equivalent to
+      <function>strip_tags</function>.
+      It is a more naive implementation that will remove all characters
+      between any <literal>&lt;</literal> and <literal>&gt;</literal>
```
kamil-tekiela, line 899: AFAIK it's only after &lt; and the &gt; is optional.

Do we really need an example for this here? This page is already getting quite long, If we'd try to explain what this filter does we'd need an entire separate page with many more examples. 

Maybe just leave `The way this filter strips tags is not equivalent to
      <function>strip_tags</function>.` and don't worry about the rest. It's already deprecated anyway.

Girgias: Yeah might as well.

reference/filter/constants.xml
```diff
@@ -879,8 +923,20 @@
...
+     the <constant>FILTER_FLAG_STRIP_<replaceable>*</replaceable></constant>
+     and <constant>FILTER_FLAG_ENCODE_<replaceable>*</replaceable></constant>
+     filter sanitization flags.
+    </simpara>
+    <tip>
+     <simpara>
+      The <function>trim</function> function can be used prior to using this
+      filter to not URL encode spaces into <literal>"%20"</literal>.
+     </simpara>
+    </tip>
```
kamil-tekiela, line 939: Why is this useful here? Surely, if someone needs to use `trim()` before, they wouldn't use this filter at all. They could just use `urlencode`. What's so special about spaces at the start and end of the input that it needs to be mentioned for this filter? Also, wouldn't this confuse beginners who want to encode spaces to `+` instead?

kamil-tekiela (approved):
Everything looks good, but I have doubts about the new tip you added.

reference/filter/constants.xml
Girgias: This was such a highly upvoted note that I add a thing, but this whole extension is... a choice.

Will remove.

PR #4145, opened 2024-11-24
Redesign is_callable() page

kamil-tekiela:


reference/var/functions/is-callable.xml
```diff
@@ -17,7 +17,8 @@
    <methodparam choice="opt"><type>string</type><parameter role="reference">callable_name</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Verify that a value is a <type>callable</type>.
+   Verifies that a value is callable, i.e. it's a valid argument to the
+   <function>call_user_func</function> function.
```
Girgias, line 21: I don't think we should be promoting `call_user_func()` considering that effectively all usages of `call_user_func()` can (and should) be replaced by `$fn()`

Also, why remove the `<type>` tag around `callable`?

reference/var/functions/is-callable.xml
```diff
@@ -73,46 +74,60 @@
...
+    &example.outputs;
+    <screen>
+<![CDATA[
+bool(true)
+string(12) "someFunction"
+]]>
+    </screen>
+   </example>
+   <example>
+    <title>Checking whether a string can be called as a function</title>
```
Girgias, line 102:
```suggestion
    <title>Checking whether an array can be called as a function</title>
```
No?

reference/var/functions/is-callable.xml
kamil-tekiela: For now, a callable type is not the only thing that is callable (English word). There are certain things that are accepted by `call_user_func` but are not of type callable. I was trying to convey that in this paragraph. I am not happy to mention `call_user_func` here either, but I saw no other way to explain what `is_callable` checks.

Girgias: Right okay, this deals with https://wiki.php.net/rfc/deprecate_partially_supported_callables

Girgias: How about:
```suggestion
   Verifies that a value is <type>callable</type>, or that it can be called using the
   <function>call_user_func</function> function.
```
?

reference/var/functions/is-callable.xml
```diff
@@ -38,10 +39,10 @@
...
       <para>
        If set to &true; the function only verifies that
-       <parameter>value</parameter> might be a function or method. It will only
-       reject simple variables that are not strings, or an array that does
-       not have a valid structure to be used as a callback. The valid ones
-       are supposed to have only 2 entries, the first of which is an object
+       <parameter>value</parameter> might be a function or method. It will
+       reject any values that are neither callable objects, strings nor arrays
+       with a valid structure to be used as a callback. A valid callable array
+       is supposed to have only 2 entries, the first of which is an object
```
Girgias, line 45:
```suggestion
       with a valid structure to be used as a callback. A valid callable array
       has 2 entries, the first of which is an object
```

reference/var/functions/is-callable.xml
kamil-tekiela: Precisely. Do you have any other ideas on how to deal with it?

reference/var/functions/is-callable.xml
```diff
@@ -17,7 +17,8 @@
    <methodparam choice="opt"><type>string</type><parameter role="reference">callable_name</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Verify that a value is a <type>callable</type>.
+   Verifies that a value is <type>callable</type>, or that it can be called using the
```
kamil-tekiela, line 20:
```suggestion
   Verifies that a value is a <type>callable</type>, or that it can be called using the
```

kamil-tekiela: @Girgias I think it should have "a" now as it did before, right?

Girgias:
```suggestion
   Verifies that <parameter>value</parameter> is <type>callable</type>, or that it can be called using the
```
Maybe?

Girgias (approved):

PR #4149, opened 2024-11-25
abstract.xml Fix the error in the last example and CS

mmalferov:
The only thing I'm not sure about is whether one empty line is required between the hooks that contain the implementation. With an empty string, readability seems to be better.

If we imagine that hooks are the same as methods, then, perhaps, it is better to leave one empty string between the "methods" :)

In this case, it might be better to add one empty line between hooks without implementation

### DIFF
--- a/language/oop5/abstract.xml
+++ b/language/oop5/abstract.xml
@@ -106,13 +106,13 @@ FOO_ConcreteClass2
 
 abstract class AbstractClass
 {
-    // Our abstract method only needs to define the required arguments
+    // An abstract method only needs to define the required arguments
     abstract protected function prefixName($name);
 }
 
 class ConcreteClass extends AbstractClass
 {
-    // Our child class may define optional parameters which are not present in the parent's signature
+    // A child class may define optional parameters which are not present in the parent's signature
     public function prefixName($name, $separator = ".")
     {
         if ($name == "Pacman") {
--- a/language/oop5/abstract.xml
+++ b/language/oop5/abstract.xml
@@ -112,7 +112,7 @@ abstract class AbstractClass
 
 class ConcreteClass extends AbstractClass
 {
-    // Our child class may define optional parameters not in the parent's signature
+    // Our child class may define optional parameters which are not present in the parent's signature
     public function prefixName($name, $separator = ".")
     {
         if ($name == "Pacman") {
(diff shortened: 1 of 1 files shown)


language/oop5/abstract.xml
```diff
@@ -96,32 +103,34 @@ FOO_ConcreteClass2
...
     abstract protected function prefixName($name);
-
 }
 
 class ConcreteClass extends AbstractClass
 {
-
-    // Our child class may define optional arguments not in the parent's signature
-    public function prefixName($name, $separator = ".") {
+    // Our child class may define optional parameters not in the parent's signature
```
kamil-tekiela, line 115:
```suggestion
    // Our child class may define optional parameters which are not present in the parent's signature
```
What is "our" referring to here, do you know?

kamil-tekiela (commented):
Was the error only s/arguments/parameters?

language/oop5/abstract.xml
mmalferov: I would remove the word 'our'

kamil-tekiela:
> > Was the error only s/arguments/parameters?
> 
> No, it's not the `s`, it's the terms. It is impossible to define arguments in the method signature, arguments are the actual values that come to the method from the outside. Only parameters are allowed to be defined in the method body.
> 
> I will allow myself one quote there:
> 
> > The two terms are used interchangeably in conversation, but they mean different things. You pass arguments when calling a method. And you use parameters within the method body.
> > As most if not all programming languages are copy-by-value, the two are equivalent, only differing in the context. The two terms exist to distinguish whether you are talking about a method definition or a method call. When you're calling a method, you pass in arguments. When you're actually in the method body, you are using parameters. Arguments get copied-by-value to parameters
> 
> Try to think of arguments as actually non-named values that are sent to the method, and of parameters as an integral part of a function or method, and you will no longer be able to say that "optional arguments have been defined in the method" :)

FYI https://phoenixnap.com/kb/sed-replace

What I meant was: was swaping the word arguments for the word parameters the only meaningful change in this PR. I understand why you did it, I just wanted to make sure I didn't miss anything more important.

mmalferov:
Oh, yes, the `sed`, now everything is clear. And I also thought: why is the sentence so strangely composed :))

kamil-tekiela (approved):
LGTM!

Issue #4151, opened 2024-11-25
Object Interfaces: Is the property really virtual?


mmalferov:
`Example #1 Interface properties example` in the [Properties section](https://www.php.net/manual/en/language.oop5.interfaces.php#language.oop5.interfaces.properties) says that `$writeable` property is virtual:

> This example creates a virtual property, which is fine

But the `ReflectionProperty` says that not.

```php
<?php

interface I
{
    public string $readable { get; }
    public string $writeable { set; }
    public string $both { get; set; }
}

class C2 implements I
{
    private string $written = '';
    private string $all = '';

    public string $readable { get => strtoupper($this->writeable); }

    // The interface only requires the property be settable,
    // but also including get operations is entirely valid.
    // This example creates a virtual property, which is fine    →    is it correct?
    public string $writeable {
        get => $this->written;
        set => $value; // NB! A property is virtual if neither its get nor set hook references the
                       // property itself using exact syntax)
    }

    public string $both {
        get => $this->all;
        set => strtoupper($value);
    }
}

$c2 = new C2();
$reflector = new ReflectionProperty($c2, 'writeable');
var_dump($reflector->isVirtual()); // Returns false
```

mmalferov:
Thanks, I created a PR https://github.com/php/doc-en/pull/4152

PR #4152, opened 2024-11-25
interfaces.xml Make the class property truly virtual

mmalferov:
In the footsteps of the issue: https://github.com/php/doc-en/issues/4151

### DIFF
--- a/language/oop5/interfaces.xml
+++ b/language/oop5/interfaces.xml
@@ -154,7 +154,6 @@ class C2 implements I
     // This example creates a virtual property, which is fine.
     public string $writeable {
         get => $this->written;
-
         set {
             $this->written = $value;
         }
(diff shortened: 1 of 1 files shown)


language/oop5/interfaces.xml
```diff
@@ -154,7 +154,10 @@ class C2 implements I
     // This example creates a virtual property, which is fine.
     public string $writeable {
         get => $this->written;
-        set => $value;
+
```
iluuu1994, line 157: Nit: I would prefer to minimize spaces.

iluuu1994 (approved):
Thank you! Btw, `$both` also looks suspicious. It only returns `$all`, but never writes to it. @Crell Can you clarify what the intention was here?

language/oop5/interfaces.xml
mmalferov: Do you mean the empty line between the get and set hooks?

iluuu1994: Yes

Crell:
Ha!  That example was copied directly from the RFC, so, oops?  Thanks for the catch.

PR #4153, opened 2024-11-25
Fix typo in example code.

Crell:
I've also made the same correction in the original RFC, where this was copied from.


iluuu1994 (approved):
Thanks!

PR #4155, opened 2024-11-25
On Regexp manual page, revise wording, "described below" -> "described in this section"

tiffany-taylor:
https://www.php.net/manual/en/regexp.introduction.php

When viewing the page from php.net, "described below" is awkward phrasing when there is no content below from which is described.

I haven't proposed changes in a while to the manual, which is why I'm making this a PR instead of pushing directly.


kamil-tekiela (approved):
Thanks.

PR #4156, opened 2024-11-25
Synchronize manual with stubs for PHP 8.4 - part 1

kocsismate:
first batch

### DIFF
--- a/reference/intl/collator/set-strength.xml
+++ b/reference/intl/collator/set-strength.xml
@@ -13,14 +13,14 @@
    &style.oop;
   </para>
   <methodsynopsis role="Collator">
-   <modifier>public</modifier> <type>bool</type><methodname>Collator::setStrength</methodname>
+   <modifier>public</modifier> <type>true</type><methodname>Collator::setStrength</methodname>
    <methodparam><type>int</type><parameter>strength</parameter></methodparam>
   </methodsynopsis>
   <para>
    &style.procedural;
   </para>
   <methodsynopsis>
-   <type>bool</type><methodname>collator_set_strength</methodname>
+   <type>true</type><methodname>collator_set_strength</methodname>
    <methodparam><type>Collator</type><parameter>object</parameter></methodparam>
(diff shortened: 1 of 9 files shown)


reference/intl/collator/set-strength.xml
```diff
@@ -13,14 +13,14 @@
    &style.oop;
   </para>
   <methodsynopsis role="Collator">
-   <modifier>public</modifier> <type>bool</type><methodname>Collator::setStrength</methodname>
+   <modifier>public</modifier> <type>true</type><methodname>Collator::setStrength</methodname>
```
cmb69, line 16: I think we need a changelog entry here.

reference/intl/locale/set-default.xml
```diff
@@ -13,14 +13,14 @@
    &style.oop;
   </para>
   <methodsynopsis role="Locale">
-   <modifier>public</modifier> <modifier>static</modifier> <type>bool</type><methodname>Locale::setDefault</methodname>
+   <modifier>public</modifier> <modifier>static</modifier> <type>true</type><methodname>Locale::setDefault</methodname>
```
cmb69, line 16: This might do without changelog entry, since the documentation already states that the method/function returns `true`.

cmb69 (commented):
Thank you!

reference/intl/collator/set-strength.xml
kocsismate: I checked these bool -> true return type changes before submitting the PR, and it looked like that `false` was only possible due to ZPP failures, so I skipped adding the changelog entries for them.

kocsismate: Yes, I've just found an actual evidence: https://github.com/php/php-src/commit/969e7a3c8b7ee3cc51502f1ff1e751ad71fd049c#diff-92deb6d32b945b2dbc99d97b472578f68656739ec6425be0b6bc70ff6316745cL123

cmb69 (approved):

PR #4158, opened 2024-11-26
Filter: Rewrite function docs and remove now useless section

Girgias:
TODO: Document properly return value of filter_*_array() functions


Girgias:
So I found some bugs in ext/filter while trying to figure out the behaviour... so I am going to ask for a review as is while I figure out the bugs and stuff

nielsdos:
I'm not really up to speed with how ext-filter works so I'm gonna wait for cmb, unless he doesn't get to it.

cmb69 (commented):
Looks generally good to me, but I think there either needs to be an example for `filter_input_array()` (or at least a link to the example of `filter_var_array()`). The API is so confusing, that without example the explanations are probably barely understandable.

And don't get me started about `FILTER_NULL_ON_FAILURE` …

Girgias:
Yeah, I will add those at one point, but will open a new issue to track this.

This whole extension is an API nightmare IMHO...

reference/filter/functions/filter-input-array.xml
```diff
@@ -18,147 +19,59 @@
...
+       The content of the superglobal that is being filtered is the original
+       "raw" content provided by the <acronym>SAPI</acronym>,
+       prior to any user modification to the superglobal.
+       To filter a modified superglobal use
+       <function>filter_var_array</function> instead.
+      </simpara>
+     </warning>
+    </listitem>
+   </varlistentry>
+   <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('function.filter-var-array')/db:refsect1[@role='parameters']//db:varlistentry[db:term[db:parameter[text()='options']]]/.)">
```
kamil-tekiela, line 43: I don't think the descriptions should be identical. They are similar but different. For example, the `$array` param doesn't exist here.

Girgias: But the `$array` param is not mentionned there?

kamil-tekiela: This is what I mean 
image

PR #4164, opened 2024-11-27
ext/bcmath: The `scale` parameter was made into a common component.

SakiTakamachi:
Since will also use it for `BcMath\Number`, will make it common.


Girgias (changes requested):

nielsdos:
Agreed, this PR shouldn't be done

SakiTakamachi:
I understand, thanks

PR #4165, opened 2024-11-27
gmp-cmp.xml Specify specific return values

mmalferov:


reference/gmp/functions/gmp-cmp.xml
```diff
@@ -41,8 +41,8 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns a positive value if <literal>a &gt; b</literal>, zero if
-   <literal>a = b</literal> and a negative value if <literal>a &lt;
+   Returns a <literal>1</literal> if <literal>a &gt; b</literal>, <literal>0</literal> if
+   <literal>a = b</literal> and a <literal>-1</literal> if <literal>a &lt;
    b</literal>.
   </para>
```
Girgias, line 47: This is incorrect, depending on the backend it might not return a normalized value, and this will only be guaranteed in PHP next.

Girgias (changes requested):

PR #4167, opened 2024-11-27
Mention `final` modifier for methods coming from traits

alexandre-daubois:
Part of #2796


language/oop5/traits.xml
```diff
@@ -545,6 +545,36 @@ class ConstantsExample {
   </example>
  </sect2>
 
+ <sect2 xml:id="language.oop5.traits.final-methods">
+  <title>Final methods</title>
+  <para>
+   As of PHP 8.3.0, the <literal>final</literal> modifier can
+   be applied to methods coming from traits.
+  </para>
```
Girgias, line 553: Maybe the final should also link to the final documentation page?
```suggestion
  <simpara>
   As of PHP 8.3.0, the <literal>final</literal> modifier can
   be applied to methods coming from traits.
  </simpara>
```

Girgias (approved):
Thank you, minor XML nit and one remark about a link but LGTM to me otherwise

language/oop5/traits.xml
alexandre-daubois: Thanks! What the rule to decide wether `para` or `simpara` should be used?

alexandre-daubois: Got it, thank you!

PR #4169, opened 2024-11-27
[PHP 8.3] Emphasize arbitrary expressions in static initializer

alexandre-daubois:
Part of #2796

Already present, but I think this could be emphasized a bit rather than just having a code snippet that says it's correct without further explanation in example 9 of https://www.php.net/manual/en/language.variables.scope.php.

### DIFF
--- a/language/variables.xml
+++ b/language/variables.xml
@@ -533,10 +533,10 @@ function foo(){
     This means that static variables in methods now behave the same way as static properties.
    </para>
 
-   <para>
+   <simpara>
     As of PHP 8.3.0, static variables can be initialized with arbitrary expressions.
     This means that method calls, for example, can be used to initialize static variables.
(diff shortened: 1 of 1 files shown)


language/variables.xml
```diff
@@ -533,6 +533,11 @@ function foo(){
     This means that static variables in methods now behave the same way as static properties.
    </para>
 
+   <para>
+    As of PHP 8.3.0, static variables can be initialized with arbitrary expressions.
+    This means that method calls, for example, can be used to initialize static variables.
+   </para>
```
Girgias, line 539:
```suggestion
   <simpara>
    As of PHP 8.3.0, static variables can be initialized with arbitrary expressions.
    This means that method calls, for example, can be used to initialize static variables.
   </simpara>
```

PR #4170, opened 2024-11-27
[PHP 8.4] add constant posix_sysconf documentation.

jdkfx:
The constant for the parameter in [posix_sysconf](https://www.php.net/manual/en/function.posix-sysconf.php
) has not been added.


devnexen (approved):
Looking good thanks !

PR #4179, opened 2024-11-27
Synchronize manual with stubs for PHP 8.4 - part 3

kocsismate:


cmb69 (commented):
Thank you!

cmb69 (approved):

reference/dir/directory.xml
```diff
@@ -85,6 +86,12 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        This class is now <literal>final</literal>.
+       </entry>
+      </row>
```
Girgias, line 94: Incorrect this is part of https://wiki.php.net/rfc/directory-opaque-object which is PHP 8.5.

Please use a worktree and the stubs from PHP-8.4 to do these changes.

reference/zip/ziparchive.xml
```diff
@@ -480,6 +480,12 @@
      <type>int</type>
      <varname linkend="ziparchive.constants.er-not-allowed">ZipArchive::ER_NOT_ALLOWED</varname>
     </fieldsynopsis>
+    <fieldsynopsis>
+     <modifier>public</modifier>
+     <modifier>const</modifier>
+     <type>int</type>
+     <varname linkend="ziparchive.constants.er-truncated-zip">ZipArchive::ER_TRUNCATED_ZIP</varname>
+    </fieldsynopsis>
```
Girgias, line 488: This needs a description below, also available as of PHP 8.4.0. (refer back to the issue tracking these changes)

reference/soap/soapclient.xml
```diff
@@ -257,42 +257,42 @@
...
-    <varlistentry xml:id="soapclient.props.--last-response-headers">
+    <varlistentry xml:id="soapclient.props.last-response-headers">
      <term><varname>__last_response_headers</varname></term>
      <listitem>
       <para>
       </para>
      </listitem>
     </varlistentry>
-    <varlistentry xml:id="soapclient.props.--soap-fault">
+    <varlistentry xml:id="soapclient.props.soap-fault">
```
Girgias, line 295: Not sure if those soap changes are correct, I'm pretty sure you are breaking links with this rn.

Girgias (changes requested):

reference/zip/ziparchive.xml
kocsismate: The description has already been added previously :)

reference/dir/directory.xml
kocsismate: Ah, I forgot that this was voted for PHP 8.5. Problem with using an older branch when generating the manual pages is that gen_stub.php usually receives improvements in master only, therefore I usually just disregard the stub changes that are not yet relevant.

reference/zip/ziparchive.xml
kocsismate: https://www.php.net/manual/en/zip.constants.php#ziparchive.constants.er-truncated-zip for reference

Girgias: Okay so this item is actually done, surprising.

reference/soap/soapclient.xml
Girgias: Can you check if those IDs are used in a `linken` attribute or not? But in the same vein, you should remove the single hyphen from the constants below.

reference/dir/directory.xml
Girgias: But you could use the stub files from PHP-8.4 with master gen_stub, no?

reference/soap/soapclient.xml
kocsismate: I found no collision. In the same time, I removed the leading single hyphens.

Girgias (approved):

PR #4181, opened 2024-11-28
Fix PASSWORD_DEFAULT constant type

Girgias:
Also remove personalization while at it

### DIFF
--- a/reference/password/constants.xml
+++ b/reference/password/constants.xml
@@ -177,7 +177,7 @@
    <varlistentry xml:id="constant.password-default">
     <term>
      <constant>PASSWORD_DEFAULT</constant>
-     (<type>mixed</type>)
+     (<type>string</type>)
     </term>
     <listitem>
      <para>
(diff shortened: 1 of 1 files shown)


reference/password/constants.xml
```diff
@@ -186,22 +186,16 @@
...
-      hash can change. Therefore, if you use <constant>PASSWORD_DEFAULT</constant>
-      you should store the resulting hash in a way that can store more than 60
-      characters (255 is the recommended width).
-     </para>
-     <para>
-      Values for this constant:
+      It is worth noting that over time this constant can change.
+      Therefore, it is important to be aware that the length of the resulting
+      hash can change.
+      Therefore, when using <constant>PASSWORD_DEFAULT</constant> the resulting
```
cmb69, line 192: I don't like the "Therefore" twice. Maybe use "Thus" or "So" for the second sentence.

cmb69 (approved):
Thank you! See wording nit below; otherwise looks good.

PR #4183, opened 2024-11-28
[PHP 8.3] Check variance of class constants

alexandre-daubois:
Part of #2796

### DIFF
--- a/language/oop5/constants.xml
+++ b/language/oop5/constants.xml
@@ -121,9 +121,41 @@ Fatal error: Uncaught Error: Cannot access private const Foo::BAZ in …
   </para>
  </note>
  <example>
+  <title>Class constant visibility variance check, as of PHP 8.3.0</title>
+  <programlisting role="php">
+<![CDATA[
+<?php
+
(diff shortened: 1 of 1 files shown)


language/oop5/constants.xml
```diff
@@ -120,10 +120,42 @@ Fatal error: Uncaught Error: Cannot access private const Foo::BAZ in …
...
+<![CDATA[
+Fatal error: Access level to MyClass::VALUE must be public (as in interface MyInterface) …
+]]>
+  </screen>
+ </example>
+ <note>
+  <simpara>
+   As of PHP 8.3.0 visibility variance is checked more strictly. Prior to this
+   version, the visibility of a class constant could be different from the
+   visibility of the constant in the implemented interface.
```
Girgias, line 152: Semantic line break
```suggestion
   As of PHP 8.3.0 visibility variance is checked more strictly.
   Prior to this version, the visibility of a class constant could be different
   from the visibility of the constant in the implemented interface.
```

PR #4188, opened 2024-11-29
Update traits.xml Mention the hierarchy of classes

mmalferov:
Static properties defined in the trait are shared only between classes of the same hierarchy. Maybe we should mention this?

### DIFF
--- a/language/oop5/traits.xml
+++ b/language/oop5/traits.xml
@@ -425,23 +425,27 @@ Example::doSomething();
     <title>Static Properties</title>
     <caution>
      <simpara>
-      Prior to PHP 8.3.0, static properties were shared across all classes
-      using the trait. As of PHP 8.3.0, each class using the trait has its
+      Prior to PHP 8.3.0, static properties were shared across a single hierarchy of classes
+      which using the trait. As of PHP 8.3.0, each class using the trait has its
       own copy of the static property.
      </simpara>
     </caution>
     <programlisting role="php">
      <![CDATA[
 <?php
-trait StaticExample {
+
+trait StaticExample
+{
     public static $static = 'foo';
 }
 
-class Example {
+class Example
+{
     use StaticExample;
 }
 
 echo Example::$static;
+
 ?>
 ]]>
     </programlisting>
@@ -458,16 +462,20 @@ echo Example::$static;
     <programlisting role="php">
 <![CDATA[
 <?php
-trait PropertiesTrait {
+
+trait PropertiesTrait
+{
     public $x = 1;
 }
 
-class PropertiesExample {
+class PropertiesExample
+{
     use PropertiesTrait;
 }
 
-$example = new PropertiesExample;
+$example = new PropertiesExample();
 $example->x;
+
 ?>
 ]]>
     </programlisting>


Girgias:
@alexandre-daubois do you have any opinion about this?

language/oop5/traits.xml
```diff
@@ -425,23 +425,27 @@ Example::doSomething();
     <title>Static Properties</title>
     <caution>
      <simpara>
-      Prior to PHP 8.3.0, static properties were shared across all classes
-      using the trait. As of PHP 8.3.0, each class using the trait has its
+      Prior to PHP 8.3.0, static properties were shared across a single hierarchy of classes
+      which using the trait. As of PHP 8.3.0, each class using the trait has its
```
alexandre-daubois, line 429:
```suggestion
      Prior to PHP 8.3.0, static properties were shared across a single hierarchy of classes
      which use the trait. As of PHP 8.3.0, each class using the trait has its
```

alexandre-daubois (commented):
Indeed! I did a little check: https://3v4l.org/ekFKM

It works **only if** the child class **also** uses the trait. If it doesn't (e.g. remove `use T;` from B), the instance is shared. It should be worth mentioning this specificity?

language/oop5/traits.xml
```diff
@@ -425,23 +425,27 @@ Example::doSomething();
...
       own copy of the static property.
```
alexandre-daubois, line 430: Given my latest comment:

```suggestion
      own copy of the static property if it explicitly uses the trait.
```

Or something equivalent

language/oop5/traits.xml
```diff
@@ -556,26 +566,37 @@ class ConstantsExample {
   <title>Final methods</title>
   <simpara>
    As of PHP 8.3.0, the <link linkend="language.oop5.final">final</link>
-   modifier can be applied to methods coming from traits.
+   modifier can be applied to methods coming from traits by same syntax with the <literal>as</literal> operator
+   which modifies traits' methods visibility in classes.
```
Girgias, line 570: Shouldn't it be this?
```suggestion
   which modifies trait's methods visibility in classes.
```

language/oop5/traits.xml
```diff
@@ -425,23 +425,29 @@ Example::doSomething();
...
     <caution>
      <simpara>
-      Prior to PHP 8.3.0, static properties were shared across all classes
-      using the trait. As of PHP 8.3.0, each class using the trait has its
-      own copy of the static property.
+      Prior to PHP 8.3.0, static properties defined in a trait were shared across a single hierarchy of classes
+      which use the trait. As of PHP 8.3.0, each class using the trait has its
+      own copy of the static property if it explicitly uses the trait,
+      i.e. the static property defined in a trait inserted into a child class override
+      a static property that the child class inherited from the parent class.
```
alexandre-daubois, line 432: As a non-native speaker, I struggle a bit to understand the sentence. Maybe it could be cut in two?

alexandre-daubois (commented):
Just one nitpick comment, otherwise it's great! Thank you for taking care of this followup 🙂

language/oop5/traits.xml
alexandre-daubois: Better to me, yes!

Issue #4175, opened 2024-11-25
str_getcsv() and friends should not explicitly tell about "optional" parameters


8ctopus:
### Description

The following code:

```php
<?php

$csv = <<<CSV
1,30,2,29
CSV;

$columns = str_getcsv($csv, ',', '', '\\');

var_dump($columns);
```

Resulted in this output in php 8.3.10:
```
array(4) {
  [0] =>
  string(1) "1"
  [1] =>
  string(2) "30"
  [2] =>
  string(1) "2"
  [3] =>
  string(2) "29"
}
```

Now in php 8.4.1, I get this:
```txt
Fatal error: Uncaught ValueError: str_getcsv(): Argument php/php-src#3 ($enclosure) must be a single character in K:\dev\github\php-nano-csv\test
.php on line 7

ValueError: str_getcsv(): Argument php/php-src#3 ($enclosure) must be a single character in K:\dev\github\php-nano-csv\test.php on line 7

Call Stack:
    0.4199     403080   1. {main}() K:\dev\github\php-nano-csv\test.php:0
    0.4200     403112   2. str_getcsv($string = '1,30,2,29', $separator = ',', $enclosure = '', $escape = '\\') K:\dev\github\php-nano
-csv\test.php:7
```

I saw that the behavior of the function was changed, not to permit an empty enclosure, however who are we meant to deal with unenclosed csv?

### PHP Version

PHP 8.4.1

### Operating System

Windows 10

PR #4190, opened 2024-11-29
Improve csv functions by mentioning that enclosure character will be ignore

8ctopus:
See https://github.com/php/doc-en/issues/4175

### DIFF
--- a/reference/filesystem/functions/fgetcsv.xml
+++ b/reference/filesystem/functions/fgetcsv.xml
@@ -77,6 +77,7 @@
       <para>
        The <parameter>enclosure</parameter> parameter sets the field enclosure character.
        It must be a single byte character.
+       If the CSV does not use an enclosure character, the parameter will be ignored.
       </para>
      </listitem>
     </varlistentry>


reference/filesystem/functions/fgetcsv.xml
```diff
@@ -77,6 +77,7 @@
       <para>
        The <parameter>enclosure</parameter> parameter sets the field enclosure character.
        It must be a single byte character.
+       If the CSV does not use an enclosure character, the parameter will be ignored.
```
cmb69, line 80: I don't think this sentence is useful information. I mean, if the CSV does not use a separator (i.e. only one column), the separator will also be "ignored".

8ctopus: I added this sentence, because while upgrading a package from php 8.3 to 8.4, I was getting this error:

```
ValueError: str_getcsv(): Argument #3 ($enclosure) must be a single character
```

The code that triggers the error is:

```php
$columns = str_getcsv($line, $this->separator, $this->enclosure, $this->escape);
```

So I started scratching my head and went back to reading the php doc which says:



So the current doc says you must set an enclosure character. Maybe for you it's obvious, that you can set it to any character, and, if the character is not present, it will be ignored. However it was not for me and I had to disturb the dev team by opening an issue on GitHub.

cmb69: Yeah, I'm aware of the background. :)

But like I've said, in the general case you cannot know whether enclosure characters are used (they are needed whenever the separator character or a line break are contained in a field value). And I've just checked latest LibreOffice Calc, and when importing a CSV, you need to choose between double- or single-quotes as enclosure character, even if the file is tab separated (and so rarely uses enclosure characters).

8ctopus: Do you think adding the extra string will do harm?

PR #4192, opened 2024-11-30
Document fallback value syntax for ini variables (closes #4191)

### LINKED ISSUE #4191: Fallback value syntax for ini variables

jimwins:


install/ini.xml
```diff
@@ -88,20 +88,23 @@
...
+   defined.
+  </simpara>
+  <example>
+   <title>&php.ini; Environment Variables</title>
+    <programlisting role="ini">
 <![CDATA[
 ; PHP_MEMORY_LIMIT is taken from environment
 memory_limit = ${PHP_MEMORY_LIMIT}
+; If PHP_MAX_EXECUTION_TIME is not defined, it will fall back to 30
+max_execution_time = ${PHP_MAX_EXECUTION_TIME!30}
```
nielsdos, line 104: Are you sure  this syntax is right? I thought it was with a colon and dash, and looking at the implementation PR seems to support that idea: https://github.com/php/php-src/pull/11351/files#diff-5deefd0bdbb9aac682cb7a131d2086fd5e9cabe8b5b8f4f9a864659cf26bfec3R148-R166

jimwins: Ah, I was looking at the original commit and not the merged PR, which used `!`.

https://github.com/php/php-src/commit/bc8b9aedf69ccec250c6dd759a8aca9fb709aec9

nielsdos (approved):
Thanks!

PR #4194, opened 2024-11-30
Closures created from magic functions handle named params (closes #4193)

### LINKED ISSUE #4193: Closures created from magic functions accept named arg

jimwins:


nielsdos:
I guess it should say closures instead of anonymus functions?

jimwins:
> I guess it should say closures instead of anonymus functions?

We aren't super consistent in our language in this section, but it leads with "Anonymous functions, also known as closures" and the other changelog entry calls them anonymous functions, so I stuck with that.

nielsdos (approved):
Eh whatever
they're not the same but I don't care

jimwins:
Fair enough, I went ahead and changed the changelog entry, but maybe we should open a new issue about cleaning up the distinction between anonymous functions and closures in this section.

Girgias:
That whole section needs to be rewritten, and split into multiple files instead of being a behemot of a file.

PR #4198, opened 2024-11-30
Document str_getcsv() and fgetcsv() change in 8.3.0 (closes #4197)

### LINKED ISSUE #4197: str_getcsv() and fgetcsv() now return an empty string instead of a string with a single null byte for the last field

jimwins:


nielsdos (approved):
Oof, I remember this bug :grimacing: 
Thanks for the docs update!

Girgias (approved):

PR #4199, opened 2024-11-30
Document when STREAM_NOTIFY_COMPLETED was implemented (closes #3813)

### LINKED ISSUE #3813: Streams can now emit the STREAM_NOTIFY_COMPLETED notification. This was previously not implemented.

jimwins:


reference/stream/constants.xml
```diff
@@ -791,6 +791,7 @@
     <listitem>
      <simpara>
       There is no more data available on the stream.
+      (Added as of PHP 8.3.0.)
```
nielsdos, line 794: No, this constant existed but never did anything

reference/stream/functions/stream-notification-callback.xml
```diff
@@ -100,6 +100,28 @@
...
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.3.0</entry>
+      <entry>
+       Support for <constant>STREAM_NOTIFY_COMPLETED</constant> added.
```
nielsdos, line 117: Maybe make clear that it actually does something now instead of the constant being unused?

PR #4203, opened 2024-11-30
Document change in how C void returns are handled (closes #4202)

### LINKED ISSUE #4202: C functions that have a return type of void now return null instead of returning the following object object(FFI\CData:void) { }

jimwins:
FFI is pretty sparsely documented as it is, so just a changelog entry here seems appropriate.


nielsdos (approved):
Right, I also see no better place to put this tbh

PR #4205, opened 2024-11-30
Document when proc_open() throws ValueError (closes #4204)

### LINKED ISSUE #4204: If using an array as the $command for proc_open(), it must now have at least one non empty element. Otherwise a ValueError is thrown.

jimwins:


reference/exec/functions/proc-open.xml
```diff
@@ -187,6 +187,15 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   As of PHP 8.3.0, throws a <classname>ValueError</classname> if
```
nielsdos, line 193:
```suggestion
   As of PHP 8.3.0, throws a <exceptionname>ValueError</exceptionname> if
```

reference/exec/functions/proc-open.xml
```diff
@@ -199,6 +208,14 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.3.0</entry>
+       <entry>
+        A <classname>ValueError</classname> will be thrown if
```
nielsdos, line 214:
```suggestion
        A <exceptionname>ValueError</exceptionname> will be thrown if
```

nielsdos (changes requested):

jimwins:
Argh, I had originally used `<exceptionname>` and then let other usages of `<classname>ValueError</classname>` convince me I was wrong.

nielsdos:
Yeah there's various "versions" of usages let's say

nielsdos (approved):

PR #4208, opened 2024-11-30
[PHP 8.4] pcntl update part 1

devnexen:


devnexen:
for #3872

Girgias (commented):
Please remove the wrapping `<para>` tags around the `<informaltable>, also it might make sense to have different changelog row for different parameters.

PR #4209, opened 2024-11-30
Synchronize manual with stubs for PHP 8.4 - part 6

kocsismate:


reference/sqlite3/sqlite3result/finalize.xml
```diff
@@ -25,10 +25,40 @@
...
   <para>
-   Returns &true;.
+   &return.true.always;.
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   An <exceptionname>Error</exceptionname> is thrown if the method is called with an uninitialized object.
```
cmb69, line 35: "with" or "on"? I think the latter is more appropriate here, but I'm fine with the wording if we already use it elsewhere.

cmb69 (commented):
Thank you!

reference/sqlite3/sqlite3result/finalize.xml
kocsismate: I'm not sure, but after reading the two alternatives, "on" sounds better for me too. @Girgias Do you have any preference?

Girgias: No real preference here, but  quick `git grep` seems to show that "called on" is more common for this situation.

reference/sqlite3/sqlite3result/finalize.xml
```diff
@@ -25,10 +25,40 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns &true;.
+   &return.true.always;.
```
Girgias, line 28:
```suggestion
   &return.true.always;
```

PR #4214, opened 2024-12-02
ext/bcmath: Avoid nesting includes

SakiTakamachi:
And added the title for error section (that forgot)

Related https://github.com/php/doc-en/pull/4163#issuecomment-2510000203

cc: @cmb69 @leonardolara

### DIFF
--- a/reference/bc/functions/bcdiv.xml
+++ b/reference/bc/functions/bcdiv.xml
@@ -52,8 +52,19 @@
  </refsect1>
 
  <refsect1 role="errors">
-  <!-- Include standard ValueErrors for num1, num2, and scale, this includes the title -->
-  <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('function.bcadd')/db:refsect1[@role='errors']/*)" />
+  &reftitle.errors;
+  <para>
+   This function throws a <exceptionname>ValueError</exceptionname> in the following cases:
+   <simplelist>
+    <member>
+     <parameter>num1</parameter> or <parameter>num2</parameter>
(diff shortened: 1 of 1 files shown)


alfsb:
We can please hold off merging this for today? I think there two issues at play here, and will need two separated general fix.

1. The current installed libxml2 we use does not expect recursive XIncludes
2. Missing or empty `<xi:fallback/>` leaves `<xi:include/>` or empty holes behind, and this fails valuation.

This explains why doc-en pass with libxml2 upgrade or looping `xinclude()`, and translations not: there is no XInclude fails on doc-en.

I think I can send a PR for doc-base fixing both issues at general level.

SakiTakamachi:
@alfsb 
Got it. It's already late in Japan, so there's no problem :)

Girgias:
Same I would rather not, `libxml` being buggy should not be the reason that we duplicate identical stuff and thus can have it out of sync.

nielsdos:
I agree with @Girgias . Although imo the real sin is distributions sticking with ancient versions, which is a pain for both users and me.

alfsb:
> I think I can send a PR for doc-base fixing both issues at general level.

Opened PR 194, linked above, that may supersedes this. Pass all languages on CI (that are building before), and seems to work well in Windows and Linux. @SakiTakamachi , please test it, to see if it fixes the manual building, without reverting the changes.

SakiTakamachi:
@alfsb 
I was able to test your patch and confirm that it works perfectly
Thanks!

SakiTakamachi:
So, I close this PR :)

PR #4221, opened 2024-12-02
Improve the example of mysqli_warning_count

kamil-tekiela:
I don't think we can provide an example that would produce a warning in both strict and non-strict mode so an example that generates a warning in strict mode is better as this is pretty much the standard. It also makes the example much simpler. The code style is aligned with the rest of mysqli examples. 

cc @Girgias


Girgias (approved):
This makes sense to me

PR #4223, opened 2024-12-02
Remove personalization from mysqli_warning_count

kamil-tekiela:


kamil-tekiela:
Ooops, I am too late.

PR #4224, opened 2024-12-02
Use CDATA tags for content of screen tags

Girgias:
Hopefully this fixes the weird empty line in https://www.php.net/manual/en/install.pecl.pear.php


jimwins (approved):
LGTM.

PR #4227, opened 2024-12-02
Synchronize manual with stubs for PHP 8.4 - part 8

kocsismate:


reference/xml/functions/xml-parse-into-struct.xml
```diff
@@ -9,7 +9,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>xml_parse_into_struct</methodname>
+   <type class="union"><type>int</type><type>false</type></type><methodname>xml_parse_into_struct</methodname>
```
kocsismate, line 12: These signatures were forgotten to be updated when the relevant parameter descriptions were changed.

reference/pgsql/functions/pg-select.xml
```diff
@@ -75,7 +75,8 @@
       <listitem>
        <para>
         An <type>array</type> whose keys are field names in the table <parameter>table_name</parameter>,
-        and whose values are the conditions that a row must meet to be retrieved.
+        and whose values are the conditions that a row must meet to be retrieved.  As of PHP 8.4.0,
+        if an empty array is provided, no conditions will apply.
```
cmb69, line 79: Maybe add that the function failed for an empty array, previously.

reference/pgsql/functions/pg-select.xml
```diff
@@ -132,6 +133,12 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        <parameter>conditions</parameter> is now nullable.
```
cmb69, line 139: Doesn't look nullable; it's just that the parameter is now optional.

cmb69 (commented):
Thanks for your continued work on this!

Besides the two issues below, this looks good to me.

reference/pgsql/functions/pg-select.xml
kocsismate: Ah no, I wanted to write "optional" :)

reference/pgsql/functions/pg-select.xml
kocsismate: I hope the added sentence doesn't sound very bad

reference/pgsql/functions/pg-select.xml
```diff
@@ -75,7 +75,9 @@
       <listitem>
        <para>
         An <type>array</type> whose keys are field names in the table <parameter>table_name</parameter>,
-        and whose values are the conditions that a row must meet to be retrieved.
+        and whose values are the conditions that a row must meet to be retrieved.  As of PHP 8.4.0,
+        if an empty array is provided, no conditions will apply.  Previously, the function failed with
+        an empty <parameter>conditions</parameter> argument.
```
Girgias, line 80: It is better to use semantic line breaks, especially here as it will just show as new lines for translations.
```suggestion
        and whose values are the conditions that a row must meet to be retrieved.
        As of PHP 8.4.0, when an empty array is provided, no conditions will apply.
        Previously, the function failed with an empty <parameter>conditions</parameter> argument.
```

Girgias (approved):

PR #4231, opened 2024-12-03
[PHP 8.3] `opcache.preload_user` is not required when run as root

alexandre-daubois:
Part of #2796

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -962,6 +962,8 @@
       startup as root before switching to an unprivileged system user. Preloading as root is not
       allowed by default for security reasons, unless this directive is explicitly set to
       <literal>root</literal>.
+      As of PHP 8.3.0, this directive does not need to be set in order to allow preloading while
+      running as root when using the &cli.sapi; or <link linkend="book.phpdbg">phpdbg SAPI</link>.
      </para>
     </listitem>
    </varlistentry>


arnaud-lb (approved):
Thank you!

reference/opcache/ini.xml
```diff
@@ -962,6 +962,13 @@
       startup as root before switching to an unprivileged system user. Preloading as root is not
       allowed by default for security reasons, unless this directive is explicitly set to
       <literal>root</literal>.
+
+      <note>
+       <simpara>
+        As of PHP 8.3.0 and when using the CLI or phpdbg SAPIs, the directive isn't required
+        by the preloading to be set when running as root.
```
jimwins, line 969:
```suggestion
        As of PHP 8.3.0, this directive does not need to be set in order to allow preloading while
        running as root when using the &cli.sapi; or <link linkend="book.phpdb">phpdbg SAPI</link>.
```

Maybe we don't even need to bump this out as a `<note>` but just add the sentence to the preceding paragraph.

jimwins (changes requested):

reference/opcache/ini.xml
```diff
@@ -961,7 +961,9 @@
       Lets the preloading to be run as the specified system user. This is useful for servers that
       startup as root before switching to an unprivileged system user. Preloading as root is not
       allowed by default for security reasons, unless this directive is explicitly set to
-      <literal>root</literal>.
+      <literal>root</literal>. As of PHP 8.3.0, this directive does not need to be set in order
+      to allow preloading while running as root when using the &cli.sapi; or
+      <link linkend="book.phpdb">phpdbg SAPI</link>.
```
jimwins, line 966: You should start the sentence on a line of its own. (It's a convention called "[semantic line breaks](https://sembr.org/)" that should help with keeping down formatting churn for translations.)

LGTM with that change.

jimwins (approved):

reference/opcache/ini.xml
alexandre-daubois: I wasn't aware this had a name. Thank you Jim, it's updated!

PR #4232, opened 2024-12-03
[PHP 8.3] `FFI::load` in preload scripts

alexandre-daubois:
Part of #2796

### DIFF
--- a/reference/ffi/ffi/load.xml
+++ b/reference/ffi/ffi/load.xml
@@ -56,6 +56,31 @@
   </para>
  </refsect1>
 
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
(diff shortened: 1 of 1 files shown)


reference/ffi/ffi/load.xml
```diff
@@ -56,6 +56,33 @@
...
+        <link linkend="opcache.preloading">preload scripts</link> when the
+        current system user is the same as the one defined in the
+        <literal>opcache.preload_user</literal> configuration directive.
+       </entry>
+      </row>
+     </tbody>
+    </tgroup>
+   </informaltable>
+  </para>
+ </refsect1>
```
Girgias, line 84: Please remove the useless wrapping `<para>` tag.

alexandre-daubois: I copied one of the last occurrences that contains it, no luck 😄 Should a follow-up PR be created to remove the remaining useless `para` in changelogs?

jimwins (approved):

reference/ffi/ffi/load.xml
Girgias: No, as that would affect translations for "no good reason". It is something that might be good to do at one point, but we need some tooling to do this automatically so that it can be replicated across translations. :)

PR #4233, opened 2024-12-03
Clarify that it represents midnight in the local timezone, not UTC.

AlexandrePlus:
Explicitly specify "local midnight" to avoid confusion. Unix timestamps represent seconds since 1970-01-01 00:00:00 UTC, and without this clarification, users might assume the timestamp corresponds to UTC midnight. In fact, the timestamp (UTC) corresponds to 00:00:00 when converted to the local timezone.


kamil-tekiela (approved):
MSTM

PR #4235, opened 2024-12-03
[PHP 8.3] `ValueError`'s previous exception is now set in `password_hash()`

alexandre-daubois:
Part of #2796

### DIFF
--- a/reference/password/functions/password-hash.xml
+++ b/reference/password/functions/password-hash.xml
@@ -188,6 +188,16 @@
      </thead>
      <tbody>
       <row>
+       <entry>8.3.0</entry>
+       <entry>
+        <function>password_hash</function> now sets the underlying
+        <exceptionname>Random\RandomException</exceptionname> as the
+        <property>Exception::$previous</property> exception when a
+        <exceptionname>ValueError</exceptionname> is thrown due to a failure
+        in the salt generation.
+       </entry>
+      </row>
+      <row>
        <entry>8.0.0</entry>
        <entry>
         <function>password_hash</function> no longer returns &false; on failure, instead a


reference/password/functions/password-hash.xml
```diff
@@ -187,6 +187,14 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.3.0</entry>
+       <entry>
+        <function>password_hash</function> now sets the underlying
+        <classname>\Random\RandomException</classname> as the previous exception when
+        a <classname>ValueError</classname> is thrown if the salt generation fails.
```
jimwins, line 195: I think both `<classname>` be `<exceptionname>`. Also:

```suggestion
        a <classname>ValueError</classname> is thrown because the salt generation fails.
```

alexandre-daubois: Updated!

jimwins: Ah, one other thing is I think you don't need the leading `\` on the exception name or it won't get linked up properly. I haven't tested this, though.

reference/password/functions/password-hash.xml
```diff
@@ -187,6 +187,14 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.3.0</entry>
+       <entry>
+        <function>password_hash</function> now sets the underlying
+        <exceptionname>\Random\RandomException</exceptionname> as the previous exception when
+        a <exceptionname>ValueError</exceptionname> is thrown because the salt generation fails.
```
kamil-tekiela, line 195:
```suggestion
        <function>password_hash</function> now sets the underlying
        <exceptionname>\Random\RandomException</exceptionname> as the previous exception when a 
        <exceptionname>ValueError</exceptionname> is thrown due to a failure in salt generation.
```

reference/password/functions/password-hash.xml
TimWolla: > Ah, one other thing is I think you don't need the leading `\` on the exception name or it won't get linked up properly. I haven't tested this, though.

Indeed.

reference/password/functions/password-hash.xml
```diff
@@ -187,6 +187,14 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.3.0</entry>
+       <entry>
+        <function>password_hash</function> now sets the underlying
+        <exceptionname>\Random\RandomException</exceptionname> as the previous exception when
```
TimWolla, line 194: Should `previous` perhaps be a `<literal>$previous</previous>`?

TimWolla (commented):
The description correctly describes the change that happened. I'll defer the English phrasing to other reviewers.

reference/password/functions/password-hash.xml
Girgias: I would prefer `<property>Exception::$previous</property>` then

Girgias (approved):

TimWolla (approved):

PR #4236, opened 2024-12-03
Add missing type for DATE_* constants

Girgias:


alexandre-daubois (commented):
Can't approve in this repo but I approve 🙂 Goes in pair with #4237

Should we put parentheses around the type? It's not done at some other places

jimwins (approved):
LGTM.

PR #4239, opened 2024-12-03
Change output of ArrayObject

kamil-tekiela:
This changed in PHP 5.3 so I think it's time to update it.

### DIFF
--- a/reference/spl/arrayobject/exchangearray.xml
+++ b/reference/spl/arrayobject/exchangearray.xml
@@ -56,8 +56,8 @@ $fruitsArrayObject = new ArrayObject($fruits);
 
 // Now exchange fruits for locations
 $old = $fruitsArrayObject->exchangeArray($locations);
-print_r($old);
-print_r($fruitsArrayObject);
+var_dump($old);
+var_dump($fruitsArrayObject);
 
 ?>
 ]]>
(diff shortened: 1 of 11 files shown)


reference/spl/arrayobject/exchangearray.xml
```diff
@@ -74,9 +74,13 @@ Array
...
-    [1] => Paris
-    [2] => London
+    [storage:ArrayObject:private] => Array
+        (
+            [0] => Amsterdam
+            [1] => Paris
+            [2] => London
+        )
+
 )
```
Girgias, line 84: I would prefer to change the example code to use `var_dump` then at the same time.

reference/spl/arrayobject/natcasesort.xml
```diff
@@ -83,23 +83,28 @@ print_r($arr2);
...
-    [3] => img2.png
+    [storage:ArrayObject:private] => Array
+        (
+            [0] => IMG0.png
+            [5] => IMG3.png
+            [4] => img1.png
+            [2] => img10.png
+            [1] => img12.png
+            [3] => img2.png
+        )
```
Girgias, line 94: Ditto

reference/spl/arrayobject/setflags.xml
```diff
@@ -97,6 +97,7 @@ var_dump($fruitsArrayObject->lemons);
     &example.outputs;
```
Girgias, line 97:
```suggestion
    &example.outputs.similar;
```

Girgias (commented):
I'm fine with this, but I prefer the use of `var_dump()` over `print_r()` as that output is not ideal IMHO.

Girgias (approved):
Thank you!

PR #4253, opened 2024-12-05
Improve filter_input_array return section

kamil-tekiela:
I have to admit I don't know if this is correct either, but it seems to make more sense. I don't know when this function actually returns false. I also don't understand under which circumstances the value of an item will be false or null.

### DIFF
--- a/reference/filter/functions/filter-input-array.xml
+++ b/reference/filter/functions/filter-input-array.xml
@@ -55,17 +55,13 @@
    On success, an &array; containing the values of the requested variables.
   </simpara>
   <simpara>
-   On failure, &false; is returned.
-   Except if the failure is that the input array designated by
-   <parameter>type</parameter> is not populated where &null; is returned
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
I don't know how to proceed here. I can't provide an accurate description because I can't understand how it's supposed to work without accurate documentation. 

I believe this function returns false only when the filter value provided is not a valid filter, except it if it is provided as a filter in an array in which case it returns an array with the requested key and value false. Except if the FILTER_NULL_ON_FAILURE flag is set and the value is null. But if the actual filter fails the item's value is always null, unless it is requested as an array in which case it's false on failure without the FILTER_NULL_ON_FAILURE flag. This whole thing is a confusing mess and I can't figure out what's a bug and what's an intended behaviour.

Girgias:
Yes, this extension is a confusing mess. I will try and document it when I get round to properly testing it so I can bugfix it.

PR #4263, opened 2024-12-06
Fill `Memcached::__construct()` parameters description

### LINKED ISSUE #4025: Document `Memcached::__construct()` parameters
This is a follow up of #3873 

The `callback` and `connection_str` parameters are not yet documented.

https://github.com/php/doc-en/blob/54eaf4ed9edc37074870a13dc0de605df4285c3b/reference/memcached/memcached/construct.xml#L40-55

alexandre-daubois:
Fix #4025

### DIFF
--- a/reference/memcached/memcached/construct.xml
+++ b/reference/memcached/memcached/construct.xml
@@ -40,16 +40,25 @@
     <varlistentry>
      <term><parameter>callback</parameter></term>
      <listitem>
-      <para>
-       <!-- TODO Document constructor params -->
-      </para>
+      <methodsynopsis>
+       <type>void</type><methodname><replaceable>callback</replaceable></methodname>
+       <methodparam><type>Memcached</type><parameter>memcached</parameter></methodparam>
+       <methodparam><type class="union"><type>string</type><type>null</type></type><parameter>persistent_id</parameter></methodparam>
+      </methodsynopsis>
+      <simpara>
+       The <parameter>callback</parameter> parameter is called when the
+       connection is established. It should be a valid PHP <type>callable</type>
+       that receive the <classname>Memcached</classname> object as its first
+       parameter, and <parameter>persistent_id</parameter> as its second.
+      </simpara>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><parameter>connection_str</parameter></term>
      <listitem>
       <para>
-       <!-- TODO Document constructor params -->
+       This parameter is used to pass additional connection options to
+       the memcache servers, such as a server weight in a cluster.
       </para>
      </listitem>
     </varlistentry>


reference/memcached/memcached/construct.xml
```diff
@@ -41,15 +41,19 @@
      <term><parameter>callback</parameter></term>
      <listitem>
       <para>
-       <!-- TODO Document constructor params -->
+       The <parameter>callback</parameter> parameter is called when the
+       connection is established. It should be a valid PHP <type>callable</type>
+       that receive the <classname>Memcached</classname> object as its first
+       parameter, and <parameter>persistent_id</parameter> as its second.
```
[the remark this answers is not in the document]
alexandre-daubois: Alright, I tried to guess the callback signature from https://github.com/php-memcached-dev/php-memcached/blob/bfcd9e7bbac40693cc055adc5f8722ef13909e38/php_memcached.c#L807

reference/memcached/memcached/construct.xml
```diff
@@ -40,16 +40,25 @@
     <varlistentry>
      <term><parameter>callback</parameter></term>
      <listitem>
-      <para>
-       <!-- TODO Document constructor params -->
-      </para>
+      <methodsynopsis>
+       <type>void</type><methodname><replaceable>callback</replaceable></methodname>
+       <methodparam><type>Memcached</type><parameter>memcached</parameter></methodparam>
+       <methodparam><type>string</type><parameter>persistent_id</parameter></methodparam>
```
Girgias, line 46: Looking at the linked source code this should be `string|null`

alexandre-daubois: Ah yes, I missed the `if`. Not a pro of PHP extensions and C, but I would say you're right 😄 Updated

reference/memcached/memcached/construct.xml
```diff
@@ -40,16 +40,25 @@
     <varlistentry>
      <term><parameter>callback</parameter></term>
      <listitem>
-      <para>
-       <!-- TODO Document constructor params -->
-      </para>
+      <methodsynopsis>
+       <type>void</type><methodname><replaceable>callback</replaceable></methodname>
+       <methodparam><type>Memcached</type><parameter>memcached</parameter></methodparam>
+       <methodparam><type>string|null</type><parameter>persistent_id</parameter></methodparam>
```
philip, line 46: We have fun markup for this, so rather than `<type>string|null</type>` here use `<type class="union"><type>string</type><type>null</type></type>` instead.

jordikroon: @alexandre-daubois would you be willing to address this last comment? If not I can do the change for you if you prefer.

alexandre-daubois: Totally forgot about it! Let me address this. Sorry for the delay!

alexandre-daubois:
Ready! Sorry it took that much time

jordikroon:
Don't worry about it. Thank you!

PR #4264, opened 2024-12-06
Adds a note about hooks + constructor property promotion

weierophinney:
Per a conversation with @crell, added a note to detail behavior of hooks
when the set hook allows more types than the property, and is used with
constructor property promotion.

### DIFF
--- a/language/oop5/property-hooks.xml
+++ b/language/oop5/property-hooks.xml
@@ -199,6 +199,69 @@ class Example
    On a backed property, omitting a <literal>get</literal> or<literal>set</literal>
    hook means the default read or write behavior will be used.
   </simpara>
+  <note>
+   <simpara>
+    Hooks can be defined when using <link linkend="language.oop5.decon.constructor.promotion">constructor
+    property promotion</link>. However, when doing so, values provided
+    to the constructor must match the type associated with the property,
+    regardless of what the <literal>set</literal> hook might allow.
+   </simpara>
+   <simpara>
+    Consider the following:
+   </simpara>
+   <programlisting role="php">
+<![CDATA[
+class Example
+{
+    public function __construct(
+        public private(set) DateTimeInterface $created {
+            set (string|DateTimeInterface $value) {
+                if (is_string($value)) {
+                    $value = new DateTimeImmutable($value);
+                }
+                $this->created = $value;
+            }
+        },
+    ) {
+    }
+}
+]]>
+   </programlisting>
+   <simpara>
+    Internally, the engine decomposes this to the following:
+   </simpara>
+   <programlisting role="php">
+<![CDATA[
+class Example
+{
+    public private(set) DateTimeInterface $created {
+        set (string|DateTimeInterface $value) {
+            if (is_string($value)) {
+                $value = new DateTimeImmutable($value);
+            }
+            $this->created = $value;
+        }
+    }
+
+    public function __construct(
+        DateTimeInterface $created,
+    ) {
+        $this->created = $created;
+    }
+}
+]]>
+   </programlisting>
+   <simpara>
+    Any attempts to set the property outside the constructor will
+    allow either <literal>string</literal> or <literal>DateTimeInterface</literal>
+    values, but the constructor will only allow <literal>DateTimeInterface<literal>,
+    as that is the type associated with the property.
+   </simpara>
+   <simpara>
+    If you need to allow this kind of behavior from the constructor, you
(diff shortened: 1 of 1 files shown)


language/oop5/property-hooks.xml
```diff
@@ -199,6 +199,69 @@ class Example
...
+        $this->created = $created;
+    }
+}
+]]>
+   </programlisting>
+   <simpara>
+    Any attempts to set the property outside the constructor will
+    allow either <literal>string</literal> or <literal>DateTimeInterface</literal>
+    values, but the constructor will only allow <literal>DateTimeInterface<literal>,
+    as that is the type associated with the property.
```
kamil-tekiela, line 258: Did you mean "parameter" instead of "property"

language/oop5/property-hooks.xml
```diff
@@ -199,6 +199,69 @@ class Example
...
+   </simpara>
+   <simpara>
+    If you need to allow this kind of behavior from the constructor, you
```
kamil-tekiela, line 261: Why is this a separate paragraph?

language/oop5/property-hooks.xml
```diff
@@ -199,6 +199,69 @@ class Example
    On a backed property, omitting a <literal>get</literal> or<literal>set</literal>
    hook means the default read or write behavior will be used.
   </simpara>
+  <note>
+   <simpara>
+    Hooks can be defined when using <link linkend="language.oop5.decon.constructor.promotion">constructor
+    property promotion</link>. However, when doing so, values provided
+    to the constructor must match the type associated with the property,
+    regardless of what the <literal>set</literal> hook might allow.
```
Girgias, line 207: Link on its own line
```suggestion
    Hooks can be defined when using
    <link linkend="language.oop5.decon.constructor.promotion">constructor property promotion</link>.
    However, when doing so, values provided
    to the constructor must match the type associated with the property,
    regardless of what the <literal>set</literal> hook might allow.
```

language/oop5/property-hooks.xml
```diff
@@ -199,6 +199,69 @@ class Example
...
+    ) {
+        $this->created = $created;
+    }
+}
+]]>
+   </programlisting>
+   <simpara>
+    Any attempts to set the property outside the constructor will
+    allow either <literal>string</literal> or <literal>DateTimeInterface</literal>
+    values, but the constructor will only allow <literal>DateTimeInterface<literal>,
```
Girgias, line 257:
```suggestion
    allow either <type>string</type> or <interfacename>DateTimeInterface</interfacename>
    values, but the constructor will only allow <interfacename>DateTimeInterface<interfacename>,
```

language/oop5/property-hooks.xml
Girgias, line 261: The manual style guide bans personalization (i.e. usage of "you").

Girgias (changes requested):

language/oop5/property-hooks.xml
weierophinney: That's the crux of the problem: it's both.

The property has a defined type of `DateTimeInterface`, but it's _set_ hook allows either string or `DateTimeInterface`. Because it's defined using constructor property promotion, the property type is used for the parameter type, regardless of what the set hook allows.

language/oop5/property-hooks.xml
weierophinney: To call out the action needed by the developer.

weierophinney: Fixed

language/oop5/property-hooks.xml
weierophinney: I've pushed an update to make this more clear.

language/oop5/property-hooks.xml
kamil-tekiela: Well, there should be no "you" there to begin with. It's also a continuation of the previous thought so it should be in the same paragraph.

Crell:
The text looks good to me, but there's some linting issues.  Once those are resolved I think we can merge this.

weierophinney:
I _think_ I've resolved the linting issues, but the build chain does not work currently (I get errors on `make` detailing `detected dubious ownership in repository at '/var/www/doc-base'`). Will need somebody to approve the workflow to validate.

Crell:
All clear now.  @Girgias do you want to clear your change request and merge?

Girgias (approved):

PR #4269, opened 2024-12-08
Fix inaccurate caution note on pack()

### LINKED ISSUE #4266: Inaccurate caution note on pack()
From https://www.php.net/manual/en/function.pack.php:

> Note that PHP internally stores [int](https://www.php.net/manual/en/language.types.integer.php) values as signed values of a machine-dependent size (C type long).

No it doesn't. The C type long is 32 bits on 64 bit Windows, even though PHP's int is 64 bits on 64 bit Windows. I think it would be good to just drop the "C type long" part.

Ankush1oo8:
Removed wrong caution From https://www.php.net/manual/en/function.pack.php

Fixes #4266

### DIFF
--- a/reference/misc/functions/pack.xml
+++ b/reference/misc/functions/pack.xml
@@ -256,7 +256,7 @@ $binarydata = pack("nvc*", 0x1234, 0x5678, 65, 66);
   &reftitle.notes;
   <caution>
    <para>Note that PHP internally stores <type>int</type> values as
-    signed values of a machine-dependent size.   
+    signed values of a machine-dependent size.
     Integer literals and operations that yield numbers outside the bounds of the  
     <type>int</type> type will be stored as <type>float</type>.
     When packing these floats as integers, they are first cast into the integer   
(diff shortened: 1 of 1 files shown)


nielsdos (approved):
I think this is sufficient. Explaining what the corresponding C type is would be cumbersome and below we already talk about 32/64 bit int.
So I think this is fine. Thanks!

reference/misc/functions/pack.xml
```diff
@@ -256,7 +256,7 @@ $binarydata = pack("nvc*", 0x1234, 0x5678, 65, 66);
   &reftitle.notes;
   <caution>
    <para>Note that PHP internally stores <type>int</type> values as
-    signed values of a machine-dependent size (C type <literal>long</literal>).   
+    signed values of a machine-dependent size.
```
nielsdos, line 259: Looks like there was some trailing whitespace here.
```suggestion
    signed values of a machine-dependent size.
```

PR #4270, opened 2024-12-08
fix: resolve PHP 8.4 deprecation warnings for getReturnType()

fabyo0:
Aligned `ezcReflectionReturnInfo::getReturnType()` with PHP 8.4 by adding the `?ReflectionType` return type. This ensures compatibility with `ReflectionFunctionAbstract::getReturnType()` and resolves deprecation warnings. Alternatively, applied `#[\ReturnTypeWillChange]` for temporary suppression.

### DIFF
--- a/language/oop5/traits.xml
+++ b/language/oop5/traits.xml
@@ -26,7 +26,7 @@
 <![CDATA[
 <?php
 trait ezcReflectionReturnInfo {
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Thanks! 

However, I analysed the example and I have several questions about it which means that this PR is probably not enough. We probably need a new example here.

PR #4276, opened 2024-12-08
Add warning about fseek()'s return values

### LINKED ISSUE #4255: fseek() documentation is confusing and incomplete, and fseek() return codes are backwards
# My experience learning fseek() from the documentation

From manual page: https://php.net/function.fseek

This is my first time using `fseek()`, so it's definitely possible I don't know what I'm doing, or am misreading the documentation.

Documentation says:

    offset
    The offset.
    
    To move to a position before the end-of-file, you need to pass a negative value in offset and set whence to SEEK_END.
    
    whence
    whence values are:
    
    SEEK_SET - Set position equal to offset bytes.
    SEEK_CUR - Set position to current location plus offset.
    SEEK_END - Set position to end-of-file plus offset.

It seems to be telling me to do this to seek 2KB before the end of the file:

`fseek($File_Handle, -2048, SEEK_END);`

However, this always returns `FALSE` for me. 

The `fopen()` before this produces no error.
The file is 1.8MB, so there is definitely data to seek in.
I can read from the file with `fgets()` or `fread()` just fine.

How do I seek to 2KB before the end of a large file?

Also, why is the word "wence" being used here?
That is horrifically confusing wording for this context.  I had to read that section 4 times to have a clue what it was even talking about.

Trying to debug this issue... trying a bunch more things out now.

These commands also fail for me:

    fseek($File_Handle, -2048, SEEK_END);
    fseek($File_Handle, 2048, SEEK_END);
    fseek($File_Handle, 0, SEEK_END);
    fseek($File_Handle, 0);

But very strangely, this command works, but starts me at position `0`:

    fseek($File_Handle, -2048);

I am at a loss for how I can be getting results like these, they seem completely arbitrary.  No warnings, notices, or errors are being thrown, the function is just returning `FALSE`.

This command shows a normal file:

    fstat($File_Handle);

It has `[size] => 1822525`

## Problem now understood

Okay, figured out the problem.  `fstat()` is not returning `FALSE`, it is returning `0` which is evaluating to `FALSE`.

Why in the world does `fseek()` return `0` (`FALSE`) on success, and `-1` (`TRUE`) on failure?

That is nonsensical and broken.

## Improvements to documentation

**At the very least update the documentation to specify this entirely crazy behaviour exists!**

This is definitely _not_ enough documentation:

    Upon success, returns 0; otherwise, returns -1.

The user should be informed that this function's return values are completely backwards from the expected return format of every other function in PHP.

This is way worse than all those functions that return `FALSE` on errors and `0` or above on success -- because PHP documentation actually mentions that special case in an info box: "`This function may return FALSE on error but may also return 0 that evaluates to boolean FALSE`", or whatever the exact wording is.

**This function needs a large warning that it returns backwards return values from what is expected of PHP functions, and an explanation of _WHY_ it chose to do things in this backwards way.**

The basic error checking code I was using (like I always use) fails because of this absurdity:

    if (!fseek($File_Handle, 0)) { die("Could not seek in file."); }

What a royal waste of time...

kamil-tekiela:
Closes #4255

### DIFF
--- a/reference/filesystem/functions/fseek.xml
+++ b/reference/filesystem/functions/fseek.xml
@@ -47,9 +47,9 @@
        The offset.
       </para>
       <para>
-       To move to a position before the end-of-file, you need to pass
-       a negative value in <parameter>offset</parameter> and
-       set <parameter>whence</parameter>
+       To move to a position before the end-of-file,
+       a negative value in the <parameter>offset</parameter> must be passed and
+       the <parameter>whence</parameter> parameter must be set
        to <constant>SEEK_END</constant>.
       </para>
      </listitem>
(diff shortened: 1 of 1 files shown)


Girgias (approved):

Girgias:
Do other I/O functions behave that way?

kamil-tekiela:
> Do other I/O functions behave that way?

I thought they did, but I can't find another example of it.

PR #4279, opened 2024-12-09
The constant classes cannot be available before the introduction of the class

Girgias:
Found while reviewing doc-fr PRs


kamil-tekiela (approved):
Makes sense but it's going to be confusing to many people.

Girgias:
Overloading PDO was already quite confusing :/

PR #4283, opened 2024-12-09
Fix: Add explicit $escape parameter in fputcsv for PHP 8.4 compatibility

fabyo0:
This commit resolves a deprecation warning in PHP 8.4 regarding the $escape parameter of the fputcsv function. The warning indicates that the default value for $escape will change in future PHP versions, requiring it to be explicitly provided.

* Changes Made

Updated the fputcsv calls to include the $escape parameter explicitly with a backslash ('\\') as the value.

* Benefits

Ensures compatibility with PHP 8.4.
 Prevents potential future errors caused by the deprecation of the default $escape parameter value.

This update makes the script more robust and ready for future PHP versions.

### DIFF
--- a/reference/filesystem/functions/fputcsv.xml
+++ b/reference/filesystem/functions/fputcsv.xml
@@ -127,19 +127,20 @@
 <![CDATA[
 <?php
 
-$list = array (
-    array('aaa', 'bbb', 'ccc', 'dddd'),
-    array('123', '456', '789'),
-    array('"aaa"', '"bbb"')
-);
+$list = [
+    ['aaa', 'bbb', 'ccc', 'dddd'],
+    ['123', '456', '789'],
+    ['"aaa"', '"bbb"']
+];
 
 $fp = fopen('file.csv', 'w');
 
 foreach ($list as $fields) {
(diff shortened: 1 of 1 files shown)


reference/filesystem/functions/fputcsv.xml
```diff
@@ -127,16 +127,16 @@
...
+    ['aaa', 'bbb', 'ccc', 'dddd'],
+    ['123', '456', '789'],
+    ['"aaa"', '"bbb"']
+];
 
 $fp = fopen('file.csv', 'w');
 
 foreach ($list as $fields) {
-    fputcsv($fp, $fields);
+    fputcsv($fp, $fields, ',', '"', '\\');
```
Girgias, line 139: If the output is the same, please use the following:
```suggestion
    fputcsv($fp, $fields, ',', '"', '');
```

I would rather promote disabling the problematic escaping mechanism than using `\\`

Girgias (changes requested):

Girgias (approved):

Girgias:
Thank you!

PR #4286, opened 2024-12-10
Integrate SPL language-snippets and rearrange SPL docs

Girgias:
They are only used once, having them as entities is useless.

Leaving this as a draft as I might decide to reorganize SPL first to get rid of the "Miscellaneous" page


cmb69 (approved):
Thank you! Small nit below; otherwise looks good to me (nice to see use-once entities to be removed!)

PR #4287, opened 2024-12-10
Idempotent build: PDO madness

alfsb:
```
$ cd phpdoc/en
$ grep -rF '.PDO.' .
$ grep -rF '.pdo.' .
(lots)
$ ls ./reference/pdo_sqlite
configure.xml  pdo  PDO  pdo-sqlite.xml  reference.xml  versions.xml
```
See de `pdo` and `PDO`? Imagine that on Windows...

This took me half of a day, and gave me very bad flashbacks from early 2000... I'm seriously Seriously SERIOUSLY considering .git ignoring any `PDO` directory where there exists an `pdo` directory.


alfsb:
This breaks the build because:
```
$ grep -F '.pdo.' file-entities.ent | wc
     75     300    8984
$ grep -F '.PDO.' file-entities.ent | wc
     16      64    2132
```

I think it is time to take the bullet and lower case all PDO dirs and entity names, across all languages.

Girgias:
I would rather renamie the `PDO` to `pdo_overloaded` as this is how the "methods" described there work.

alfsb:
I will open a issue, for someone take this later on. I was only testing idempotent2 on Windows when I found out this other case.

PR #4115, opened 2024-11-20
Document new PDO classes and methods

Girgias:
Reference: https://wiki.php.net/rfc/pdo_driver_specific_subclasses and https://wiki.php.net/rfc/pdo_driver_specific_parsers

Still needs MySQLi, PGSQL, and OCI8 (although that one is going to be low priority)

@SakiTakamachi can you document, either as inline comments or in a follow-up PR the transaction attributes?

PR #4288, opened 2024-12-10
Rename folder that contain overloaded PDO methods

Girgias:
@alfsb this should fix the issue


alfsb:
Thanks. I was almost at my wits end with these renames. I only found out a few minutes ago that this *also* is affected by file-entities. I merge doc-base first.

This may need a heads uip on all translations, as it probally will break all (and also futher confuse translations that are done on Windows)

alfsb:
I will merge others idempotent first. And I just fixed building on Windows (at least on my machine), so will be possible to start from a building state.

As bad this error can be, it' 20 somewhat old situation... It's maybe better to wait for 8.4 changes sloes down.

alfsb:
Confirmed that changes merged today builds ok CI *and* Windows. So it's possible to start building state, in principle.

> I can take care of translations, just let me know.

I ask to wait for next week, to allow a cooldown for big changes merged today (doc-base 201/202) and one planned for Thursday (doc-base 198). And to allow translations recover from `reference.spl.entities.countable`. I will do some tests in the meanwhile.

Starting next week, please, take care of translations. I can't even imagine how many man-*years* were wasted dealing with the problems caused by these similarly named folders.

Girgias:
> Confirmed that changes merged today builds ok CI _and_ Windows. So it's possible to start building state, in principle.
> 
> > I can take care of translations, just let me know.
> 
> I ask to wait for next week, to allow a cooldown for big changes merged today (doc-base 201/202) and one planned for Thursday (doc-base 198). And to allow translations recover from `reference.spl.entities.countable`. I will do some tests in the meanwhile.
> 
> Starting next week, please, take care of translations. I can't even imagine how many man-_years_ were wasted dealing with the problems caused by these similarly named folders.

The new `pdo` sub folders got introduced with https://github.com/php/doc-en/pull/4115 so it is not a new thing. And I was aware of the potential Windows issues, but to have things in a namespace and `gen_stub` not complain it needed to be in `pdo/`.

Arguably, this _is_ a PHP 8.4 related change.

alfsb:
> Arguably, this _is_ a PHP 8.4 related change.

Ah, I was remembering, rather vividly, when a similar thing happened back in PHP 5 days. Since these are new, this explains something I wasn't able to understand: how file-entities, old or new, would be able to generate different names for paths that don't exist in Windows.

I only managed to make the manual compilable on Windows with this [ugly hack](https://github.com/php/doc-base/commit/df18cd280ca4cb21ee0cacad8e4f43f571829388#diff-6c87a6e32d355c274d829beb3d1f6461d84cb02a709af1d8c9a604a5afe631c4R165), already scheduled to be removed in an issue.

PR #4296, opened 2024-12-12
Fix ext/dom constant aliases

kocsismate:
I noticed failures when I tried to run `./build/gen_stub.php --verify-manual ./ ../doc-en/`, and it turned out that the newly added ext/dom constant aliases cause this. I found no way to add support for the `DOM_INDEX_SIZE_ERR / Dom\INDEX_SIZE_ERR ` format in gen_stub.php (gen_stub.php can only replace/fix constants one-by-one), so I had to duplicate the constant definitions in the manual.

Since the constant IDs would clash in most of the cases, I had to add an extra `dom-` prefix to them. It may be possible to fix these via phd and use `.` as separator instead of `-` in case of namespaces (?).

### DIFF
--- a/reference/dom/constants.xml
+++ b/reference/dom/constants.xml
@@ -302,7 +302,7 @@
      </row>
      <row xml:id="constant.dom-index-size-err">
       <entry>
-       <constant>DOM_INDEX_SIZE_ERR</constant> / <constant>Dom\INDEX_SIZE_ERR</constant>
+       <constant>DOM_INDEX_SIZE_ERR</constant>
        (<type>int</type>)
       </entry>
       <entry>1</entry>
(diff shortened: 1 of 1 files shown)


Girgias:
Known when reviewing the PR: https://github.com/php/doc-en/pull/4212#discussion_r1865855179

I don't think duplicating the test is useful, or at minimum use XIncludes so they do not get out of sync.

For the linking aspect I would prefer to have @haszi's input, considering they handled the whole constant linking thing.

Girgias:
I just thought of the good solution, which is to convert the table to a `<variablelist>` and have multiple `<term>` elements. Will do that soonish.

haszi:
> For the linking aspect I would prefer to have @haszi's input, considering they handled the whole constant linking thing.

All the linking needs is an `xml:id` so the simplest way to do this is to use one `<term>` / `<entry>` tag per constant and add an `xml:id` to the alias's tag. This will work as long as we make sure PhD renders those IDs.

E.g.:
```xml
     <row xml:id="constant.dom-index-size-err">
      <entry>
       <constant>DOM_INDEX_SIZE_ERR</constant>
       (<type>int</type>)
      </entry>
      <entry xml:id="dom.constants.index-size-err">
       <constant>Dom\INDEX_SIZE_ERR</constant>
       (<type>int</type>)
      </entry>
```

Girgias:
> @haszi Thank you for the hint! @Girgias Are you still planning to convert the table to a list?

Still planned, but feel free to do @haszi suggestion in the meantime :)

kocsismate:
Closing as per our conversation in https://github.com/php/doc-en/pull/5516#issuecomment-4398149617

PR #4297, opened 2024-12-12
Synchronize manual with stubs for PHP 8.4 - part 10

kocsismate:
Adding docs for the `StreamBucket` class.

### DIFF
--- a/reference/stream/functions/stream-bucket-make-writeable.xml
+++ b/reference/stream/functions/stream-bucket-make-writeable.xml
@@ -12,7 +12,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type class="union"><type>object</type><type>null</type></type><methodname>stream_bucket_make_writeable</methodname>
+   <type class="union"><type>StreamBucket</type><type>null</type></type><methodname>stream_bucket_make_writeable</methodname>
    <methodparam><type>resource</type><parameter>brigade</parameter></methodparam>
   </methodsynopsis>
   <para>
@@ -55,17 +55,52 @@
     </varlistentry>
     <varlistentry>
      <term>
-      <property>datalen</property>
-      (<type>integer</type>)
+      <property>dataLength</property>
+      (<type>int</type>)
      </term>
      <listitem>
       <para>
-       <parameter>datalen</parameter> <literal>bucket</literal> The length of the string in the bucket.
+       <parameter>dataLength</parameter> <literal>bucket</literal> The length of the string in the bucket.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term>
+      <property>datalen</property>
+      (<type>int</type>)
+     </term>
+     <listitem>
+      <para>
+       <parameter>datalen</parameter> <literal>bucket</literal> The length of the string in the bucket. Deprecated as of PHP 8.4
+       in favor of <literal>StreamBucket::$dataLength</literal>.
       </para>
      </listitem>
     </varlistentry>
    </variablelist>
   </para>
+ </refsect1>
+
+ <refsect1 role="changelog">
+  &reftitle.changelog;
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       This function returns a <classname>StreamBucket</classname> instance now;
+       previously, an <classname>stdClass</classname> was returned.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
  </refsect1>
 
  <refsect1 role="seealso">
@@ -78,7 +113,6 @@
   </para>
  </refsect1>
 </refentry>
-
 <!-- Keep this comment at the end of the file
 Local variables:
 mode: sgml
--- a/reference/stream/streambucket.xml
+++ b/reference/stream/streambucket.xml
@@ -1,0 +1,123 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<reference xml:id="class.streambucket" role="class" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
+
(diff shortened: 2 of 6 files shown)


reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -53,21 +53,56 @@
...
+     <term>
+      <property>dataLength</property>
+      (<type>int</type>)
+     </term>
+     <listitem>
+      <para>
+       <parameter>dataLength</parameter> <literal>bucket</literal> The length of the string in the bucket.
+      </para>
+     </listitem>
+    </varlistentry>
```
Girgias, line 66: Can't we just remove the property description now? I would rather have something along the lines of:

> Prior to PHP 8.4.0, an `stdClass` containing the same properties as `StreamBucket` was returned.

Then duplicating documentation.

reference/stream/streambucket.xml
```diff
@@ -0,0 +1,123 @@
...
+ <section role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>stream_bucket_new</function></member>
+    <member><function>stream_bucket_append</function></member>
+    <member><function>stream_bucket_prepend</function></member>
+    <member><function>stream_bucket_make_writeable</function></member>
+   </simplelist>
+  </para>
```
Girgias, line 96: No useless wrapping para tags
```suggestion
  <simplelist>
   <member><function>stream_bucket_new</function></member>
   <member><function>stream_bucket_append</function></member>
   <member><function>stream_bucket_prepend</function></member>
   <member><function>stream_bucket_make_writeable</function></member>
  </simplelist>
```

reference/stream/streambucket.xml
```diff
@@ -0,0 +1,123 @@
...
+     <term>int <varname>dataLength</varname></term>
+     <listitem>
+      <para>The length of the string in the bucket.</para>
+     </listitem>
+    </varlistentry>
+   </variablelist>
+  </section>
+
+ <section role="seealso">
+  &reftitle.seealso;
```
Girgias, line 88: Indent is off

reference/stream/streambucket.xml
```diff
@@ -0,0 +1,123 @@
...
+   </classsynopsis>
+  </section>
+
+  <section xml:id="streambucket.props">
+   &reftitle.properties;
+   <variablelist>
+    <varlistentry xml:id="streambucket.props.bucket">
+     <term>resource <varname>bucket</varname></term>
+     <listitem>
+      <para>A <literal>userfilter.bucket</literal> resource.</para>
```
Girgias, line 60: Please use simpara now, and use new lines for the descriptions. Apply for the other properties.
```suggestion
      <simpara>
       A <literal>userfilter.bucket</literal> resource.
      </simpara>
```

reference/stream/streambucket.xml
```diff
@@ -0,0 +1,123 @@
...
+ <title>The StreamBucket class</title>
+ <titleabbrev>StreamBucket</titleabbrev>
+
+ <partintro>
+
+  <section xml:id="streambucket.intro">
+   &reftitle.intro;
+   <para>
+    A stream bucket is a chunk of stream which can be extracted from bucket brigades.
+   </para>
```
Girgias, line 14:
```suggestion
   <simpara>
    A stream bucket is a chunk of stream which can be extracted from bucket brigades.
   </simpara>
```

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -53,21 +53,56 @@
...
+      <entry>8.4.0</entry>
+      <entry>
+       This function returns a <classname>StreamBucket</classname> instance now;
+       previously, an <classname>stdClass</classname> was returned.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
+ </refsect1>
```
Girgias, line 104: Feels like the changelogs could just XInclude the one from `stream_bucket_append()` to prevent duplication.

Girgias (changes requested):

reference/stream/functions/stream-bucket-make-writeable.xml
cmb69: Scream-shot from the current docs:

bucket

What is the meaning of `data(len) bucket` there? This looks like pure noise.

reference/stream/streambucket.xml
cmb69, line 14:
```suggestion
   <simpara>
    A stream bucket is a chunk of a stream which can be extracted from bucket brigades.
   </simpara>
```

reference/stream/functions/stream-bucket-make-writeable.xml
kocsismate: Yeah, good point! I wasn't sure if we want to keep these docs or not, so I left it there. But I agree, these are just noise after adding the StreamBucket page.

reference/stream/streambucket.xml
kocsismate: What's the difference between the two?

reference/stream/functions/stream-bucket-make-writeable.xml
kocsismate: TBH I'm not sure it's worth the effort for only two places to include. 🤔

Girgias: Xincluding is not really more difficult than copy pasting.

reference/stream/functions/stream-bucket-make-writeable.xml
```diff
@@ -40,34 +40,33 @@
...
-      (<type>integer</type>)
-     </term>
-     <listitem>
-      <para>
-       <parameter>datalen</parameter> <literal>bucket</literal> The length of the string in the bucket.
-      </para>
-     </listitem>
-    </varlistentry>
-   </variablelist>
+   Returns a bucket object or &null;.
```
Girgias, line 43: When does it return `null` actually?

kocsismate: As far as I managed to understand https://github.com/php/php-src/blob/d6d78545ea405e89a44445a9a32d77a75fd8bb6b/ext/standard/user_filters.c#L356, it returns null if the brigade is empty (?). The 2nd part of the condition seems to be always non-null (`bucket` variable).

kocsismate:
@Girgias Is there anything else I should fix in order to merge this?

Girgias (approved):

PR #4301, opened 2024-12-13
Document touch() doesn't clear the stat cache.

Crell:
Thanks to @derickr for pointing this out to me.  It drove me batty for days. :smile:

### DIFF
--- a/reference/filesystem/functions/clearstatcache.xml
+++ b/reference/filesystem/functions/clearstatcache.xml
@@ -57,8 +57,9 @@
    <function>filegroup</function>,
    <function>fileowner</function>,
    <function>filesize</function>,
-   <function>filetype</function>, and
-   <function>fileperms</function>.
+   <function>filetype</function>,
+   <function>fileperms</function>, and.
+   <function>touch</function>.
   </para>
  </refsect1>
 
--- a/reference/filesystem/functions/touch.xml
+++ b/reference/filesystem/functions/touch.xml
@@ -135,6 +135,27 @@ if (!touch('some_file.txt', $time)) {
  <refsect1 role="notes">
   &reftitle.notes;
   &note.filesystem-time-res;
+  <note>
+   <simpara>
+    Updates to the <literal>mtime</literal> or <literal>atime</literal>
+    of a file will not be reflected immediately in filesystem operations
+    such as <function>filemtime</function>.  That's because the information
+    is cached, and this function does not automatically clear the cache.
+    To have changes reflected in the application immediately, call
+    <function>clearstatcache</function> on the file.
+   </simpara>
+  </note>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>clearstatcache</function></member>
+    <member><function>fileatime</function></member>
+    <member><function>filemtime</function></member>
+   </simplelist>
+  </para>
  </refsect1>
  
 </refentry>
--- a/reference/filesystem/functions/touch.xml
+++ b/reference/filesystem/functions/touch.xml
@@ -135,16 +135,7 @@ if (!touch('some_file.txt', $time)) {
  <refsect1 role="notes">
   &reftitle.notes;
   &note.filesystem-time-res;
-  <note>
-   <simpara>
-    Updates to the <literal>mtime</literal> or <literal>atime</literal>
-    of a file will not be reflected immediately in filesystem operations
-    such as <function>filemtime</function>.  That's because the information
-    is cached, and this function does not automatically clear the cache.
-    To have changes reflected in the application immediately, call
-    <function>clearstatcache</function> on the file.
-   </simpara>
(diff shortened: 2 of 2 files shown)


TimWolla (changes requested):
This is not unique about `touch()`. None of the file operations clear the stat cache:

```
<?php
var_dump(filemtime('/tmp/foo.txt'), filesize('/tmp/foo.txt'));
file_put_contents('/tmp/foo.txt', str_repeat('x', random_int(0, 999)));
clearstatcache();
var_dump(filemtime('/tmp/foo.txt'), filesize('/tmp/foo.txt'));
```

Crell:
Seems like we should add the same warning to other functions, then.  Is it worth making an entity out of it, or just copy-paste-tweak?

Girgias:
We already have the `note.clearstatcache` entity. Is that insufficient?

reference/filesystem/functions/touch.xml
```diff
@@ -135,6 +135,27 @@ if (!touch('some_file.txt', $time)) {
...
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
+    <member><function>clearstatcache</function></member>
+    <member><function>fileatime</function></member>
+    <member><function>filemtime</function></member>
+   </simplelist>
+  </para>
```
Girgias, line 158:
```suggestion
  &reftitle.seealso;
  <simplelist>
   <member><function>clearstatcache</function></member>
   <member><function>fileatime</function></member>
   <member><function>filemtime</function></member>
  </simplelist>
```

Crell:
What is insufficient is my awareness that entity already exists. :smile: Switched to using that, and added `touch` to the list on the `clearstatcache` page, which was missing.

reference/filesystem/functions/clearstatcache.xml
```diff
@@ -57,8 +57,9 @@
    <function>filegroup</function>,
    <function>fileowner</function>,
    <function>filesize</function>,
-   <function>filetype</function>, and
-   <function>fileperms</function>.
+   <function>filetype</function>,
+   <function>fileperms</function>, and.
+   <function>touch</function>.
```
TimWolla, line 62: This change is wrong. This list lists the functions that _read_ from the stat cache. `touch()` is a _setter_.

reference/filesystem/functions/touch.xml
```diff
@@ -135,6 +135,16 @@ if (!touch('some_file.txt', $time)) {
  <refsect1 role="notes">
   &reftitle.notes;
   &note.filesystem-time-res;
+  &note.clearstatcache;
```
TimWolla, line 138: And likewise is this change wrong.

reference/filesystem/functions/touch.xml
```diff
@@ -135,6 +135,16 @@ if (!touch('some_file.txt', $time)) {
...
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
```
TimWolla, line 142: This should probably list `filemtime` first, because the mtime is the relevant field that is modified by `touch`.

Crell: touch can modify both properties, so I listed them in alphabetical order.

reference/filesystem/functions/touch.xml
Crell: Then do we need a new entity for statcache impacting write functions?  Because the root issue is I didn't realize touch didn't reset the cache, so couldn't understand why my touch() call didn't seem to do anything.  So I want *something* on the touch page to indicate that.  You and @Girgias figure out what you want there, I've already made two proposals and both of them someone has objected to.

reference/filesystem/functions/clearstatcache.xml
Crell: The description says nothing about read/  It just says "Affected."

TimWolla: The context is:

> When you use stat(), lstat(), or any of the other functions listed in the affected functions list (below), PHP caches the information those functions return in order to provide faster performance.

`touch()` is not an affected function, because it does not interact with the stat cache at all. It does not (need to) know about it.

reference/filesystem/functions/touch.xml
TimWolla: > Then do we need a new entity for statcache impacting write functions?

None of them impact the stat*cache* (except for `unlink()`, which automatically clears it according to the manual). But virtually all of them impact the stat*data*, because writes implicitly affect the mtime (and usually also the file size).

Note that `exec()` and friends might also affect the stat*data*, because you could `exec('touch foo.txt')`. `Dom\HTMLDocument::saveHtml()` does as well.

I do not see a value-add in adding a reference to the stat*cache* to every function that possibly affects the stat*data* and as far as I can tell all the functions reading from the statcache have the note already.

lacatoire:
Closing as obsolete

PR #4303, opened 2024-12-13
Example for Socket class in PHP 8+ in socket.xml

KERNAMON:
Added a detailed example of creating and interacting with the Socket class introduced in PHP 8+

### DIFF
--- a/reference/sockets/socket.xml
+++ b/reference/sockets/socket.xml
@@ -32,7 +32,46 @@
 
  </partintro>
 
- <!-- &reference.sockets.entities.socket; -->
+ <!-- {{{ Example Section -->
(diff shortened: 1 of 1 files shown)


KERNAMON:
The current documentation lacks a concrete example for the Socket class and this example serves as a bridge, showing how to adapt to the new object-based implementation introduced in PHP 8+

Girgias:
This is not relevant at all.

There are existing examples, located in `reference/socket/examples.xml` that *might* need some modernization. But the fact this is now an opaque object instead of a resource doesn't change the fundamental way how one use the socket functions.

You cannot create a socket class unless calling one of the socket functions.

PR #4304, opened 2024-12-13
basic-syntax.xml Specify when it is better to skip closing PHP tags

mmalferov:
It seemed to me that limiting the skipping of closing PHP tags to only files that contain only PHP code is misleading. After all, we skip closing tags in files with mixed content too, and the condition for skipping is not whether the file consists only of PHP code, but whether the file ends with PHP code


language/basic-syntax.xml
```diff
@@ -53,7 +53,7 @@
    </para>
 
    <para>
-    If a file contains only PHP code, it is preferable to omit the PHP closing tag
+    If a file ends with PHP code, it is preferable to omit the PHP closing tag
     at the end of the file.  This prevents accidental whitespace or new lines
     being added after the PHP closing tag, which may cause unwanted effects
     because PHP will start output buffering when there is no intention from
```
Girgias, line 59: If the file is not just PHP code the following sentence doesn't make any sense.

mmalferov:
And still, the current wording doesn't let me sleep peacefully :)

The current documentation states:

> "If a file contains only PHP code, it is preferable to omit the PHP closing tag."

While this rule is valid, it places emphasis on whether a file contains **only PHP code** when deciding whether to omit the closing tag. However, this can lead to confusion in cases where files contain mixed content (PHP + HTML) but still end with PHP code.

#### Example

Consider the following file:

```php
<?php
$var = 42;
?>

<p>Hello, world!</p>

<?php
echo "This is PHP again.";
// Should we add a closing tag here?
```

Now, let’s analyze:

1. Does this file contain **only PHP code**? Clearly not—it contains both PHP and HTML.
2. Does this file meet the condition of containing "only PHP code" as described in the current documentation? No, it does not.
3. Should we include a closing `?>` tag at the end of this file simply because it contains **not only PHP code**?

#### Question

Does the current formulation remain accurate and precise in such cases? Or would it be better to revise the wording to clarify that:

- The decision to omit the closing tag depends not on whether the file contains **only PHP code**, but on whether the file **ends with PHP code**.
- This applies even to files with mixed content (PHP + HTML), as long as the last block is PHP.

PR #4313, opened 2024-12-15
Use same case for access.format "%T" in install.fpm.configuration.php 

danny50610:


bukka (approved):
This is correct change. Thanks.

bukka:
The failure is unrelated so merging.

PR #4320, opened 2024-12-16
add Xpass extension

remicollet:


remicollet:
chicken and egg issue with https://github.com/php/doc-base/pull/209

Build passes locally

PR #4328, opened 2024-12-19
Update unserialize.xml Fix typo

mmalferov:


mmalferov:
And shouldn't we change the type for the `allowed_classes` element to `array|bool` instead `mixed` in line 92?

sy-records (approved):

kamil-tekiela (approved):

kamil-tekiela:
> And shouldn't we change the type for the `allowed_classes` element to `array|bool` instead `mixed` in line 92?

I think we should. Could you please submit a new PR?

PR #4331, opened 2024-12-21
Modernize RPM docs

Girgias:
Use more appropriate DocBook tags, add constant tags which will handle automatic linking, use entities and literal tags, fix indentation, group related constants together.


philip:
@Girgias imho it can be pushed, what say you?

Girgias:
Yeah, worst case we can always PR and do more changes later.

PR #4332, opened 2024-12-21
Use "singleton type" instead of "value type"

### LINKED ISSUE #4077: null is also a value type
but not listed here: https://www.php.net/manual/en/language.types.value.php

kamil-tekiela:
Fixes #4077

https://en.wikipedia.org/wiki/Unit_type

### DIFF
--- a/language/types/type-system.xml
+++ b/language/types/type-system.xml
@@ -28,9 +28,6 @@
      <simpara>Built-in types</simpara>
      <itemizedlist>
       <listitem>
-       <simpara><type>null</type> type</simpara>
-      </listitem>
-      <listitem>
        <simpara>
         Scalar types:
        </simpara>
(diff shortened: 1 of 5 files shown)


language/types/type-system.xml
```diff
@@ -70,18 +67,21 @@
...
+         <simpara><type>false</type></simpara>
+        </listitem>
+        <listitem>
+         <simpara><type>true</type></simpara>
+        </listitem>
+        <listitem>
+         <simpara><type>null</type></simpara>
+        </listitem>
+       </itemizedlist>
       </listitem>
```
TimWolla, line 85: I disagree with this change, since `gettype(false)` returns boolean. Thus the boolean type is not simply the union of the `true` and the `false` type, but an independent type.

kamil-tekiela: I don't think I made that claim anywhere in this PR. What exactly in this PR do you disagree with? Your comment points to the lines where I changed "Value type" to "Unit type", added a null entry and indented the list to be inside of "Built-in types". I don't think either one of these makes it seem as if boolean is a union of false and true.

TimWolla: > I don't think either one of these makes it seem as if boolean is a union of false and true.

My reasoning goes like this:

- Any given value can have only one type.
- A unit type can contain only one value.
- `gettype(false)` indicates that the value `false` is of type `boolean`.
- That means it cannot be of type `false`.
- That means that the type `false` has no members.
- That means that the type `false` cannot actually exist as a unit type.

Girgias:
To continue on this example, the following enumerated type:
```
enum Suit
{
    case Hearts;
    case Diamonds;
    case Clubs;
    case Spades;
}
```
has 4 singleton types:
- `Suit::Hearts`
- `Suit::Diamonds`
- `Suit::Clubs`
- `Suit::Spades`

And it is possible to create a new unit type:
```
enum Unit
{
    case Unit;
}
```
which is only inhabited by a single type, the singleton `Unit::Unit` type.

Girgias (approved):

PR #4333, opened 2024-12-21
unserialize.xml Specify a strict type for the `allowed_classes` element

mmalferov:


kamil-tekiela (approved):
Thanks. I think it's fine to merge this.

PR #4337, opened 2024-12-22
Add warnings to use `php_sapi_name()` to test when running as CLI

jimwins:
This has been a recurring security problem encountered by packages, most recently Craft CMS, so it's probably worth some warnings in the documentation.


kamil-tekiela:
Perhaps an example would be good here because I don't actually know why that warning applies. I assume a lot of people will be confused too.

kamil-tekiela (approved):

PR #4338, opened 2024-12-23
array-column.xml Replace `array()` to `[]`, and make ex. #2 executable

mmalferov:


mmalferov:
Do you think it would be helpful to change the wording of the sentence in the `returnvalues` ​​section?

From:

`Returns an array of values representing a single column from the input array.`

For example, on:

`The function iterate over elements of the input array, pulls one value from each array by key or from each object by property name, and returns an array with the results.`

kamil-tekiela (approved):

kamil-tekiela:
> Do you think it would be helpful to change the wording...

No I think it's fine as it is now.

PR #4342, opened 2024-12-26
Fix typo in pattern.modifiers.xml

nyan-dot:
This section compares cases where the `r` flag is not used and where it is used. While the `u` flag is also in effect, it seems likely that "u" is a mistake and should be "r".


kamil-tekiela (approved):
MSTM!

PR #4357, opened 2025-01-02
[skip-revcheck] Remove xi:fallback, partial fix for doc-it

alfsb:
This, and https://github.com/php/doc-it/issues/55 fixes `doc-it` into a building state.


alfsb:
Could you also make the changes of https://github.com/php/doc-it/issues/55, so `doc-it` start building again (and all `doc-base` tests starts going all green)?

PR #4360, opened 2025-01-05
PDO(FETCH_CLASS): private and read-only properties are populated too

pereorga:
I'm not sure if this behavior is intentional, but it's something I rely on.

### DIFF
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -251,7 +251,7 @@
       called, allowing the population of properties regardless of their
       visibility or whether they are marked as <literal>readonly</literal>. If
       a property does not exist in the class, the magic
-      <link linkend="language.oop5.overloading.members"><methodname>__set</methodname></link>
+      <link linkend="object.set">__set()</link>
       method will be invoked if it exists; otherwise, a dynamic public property
       will be created. However, when <constant>PDO::FETCH_PROPS_LATE</constant>
       is also given, the constructor is called <emphasis>before</emphasis> the
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -253,7 +253,9 @@
       a property does not exist in the class, the magic
       <link linkend="language.oop5.overloading.members"><methodname>__set</methodname></link>
       method will be invoked if it exists; otherwise, a dynamic public property
-      will be created.
+      will be created. However, when <constant>PDO::FETCH_PROPS_LATE</constant>
+      is also given, the constructor is called <emphasis>before</emphasis> the
+      properties are populated.
      </simpara>
     </note>
    </listitem>
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -243,6 +243,7 @@
     <simpara>
      Specifies that the fetch method shall return a new instance of the
      requested class, mapping the columns to named properties in the class.
+     Private and read-only properties will also be populated.
     </simpara>
     <note>
      <simpara>
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -242,14 +242,18 @@
    <listitem>
     <simpara>
      Specifies that the fetch method shall return a new instance of the
-     requested class, mapping the columns to named properties in the class.
-     Private and read-only properties will also be populated.
+     requested class.
     </simpara>
     <note>
      <simpara>
-      The magic
+      The object is initialized by mapping the columns from the result set to
+      properties in the class. This process occurs before the constructor is
+      called, allowing the population of properties regardless of their
+      visibility or whether they are marked as <literal>readonly</literal>. If
+      a property does not exist in the class, the magic
       <link linkend="language.oop5.overloading.members"><methodname>__set</methodname></link>
-      method is called if the property doesn't exist in the requested class
+      method will be invoked if it exists; otherwise, a dynamic public property
+      will be created.
      </simpara>
     </note>
    </listitem>
--- a/reference/pdo/pdostatement/fetch.xml
+++ b/reference/pdo/pdostatement/fetch.xml
@@ -271,11 +271,11 @@ Reading backwards:
 
    <example><title>Construction order</title>
     <simpara>
-     When objects are fetched via <literal>PDO::FETCH_CLASS</literal> the object
+     When objects are fetched via <class>PDO::FETCH_CLASS</class>, the object
      properties are assigned first, and then the constructor of the class is
-     invoked. If <literal>PDO::FETCH_PROPS_LATE</literal> is also given, this
-     order is reversed, i.e. first the constructor is called, and afterwards the
-     properties are assigned.
+     invoked. However, when <class>PDO::FETCH_PROPS_LATE</class> is also given,
+     this order is reversed, i.e. first the constructor is called, and
+     afterwards the properties are assigned.
     </simpara>
     <programlisting role="php">
 <![CDATA[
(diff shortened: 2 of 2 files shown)


reference/pdo/constants.xml
```diff
@@ -243,6 +243,7 @@
     <simpara>
      Specifies that the fetch method shall return a new instance of the
      requested class, mapping the columns to named properties in the class.
+     Private and read-only properties will also be populated.
```
kamil-tekiela, line 246: You are correct, but why wouldn't it fill private and readonly properties? I guess the current description is a bit unclear and misleading. 

> mapping the columns to named properties in the class

That's not even true. If a property doesn't exist it will call `__set` and if that doesn't exist it will create a dynamic public property. 

The object is created with the properties populated before even the constructor is called. The visibility doesn't matter. And when the object is created the property has never been written to before so readonly doesn't matter either. 

I can merge it as you propose but I feel like we could do better here. However, I am incapable of doing it better myself.

reference/pdo/pdostatement/fetch.xml
```diff
@@ -60,6 +60,7 @@
          includes PDO::FETCH_CLASSTYPE (e.g. <literal>PDO::FETCH_CLASS |
          PDO::FETCH_CLASSTYPE</literal>) then the name of the class is
          determined from a value of the first column.
+         Private and read-only properties are also populated.
```
kamil-tekiela, line 63:
```suggestion
         Private and <literal>readonly</literal> properties are also populated.
```
I know that in English we would say "read-only" but we are talking here about PHP modifier so it should be one word.

reference/pdo/constants.xml
pereorga: Thank you for your feedback. I’ve tried to improve the explanation. Please let me know if it is correct

reference/pdo/pdostatement/fetch.xml
```diff
@@ -53,13 +53,18 @@
...
-         <literal>PDO::FETCH_PROPS_LATE</literal> is also given.
-         If <parameter>mode</parameter>
-         includes PDO::FETCH_CLASSTYPE (e.g. <literal>PDO::FETCH_CLASS |
-         PDO::FETCH_CLASSTYPE</literal>) then the name of the class is
-         determined from a value of the first column.
+         requested class. The object is initialized by mapping the columns of
+         the result set to properties in the class. This occurs before the
+         constructor is called, allowing properties to be populated regardless
+         of their visibility or whether they are marked as <literal>readonly</literal>.
+         If a property does not exist in the class, the magic <methodname>__set</methodname>
```
kamil-tekiela, line 60: @Girgias Do you know if we can use `methodname` with `__set`? The other one in this PR links to a different page. I think all the other ones in the repo link to `<link linkend="object.set">__set()</link>`. What's the correct approach?

Girgias: I think currently one must use `<link linkend="object.set">__set()</link>`, but it might make sense to add support for magic method linking to PhD.

I'll open an issue on it.

reference/pdo/constants.xml
```diff
@@ -242,13 +242,18 @@
...
      Specifies that the fetch method shall return a new instance of the
-     requested class, mapping the columns to named properties in the class.
+     requested class.
     </simpara>
     <note>
      <simpara>
-      The magic
+      The object is initialized by mapping the columns from the result set to
+      properties in the class. This process occurs before the constructor is
+      called, allowing the population of properties regardless of their
```
Girgias, line 251: This should probably also mention `PDO::FETCH_PROPS_LATE`

reference/pdo/pdostatement/fetch.xml
```diff
@@ -53,13 +53,18 @@
...
+         method will be invoked if it exists; otherwise, a dynamic public
+         property will be created. If <literal>PDO::FETCH_PROPS_LATE</literal>
+         is also given, the constructor is called <emphasis>before</emphasis>
+         the properties are populated. If <parameter>mode</parameter> includes
+         <literal>PDO::FETCH_CLASSTYPE</literal> (e.g.
```
Girgias, line 65:
```suggestion
         property will be created. If <constant>PDO::FETCH_PROPS_LATE</constant>
         is also given, the constructor is called <emphasis>before</emphasis>
         the properties are populated. If <parameter>mode</parameter> includes
         <constant>PDO::FETCH_CLASSTYPE</constant> (e.g.
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -53,13 +53,18 @@
         </para></listitem>
         <listitem><para>
```
Girgias, line 54: Complete aside, but the current XML structure is bonkers :|

reference/pdo/constants.xml
pereorga: Please see my update

reference/pdo/pdostatement/fetch.xml
pereorga: Also, dozens of literal tags that could be constant tags.

reference/pdo/pdostatement/fetch.xml
```diff
@@ -266,11 +271,11 @@ Reading backwards:
 
    <example><title>Construction order</title>
     <simpara>
-     When objects are fetched via <literal>PDO::FETCH_CLASS</literal> the object
+     When objects are fetched via <class>PDO::FETCH_CLASS</class>, the object
```
kamil-tekiela, line 274:
```suggestion
     When objects are fetched via <constant>PDO::FETCH_CLASS</constant>, the object
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -266,11 +271,11 @@ Reading backwards:
...
      properties are assigned first, and then the constructor of the class is
-     invoked. If <literal>PDO::FETCH_PROPS_LATE</literal> is also given, this
-     order is reversed, i.e. first the constructor is called, and afterwards the
-     properties are assigned.
+     invoked. However, when <class>PDO::FETCH_PROPS_LATE</class> is also given,
```
kamil-tekiela, line 276:
```suggestion
     invoked. However, when <constant>PDO::FETCH_PROPS_LATE</constant> is also given,
```

kamil-tekiela (changes requested):

reference/pdo/pdostatement/fetch.xml
kamil-tekiela: Could you please change it to `<link linkend="object.set">__set()</link>` for now. If support for magic methods gets added we can fix it later.

reference/pdo/constants.xml
```diff
@@ -242,13 +242,20 @@
...
+      visibility or whether they are marked as <literal>readonly</literal>. If
+      a property does not exist in the class, the magic
       <link linkend="language.oop5.overloading.members"><methodname>__set</methodname></link>
```
kamil-tekiela, line 254:
```suggestion
      <link linkend="object.set">__set()</link>
```

reference/pdo/pdostatement/fetch.xml
kamil-tekiela, line 60:
```suggestion
         If a property does not exist in the class, the magic <link linkend="object.set">__set()</link>
```

pereorga: Applied suggestions, thanks again!

kamil-tekiela (approved):
It looks to me. If @Girgias is ok with this then we can merge.

Girgias (approved):
Thank you! Sorry for the delay.

PR #4363, opened 2025-01-07
`ext/curl`: add `curl_share_init_persistent`

ericnorris:
This PR adds documentation for `curl_share_init_persistent` and its respective `CurlSharePersistentHandle` class.

RFC: https://wiki.php.net/rfc/curl_share_persistence_improvement.

I wrote the documentation by running `build/gen_stub.php` and also copying bits from other `curl` documentation pages. Since this is my first time contributing to the PHP documentation, I expect I may have made some mistakes or missed some common patterns that are usually followed.

### DIFF
--- a/reference/curl/curlsharepersistenthandle.xml
+++ b/reference/curl/curlsharepersistenthandle.xml
@@ -58,22 +58,22 @@
 
 </reference>
 <!-- Keep this comment at the end of the file
-     Local variables:
-     mode: sgml
-     sgml-omittag:t
-     sgml-shorttag:t
-     sgml-minimize-attributes:nil
-     sgml-always-quote-attributes:t
-     sgml-indent-step:1
-     sgml-indent-data:t
-     indent-tabs-mode:nil
-     sgml-parent-document:nil
-     sgml-default-dtd-file:"~/.phpdoc/manual.ced"
-     sgml-exposed-tags:nil
-     sgml-local-catalogs:nil
-     sgml-local-ecat-files:nil
-     End:
-     vim600: syn=xml fen fdm=syntax fdl=2 si
-     vim: et tw=78 syn=sgml
-     vi: ts=1 sw=1
+Local variables:
+mode: sgml
+sgml-omittag:t
+sgml-shorttag:t
+sgml-minimize-attributes:nil
+sgml-always-quote-attributes:t
+sgml-indent-step:1
+sgml-indent-data:t
+indent-tabs-mode:nil
+sgml-parent-document:nil
+sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+sgml-exposed-tags:nil
+sgml-local-catalogs:nil
+sgml-local-ecat-files:nil
+End:
+vim600: syn=xml fen fdm=syntax fdl=2 si
+vim: et tw=78 syn=sgml
+vi: ts=1 sw=1
 -->
--- a/reference/curl/curlsharepersistenthandle.xml
+++ b/reference/curl/curlsharepersistenthandle.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<reference xml:id="class.curlsharepersistenthandle" role="class" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
+
+  <title>The CurlSharePersistentHandle class</title>
+  <titleabbrev>CurlSharePersistentHandle</titleabbrev>
+
+  <partintro>
+
+    <!-- {{{ CurlSharePersistentHandle intro -->
+    <section xml:id="curlsharepersistenthandle.intro">
+      &reftitle.intro;
+      <para>
+        Represents a persistent cURL "share" handle.
+      </para>
+    </section>
+    <!-- }}} -->
+
+    <section xml:id="curlsharepersistenthandle.synopsis">
+      &reftitle.classsynopsis;
+
+      <!-- {{{ Synopsis -->
+      <classsynopsis class="class">
+        <ooclass>
+          <modifier>final</modifier>
+          <classname>CurlSharePersistentHandle</classname>
+        </ooclass>
+
+        <classsynopsisinfo role="comment">&Properties;</classsynopsisinfo>
+        <fieldsynopsis>
+          <modifier>public</modifier>
+          <modifier>readonly</modifier>
+          <type>array</type>
+          <varname linkend="curlsharepersistenthandle.props.options">options</varname>
+        </fieldsynopsis>
+      </classsynopsis>
+      <!-- }}} -->
+
+    </section>
+
+    <!-- {{{ CurlSharePersistentHandle properties -->
+    <section xml:id="curlsharepersistenthandle.props">
+      &reftitle.properties;
+      <variablelist>
+        <varlistentry xml:id="curlsharepersistenthandle.props.options">
+          <term><varname>options</varname></term>
+          <listitem>
+            <para>An array of <constant>CURL_LOCK_DATA_*</constant> constants shared in this handle.</para>
+          </listitem>
+        </varlistentry>
+      </variablelist>
+    </section>
+    <!-- }}} -->
+
+
+  </partintro>
+
+  <!-- &reference.curl.entities.curlsharepersistenthandle; -->
+
+</reference>
+<!-- Keep this comment at the end of the file
+     Local variables:
+     mode: sgml
+     sgml-omittag:t
+     sgml-shorttag:t
+     sgml-minimize-attributes:nil
+     sgml-always-quote-attributes:t
+     sgml-indent-step:1
+     sgml-indent-data:t
+     indent-tabs-mode:nil
+     sgml-parent-document:nil
+     sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+     sgml-exposed-tags:nil
+     sgml-local-catalogs:nil
+     sgml-local-ecat-files:nil
+     End:
+     vim600: syn=xml fen fdm=syntax fdl=2 si
+     vim: et tw=78 syn=sgml
+     vi: ts=1 sw=1
+-->
--- a/reference/curl/curlsharepersistenthandle.xml
+++ b/reference/curl/curlsharepersistenthandle.xml
@@ -45,7 +45,7 @@
         <varlistentry xml:id="curlsharepersistenthandle.props.options">
           <term><varname>options</varname></term>
           <listitem>
-            <para>The <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> constants shared between requests using this handle.</para>
+            <para>The <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> caches shared between requests using this handle.</para>
           </listitem>
         </varlistentry>
       </variablelist>
--- a/reference/curl/curlsharepersistenthandle.xml
+++ b/reference/curl/curlsharepersistenthandle.xml
@@ -45,7 +45,7 @@
         <varlistentry xml:id="curlsharepersistenthandle.props.options">
           <term><varname>options</varname></term>
           <listitem>
-            <para>An array of <constant>CURL_LOCK_DATA_*</constant> constants shared in this handle.</para>
+            <para>The <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> constants shared between requests using this handle.</para>
           </listitem>
         </varlistentry>
       </variablelist>
--- a/reference/curl/curlsharepersistenthandle.xml
+++ b/reference/curl/curlsharepersistenthandle.xml
@@ -1,61 +1,60 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
 <reference xml:id="class.curlsharepersistenthandle" role="class" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
+ <title>The CurlSharePersistentHandle class</title>
+ <titleabbrev>CurlSharePersistentHandle</titleabbrev>
 
-  <title>The CurlSharePersistentHandle class</title>
-  <titleabbrev>CurlSharePersistentHandle</titleabbrev>
+ <partintro>
 
-  <partintro>
+  <!-- {{{ CurlSharePersistentHandle intro -->
+  <section xml:id="curlsharepersistenthandle.intro">
+   &reftitle.intro;
+   <para>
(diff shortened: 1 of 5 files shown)


reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,120 @@
...
+  </refnamediv>
+
+  <refsect1 role="description">
+    &reftitle.description;
+    <methodsynopsis>
+      <type>CurlSharePersistentHandle</type><methodname>curl_share_init_persistent</methodname>
+      <methodparam><type>array</type><parameter>share_options</parameter></methodparam>
+    </methodsynopsis>
+    <para>
+      Initialize a <emphasis role="bold">persistent</emphasis> cURL share handle with the given share options. Unlike <function>curl_share_init</function>, handles created by this function will not be destroyed at the end of the PHP request. If a persistent share handle with the same set of <literal>$share_options</literal> is found, it will be reused.
```
pereorga, line 15: Should it be mentioned that this behavior might vary depending on the Server API (e.g., CLI, FPM, Apache Module)?

pereorga: Good question, I don't know. It would definitely be useful to have this documented on php.net.

reference/curl/curlsharepersistenthandle.xml
```diff
@@ -0,0 +1,80 @@
...
+  <title>The CurlSharePersistentHandle class</title>
+  <titleabbrev>CurlSharePersistentHandle</titleabbrev>
+
+  <partintro>
+
+    <!-- {{{ CurlSharePersistentHandle intro -->
+    <section xml:id="curlsharepersistenthandle.intro">
+      &reftitle.intro;
+      <para>
+        Represents a persistent cURL "share" handle.
```
TimWolla, line 14: The description should probably be synced with the one of `CurlShareHandle` (it's fine to update the other if the current wording is not good, but given they are so similar, they should have a similar documentation).

reference/curl/curlsharepersistenthandle.xml
```diff
@@ -0,0 +1,80 @@
...
+    </section>
+
+    <!-- {{{ CurlSharePersistentHandle properties -->
+    <section xml:id="curlsharepersistenthandle.props">
+      &reftitle.properties;
+      <variablelist>
+        <varlistentry xml:id="curlsharepersistenthandle.props.options">
+          <term><varname>options</varname></term>
+          <listitem>
+            <para>An array of <constant>CURL_LOCK_DATA_*</constant> constants shared in this handle.</para>
```
TimWolla, line 48:
```suggestion
            <para>An array of <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> constants shared in this handle.</para>
```

Something also bothers me about this wording, but I don't have a better suggestion.

reference/curl/functions/curl-share-init-persistent.xml
TimWolla, line 15: Please hard wrap this around 80 characters or so. Here's an example of a documentation page with long-form text: https://github.com/php/doc-en/blob/master/reference/random/random/randomizer/getfloat.xml

(Same applies to the entire file)

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,120 @@
...
+  </refsect1>
+
+  <refsect1 role="errors">
+    &reftitle.errors;
+    <simpara>
+      A <exceptionname>ValueError</exceptionname> is thrown when <literal>$share_options</literal> either: is empty, contains <constant>CURL_LOCK_DATA_COOKIE</constant>, or contains a value not matching a <constant>CURL_LOCK_DATA_*</constant> constant.
+    </simpara>
+    <simpara>
+      A <exceptionname>TypeError</exceptionname> is thrown when <literal>$share_options</literal> contains a value that cannot be cast to an integer.
+    </simpara>
```
TimWolla, line 52: Not sure if there's any hard rule around this (@Girgias?), but for the ext/random documentation I've made this a proper list and separate bullet points for each case. See: https://github.com/php/doc-en/blob/4d8af0862acd95744e96db9e611a9e6c1b02521e/reference/random/random/randomizer/getfloat.xml#L277-L295

In this case:
- `If <parameter>share_options</parameters> is empty, a <classname>ValueError</classname> will be thrown.`
- `If <parameter>share_options</parameters> contains a value not matching a <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant>, a <classname>ValueError</classname> will be thrown.`
- `If <parameter>share_options</parameters> contains <constant>CURL_LOCK_DATA_COOKIE</constant>, a <classname>ValueError</classname> will be thrown.`

The TypeError can possibly even be left out. While it's not strictly part of the method signature, passing invalid types will lead to value errors and the bit about “if you don't pass a CURL_LOCK_DATA_ constant you get an error” is probably sufficient.

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,120 @@
...
+
+]]>
+      </programlisting>
+    </example>
+  </refsect1>
+
+  <refsect1 role="seealso">
+    &reftitle.seealso;
+    <simplelist>
+      <member><function>curl_setopt</function></member>
```
TimWolla, line 96: This should probably reference `curl_share_init` (and vice versa).

TimWolla (commented):
Some remarks.

reference/curl/curlsharepersistenthandle.xml
ericnorris: I see, my hesitation is that the description of `CurlShareHandle` is somewhat unique in that it is the result of converting a resource to an object, and that seems noteworthy enough to remain that way. What might you suggest changing `CurlShareHandle` to? Something like:

```
Represents a cURL "share" handle. Prior to PHP 8.0.0, this was a resource.
```

reference/curl/curlsharepersistenthandle.xml
ericnorris: I've changed it to:

```
<para>The <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> constants shared between requests using this handle.</para>
```

How does that sound?

reference/curl/functions/curl-share-init-persistent.xml
ericnorris: Thank you, this is exactly the kind of thing I was looking for feedback on. I decided to leave the `TypeError` in since you can trigger it by doing something like `curl_share_init_persistent(["q"]);`.

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,152 @@
...
+      Returns a persistent cURL share handle.
+    </simpara>
+  </refsect1>
+
+  <refsect1 role="errors">
+    &reftitle.errors;
+    <itemizedlist>
+      <listitem>
+        <simpara>
+          If <parameter>$share_options</parameter> is empty, this function throws
```
TimWolla, line 57:
```suggestion
          If <parameter>share_options</parameter> is empty, this function throws
```
No dollar sign, same for the others.

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,152 @@
...
+    <methodsynopsis>
+      <type>CurlSharePersistentHandle</type><methodname>curl_share_init_persistent</methodname>
+      <methodparam><type>array</type><parameter>share_options</parameter></methodparam>
+    </methodsynopsis>
+    <para>
+      Initialize a <emphasis role="bold">persistent</emphasis> cURL share handle
+      with the given share options. Unlike <function>curl_share_init</function>,
+      handles created by this function will not be destroyed at the end of the
+      PHP request. If a persistent share handle with the same set of
+      <parameter>$share_options</parameter> is found, it will be reused.
```
TimWolla, line 19:
```suggestion
      <parameter>share_options</parameter> is found, it will be reused.
```

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,152 @@
...
+curl_setopt($ch1, CURLOPT_SHARE, $sh);
+
+// Execute the first cURL handle. This may reuse the connection from an earlier SAPI request.
+curl_exec($ch1);
+
+// Initialize the second cURL handle and assign the share handle to it.
+$ch2 = curl_init("http://example.com/");
+curl_setopt($ch2, CURLOPT_SHARE, $sh);
+
+// Execute the second cURL handle. This will reuse the connection from $ch2.
```
TimWolla, line 111:
```suggestion
// Execute the second cURL handle. This will reuse the connection from $ch1.
```
?

reference/curl/curlsharepersistenthandle.xml
TimWolla: That's a clear improvement. Perhaps some other word than “constants” can be found, because it's not sharing the constant, but the thing *represented by* the constant?

ericnorris: I thought the same thing, but I'm having a hard time coming up with that word. Re-reading https://everything.curl.dev/helpers/sharing.html, I see the word cache used, e.g.:

> `CURL_LOCK_DATA_DNS` - the DNS cache is where libcurl stores addresses for resolved hostnames for a while to make subsequent lookups faster.

The only time they don't use the word "cache" is for `CURL_LOCK_DATA_COOKIE`, which they call a cookie jar. Of course we don't allow that constant :).

So maybe "caches":

```
<para>The <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> caches shared between requests using this handle.</para>
```

?

ericnorris: I've went ahead with `caches`, let me know if you disagree or have a better suggestion.

reference/curl/curlsharepersistenthandle.xml
```diff
@@ -0,0 +1,80 @@
...
+
+  <title>The CurlSharePersistentHandle class</title>
+  <titleabbrev>CurlSharePersistentHandle</titleabbrev>
+
+  <partintro>
+
+    <!-- {{{ CurlSharePersistentHandle intro -->
+    <section xml:id="curlsharepersistenthandle.intro">
+      &reftitle.intro;
+      <para>
```
Girgias, line 13: New manual style is to prefer `simpara` over `para` when possible (i.e. when not using other "div" like structural elements)

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,152 @@
...
+        <listitem>
+          <para>
+            A non-empty array of <constant>CURL_LOCK_DATA_<replaceable>*</replaceable></constant> constants.
+          </para>
+          <para>
+            <emphasis role="bold">Note:</emphasis> <constant>CURL_LOCK_DATA_COOKIE</constant>
+            is not allowed and, if specified, this function will throw a
+            <exceptionname>ValueError</exceptionname>. Sharing cookies between PHP
+            requests may lead to inadvertently mixing up sensitive cookies between users.
+          </para>
```
Girgias, line 38: I'm guessing you want a `<note>` tag here?

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,152 @@
...
+        </listitem>
+      </varlistentry>
+    </variablelist>
+
+  </refsect1>
+
+  <refsect1 role="returnvalues">
+    &reftitle.returnvalues;
+    <simpara>
+      Returns a persistent cURL share handle.
```
Girgias, line 48: Could use `<classname>` with the actual class instance being returned?

Girgias (changes requested):
Mainly XML markup things:

- XML indent is 1 space, please fix
- Please prefer `simpara` over `para` as per the new manual style.

reference/curl/functions/curl-share-init-persistent.xml
ericnorris: I did, but didn't know that tag existed - thanks!

ericnorris:
Thanks for the review @Girgias, and let me know if my recent commits addressed your comments!

reference/curl/curlsharepersistenthandle.xml
```diff
@@ -0,0 +1,79 @@
...
+     sgml-parent-document:nil
+     sgml-default-dtd-file:"~/.phpdoc/manual.ced"
+     sgml-exposed-tags:nil
+     sgml-local-catalogs:nil
+     sgml-local-ecat-files:nil
+     End:
+     vim600: syn=xml fen fdm=syntax fdl=2 si
+     vim: et tw=78 syn=sgml
+     vi: ts=1 sw=1
+-->
```
Girgias, line 79: Nit: this comment should not have any indentation.

reference/curl/functions/curl-share-init-persistent.xml
```diff
@@ -0,0 +1,154 @@
...
+  &reftitle.seealso;
+  <simplelist>
+   <member><function>curl_setopt</function></member>
+   <member><function>curl_share_init</function></member>
+  </simplelist>
+ </refsect1>
+
+</refentry>
+<!-- Keep this comment at the end of the file
+     Local variables:
```
Girgias, line 136: Ditto

Girgias (approved):
Minor remaining nit, but LGTM now!

Thank you!

philip:
Looks good, I clicked the button, thanks @ericnorris 👍

ericnorris:
@pentium10 I don't believe this is true, per the [curl documentation](https://curl.se/libcurl/c/curl_multi_add_handle.html) for `curl_multi_add_handle`:

> If the easy handle is not set to use a shared ([CURLOPT_SHARE](https://curl.se/libcurl/c/CURLOPT_SHARE.html)) cache, it is made to use a DNS cache that is shared between all easy handles within the multi handle when [curl_multi_add_handle](https://curl.se/libcurl/c/curl_multi_add_handle.html) is called.

I've taken this to mean that you need to set the share object on each easy handle before adding the easy handle to the multi handle.

I may not be able to test this for a bit due to the US holiday, but I will try and respond to that GitHub issue when possible.

pentium10:
Thanks. I have a question, on a heavy request app, does it makes sense to switch from file_get_contents to CURL after this change? And prefacing this question, I am not well in PHP internal, but I assume file_get_contents doesn't have these improvements, and its a way slower than this? Maybe one day file_get_contents gets these added on its own way.

ericnorris:
I think you'd have to do some performance testing yourself, but yes, it's likely that you'd see a performance benefit from switching to something that involved `curl_share_init_persistent` instead of `file_get_contents`.

I believe that `file_get_contents` uses the stream API under the hood, so it may be that something already exists or that someone could add functionality to enable persistent streams, but I'm less familiar with that. Someone who is more familiar than me with that aspect of PHP internals could maybe chime in.

In any case, this is somewhat outside of the scope of this particular issue, so it's maybe worth opening a discussion elsewhere?

PR #4368, opened 2025-01-07
Copy setup code into second example to enable "run code"

ravage84:
While browsing the docs, I tried to "run [the] code" examples of [`is_a`](https://www.php.net/manual/en/function.is-a.php).

The firs worked but the second failed with `Warning: Undefined variable $WF in php-wasm run script on line 2` due to missing setup code.

If the "run code" functionality should actually be usable, code examples must be self-containing.


Girgias (changes requested):
I feel combining both examples into one seems better?

But thank you for noticing the issue and submitting a PR!

philip:
:+1:

Issue #4185, opened 2024-11-28
Update `ini_get()` to make use of `ini_parse_quantity()`


TimWolla:
... and rewrite the note about "When querying memory size values".

PR #4372, opened 2025-01-08
ini_get: Update example to use `ini_parse_quantity()`

edorian:
- Use ini_parse_quantity instead of custom code for ini value parsing.
- Replace register_globals ini value with something more current, as register_globals was removed in, I think, PHP 5.4.
- Change the wording for the ini_parse_quantity example to not contain “similar to” as it's deterministic output.

Resolves php/doc-en#4185

### DIFF
--- a/reference/info/functions/ini-get.xml
+++ b/reference/info/functions/ini-get.xml
@@ -55,37 +55,15 @@
 Our php.ini contains the following settings:
 
 display_errors = On
-register_globals = Off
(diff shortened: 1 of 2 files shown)


TimWolla (approved):
Created this PR together with edorian to showcase how editing the docs works. Requesting another review for wording approval.

Girgias (approved):
Thank you!

PR #4377, opened 2025-01-10
describe behavior of static variables inside anonymous functions

pereorga:
While working with static variables inside anonymous functions, I unexpectedly found that the static variable would not persist across calls because the anonymous function was recreated each time. This behavior was counterintuitive for me, so I updated the documentation to describe it.


Girgias:
Thank you!

PR #4385, opened 2025-01-13
Update ini.list.xml

Faabvk:
Added two Zend ini directives that were made available in PHP 8.3, but aren't on this list.
I tested manually to determine the changeable mode, but also looked at source code.

### DIFF
--- a/appendices/ini.list.xml
+++ b/appendices/ini.list.xml
@@ -777,10 +777,22 @@
        <entry></entry>
       </row>
       <row>
+      <entry><link linkend="ini.zend.max_allowed_stack_size">zend.max_allowed_stack_size</link></entry>
+       <entry><literal>"0"</literal></entry>
+       <entry><constant>INI_SYSTEM</constant></entry>
+       <entry>Available as of PHP 8.3.0</entry>
+      </row>
+      <row>
        <entry><link linkend="ini.zend.multibyte">zend.multibyte</link></entry>
        <entry><literal>"0"</literal></entry>
        <entry><constant>INI_PERDIR</constant></entry>
        <entry></entry>
+      </row>
+      <row>
+       <entry><link linkend="ini.zend.reserved_stack_size">zend.reserved_stack_size</link></entry>
+       <entry><literal>"0"</literal></entry>
+       <entry><constant>INI_SYSTEM</constant></entry>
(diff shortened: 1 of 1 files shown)


Girgias:
These also need to be documented somewhere else I think.

@arnaud-lb did you introduce those?

arnaud-lb:
Oh, I don't know why but I was pretty sure these where documented. I will take care of it.

arnaud-lb (approved):

appendices/ini.list.xml
```diff
@@ -776,12 +776,24 @@
        <entry><constant>INI_ALL</constant></entry>
        <entry></entry>
       </row>
+      <row>
+      <entry><link linkend="ini.zend.max_allowed_stack_size">zend.max_allowed_stack_size</link></entry>
```
arnaud-lb, line 780: For consistency with other entries, the linkend should use regular dashes (`-`) instead of underscores:

``` suggestion
      <entry><link linkend="ini.zend.max-allowed-stack-size">zend.max_allowed_stack_size</link></entry>
```

appendices/ini.list.xml
```diff
@@ -776,12 +776,24 @@
...
+       <entry>Available as of PHP 8.3.0</entry>
+      </row>
       <row>
        <entry><link linkend="ini.zend.multibyte">zend.multibyte</link></entry>
        <entry><literal>"0"</literal></entry>
        <entry><constant>INI_PERDIR</constant></entry>
        <entry></entry>
       </row>
+      <row>
+       <entry><link linkend="ini.zend.reserved_stack_size">zend.reserved_stack_size</link></entry>
```
arnaud-lb, line 792: Here too

appendices/ini.list.xml
Faabvk: Done!

appendices/ini.list.xml
Faabvk: Also done :)

arnaud-lb (approved):
The build fails because the linkends do not exist yet (https://github.com/php/doc-en/pull/4398 fixes that), but otherwise this looks good to me.

I would suggest to merge https://github.com/php/doc-en/pull/4398 directly, as it includes this PR.

Girgias:
Closing as #4398 which contains this has been merged.

PR #4392, opened 2025-01-14
Fix hooks example

### LINKED ISSUE #4336: identical examples are confusing
From manual page: https://php.net/language.oop5.property-hooks

---

The code examples for "Hook inheritance" (`#9`) and "Parent hook access (set)" (`#10`) are identical; it appears they were added this way in the initial commit.  This is confusing, especially with the comment between them stating the former _"could be rewritten more efficiently"_ as the latter, which is also present in the initial commit and not materially changed thereafter.

Please update either/both of these examples to show distinct approaches, or remove one of them along with the reference to rewriting.  Thank you!

Crell:
Fixes #4336

Copy-pasta error on my part when cloning the RFC to the docs page.  This should resolve it.

### DIFF
--- a/language/oop5/property-hooks.xml
+++ b/language/oop5/property-hooks.xml
@@ -498,7 +498,10 @@ class PositivePoint extends Point
     Hooks may not access any other hook except their own parent on their own property.
    </simpara>
    <simpara>
-    The example above could be rewritten more efficiently as follows.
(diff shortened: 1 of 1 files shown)


Girgias (approved):
Thanks, sorry for the delay.

PR #4393, opened 2025-01-14
Minor corection to enum traits.

Crell:
As called out by a comment, that I have since deleted as this PR will resolve it.


kamil-tekiela (approved):
Thanks.

PR #4395, opened 2025-01-14
gd: explain why imagedestroy() is NOP

pereorga:


kamil-tekiela:
I'm not sure about this change. Are you saying that prior to PHP 8.0 images weren't automatically destroyed? If so, what do you base this on?

pereorga:
@kamil-tekiela Not necessarily, I believe this only states that as of 8.0.0, images are destroyed automatically when no longer referenced (without making it clear what happened earlier, in unsupported versions). But happy to make it less confusing, should I remove the version number?

kamil-tekiela:
> images are destroyed automatically when no longer referenced

I'm pretty sure that was true before too. I haven't checked but I have no reason to believe anything changed in this regard in PHP 8. All the information I was able to find online shows that the resources were destroyed automatically when no longer referenced, in PHP 7 too. See for example https://www.php.net/manual/en/language.types.resource.php#language.types.resource.self-destruct

The reason why this function is no-op is a technical one in my opinion. While resources had a feature that allowed for them to be closed (it's a separate type), objects do not have such a feature. Some extensions that wanted to have this feature had to implement it on an extension level, e.g. mysqli can close objects but this feature is part of that extension. 

There are proponents of closing objects, but in general, it doesn't make sense for objects to have such a feature. Hence, when changing GD to objects, it was not possible to carry the old behaviour of `imagedestroy` without implementing this extra feature in the GD extension. 

So if you want to clarify why this function is a no-op, it should say that it does nothing because as of PHP 8, GD uses objects instead of resources and objects cannot be closed.

reference/image/functions/imagedestroy.xml
```diff
@@ -13,9 +13,11 @@
...
   &note.resource-migration-8.0-dead-function;
   <para>
-   Prior to PHP 8.0.0, <function>imagedestroy</function> freed any memory associated
-   with image <parameter>image</parameter>. As of 8.0.0, images are automatically
-   destroyed when no longer referenced.
+   Prior to PHP 8.0.0, <function>imagedestroy</function> freed any memory
+   associated with the image resource <parameter>image</parameter>. As of 8.0.0,
+   the GD extension uses objects instead of resources for images, and objects
+   cannot be explicitly closed; instead, they are automatically destroyed when
+   no longer referenced.
```
kamil-tekiela, line 20:
```suggestion
   Prior to PHP 8.0.0, <function>imagedestroy</function> freed any memory
   associated with the <parameter>image</parameter> resource. As of 8.0.0,
   the GD extension uses objects instead of resources, and objects
   cannot be explicitly closed.
```
How about this?

kamil-tekiela (approved):
Thanks

PR #4396, opened 2025-01-14
gd: do not use or unnecessary reference imagedestroy()

pereorga:
Because `imagedestroy()` is NOP in all supported PHP versions (as of 8.0.0), there is no need to reference it.


kamil-tekiela (approved):
Thanks. I feel sorry for translators but this is a good change that we'd have to make sooner or later.

PR #4400, opened 2025-01-16
Replace concatenations with argument list in examples, add the missing dot char

mmalferov:


kamil-tekiela:
I'm going to say no to this change. Leave the dots as they are now. There's no point in changing them. If anything what's more annoying to me is the use of `PHP_EOL`.

PR #4402, opened 2025-01-17
Match expressions doesn't need always to be terminated by a semicolon

driade:
I'm not sure if this PR is important for the documentation, but "match" doesn't always need to be terminated by semicolon. 

I've learned this today :)

### DIFF
--- a/language/control-structures/match.xml
+++ b/language/control-structures/match.xml
@@ -86,13 +86,38 @@ string(8) "Teenager"
     The result of a <literal>match</literal> expression does not need to be used.
    </simpara>
   </note>
-  <note>
-   <simpara>
+  <para>
     A <literal>match</literal> expression <emphasis>must</emphasis> be
-    terminated by a semicolon <literal>;</literal>.
-   </simpara>
-  </note>
- </example>
+    terminated by a semicolon <literal>;</literal> in most scenarios.
+    There may be some exceptions where a semicolon is not required, such as
+    nested matches
+    <informalexample>
+   <programlisting role="php">
+<![CDATA[
+<?php
+match ($a) {
+    match (true) {1 => 2} => match (true) {1 => 2}
+};
+?>
+]]>
+   </programlisting>
+  </informalexample>
+
+  or operating with the result of the expression
+
+  <informalexample>
+   <programlisting role="php">
+<![CDATA[
+<?php
+$output = match ($food) {
+    'apple' => 2,
+    default => 0
+} + 3;
+?>
+]]>
+   </programlisting>
+  </informalexample>
+</para>
 
  <para>
   The <literal>match</literal> expression is similar to a


driade:
Some examples

https://github.com/driade/phpfmt8/commit/d5e484b281fd80be99e2d5e66458e99c244369a7#diff-f9724eb0e73ef20a68d492baa1516f2c47019efdda7bd213ee856d0a8a045724

kamil-tekiela:
> Some expressions can be considered as statements. In this case, a statement has the form of 'expr ;' that is, an expression followed by a semicolon. In $b = $a = 5;, $a = 5 is a valid expression, but it's not a statement by itself. $b = $a = 5;, however, is a valid statement.

https://www.php.net/manual/en/language.expressions.php 

I'm not sure why we repeat that information on the `match` page. Perhaps it should be removed instead of expanded.

language/control-structures/match.xml
```diff
@@ -86,13 +86,39 @@ string(8) "Teenager"
     The result of a <literal>match</literal> expression does not need to be used.
    </simpara>
   </note>
-  <note>
-   <simpara>
+
+  <para>
     A <literal>match</literal> expression <emphasis>must</emphasis> be
-    terminated by a semicolon <literal>;</literal>.
```
Girgias, line 92: I think just rephrasing the existing note to

> When a match expression is used as a standalone expression it must be terminated by a semicolon.

Covers the nuance just fine without needing to add examples.

You could use a match expression as a [function argument](https://3v4l.org/ZMdQW), and you don't need a semi colon there either.

driade: Here we go then!

language/control-structures/match.xml
```diff
@@ -88,8 +88,7 @@ string(8) "Teenager"
   </note>
   <note>
    <simpara>
-    A <literal>match</literal> expression <emphasis>must</emphasis> be
-    terminated by a semicolon <literal>;</literal>.
+    When a <literal>match</literal> expression is used as a standalone expression it <emphasis>must</emphasis> be terminated by a semicolon<literal>;</literal>.
```
kamil-tekiela, line 91:
```suggestion
    When a <literal>match</literal> expression is used as a standalone
    expression it <emphasis>must</emphasis> be terminated
    by a semicolon<literal>;</literal>.
```

language/control-structures/match.xml
```diff
@@ -88,8 +88,9 @@ string(8) "Teenager"
   </note>
   <note>
    <simpara>
-    A <literal>match</literal> expression <emphasis>must</emphasis> be
-    terminated by a semicolon <literal>;</literal>.
+    When a <literal>match</literal> expression is used as a standalone
+    expression it <emphasis>must</emphasis> be terminated
+    by a semicolon<literal>;</literal>.
```
Girgias, line 93:
```suggestion
    by a semicolon <literal>;</literal>.
```

PR #4404, opened 2025-01-18
unserialize(): fix XML tags for union type

nsfisis:
Fix `unserialize()` manual: modify XML tags for union type to make each type name hyperlink.

https://www.php.net/manual/en/function.unserialize.php

`<type>array|bool</type>` is recognized as single type of `array|bool`, which is not defined. `<type class="union">` tag is appropriate for union types.


Girgias:
Thank you!

PR #4408, opened 2025-01-20
createfromstring.xml Fix typo: document → string + CS

mmalferov:


nielsdos (approved):

reference/dom/dom/htmldocument/createfromstring.xml
```diff
@@ -48,7 +48,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <simpara>
-   The parsed document as an <classname>Dom\HTMLDocument</classname> instance.
+   The parsed string as an <classname>Dom\HTMLDocument</classname> instance.
```
nielsdos, line 51: Actually https://www.php.net/manual/en/dom-htmldocument.createfromfile.php uses the same phrasing, and I kinda used "parsed document" on purpose. But I don't object changing this (if done consistently)

PR #4413, opened 2025-01-22
Update enumerations.xml

exakat:
Mentionned the special behavior that is limited to PHP 8.1

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -192,9 +192,10 @@ enum Suit: string
   </para>
 
   <para>
-   The equivalent values may be a constant scalar expression, since PHP 8.2. 
-   In PHP 8.1, the equivalent values had to be literals or literal expressions. Constants and constant expressions
-   were not supported.  That is, <literal>1 + 1</literal> is allowed, but <literal>1 + SOME_CONST</literal>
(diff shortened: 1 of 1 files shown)


language/enumerations.xml
```diff
@@ -192,8 +192,9 @@ enum Suit: string
   </para>
 
   <para>
-   Equivalent values must be literals or literal expressions. Constants and constant expressions
-   are not supported.  That is, <literal>1 + 1</literal> is allowed, but <literal>1 + SOME_CONST</literal>
+   The equivalent values may be a constant scalar expression, since PHP 8.2. 
+   In PHP 8.1, the equivalent values had to be literals or literal expressions. Constants and constant expressions
+   were not supported.  That is, <literal>1 + 1</literal> is allowed, but <literal>1 + SOME_CONST</literal>
```
Girgias, line 197:
```suggestion
   The equivalent values may be a constant scalar expression.
   Prior to PHP 8.2.0, the equivalent values had to be literals or literal expressions.
   This means that constants and constant expressions were not supported.
   That is, <code>1 + 1</code> was allowed, but not <code>1 + SOME_CONST</code>.
```

PR #4416, opened 2025-01-24
ArrayObject::__construct Replace tag and implement → extend + CS

mmalferov:


Girgias (approved):

mmalferov:
Maybe create a separate pull request that corrects the categorical requirement to extend the iterator class?

> The class must extend the [ArrayIterator](https://www.php.net/manual/en/class.arrayiterator.php) class.

To:

> The class must belong to or extend the [ArrayIterator](https://www.php.net/manual/en/class.arrayiterator.php) class.

```php
<?php

$array = [];

// The third argument do not extends the ArrayIterator class, while "MUST"
$arrayObject = new ArrayObject($array, 0, 'ArrayIterator');```

Girgias:
I suppose
> must be a subclass (or subtype) of ArrayIterator
would work better?

PR #4420, opened 2025-01-28
Missing function and method definitions in Example #1 for register_tick_function()

UsingSession:
The example for [register_tick_function()](https://www.php.net/manual/en/function.register-tick-function.php) is incomplete. It is missing the definitions for the callback function my_function and the class method my_method.

Additionally, due to the missing definitions, the "Sandbox Example" code provided on the PHP documentation page cannot be run directly, making it difficult for users to test or understand the functionality.

Missing function and method definitions in Example #1 for register_tick_function()

### DIFF
--- a/reference/funchand/functions/register-tick-function.xml
+++ b/reference/funchand/functions/register-tick-function.xml
@@ -59,13 +59,11 @@
 <?php
 declare(ticks=1);
 
-// Callback function
-function my_function($param) {
+function my_tick_function($param) {
     echo "Tick callback function called with param: $param\n";
 }
 
-// using a function as the callback
-register_tick_function('my_function', true);
+register_tick_function('my_tick_function', true);
 ?>
 ]]>
     </programlisting>
--- a/reference/funchand/functions/register-tick-function.xml
+++ b/reference/funchand/functions/register-tick-function.xml
@@ -64,19 +64,8 @@ function my_function($param) {
     echo "Tick callback function called with param: $param\n";
 }
 
-// Class with a method
-class my_class {
-    public function my_method($param) {
-        echo "Tick callback method called with param: $param\n";
-    }
-}
-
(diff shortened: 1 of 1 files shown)


Girgias:
I think the whole example should probably reworked at the same time, I don't really see the point in showcasing the different ways to pass a callable as that would also require examples with closures.

But yes, many examples cannot be executed as the WASM runner is new and many examples predate it by years/decades.

UsingSession:
@Girgias 
I agree that the entire example can be reviewed and improved.

My main idea was to make the current code functional without major changes.

I accidentally came across this function on Stack Overflow and decided to learn more on the documentation site. Seeing the possibility to run the code, I decided to test it, but encountered an error, so ... yeah🙂

Girgias:
The ability to run code is only a recent feature to the website and many example do not work.

I would prefer to remove the various ways to set a callable (as that is general PHP behaviour rather than a register tick one) and just have a single function, which is removing code from the example to make it functional.

UsingSession:
@Girgias

Sorry, but I did a mistake. I accidentally did a merge. Should I create a new PR?

Girgias:
I don't see the mistake? The PR seems fine.

reference/funchand/functions/register-tick-function.xml
```diff
@@ -59,12 +59,13 @@
 <?php
 declare(ticks=1);
 
+// Callback function
+function my_function($param) {
+    echo "Tick callback function called with param: $param\n";
+}
+
 // using a function as the callback
 register_tick_function('my_function', true);
```
Girgias, line 68:
```suggestion
function my_tick_function($param) {
    echo "Tick callback function called with param: $param\n";
}

register_tick_function('my_tick_function', true);
```

Girgias:
Thank you!

PR #4423, opened 2025-01-29
ext/bcmath: Moved scale description from language-snippets to bcadd

SakiTakamachi:


nielsdos:
I'm not sure if there's much point in doing this, maybe I missed some discussion somewhere; but I won't object.

SakiTakamachi:
@nielsdos 
Oh, sorry
the discussion is here:
https://github.com/php/doc-en/pull/4187#discussion_r1887141524

nielsdos:
I see

But I think you missed that you can giev it a special xml:id:

> xml:id attribute like "bcmath-number.add..parameters.scale"

SakiTakamachi:
Oh, it was my mistake to close this.
There is no need to forcefully assign an ID, so may be fine with the current PR status.

SakiTakamachi:
If don't assign IDs, I may not need to make this change...

alfsb:
> hmm
> 
> ```
>   Random removing duplicated xml:id: function.bcadd.parameters.scale
> ```
> 
> Since the includes are nested, duplicate ids seem to occur. I don't think it's necessary to force this, so I'll close it for now.

Per https://github.com/php/doc-base/blob/master/docs/structure.md#xmlid-structure , the xml:id would be:
```
function.bcadd..parameters.scale
```
not
```
function.bcadd.parameters.scale
```
Mind the double dot.

SakiTakamachi:
@alfsb 
I was wrong.
Using two dots worked, I didn't understand the documentation correctly...
Thanks!

SakiTakamachi:
I can probably use IDs elsewhere as well, but does it seem ok to add the commit to this PR?

SakiTakamachi:
Leave this alone for now and create a follow-up PR for the others.

alfsb:
> I'm not sure if there's much point in doing this, maybe I missed some discussion somewhere; but I won't object.

Only to clarify, this change is good because `language-snippets.ent` is *huge*, is the most frequently changed file, and it's not normal XML. All these points make this file a real PITA for translations with small teams. Avoiding adding things on it, and even better, having these "snippets" as normal XML, in normal XML files, that are infrequently changed, helps translations big time.

Also, it is way easier to debug `xml:id` breakages instead of DTD entity references breakages, even the nested ones.

alfsb:
> I didn't understand the documentation correctly... Thanks!

I will try to change it to make more detailed.

nielsdos (approved):

PR #4425, opened 2025-01-29
Update `round()` for PHP-8.4 and Minor fixes for BCMath's round

SakiTakamachi:


SakiTakamachi:
I just noticed, does this conflict with #4107 ?

reference/math/functions/round.xml
```diff
@@ -321,6 +328,86 @@ float(-1.5)
 ]]>
     </screen>
    </example>
+  <example>
+   <title>Example of using <enumname>RoundingMode</enumname></title>
```
TimWolla, line 332: I think the example might be more useful when it looks like this test: https://github.com/php/php-src/blob/master/ext/standard/tests/math/round_RoundingMode.phpt.

It clearly shows the name of the rounding mode that is used and then both the original value and the rounded value.

reference/math/functions/round.xml
```diff
@@ -112,6 +112,7 @@
          </tbody>
         </tgroup>
        </informaltable>
+       However, please note that some newly added modes only exist in <link linkend="enum.roundingmode">RoundingMode</link>.
```
TimWolla, line 115:
```suggestion
       However, please note that some newly added modes only exist in <enumname>RoundingMode</enumname>.
```

reference/bc/bcmath/number/round.xml
```diff
@@ -25,11 +25,11 @@
   <variablelist>
    <!-- precision parameter -->
    <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('function.round')/db:refsect1[@role='parameters']/descendant::db:varlistentry[2])" />
-   <varlistentry>
+   <varlistentry xml:id="bcmath-number.round..parameters.mode">
     <term><parameter>mode</parameter></term>
     <listitem>
      <simpara>
-      Specifies the rounding mode.
+      Specifies the rounding mode. For more information about modes, see <link linkend="enum.roundingmode">RoundingMode</link>.
```
TimWolla, line 32:
```suggestion
      Specifies the rounding mode. For more information about modes, see <enumname>RoundingMode</enumname>.
```

TimWolla (commented):
I leave the Docbook / wording review to Girgias.

reference/math/functions/round.xml
SakiTakamachi: I can't access the enum page because there is no link, is this something plan to address?

reference/math/functions/round.xml
SakiTakamachi: fixed in https://github.com/php/doc-en/pull/4425/commits/071f4a2d4b1acb5c864f96cb84ac895900ddb831

reference/math/functions/round.xml
Girgias: Yes, but currently not supported :/

Girgias (commented):
Please use `<enumname>` tags instead of `<link>` the linking can be resolved out-of-band.

SakiTakamachi:
Fixed.
I decided not to use the `link` tag because it seems like I could support automatic linking of `enumname`.

Girgias (approved):

PR #4442, opened 2025-02-05
Update shmop-read.xml

mikepultz:
shmop_read doesn't return false on failure.


Girgias (approved):
Indeed, this used to be the case in PHP 7.

PR #4449, opened 2025-02-07
late-static-bindings.xml Adjust the output of the 3rd code example + CS

mmalferov:


language/oop5/late-static-bindings.xml
```diff
@@ -131,31 +147,40 @@ B
...
+        /* Original method is replaced; the scope of the new one is C */
     }
 }
 
 $b = new B();
 $b->test();
+
 $c = new C();
-$c->test();   //fails
+$c->test();   // Fails
```
Girgias, line 183: While at it?
```suggestion
$c->test(); // Fails
```
Or maybe even better, just `try-catch` it and echo the exception class followed by the error message?

mmalferov: I made changes, is that what you meant? ;-)

Girgias (approved):
Indeed! thank you

Issue #4453, opened 2025-02-09
Incorrect indentation in file `reference/datetime/datetimeinterface.xml`


Marqitos:
Incorrect indentation in file `reference/datetime/datetimeinterface.xml`.

From line 356

Marqitos:
Detected incorrect indentation in file `reference/datetime/datetime.xml`

From line 67 - 105

Marqitos:
Detected incorrect indentation in file `reference/datetime/functions/strptime.xml`

From line 153 - 227

vrana:
There are many files with incorrect indentation. Create a pull request for all. Start the commit message with `[skip-revcheck]`.

PR #4454, opened 2025-02-09
Correct indentation on file `reference/datetime/datetimeinterface.xml`

Marqitos:
#4453 - Correct indentation on file `reference/datetime/datetimeinterface.xml`
From lines 356 - 390


kamil-tekiela:
You're correct, but is there a reason why you picked these files and only made this change? There are hundreds of files with wrong indentation and changing them all without reason is probably too much churn.

Marqitos:
Sorry if this is not correct.
I am updating the translation of the documentation into Spanish.
And when I translate a file that has the wrong indentation, at the same time that I put it correctly in the Spanish documentation, I modify the original file, so that it is correct.

PR #4455, opened 2025-02-09
password_hash: Update for PHP 8.4

TimWolla:
This updates the `password_hash()` documentation for the updated default cost of bcrypt in PHP 8.4. It also performs some general cleanup.

### DIFF
--- a/reference/password/functions/password-hash.xml
+++ b/reference/password/functions/password-hash.xml
@@ -34,9 +34,9 @@
     </listitem>
     <listitem>
      <simpara>
-      <constant>PASSWORD_BCRYPT</constant> - Use the <constant>CRYPT_BLOWFISH</constant> algorithm to
+      <constant>PASSWORD_BCRYPT</constant> - Use the bcrypt algorithm to
       create the hash. This will produce a standard <function>crypt</function> compatible hash using
-      the "$2y$" identifier. The result will always be a 60 character string, &return.falseforfailure;.
+      the <literal>$2y$</literal> identifier. The result will always be a 60 character string, &return.falseforfailure;.
      </simpara>
     </listitem>
     <listitem>
@@ -81,8 +81,8 @@
       Examples of these values can be found on the <function>crypt</function> page.
      </para>
      <para>
-      If omitted, a default value of <literal>10</literal> will be used. This is a good
-      baseline cost, but you may want to consider increasing it depending on your hardware.
+      If omitted, a default value of <literal>12</literal> will be used. This is a good
+      baseline cost, but you may want to consider adjusting it depending on your hardware.
      </para>
     </listitem>
    </itemizedlist>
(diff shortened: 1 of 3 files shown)


reference/password/functions/password-hash.xml
```diff
@@ -34,9 +34,9 @@
     </listitem>
     <listitem>
      <simpara>
-      <constant>PASSWORD_BCRYPT</constant> - Use the <constant>CRYPT_BLOWFISH</constant> algorithm to
+      <constant>PASSWORD_BCRYPT</constant> - Use the bcrypt algorithm to
       create the hash. This will produce a standard <function>crypt</function> compatible hash using
-      the "$2y$" identifier. The result will always be a 60 character string, &return.falseforfailure;.
+      the <literal>$2y$</literal> identifier. The result will always be a 60 character string, &return.falseforfailure;.
```
nielsdos, line 39: I think we mostly still put "" around the string literals, like so:
```suggestion
      the <literal>"$2y$"</literal> identifier. The result will always be a 60 character string, &return.falseforfailure;.
```

nielsdos (approved):
LGTM, I have one remark, I'm not sure if we have a real guideline about that but it's what I'm used to.

reference/password/functions/password-hash.xml
TimWolla: In this case it's *not really* but more of an “abstract identifier”.

reference/password/functions/password-hash.xml
```diff
@@ -81,8 +81,8 @@
       Examples of these values can be found on the <function>crypt</function> page.
      </para>
      <para>
-      If omitted, a default value of <literal>10</literal> will be used. This is a good
-      baseline cost, but you may want to consider increasing it depending on your hardware.
+      If omitted, a default value of <literal>12</literal> will be used. This is a good
+      baseline cost, but you may want to consider adjusting it depending on your hardware.
```
Girgias, line 85: Maybe at the same time remove personalization (i.e. usage of "you")?

reference/password/functions/password-hash.xml
Girgias: I'm happy to have it without the quotes, as its a sort of prefix identifier in the generated string.

reference/password/functions/password-hash.xml
TimWolla: It's all over the document. I'll do this in a follow-up.

siliconforks:
I think [constants.xml](https://github.com/php/doc-en/blob/master/reference/password/constants.xml) needs to be updated too?

https://github.com/php/doc-en/blob/51ea49d0a56569458483a36478f6e5571a59bdc8/reference/password/constants.xml#L46

nielsdos:
> I think [constants.xml](https://github.com/php/doc-en/blob/master/reference/password/constants.xml) needs to be updated too?
> 
> https://github.com/php/doc-en/blob/51ea49d0a56569458483a36478f6e5571a59bdc8/reference/password/constants.xml#L46

Good find, thanks. Fixed in https://github.com/php/doc-en/commit/66aff414be91c5a0446be585aa6ac78121de1e67

PR #4465, opened 2025-02-14
`ReflectionClassConstant::getType()`: replace "parameter" reference

DanielEScherzer:


DanielEScherzer:
CC @Girgias - follow-up to #4049

Issue #4467, opened 2025-02-15
Usual comments are missing in files `language/operators/*.xml`.


Marqitos:
All files in `language/operators/` does not contain the usual comments at the beginning and end of it.

philip:
The PR was closed with comment, so closing this issue too.

PR #4468, opened 2025-02-15
Add initial and end comments to files in `language/operators` (#4467)

Marqitos:
Add usuals initial and end comments to all files in `language/operators` (#4467)


Girgias:
Can we not? Those are useless and were a conscious decision.

Removing them for all other files is just useless churn for translations.

Marqitos:
Sorry, I just saw those without the comments, and I thought I was a mistake.

I'll keep that in mind.

Thanks.

Girgias:
No worries, those comments (especially trailing ones) are just for some text editors to understand how to fold stuff... maybe we should get rid of them completely. But that's a documentation wide discussion.

PR #4472, opened 2025-02-18
Re-add mysqli::__construct return value description

DeveloperRob:
In https://github.com/php/doc-en/commit/390278d79b453af8f2f0837596c4cd25882b951a, lines 126-129 were incorrectly removed - `__construct` will still return an object in all circumstances.

### DIFF
--- a/reference/mysqli/mysqli/construct.xml
+++ b/reference/mysqli/mysqli/construct.xml
@@ -124,6 +124,10 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
+   <methodname>mysqli::__construct</methodname> always returns an object which represents the connection to a MySQL Server,
+   regardless of it being successful or not.
+  </para>
+  <para>
    <function>mysqli_connect</function> returns an object which represents the connection to a MySQL Server,
    &return.falseforfailure;.
   </para>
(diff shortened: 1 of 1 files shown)


DeveloperRob:
If it is useful for context, this came about because one of my team thought that the constructor behaved different because of the return type notes and that the `mysqli` extension has a number of other non-standard behaviours. Thinking of other similar classes; the only other example I can think of where the extension/class has an OOP and procedural version is `DateTime`; and `DateTime::__construct` does have a return section (https://www.php.net/manual/en/datetime.construct.php).

DeveloperRob:
(Fixed the whitespace issue)

reference/mysqli/mysqli/construct.xml
```diff
@@ -123,6 +123,10 @@
 
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
+  <para>
+   <methodname>mysqli::__construct</methodname> always returns an object which represents the connection to a MySQL Server,
```
kamil-tekiela, line 127:
```suggestion
   <methodname>mysqli::__construct</methodname> always returns an object
   which represents the connection to a MySQL Server,
```

kamil-tekiela (approved):
Ok, I'm sold. Let's add it back in.

PR #4474, opened 2025-02-19
Clarify socket_write()

nielsdos:
The note was not parseable for me, and the use of "buffer" for both the network buffers and the input data (in the same sentence) made it too confusing for me to interpret initially.


philip:
👍

PR #4479, opened 2025-02-20
Adjust column width in resources.xml for improved layout

### LINKED ISSUE #4471: resource documentation page not redered correctly
From manual page: https://php.net/resource

---

This page might like some love, somthing went wrong here :)

sy-records:
Fix #4471

```html
     <col width="1*"/>
     <col style="width: 1.4*;"/>
```

Firefox renders the style. But it's an invalid property value in Chrome.

https://github.com/php/phd/blob/3c00a1a7539b660a5d792d4306d4c5a355676eea/phpdotnet/phd/Package/Generic/XHTML.php#L2054-L2063

### DIFF
--- a/appendices/resources.xml
+++ b/appendices/resources.xml
@@ -13,7 +13,7 @@
    <tgroup cols="5">
      <colspec colwidth='1*'/>
      <colspec colwidth='1*'/>
-     <colspec colwidth='1.4*'/>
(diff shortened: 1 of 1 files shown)


Girgias (approved):
I mean I guess this works, can always alter it later.

PR #4491, opened 2025-02-24
memory-get-usage.xml Clarify what kind of memory we are talking about

mmalferov:
The current description confuses the reader and does not explain who allocates memory to whom. I'm not sure about the accuracy of the proposed formulations, but it seems that the changes better explain the difference between function calls without arguments or with the value `false`, or with the argument `true`.

I think that mentioning the engine, Zend Memory Manager (ZMM), or other internal details would be redundant, and the current description is enough to understand the difference.

Please check the proposal and make corrections to eliminate linguistic or semantic inaccuracies, if any, in the updated text.

And if the suggestion is OK, I will make the same changes with `memory_get_peak_usage()` ;-)

### DIFF
--- a/reference/info/functions/memory-get-usage.xml
+++ b/reference/info/functions/memory-get-usage.xml
@@ -3,9 +3,9 @@
 <refentry xml:id="function.memory-get-usage" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>memory_get_usage</refname>
-  <refpurpose>Returns the amount of memory allocated to PHP</refpurpose>
+  <refpurpose>Returns the amount of memory consumed by PHP script or allocated by system to PHP process</refpurpose>
  </refnamediv>
- 
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
@@ -13,8 +13,8 @@
    <methodparam choice="opt"><type>bool</type><parameter>real_usage</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Returns the amount of memory, in bytes, that's currently being
-   allocated to your PHP script.
+   Returns the amount of memory, in bytes, that the PHP script is
+   currently using.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/info/functions/memory-get-usage.xml
```diff
@@ -3,18 +3,18 @@
 <refentry xml:id="function.memory-get-usage" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>memory_get_usage</refname>
-  <refpurpose>Returns the amount of memory allocated to PHP</refpurpose>
+  <refpurpose>Returns the amount of memory consumed by PHP script or allocated by system to PHP process</refpurpose>
```
Crell:
```suggestion
  <refpurpose>Returns the amount of memory consumed by the PHP script or allocated by the system to PHP process</refpurpose>
```

reference/info/functions/memory-get-usage.xml
```diff
@@ -3,18 +3,18 @@
...
   &reftitle.description;
   <methodsynopsis>
    <type>int</type><methodname>memory_get_usage</methodname>
    <methodparam choice="opt"><type>bool</type><parameter>real_usage</parameter><initializer>&false;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Returns the amount of memory, in bytes, that's currently being
-   allocated to your PHP script.
+   Returns the amount of memory, in bytes, that the PHP script is
+   currently using.
```
Crell: This is a subtle change in meaning.  I'm not actually sure which is accurate.  The script could be using X bytes, but the OS has given it access to X+Y bytes.  (Especially if your script has cleaned up some memory.)

Can someone verify which value is measured here?  Tagging @Girgias

reference/info/functions/memory-get-usage.xml
```diff
@@ -54,18 +56,19 @@
     <programlisting role="php">
 <![CDATA[
 <?php
+
 // This is only an example, the numbers below will
-// differ depending on your system
+// differ depending on the specific system
 
-echo memory_get_usage() . "\n"; // 36640
+echo memory_get_usage(), "\n"; // 36640
```
Crell, line 77: These changes are stylistic, and I don't think necessary.

PR #4497, opened 2025-02-26
rewind.xml Give more details about the generator's rewinding

mmalferov:


philip:
Thanks for the work here, @mmalferov !

PR #4505, opened 2025-02-27
Update idn-to-ascii.xml: Adding info about conversion to lowercase

ThomasLandauer:
Page: https://www.php.net/manual/en/function.idn-to-ascii.php

This would close https://bugs.php.net/bug.php?id=80922

### DIFF
--- a/reference/intl/idn/idn-to-ascii.xml
+++ b/reference/intl/idn/idn-to-ascii.xml
@@ -137,7 +137,7 @@ xn--tst-qla.de
 <![CDATA[
 <?php
 
-echo idn_to_ascii('Example.com'); 
+var_dump(idn_to_ascii('Example.com'));
 
 ?>
 ]]>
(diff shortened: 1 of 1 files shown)


reference/intl/idn/idn-to-ascii.xml
```diff
@@ -129,6 +129,24 @@ echo idn_to_ascii('täst.de');
...
 xn--tst-qla.de
+]]>
+  </screen>
+  <example>
+   <title>All-ASCII domain names are just converted to lowercase</title>
+   <programlisting role="php">
+<![CDATA[
+<?php
+
+echo idn_to_ascii('Example.com');
```
Girgias, line 140: I usually prefer `var_dump` over a simple echo.

ThomasLandauer: OK

PR #4506, opened 2025-02-27
array_key_exists: Move BC-break note into changelog, make examples output friendly.

zimzat:


philip:
Thanks @zimzat for the useful contribution!

PR #4508, opened 2025-02-27
property_exists: Update example to show more scenarios.

zimzat:
The main thing to highlight is that property visibility and `__get` and `__isset` is not checked.

This example may be a little too complex and hides the intent. Would it be better to explicitly state the lack of visibility check in the description instead?


Girgias (commented):
I think it would be better to split this into a _new_ example, and have that one have a description to explain the intent

PR #4511, opened 2025-02-28
Add type declarations changelog for 7.0

vrana:


Girgias:
The PHP documentation assumes PHP 7.0.0 now, so any changelog for PHP 7.0.0 (and prior) should be removed.

vrana:
PHP 7.0 changes are still documented e.g. at https://www.php.net/manual/en/doc.changelog.php.

Girgias:
Those should be removed.

PR #4514, opened 2025-03-04
throw.xml Add an example

mmalferov:
Can we add an example that shows how the `Fiber::throw()` method passes an exception to the fiber?

### DIFF
--- a/language/predefined/fiber/throw.xml
+++ b/language/predefined/fiber/throw.xml
@@ -41,6 +41,43 @@
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


Girgias:
@trowski would you be able to write such an example?

trowski:
The example supplied looks fine. This was a PR, not an issue. 😄

lacatoire (approved):

jordikroon:
I have changed this to an `informalpara` as there is this is the default example. Plus removed the outer `<para>`.

PR #4520, opened 2025-03-07
Simplify xdiff_string_diff example

staabm:
since the method can only return `string`s, there is no point in checking the return type with `is_string`.


sy-records:
Returns string with resulting diff or false if an internal error happened. 

see https://www.php.net/manual/en/function.xdiff-string-diff.php#refsect1-function.xdiff-string-diff-returnvalues

staabm:
return type is wrong then

grafik

sy-records:
yep, see https://github.com/php/pecl-text-xdiff/blob/main/xdiff.stub.php#L4-L5

sy-records:
It did before, but was revert. https://github.com/php/doc-en/commit/4754397753fd79f1c846868b66a2448babab1c54

PR #4522, opened 2025-03-07
Improve $_POST/$_GET definition

pereorga:


language/predefined/variables/get.xml
```diff
@@ -46,7 +48,7 @@ Hello Hannes!
   &note.is-superglobal;
   <note>
    <para>
-    The GET variables are passed through <function>urldecode</function>.
+    The values in <varname>$_GET</varname> are automatically passed through <function>urldecode</function>.
```
pereorga, line 51: I checked if both the keys and values went through urldecode(), and I noticed this different behaviour. Not sure if this is documented, or whether it should be clarified:

```
$ curl http://localhost/test.php?user%20name=John%20Doe
Array
(
    [user_name] => John Doe
)
```

kamil-tekiela: Yes, it is well-known and documented here https://www.php.net/manual/en/language.variables.external.php

language/predefined/variables/get.xml
```diff
@@ -4,15 +4,17 @@
...
  <refsect1 role="description">
   &reftitle.description;
   <para>
-   An associative array of variables passed to the current script
-   via the URL parameters (aka. query string). Note that the array is  not only
-   populated for GET requests, but rather for all requests with a query string.
+   An associative array of variables passed to the current script via the URL
+   parameters (also known as the query string). Note that this array is
+   populated whenever a query string is present, regardless of the HTTP request
+   method. Even in POST, PUT, or other types of requests, query string
```
kamil-tekiela, line 16: Why the use of "even"? It seems like you are emphasising the same thing 3 times. Maybe a little less repetition?

language/predefined/variables/get.xml
```diff
@@ -29,7 +31,7 @@ echo 'Hello ' . htmlspecialchars($_GET["name"]) . '!';
 ]]>
     </programlisting>
     <simpara>
-     Assuming the user entered http://example.com/?name=Hannes
+     Assuming the user entered http://example.com/?name=Hannes.
```
kamil-tekiela, line 34: I think this requires something like `<literal>` or something to delimit the URL from the rest.

language/predefined/variables/post.xml
```diff
@@ -4,7 +4,7 @@
 <refentry role="variable" xml:id="reserved.variables.post" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>$_POST</refname>
-  <refpurpose>HTTP POST variables</refpurpose>
+  <refpurpose>Form Data from HTTP POST Requests</refpurpose>
```
kamil-tekiela, line 7: I don't know about that. It's often from a form. but HTTP doesn't have a concept of forms AFAIK. So technically this is incorrect. It should rather say something like "Data coming from HTTP body of a POST request".

pereorga: But not in all POST requests? The request needs to have application/x-www-form-urlencoded or multipart/form-data as the HTTP Content-Type, as per the documentation of this page. That's why I put "Form" (although it may not be the best solution)

language/predefined/variables/get.xml
pereorga: I agree it's too verbose, I'll remove the last sentence. The reason I wanted to be super clear is because the variable is so badly named that it may not be obvious that POST requests can have "HTTP GET variables" (quoting the current page)

language/predefined/variables/post.xml
kamil-tekiela: No, it can be any type. https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST

pereorga: This is about PHP behaviour, not the HTTP standard (see that page, and the comments). I just verified it doesn't work if you set something else as the content type:

Code:
```
<?php
header('Content-Type: text/plain');

echo "Request Method: " . $_SERVER['REQUEST_METHOD'] . "\n";
echo "POST Data:\n";
print_r($_POST);
```

OK:
```
$ curl -X POST -d "name=John" -H "Content-Type: application/x-www-form-urlencoded" http://localhost:8090/test.php
Request Method: POST
POST Data:
Array
(
    [name] => John
)

KO:
$ curl -X POST -d "name=John" -H "Content-Type: text/plain" http://localhost:8090/test.php
Request Method: POST
POST Data:
Array
(
)
```

(note that curl sets a content type by default, e.g. https://everything.curl.dev/http/post/content-type.html)

pereorga: what about: HTTP POST form-encoded data

kamil-tekiela: That's better, but it could make somebody think that they need to decode the data. Why can it not be just "HTTP POST data" or even left as it is now? We're not really making it easier to understand what it is.

pereorga: The problem is that `$_POST` only works with 2 form-specific content types. For example, it does not work with `application/json`, which is widely used in modern web APIs (RESTful, GraphQL, etc.). Because of this, I believe the original text leads to confusion.

kamil-tekiela: Right then, if this isn't mentioned anywhere in the manual yet, then we should make a note informing the users that `$_POST` is only populated in those two cases. Changing this little description is not enough.

pereorga: It is mentioned later in that page. I've now added a reference to `php://input`, and I've provided an example there that works with [`curl --json`](https://everything.curl.dev/http/post/json.html).

pereorga: For the `refpurpose`, I believe "Form data from HTTP POST requests" is the best alternative I've come up with.

language/wrappers/php.xml
```diff
@@ -368,6 +368,31 @@ file_put_contents("php://filter/write=string.rot13/resource=example.txt","Hello
...
+  <example>
+   <title>php://input to read JSON data from the request body</title>
+   <para>
+    This example demonstrates how to read JSON data from POST, PUT and PATCH
+    requests using <filename>php://input</filename>.
+   </para>
+   <programlisting role="php">
+<![CDATA[
+<?php
+header('Content-Type: text/plain');
```
kamil-tekiela, line 383: Please remove, not part of the example.

language/wrappers/php.xml
```diff
@@ -368,6 +368,31 @@ file_put_contents("php://filter/write=string.rot13/resource=example.txt","Hello
...
+
+// Read raw input data from php://input
```
kamil-tekiela, line 385: Redundant comment since that is exactly what the example is about.

language/wrappers/php.xml
```diff
@@ -368,6 +368,31 @@ file_put_contents("php://filter/write=string.rot13/resource=example.txt","Hello
...
+$jsonData = file_get_contents("php://input");
+
+// Decode JSON data into a PHP array
+$data = json_decode($jsonData, true);
```
kamil-tekiela, line 389: Use [JSON_THROW_ON_ERROR](https://www.php.net/manual/en/json.constants.php#constant.json-throw-on-error) instead of checking the return value.

language/wrappers/php.xml
```diff
@@ -368,6 +368,33 @@ file_put_contents("php://filter/write=string.rot13/resource=example.txt","Hello
...
+   <para>
+    This example demonstrates how to read raw JSON data from POST, PUT and
+    PATCH requests using <filename>php://input</filename>.
+   </para>
+   <programlisting role="php">
+<![CDATA[
+<?php
+$jsonString = file_get_contents("php://input");
+
+try {
```
kamil-tekiela, line 385: The try-catch isn't relevant to the example, either. In most cases, the developer probably will not want to catch the exception so the try catch isn't even suggesting a common scenario. It could lead to users blindly copying the whole example and assuming that try-catch is necessary. Please stick to only the bare essentials in the examples.

language/predefined/variables/post.xml
pereorga: Let me know if there's anything else you'd like to adjust.

PR #4524, opened 2025-03-09
Update constants.xml

sandacode:
The previous commit added the new error constant at the bottom of the page, instead of the section of the page listing error constants. Moved it there.


sy-records (approved):

sy-records:
Thank you!

PR #4535, opened 2025-03-13
objects does not have to be Traversable to be ForEach'ed.

### LINKED ISSUE #4534: ForEach does not issue an error for non-iterable non-Traversable objects
### Description

The following code:

```php
<?php
declare(strict_types=1);
error_reporting(E_ALL);
\set_error_handler(function ($errno, $errstr, $errfile, $errline) {
    if (\error_reporting() & $errno) {
        throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
    }
});
$a = ["foo" => "bar"];
$o = (object)$a;
$isIterable = is_iterable($o); // false
$isTraversable = $o instanceof Traversable; // false
var_dump(["isIterable" => $isIterable, "isTraversable" => $isTraversable]);
/*
Quoting https://www.php.net/manual/en/control-structures.foreach.php
The foreach construct provides an easy way to iterate over arrays and Traversable objects.
>>>foreach will issue an error when used with a variable containing a different data type<<<
*/
foreach ($o as $key => $val) {
    var_dump(["key" => $key, "val" => $val]);
}
?>

```

Resulted in this output:
```
array(2) {
  ["isIterable"]=>
  bool(false)
  ["isTraversable"]=>
  bool(false)
}
array(2) {
  ["key"]=>
  string(3) "foo"
  ["val"]=>
  string(3) "bar"
}
```

But given [the ForEach documentation](https://www.php.net/manual/en/control-structures.foreach.php), I expected this output instead:
```
array(2) {
  ["isIterable"]=>
  bool(false)
  ["isTraversable"]=>
  bool(false)
}
Fatal error: Uncaught TypeError: ForEach: Argument php/php-src#1 (iterable_expression) must be iterable or instance of Traversable
```
(some variant of that)

So it seems either the documentation or the implementation is wrong here.

### PHP Version

PHP 8.4.5

### Operating System

Ubuntu24.04

divinity76:
fix https://github.com/php/doc-en/issues/4534

### DIFF
--- a/language/control-structures/foreach.xml
+++ b/language/control-structures/foreach.xml
@@ -5,7 +5,7 @@
  <?phpdoc print-version-for="foreach"?>
  <para>
   The <literal>foreach</literal> construct provides an easy way to
-  iterate over <type>array</type>s and <interfacename>Traversable</interfacename> objects.
+  iterate over <type>array</type>s and objects.
   <literal>foreach</literal> will issue an error when used with
   a variable containing a different data type or with an uninitialized variable.
   <informalexample>
--- a/language/control-structures/foreach.xml
+++ b/language/control-structures/foreach.xml
@@ -26,7 +26,7 @@ foreach (iterable_expression as $key => $value) {
   </informalexample>
  </para>
  <simpara>
-  The first form traverses the iterable given by
+  The first form traverses the value given by
   <literal>iterable_expression</literal>. On each iteration, the value of
   the current element is assigned to <literal>$value</literal>.
  </simpara>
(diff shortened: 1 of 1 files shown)


language/control-structures/foreach.xml
```diff
@@ -5,7 +5,7 @@
  <?phpdoc print-version-for="foreach"?>
  <para>
   The <literal>foreach</literal> construct provides an easy way to
-  iterate over <type>array</type>s and <interfacename>Traversable</interfacename> objects.
+  iterate over <type>array</type>s and objects.
```
tiffany-taylor, line 8: Should specify that it's only object properties that can be iterated over.

PR #4536, opened 2025-03-13
set-error-handler.xml Remove the misleading statement

mmalferov:
For the method of the class as an error handler, the `set_error_handler` function may return a `callable`:

a) a string

b) a Closure

c) an indexed array (with the name of the class or an instance of the class in the first element, and the name of the method in the second one)

However, it is not limited to just "an indexed array with the class and method name".

I suggest removing the mention of the array as a return value, otherwise we will have to list every possible type of callable value :)


arnaud-lb (approved):
This looks good to me!

PR #4539, opened 2025-03-16
pcntl_signal: Switch to pcntl_async_signals and improve documentation

AllenJB:
This merges several user notes into the current documentation.

### DIFF
--- a/reference/pcntl/functions/pcntl-signal.xml
+++ b/reference/pcntl/functions/pcntl-signal.xml
@@ -195,7 +195,7 @@ echo "Done\n";
     <simplelist>
      <member>Asynchronous dispatch with <function>pcntl_async_signals</function> enabled. This is the recommended method</member>
      <member>Setting <link linkend="control-structures.declare.ticks">tick</link> frequency</member>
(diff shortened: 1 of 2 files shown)


reference/pcntl/functions/pcntl-signal.xml
```diff
@@ -189,13 +188,28 @@ echo "Done\n";
...
+   <title>Dispatch Methods</title>
+   <para>
+    There are several methods of dispatching signal handlers:
+    <simplelist>
+     <member>Asynchronous dispatch with <function>pcntl_async_signals</function> enabled. This is the recommended method</member>
+     <member>Setting <link linkend="control-structures.declare.ticks">tick</link> frequency</member>
+     <member>Manual dispatch with <function>pcntl_signal_dispatch</function></member>
+    </simplelist>
+   </para>
+   <para>When ticks are dispatched asynchronously or using ticks, blocking functions like <function>sleep</function> may be interrupted.</para>
```
devnexen, line 201: nit: I would prefer something like 

`When signals are dispatched asynchronously or using ticks-based execution, blocking functions like <function>sleep</function> may be interrupted.`

devnexen (approved):

PR #4541, opened 2025-03-16
See also is_link and symlink

vrana:


philip:
@vrana Nice to have you back, I think you should regain commit rights... a true phpdoc legend :)

PR #4546, opened 2025-03-18
Fix typo

msamesch:
Shouldn't this be tagged type instead of parameter?

### DIFF
--- a/reference/array/functions/array-key-exists.xml
+++ b/reference/array/functions/array-key-exists.xml
@@ -73,9 +73,9 @@
       <entry>8.0.0</entry>
       <entry>
        The <parameter>key</parameter> parameter now accepts
-       <parameter>bool</parameter>, <parameter>float</parameter>, <parameter>int</parameter>,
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-key-exists.xml
```diff
@@ -81,13 +81,13 @@
      <row>
       <entry>8.0.0</entry>
       <entry>
-       Passing an <parameter>object</parameter> to the <parameter>array</parameter> parameter is no longer supported.
+       Passing an <type>object</type> to the <parameter>array</parameter> parameter is no longer supported.
```
sy-records, line 84: You're right. L76-L78 also need to be modified.

```
       <type>bool</type>, <type>float</type>, <type>int</type>,
       <type>null</type>, <type>resource</type>, and
       <type>string</type> as arguments.
```

msamesch: Thanks for the review.

PR #4549, opened 2025-03-20
Fix buggy code examples for `ReflectionParameter::getAttributes`

daniil-berg:
- Fix calls to non-existent `getParameter` method of `ReflectionFunction`.
- Add missing `Red` attribute class definition implementing `Color`.

---

[Current page](https://www.php.net/manual/en/reflectionparameter.getattributes.php) in question.

Although I like the idea of being able to filter/get a specific function parameter by name, as far as I can tell, there is no such method on [`ReflectionFunctionAbstract`](https://www.php.net/manual/en/class.reflectionfunctionabstract.php) or its subclasses.

Also, the entire point of the [third example](https://www.php.net/manual/en/reflectionparameter.getattributes.php#example-5485) presumably was to illustrate that sub-type relationships are honored by passing the `IS_INSTANCEOF` flag to `getAttributes`. Therefore including the `Red` class in this example and making it implement `Color` is surely necessary.

### DIFF
--- a/reference/reflection/reflectionparameter/getattributes.xml
+++ b/reference/reflection/reflectionparameter/getattributes.xml
@@ -57,7 +57,7 @@ function fruitBasket(
 ) { }
 
 $reflection = new ReflectionFunction('fruitBasket');
-$parameter = $reflection->getParameter('apple');
+$parameter = $reflection->getParameters()[0];
 $attributes = $parameter->getAttributes();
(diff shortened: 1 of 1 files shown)


tiffany-taylor (approved):
LGTM, thank you for fixing this.

Using https://3v4l.org/KfXVW to show that the `getParameter()` method doesn't exist, and [with the proposed change](https://3v4l.org/kDBP8), the example executes without error.

PR #4550, opened 2025-03-20
Fix property hooks example

BackEndTea:
Property hooks are incompatible with read only properties, yet an example was given with a read only class.


TimWolla (approved):
Thanks!

PR #4556, opened 2025-03-22
Modernise parse_str description

IMSoP:
Most of the text still acted as though "create a bunch of local variables" was the normal behaviour, even though it's been discouraged for years, and isn't even supported in PHP 8.

There were also a few incorrect or questionable uses of markup.

### DIFF
--- a/reference/strings/functions/parse-str.xml
+++ b/reference/strings/functions/parse-str.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="function.parse-str" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>parse_str</refname>
-  <refpurpose>Parses the string into variables</refpurpose>
+  <refpurpose>Parse a string as a URL query string</refpurpose>
  </refnamediv>
  
  <refsect1 role="description">
@@ -15,8 +15,9 @@
   </methodsynopsis>
   <para>
    Parses <parameter>string</parameter> as if it were the query string
-   passed via a URL and sets variables in the current scope (or in the array
-   if <parameter>result</parameter> is provided).
+   passed via a URL and sets keys in the provided <parameter>result</parameter>
+   array. If no <parameter>result</parameter> is passed, values are instead
+   set as variables in the current scope.
   </para>
  </refsect1>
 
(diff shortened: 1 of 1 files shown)


reference/strings/functions/parse-str.xml
```diff
@@ -153,13 +157,14 @@ echo $output['My_Value']; // Something
 
   <note>
    <para>
-    All variables created (or values returned into array if second parameter is set)
-    are already <function>urldecode</function>d.
+    All values populated in the <parameter>result</parameter> array
+    (or variables created if second parameter is not set)
+    are already URL-decoded as though run through <function>urldecode</function>.
```
TimWolla, line 162:
```suggestion
    are already URL-decoded as if run through <function>urldecode</function>.
```
Not a native speaker, the suggestion is easier to understand for me / simpler language.

reference/strings/functions/parse-str.xml
```diff
@@ -118,9 +121,10 @@ echo $arr[1]; // baz
    </example>
   </para>
   <para>
-   Because variables in PHP can't have dots and spaces in their names,
-   those are converted to underscores. Same applies to naming of
-   respective key names in case of using this function with
+   Any spaces and dots in parameter names are converted to underscores
+   when creating array keys or local variables.
+   This is because variables in PHP can't have dots and spaces in their names,
```
TimWolla, line 126: Are contractions legal within docs or should this be "cannot"?

TimWolla (commented):
This is a clear improvement already.

reference/strings/functions/parse-str.xml
IMSoP: Reworded to be less abstract, hopefully simpler?

reference/strings/functions/parse-str.xml
IMSoP: Avoided both by using "are not allowed"

IMSoP:
Incidentally, it surprised me that this is in the "String Functions" section of the manual rather than "URL Functions" alongside `http_build_query`, which is its reverse (despite having a completely different name and signature).

Presumably that's because the manual is following the source code, and `PHP_FUNCTION(parse_str)` is defined in "string.c". I'm not sure what the implications of changing that would be.

TimWolla (approved):
I don't have further remarks, but defer to Girgias for final approval.

Girgias (approved):

PR #4559, opened 2025-03-23
random: Fix right-open / left-open confusion for Random\IntervalBoundary

TimWolla:
This was correct in my initial suggestions (https://github.com/php/doc-en/pull/3419#discussion_r1617712920), but incorrect in my later suggestions (https://github.com/php/doc-en/pull/3419#discussion_r1845547942) 😔 

---------------

see #3419


tiffany-taylor:
I should have commented before merging. Meant to say LGTM prior to merging.

PR #4579, opened 2025-04-03
Add example that shows array_chunk() within a loop and without a whole multiple

tiffany-taylor:
https://www.php.net/manual/en/function.array-chunk.php

The intent with this example is to show how to use `array_chunk()` within a loop, and what happens when the total amount of elements doesn't divide evenly with the `$length` argument.

### DIFF
--- a/reference/array/functions/array-chunk.xml
+++ b/reference/array/functions/array-chunk.xml
@@ -159,6 +159,55 @@ Array
     </screen>
    </example>
   </para>
+  <para>
+   <example>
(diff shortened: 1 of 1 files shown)


lacatoire:
I am closing this since there has been no follow-up. Both points raised above hold: the call is in the `foreach` header, so it runs once before iteration rather than inside the loop, and the partial-chunk case is already shown by the existing example and stated in the description. Happy to see this reopened around a case the page doesn't already cover.

PR #4591, opened 2025-04-09
Update ibase-trans.xml

MartinKoeditz:
Removed wrong option. You have to put transaction args at first place.


vrana:
[Source code](https://github.com/FirebirdSQL/php-firebird/blob/93c8e50221bf62f23a7668cbd826bbb64073d6b1/interbase.c#L1289) suggests that the order could be arbitrary.

PR #4604, opened 2025-04-10
Swoole: Fix typo in Swoole\Server::__construct

AkameOuO:
Reference: https://wiki.swoole.com/en/#/server/methods?id=__construct


sy-records (approved):
Thanks!

PR #4605, opened 2025-04-10
Add back authors previously featured on the frontpage

vrana:
I just couldn't believe that @haszi removed people with the biggest contributions in fde7fc4. Only two of them were added in 169faf3.

### DIFF
--- a/contributors.xml
+++ b/contributors.xml
@@ -16,6 +16,7 @@
 
   <para>
    &Credit.Past.Authors.Text;
+   Mehdi Achour,            <!-- didou -->
    Bill Abt,                <!-- pdo howto via ibm -->
    Jouni Ahto,              <!-- jah -->
    Alexander Aulbach,       <!-- ssilk [at] ssilk.de  [no cvs user] -->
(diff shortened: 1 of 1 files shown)


contributors.xml
```diff
@@ -53,15 +57,17 @@
...
    Sascha Schumann,         <!-- sas -->
    Dan Scott,               <!-- dbs -->
+   Damien Seguy,            <!-- dams -->
    Masahiro Takagi,         <!-- takagi -->
    Yoshinari Takaoka,       <!-- mumumu -->
    Yannick Torres,          <!-- yannick -->
+   Jakub Vrana,             <!-- vrana -->
    Michael Wallner,         <!-- mike -->
    Lars Torben Wilson,      <!-- torben and also cslawi -->
    Jim Winstead,            <!-- jimw and also jim(?) -->
```
jimwins, line 73: Not necessarily related to this change, but I'm pretty sure I've only ever been `jimw` in PHP-related work, and there hasn't been a bare `jim` with a VCS account.

jimwins (approved):

PR #4621, opened 2025-04-19
wkhtmltox: Inline invalid XML

vrana:
I process the manual sources externally and it's a PITA to work with invalid XML files. This inlines the only invalid files we have and XIncludes it to the other place.

This seems also cleaner, we use this approach elsewhere, e.g. [in bcdiv](https://github.com/php/doc-en/blob/master/reference/bc/functions/bcdiv.xml#L42).

Cc @krakjoe.


lacatoire (approved):

jordikroon:
Thank you for this change.

PR #4624, opened 2025-04-21
Fix return type of SolrUtils::escapeQueryChars

bix0r:
As noted in https://github.com/php/pecl-search_engine-solr/issues/83, and fixed in https://github.com/php/pecl-search_engine-solr/pull/84 the method actually never returns anything else then a string.


vrana:
It can still [return false](https://github.com/php/pecl-search_engine-solr/blob/9a8452d7ddc06f18f97752b34198d9b36499d29d/src/php_solr_utils.c#L31).

PR #4625, opened 2025-04-22
Update decrement.xml

Elompenta:
Fix typo (missing spacer) and allign naming to the sister-function increment()
https://github.com/php/doc-en/blob/master/reference/memcache/memcache/increment.xml


kamil-tekiela:
Thanks for trying to contribute, but I don't understand what you were trying to do here. The new sentence is grammatically incorrect, technically wrong, and contains unnecessary whitespace.

PR #4631, opened 2025-04-25
Use explicit nullable type in language/exceptions.xml

benfaerber:
PHP 8.4 deprecates inferred nullable parameters. Many people copy this code directly into their codebase so it's good to make sure it won't break anything soon.


TimWolla (approved):
Thanks!

PR #4633, opened 2025-04-25
Sort INI options

vrana:
https://bugs.php.net/80315
Created by a script: https://gist.github.com/vrana/eb84b267c3352b7e9f9adca888b8665b

### DIFF
--- a/appendices/ini.core.xml
+++ b/appendices/ini.core.xml
@@ -31,35 +31,15 @@
       </thead>
       <tbody>
        <row>
-        <entry><link linkend="ini.short-open-tag">short_open_tag</link></entry>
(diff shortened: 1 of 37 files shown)


lacatoire:
@vrana Sorry for the long delay on this one. The change is correct, and publishing the script makes it easy to redo, which is what this needs. Two things stand in the way of merging it as is.

15 of the 37 files belong to extensions that are migrating to [php/doc-extensions](https://github.com/php/doc-extensions/). Those would need a separate pull request against `php/doc-extensions`.

The rest conflicts with master, which is unavoidable for a 966 line patch across that many files. Rerunning the script on current master would be quicker than resolving it.

PR #4637, opened 2025-04-26
Fix return values of streamWrapper functions

nielsdos:


vrana:
https://github.com/php/php-src/blob/php-8.4.6/main/streams/userspace.c#L1327 suggests that it might return also `true`.

nielsdos:
Interesting find. I'm not sure it's a good idea to document that insane behaviour as it would promote people to use that. I prefer to stick to `false` for errors. But cc'ing @Girgias just in case

Girgias:
I think documenting the `bool` type with a massive warning to indicate that both `true` and `false` ignore casting, but it is not recommended to return `true` is the correct thing to do.

But this is very much insane behaviour

nielsdos:
Thank you both. I pushed a commit to clarify this in accordance to Girgias's comment. Let me know if this works.

Girgias (approved):

reference/stream/streamwrapper/stream-cast.xml
```diff
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>resource</type><methodname>streamWrapper::stream_cast</methodname>
+   <modifier>public</modifier> <type class="union"><type>resource</type><type>false</type></type><methodname>streamWrapper::stream_cast</methodname>
```
vrana, line 13: https://github.com/php/php-src/blob/php-8.4.6/main/streams/userspace.c#L1416 suggests that the special value is `true`, not `false`.

nielsdos: Hm, it checks truthiness and if it's not true, it does not continue. So the special value is `false` (or technically, anything that evaluates to a falsey value).

PR #4640, opened 2025-04-27
IntlDateFormatter::localtime: Fix example

### LINKED ISSUE #4636: IntlDateFormatter::localtime broken example
**Affected page**

https://www.php.net/manual/en/intldateformatter.localtime.php

**Issue description**

Run the example code, you'll get an error. This makes sense because the $offset is passed as an integer but you can't because the argument is a reference.

**Steps to reproduce**

Run the examples.

**Suggested fix**

Probably introduce a variable $offset, set it to 0, and pass that.

leonardolara:
Date/time format adjusted to match the specified one in the first line.
Last parameter is a reference.


leonardolara:
Closes #4636

PR #4644, opened 2025-04-30
these are not available in 32bit pack: q, Q, J, P

divinity76:
see https://github.com/php/php-src/blob/156d034d3da314925a77115538b55c7097c662cc/ext/standard/pack.c#L320-L330

### DIFF
--- a/reference/misc/functions/pack.xml
+++ b/reference/misc/functions/pack.xml
@@ -255,6 +255,9 @@ $binarydata = pack("nvc*", 0x1234, 0x5678, 65, 66);
  <refsect1 role="notes">
   &reftitle.notes;
   <caution>
+    <para>Warning: format codes <literal>q</literal>, <literal>Q</literal>, <literal>J</literal> and <literal>P</literal> are not available on 32-bit PHP builds.</para>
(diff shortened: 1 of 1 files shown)


nielsdos:
Thanks!

Note that this renders a bit weirdly because of the explicit "Warning: " you put in front.

Screenshot from 2025-05-06 22-22-58

I think it's better to:
* Drop "Warning: "
* Use simpara tag instead of para.

divinity76:
@nielsdos done!

reference/misc/functions/pack.xml
```diff
@@ -254,6 +254,9 @@ $binarydata = pack("nvc*", 0x1234, 0x5678, 65, 66);
  
  <refsect1 role="notes">
   &reftitle.notes;
+  <caution>
+    <simpara>format codes <literal>q</literal>, <literal>Q</literal>, <literal>J</literal> and <literal>P</literal> are not available on 32-bit PHP builds.</simpara>
```
nielsdos, line 258: Capital letter for f? :)

nielsdos (approved):
Thanks!

PR #4646, opened 2025-05-03
Language snippets: fix `odbc.changelog.connection-return`

DanielEScherzer:
Don't include `odbc.changelog.connection-param`, not every function that returns a connection has a connection parameter (and, in fact, neither of the existing uses of the connection-return entity, for `odbc_connect()` and `odbc_pconnect()`, has such a parameter).


DanielEScherzer:
If there are no objections I plan to merge this next time I remember it exists

DanielEScherzer:
I remembered again, and confirmed it is still just those two uses

PR #4647, opened 2025-05-03
String comparison / similarity doc improvements

AllenJB:
Improved the documentation of strcmp to make it clear it's not locale aware and direct to alternatives.

Improved the strcmp see also list:

* Broke the list up into clear categories
* Add the string similarity functions
* Removed substr because I don't think it belongs here

Improved the see also lists for the the string similarity functions by cross-linking them all.

### DIFF
--- a/reference/strings/functions/strcmp.xml
+++ b/reference/strings/functions/strcmp.xml
@@ -14,10 +14,12 @@
    <methodparam><type>string</type><parameter>string2</parameter></methodparam>
   </methodsynopsis>
   <simpara>
-   Note that this comparison is case sensitive. For case-insensitive comparison, see <function>strcasecmp</function>.
+   Note that this comparison is case sensitive. For case-insensitive comparison,
+   see <function>strcasecmp</function>.
   </simpara>
   <simpara>
-   Note that this comparison is not locale aware. See <function>strcoll</function> or <methodname>Collator::compare</methodname>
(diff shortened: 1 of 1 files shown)


reference/strings/functions/strcmp.xml
```diff
@@ -14,7 +14,10 @@
    <methodparam><type>string</type><parameter>string2</parameter></methodparam>
   </methodsynopsis>
   <simpara>
-   Note that this comparison is case sensitive.
+   Note that this comparison is case sensitive. See <function>strcasecmp</function>
```
kamil-tekiela:
```suggestion
   Note that this comparison is case sensitive. For case-insensitive comparison, see <function>strcasecmp</function>.
```

reference/strings/functions/strcmp.xml
```diff
@@ -14,7 +14,10 @@
...
+  </simpara>
+  <simpara>
+   Note that this comparison is not locale aware. See <function>strcoll</function> or <methodname>Collator::compare</methodname>
```
kamil-tekiela: Same as above, please explain why one should see these functions.

kamil-tekiela (changes requested):

reference/strings/functions/strcmp.xml
AllenJB: Updated.

Crell (approved):

Crell:
Thanks!

PR #4650, opened 2025-05-05
$php_errormsg was removed in PHP 8.0

DanielEScherzer:
See php/php-src@920b4b249f71e6cbfd795f81c6a08126a33c659e
Use `&warn.deprecated.feature-7-2-0.removed-8-0-0;` instead of `&warn.deprecated.feature-7-2-0;`


TimWolla (approved):
It's also already in the changelog on this page.

PR #4653, opened 2025-05-05
SensitiveParameterValue::getValue() is documented, remove warning

DanielEScherzer:


language/predefined/sensitiveparametervalue/getvalue.xml
```diff
@@ -14,9 +14,6 @@
   <para>
```
TimWolla, line 15: It isn't actually documented: A description is missing.

DanielEScherzer: I guess I thought the "Returns the sensitive value" part was a description, but I've added one - if you approve this I'll squash with a nice message

TimWolla (approved):
I had troubles writing an appropriate description (that's why it was missing). I guess that's okay, there's only *so much* you can explain here.

PR #4654, opened 2025-05-05
Array type: Add a top of page link to the array functions docs

AllenJB:
It's not uncommon to end up on the array type documentation page - for example, if you go to https://php.net/array - when you really want the functions documentation.

While there are links further down the page, you currently need to know they're there (and where they are), and then scroll down to them for access.

A top of page link allows for users to quickly get to where they want to go.

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -4,7 +4,7 @@
  <title>Arrays</title>
 
  <para>
-  See also the <link linkend="ref.array">array functions</link> documentation.
+  For a list of all array functions, see the <link linkend="ref.array">array functions</link> chapter in the documentation.
  </para>
 
  <para>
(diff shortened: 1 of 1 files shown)


language/types/array.xml
```diff
@@ -3,6 +3,10 @@
 <sect1 xml:id="language.types.array">
  <title>Arrays</title>
 
+ <para>
+  See also the <link linkend="ref.array">array functions</link> documentation.
```
kamil-tekiela:
```suggestion
  For a list of all array functions, see the <link linkend="ref.array">array functions</link> chapter in the documentation.
```

kamil-tekiela (changes requested):

kamil-tekiela (approved):

PR #4658, opened 2025-05-06
Document `Pdo\Dblib::ATTR_DATETIME_CONVERT`

NattyNarwhal:
I've had users ask me about how to do this in PDO_DBLIB, and it turns out it can do this, but it's not documented. Do that.


jordikroon:
@NattyNarwhal Could you rebase this with master?

PR #4664, opened 2025-05-08
random: clarify that `getrandmax` was small only before PHP 7.1

takaram:
Since PHP 7.1 `getrandmax` is an alias of `mt_getrandmax`, which returns 2**31 - 1


TimWolla (approved):
Thanks!

PR #4666, opened 2025-05-10
optimize swoole docs

NathanFreeman:


reference/swoole/ini.xml
```diff
@@ -18,58 +18,40 @@
     </thead>
     <tbody>
      <row>
-      <entry><link linkend="ini.swoole.aio-thread-num">swoole.aio_thread_num</link></entry>
-      <entry>2</entry>
+      <entry><link linkend="ini.swoole.enable_library">swoole.enable_library</link></entry>
```
sy-records, line 21: Use `-` instead of `_` in xml:id

sy-records (approved):

PR #4696, opened 2025-05-22
Update the definition of the Z pack format code

27pchrisl:
a and Z had the same definition, but Z is the NUL-terminated variation.

### DIFF
--- a/reference/misc/functions/pack.xml
+++ b/reference/misc/functions/pack.xml
@@ -166,7 +166,7 @@
           </row>
           <row>
            <entry>Z</entry>
-           <entry>NUL-terminated (ASCIZ) string, will be NUL padded</entry>
+           <entry>NUL-terminated (ASCIIZ) string, will be NUL padded</entry>
           </row>
           <row>
            <entry>@</entry>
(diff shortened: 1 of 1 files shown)


nielsdos:
It's actually NUL terminated, _and_ padded in some cases: https://3v4l.org/WGYGn
Not sure how to communicate this in the best way

27pchrisl:
Perl’s doc says “A null-terminated (ASCIZ) string, will be null padded.”, that seems clear? https://perldoc.perl.org/functions/pack. - could use their definition for A as well to make the distinction

nielsdos:
Yes that description is better

reference/misc/functions/pack.xml
```diff
@@ -166,7 +166,7 @@
           </row>
           <row>
            <entry>Z</entry>
-           <entry>NUL-padded string</entry>
+           <entry>NUL-terminated (ASCIZ) string, will be NUL padded</entry>
```
ndossche: ASCII is with double I
```suggestion
           <entry>NUL-terminated (ASCIIZ) string, will be NUL padded</entry>
```

nielsdos (approved):
Thanks

PR #4704, opened 2025-05-24
Rephrased advice in Heredoc string example 6

Andrew-Staves-Activ:
for clarity.

### DIFF
--- a/language/types/string.xml
+++ b/language/types/string.xml
@@ -397,8 +397,8 @@ Parse error: syntax error, unexpected identifier "ING", expecting "]" in example
 
     <simpara>
      To avoid this problem, it is safe to follow the simple rule:
-     <emphasis>do not choose a word that appears in the body
-     of the text as a closing identifier</emphasis>.
+     <emphasis>do not choose a word that appears in the body of the text
+     as a closing identifier</emphasis>.
     </simpara>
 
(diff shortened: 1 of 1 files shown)


language/types/string.xml
```diff
@@ -397,8 +397,8 @@ Parse error: syntax error, unexpected identifier "ING", expecting "]" in example
 
     <simpara>
      To avoid this problem, it is safe to follow the simple rule:
-     <emphasis>do not choose as a closing identifier if it appears in the body
-     of the text</emphasis>.
+     <emphasis>do not choose a word that appears in the body
+     of the text as a closing identifier</emphasis>.
```
tiffany-taylor: It should be possible to move "of the text" from line 401 onto the end of line 400, without going over the line length that we use per style guide.

```suggestion
     <emphasis>do not choose a word that appears in the body of the text
     as a closing identifier</emphasis>.
```

Andrew-Staves-Activ: I've applied the suggestion, thanks. I don't see an obvious way to set this pull request to squash the commits upon merging (like GitLab has), otherwise I would set that. Perhaps that's due to a (quite reasonable) lack of privileges to merge though.

Please let me know if you need me to manually squash the commits in the branch.

PR #4707, opened 2025-05-27
fix: typo in Constructors and Destructors example #5

adjsky:


Fan2Shrek:
Nice catch! Thanks you :)

PR #4712, opened 2025-05-30
Fix class for getIanaID method

SergiiDolgushev:
https://github.com/php/php-src/blob/3cb7d1bd8ac352d7c97237e509da3b4c1db1aed8/ext/intl/timezone/timezone.stub.php#L119


sy-records (approved):

devnexen (approved):
Thanks !

PR #4721, opened 2025-06-08
Update json-last-error-msg.xml - json_validate()

juan-morales:
adding json_validate()


reference/json/functions/json-last-error-msg.xml
```diff
@@ -13,7 +13,7 @@
    <void/>
   </methodsynopsis>
   <para>
-   Returns the error string of the last <function>json_encode</function> or <function>json_decode</function>
+   Returns the error string of the last <function>json_validate()</function>, <function>json_encode</function> or <function>json_decode</function>
```
ndossche: Function names should not have () at the end such that linking works

```suggestion
   Returns the error string of the last <function>json_validate</function>, <function>json_encode</function> or <function>json_decode</function>
```

PR #4733, opened 2025-06-19
Reverse hash_equals parameters

dmgsoftware:
As stated in the document, the user provided string is supposed to be the second parameter. Whereas before this pull request, the user provided string is the first parameter.


TimWolla:
This change is incorrect. The `$signature` is the user-provided string, which even is explicitly stated in the comment.

dmgsoftware:
Sorry about the confusion. I should have paid closer attention to the example.

PR #4738, opened 2025-06-22
#[\Override]: document restriction on __construct()

DanielEScherzer:
Prompted by a note on the manual page

### DIFF
--- a/language/predefined/attributes/override.xml
+++ b/language/predefined/attributes/override.xml
@@ -20,7 +20,7 @@
    <simpara>
     The attribute cannot be used on the
     <link linkend="object.construct">__construct()</link>
-    method.
+    method, which is exempt from signature checks.
    </simpara>
   </section>
 


DanielEScherzer:
This was part of the details on https://wiki.php.net/rfc/marking_overriden_methods but not noted in the docs page

language/predefined/attributes/override.xml
```diff
@@ -17,6 +17,11 @@
     in an implemented interface a compile-time error will be
     emitted.
    </simpara>
+   <simpara>
+    The attribute cannot be used on a
+    <link linkend="object.construct">__construct()</link>
+    method.
```
TimWolla:
```suggestion
    The attribute cannot be used on the
    <link linkend="object.construct">__construct()</link>
    method.
```
From what I see the definite article is used everywhere else and it also feels more correct to me, since an individual class can only have one constructor.

Perhaps also with short explanation:

> […], as `__construct()` is exempt from signature checks.

or something like this?

TimWolla (commented):
No particularly strong feelings on the exact wording, but I'd like to see an explanation for the restriction. Feel free to merge after adding something.

TimWolla (approved):

language/predefined/attributes/override.xml
DanielEScherzer: Done

PR #4739, opened 2025-06-22
Namespaces: note that enums are affected

DanielEScherzer:
Prompted by a note on the manual page

### DIFF
--- a/language/namespaces.xml
+++ b/language/namespaces.xml
@@ -91,7 +91,8 @@ echo constant($d); // see "Namespaces and dynamic language features" section
   <?phpdoc print-version-for="namespaces"?>
   <para>
    Although any valid PHP code can be contained within a namespace, only the following
-   types of code are affected by namespaces: classes (including abstracts and traits), interfaces, functions and constants.
+   types of code are affected by namespaces: classes (including abstracts and traits), interfaces, enums (since PHP 8.1),
+   functions and constants.
   </para>
   <para>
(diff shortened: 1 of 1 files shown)


DanielEScherzer:
See https://www.php.net/manual/en/language.namespaces.definition.php#130252

language/namespaces.xml
```diff
@@ -91,7 +91,8 @@ echo constant($d); // see "Namespaces and dynamic language features" section
   <?phpdoc print-version-for="namespaces"?>
   <para>
    Although any valid PHP code can be contained within a namespace, only the following
-   types of code are affected by namespaces: classes (including abstracts and traits), interfaces, functions and constants.
+   types of code are affected by namespaces: classes (including abstracts and traits), interfaces, enums (since PHP 8.1),
+   functions and constants.
```
TimWolla:
```suggestion
   types of code are affected by namespaces: classes (including abstract classes, traits and enums), interfaces,
   functions and constants.
```

The version indication implied to me what enums were there before 8.1, but could only be used in namespaces with 8.1. I think leaving out the version here is just fine, since when you don't have enums, clearly they can't be in a namespace.

Also, either both traits and enums should be mentioned separately, or they both should be considered a subset of classes. No strong feeling either way. Enums are more similar to classes than traits are, though.

DanielEScherzer: Added your suggested language

PR #4740, opened 2025-06-22
fix(reference/stream): HTTP Headers are U+0D U+0A (CRLF) separated

hakre:
compare Section 4.1 of RFC-2616 etc.

        generic-message = start-line
                          *(message-header CRLF)
                          CRLF
                          [ message-body ]

RFC-2616: https://www.rfc-editor.org/rfc/rfc2616#section-4.1
ok: /reference/stream/functions/stream-context-set-default.xml etc.

### DIFF
--- a/language/context/http.xml
+++ b/language/context/http.xml
@@ -208,19 +208,19 @@
 <?php
 
 $postdata = http_build_query(
-    array(
+    [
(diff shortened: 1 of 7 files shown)


nielsdos (approved):
Nice catch, and nice improvements

Issue #4355, opened 2024-12-31
Callables: clarify behavior on instance callables


Vectorial1024:
From manual page: https://www.php.net/manual/en/language.types.callable.php

---

While debugging why a piece of code involving object instance callables acts as if it was never called, I have discovered a behavior that is only implied by the docs. It is best if this behavior is directly pointed out.

Consider:

```php
class SignalListener {
    public function __construct() {
        pcntl_signal(SIGTERM, [$this, 'signalHandler']);
    }

    private function signalHandler() {
        echo 'signal!';
    }
}
```

Then:

```php
$listener = new SignalListener();
posix_kill(getmypid(), SIGTERM);
```

One may expect that `$listener->signalHandler()` is called since the callable might actually be referenced directly by the runtime somehow, so perhaps the PCNTL function in question might actually be able to call `$listener->signalHandler()`. This misconception therefore prompts the following expected output:

```
signal!
```

...but actually, this triggers the following error:

```
Invalid callback SignalListener::signalHandler, cannot access private method SignalListener::signalHandler [...]
```

------

Currently, the docs mention that "accessing protected and private methods from within a class is allowed", but it does not mention what happens when protected/private methods are being passed as callables to outside a class, such as the PCNTL module functions that exists freely with no namespaces or classes.

It is best if the docs can explicitly mention that protected/private members inside a class are still not callable outside its own class even via the callables syntax.

It does not seem like this behavior will change, and this can be a gotcha moment for inexperienced devs, so this is worthy of being put into the official docs.

damianwadley:
The answer is fairly straightforward: `callable` means the thing receiving the array (or whatever) is able to call it. In other words, this
```php
function execute_the_callable(callable $thing) {
    $thing(/* whatever args are necessary */);
}
```
should always be able to work. If the thing receiving the "callable" isn't able to invoke the value then the value isn't valid as a `callable`.

So while `[$this, "signalHandler"]` is certainly callable from within the SignalListener class, it's referencing a private method so it won't be callable from outside that class. Including from pcntl_signal [1]. So that means it won't be valid for pcntl_signal's `callable` second argument.
For the record, you can avoid having to make the handler public by (a) using an inline function instead of a discrete method on the class, as in `pcntl_signal(SIGTERM, function() { ... })` [2], or (b) using [Closure::fromCallable](https://www.php.net/manual/en/closure.fromcallable.php) [3] to create a Closure object which you can then pass as the handler argument.

[...]

PR #4752, opened 2025-06-28
Improve documentation on callables

Vectorial1024:
This implements #4355 :
- added a note that warns about passing protected/private class/instance methods
- provide additional examples of manually invoking the callbacks

The goal is that, by providing examples of how callbacks may be manually used, the reader becomes convinced that passing protected/private methods as callables to the outside world is generally a bad idea.

### DIFF
--- a/language/types/callable.xml
+++ b/language/types/callable.xml
@@ -205,6 +205,25 @@ class NumberTripler
 // prints "15"
 $tripler = new NumberTripler();
 processNumber($tripler, 5);
+
+// define a class with a private instance method, to be called with the callable syntax
+class HiddenFactory
+{
+    private function hiddenWork()
+    {
+        echo "256" . PHP_EOL;
+    }
+
+    public function getProof()
+    {
+        return Closure::fromCallable([$this, 'hiddenWork']);
+    }
+}
+
+// prints "256"
+$factory = new HiddenFactory();
+$theProof = $factory->getProof();
+invokeCallback($theProof);
 ?>
     </programlisting>
     &example.outputs;
@@ -213,6 +232,7 @@ processNumber($tripler, 5);
 42
 72
 15
+256
 ]]>
     </screen>
    </example>


PhilDaiguille (approved):

language/types/callable.xml
```diff
@@ -30,10 +30,18 @@
...
-   method name at index 1. Accessing protected and private methods from
-   within a class is allowed.
+   method name at index 1. Passing protected and private methods
+   with the <type>callable</type> syntax is allowed.
   </para>
 
+  <note>
+   <para>
+    If a callable of a protected or a private method is invoked from outside the class
+    with no visibility of said methods (e.g. being called from a POSIX signal handler),
```
Crell, line 40: A POSIX signal handler is a very niche example here. I know it's the one that prompted this PR, but I'd use something more common or self-evident.

language/types/callable.xml
```diff
@@ -156,6 +164,80 @@ print implode(' ', $new_numbers);
...
+class HiddenFactory
+{
+    private function hiddenWork()
+    {
+        echo "256" . PHP_EOL;
+    }
+
+    public function getProof()
+    {
+        return Closure::fromCallable([$this, 'hiddenWork']);
```
Crell, line 219: Using `$this->hiddenWork(...)` is preferred these days.

language/types/callable.xml
```diff
@@ -156,6 +164,80 @@ print implode(' ', $new_numbers);
    </example>
   </para>
 
+  <para>
+   <example>
+    <title>
+     Manually invoking callbacks
```
Crell, line 170: Is that the appropriate title here?  The example feels a big long and involved, but I don't quite know what it's trying to demonstrate.

language/types/callable.xml
```diff
@@ -30,10 +30,18 @@
...
-   within a class is allowed.
+   method name at index 1. Passing protected and private methods
+   with the <type>callable</type> syntax is allowed.
   </para>
 
+  <note>
+   <para>
+    If a callable of a protected or a private method is invoked from outside the class
+    with no visibility of said methods (e.g. being called from a POSIX signal handler),
+    a runtime error is thrown because the callable is invalid at the time of invocation.
```
kamil-tekiela, line 41: What kind of error?

language/types/callable.xml
```diff
@@ -156,6 +164,80 @@ print implode(' ', $new_numbers);
...
+// prints "15"
+$tripler = new NumberTripler();
+processNumber($tripler, 5);
+
+// define a class with a private instance method, to be called with the callable syntax
+class HiddenFactory
+{
+    private function hiddenWork()
+    {
+        echo "256" . PHP_EOL;
```
kamil-tekiela, line 214: Please don't use `PHP_EOL` for displaying newline. This constant is for parsing text. Use `"\n"`.

language/types/callable.xml
kamil-tekiela: I don't think this means what you wanted to say. A callable or a closure of a private member can be invoked from outside the class. The problem is only with the pseudo-callables in the form of array which are converted into the actual closure at the invokation time only and therefore refer to an inaccessible member. 

This page should probably be entirely rewritten to explain the modern concepts better.

language/types/callable.xml
Crell: Uh, what?  No, `PHP_EOL` is entirely fine to use for output.

kamil-tekiela: It creates inconsistent output on different platforms and could confuse new users. The standard `"\n"` is shorter, unchangeing and universally understood by most programmers. We should avoid using `PHP_EOL` in the PHP manual.

Vectorial1024:
@kamil-tekiela I do receive emails on comments of this PR, but often times I switched to look at something else and then forgot to come back to this. I'll take a look at the new PR.

PR #4753, opened 2025-06-28
Clarify purpose of $data in updateTimestamp()

lt-columbo:
The current documentation only says "$data — The session data" without context.

This update explains that $data contains the serialized session and may be used  to conditionally update the session expiry. It also clarifies that most  implementations can ignore it.


lacatoire (approved):

lacatoire (approved):

lacatoire:
LGTM, adds useful context without overcomplicating the documentation.
see https://github.com/php/php-src/blob/master/ext/session/session.c
Thnks @It-columbo

PR #4765, opened 2025-07-08
document setlocale(LC_ALL, null) to retrieve current setting

remicollet:
Document how it works (since always)

ping @nielsdos as related to https://github.com/php/php-src/issues/19070

### DIFF
--- a/reference/strings/functions/setlocale.xml
+++ b/reference/strings/functions/setlocale.xml
@@ -11,7 +11,7 @@
   <methodsynopsis>
    <type class="union"><type>string</type><type>false</type></type><methodname>setlocale</methodname>
    <methodparam><type>int</type><parameter>category</parameter></methodparam>
-   <methodparam><type>string</type><parameter>locales</parameter></methodparam>
+   <methodparam><type class="union"><type>string</type><type>null</type></type><parameter>locales</parameter></methodparam>
    <methodparam rep="repeat"><type>string</type><parameter>rest</parameter></methodparam>
   </methodsynopsis>
   <simpara>Alternative signature (not supported with named arguments):</simpara>
@@ -102,7 +102,7 @@
        categories, or from "LANG".
       </para>
       <para>
-       If <parameter>locales</parameter> is <literal>"0"</literal>,
+       If <parameter>locales</parameter> is &null;,
        the locale setting is not affected, only the current setting is returned.
       </para>
       <para>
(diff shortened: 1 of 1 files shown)


reference/strings/functions/setlocale.xml
```diff
@@ -102,7 +102,7 @@
        categories, or from "LANG".
       </para>
       <para>
-       If <parameter>locales</parameter> is <literal>"0"</literal>,
+       If <parameter>locales</parameter> is &null;,
```
ndossche, line 105: This is right, but "0" works too.

reference/strings/functions/setlocale.xml
```diff
@@ -184,6 +184,21 @@ echo strftime("%A %e %B %Y", mktime(0, 0, 0, 12, 22, 1978));
...
 echo "Preferred locale for german on this system is '$loc_de'";
 ?>
+]]>
+    </programlisting>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title><function>setlocale</function> Retrieve current setting</title>
+    <programlisting role="php" annotations="non-interactive">
```
ndossche, line 194: Is this actually non-interactive?

reference/strings/functions/setlocale.xml
ndossche: Ah no, I now see your comment on the php-src issue, nvm this!

reference/strings/functions/setlocale.xml
remicollet: Don't know (1st time I see this attribute)
This is a copy-paste of the other example ;)

ndossche: It is for enabling running the example code on the docs website. But okay, if the other does not have it, it likely has a reason.

nielsdos:
Unfortunately, it seems more complicated than that.

A locale of "0" maps to NULL in the C code:
https://github.com/php/php-src/blob/c4183fba003a8601fcdf4d9530fa2f38e3e0c100/ext/standard/string.c#L4850-L4852

This means that `setlocale(..., "0")` in PHP does `setlocale(..., NULL)` in C.
NULL actually means returning the current locale without changing it. So the docs description was correct.

Passing NULL in PHP transforms it to the string "" in the C code, i.e. `setlocale(..., NULL)` in PHP does `setlocale(..., "")` in C, which means setting the locale to the user-preferred locale.

The difference between C/PHP should be cleared up in this page.

remicollet:
Sorry, I merged the fix before seing your last comment.

> This means that setlocale(..., "0") in PHP does setlocale(..., NULL) in C.

Yes, I noticed this part of the code, But this doesn't work

```
$ sapi/cli/php -r 'var_dump(setlocale(LC_ALL, null));'
string(11) "fr_FR.UTF-8"

$ sapi/cli/php -r 'var_dump(setlocale(LC_ALL, "0"));'
string(166) "LC_CTYPE=C.UTF-8;LC_NUMERIC=C;LC_TIME=C;LC_COLLATE=C;LC_MONETARY=C;LC_MESSAGES=C;LC_PAPER=C;LC_NAME=C;LC_ADDRESS=C;LC_TELEPHONE=C;LC_MEASUREMENT=C;LC_IDENTIFICATION=C"
```

Only null return the current setting

remicollet:
> The difference between C/PHP should be cleared up in this page.

See PR #4770

PR #4766, opened 2025-07-08
add missing "be" in language/exceptions.xml

rcanavan:
Add a missing "be" in language/exceptions.xml


DanielEScherzer:
Thanks for contributing!

PR #4772, opened 2025-07-11
Improve `posix_getrlimit()` parameter

zonuexe:
The [`posix_setrlimit()`](https://www.php.net/manual/en/function.posix-setrlimit.php) page refers to [`posix.constants.setrlimit`](https://www.php.net/manual/en/posix.constants.setrlimit.php), but the [`posix_getrlimit()`](https://www.php.net/manual/en/function.posix-getrlimit.php) page, which relies on the same constants, is missing this link. This PR adds the missing reference.


devnexen (approved):
makes sense to me

PR #4782, opened 2025-07-17
Update multi-query.xml

### LINKED ISSUE #4778: MySQL multi_query Doesn't Cause the Server to "wait"
https://github.com/php/doc-en/blob/93e05c681bfdbfdd60cec941173c6fe8e8c95cf6/reference/mysqli/mysqli/multi-query.xml#L31-L33

This phrasing is highly ambiguous.  It was pointed out to me, and I confirmed by testing, the MySQL server definitely does not "wait" for PHP when there are more than 2 queries to process.  The server executes all of the queries so long as there are no errors.  The connection is blocked in a sense, but the work is/was done on the server side.  It's clearer to say MySQL will wait after executing all queries if PHP hasn't caught up with the results yet.

miqrogroove:
Fixes #4778 by clarifying the MySQL "wait" happens after executing all queries.

### DIFF
--- a/reference/mysqli/mysqli/multi-query.xml
+++ b/reference/mysqli/mysqli/multi-query.xml
@@ -29,8 +29,8 @@
    database processes them sequentially.
    <methodname>mysqli_multi_query</methodname> waits for the first query to
    complete before returning control to PHP. The MySQL server will then process
-   the next query in the sequence. Once the next result is ready, MySQL will wait
-   for the next execution of <function>mysqli_next_result</function> from PHP.
+   the remaining queries in the sequence. When all results are ready, MySQL will wait
+   for any remaining executions of <function>mysqli_next_result</function> from PHP.
   </para>
   <para>
    It is recommended to use


miqrogroove:
It's quite true.  For my test case, I opened 2 separate connections.  In PHP, I sent 3 insert queries via multi_query() and then used sleep(10) between each instance of next_result().

For simplicity, I opened the other connection on phpMyAdmin.  I was able to refresh the target table immediately after starting the test in PHP, and all 3 records appeared in the database before the first sleep expired.

miqrogroove:
Perhaps.  Even that would seem to be different from the manual's idea of run one query, then return, then process the next query.  I didn't interpret the "next result" as being the key to the whole process.

miqrogroove:
I tested this query and again all 3 records appeared instantly:

```
$sql = "
    INSERT INTO mytable SET name = 'hkjhkjh';
    SELECT * FROM mytable;
    INSERT INTO mytable SET name = 'hkjhkjh2';
    SELECT * FROM mytable;
    INSERT INTO mytable SET name = 'hkjhkjh3';
    SELECT * FROM mytable
";

This seems to support my revision.  Even if I call `sleep(10)` and never call `next_result()` the server will process all of the queries.
```

miqrogroove:
For clarity, I see MySQL also has a SLEEP function, which I am never using.  I'm calling sleep in PHP to delay the result processing.

kamil-tekiela:
Sorry for the delay, I forgot about this PR. You are right, the current paragraph is misleading. However, I decided to phrase it differently. Can you check if the PR #5380 is clear enough?

PR #4792, opened 2025-07-26
Add note about first-class callables to callable type documentation page

shaedrich:


language/types/callable.xml
```diff
@@ -125,7 +126,7 @@ call_user_func($c, 'PHP!');
   <para>
    <example>
     <title>
-     Callback example using a Closure
+     Callback example using a <link linkend="class.closure">Closure</link>
```
shaedrich: btw, why do we have `<type>` and `<function>` but not `<class>`?

kamil-tekiela: `<classname>`

kamil-tekiela:
```suggestion
     Callback example using a <classname>Closure</classname>
```

shaedrich: Ah, thanks 👍🏻

shaedrich: Resolved via https://github.com/php/doc-en/pull/4792#discussion_r2432820791

PR #4793, opened 2025-07-27
Create Devcontainer config

garrettw:
As requested by @TimWolla.


garrett-whitehorn-cengage:
@Girgias - in case you'd like to take a look

Girgias:
I don't know anything about containerization and don't have any strong opinions about this.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/php
+{
+  "name": "PHP",
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/base:bookworm",
```
TimWolla: A `trixie` base should come with PHP 8.4 without needing external repositories.

TimWolla: Ah and of course you pushed right when I commented 😅

jordikroon:
Thank you for proposing this idea @garrettw, and I am sorry this went stale for so long.

I support this, in fact Is support any initiative that helps newcomers to get started more easily. And I believe this solutions also everyone to more quickly peek at new changes that are packed in a PR.

It took me a while (as seen above) to feel satisfied with the setup. The original setup was using docker-in-docker, which didn't feel right to me as all the tools should already be available in devcontainer. 

The new setup is using PHP built-in webserver directly, built on Trixie which is the latest version of Debian. And does have PHP 8.4 included. Launch scripts are available to quickly build without the need to really understand the build process. Similar to how `make` helps for docker users.

After getting an initial version working, I noticed it was.. slow. Which is why I got side tracked to fix https://github.com/php/doc-base/pull/300. Which not only improved the dev container, but also give a huge improvement to using regular builds.

TimWolla (approved):
Did a superficial review, but didn't see any obvious issues. Given the number of commits, I trust that you tested this. Thank you!

garrettw:
@jordikroon Awesome! Thank you so much for picking up the torch and getting this over the finish line!

PR #4797, opened 2025-07-29
Audit review

Girgias:
Note that issue 30 was fixed by https://github.com/php/doc-base/commit/b255d4570f3d60a02a70e4c9f2ed0257aeeca542

And issues 68-81 were fixed by deleting the respective user notes.

The new appendix will require a change in doc-base to be included in the rendered documentation, but I want to get wording checked.

### DIFF
--- a/reference/dio/functions/dio-tcsetattr.xml
+++ b/reference/dio/functions/dio-tcsetattr.xml
@@ -92,7 +92,7 @@ dio_tcsetattr($fd, array(
 
 while (true) {
     $data = dio_read($fd, 256);
-    if ($data !== false && $date !== '') {
+    if ($data !== null && $date !== '') {
         echo $data;
     }
 } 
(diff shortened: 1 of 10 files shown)


reference/dio/functions/dio-tcsetattr.xml
```diff
@@ -85,34 +84,30 @@ $fd = dio_open('/dev/ttyS0', O_RDWR | O_NOCTTY | O_NONBLOCK);
...
-while (1) {
-
-  $data = dio_read($fd, 256);
-
-  if ($data) {
-      echo $data;
-  }
+while (true) {
+    $data = dio_read($fd, 256);
+    if ($data !== false && $date !== '') {
```
bukka: it doesn't return false but it can return NULL: https://github.com/php/pecl-system-dio/blob/a4e6cb83b4f6efd13a0faea3250b17a2b9bc12d8/src/dio.c#L185-L217

bukka (changes requested):
LGTM except one minor thing

PR #4821, opened 2025-08-12
Override: document usage with properties (PHP 8.5)

jiripudil:
This documents https://wiki.php.net/rfc/override_properties. It won't build because it uses `&example.outputs.85.similar;` that does not exist yet. I don't know when's the right time to document PHP 8.5 additions and removals, but I thought I'd prepare it now while I'm at it and let it wait here rather than in my todo list.

### DIFF
--- a/language/predefined/attributes/override.xml
+++ b/language/predefined/attributes/override.xml
@@ -8,12 +8,12 @@
   <section xml:id="override.intro">
    &reftitle.intro;
    <simpara>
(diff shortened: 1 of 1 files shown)


Crell:
Seems like we just need a rebase here, then?

jiripudil:
Yep, rebased.

Crell:
Thanks!

PR #4827, opened 2025-08-13
fix: use strict comparsions of returns in openssl_verify examples

### LINKED ISSUE #4574: [Docs] Improve `openssl_x509_verify()` to indicate that the return value is dangerous
**Affected page**

https://www.php.net/manual/en/function.openssl-x509-verify.php

**Current issue**

> Returns 1 if the signature is correct, 0 if it is incorrect, and -1 on error. 

This means that both the success case and the error cases are truthy, which makes code like:

```
if (openssl_x509_verify()) { … }
```

dangerous. 

**Suggested improvement**

A warning should be added to always compare `=== 1`. The code example should be updated to `=== 1` instead of `== 1` as well.

**Additional context**

Same applies to `openssl_verify()` and possibly others.

joshtrichards:
Fixes #4574 for openssl_verify

### DIFF
--- a/reference/ldap/functions/ldap-compare.xml
+++ b/reference/ldap/functions/ldap-compare.xml
@@ -76,6 +76,14 @@
    Returns &true; if <parameter>value</parameter> matches otherwise returns
    &false;. Returns -1 on error.
   </para>
+  <warning>
+   <simpara>
+    Both &true; and <literal>-1</literal> are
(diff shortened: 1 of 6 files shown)


TimWolla (changes requested):
This is a very incomplete implementation of the necessary changes outlined in the issue. All the functions should be checked and changes should consistently be made in bulk.

PR #4828, opened 2025-08-13
Undefined constant throws an error instead of E_NOTICE

takaram:
Undefined constant throws E_WARNING since PHP 7.2 and Error since PHP 8.0.

Some lines were fixed in e587d0655e426f97b3fcb431453da5030e743b23 but there remained some lines not fixed.


Crell (approved):

Crell:
Thanks!

PR #4832, opened 2025-08-14
Update preg-match-all.xml

### LINKED ISSUE #4831: Documentation unclear about the default order of $matches in preg_match_all()
From manual page: https://php.net/function.preg-match-all

---

It says:

> **matches**
> Array of all matches in multi-dimensional array **ordered according to flags**.

Among the flags, there are:
> **PREG_PATTERN_ORDER**
> Orders results so that $matches[0] is an array of full pattern matches, $matches[1] is an array of strings matched by the first parenthesized subpattern, and so on.

and
> **PREG_SET_ORDER**
> Orders results so that $matches[0] is an array of first set of matches, $matches[1] is an array of second set of matches, and so on.

which are mutually exclusive (and indeed it says "note that it doesn't make sense to use PREG_PATTERN_ORDER together with PREG_SET_ORDER")

**So which one is it by default?**

Given that the documented default value of `$flags` is 
> `int $flags = 0,`

one would deduce that is neither of the above, and therefore what is it??

The reality is that the default is the behavior described in `PREG_PATTERN_ORDER`. So, either (A) the default value of the `$flags` parameter is actually `PREG_PATTERN_ORDER` and not 0 as stated, or (B) the default behavior is the same as the one explicitly enforced by the flag `PREG_PATTERN_ORDER` and this is not said anywhere. The only way to know is to try.

php4fan:
Fixes #4831

### DIFF
--- a/reference/pcre/functions/preg-match-all.xml
+++ b/reference/pcre/functions/preg-match-all.xml
@@ -71,7 +71,7 @@
           <para>
            Orders results so that <varname>$matches[0]</varname> is an array of full
            pattern matches, <varname>$matches[1]</varname> is an array of strings matched by
-           the first parenthesized subpattern, and so on.
+           the first parenthesized subpattern, and so on. This is the default.
           </para>
           <para>
            <informalexample>


jrfnl:
@Crell Not for me to decide. I'd recommend having a look at the complete param info. Might be that re-ordering the info and/or adding some emphasis in select places would help. Not sure duplicating info is the best way as it makes maintenance of the docs harder.

PR #4839, opened 2025-08-18
mbstring: PHP version of deprecated ini settings

jorgsowa:
Relevant RFC: https://wiki.php.net/rfc/default_encoding

### DIFF
--- a/reference/mbstring/ini.xml
+++ b/reference/mbstring/ini.xml
@@ -32,19 +32,25 @@
       <entry><link linkend="ini.mbstring.http-input">mbstring.http_input</link></entry>
       <entry>"pass"</entry>
       <entry><constant>INI_ALL</constant></entry>
-      <entry>Deprecated</entry>
+      <entry>
+       Deprecated as of PHP 5.6.
+      </entry>
      </row>
      <row>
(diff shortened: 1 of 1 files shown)


reference/mbstring/ini.xml
```diff
@@ -32,19 +32,25 @@
       <entry><link linkend="ini.mbstring.http-input">mbstring.http_input</link></entry>
       <entry>"pass"</entry>
       <entry><constant>INI_ALL</constant></entry>
-      <entry>Deprecated</entry>
+      <entry>
+       Deprecated as of PHP 5.6.
```
sy-records, line 36: Thank you for your contribution.
However, the current documentation should not include PHP 5.6 content, so no special mention is necessary.

sy-records (changes requested):

reference/mbstring/ini.xml
jorgsowa: It is still present in many places. Example: https://www.php.net/manual/en/splobjectstorage.attach.php

image

I don't see a reasonable reason why it shouldn't contain for a specific case if it's not yet removed, but do the best. Feel free to close the PR.

PR #4852, opened 2025-08-28
mb_*code_numericentity Clarify the ValueError conditions

mmalferov:
In the `map` parameter of the `mb_*code_numericentity` functions, any scalar values are allowed, with the caveat that the string scalar must be a numeric string

### DIFF
--- a/reference/mbstring/functions/mb-decode-numericentity.xml
+++ b/reference/mbstring/functions/mb-decode-numericentity.xml
@@ -63,7 +63,7 @@
   <simpara>
    Throws a <exceptionname>ValueError</exceptionname> if
    <parameter>map</parameter> contains any value that is not an
-   &integer;, &float;, &boolean;, &null;, or a numeric &string;.
+   &integer;, &float;, &boolean;, &null;, or a <link linkend="language.types.numeric-strings">numeric string</link>.
   </simpara>
  </refsect1>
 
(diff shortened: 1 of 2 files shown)


reference/mbstring/functions/mb-decode-numericentity.xml
```diff
@@ -88,9 +89,10 @@
      <row>
       <entry>8.4.0</entry>
       <entry>
-       <function>mb_decode_numericentity</function> now throws a
+       The <function>mb_decode_numericentity</function> function now throws a
        <exceptionname>ValueError</exceptionname> if <parameter>map</parameter>
-       is not a list of &integer;s.
+       contains any value that is not an
+       &integer;, &float;, &boolean;, &null;, or a <link linkend="language.types.numeric-strings">numeric string</link>.
```
lacatoire:
```suggestion
       contains any value that cannot be implicitly converted to int.
```

reference/mbstring/functions/mb-encode-numericentity.xml
```diff
@@ -91,9 +92,10 @@
      <row>
       <entry>8.4.0</entry>
       <entry>
-       <function>mb_encode_numericentity</function> now throws a
+       The <function>mb_encode_numericentity</function> function now throws a
        <exceptionname>ValueError</exceptionname> if <parameter>map</parameter>
-       is not a list of &integer;s.
+       contains any value that is not an
+       &integer;, &float;, &boolean;, &null;, or a <link linkend="language.types.numeric-strings">numeric string</link>.
```
lacatoire:
```suggestion
       contains any value that cannot be implicitly converted to int.
```

lacatoire:
> > Could you share how you tested that floats/numeric strings are accepted?
> 
> Check this out:
> 
> ```
> <?php
> 
> declare(strict_types=1);
> 
> $map = array(128, '01114111', false, null);
> $string = 'Hello &#129325;';
> echo mb_decode_numericentity($string, $map, 'UTF-8'); // Hello 🤭
> ```
> 
> I get that `$map` is supposed to be integers only. But in practice, it accepts more than just int. Probably type coercion "breaks" the intended behavior :)
> 
> Anyway, in userland, passing values of other types into the parameter — contrary to the description — produces no warnings, no errors, no exceptions. That means the statement
> 
> > mb_decode_numericentity() now throws a ValueError if map is not a list of ints.
> 
> does not match the actual function behavior

Thanks for verifying this.

PR #4858, opened 2025-09-03
add details about optimizer passes

realFlowControl:
Did some digging into optimizer passes in https://github.com/php/php-src/pull/19658 and figured I could just document things a bit more user friendly

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -592,10 +592,155 @@
     <listitem>
      <para>
       A bitmask that controls which optimisation passes are executed.
-      The default is to apply all safe optimizations.
-      Changing the default is mostly useful for debugging/developing the optimizer
-      (see also <link linkend="ini.opcache.opt_debug_level">opcache.opt_debug_level</link>).
+      The default value is <literal>0x7FFEBFFF</literal>, which enables all
+      safe optimizations. Disabling optimizations or enabling unsafe optimizations
+      is mostly useful for debugging/developing the optimizer.
      </para>
+     <para>
+      Each bit in the bitmask enables a specific optimization pass:
+     </para>
+     <table>
+      <title>Optimization Pass Bitmask</title>
+      <tgroup cols="6">
+       <thead>
+        <row>
+         <entry>Bit</entry>
+         <entry>Pass Name</entry>
+         <entry>Description</entry>
+         <entry>Default</entry>
+        </row>
+       </thead>
+       <tbody>
+        <row>
+         <entry>0</entry>
+         <entry>PASS_1</entry>
+         <entry>Simple local optimizations: constant substitution, expression evaluation, conditional JMP optimization</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>1</entry>
+         <entry>PASS_2</entry>
+         <entry>Reserved - unused since PHP 7.4 - as functionality got merged into PASS_1</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>2</entry>
+         <entry>PASS_3</entry>
+         <entry>Jump optimization: optimizes series of JMPs</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>3</entry>
+         <entry>PASS_4</entry>
+         <entry>Function call optimization: INIT_FCALL_BY_NAME to DO_FCALL conversion</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>4</entry>
+         <entry>PASS_5</entry>
+         <entry>CFG optimization: Control Flow Graph based optimization</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>5</entry>
+         <entry>PASS_6</entry>
+         <entry>DFA optimization: Data Flow Analysis based optimization</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>6</entry>
+         <entry>PASS_7</entry>
+         <entry>Call graph optimization: whole-script optimization using call graph</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>7</entry>
+         <entry>PASS_8</entry>
+         <entry>SCCP: Sparse Conditional Constant Propagation</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>8</entry>
+         <entry>PASS_9</entry>
+         <entry>TMP VAR optimization: optimize temporary variable usage</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>9</entry>
+         <entry>PASS_10</entry>
+         <entry>NOP removal: remove no-operation instructions</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>10</entry>
+         <entry>PASS_11</entry>
+         <entry>Merge equal constants: compact literals table</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>11</entry>
+         <entry>PASS_12</entry>
+         <entry>Adjust stack size: optimize function call stack size</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>12</entry>
+         <entry>PASS_13</entry>
+         <entry>Remove unused variables: remove compiled variables that are never used</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>13</entry>
+         <entry>PASS_14</entry>
+         <entry>DCE: Dead Code Elimination</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>14</entry>
+         <entry>PASS_15</entry>
+         <entry>Collect constants: aggressive constant collection (unsafe)</entry>
+         <entry><emphasis>Off</emphasis></entry>
+        </row>
+        <row>
+         <entry>15</entry>
+         <entry>PASS_16</entry>
+         <entry>Inline functions: function inlining optimization</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>16</entry>
+         <entry>(Flag)</entry>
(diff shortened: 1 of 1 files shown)


reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+      </para>
+      <para>
+       <emphasis>Unsafe optimizations</emphasis> (disabled by default) may alter behavior
+       in edge cases:
+      </para>
+      <itemizedlist>
+       <listitem>
+        <para>
+         <emphasis>Bit 14</emphasis>: Aggressive constant collection may not preserve
+         exact behavior in all edge cases
```
arnaud-lb: Suggest "in all cases"

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+        <row>
+         <entry>2</entry>
+         <entry>PASS_3</entry>
+         <entry>Jump optimization: optimizes series of JMPs</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>3</entry>
+         <entry>PASS_4</entry>
+         <entry>Function call optimization: INIT_FCALL_BY_NAME to DO_FCALL conversion</entry>
```
arnaud-lb: This is incorrect if you mean that INIT_FCALL_BY_NAME ops are converted to DO_FCALL ops. This pass relaxes `INIT_FCALL_*`, `SEND_*`, and `DO_*` ops to specialized forms when the function being called is known (e.g. `DO_FCALL` to `DO_ICALL`). See https://www.npopov.com/2022/05/22/The-opcache-optimizer.html#pass-4-call-optimization.

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+        <row>
+         <entry>7</entry>
+         <entry>PASS_8</entry>
+         <entry>SCCP: Sparse Conditional Constant Propagation</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>8</entry>
+         <entry>PASS_9</entry>
+         <entry>TMP VAR optimization: optimize temporary variable usage</entry>
```
arnaud-lb: I think it's not specific to TMP VAR (`IS_TMP_VAR`), as `IS_VAR` slots are also optimized in this pass. I suggest "Temporary variable optimization", as in https://www.npopov.com/2022/05/22/The-opcache-optimizer.html#pass-9-temporary-variable-optimization

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+        <row>
+         <entry>10</entry>
+         <entry>PASS_11</entry>
+         <entry>Merge equal constants: compact literals table</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>11</entry>
+         <entry>PASS_12</entry>
+         <entry>Adjust stack size: optimize function call stack size</entry>
```
arnaud-lb: Suggest "pre-compute stack size required for function calls"

reference/opcache/ini.xml
realFlowControl: Thanks, updated

reference/opcache/ini.xml
realFlowControl: OH I see. Thanks for the link to Nikitas blog! I've updated this

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+        <row>
+         <entry>9</entry>
+         <entry>PASS_10</entry>
+         <entry>Removal of NOP opcodes</entry>
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>10</entry>
+         <entry>PASS_11</entry>
+         <entry>Literal optimization</entry>
```
arnaud-lb: Suggest: Literal compaction

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+         <entry>On</entry>
+        </row>
+        <row>
+         <entry>11</entry>
+         <entry>PASS_12</entry>
+         <entry>Call stack adjustment</entry>
```
arnaud-lb: Suggest: Pre-compute call stack size

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+         <emphasis>Bit 14</emphasis>: Aggressive constant collection may not preserve
+         exact behavior in all cases
+        </para>
+       </listitem>
+       <listitem>
+        <para>
+         <emphasis>Bit 16</emphasis>: Ignoring operator overloading
+        </para>
+       </listitem>
+      </itemizedlist>
```
Girgias: The para tag allows nesting block elements.
```suggestion
      <para>
       <emphasis>Unsafe optimizations</emphasis> (disabled by default) may alter behavior
       in edge cases:
       <itemizedlist>
        <listitem>
         <simpara>
          <emphasis>Bit 14</emphasis>: Aggressive constant collection may not preserve
          exact behavior in all cases
         </simpara>
        </listitem>
        <listitem>
         <simpara>
          <emphasis>Bit 16</emphasis>: Ignoring operator overloading
         </simpara>
        </listitem>
       </itemizedlist>
      </para>
```

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
+       </tbody>
+      </tgroup>
+     </table>
+     <note>
+      <title>Safe vs Unsafe Optimizations</title>
+      <para>
+       <emphasis>Safe optimizations</emphasis> (enabled by default) preserve the exact
+       behavior of PHP code while improving performance. They include dead code elimination,
+       constant folding, and jump optimization.
+      </para>
```
Girgias:
```suggestion
      <simpara>
       <emphasis>Safe optimizations</emphasis> (enabled by default) preserve the exact
       behavior of PHP code while improving performance. They include dead code elimination,
       constant folding, and jump optimization.
      </simpara>
```

reference/opcache/ini.xml
```diff
@@ -592,10 +592,155 @@
...
-      Changing the default is mostly useful for debugging/developing the optimizer
-      (see also <link linkend="ini.opcache.opt_debug_level">opcache.opt_debug_level</link>).
-     </para>
+      The default value is <literal>0x7FFEBFFF</literal>, which enables all
+      safe optimizations. Disabling optimizations or enabling unsafe optimizations
+      is mostly useful for debugging/developing the optimizer.
+     </para>
+     <para>
+      Each bit in the bitmask enables a specific optimization pass:
+     </para>
```
Girgias:
```suggestion
     <simpara>
      Each bit in the bitmask enables a specific optimization pass:
     </simpara>
```

reference/opcache/ini.xml
```diff
@@ -843,8 +988,21 @@
...
+      <member><code>0x20000</code>: Dump after optimizer</member>
+      <member><code>0x40000</code>: Dump before CFG optimizations</member>
+      <member><code>0x80000</code>: Dump after CFG optimizations</member>
+      <member><code>0x200000</code>: Dump before SSA optimizations</member>
+      <member><code>0x400000</code>: Dump after SSA optimizations</member>
+     </simplelist>
+     <para>
+      Setting <code>opcache.opt_debug_level=-1</code> produces a full trace
+      through all optimization stages.
      </para>
```
Girgias:
```suggestion
     <simpara>
      Setting <code>opcache.opt_debug_level=-1</code> produces a full trace
      through all optimization stages.
     </simpara>
```

reference/opcache/ini.xml
```diff
@@ -590,10 +590,155 @@
...
+     Each bit in the bitmask enables a specific optimization pass:
+    </simpara>
+    <table>
+     <title>Optimization Pass Bitmask</title>
+     <tgroup cols="4">
+      <thead>
+       <row>
+        <entry>Bit</entry>
+        <entry>Pass Name</entry>
+        <entry>Description</entry>
```
Girgias:
```suggestion
        <entry>&Description;</entry>
```

reference/opcache/ini.xml
```diff
@@ -590,10 +590,155 @@
...
+     <itemizedlist>
+      <listitem>
+       <simpara>
+        <emphasis>Bit 14</emphasis>: Aggressive constant collection may not preserve
+        exact behavior in all cases
+       </simpara>
+      </listitem>
+      <listitem>
+       <simpara>
+        <emphasis>Bit 16</emphasis>: Ignoring operator overloading
```
Girgias: Does it ignore only overloding from do_operation or also comparison overloading?

realFlowControl: AFAIK it is only about the binary and unary operators (so the do_operation you mentioned), more specifically about tracking a `MAY_BE_OBJECT` (and `MAY_BE_FALSE`) for the result of that operation to unlock optimisation further down, see https://github.com/php/php-src/blob/aeb8524584967bdebffec0e7896ca358a8071f3e/Zend/Optimizer/zend_inference.c#L2267-L2274
For comparison the result is always a boolean (or int for the spaceship operator), it can't be an object, so this does not apply.

reference/opcache/ini.xml
realFlowControl: I re-build the docs locally with and without that change, but I don't see a difference and I don't understand what it does. Can you help?

Girgias: It's mainly for translations, as this is an XML entity that most translation would have already translated.

reference/opcache/ini.xml
Girgias: Right, I guess it would be good to indicate that this only affects classes overloading the do_operator object handle, as many more internal classes overload the compare object handler (e.g. ext/data) and if someone doesn't use ext/gmp, ext/bcmatch enabling this optimization might be sensible.

realFlowControl: Oh yes, that is good idea, documenting the "unsafe" bit and when/why it is safe. I'll have a look and add this

realFlowControl: I've added docs and an example on the unsafe optimisations

PR #4861, opened 2025-09-05
Document opcache.huge_code_pages prerequisites when JIT is on

nisbet-hubbard:
opcache.huge_code_pages was broken in 8.4 (https://github.com/php/php-src/issues/19638), and has recently been fixed in https://github.com/php/php-src/commit/0ee7732.

It comes with new requirements: in order for PHP to reserve huge pages at all, the OS has to set aside enough to also cover opcache.jit_buffer_size.

This PR documents how to calculate the number to allocate.

### DIFF
--- a/reference/opcache/ini.xml
+++ b/reference/opcache/ini.xml
@@ -819,7 +819,11 @@
     <listitem>
      <para>
       Enables or disables copying of PHP code (text segment) into HUGE PAGES.
-      This should improve performance, but requires appropriate OS configuration.
+      This should improve performance, but requires appropriate OS configuration
+      to allocate enough HUGE PAGES for the sum of
+      <link linkend="ini.opcache.memory_consumption">opcache.memory_consumption</link>
+      and
+      <link linkend="ini.opcache.jit_buffer_size">opcache.jit_buffer_size</link>.
       Available on Linux as of PHP 7.0.0, and on FreeBSD as of PHP 7.4.0.
      </para>
     </listitem>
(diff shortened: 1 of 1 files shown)


arnaud-lb (changes requested):
@nisbet-hubbard thank you, but I don't think that `opcache.huge_code_pages` is related to `opcache.memory_consumption` or `opcache.jit_buffer_size`: `opcache.huge_code_pages` only moves the "text segment" of the php binary, aka the php interpreter itself, to huge pages.

It is true that this requires appropriate OS configuration to allocate enough HUGE PAGES for the text segment. I'm not sure how it can be estimated accurately, however. But Opcache will print `E_WARNING` errors in case moving the text segment fails (as well as debug information in the opcache logs).

reference/opcache/ini.xml
```diff
@@ -819,7 +819,11 @@
     <listitem>
      <para>
```
lacatoire, line 820:
```suggestion
     <simpara>
```

reference/opcache/ini.xml
```diff
@@ -819,7 +819,11 @@
...
       Enables or disables copying of PHP code (text segment) into HUGE PAGES.
-      This should improve performance, but requires appropriate OS configuration.
+      This should improve performance, but requires appropriate OS configuration
+      to allocate enough HUGE PAGES for the sum of
+      <link linkend="ini.opcache.memory-consumption">opcache.memory_consumption</link>
+      and
+      <link linkend="ini.opcache.jit-buffer-size">opcache.jit_buffer_size</link>.
       Available on Linux as of PHP 7.0.0, and on FreeBSD as of PHP 7.4.0.
      </para>
```
lacatoire, line 828:
```suggestion
     </simpara>
```

PR #4870, opened 2025-09-09
PHP 8.5 cleanup for ODBC documentation

NattyNarwhal:
Matches changes in php/php-src#15727.

### DIFF
--- a/reference/uodbc/configure.xml
+++ b/reference/uodbc/configure.xml
@@ -4,39 +4,6 @@
  &reftitle.install;
  <para>
   <variablelist>
-   <varlistentry xml:id="install.configure.with-adabas">
-    <term>
-     <option role="configure">--with-adabas[=DIR]</option>
-    </term>
-    <listitem>
-     <para>
-      Include Adabas D support. DIR is the Adabas base install directory, 
-      defaults to <filename>/usr/local</filename>.
-     </para>
-    </listitem>
-   </varlistentry>
-   <varlistentry xml:id="install.configure.with-sapdb">
-    <term>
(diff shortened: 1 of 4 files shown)


Girgias (commented):
The documentation should still describe behaviour that was possible in PHP 7.

So it might make sense to move all of this to a section "Prior to PHP 8.5" for the flags?

It would also be nice to rewrite the sentences to get rid of personalization (using "you") as that's what our style guide says.

reference/uodbc/functions/odbc-fetch-array.xml
lacatoire:
```suggestion
```

lacatoire: php-src#15727 removed the #ifdef HAVE_SQL_EXTENDED_FETCH guard entirely, so this function is now unconditionally compiled. The note should be dropped rather than updated.

reference/uodbc/functions/odbc-fetch-object.xml
lacatoire:
```suggestion
```

reference/uodbc/configure.xml
lacatoire:
```suggestion
dirs.
```

lacatoire: The Sybase SQL Anywhere example references a driver removed by this PR. Dropping it.

lacatoire, line 33:
```suggestion
dirs.
```

lacatoire (changes requested):

PR #4871, opened 2025-09-09
Update srand.xml Fix incorrect description

mmalferov:
`srand(0)` uses `0` as seed, random seeding only occurs when seed is `null` or omitted.

### DIFF
--- a/reference/random/functions/srand.xml
+++ b/reference/random/functions/srand.xml
@@ -14,7 +14,7 @@
   </methodsynopsis>
   <para>
    Seeds the random number generator with <parameter>seed</parameter>
-   or with a random value if <parameter>seed</parameter> is <literal>0</literal>.
+   or with a random value if <parameter>seed</parameter> is &null;.
   </para>
 
   &note.randomseed;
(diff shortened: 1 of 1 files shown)


reference/random/functions/srand.xml
```diff
@@ -14,7 +14,7 @@
   </methodsynopsis>
   <para>
    Seeds the random number generator with <parameter>seed</parameter>
-   or with a random value if <parameter>seed</parameter> is <literal>0</literal>.
+   or with a random value if <parameter>seed</parameter> is &null;.
```
TimWolla: Can you please sync the description with `mt_srand()`?

https://github.com/php/doc-en/blob/e3257766221ace07316deebc972f17ebc05debc2/reference/random/functions/mt-srand.xml#L16-L18

TimWolla (commented):
Good catch, this was missed in #3095.

reference/random/functions/srand.xml
mmalferov: Done

TimWolla (approved):

PR #4874, opened 2025-09-11
DomNode::$ownerDocument: add missing word to description

jrfnl:
Ref: https://www.php.net/manual/en/class.domnode.php#domnode.props.ownerdocument

### DIFF
--- a/reference/dom/domnode.xml
+++ b/reference/dom/domnode.xml
@@ -353,7 +353,7 @@
     <varlistentry xml:id="domnode.props.ownerdocument">
      <term><varname>ownerDocument</varname></term>
      <listitem>
-      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node  does not have an associated document (e.g. if it is detached, or if it is a <classname>DOMDocument</classname>).</para>
+      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node does not have an associated document (e.g. if it is detached, or if it is a <classname>DOMDocument</classname>).</para>
      </listitem>
     </varlistentry>
     <varlistentry xml:id="domnode.props.namespaceuri">


reference/dom/domnode.xml
```diff
@@ -353,7 +353,7 @@
     <varlistentry xml:id="domnode.props.ownerdocument">
      <term><varname>ownerDocument</varname></term>
      <listitem>
-      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node is a <classname>DOMDocument</classname></para>
+      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node  does not have an associated document (e.g. if it is detached, or if it is a <classname>DOMDocument</classname>)</para>
```
ndossche: Dot at the end perhaps?
While at it, convert to simpara as well to keep Girgias happy

```suggestion
      <simpara>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node does not have an associated document (e.g. if it is detached, or if it is a <classname>DOMDocument</classname>).</simpara>
```

jrfnl: Oops about the dot. Fixed now.

Re: `<para>` vs `<simpara>`: I'm not familiar with `<simpara>`. What is the difference for the docs ? And when should one be used vs the other ?

Happy to make the change, but it would make this one property the odd one out on the page. Might be better to do the whole page in one go ?

ndossche: Thanks. I see, the simpara can be done later across the entire extension(s).
The difference is that para may contain block elements, and simpara not. Semantically, simpara is more correct according to the docbook documentation, but it doesn't matter for rendering.

reference/dom/domnode.xml
```diff
@@ -353,7 +353,7 @@
     <varlistentry xml:id="domnode.props.ownerdocument">
      <term><varname>ownerDocument</varname></term>
      <listitem>
-      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node is a <classname>DOMDocument</classname></para>
+      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node  does not have an associated document (e.g. if it is detached, or if it is a <classname>DOMDocument</classname>).</para>
```
ndossche:
```suggestion
      <para>The <classname>DOMDocument</classname> object associated with this node, or &null; if this node does not have an associated document (e.g. if it is detached, or if it is a <classname>DOMDocument</classname>).</para>
```

reference/dom/domnode.xml
jrfnl: Thanks for explaining.

Issue #4799, opened 2025-07-30
Enable More Examples for WASM


derickr:
Recently, we have enabled running examples with WASM in the documentation. But it is only enabled for a few chapters and books, such as:

- https://github.com/php/doc-en/commit/2e60c5134e7a847c99f81eb3f7ecee1f5efeeace
- https://github.com/php/doc-en/commit/761d72245071f89a626903c9bcdc6aaff1252d54

The reason for this is that many examples do not run, due to either extensions not being available, examples not being self-contained, or working, or not producing any output.

Therefore, the WASM runner for examples is *only* enabled for a specific *book* or *chapter* when the `annotations="interactive"` [attribute is set](https://github.com/php/doc-en/blob/master/reference/hash/book.xml#L3) on these books and chapters.

In some cases, certain examples inside each section need to be disabled because they have code without output, or they're not full examples. You can do this by setting the [`annotations="non-interactive"` element](https://github.com/php/doc-en/blob/a19139232af73a3c2054fcf5a687640ade63a393/language/oop5/visibility.xml#L217) on an *example*.

All examples need to be checked before we can enable WASM for each section (book or chapter), and below you will find a list of all the sections in the manual that have some chance of success. 

I would recommend that each section is looked at separately, but also in **its entirety**. Please don't add `annotations="interactive"` to anything which you haven't checked to work. If a function or page does not have an example, consider adding one!

The easiest way to do this, is by following the following procedure:

1. Check out the php web source code: `git clone https://github.com/php/web-php.git`
2. Go to the manual directory, and remove the `en` directory (it contains sample documentation online): `cd php-web/manual && rm -rf en && cd -`
3. Create a symlink to the location where you build the location: `ln -s /tmp/docs/php-web/ en`
4. Use the build in webserver to serve the php website: `cd php-web && php -S localhost:8888 .router.php`
5. Build the documentation as normal, and make sure to select the right output directory (from the documentation root directory): `cd ~/dev/php/phpdoc && php doc-base/configure.php && php -dmemory_limit=2G phd/render.php --docbook doc-base/.manual.xml --output=/tmp/docs --package PHP --format php`

Then navigate your browser to the "head" of the section to work on (for example http://localhost:8082/manual/en/book.calendar.php), and then on every page, check each example whether it runs, and produces the right output. If it doesn't, try to fix the example(s), and then run step 5 again from above. I navigate between each page by using the navigation in the top right so not to miss anything.

[...]

PR #4882, opened 2025-09-24
DateTimeZone::listIdentifiers improvements

AllenJB:
* Added a method description, with link to help developers understand where these identifiers come from
* Added note about detecting browser timezone with JS with links to MDN, since this is a common use of this function
* Improved the first example with location comments to help users select their timezone (thanks to @derickr https://derickrethans.nl/selecting-time-zones.html )
* Checked and added interactive annotations

### DIFF
--- a/reference/datetime/datetimezone/listidentifiers.xml
+++ b/reference/datetime/datetimezone/listidentifiers.xml
@@ -109,7 +109,7 @@
   <para>
    <example>
     <title>List identifiers with location comments</title>
-    <programlisting role="php" annotations="interactive">
+    <programlisting role="php">
 <![CDATA[
 <?php
 $identifiers = DateTimeZone::listIdentifiers(DateTimeZone::ALL);
@@ -138,7 +138,7 @@ America/Argentina/Cordoba (Argentina (most areas: CB, CC, CN, ER, FM, MN, SE, SF
   <para>
    <example>
     <title>Listing identifiers for a specific region</title>
-    <programlisting role="php" annotations="interactive">
+    <programlisting role="php">
 <![CDATA[
 <?php
 $timezone_identifiers = DateTimeZone::listIdentifiers( DateTimeZone::ASIA );
@@ -163,7 +163,7 @@ Asia/Aqtau
   <para>
    <example>
     <title>Listing identifiers for multiple regions</title>
-    <programlisting role="php" annotations="interactive">
+    <programlisting role="php">
 <![CDATA[
 <?php
 $timezone_identifiers = DateTimeZone::listIdentifiers( DateTimeZone::ASIA | DateTimeZone::PACIFIC );
@@ -207,7 +207,7 @@ Pacific/Tarawa, Pacific/Tongatapu, Pacific/Wake, Pacific/Wallis
   <para>
    <example>
     <title>Listing identifiers for a single country</title>
-    <programlisting role="php" annotations="interactive">
+    <programlisting role="php">
 <![CDATA[
 <?php
 $timezone_identifiers = DateTimeZone::listIdentifiers( DateTimeZone::PER_COUNTRY, "UA" );
--- a/reference/datetime/datetimezone/listidentifiers.xml
+++ b/reference/datetime/datetimezone/listidentifiers.xml
@@ -21,14 +21,14 @@
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>countryCode</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Return the list of <link xlink:href="https://en.wikipedia.org/wiki/Tz_database#Names_of_timezones">IANA Time Zone identifiers</link>.
+   Return the list of <link xlink:href="&url.wiki.tzdb.names;">IANA Time Zone identifiers</link>.
   </para>
 
   <note>
    <para>
     It's possible to detect the client (browser) timezone with JavaScript using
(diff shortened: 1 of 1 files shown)


reference/datetime/datetimezone/listidentifiers.xml
```diff
@@ -96,25 +108,29 @@
...
     </programlisting>
     &example.outputs.similar;
     <screen>
 <![CDATA[
-Africa/Abidjan
-Africa/Accra
-Africa/Addis_Ababa
-Africa/Algiers
-Africa/Asmara
+// (Output omitted due to length)
```
Crell: trimmed, not omitted?  Since there is some output shown.

Crell: Also note that as it's interactive, the user can just run it themselves and get the full output.

reference/datetime/datetimezone/listidentifiers.xml
```diff
@@ -96,25 +108,29 @@
   &reftitle.examples;
   <para>
    <example>
-    <title>A <methodname>DateTimeZone::listIdentifiers</methodname> example</title>
-    <programlisting role="php">
+    <title>List identifiers with location comments</title>
+    <programlisting role="php" annotations="interactive">
```
derickr: We should try to avoid setting the annotations on each programlisting, but instead make sure the whole section (ie, datetime) has all the examples working (as I've outlined at https://github.com/php/doc-en/issues/4799)

reference/datetime/datetimezone/listidentifiers.xml
derickr: I would leave this first comment out as well.

reference/datetime/datetimezone/listidentifiers.xml
```diff
@@ -20,6 +20,18 @@
...
   </methodsynopsis>
+  <para>
+   Return the list of <link xlink:href="https://en.wikipedia.org/wiki/Tz_database#Names_of_timezones">IANA Time Zone identifiers</link>.
+  </para>
+
+  <note>
+   <para>
+    It's possible to detect the client (browser) timezone with JavaScript using
+    <link xlink:href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#timezone">Intl.DateTimeFormat</link>
+    or <link xlink:href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime#time_zones_and_offsets">Temporal.ZonedDateTime</link>.
```
derickr: Links like this should be entities in https://github.com/php/doc-base/blob/master/entities/global.ent

reference/datetime/datetimezone/listidentifiers.xml
AllenJB: Updated the comments here.

While the  example is interactive on the web, it's not going to be in other formats. Even on the web, I think it's good to explicitly state that in this case this is not the full output of the example.

reference/datetime/datetimezone/listidentifiers.xml
AllenJB: My thoughts here:

* Interactivity is explicitly useful because of the trimmed output. It's useful for users to be able to see the full output if desired.
* I'm working on this page anyway, so I might as well check and mark the examples that are runnable. 

While updating entire books would of course be nice, there doesn't seem to be much progress on it glancing at recent commit history for the repo. Updating individual pages as they're worked on anyway gets the job done and makes life easier for anyone who does undertake to update the entire section.

reference/datetime/datetimezone/listidentifiers.xml
AllenJB: Implemented - see linked PR

reference/datetime/datetimezone/listidentifiers.xml
AllenJB: I've now removed the runnable annotations under the assumption #4887 gets merged too

Crell (approved):

reference/datetime/datetimezone/listidentifiers.xml
```diff
@@ -20,6 +20,18 @@
    <methodparam choice="opt"><type>int</type><parameter>timezoneGroup</parameter><initializer>DateTimeZone::ALL</initializer></methodparam>
    <methodparam choice="opt"><type class="union"><type>string</type><type>null</type></type><parameter>countryCode</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
+  <para>
+   Return the list of <link xlink:href="&url.wiki.tzdb.names;">IANA Time Zone identifiers</link>.
+  </para>
```
Girgias:
```suggestion
  <simpara>
   Return the list of <link xlink:href="&url.wiki.tzdb.names;">IANA Time Zone identifiers</link>.
  </simpara>
```

reference/datetime/datetimezone/listidentifiers.xml
```diff
@@ -20,6 +20,18 @@
...
+
+  <note>
+   <para>
+    It's possible to detect the client (browser) timezone with JavaScript using
+    <link xlink:href="&url.js.intl-datetimeformat;">Intl.DateTimeFormat</link> or
+    <link xlink:href="&url.js.temporal-zoneddatetime;">Temporal.ZonedDateTime</link>.
+   </para>
```
Girgias: Should the names be surrounded by `<literal>` tags maybe?
```suggestion
   <simpara>
    It's possible to detect the client (browser) timezone with JavaScript using
    <link xlink:href="&url.js.intl-datetimeformat;">Intl.DateTimeFormat</link> or
    <link xlink:href="&url.js.temporal-zoneddatetime;">Temporal.ZonedDateTime</link>.
   </simpara>
```

Girgias (approved):

PR #4887, opened 2025-09-28
Enable runnable (WASM) examples for datetime

AllenJB:
Related to #4799

Including fixing markup to make sure output replacement works correctly.

Additional changes I made:

- Normalize examples to omit unnecessary trailing `?>` (PER coding style, also I see new users using this too often and suspect the manual is partly to blame)
- Moved an example from date() to mktime() (it primarily uses mktime(), and there's alternative (IMO better) ways to do that);
- Added missing output examples where appropriate;
- Modified microtime() examples to more reliably output something other than "0";
- Corrected plurality of "above example(s)";
- Misc. other minor typos and updates for deprecations

The first example on strftime() isn't working. I suspect this is due to the WASM environment not having locales or something. Since the function is deprecated I decided not to both investigating too deeply and just marked it as not runnable.

### DIFF
--- a/features/persistent-connections.xml
+++ b/features/persistent-connections.xml
@@ -3,6 +3,8 @@
  <chapter xml:id="features.persistent-connections" xmlns="http://docbook.org/ns/docbook">
   <title>Persistent Database Connections</title>
 
+ <simplesect>
+  <title>What are Persistent Connections?</title>
   <simpara>
    Persistent connections are links that do not close when the
    execution of your script ends. When a persistent connection is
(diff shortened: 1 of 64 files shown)


PhilDaiguille (approved):

reference/datetime/dateperiod/construct.xml
```diff
@@ -231,6 +230,7 @@ foreach ($period as $date) {
     <programlisting role="php">
 <![CDATA[
 <?php
+Deprecated: Calling DatePeriod::__construct(string $isostr, int $options = 0) is deprecated, use DatePeriod::createFromISO8601String() instead in script on line 11
```
kamil-tekiela: Is this right? Did you accidently put this in the wrong place?

reference/datetime/datetime/modify.xml
```diff
@@ -92,20 +92,24 @@
 $date = new DateTime('2006-12-12');
 $date->modify('+1 day');
 echo $date->format('Y-m-d');
-?>
 ]]>
    </programlisting>
+   &example.outputs;
+   <screen>
+    <![CDATA[
```
kamil-tekiela:
```suggestion
<![CDATA[
```

reference/datetime/datetime/settimezone.xml
```diff
@@ -70,9 +70,15 @@ echo $date->format('Y-m-d H:i:sP') . "\n";
 
 $date->setTimezone(new DateTimeZone('Pacific/Chatham'));
 echo $date->format('Y-m-d H:i:sP') . "\n";
-?>
 ]]>
    </programlisting>
+   &example.outputs;
+   <screen>
+    <![CDATA[
```
kamil-tekiela:
```suggestion
<![CDATA[
```

kamil-tekiela: Why did you split it up? Both examples should produce identical output and there should only be one output section.

AllenJB: There's an issue with the interactive examples and the way they append / replace output in cases where there's multiple examples in the same example block.

Without an output block between them, they replace the next example instead.

I asked for advice and Derick suggested either duplicating the output or eliminating the procedural example.

There's probably a fix that can be applied to interactive-examples.js in web-php for this, but when I started work on this I had no previous experience with WASM and hadn't looked at the JS. I'm still wary of making changes to the interactive-examples JS that might possibly adversely affect something elsewhere in the manual.

PR #4890, opened 2025-10-02
Document the pipe operator.

Crell:
Most of the text and examples here are pulled straight from the RFC.

### DIFF
--- a/language/operators/functional.xml
+++ b/language/operators/functional.xml
@@ -1,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<sect1 xml:id="language.operators.functional">
+ <title>Functional Operators</title>
+ <titleabbrev>Functional</titleabbrev>
+ <para>
+  PHP 8.5 and later supports one operator that works directly on callables.  The <literal>|></literal>
+  operator, or “pipe,” accepts a single-parameter callable on the right and passes
+  the left-side value to it, evaluating to the callable's result.  The callable
+  on the right may be any valid PHP callable: a <classname>Closure</classname>,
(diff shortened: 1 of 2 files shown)


Fan2Shrek (approved):

language/operators/functional.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<sect1 xml:id="language.operators.functional">
+ <title>Functional Operators</title>
+ <titleabbrev>Functional</titleabbrev>
+ <para>
+  PHP 8.5 and later supports one operator that works directly on callables.  The <literal>|></literal>
```
TimWolla, line 6:
```suggestion
  PHP 8.5 and later supports one operator that works directly on callables.  The <literal>|&gt;</literal>
```
Bare `>` in XML is fugly (same below).

language/operators/functional.xml
```diff
@@ -0,0 +1,116 @@
...
+  <example>
+   <title>Using <literal>|></literal></title>
+   <programlisting role="php">
+<![CDATA[
+<?php
+$result = "Hello World" |> strlen(...);
+print $result . PHP_EOL;
+
+$result = strlen("Hello World");
+print $result . PHP_EOL;
```
TimWolla, line 24: Prefer `echo` and `, PHP_EOL` (with the comma).

TimWolla (changes requested):
Needs to be added to the precedence list: https://www.php.net/manual/en/language.operators.precedence.php

Also not particularly happy with calling this “Functional Operators”.

Crell:
I wasn't sold on "functional operator" either, but in Discord we didn't think of anything better.  I'm open to suggestions.

PR #4892, opened 2025-10-03
Mark xml_parser_free as deprecated.

Crell:
Not sure if anything else is really needed here.

### DIFF
--- a/reference/xml/functions/xml-parser-free.xml
+++ b/reference/xml/functions/xml-parser-free.xml
@@ -21,12 +21,6 @@
   <para>
    Frees the given XML <parameter>parser</parameter>.
   </para>
-  <warning>
-   <simpara>
-    This function is deprecated as of PHP 8.5.0.
-    It is no longer necessary.
-   </simpara>
-  </warning>
   <caution>
    <para>
     In addition to calling <function>xml_parser_free</function> when the parsing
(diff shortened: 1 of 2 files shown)


reference/xml/functions/xml-parser-free.xml
```diff
@@ -16,6 +16,12 @@
   <para>
    Frees the given XML <parameter>parser</parameter>.
   </para>
+  <warning>
+   <simpara>
+    This function is deprecated as of PHP 8.5.0.
+    It is no longer necessary.
+   </simpara>
+  </warning>
```
TimWolla: Use `&warn.deprecated.function-8-5-0;` at the very top. See:

https://github.com/php/doc-en/blob/84f25609071c1da9110a54776d0eda3d3709cacb/reference/random/functions/lcg-value.xml#L9-L11

reference/xml/functions/xml-parser-free.xml
```diff
@@ -6,16 +6,27 @@
...
   &note.resource-migration-8.0-dead-function;
   <para>
    Frees the given XML <parameter>parser</parameter>.
   </para>
+  <warning>
+   <simpara>
+    This function is deprecated as of PHP 8.5.0.
+    It is no longer necessary.
+   </simpara>
+  </warning>
```
TimWolla: This is redundant with changelog / warning above.
```suggestion
```

TimWolla (approved):

PR #4893, opened 2025-10-03
Note deprecated get_defined_functions argument.

Crell:


reference/funchand/functions/get-defined-functions.xml
```diff
@@ -24,7 +24,8 @@
     <term><parameter>exclude_disabled</parameter></term>
     <listitem>
      <para>
-      Whether disabled functions should be excluded from the return value.
+      Whether disabled functions should be excluded from the return value. This parameter has no effect
+      since PHP 8.0.0 and is deprecated since 8.5.0.  It should not be used.
```
TimWolla: Please make this consistent with:

https://github.com/php/doc-en/blob/84f25609071c1da9110a54776d0eda3d3709cacb/reference/random/functions/mt-srand.xml#L43-L62

Crell: Other than adding the entity reference, I'm not sure what else would be needed here?  The other is listing constants, this is just a boolean that no longer does anything.

TimWolla: I was referring to the entity reference / formatting to use. Sorry for being too succinct.

reference/funchand/functions/get-defined-functions.xml
```diff
@@ -53,11 +55,20 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.5.0</entry>
+      <entry>
+       The <parameter>exclude_disabled</parameter> parameter has been deprecated,
+       as it no longer has any effect.  It should no longer be used.
```
TimWolla: I think that is implied.
```suggestion
       as it no longer has any effect.
```

reference/funchand/functions/get-defined-functions.xml
```diff
@@ -24,8 +24,10 @@
     <term><parameter>exclude_disabled</parameter></term>
     <listitem>
      <para>
-      Whether disabled functions should be excluded from the return value.
+      Whether disabled functions should be excluded from the return value. This parameter has no effect
+      since PHP 8.0.0 and is deprecated since 8.5.0.  It should not be used.
```
TimWolla:
```suggestion
      since PHP 8.0.0.
```
The deprecation is mentioned below and the “don't use” is implied.

TimWolla (commented):
Please make

TimWolla (approved):

PR #4895, opened 2025-10-03
Add deprecation/changelog for switch statement semicolons.

Crell:
I'm not sure if it makes sense to use the "deprecated" entities here, since it's just a tiny part of the syntax that's deprecated, not the whole thing.  The "feature" warning seems like it would be overkill.  So I just adjusted the text and added a changelog.

### DIFF
--- a/language/control-structures/switch.xml
+++ b/language/control-structures/switch.xml
@@ -290,7 +290,7 @@ switch($beer)
 ]]>
    </programlisting>
   </informalexample>
-  This syntax has been deprecated as of PHP 8.5.0.
+  &warn.deprecated.feature-8-5-0;
  </para>
 
 
--- a/language/control-structures/switch.xml
+++ b/language/control-structures/switch.xml
@@ -269,7 +269,7 @@ endswitch;
   </informalexample>
  </para>
  <para>
-  It's possible to use a semicolon instead of a colon after a case like:
+  Prior to PHP 8.5.0, it was possible to use a semicolon instead of a colon after a case like:
   <informalexample>
    <programlisting role="php">
 <![CDATA[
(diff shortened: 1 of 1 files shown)


TimWolla:
> The "feature" warning seems like it would be overkill.

I find it appropriate. This what it's for. See also: https://www.php.net/manual/en/info.configuration.php#ini.assert.active where the snippet is used for various INIs.

language/control-structures/switch.xml
```diff
@@ -269,7 +269,7 @@ endswitch;
   </informalexample>
  </para>
  <para>
-  It's possible to use a semicolon instead of a colon after a case like:
+  Prior to PHP 8.5.0, it was possible to use a semicolon instead of a colon after a case like:
```
Girgias: But it is still possible, just deprecated.

Girgias (commented):
Using the XML entity seems fine.

language/control-structures/switch.xml
cweiske: @Crell could you please reword it?

Crell: Hm.  If we're adding the deprecation marker, should this line change at all?

Girgias: I guess the line can remain the same, makes also less work for translators.

language/control-structures/switch.xml
```diff
@@ -290,8 +290,32 @@ switch($beer)
 ]]>
    </programlisting>
   </informalexample>
+  &warn.deprecated.feature-8-5-0;
  </para>
 
+
+ <section role="changelog">
```
lacatoire:
```suggestion
 <sect2 role="changelog">
```

language/control-structures/switch.xml
```diff
@@ -290,8 +290,32 @@ switch($beer)
...
+     <row>
+      <entry>8.5.0</entry>
+      <entry>
+       The alternate syntax using a semicolon after <literal>case</literal> has been deprecated.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
+ </section>
```
lacatoire:
```suggestion
 </sect2>
```

language/control-structures/switch.xml
Girgias:
```suggestion
  It's possible to use a semicolon instead of a colon after a case like:
```

Girgias (approved):

Issue #4540, opened 2025-03-16
[Docs] PDO MySQL behavior with persistent connections (compared to mysqli)


AllenJB:
**Affected page**

https://www.php.net/manual/en/pdo.connections.php#example-1063

**Current issue**

https://www.php.net/manual/en/mysqli.persistconns.php details several behaviors when using persistent connections with mysqli.

It's not clear/documented if PDO MySQL implements these same behaviors or not.

**Suggested improvement**

Clearly state in the PDO MySQL documentation whether any cleanup is performed when using persistent connections. If not, this should be clearly stated so developers are aware of the potential issues of persistent connections.

**Additional context**

PR #4897, opened 2025-10-04
Improve docs on Persistent (DB) Connections

AllenJB:
See #4540

### DIFF
--- a/features/persistent-connections.xml
+++ b/features/persistent-connections.xml
@@ -15,16 +15,22 @@
    the same username and the same password (where applicable).
   </simpara>
   <simpara>
-   People who aren't thoroughly familiar with the way web servers work
-   and distribute the load may mistake persistent connects for what
-   they're not. In particular, they do <emphasis>not</emphasis> give
-   you an ability to open 'user sessions' on the same link, they
-   do <emphasis>not</emphasis> give you an ability to build up a
-   transaction efficiently, and they don't do a whole lot of other
-   things. In fact, to be extremely clear about the subject,
-   persistent connections don't give you <emphasis>any</emphasis>
-   functionality that wasn't possible with their non-persistent
-   brothers.
+   There's no method of requesting a specific connection, or guaranteeing
+   whether you get an existing connection or a brand new one (if all existing
+   connections are in use, or the request is being served by a different worker,
+   which has a separate pool of connections).
+  </simpara>
+  <simpara>
+   This means that you cannot use PHP's persistent connections to, for example:
+  </simpara>
+  <simplelist>
+   <member>assign a specific database session to a specific web user</member>
+   <member>create a large transaction across multiple requests</member>
+   <member>initiate a query on one request and collect the results on another</member>
+  </simplelist>
+  <simpara>
+   Persistent connections do not give you <emphasis>any</emphasis>
+   functionality that wasn't possible with non-persistent connections.
   </simpara>
  </simplesect>
 
@@ -41,7 +47,7 @@
    Because it is destroyed after every request, any resources that it
    acquires (such as a link to an SQL database server) are closed when
    it is destroyed. In this case, you do not gain anything from trying
-   to use persistent connections -- they simply don't persist.
+   to use persistent connections - they simply don't persist.
   </simpara>
   <simpara>
    The second, and most popular, method is to run PHP-FPM, or PHP as a module
@@ -56,6 +62,18 @@
    every following page requesting SQL services can reuse the same
    established connection to the SQL server.
   </simpara>
+  <note>
+   <para>
+   You can check which method your web requests use by checking the value of
+   "Server API" in the output of <function>phpinfo</function> or the value of
+   <constant>PHP_SAPI</constant>, run from a web request.
+   </para>
+   <para>
+    If the Server API is "Apache 2 Handler" or "FPM/FastCGI", then persistent
+    connections will be used across requests served by the same worker. For any
+    other value, persistent connections will not persist after each request.
+   </para>
+  </note>
  </simplesect>
 
  <simplesect xml:id="persistent-connections.cli">
@@ -65,8 +83,8 @@
    connections are not shared between command-line scripts, so there is no
    value in using them in transient scripts such as crons or commands.
    However, they may be useful if, for example, you're writing a long-running
-   application server that serves many requests (or other connections) and each
-   may need their own database connection.
+   application server that serves many requests or tasks and each may need
+   their own database connection.
   </simpara>
  </simplesect>
 
@@ -101,18 +119,35 @@
    threads attempt to connect, one will not be able to. If there are
    bugs in your scripts which do not allow the connections to shut
    down (such as infinite loops), the database with only 16 connections
-   may be rapidly swamped. Check your database documentation for
-   information on handling abandoned or idle connections (timeouts).
+   may be rapidly swamped.
+  </simpara>
+  <simpara>
+   Persistent connections will usually increase the number of connections open
+   at any given time because idle workers will still hold the connections for
+   the previous requests they served. If a large number of workers is spun up to
+   handle an influx of requests, the connections they opened will remain until
+   the worker is killed or the database server closes the connection.
+  </simpara>
+  <simpara>
+   Ensure that the maximum number of connections allowed by the database server
+   is greater than the maximum number of web request workers (plus any other
+   usage such as crons or administrative connections).
+  </simpara>
+  <simpara>
+   Check your database documentation for information on handling abandoned or
+   idle connections (timeouts). Long timeouts may significantly increase the
+   number of persistent connections open at any one time.
   </simpara>
  </simplesect>
 
  <simplesect xml:id="persistent-connections.drawbacks.state">
   <title>Potential Drawbacks: Maintaining Connection State</title>
   <simpara>
-   Depending on your application and the database or extension you're using you
-   may need to ensure that the state of connections is cleaned up before the
-   script exits. Changes that may leave connections in an unexpected state
-   include:
+   Some database extensions perform automatic cleanup when the connection is
+   reused; others leave this task at the discretion of the application developer.
+   Depending on the chosen database extension and the application design, manual
+   cleanup may be needed before the script exits. Changes that may leave
+   connections in an unexpected state include:
   </simpara>
   <simplelist>
    <member>Selected / default database</member>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Though I may add that this is a little bit ambiguous for a PHP manual. If anyone is using PDO persistent connection right now then I assume they already know how to keep the connections clean or they don't run into this problem much. So just saying that PDO doesn't do any clean up isn't as useful as it could be. For a new user some questions arise: what is this clean up? how can I do it manually? when do I need to do it?

features/persistent-connections.xml
```diff
@@ -20,13 +22,14 @@
    do <emphasis>not</emphasis> give you an ability to build up a
    transaction efficiently, and they don't do a whole lot of other
    things. In fact, to be extremely clear about the subject,
-   persistent connections don't give you <emphasis>any</emphasis>
+   persistent connections don't give you<emphasis>any</emphasis>
```
kamil-tekiela: Missing space.

features/persistent-connections.xml
```diff
@@ -20,13 +22,14 @@
...
-  <simpara>
-   Why?
-  </simpara> 
+ </simplesect>
+
+ <simplesect xml:id="persistent-connections.web">
+  <title>Web Requests</title>
   <simpara>
    This has to do with the way web servers work. There are three ways
    in which your web server can utilize PHP to generate web pages.
```
kamil-tekiela: This whole paragraph was confusing and it's even more so now after you change the Why? to Web Requests.

features/persistent-connections.xml
```diff
@@ -42,33 +45,43 @@
   </simpara>
   <simpara>
    The second, and most popular, method is to run PHP as a module in a
-   multiprocess web server, which currently only includes Apache. A
-   multiprocess server typically has one process (the parent) which
+   multiprocess web server, which currently only includes Apache, or PHP-FPM.
```
kamil-tekiela: I am not sure how to understand the addition "or PHP-FPM".

features/persistent-connections.xml
```diff
@@ -42,33 +45,43 @@
...
+   for the multiprocess model described before.
   </simpara>
+ </simplesect>
+
+ <simplesect xml:id="persistent-connections.cli">
+  <title>Commandline Processes</title>
   <simpara>
-   If persistent connections don't have any added functionality, what
-   are they good for?
+   As commandline PHP uses a new process for each script, persistent
```
kamil-tekiela:
```suggestion
   As command-line PHP uses a new process for each script, persistent
```
or should it be "command line"?

features/persistent-connections.xml
```diff
@@ -42,33 +45,43 @@
...
+   connections are not shared between commandline scripts, so there is no
```
kamil-tekiela:
```suggestion
   connections are not shared between command-line scripts, so there is no
```

features/persistent-connections.xml
```diff
@@ -42,33 +45,43 @@
...
+   value in using them in (short-lived) scripts such as crons or commands.
```
kamil-tekiela: What about long-lived scripts? How do they help then?

features/persistent-connections.xml
```diff
@@ -92,40 +109,60 @@
...
    may be rapidly swamped. Check your database documentation for
-   information on handling abandoned or idle connections.
+   information on handling abandoned or idle connections (timeouts).
+  </simpara>
+ </simplesect>
+
+ <simplesect xml:id="persistent-connections.drawbacks.state">
+  <title>Potential Drawbacks: Cleanup / State Management</title>
+  <simpara>
+   Depending on your application and the database / extension you're using you
```
kamil-tekiela:
```suggestion
   Depending on your application and the database/extension you're using you
```
As I understand, in English we do not put spaces around slashes.

features/persistent-connections.xml
```diff
@@ -92,40 +109,60 @@
...
-    simple cleanup function to unlock your tables or roll back your
-    transactions. Better yet, avoid the problem entirely by not using
-    persistent connections in scripts which use table locks or
-    transactions (you can still use them elsewhere).
-   </simpara>
-  </warning>
+  <simplelist>
+   <member>Selected / default database</member>
+   <member>Table locks</member>
+   <member>Unclosed transactions</member>
```
kamil-tekiela:
```suggestion
   <member>Uncommitted transactions</member>
```

features/persistent-connections.xml
```diff
@@ -92,40 +109,60 @@
...
-   information on handling abandoned or idle connections.
+   information on handling abandoned or idle connections (timeouts).
+  </simpara>
+ </simplesect>
+
+ <simplesect xml:id="persistent-connections.drawbacks.state">
+  <title>Potential Drawbacks: Cleanup / State Management</title>
+  <simpara>
+   Depending on your application and the database / extension you're using you
+   may need to ensure that the state of connections is cleaned up. Changes
```
kamil-tekiela: cleaned up when?

features/persistent-connections.xml
```diff
@@ -92,40 +109,60 @@
...
+  </simpara>
+  <simpara>
+   If temporary tables are not cleaned up, subsequent requests will not be able
+   to recreate the same table.
+  </simpara>
+  <simpara>
+   You can implement cleanup using class destructors or
+   <function>register_shutdown_function</function>. You may also want to
+   consider dedicated connection pooling proxies that include this as part of
+   their functionality.
```
kamil-tekiela, line 177: I'd remove this. Without a link or further explanation, we leave the user confused as to what it is and how to use it.

kamil-tekiela (changes requested):
I like your changes. They are very good, but I think the page needs more work after this. I'll try to think about this later.

features/persistent-connections.xml
AllenJB: I've reworded this. 

I also ended up removing the entire "third way" because as far as I can tell it no longer exists from PHP's point of view - the NSAPI and ISAPI (which I'm pretty sure is what WSAPI is referring to here - "WSAPI PHP" on google gives no useful results) SAPIs no longer exist. IIS just uses FastCGI.

features/persistent-connections.xml
AllenJB: Reworded so it's hopefully clearer. I'm simply trying to clearly indicate what PHP-FPM users can expect.

features/persistent-connections.xml
AllenJB: They're mentioned in the next sentence. I've changed this to "transient" which may be a better fit here.

features/persistent-connections.xml
AllenJB: I think it's useful to make users aware these tools (with these capabilities) exist.

I don't think there's a necessity to link to specific tools - as I mentioned above I'm both inexperienced with these and want to avoid the manual appearing to recommend specific tools (which may then become unmaintained, or have better alternatives available).

(As someone who hangs out on various forums and chats, there are still have users picking up XAMPP, now 2 years after its last release, and I've been specifically told by at least 1 recently that it was because it was the first tool mentioned in the manual - and still is today!)

Also, the tools are going to be different depending on what database (and/or cloud provider) is used, so this would end up a list of different tools.

I think this provides just enough information to make users aware of these tools and give them the terms to search for them.

features/persistent-connections.xml
```diff
@@ -92,40 +102,61 @@
...
-    persistent connections in scripts which use table locks or
-    transactions (you can still use them elsewhere).
-   </simpara>
-  </warning>
+ </simplesect>
+
+ <simplesect xml:id="persistent-connections.drawbacks.state">
+  <title>Potential Drawbacks: Maintaining Connection State</title>
   <simpara>
-   An important summary. Persistent connections were designed to have
```
adrian-enspired, line 116: I would not simply remove this paragraph, especially since you're adding a "drawbacks" section.
the biggest issue i've seen with persistent connections is not people failing to clean up after themselves, but with how they behave more generally (i.e., the "one-to-one" concept mentioned here).

i've seen many beginners, or even experienced programmers that have little exposure to managing sapis/databases/etc., start using persistent connections without sufficient resources (e.g., in a cheap shared hosting environment) or without careful configuration (e.g., assuming you only need one persistent connection because it can be "shared"). in these cases, you're likely to see _worse_ performance. this is actually the only thing i've ever had to explain to a programmer seeking help with persistent connections, and i've had to explain it quite a few times.

AllenJB: The problem I had with this paragraph (and this appears to me to be [echoed by at least one user comment](https://www.php.net/manual/en/features.persistent-connections.php#117523)) is it reads to me as "you can simply enable persistent connections and not have to worry about making any other changes", which obviously isn't true, particularly in cases where PHP is not performing any additional management (eg. PDO MySQL).

Persistent connections can change behavior of scripts when cleanup is not performed.

The situation is not helped by apparent differences in behavior between different extensions. The cleanup performed by mysqli is just one manifestation.

Comparing https://github.com/php/php-src/blob/master/ext/pdo_mysql/tests/pdo_mysql_pconnect.phpt and https://github.com/php/php-src/blob/master/ext/mysqli/tests/mysqli_pconn_reuse.phpt shows PDO MySQL will happily reuse connections even if they are already in use by an existing PDO instance in the same process, while mysqli will not reuse connections until they're "closed". I haven't investigated whether it's required to explicitly close the connection or if this also happens if the mysqli object goes out of scope and is gc'd - I don't know whether there's a difference internally between these 2 actions when dealing with persistent connections.

I did have a quick check to see if there's similar tests that show how PDO PgSQL behaves, but couldn't see any, and I don't have a local postgres server running to perform my own tests. According to https://github.com/php/php-src/blob/master/ext/pgsql/tests/connection_reuse.phpt and https://github.com/php/php-src/blob/master/ext/pgsql/tests/gh13519.phpt both pg_connect and pg_pconnect will give you the same connection instance even if it's in active use by another PgSql\Connection object unless you explicitly request a new connection.

As far as I can see, "persistent connections were designed to be a one to one mapping to regular connections" is not (always) true (depending on the extension / database you're using) in any sense.

It's possible I am misunderstanding what "one-to-one mapping" is supposed to mean in this context, but I appear to not be the only one. Please feel free to submit suggestions for an alternative if you believe it's something not covered by the new version of this page, but I do not believe this paragraph should be restored.

kamil-tekiela: I think neither option is good. The previous paragraph was confusing, but the lack of it isn't good either. For now, we can remove it and we can add something better later.

features/persistent-connections.xml
```diff
@@ -41,34 +44,37 @@
...
-  </simpara>
-  <simpara>
-   If persistent connections don't have any added functionality, what
-   are they good for?
+   As command-line PHP uses a new process for each script, persistent
+   connections are not shared between command-line scripts, so there is no
+   value in using them in transient scripts such as crons or commands.
+   However, they may be useful if, for example, you're writing a long-running
+   application server that serves many requests (or other connections) and each
+   may need their own database connection.
```
kamil-tekiela: I still don't like this part. Are you saying that persistent connections should only be used with command-line scripts if they take a long time to execute? Also, doesn't this information belong in the next section?

features/persistent-connections.xml
```diff
@@ -92,40 +102,61 @@
...
-   one-to-one mapping to regular connections. That means that you
-   should <emphasis>always</emphasis> be able to replace persistent
-   connections with non-persistent connections, and it won't change
-   the way your script behaves. It <emphasis>may</emphasis> (and
-   probably will) change the efficiency of the script, but not its
-   behavior!
+   Depending on your application and the database or extension you're using you
+   may need to ensure that the state of connections is cleaned up before the
+   script exits. Changes that may leave connections in an unexpected state
+   include:
```
kamil-tekiela:
```suggestion
   Some database extensions perform automatic cleanup when the connection is reused;
   others leave this task at the discretion of the application developer.
   Depending on the chosen database extension and the application design,
   manual cleanup may be needed before the
   script exits. Changes that may leave connections in an unexpected state
   include:
```

features/persistent-connections.xml
AllenJB: I've made some further amendments across the article that I hope help to clarify the behavior / make the sections on behavior easier to quickly read, and added a "Final Words" section to reiterate the potential for issues and alternative (probably better) solutions.

PR #4903, opened 2025-10-06
isinstance.xml Make the example workable and semantic

mmalferov:
Let's name the variable not $arg, but $object, according to the parameter name, and add the missing definitions

### DIFF
--- a/reference/reflection/reflectionclass/isinstance.xml
+++ b/reference/reflection/reflectionclass/isinstance.xml
@@ -49,7 +49,6 @@
 <![CDATA[
 <?php
 
-// Example usage
 class Foo {}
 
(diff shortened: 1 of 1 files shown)


reference/reflection/reflectionclass/isinstance.xml
```diff
@@ -48,20 +48,25 @@
     <programlisting role="php">
 <![CDATA[
 <?php
+
 // Example usage
```
kamil-tekiela:
```suggestion
```
By definition, an example shows example usage.

kamil-tekiela (changes requested):

kamil-tekiela (approved):

PR #4908, opened 2025-10-07
CLI Web Server: Add a note about `php --help` (fixes #4587)

### LINKED ISSUE #4587: Dev Server command argument order issue undocumented
From manual page: https://php.net/features.commandline.webserver

---
When invoking the command line development server, if using a routing script, the script file or path MUST be specified AFTER  the -t argument specifying the document root path, if both are being used. Reversing the order of these arguments causes failure to serve requested resources unless explicitly included, read, or fileread by the script.

I could not find anything on https://php.net specifying this or indicating any issue.

Environment: 
    MacBook Air 2025 - M4
    PHP v8.4.4
    MacOS 15.4 (24E248) Sequoia
    Firefox 136.0.4
    Chrome 135.0.7049.42 (Official Build) (arm64)

AllenJB:


TimWolla (commented):
I'm not sure how I feel about specifically calling out `--help` for the internal web server, particularly since the explanation also requires the “not all options will apply” caveat.

AllenJB:
Specifically useful here, the output of `--help` includes the required order for parameters for `-S` usage.

I think users are likely to end up directly on this page if they're having problems with the built-in server (as in the linked issue) and teaching them where they can find this information right at the commandline (and specific to their version of PHP) is useful.

Many users are not familiar with the common conventions of (*nix) commands.

I think the caveat is helpful, and once aware users can be expected to work out which options likely can be applied when running the built-in server.

I'm open to suggestions for how to modify the documentation to help users in cases like the linked issue, keeping in mind that there are already examples of correct usage on the page.

lacatoire (approved):

PR #4912, opened 2025-10-09
Improve the documentation of mysqli_stmt_result_metadata()

kamil-tekiela:


reference/mysqli/mysqli_stmt/result-metadata.xml
```diff
@@ -25,8 +25,15 @@
    that can be used to process the meta information such as total number of fields
    and individual field information.
   </para>
+  <para>
+   This function returns an empty <classname>mysqli_result</classname> object
+   which can be used to access metadata information from the prepared statement
+   without having to fetch the actual rows of data. There is no need to use
+   this function when using <function>mysqli_stmt_get_result</function> to
+   retrieve the entire result set from a prepared statement as a result object.
+  </para>
```
Girgias:
```suggestion
  <simpara>
   This function returns an empty <classname>mysqli_result</classname> object
   which can be used to access metadata information from the prepared statement
   without having to fetch the actual rows of data. There is no need to use
   this function when using <function>mysqli_stmt_get_result</function> to
   retrieve the entire result set from a prepared statement as a result object.
  </simpara>
```

reference/mysqli/mysqli_stmt/result-metadata.xml
```diff
@@ -25,8 +25,15 @@
...
   <note>
-   <para>This result set pointer can be passed as an argument to any of the
+   <para>This result set object can be passed as an argument only to the
```
Girgias:
```suggestion
   <para>
    This result set object can be passed as an argument only to the
```

reference/mysqli/mysqli_stmt/result-metadata.xml
```diff
@@ -40,15 +47,12 @@
     </itemizedlist>
    </para>
   </note>
-  <para>
-   The result set structure should be freed when you are done with it,
-   which you can do by passing it to <function>mysqli_free_result</function>
-  </para>
   <note>
    <para>
```
Girgias:
```suggestion
   <simpara>
```

reference/mysqli/mysqli_stmt/result-metadata.xml
```diff
@@ -40,15 +47,12 @@
...
-    The result set returned by <function>mysqli_stmt_result_metadata</function> contains only
-    metadata. It does not contain any row results. The rows are obtained by using the
-    statement handle with <function>mysqli_stmt_fetch</function>.
+    The result set returned by <function>mysqli_stmt_result_metadata</function>
+    contains only metadata. It does not contain any row results.
+    The rows are obtained by calling <function>mysqli_stmt_get_result</function>
+    on the statement handle or with <function>mysqli_stmt_fetch</function>.
    </para>
```
Girgias:
```suggestion
   </simpara>
```

reference/mysqli/mysqli_stmt/result-metadata.xml
```diff
@@ -66,6 +70,7 @@
   &reftitle.returnvalues;
   <para>
    Returns a result object or &false; if an error occurred.
+   If the statement does not produce a result set, &false; is returned as well.
   </para>
```
Girgias:
```suggestion
  <simpara>
   Returns a result object or &false; if an error occurred.
   If the statement does not produce a result set, &false; is returned as well.
  </simpara>
```

Issue #4803, opened 2025-07-31
Improve finfo_file documentation to state that the file must be local


socivdias:
**Affected page**

https://www.php.net/manual/en/function.finfo-file.php

**Current issue**

Many functions, such as `file_get_contents`, have a `$filepath` parameter that can be a remote URL (thanks to PHP stream wrappers).

`finfo_file` has a `$filepath` parameter that needs to be a path to a local file, because behind the scenes it does not use the stream wrappers (AFAIK).

This could be clearer in the documentation to prevent surprises.

**Suggested improvement**

Clearly mention in the documentation that `$filepath` must be a local file and remote URLs are not accepted.

ndossche:
The docs ain't wrong, php-src is wrong. The stat call doesn't work for remote resources, causing issues.
I'll open a php-src issue.

PR #4913, opened 2025-10-09
edit description for filename parameter in finfo-file.xml (php#4803)

vera-ju:


reference/fileinfo/functions/finfo-file.xml
```diff
@@ -44,7 +44,7 @@
      <term><parameter>filename</parameter></term>
      <listitem>
       <para>
-       Name of a file to be checked.
+       Name of a local file to be checked. Remote URLs are not supported.
```
kamil-tekiela, line 47: Should it not say URIs? Also, what is the source of this information?

vera-ju: To be completely honest, I'm not a PHP developer — I just came across this issue during Hacktoberfest.

From my point of view, there are three indicators that confirm this is correct:

1) The open issue states it very clearly: (https://github.com/php/doc-en/issues/4803)
“Suggested improvement: Clearly mention in the documentation that $filepath must be a local file and remote URLs are not accepted.”

2) I asked ChatGPT about it, and it confirmed the same.

3) A PHP colleague double-checked it and tested the function with an external URL — it didn’t work. However, using a local file worked correctly.

But if you don’t want to have it written that way in the documentation, feel free to just decline the PR.

Of course, I don’t want anything “incorrect” to be added either.
If you’d like, you can double-check it yourselves and adjust the wording however you think fits best 🙂

kamil-tekiela: I don't know why it doesn't work with remote files, because it does use streams. Perhaps someone else more knowledge can review this PR.

ndossche: No, this looks like a bug in php-src. It _does_ use streams, but the fact it calls stat first fucks it up.

PR #4915, opened 2025-10-11
session_gc: Add note regarding session.save_path and file access / permissions

### LINKED ISSUE #2263: session_gc works per-sapi
I'm not even sure if I am saying this correctly but bear with me.  We are moving an application from Ubuntu Server to RHEL.  In doing so, we are looking at ways to re-implement a similar way that Debian-based systems clear PHP sessions.  I, like a few others, thought that the [session_gc()](https://www.php.net/manual/en/function.session-gc.php) function would meet the need.  The first example even states that it would be ran via task manager such as cron.  But, as others have found[^1][^2], this does not work.  I am unable to find a reference to SAPI specific handling in the general [Session](https://www.php.net/manual/en/book.session.php) documentation.  The only piece that seems of minor relevance is this part on [Session Data Deletion](https://www.php.net/manual/en/features.session.security.management.php).

I think the documentation needs to be updated to reflect the scenarios where `session_gc` actually works.  The first example may even need to be removed entirely unless it can actually work with some changes.

[^1]: https://stackoverflow.com/questions/67543926/php-garbage-collection-triggered-by-cron-not-working#comment119422612_67543926
[^2]: https://www.drupal.org/project/drupal/issues/3033791#comment-14476110

AllenJB:
Fixes #2263

### DIFF
--- a/reference/session/functions/session-gc.xml
+++ b/reference/session/functions/session-gc.xml
@@ -36,7 +36,7 @@
     <link linkend="ini.session.save-path">session.save_path</link> must be set
     to the same value as web requests and the script must have access and
     permissions for the same files. This may be affected by the user the script
-    runs as and container / sandbox features such as SystemD's PrivateTmp.
+    runs as and container / sandboxing features such as systemd's <literal>PrivateTmp=</literal> option.
    </para>
   </note>
  </refsect1>


reference/session/functions/session-gc.xml
```diff
@@ -30,6 +30,15 @@
    <link linkend="ini.session.gc-probability">session.gc_probability</link> 
    to 0.
   </para>
+  <note>
+   <para>
+    For <function>session_gc</function> to work correctly, the
+    <link linkend="ini.session.save-path">session.save_path</link> must be set
+    to the same value as web requests and the script must have access and
+    permissions for the same files. This may be affected by the user the script
+    runs as and container / sandbox features such as SystemD's PrivateTmp.
```
TimWolla:
```suggestion
    runs as and container / sandboxing features such as systemd's <literal>PrivateTmp=</literal> option.
```
Something like this?

Girgias: The second sentence reads weird to me, could you improve it?

PR #4917, opened 2025-10-11
pcntl_wait / pcntl_waitpid: Document the rusage parameter (fixes #861)

### LINKED ISSUE #861: Undocumented resource_usage parameter of pcntl_wait[pid] function
https://www.php.net/manual/en/function.pcntl-waitpid.php
https://www.php.net/manual/en/function.pcntl-wait.php

pcntl_wait[pid] take a parameter `array &$resource_usage = []`, which stores `struct rusage` [*1] value.

[*1] https://man7.org/linux/man-pages/man2/getrusage.2.html

AllenJB:


reference/pcntl/functions/pcntl-wait.xml
```diff
@@ -99,12 +109,71 @@
   </para>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title><function>pcntl_wait</function> example</title>
```
TimWolla: Given that `pcntl_wait` is documented to be equivalent to `pcntl_waitpid` with specific parameters, I don't think we should duplicate an example as complex as this for maintenance purposes.

TimWolla (changes requested):
`pcntl_waitid` already has the parameter documented. Can you align the explanation: https://www.php.net/manual/en/function.pcntl-waitid.php?

reference/pcntl/functions/pcntl-wait.xml
AllenJB: Removed

reference/pcntl/functions/pcntl-waitpid.xml
```diff
@@ -143,12 +157,71 @@
...
+    $deadPid = pcntl_waitpid($pid, $status, $flags, $rusage);
+    if ($deadPid === -1) {
+        $errNo = pcntl_get_last_error();
+        $errMsg = pcntl_strerror($errNo);
+        print "PARENT pcntl_waitpid FAILED ({$deadPid}): {$errMsg}\n";
+    } elseif ($deadPid === 0) {
+        print "PARENT pcntl_waitpid: No exited children.\n";
+    } else {
+        print "PARENT pcntl_waitpid: Child PID {$deadPid} exited with status: "
+            . pcntl_wexitstatus($status) ."\n";
```
TimWolla: The documentation for wexitstatus says:

> This function is only useful if [pcntl_wifexited()](https://www.php.net/manual/en/function.pcntl-wifexited.php) returned [true](https://www.php.net/manual/en/reserved.constants.php#constant.true).

So, is a `wifexited` call missing here? Looking more carefully at the example, it uses quite a few functions from the pcntl extension. It might make sense to include it in https://www.php.net/manual/en/pcntl.example.php instead.

TimWolla (commented):
Except for the example this LGTM. Perhaps it makes sense to split the example off in another PR?

reference/pcntl/functions/pcntl-waitpid.xml
```diff
@@ -143,12 +157,71 @@
   </para>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
```
Girgias: Useless wrapping para tag

reference/pcntl/functions/pcntl-wait.xml
```diff
@@ -86,6 +86,20 @@
...
+       The <parameter>resource_usage</parameter> parameter is set to an
+       associative <type>array</type> containing resource usage statistics
+       from the child process.
+
+       This is supported either if the wait6 system call is available
+       (e.g. on FreeBSD), or on Linux through the raw waitid system call. For
+       information on the contents see <function>getrusage</function>.
+      </para>
+     </listitem>
+    </varlistentry>
```
Girgias, line 106: Considering how often this is repeated, an XInclude or XML entity to have consistent wording would be better

PR #4918, opened 2025-10-11
Missing function documentation for pcntl_wifcontinued and pcntl_forkx

AllenJB:
Related to #2163 

As I don't have a Solaris box, the pcntl_forkx documentation is based on the php-src code and comments, combined with [the Solaris documentation](https://docs.oracle.com/cd/E86824_01/html/E54765/forkx-2.html)

pcntl_wifcontinued is wrapped in an ifdef in php-src. Based on commit / NEWS history this was because it wasn't available on NetBSD (but based on search results appears to be in more recent versions). Is this significant enough to note? eg. A note along the lines of:

> This function may not be available on some systems. Check you systems fork(2) man page for availability.

### DIFF
--- a/reference/pcntl/functions/pcntl-forkx.xml
+++ b/reference/pcntl/functions/pcntl-forkx.xml
@@ -1,0 +1,135 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.pcntl-forkx" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>pcntl_forkx</refname>
+  <refpurpose>Forks the currently running process, with additional options (Solaris only)</refpurpose>
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>int</type><methodname>pcntl_forkx</methodname>
+   <methodparam><type>int</type><parameter>flags</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   The <function>pcntl_forkx</function> function creates a child
+   process that differs from the parent process only in its PID and
+   PPID. Please see your system's forkx(2) man page for specific details as to
+   how forkx works on your system.
+  </para>
+  <para>
+   This function is only available on <productname>Solaris</productname>.
+  </para>
+ </refsect1>
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <para>
+   <variablelist>
+    <varlistentry>
+     <term><parameter>flags</parameter></term>
+     <listitem>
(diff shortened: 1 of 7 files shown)


reference/pcntl/functions/pcntl-forkx.xml
```diff
@@ -0,0 +1,135 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
+<refentry xml:id="function.pcntl-forkx" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
+  <refname>pcntl_forkx</refname>
+  <refpurpose>Forks the currently running process, with additional options (Solaris only)</refpurpose>
```
devnexen, line 6: nit: ```Solaris only```, illumos systems support it too

reference/pcntl/functions/pcntl-forkx.xml
```diff
@@ -0,0 +1,135 @@
...
+ </refnamediv>
+
+ <refsect1 role="description">
+  &reftitle.description;
+  <methodsynopsis>
+   <type>int</type><methodname>pcntl_forkx</methodname>
+   <methodparam><type>int</type><parameter>flags</parameter></methodparam>
+  </methodsynopsis>
+  <para>
+   The <function>pcntl_forkx</function> function creates a child
```
devnexen, line 16: note that forkx, `x` stands for `extended version of`. Also what is mentioned here is true, unless you pass flags so even though there is 'read the manual if you want to know more', a little mention of these would not kill :)

reference/pcntl/functions/pcntl-forkx.xml
AllenJB: At a glance illumos is OpenSolaris, so surely it just comes under the "Solaris" banner, so does it need explicitly mentioning? Since refpurpose is commonly used in places such as search result description and "see also" lists, shouldn't it be kept as short as possible?

I don't actually use Solaris myself. I just had some time to contribute and put that towards documenting undocumented functions as best I could.

reference/pcntl/functions/pcntl-forkx.xml
AllenJB: I'm not sure what you want me to add here. As far as I can tell from the man page and PHP source code none of the flags (which I've already included documentation for) substantially change this description.

I was also trying to avoid unnecessary (and poorly summarized) reproduction of information that is better documented elsewhere. I could include links to man pages on the internet, but (as well as requiring an additional PR to doc-base) these may not necessarily be up-to-date and I should think most users looking to use these functions are aware (or can quickly find) how to use man pages.

AllenJB:
Closing: This documentation has since been added by #5350

PR #4919, opened 2025-10-12
setisodate.xml Fix an example

mmalferov:
The first example seems to have been moved from a mutable DateTime :) We need to assign the result of the method to a variable


kamil-tekiela (approved):
Thanks!

PR #4924, opened 2025-10-13
escapeshellarg / escapeshellcmd: Document multibyte string behavior

### LINKED ISSUE #1452: escapeshellarg documentation is missing the locale dependency
The following paragraph should be added:

escapeshellarg() will strip all invalid characters according to your locale settings (e.g. latin-1 characters are stripped when locale/LC_CTYPE is UTF-8). Please keep in mind that the locale support depends on you C standard library while compiling. This might result in strange behavior on embedded systems.

### LINKED ISSUE #3052: escapeshellarg() strips off bytes which are not valid characters (leading to file not found errors)
### Description

It seems `escapeshellarg()` strips off bytes which are >= 128 probably because they are not valid UTF-8 characters.

Because it's not possible to provide an encoding to `escapeshellarg()` and this behaviour is not [documented](https://www.php.net/manual/en/function.escapeshellarg.php) I think it's a bug.

Also if this is expected behaviour then which function could be used instead? There is no other function which could do this.

Please note that the code example runs on Linux only because Windows (probably) doesn't have the `cat` command. Also I tested this on Linux only. Also note it's not necessary to close the php tag.

```php
<?php
$fname = "asdf.txt".chr(128);
file_put_contents($fname, "hey!");
$cmd = "cat ".escapeshellarg($fname);
echo "cmd: $cmd\n";
passthru($cmd);
```

Resulted in this output:
```
cmd: cat 'asdf.txt'
cat: asdf.txt: No such file or directory
```

```
cmd: cat 'asdf.txt'$'\200'
hey!
```

### PHP Version

PHP 8.2.7 (cli) (built: Jun  9 2023 19:37:27) (NTS)

### Operating System

Debian GNU/Linux 12 (bookworm)

AllenJB:
Fixes #1452, Fixes #3052


lacatoire (approved):

reference/exec/functions/escapeshellarg.xml
```diff
@@ -29,6 +29,11 @@
    Furthermore, each streak of consecutive backslashes (<literal>\</literal>)
    is escaped by one additional backslash.
   </para>
+  <para>
+   The behavior of this function with multibyte strings depends on the current
+   <constant>LC_CTYPE</constant> locale setting. Unrecognized characters will be
+   discarded. See <function>setlocale</function>
```
cweiske: Add a . at the end. Maybe even "discarded - see .."

cweiske (changes requested):

lacatoire (approved):

PR #4933, opened 2025-10-16
Rewrite the callback and callable page

kamil-tekiela:


language/types/callable.xml
```diff
@@ -1,25 +1,124 @@
...
+  <para>
+   Callables can be created in several different ways:
+  </para>
+
+  <itemizedlist>
+   <listitem>
+    <simpara><classname>Closure</classname> object</simpara>
+   </listitem>
+   <listitem>
+    <simpara>String containing the name of a function or a method</simpara>
```
Vectorial1024: > String containing the name of a function or a method

I think you are trying to refer to e.g. `usort(...)`? Then, iirc these are functions. iirc a method is always a member of a class, and so methods can't exist as itself out in the open.

kamil-tekiela: Static methods can.

language/types/callable.xml
```diff
@@ -1,25 +1,124 @@
...
+   </listitem>
+   <listitem>
+    <simpara>Array containing an <type>object</type> or class name
```
Vectorial1024: > Array containing an <type>object</type> or class name

Recommend changing the order to be "a class name or an object", to avoid English ambiguity (e.g. "object name or class name"?)

language/types/callable.xml
```diff
@@ -1,25 +1,124 @@
...
+
+  <example>
+   <title>
+    Callback example using a <classname>Closure</classname>
+   </title>
+  <programlisting role="php">
+<![CDATA[
+<?php
+// Using anonymous function syntax
+$double1 = function($a) {
```
Vectorial1024: Because arrow functions are mentioned, perhaps we should also include an example of arrow functions for completeness.

```php
// Using arrow function syntax
$double2 = fn ($a) => $a * 2;
```

language/types/callable.xml
```diff
@@ -29,133 +128,133 @@
...
+        echo 'Hello World!', "\n";
     }
 }
 
 // Type 1: Simple callback
 call_user_func('my_callback_function');
 
 // Type 2: Static class method call
-call_user_func(array('MyClass', 'myCallbackMethod'));
+call_user_func(['MyClass', 'myCallbackMethod']);
```
Vectorial1024, line 221: > call_user_func(['MyClass', 'myCallbackMethod']);

I think the following syntax is also valid/equivalent, and can also be included:

```php
call_user_func([MyClass::class, 'myCallbackMethod']);
```

This syntax has the added benefit that, because this requires an import or a fully qualified name, the user can simply follow the class symbol to go to the class definition in modern IDEs.

language/types/callable.xml
Vectorial1024: I see. I forgot about static methods upon first reading.

language/types/callable.xml
```diff
@@ -1,25 +1,131 @@
...
+function foo(callable $callback) {
+    $callback();
+}
+?>
+]]>
+  </programlisting>
+ </informalexample>
 
  <para>
   Some functions like <function>call_user_func</function> or
```
TimWolla: Should we replace `call_user_func()` with something else? I don't think `call_user_func()` has a place in modern code. Perhaps `preg_replace_callback()`?

language/types/callable.xml
```diff
@@ -1,25 +1,131 @@
...
-  parameter. Callback functions can not only be simple functions, but also
-  <type>object</type> methods, including static class methods.
+  parameter.
  </para>
 
  <sect2 xml:id="language.types.callable.passing">
-  <title>Passing</title>
+  <title>Creation of callables</title>
+
+  <para>
```
TimWolla: I think these should all be `<simpara>`.

language/types/callable.xml
```diff
@@ -1,25 +1,131 @@
...
+
+  <example>
+   <title>
+    Callback example using a <classname>Closure</classname>
+   </title>
+  <programlisting role="php">
+<![CDATA[
+<?php
+// Using anonymous function syntax
+$double1 = function($a) {
```
TimWolla: I think accepted style is this:
```suggestion
$double1 = function ($a) {
```

language/types/callable.xml
TimWolla: `exit()` is a function nowadays and should be removed from the list.

language/types/callable.xml
```diff
@@ -1,25 +1,131 @@
...
+   </listitem>
+   <listitem>
+    <simpara>String containing the name of a function or a method</simpara>
+   </listitem>
+   <listitem>
+    <simpara>Array containing a class name or an <type>object</type>
+    in index 0 and the method name in index 1</simpara>
+   </listitem>
+   <listitem>
+    <simpara>Object implementing <link linkend="object.invoke">__invoke()</link></simpara>
```
TimWolla:
```suggestion
    <simpara>Object implementing the <link linkend="object.invoke">__invoke()</link> magic method</simpara>
```
Or something like this?

language/types/callable.xml
```diff
@@ -29,133 +135,137 @@
...
+    Callback function examples
+   </title>
+   <programlisting role="php">
 <![CDATA[
 <?php
 
 // An example callback function
 function my_callback_function() {
-    echo 'hello world!', PHP_EOL;
+    echo 'hello world!', "\n";
```
TimWolla: The usual style in the documentation is `PHP_EOL`.

language/types/callable.xml
kamil-tekiela: Both appear in the manual and there's more of the one without space. I will change it though.

language/types/callable.xml
kamil-tekiela: I will use `usort` as I think it's easier.

kamil-tekiela: Actually, what do you think, is it better with two examples or just one?

language/types/callable.xml
kamil-tekiela: I don't think so. There is definitely more `"\n"` than `PHP_EOL` and I see no reason why we should use this constant here. We have been removing it in previous PRs so there is a precedent for it already.

kamil-tekiela: I have reverted it back per your request, but I believe that it is a bad practice to use this constant in a write context, and it's dangerous. We should not be teaching new users this.

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
- <para>
-  Some functions like <function>call_user_func</function> or
-  <function>usort</function> accept user-defined callback functions as a
-  parameter. Callback functions can not only be simple functions, but also
-  <type>object</type> methods, including static class methods.
- </para>
+ <simpara>
+  Some functions accept user-defined callback functions as a parameter, e.g.
+  <function>array_map()</function>, <function>usort()</function>, or
+  <function>preg_replace_callback()</function>.
```
TimWolla:
```suggestion
  <function>array_map</function>, <function>usort</function>, or
  <function>preg_replace_callback</function>.
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
+
+  <itemizedlist>
+   <listitem>
+    <simpara><classname>Closure</classname> object</simpara>
+   </listitem>
+   <listitem>
+    <simpara>String containing the name of a function or a method</simpara>
+   </listitem>
+   <listitem>
+    <simpara>Array containing a class name or an <type>object</type>
```
TimWolla:
```suggestion
    <simpara>&array; containing a class name or an <type>object</type>
```

TimWolla (approved):
I'll let someone else take another look, but LGTM from my side.

language/types/callable.xml
kamil-tekiela: Won't that make it lowercase?

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
+  <simpara>
+   A callable can also be a string containing the name of a function or
+   a static method.
    Any built-in or user-defined function can be used, except language constructs
    such as: <function>array</function>, <function>echo</function>,
    <function>empty</function>, <function>eval</function>,
-   <function>exit</function>, <function>isset</function>,
+   <function>isset</function>,
    <function>list</function>, <function>print</function> or
    <function>unset</function>.
```
Girgias, line 140: It should be noted somewhere that `exit()`/`die()` were language constructs prior to 8.4.

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
 <sect1 xml:id="language.types.callable">
- <title>Callbacks / Callables</title>
+ <title>Callbacks and Callables</title>
```
Girgias: I feel we should just get rid of the notion of "callbacks" here, as that's a programming pattern rather than a "construction".
```suggestion
 <title>Callables</title>
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
 
  <para>
+  A callback is a reference to a function or method that is passed to
+  another function as an argument.
   Callbacks can be denoted by the <type>callable</type> type declaration.
  </para>
```
Girgias:
```suggestion
<simpara>
  A callable is a reference to a function or method that is passed to
  another function as an argument.
  They are represented with the <type>callable</type> type declaration.
</simpara>
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
+ </informalexample>
 
- <para>
-  Some functions like <function>call_user_func</function> or
-  <function>usort</function> accept user-defined callback functions as a
-  parameter. Callback functions can not only be simple functions, but also
-  <type>object</type> methods, including static class methods.
- </para>
+ <simpara>
+  Some functions accept user-defined callback functions as a parameter, e.g.
```
Girgias: They don't care that they are user defined.
```suggestion
  Some functions accept callback functions as a parameter, e.g.
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
 
  <sect2 xml:id="language.types.callable.passing">
-  <title>Passing</title>
+  <title>Creation of callables</title>
+
+  <simpara>
+   A callable is a type that represents something that can be invoked.
+   Callables can be passed as arguments to functions or methods which
+   expect a callback parameter or they can be invoked directly. The callable
+   type cannot be used as a type declaration for class properties.
```
Girgias:
```suggestion
   expect a callback parameter or they can be invoked directly.
   The <type>callable</type> type cannot be used as a type declaration for class properties.
```
And maybe we even should explain why they cannot or recommend using `Closure` for properties.

language/types/callable.xml
Girgias: It will make it lowercase, but will also wrap it in a `<type>` tag to link to the page.
So I'm +-0 on this change.

Although please indent then content if it is going to span multiple lines, see review comment for invoke.

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
+   <listitem>
+    <simpara>String containing the name of a function or a method</simpara>
+   </listitem>
+   <listitem>
+    <simpara>Array containing a class name or an <type>object</type>
+    in index 0 and the method name in index 1</simpara>
+   </listitem>
+   <listitem>
+    <simpara>Object implementing the
+    <link linkend="object.invoke">__invoke()</link> magic method</simpara>
```
Girgias:
```suggestion
    <simpara>
     Object implementing the <link linkend="object.invoke">__invoke()</link> magic method
    </simpara>
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
+$new_numbers = array_map($double4, range(1, 5));
+print implode(' ', $new_numbers);
 
-  <para>
-   A PHP function is passed by either its name as a <type>string</type> or by
-   a <link linkend="functions.first_class_callable_syntax">first-class callable</link>.
+?>
+]]>
+   </programlisting>
+   &example.outputs;
```
Girgias: As you use FCC.
```suggestion
   &example.outputs.8.1;
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
+
+  <simpara>
+   A method of an instantiated <type>object</type> can be a callable
+   when provided as an array with the <type>object</type> at index 0 and
+   the method name at index 1.
+  </simpara>
+
+  <simpara>
+   The main difference between a <classname>Closure</classname> object and the
+   callable type is that a <classname>Closure</classname> object is
```
Girgias:
```suggestion
   <type>callable</type> type is that a <classname>Closure</classname> object is
```

language/types/callable.xml
```diff
@@ -107,55 +238,35 @@ class B extends A {
...
+// Type 7: Objects implementing __invoke can be used as callables
 class C {
     public function __invoke($name) {
-        echo 'Hello ', $name, PHP_EOL;
+        echo 'Hello ', $name;
     }
 }
 
 $c = new C();
 call_user_func($c, 'PHP!');
```
Girgias, line 256: Maybe?
```suggestion
$c('PHP!');
```

language/types/callable.xml
```diff
@@ -1,71 +1,199 @@
...
-
-  <para>
-   <example>
-    <title>
-     Callback function examples
-    </title>
-    <programlisting role="php">
+  <example>
+   <title>
+    Callback function examples
```
Girgias: Although I would prefer not encouraging the use of CUF{A}
```suggestion
    Calling various types of callables with <function>call_user_function</function>
```

language/types/callable.xml
kamil-tekiela: Yes, but not here. This is just a list of examples and it's not exhaustive. For example, `return` isn't there.

language/types/callable.xml
kamil-tekiela: I'm not sure. I'd have to replace it everywhere in this example otherwise it would lose cohesiveness. Then I should probably also rename the variable from `$c` to `$callable`. And we would also lose number 6. I think it's too much for this PR. Maybe we can think of replacing this example at a later date.

PR #4936, opened 2025-10-17
PDO Fetch Modes

### LINKED ISSUE #1425: PDO::FETCH_KEY_PAIR is undocumented on PDO::fetchAll
Without consulting the user comments or the PDO constants list, users don't know that FETCH_KEY_PAIR exists to read a key-value list into an associative array. The PDOStatement::fetchAll page refers to PDOStatement::fetch for a list of FETCH-Values, but FETCH_KEY_PAIR makes sense only for fetchAll. So you should add something like this

<para>
To return a key-value query as an associative array, use <constant>PDO::FETCH_KEY_PAIR</constant> as query mode. The
query must retrieve exacty two columns for this to work.
</para>

AllenJB:
Fixes #1425 (and numerous user comments on various pages)

The documentation for PDO fetch modes is currently distinctly lacking. There's some specific documentation and examples on various pages, but nothing decent "all in one place".

https://phpdelusions.net/pdo/fetch_modes has long been oft referenced by myself and by many others across chats and forums. If it ever goes missing for any reason we have nothing comparable in the official documentation.

This is my attempt to fix that. While my work is based on the above link at a high level, I have written this documentation "from scratch", performing my own tests to verify behavior, along with referencing php-src code and tests.

I have specifically written this documentation so that wherever the PDO::FETCH_* constants appear, you can click them and they take you right to the detailed docs.

There's a summary at the top of the page so people can still quickly see all the available fetch modes and jump to the detailed documentation.

Test scripts I wrote while writing this article: https://github.com/AllenJB/sandbox/tree/main/php-docs/pdo/fetch-modes

# Undefined Behaviors

There are some behaviors I've currently documented as specifically undefined. These are:

- Duplicated columns - which value is returned
- NAMED - order of duplicated values
- KEY_PAIR - which values are overwritten if first column is not unique
- UNIQUE - which values are overwritten

I couldn't see any obvious tests that were explicitly testing these behaviors in php-src. If you think I should document that current behavior as the define behavior, please let me know.

# FETCH_COLUMN | FETCH_GROUP

I've filed this as a bug: https://github.com/php/php-src/issues/20215

I have specifically not documented the behavior of FETCH_COLUMN | FETCH_GROUP because I think it's non-sensical and probably isn't supposed to work the way it currently does.

To see what I'm talking about, please refer to https://github.com/AllenJB/sandbox/blob/main/php-docs/pdo/fetch-modes/column/group.phpt (3v4l version: https://3v4l.org/2f7aV )

When you use FETCH_COLUMN, you can specify the additional column parameter to fetchAll() to tell PDO which column to retrieve values from. However, when you combine it with FETCH_GROUP, the column number is used for the grouping key and the values returned are always the first column.

This does not match my personal expected behavior, especially given that the column parameter can only be used with the COLUMN fetch mode - no other fetch mode allows you to influence how GROUP works in this way.

It also does not match the current documentation for PDOStatement::fetchAll()

You can see from the 3v4l link above that COLUMN | GROUP works as I expected in PHP <5.2.6.

I could find no tests in php-src for the combination of COLUMN | GROUP and specifying the column parameter.

The following is a timeline I put together for this:

* 2005-08-31: PDOStatement::fetchAll docs updated to include FETCH_GROUP, but the behavior described does not match the then actual behavior
* 2008-05-01: PHP 5.2.6 changes the behavior to somewhat match the docs (I believe the bug that triggered this is https://bugs.php.net/bug.php?id=44362 ) (if you read it one way, but this just ends up confusing because if you read it this way to column parameter has 2 different purposes at the same time)
* 2024-10-06: PDOStatement::fetchAll docs changed to describe the original < 5.2.6 behavior

I have no idea what to do about the documentation here. The current documentation for PDOStatement::fetchAll is wrong, but the current code behavior is non-sensical in my opinion.

### DIFF
--- a/reference/pdo/constants.fetch-modes.xml
+++ b/reference/pdo/constants.fetch-modes.xml
@@ -990,7 +990,7 @@ object(TestEntity)#5 (4) {
   <para>
    This fetch mode does not directly return a result, but binds values to
    variables specified with <function>PDOStatement::bindColumn</function>. The
-   called fetch method returns <literal>TRUE</literal>.
+   called fetch method returns &true;.
   </para>
   <note>
    <para>


kamil-tekiela:
Wow, this is a lot of work. Thank you. I don't know when I will have the time to review this. I would say that you shouldn't be concerned too much with documenting broken or undefined behaviour. We can make issues for these and revisit them later.

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+      <entry>
+       Array indexed by column number only.
+      </entry>
+     </row>
+     <row>
+      <entry>
+       <constant>PDO::FETCH_COLUMN</constant>
+      </entry>
+      <entry>
+       A single value / column.
```
kamil-tekiela:
```suggestion
       A single column.
```

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+      </entry>
+     </row>
+     <row>
+      <entry>
+       <constant>PDO::FETCH_KEY_PAIR</constant>
+      </entry>
+      <entry>
+       Key-Value pairs, indexed by the first column.
```
kamil-tekiela:
```suggestion
       Key-value pairs, indexed by the first column.
```

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+       An object of a specified class.
+      </entry>
+     </row>
+    </tbody>
+   </tgroup>
+  </informaltable>
+ </section>
+
+ <section xml:id="pdo.fetch-modes.class-flags" annotations="chunk:false">
+  <title><constant>PDO::FETCH_CLASS</constant> options</title>
```
kamil-tekiela:
```suggestion
  <title>PDO::FETCH_CLASS options</title>
```
It renders weird.

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+ </section>
+
+ <section xml:id="pdo.constants.fetch-func" annotations="chunk:false">
+  <title>PDO::FETCH_FUNC (<type>int</type>)</title>
+  <para>
+   Specify a function to create the returned value. This mode can only be used
+   with <function>PDOStatement::fetchAll</function>.
+  </para>
+  <para>
+   The function received the values as parameters. There's no way to retrieve
```
kamil-tekiela:
```suggestion
   The function receives the values as parameters. There's no way to retrieve
```

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+   the column a given value was associated with. You must make sure the column
```
kamil-tekiela:
```suggestion
   the column name a given value was associated with. You must make sure the column
```

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+    function is called.
+   </para>
+  </note>
+  <informalexample>
+   <programlisting role="php">
+<![CDATA[
+<?php
+function valueCreator($col1, $col2, $col3)
+{
+    $col2 = strtoupper($col2);
```
kamil-tekiela: I am not sure why you have this twice here.

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+
+ <section xml:id="pdo.constants.fetch-bound" annotations="chunk:false">
+  <title>PDO::FETCH_BOUND (<type>int</type>)</title>
+  <para>
+   This fetch mode can not be used with
+   <function>PDOStatement::fetchAll</function>.
+  </para>
+  <para>
+   This fetch mode does not directly return a result, but binds values to
+   variables specified with <function>PDOStatement::bindParam</function> or
```
kamil-tekiela: I don't understand how it can be used with `bindParam`

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+}
+]]>
+   </screen>
+  </informalexample>
+ </section>
+
+ <section xml:id="pdo.constants.fetch-bound" annotations="chunk:false">
+  <title>PDO::FETCH_BOUND (<type>int</type>)</title>
+  <para>
+   This fetch mode can not be used with
```
kamil-tekiela:
```suggestion
   This fetch mode cannot be used with
```

It appears 5 times on this page. See https://www.reddit.com/r/ENGLISH/comments/194zq37/why_is_it_cannot_instead_of_can_not/ for easy explanation.

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+  <programlisting role="php">
+<![CDATA[
+<?php
+$stmt = $db->query("SELECT userid, name, country, referred_by_userid FROM users");
+$result = $stmt->fetch(\PDO::FETCH_LAZY);
+
+print "ID: ". $result[0] ."\n";
+print "Name: {$result->name}\n";
+print "Country: " . $result['country'] ."\n";
+print "Does not exist: " . ($result->does_not_exist ?? "NULL") . "\n";
```
kamil-tekiela: The `??` hides what you were trying to show. Perhaps a ternary would be better?

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+   <constant>PDO::FETCH_GROUP</constant> returns lists of associative arrays,
+   indexed by a (non-unique) column. This fetch mode only works with
+   <function>PDOStatement::fetchAll</function>.
+  </para>
+  <para>
+   When combined with <constant>PDO::FETCH_UNIQUE</constant>, both modes will
+   use the same column, rendering the combination of these modes useless.
+  </para>
+  <para>
+   This fetch should be combined with one of
```
kamil-tekiela, line 1177: You say it should be combined but then your examples proves you wrong.

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+   for the <classname>PDOStatement</classname> will be used.
+  </para>
+  <para>
+   When used with a column that is known to be unique (such as record ID), this
+   mode provides the ability to quickly return results indexed by that value.
+  </para>
+  <note>
+   <para>
+    If the first column is not unique, values will be lost. Which value(s) are
+    lost / retained should be considered undefined.
```
kamil-tekiela:
```suggestion
    lost/retained should be considered undefined.
```
https://english.stackexchange.com/questions/2131/should-i-write-module-theme-or-module-theme

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1452 @@
...
+  <title>PDO::FETCH_KEY_PAIR (<type>int</type>)</title>
+  <para>
+   <constant>PDO::FETCH_KEY_PAIR</constant> returns pairs of values, indexed by
+   the first column. The results must contain only 2 columns. This fetch mode
+   only makes sense with <function>PDOStatement::fetchAll</function>.
+  </para>
+  <note>
+   <para>
+    If the first column is not unique, values will be lost. Which value(s) are
+    lost / retained should be considered undefined.
```
kamil-tekiela:
```suggestion
    lost/retained should be considered undefined.
```
https://english.stackexchange.com/questions/2131/should-i-write-module-theme-or-module-theme

kamil-tekiela (changes requested):
It looks great. I think this is going to be a huge improvement. The only thing that feels weird is that the fetch modes are now on a separate page behind a link, but I have no idea how to present it better. Maybe "Predefined Constants" should be split into 3 pages with "cursors" and "other constants" getting their separate pages too?

reference/pdo/constants.fetch-modes.xml
AllenJB: I've changed this to use `var_export` and added a comment

reference/pdo/constants.fetch-modes.xml
AllenJB: Why I have what twice here? I'm not sure what you're pointing to and I can't see anything that's been duplicated here.

AllenJB:
> It looks great. I think this is going to be a huge improvement. The only thing that feels weird is that the fetch modes are now on a separate page behind a link, but I have no idea how to present it better. Maybe "Predefined Constants" should be split into 3 pages with "cursors" and "other constants" getting their separate pages too?

There are precedents for constants sections with separate pages - see openssl, posix and radius. (curl also has them split across multiple files in the docs repo but not in the published html docs).

I considered having separate pages for "cursors" and "other" constants here, but I didn't think there is currently enough content to justify it (and was also consciously avoiding making "unrelated changes" as much as possible)

reference/pdo/constants.fetch-modes.xml
kamil-tekiela: `strtoupper($col2)` twice. Here and two lines below

reference/pdo/constants.fetch-modes.xml
```diff
@@ -0,0 +1,1456 @@
...
+ <section xml:id="pdo.constants.fetch-bound" annotations="chunk:false">
+  <title>PDO::FETCH_BOUND (<type>int</type>)</title>
+  <para>
+   This fetch mode cannot be used with
+   <function>PDOStatement::fetchAll</function>.
+  </para>
+  <para>
+   This fetch mode does not directly return a result, but binds values to
+   variables specified with <function>PDOStatement::bindColumn</function>. The
+   called fetch method returns <literal>TRUE</literal>.
```
kamil-tekiela:
```suggestion
   called fetch method returns &true;.
```

Wouldn't this be better?

reference/pdo/constants.fetch-modes.xml
AllenJB: Fixed

PR #4943, opened 2025-10-21
Make object a parameter

mmalferov:


reference/spl/splobjectstorage/attach.xml
```diff
@@ -31,15 +31,15 @@
      <term><parameter>object</parameter></term>
      <listitem>
       <para>
-       The <type>object</type> to add.
+       The <parameter>object</parameter> to add.
```
kamil-tekiela, line 34: This is wrong. You have now created recursion within this page. I think it's ok to leave the page as it is now.

kamil-tekiela:
Thank you, but this doesn't improve anything.

PR #4945, opened 2025-10-22
stream_context_set_option(): fix incorrect parameter name

jrfnl:
I vaguely remember these signatures are normally updated automatically from the stub files ? Not sure what happened here, but this one is clearly wrong.

Page: https://www.php.net/manual/en/function.stream-context-set-option.php

Signature as per the stub: https://github.com/php/php-src/blob/40a42cffd81b7f01fed25b3f7d02cb4d2ead700d/ext/standard/basic_functions.stub.php#L3384C78-L3384C103


TimWolla:
> I vaguely remember these signatures are normally updated automatically from the stub files

They *can* be updated automatically, but it's still a manual process to start the automated update.

PR #4946, opened 2025-10-23
parse_ini_*(): Add warning that the functions are unsafe with untrusted inputs

TimWolla:


nielsdos (commented):
Not sure about warning vs caution tag but text lgtm.
Please change the para tags to simpara.

Girgias (approved):
Warning MSTM

PR #4948, opened 2025-10-23
curl-escape.xml Amend comment

mmalferov:
The `curl_close()` call has been removed — the fix was there, just not the comment :)


TimWolla (approved):
Good catch. Thank you!

PR #4953, opened 2025-10-24
Simplifiying the "A simple tutorial" page

CViniciusSDias:
Simplifiying the "A simple tutorial" page  by removing the Apache and MySQL pre-requisites and using php's builtin server

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -51,7 +51,7 @@ php -S localhost:8000
      <simpara>
       Use your browser to access the file with your web server's URL, ending
       with the <literal>/hello.php</literal> file reference.
-      According to the previous command executed, the URL will be 
+      According to the previous command executed, the URL will be
       <literal>http://localhost:8000/hello.php</literal>.
       If everything is configured correctly, this file will be parsed by PHP
       and you will see the "Hello World!" output displayed in your browser.


chapters/tutorial.xml
```diff
@@ -73,17 +36,25 @@
...
-?>
+]]>
+     </programlisting>
+     <simpara>
+      Using your terminal, navigate to the directory containing this file and
+      start a development server with the following command:
+     </simpara>
+     <programlisting role="shell">
+<![CDATA[
+$ php -S localhost:8000
```
TimWolla:
```suggestion
php -S localhost:8000
```

chapters/tutorial.xml
```diff
@@ -206,7 +164,7 @@ echo "Hello World!";
      <info><title>Get system information from PHP</title></info>
      <programlisting role="php">
 <![CDATA[
-<?php phpinfo(); ?>
+<?php phpinfo();
```
TimWolla:
```suggestion
<?php

phpinfo();
```

chapters/tutorial.xml
```diff
@@ -16,52 +16,15 @@
...
-     xlink:href="&url.apt-get;">apt-get</link> to find packages for Debian.
-   </para>
-  </section>
-
   <section xml:id="tutorial.firstpage">
    <info><title>Your first PHP-enabled page</title></info>
+   <simpara>
+    This tutorial assumes PHP is already installed.
+    If not, you can download and install it
+    <link xlink:href="&url.php.downloads;">here</link>.
```
TimWolla: Please use a more useful link text than just "here". Perhaps:


```suggestion
    This tutorial assumes PHP is already installed.
    Installation instructions can be found on the
    <link xlink:href="&url.php.downloads;">download page</link>.
```

chapters/tutorial.xml
```diff
@@ -73,17 +36,25 @@
...
-      with the <literal>/hello.php</literal> file reference. When developing locally this
-      URL will be something like <literal>http://localhost/hello.php</literal>
-      or <literal>http://127.0.0.1/hello.php</literal> but this depends on the
-      web server's configuration. If everything is configured correctly, this
-      file will be parsed by PHP and you will see the "Hello World" output displayed
-      in your browser.
+      with the <literal>/hello.php</literal> file reference.
+      According to the previous command executed, the URL will be 
+      <literal>http://localhost:8000/hello.php</literal>
+      or <literal>http://127.0.0.1:8000/hello.php</literal>.
```
TimWolla: 127.0.0.1 might not necessarily be bound for IPv6 hosts. Just `localhost`, matching the actual command should be sufficient.

chapters/tutorial.xml
```diff
@@ -73,17 +36,25 @@
 
 echo "Hello World!";
 
-?>
```
TimWolla, line 76: I'm not sure how I feel about removing the closing tag here. All the following pages in the tutorial still have it (and the reference documentation does as well) - the HTML example further below necessarily also has it. I feel leaving the closing tag out at this point is doing more harm than it helps.

TimWolla (changes requested):
Thank you. I have some comments, but this definitely looks like an improvement to me.

chapters/tutorial.xml
CViniciusSDias: Makes sense. I added the closing tags in the othe example too.

chapters/tutorial.xml
```diff
@@ -74,16 +37,24 @@
...
+     </programlisting>
+     <simpara>
+      Using your terminal, navigate to the directory containing this file and
+      start a development server with the following command:
+     </simpara>
+     <programlisting role="shell">
+<![CDATA[
+php -S localhost:8000
 ]]>
      </programlisting>
```
Girgias, line 50: Does the "example" tag+title even provide any benefit?

Maybe just moving all of it into the normal flow is better? But I don't have strong opinions about this and the new text is an improvement over the current one.

chapters/tutorial.xml
derickr: I agree, but it *is* in the documentation guidelines to leave them out: https://doc.php.net/guide/cs-for-examples.md (even though the examples on that page have them)

kamil-tekiela: Can you quote which part talks about it? I cannot find it.

PR #4960, opened 2025-10-29
Fix misleading wording

ndossche:
Since PHP8 this is no longer a warning but an actual exception. Throw was never really right in the first place however, as warnings are _emitted_, not _thrown_.

### DIFF
--- a/reference/array/functions/array-walk.xml
+++ b/reference/array/functions/array-walk.xml
@@ -56,7 +56,7 @@
       <note>
         <para>
           Many internal functions (for example <function>strtolower</function>)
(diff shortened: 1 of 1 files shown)


reference/array/functions/array-walk.xml
```diff
@@ -56,7 +56,7 @@
       <note>
         <para>
           Many internal functions (for example <function>strtolower</function>)
-          will throw a warning if more than the expected number of argument
+          will throw if more than the expected number of argument
```
TimWolla:
```suggestion
          will throw if more than the expected number of arguments
```

PR #4962, opened 2025-10-30
Add a Docker Compose setup for building and serving PHP documentation locally

lacatoire:
### Description

This PR adds a simple **Docker Compose workflow** to build and preview the PHP manual locally.
It provides a lightweight, cross-platform alternative to the existing make setup and works on Linux, macOS, and Windows (including PowerShell).

The goal is to make contributing to the documentation easier for new and existing contributors by removing installation steps and dependency issues.

### Highlights

* Uses [PhD](https://github.com/php/phd) inside a clean Docker environment
* Default language is English (en), but others can be built easily:
  * Linux/macOS: LANGUAGE=fr docker compose up --build
  * PowerShell: $env:LANGUAGE="fr"; docker compose up --build
* Output format (e.g. xhtml, php-chunked-xhtml) can be changed in the compose.yaml file
* Automatically handles CRLF normalization on Windows

### Technical notes

- Based on php:8.2-cli-alpine for stability and minimal dependencies.
- PHP 8.3 and 8.4 were skipped to avoid compatibility issues with the PhD toolchain.
- **FrankenPHP** was considered for serving the output but would add unnecessary complexity the built-in PHP server is enough here.

Does not modify the existing make build process this is an optional, self-contained alternative.

### Impact

This setup should lower the barrier to entry for contributors and make it easier to test translations and improvements locally, without manual setup.

### DIFF
--- a/.docker/entrypoint.sh
+++ b/.docker/entrypoint.sh
@@ -43,6 +43,14 @@ if [ -f "$MANUAL_PATH" ]; then
   rm -f "$MANUAL_PATH" || true
 fi
 
+if [ -d "/var/www/doc-en" ]; then
+  # Make sure that doc-en is mounted
+  if [ ! -d "/var/www/en" ] && [ ! -L "/var/www/en" ]; then
+    ln -s /var/www/doc-en /var/www/en
+    echo "🩹 Created symlink for English: /var/www/en -> /var/www/doc-en"
+  fi
+fi
+
 echo "ℹ️ Generating .manual.xml for $LANGUAGE..."
 cd /var/www/doc-base
 php configure.php \


README.md
```diff
@@ -31,6 +31,48 @@ already exist.
...
+### 🐋 Building with Docker Compose (alternative modern setup)
+
+A `compose.yaml` file is available for contributors who prefer a simple,
+cross-platform Docker workflow instead of `make`.
+
+It builds and serves the PHP manual locally in any translation
+(English by default, or another language via `LANGUAGE`).
+
+#### Prerequisites
+- Docker ≥ 24 and Docker Compose ([install guide](https://docs.docker.com/get-docker/))
```
afilina, line 44:
```suggestion
#### Prerequisites

- Docker ≥ 24 and Docker Compose ([install guide](https://docs.docker.com/get-docker/))
```

Nitpick for document consistency.

README.md
```diff
@@ -31,6 +31,48 @@ already exist.
...
+- Local clone structure:
+```
+php-doc/
+├─ doc-base/ ← common build files, scripts, entities, configure.php
+├─ doc-en/ ← English manual (contains this Docker setup)
+└─ doc-fr/ ← French manual or you official language repository (doc-es, doc-ja...)
+```
+
+#### Usage
+From inside `php-doc/doc-en/`:
```
afilina, line 55:
```suggestion
#### Usage

From inside `php-doc/doc-en/`:
```

Nitpick for document consistency.

.docker/entrypoint.sh
```diff
@@ -0,0 +1,82 @@
...
+OUTDIR=""
+for d in php-chunked-xhtml xhtml phpweb php; do
+  if [ -d "/var/www/$LANGUAGE/output/$d" ]; then
+    OUTDIR="/var/www/$LANGUAGE/output/$d"
+    break
+  fi
+done
+
+if [ -d "$OUTDIR" ]; then
+  echo "✅ Build complete: $OUTDIR"
```
afilina, line 104:
```suggestion
  echo "✅ Build complete: $OUTDIR"
  echo "View the documentation on http://localhost:8000"
```
Once the build is complete, print the URL with next step. Then the user wouldn't need to go back to the README.

compose.yaml
```diff
@@ -0,0 +1,22 @@
...
+  phpdoc:
+    build:
+      dockerfile: .docker/Dockerfile
+    container_name: phpdoc-builder
+    working_dir: /var/www
+    user: "0:0"
+    volumes:
+      - ../doc-${LANGUAGE:-en}:/var/www/${LANGUAGE:-en}
+      - ../doc-en:/var/www/en
+      - ../doc-base:/var/www/doc-base
```
afilina, line 11: When the prerequisites (`doc-base`, `doc-fr`) don't exist, it would make sense to warn the user instead of the default behavior which mounts empty directories.

Otherwise, the subsequent error message `Could not open input file: configure.php` provides no context and is confusing. User must then fix by deleting empty folders, cloning prerequisites, and running docker with --force-recreate.

README.md
```diff
@@ -31,6 +31,48 @@ already exist.
...
+```bash
+$env:LANGUAGE="fr"; docker compose up --build
+
+Remove-Item Env:LANGUAGE # To remove the variable:
+```
+The process will:
+
+1. Build a minimal PHP 8.2 environment with PhD
+2. Generate `.manual.xml` and render the manual in HTML (php-chunked-xhtml).
+3. Start a local PHP web server on http://localhost:8000
```
afilina:
```suggestion
1. Build a minimal PHP 8.2 environment with PhD.
2. Generate `.manual.xml` and render the manual in HTML (php-chunked-xhtml).
3. Start a local PHP web server on http://localhost:8000 where you can view the documentation.
```

README.md
```diff
@@ -31,6 +31,48 @@ already exist.
...
+To build another translation:
+**Linux / macOS**
+```
+LANGUAGE=fr docker compose up --build
+```
+**Windows PowerShell**
+```bash
+$env:LANGUAGE="fr"; docker compose up --build
+
+Remove-Item Env:LANGUAGE # To remove the variable:
```
afilina:
```suggestion

Remove-Item Env:LANGUAGE # Revert back to English
```

README.md
```diff
@@ -31,6 +31,48 @@ already exist.
...
+(English by default, or another language via `LANGUAGE`).
+
+#### Prerequisites
+- Docker ≥ 24 and Docker Compose ([install guide](https://docs.docker.com/get-docker/))
+- Local clone structure:
+```
+php-doc/
+├─ doc-base/ ← common build files, scripts, entities, configure.php
+├─ doc-en/ ← English manual (contains this Docker setup)
+└─ doc-fr/ ← French manual or you official language repository (doc-es, doc-ja...)
```
afilina:
```suggestion
└─ doc-fr/ ← French manual or your official language repository (doc-es, doc-ja...)
```

.docker/Dockerfile
```diff
@@ -1,19 +1,49 @@
...
+# Grant full permissions to /var/www (useful on Windows/WSL2)
+RUN mkdir -p /var/www/output && chmod -R 777 /var/www
+
+# Entry script
+COPY .docker/entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN sed -i 's/\r$//' /usr/local/bin/entrypoint.sh && chmod +x /usr/local/bin/entrypoint.sh
+
+EXPOSE 8000
+RUN chmod -R 777 /var/www
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
```
afilina:
```suggestion
```

CI complaint: https://github.com/php/doc-en/actions/runs/18949333125/job/54494031296?pr=4962

compose.yaml
lacatoire: HI @afilina thanks for your review, i've added some explanation in the entrypoint.sh
https://github.com/php/doc-en/pull/4962/files?new_files_changed=true#diff-d464f2255f2f24bca648f9c24fc26ecba642be2efd7afc78fb080232dee04524R17

.docker/entrypoint.sh
```diff
@@ -0,0 +1,103 @@
...
+MANUAL_PATH="/var/www/$LANGUAGE/.manual.xml"
+
+# 🧩 Step 1. Verify prerequisites (avoid silent empty mounts)
+
+REQUIRED_DIRS="/var/www/doc-base /var/www/$LANGUAGE"
+
+echo "🔍 Checking required documentation directories..."
+for DIR in $REQUIRED_DIRS; do
+  if [ ! -d "$DIR" ]; then
+    echo "❌ Missing required directory: $DIR"
```
afilina: I wasn't able to hit this line even when I had no directories. It's checking them after the volumes are mounted, so there's always a folder, just empty.

I always get "✅ All prerequisites found." and then it crashes later.

lacatoire: Good catch @afilina !
Docker creates empty folders on volume mount, so the -d check always passed.
I changed it to check for real files (configure.php, entities/) instead.

.docker/Dockerfile
```diff
@@ -1,19 +1,48 @@
...
 
-WORKDIR /var/www
+ENV LANG=C.UTF-8 \
+    LC_ALL=C.UTF-8 \
+    DOC_DIR=/var/www \
+    PHD_DIR=/opt/phd
 
-ADD https://api.github.com/repos/php/phd/git/refs/heads/master version-phd.json
-ADD https://api.github.com/repos/php/doc-base/git/refs/heads/master version-doc-base.json
+# 🔧 Dependancies to PhD
```
afilina:
```suggestion
# 🔧 PhD dependencies
```

.docker/Dockerfile
```diff
@@ -1,19 +1,48 @@
...
+    libxslt-dev \
+    gettext \
+    autoconf \
+    g++ \
+    pkgconfig \
+ && docker-php-ext-install intl xsl
 
-RUN git clone --depth 1 https://github.com/php/phd.git &&  \
-    git clone --depth 1 https://github.com/php/doc-base.git
+# 📦 Installer Composer
```
afilina:
```suggestion
# 📦 Install Composer
```

.docker/entrypoint.sh
```diff
@@ -0,0 +1,109 @@
...
+if [ ! -f /var/www/doc-base/configure.php ]; then
+  echo "❌ doc-base is missing or incomplete!"
+  echo "👉 Please clone it before running Docker:"
+  echo "   git clone https://github.com/php/doc-base ../doc-base"
+  echo ""
+  exit 1
+fi
+
+# Check language documentation presence (must contain entities/)
+if [ ! -d /var/www/$LANGUAGE/entities ]; then
```
afilina: When a user first clones the translation, there is no `entities` directory there. Was this test meant to be performed at a later time, and we only need to check for `/var/www/$LANGUAGE`?

Otherwise it will just constantly stop here. I would suggest testing the workflow in a fresh directory structure.

.docker/entrypoint.sh
```diff
@@ -0,0 +1,109 @@
...
+  echo "❌ doc-${LANGUAGE} is missing or incomplete!"
+  echo "👉 Please clone it before running Docker:"
+  echo "   git clone https://github.com/php/doc-${LANGUAGE} ../doc-${LANGUAGE}"
+  echo ""
+  echo "💡 Tip: remove any empty folders and recreate containers with:"
+  echo "   docker compose down --volumes && docker compose up --force-recreate"
```
afilina: Shouldn't this command include the `LANGUAGE` env var as well, otherwise it will default to English again? The only problem is that syntax changes based on platform, so perhaps to avoid overengineering, we can simply mention to add `--force-recreate` when running the next docker command, instead of giving the full command? Just an idea.

.docker/entrypoint.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# Normalize Windows CRLF endings and adjust doc-base paths
+echo "🩹 Patching entity paths to use doc-base…"
+tr -d '\r' < "$MANUAL_PATH" > "$MANUAL_PATH.clean" && mv "$MANUAL_PATH.clean" "$MANUAL_PATH"
+sed -i "s#/var/www/$LANGUAGE/entities/#/var/www/doc-base/entities/#g" "$MANUAL_PATH"
+sed -i "s#/var/www/$LANGUAGE/version.xml#/var/www/doc-base/version.xml#g" "$MANUAL_PATH"
+sed -i "s#/var/www/$LANGUAGE/sources.xml#/var/www/doc-base/sources.xml#g" "$MANUAL_PATH"
+head -n 25 "$MANUAL_PATH"
+
+# 🧩 Inclure le package PHP pour PhD
```
afilina:
```suggestion
# 🧩 Include the PHP package for PhD
```

afilina (changes requested):
It still says "doc-fr is missing or incomplete!" even after I clone it. Please test the full workflow (starting at missing doc-fr and doc-base) before submitting for review again. This will allow you to catch and fix errors early.

Also if you could add changes as a separate commit for each review iteration, as squashing means I have to re-read everything each time, and I've done this multiple times already. Thanks.

afilina:
I'm on my way to a conference overseas, so I won't be able to give it another review until Sunday.

PR #4964, opened 2025-10-30
mb-strpos.xml Use parameter names instead of types, fix typos and XML syntax

mmalferov:


mmalferov:
And I'm not sure whether to leave the sentence "In contrast with strpos(), numeric values are not applied as the ordinary value of a character" in the description of the `needle` parameter.

reference/strings/functions/strripos.xml
```diff
@@ -141,25 +141,27 @@
...
 <?php
+
 $haystack = 'ababcd';
 $needle   = 'aB';
 
 $pos      = strripos($haystack, $needle);
 
 if ($pos === false) {
-    echo "Sorry, we did not find ($needle) in ($haystack)";
+    echo "Sorry, we did not find `$needle` in `$haystack`";
```
kamil-tekiela, line 151: This doesn't match the output and I am not clear why you changed it. https://3v4l.org/nll6U

mmalferov: Oh, you're right, I deleted the extra spaces in the CDATA block, but I didn't correct the output, I didn't replace the brackets that the eye stumbles over with a backtick. Fixed it now

kamil-tekiela (approved):

PR #4983, opened 2025-11-04
Fix typo

Marqitos:
'resoluition' to 'resolution' in `reference/wkhtmltox/wkhtmltox/pdf/converter/construct.xml`


sy-records (approved):
Thanks!

PR #4989, opened 2025-11-10
Rework documentation for dir functions

Girgias:
Include 8.5 deprecation for passing null to use the last opened directory.

### DIFF
--- a/reference/dir/functions/opendir.xml
+++ b/reference/dir/functions/opendir.xml
@@ -104,7 +104,6 @@
 <?php
 
 if ($handle = opendir('/path/to/files')) {
-    echo "Directory handle: $handle\n";
     echo "Entries:\n";
 
     /* Correctly handling directory entries that may be considered falsy */
@@ -123,11 +122,12 @@ if ($handle = opendir('/path/to/files')) {
    &example.outputs.similar;
    <screen>
 <![CDATA[
-filename: . : filetype: dir
-filename: .. : filetype: dir
-filename: apache : filetype: dir
-filename: cgi : filetype: dir
-filename: cli : filetype: dir
+Entries:
+base
+en
+fr
+output.md
+test.php
 ]]>
    </screen>
   </example>


reference/dir/functions/opendir.xml
```diff
@@ -89,49 +85,65 @@
...
+   </title>
+   <simpara>
+    Because file and directory names can be strings that PHP considers "falsy"
+    (e.g. a directory named <literal>"0"</literal>) and
+    <function>readdir</function> returns &false; when it has read all entries
+    in a directory one needs to use the <literal>===</literal>
+    <link linkend="language.operators.comparison">comparison operator</link>
+    to properly distinguish between a directory entry whose name is "falsy"
+    and having read all entries of the directory.
+   </simpara>
```
kamil-tekiela, line 101: Shouldn't this need to be a warning or a note?

reference/dir/functions/opendir.xml
```diff
@@ -89,49 +85,65 @@
...
+   </programlisting>
+   &example.outputs.similar;
+   <screen>
 <![CDATA[
 filename: . : filetype: dir
 filename: .. : filetype: dir
 filename: apache : filetype: dir
 filename: cgi : filetype: dir
 filename: cli : filetype: dir
 ]]>
```
kamil-tekiela, line 131: The output doesn't match the code.

Girgias: We can't really have any output because it might be any filesystem, that's why we use the `&example.outputs.similar;` entity rather than `&example.outputs;`.

reference/dir/functions/opendir.xml
Girgias: I guess so, maybe could even be a caution.

reference/dir/functions/opendir.xml
kamil-tekiela: I don't mean the file names. I meant the overal output structure is from the old example. You copied the other example without the copying the output also.

reference/dir/functions/readdir.xml
```diff
@@ -71,77 +74,23 @@
...
-
-    /* This is the WRONG way to loop over the directory. */
-    while ($entry = readdir($handle)) {
-        echo "$entry\n";
-    }
-
-    closedir($handle);
-}
-?>
-]]>
```
kamil-tekiela, line 107: You copied it from here and this didn't have the output section.

kamil-tekiela (approved):

PR #4998, opened 2025-11-15
Fix typo add ref `warn.deprecated.feature-8-1-0`

mmalferov:


reference/pdo/constants.fetch-modes.xml
```diff
@@ -867,8 +867,7 @@ object(TestEntity)#3 (4) {
   <title>PDO::FETCH_SERIALIZE (<type>int</type>)</title>
   <warning>
    <simpara>
```
kamil-tekiela: You need to remove these too

mmalferov: You're right, my mistake. Just a minute…

kamil-tekiela (approved):

PR #5000, opened 2025-11-15
constants.fetch-modes.xml Fix output

mmalferov:


sy-records (approved):

AllenJB:
I've submitted #5001 as an alternative fix for this. As this additional output serves no purpose in this example, I think it would be better to remove the print line from the code, rather than correcting the output.

Thank you for pointing out this issue.

PR #5014, opened 2025-11-23
gmstrftime.xml Fix example to match description and demonstrate local…

mmalferov:
…e handling

The description says we should see different times for local vs GMT, but both lines print the same. Also, using English locale doesn't really show what `gmstrftime` can do

### DIFF
--- a/reference/datetime/functions/gmstrftime.xml
+++ b/reference/datetime/functions/gmstrftime.xml
@@ -115,7 +115,7 @@ diciembre 31 1998 20:00:00
 Deprecated: Function gmstrftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 7
 enero 01 1999 01:00:00
 ]]>
-   </screen
+   </screen>
    </example>
   </para>
  </refsect1>
--- a/reference/datetime/functions/gmstrftime.xml
+++ b/reference/datetime/functions/gmstrftime.xml
@@ -110,7 +110,7 @@ echo gmstrftime("%B %d %Y %H:%M:%S", mktime(20, 0, 0, 12, 31, 98));
    <screen>
 <![CDATA[
 Deprecated: Function strftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 6
-diciembre 31 1998 20:00:00
+Dec 31 1998 20:00:00
 
 Deprecated: Function gmstrftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 7
 enero 01 1999 01:00:00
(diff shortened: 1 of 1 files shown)


reference/datetime/functions/gmstrftime.xml
```diff
@@ -99,11 +99,23 @@
...
+   &example.outputs;
+   <screen>
+<![CDATA[
+Deprecated: Function strftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 6
+diciembre 31 1998 20:00:00
+
+Deprecated: Function gmstrftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 7
+enero 01 1999 01:00:00
+]]>
+   </screen>
```
Girgias, line 118: While at it, maybe remove the useless `<para>` tag around the example tag and reindent everything properly?

reference/datetime/functions/gmstrftime.xml
```diff
@@ -99,11 +99,23 @@
...
+
+echo strftime("%B %d %Y %H:%M:%S",   mktime(20, 0, 0, 12, 31, 98)) . "\n";
+echo gmstrftime("%B %d %Y %H:%M:%S", mktime(20, 0, 0, 12, 31, 98));
 ]]>
     </programlisting>
+   &example.outputs;
+   <screen>
+<![CDATA[
+Deprecated: Function strftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 6
+diciembre 31 1998 20:00:00
```
lacatoire:
```suggestion
Dec 31 1998 20:00:00
```

reference/datetime/functions/gmstrftime.xml
```diff
@@ -99,11 +99,23 @@
...
+
+Deprecated: Function gmstrftime() is deprecated since 8.1, use IntlDateFormatter::format() instead in script on line 7
+enero 01 1999 01:00:00
```
lacatoire, line 116:
```suggestion
Jan 01 1999 01:00:00
```

reference/datetime/functions/gmstrftime.xml
mmalferov: Oh, I wish I could remember the rules for proper indentation :) I deleted the extra `<para>` tag and indented and tagged it by analogy with another file ))

reference/datetime/functions/gmstrftime.xml
jordikroon: The initial change was correct as `LC_TIME` is set to `es_ES.UTF-8`. Changing to `Jan` end `Dec` is incorrect. Could you undo these changes?

Other than that, this LGTM.

reference/datetime/functions/gmstrftime.xml
jordikroon: Same here. Should be `enero`.

jordikroon (approved):

PR #5022, opened 2025-11-27
Rewrite PDOStatement::fetch() docs

Girgias:
Remove personalization
Add constant tags
Reorder fetch modes in a more logical way
Update an example and make it runnable
Remove useless wrapping `<para>`
Use `<simpara>` when possible

Now I do wonder if it makes sense to detail this all here or if we should refer back to https://www.php.net/manual/en/pdo.constants.fetch-modes.php instead?

### DIFF
--- a/reference/pdo/pdostatement/fetch.xml
+++ b/reference/pdo/pdostatement/fetch.xml
@@ -11,141 +11,172 @@
   &reftitle.description;
   <methodsynopsis role="PDOStatement">
    <modifier>public</modifier> <type>mixed</type><methodname>PDOStatement::fetch</methodname>
-   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer>PDO::FETCH_DEFAULT</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter>cursorOrientation</parameter><initializer>PDO::FETCH_ORI_NEXT</initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>mode</parameter><initializer><constant>PDO::FETCH_DEFAULT</constant></initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>cursorOrientation</parameter><initializer><constant>PDO::FETCH_ORI_NEXT</constant></initializer></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>cursorOffset</parameter><initializer>0</initializer></methodparam>
   </methodsynopsis>
 
-  <para>
-   Fetches a row from a result set associated with a PDOStatement object. The
-   <parameter>mode</parameter> parameter determines how PDO returns
-   the row.
-  </para>
+  <simpara>
+   Fetches a row from a result set associated with a
+   <classname>PDOStatement</classname> instance.
+   The<parameter>mode</parameter> parameter determines how PDO returns the row.
+  </simpara>
  </refsect1>
 
  <refsect1 role="parameters">
   &reftitle.parameters;
-  <para>
-   <variablelist>
-    <varlistentry>
-     <term><parameter>mode</parameter></term>
-     <listitem>
-      <para>
-       Controls how the next row will be returned to the caller. This value
-       must be one of the <literal>PDO::FETCH_*</literal> constants,
-       defaulting to value of <literal>PDO::ATTR_DEFAULT_FETCH_MODE</literal>
-       (which defaults to <literal>PDO::FETCH_BOTH</literal>).
-       <itemizedlist>
-        <listitem><para>
-         <literal>PDO::FETCH_ASSOC</literal>: returns an array indexed by column
-         name as returned in your result set
-        </para></listitem>
-        <listitem><para>
-         <literal>PDO::FETCH_BOTH</literal> (default): returns an array indexed by
-         both column name and 0-indexed column number as returned in your
-         result set
-        </para></listitem>
-        <listitem><para>
-         <literal>PDO::FETCH_BOUND</literal>: returns &true; and assigns the
-         values of the columns in your result set to the PHP variables to which
-         they were bound with the <methodname>PDOStatement::bindColumn</methodname>
-         method
-        </para></listitem>
-        <listitem><para>
-         <literal>PDO::FETCH_CLASS</literal>: returns a new instance of the
-         requested class. The object is initialized by mapping the columns of
-         the result set to properties in the class. This occurs before the
-         constructor is called, allowing properties to be populated regardless
-         of their visibility or whether they are marked as <literal>readonly</literal>.
-         If a property does not exist in the class, the magic <link linkend="object.set">__set()</link>
+  <variablelist>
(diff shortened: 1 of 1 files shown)


reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
 
-  <para>
-   Fetches a row from a result set associated with a PDOStatement object. The
-   <parameter>mode</parameter> parameter determines how PDO returns
-   the row.
-  </para>
+  <simpara>
+   Fetches a row from a result set associated with a
+   <classname>PDOStatement</classname> instance.
+   The<parameter>mode</parameter> parameter determines how PDO returns the row.
```
kamil-tekiela:
```suggestion
   The <parameter>mode</parameter> parameter determines how PDO returns the row.
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
+         column name as returned in the result set.
+        </simpara>
+       </listitem>
+       <listitem>
+        <simpara>
+         <constant>PDO::FETCH_NAMED</constant>: returns an array with the same
+         form as <constant>PDO::FETCH_ASSOC</constant>,
+         except that if there are multiple columns with the same name,
+         the value referred to by that key will be an array of all the values
+         in the row that had that column name
```
kamil-tekiela:
```suggestion
         in the row that had that column name.
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
+         <classname>stdClass</classname> with property names that correspond
+         to the column names returned in the result set.
+        </simpara>
+       </listitem>
+       <listitem>
+        <simpara>
+         <constant>PDO::FETCH_CLASS</constant>: returns a new instance of the
+         requested class.
+         If <constant>PDO::FETCH_CLASSTYPE</constant> is included,
+         the class to instantiated is determined by the value of the
```
kamil-tekiela:
```suggestion
         the class to instantiate is determined by the value of the
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
+         property will be created.
+        </simpara>
+        <simpara>
+         It is possible to change this behaviour by using the
+         <constant>PDO::FETCH_PROPS_LATE</constant> flag to call the
+         constructor <emphasis>before</emphasis> the properties are populated.
+        </simpara>
+        <caution>
+         <simpara>
+          The properties that will be populated are <emphasis>not</emphasis>
```
kamil-tekiela, line 112: Do you mean the column names or the values that were assigned before calling the constructor?

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
+          passed as parameters to the constructor, regardless of the flag
+          <constant>PDO::FETCH_PROPS_LATE</constant> being used or not.
+         </simpara>
+        </caution>
+       </listitem>
+       <listitem>
+        <simpara>
+         <constant>PDO::FETCH_INTO</constant>: updates an existing instance
          of the requested class, mapping the columns of the result set to
          named properties in the class
```
kamil-tekiela:
```suggestion
         named properties in the class.
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
+    <term><parameter>cursorOffset</parameter></term>
+    <listitem>
+     <simpara>
+      If the value of the <parameter>cursorOrientation</parameter> parameter
+      is <constant>PDO::FETCH_ORI_ABS</constant>, this value specifies the
+      absolute number of the row in the result set that shall be fetched.
+     </simpara>
+     <simpara>
+      If the value of the <parameter>cursorOrientation</parameter> parameter
+      is <constant>PDO::FETCH_ORI_REL</constant>,this value specifies the
```
kamil-tekiela:
```suggestion
      is <constant>PDO::FETCH_ORI_REL</constant>, this value specifies the
```

reference/pdo/pdostatement/fetch.xml
```diff
@@ -155,74 +186,95 @@
...
-  <para>
-   <example><title>Fetching rows using different fetch styles</title>
-    <programlisting role="php">
+  <example>
+   <title>Fetching rows using different fetch styles</title>
+   <programlisting role="php">
 <![CDATA[
 <?php
-$sth = $dbh->prepare("SELECT name, colour FROM fruit");
+$db = new PDO('sqlite::memory');
```
kamil-tekiela:
```suggestion
$db = new PDO('sqlite::memory:');
```

kamil-tekiela (changes requested):

reference/pdo/pdostatement/fetch.xml
Girgias: I mean the values, but the column names are not passed either to the constructor AFAIK?

reference/pdo/pdostatement/fetch.xml
```diff
@@ -11,141 +11,172 @@
...
+         to the column names returned in the result set.
+        </simpara>
+       </listitem>
+       <listitem>
+        <simpara>
+         <constant>PDO::FETCH_CLASS</constant>: returns a new instance of the
+         requested class.
+         If <constant>PDO::FETCH_CLASSTYPE</constant> is included,
+         the class to instantiate is determined by the value of the
+         first column.
```
kamil-tekiela, line 91: I know this exists in the current version too, but where else would the name of the class come from? This method doesn't have a parameter to provide the name of a class. IMHO it can only come from `PDO::FETCH_CLASSTYPE`. Am I wrong?

reference/pdo/pdostatement/fetch.xml
kamil-tekiela: Ok, could you clarify this paragraph and say something along the lines that in either case no arguments are provided to the constructor?

reference/pdo/pdostatement/fetch.xml
Girgias: > I know this exists in the current version too, but where else would the name of the class come from? This method doesn't have a parameter to provide the name of a class. IMHO it can only come from `PDO::FETCH_CLASSTYPE`. Am I wrong?

Yeah good point, I guess we should cross ref with `PDOStatement::fetchObject()`

kamil-tekiela: Well, I guess one could also call `$sth->setFetchMode(PDO::FETCH_CLASS, 'Fruit');` but in that case why pass an argument to `fetch()`? Can this maybe be clarified? Maybe something like: "Unless the class name was specified in `setFetchMode`, this mode needs to be combined with`PDO::FETCH_CLASSTYPE`?

lacatoire (approved):

PR #5033, opened 2025-12-09
Document openssl.libctx INI

bukka:
New INI introduced in PHP 8.5

### DIFF
--- a/reference/openssl/ini.xml
+++ b/reference/openssl/ini.xml
@@ -76,14 +76,15 @@
      <type>string</type>
     </term>
     <listitem>
-     <para>
-      Specifies the type of OpenSSL library context to use. The default value,
-      <literal>custom</literal>, creates a separate library context for each
-      worker or thread. This improves isolation from other libraries using
-      OpenSSL and, in ZTS builds, increases separation between threads.
+     <simpara>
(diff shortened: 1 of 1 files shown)


reference/openssl/ini.xml
```diff
@@ -64,6 +70,22 @@
...
+    </term>
+    <listitem>
+     <para>
+      Specifies the type of OpenSSL library context to use. The default value,
+      <literal>custom</literal>, creates a separate library context for each
+      worker or thread. This improves isolation from other libraries using
+      OpenSSL and, in ZTS builds, increases separation between threads.
+      It is also possible to use the <literal>default</literal> value, which
+      causes PHP to use OpenSSL's global default library context.
+     </para>
```
Girgias:
```suggestion
     <simpara>
      Specifies the type of OpenSSL library context to use.
      The default value, <literal>custom</literal>, creates a separate
      library context for each worker or thread.
      This improves isolation from other libraries using OpenSSL and,
      in ZTS builds, increases separation between threads.
      It is also possible to use the <literal>default</literal> value, which
      causes PHP to use OpenSSL's global default library context.
     </simpara>
```

PR #5041, opened 2025-12-14
Add NoDiscard attribute

sasezaki:
- https://wiki.php.net/rfc/marking_return_value_as_important

### DIFF
--- a/language/predefined/attributes/nodiscard.xml
+++ b/language/predefined/attributes/nodiscard.xml
@@ -21,13 +21,15 @@
     cast to suppress the warning.
    </simpara>
    <note>
-    <para>
-     Since attributes are designed to be backwards compatible, #[\NoDiscard] can
-     also be added to functions and methods even when PHP 8.4 or below are supported,
-     it just won't do anything. On PHP 8.5 and above a warning will be emitted if
-     the result is unused. To suppress the warning without using (void), which
-     is not supported before PHP 8.5, consider using a variable like $_.
-    </para>
+    <simpara>
+     Since attributes are designed to be backwards compatible,
+     <code>#[\NoDiscard]</code> can be added to functions and methods
+     even when PHP 8.4 or below are supported, it just won't do anything.
+     On PHP 8.5 and above a warning will be emitted if the result is unused.
+     To suppress the warning without using <code>(void)</code>,
+     which is not supported before PHP 8.5,
+     consider using a variable like <code>$_</code>.
+    </simpara>
    </note>
   </section>
 
@@ -62,9 +64,9 @@
     <varlistentry xml:id="nodiscard.props.message">
      <term><varname>message</varname></term>
      <listitem>
-      <para>
+      <simpara>
        An optional message explaining why the return value should not be discarded.
-      </para>
+      </simpara>
      </listitem>
     </varlistentry>
    </variablelist>
--- a/language/predefined/attributes/nodiscard/construct.xml
+++ b/language/predefined/attributes/nodiscard/construct.xml
@@ -23,9 +23,9 @@
    <varlistentry>
     <term><parameter>message</parameter></term>
     <listitem>
-     <para>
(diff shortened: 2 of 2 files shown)


language/predefined/attributes/nodiscard.xml
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		$results[$key] = $error;
+	}
+
+	return $results;
+}
+
+?>
+]]>
```
DanielEScherzer, line 114: presumably you also have to call the function somewhere

language/predefined/attributes/nodiscard.xml
```diff
@@ -0,0 +1,160 @@
...
+    return 1;
+}
+
+(void) some_command();
+
+?>
+]]>
+    </programlisting>
+   </example>
+  </section>
```
DanielEScherzer, line 144: I suggest also documented somewhere (maybe just an example) for a `$_` or similar being assigned so that the warning isn't shown *without* needing `(void)`, which is only 8.5+, e.g.

> Since attributes are designed to be backwards compatible, `#[\NoDiscard]` can also be added to functions and methods even when PHP 8.4 or below are supported, it just won't do anything. On PHP 8.5 and above a warning will be emitted if the result is unused. To suppress the warning without using `(void)`, which is not supported before PHP 8.5, consider using a variable like `$_`.

> `$_ = some_command();`

language/predefined/attributes/nodiscard.xml
sasezaki: Thanks for noticing! I'll add it.

language/predefined/attributes/nodiscard/construct.xml
```diff
@@ -0,0 +1,53 @@
...
+
+ <refsect1 role="parameters">
+  &reftitle.parameters;
+  <variablelist>
+   <varlistentry>
+    <term><parameter>message</parameter></term>
+    <listitem>
+     <para>
+      The value of the <property linkend="nodiscard.props.message">message</property> property.
+     </para>
```
Girgias:
```suggestion
     <simpara>
      The value of the <property linkend="nodiscard.props.message">message</property> property.
     </simpara>
```

language/predefined/attributes/nodiscard.xml
```diff
@@ -0,0 +1,175 @@
...
+
+  <section xml:id="nodiscard.props">
+   &reftitle.properties;
+   <variablelist>
+    <varlistentry xml:id="nodiscard.props.message">
+     <term><varname>message</varname></term>
+     <listitem>
+      <para>
+       An optional message explaining why the return value should not be discarded.
+      </para>
```
Girgias:
```suggestion
      <simpara>
       An optional message explaining why the return value should not be discarded.
      </simpara>
```

language/predefined/attributes/nodiscard.xml
```diff
@@ -0,0 +1,175 @@
...
+    cast to suppress the warning.
+   </simpara>
+   <note>
+    <para>
+     Since attributes are designed to be backwards compatible, #[\NoDiscard] can
+     also be added to functions and methods even when PHP 8.4 or below are supported,
+     it just won't do anything. On PHP 8.5 and above a warning will be emitted if
+     the result is unused. To suppress the warning without using (void), which
+     is not supported before PHP 8.5, consider using a variable like $_.
+    </para>
```
Girgias:
```suggestion
    <simpara>
     Since attributes are designed to be backwards compatible,
     <code>#[\NoDiscard]</code> can be added to functions and methods
     even when PHP 8.4 or below are supported, it just won't do anything.
     On PHP 8.5 and above a warning will be emitted if the result is unused.
     To suppress the warning without using <code>(void)</code>,
     which is not supported before PHP 8.5,
     consider using a variable like <code>$_</code>.
    </simpara>
```

PR #5043, opened 2025-12-15
PHP 8.1 | Migration guide: add IMAGETYPE_AVIF constant

rodrigoprimo:
Add the `IMAGETYPE_AVIF` constant to the PHP 8.1 migration guide.

The Standard section was placed at the end of the file, after Tokenizer, following the pattern used in the PHP 8.5 migration guide. I'm not sure if this is correct or not.

Refs:
* https://github.com/php/php-src/pull/7091

### DIFF
--- a/appendices/migration81/constants.xml
+++ b/appendices/migration81/constants.xml
@@ -145,22 +145,22 @@
   </itemizedlist>
  </sect2>
 
- <sect2 xml:id="migration81.constants.tokenizer">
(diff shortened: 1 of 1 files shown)


lacatoire:
LGTM! Thanks @rodrigoprimo IMAGETYPE_AVIF confirmed in [src](https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php#L633)

PR #5057, opened 2025-12-23
Fix outdated documentation for ReflectionClass::setStaticPropertyValue visibility handling

### LINKED ISSUE #4600: `ReflectionClass::setStaticPropertyValue()` documentation is outdated
From manual page: https://php.net/reflectionclass.setstaticpropertyvalue

---
> Sets the value of a public static property. If the property is private or protected, the method will fail.

From https://3v4l.org/mSHWS this has not been the case since 7.4.9+, see https://github.com/php/php-src/commit/26aefb750a614391cdb8ee3f63d66f46ed57bd55

Description should probably be

> Sets the value of a static property, regardless of visibility.

With a changelog entry (not sure if the changelog should be 7.4.9, 8.0, or both?)

lacatoire:
Fixes https://github.com/php/doc-en/issues/4600
The documentation for `ReflectionClass::setStaticPropertyValue()` is outdated.

### DIFF
--- a/reference/reflection/reflectionclass/setstaticpropertyvalue.xml
+++ b/reference/reflection/reflectionclass/setstaticpropertyvalue.xml
@@ -3,7 +3,7 @@
 <refentry xml:id="reflectionclass.setstaticpropertyvalue" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <refnamediv>
   <refname>ReflectionClass::setStaticPropertyValue</refname>
-  <refpurpose>Sets public static property value</refpurpose>
+  <refpurpose>Sets static property value</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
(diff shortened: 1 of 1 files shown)


reference/reflection/reflectionclass/setstaticpropertyvalue.xml
```diff
@@ -14,12 +14,11 @@
...
   <simpara>
-   Sets the value of a public static property.
-   If the property is private or protected, the method will fail.
+   Sets the value of a static property, regardless of its visibility.
   </simpara>
   <simpara>
-   <methodname>ReflectionProperty::setValue</methodname> allows setting
-   the value of public, private, and protected properties.
+   <methodname>ReflectionProperty::setValue</methodname> can also be
+   used to set the value of static properties.allows setting.
```
DanielEScherzer:
```suggestion
   used to set the value of static properties.
```

looks like this was left over

reference/reflection/reflectionclass/setstaticpropertyvalue.xml
```diff
@@ -73,6 +72,12 @@
        threw a <classname>ReflectionException</classname>.
       </entry>
      </row>
+     <row>
+      <entry>7.4.9</entry>
```
DanielEScherzer: there should also be documentation of which minor (non-patch) version of PHP this was first changed in, e.g. it could be that 7.4.9 was released after 8.0.0 and thus PHP 8.0 only allows public static properties, or that 7.4.9 was before 8.0.0 and the change was also in 8.0.0, but without further detail a reader would need to investigate

DanielEScherzer (changes requested):

Issue #4886, opened 2025-09-26
PHP 8.5 documentation tracker


DanielEScherzer:
Pages which need to be added/modified for complete documentation of PHP 8.5:

> [!NOTE]
> This issue is still a Work In Progress

> [!IMPORTANT]
> Priority should be given to documenting Core language and ext/standard behavioral changes

rbowen:
Docs contribution onboarding could use some love. It points to the phpdoc@ mailing list which appears to have been largely abandoned as a place to discuss docs. The IRC channel is lonely, and the discord invite link has expired. (I did docs stuff in ancient times, and have been inspired to start helping out again, but can't find where the discussion happens ... unless this is it here?) I'm happy to make these updates but I'm not sure where to find current info.

Sorry if I'm putting this note the wrong place, but it *seemed* like the right place.

jordikroon:
@rbowen Thank you for pointing this out. It is actually on my list to improve, together with some other changes to make building the documentation easier across translations.

Just one thing to confirm. Are you referring to the Discord link at https://phpc.chat? As that seems to be working fine, at least on my end.

rbowen:
I'll have to check again. I met some old php friends at fossy and was going through the docs onboarding. I'll have to post a more useful/specific report with links. I'll get that done tomorrow.

PR #5058, opened 2025-12-25
Add PHP_BUILD_DATE and PHP_BUILD_PROVIDER constants

jordikroon:
Document two new predefined constants introduced in PHP 8.5.0:
- PHP_BUILD_DATE: Returns the date and time when PHP was built
- PHP_BUILD_PROVIDER: Returns the provider who built PHP

### DIFF
--- a/appendices/reserved.constants.core.xml
+++ b/appendices/reserved.constants.core.xml
@@ -87,7 +87,7 @@
    </term>
    <listitem>
     <simpara>
-     The date and time when PHP was built, in "M d Y H:i:s" format. Available as of PHP 8.5.0.
+     The date and time when PHP was built, in <literal>"M d Y H:i:s"</literal> format. Available as of PHP 8.5.0.
     </simpara>
    </listitem>
   </varlistentry>


appendices/reserved.constants.core.xml
```diff
@@ -80,6 +80,28 @@
...
+    <constant>PHP_BUILD_DATE</constant>
+    (<type>string</type>)
+   </term>
+   <listitem>
+    <simpara>
+     The date and time when PHP was built, in "M d Y H:i:s" format. Available as of PHP 8.5.0.
+    </simpara>
+   </listitem>
+  </varlistentry>
+  <varlistentry xml:id="constant.php-build-provider">
```
DanielEScherzer, line 94: The build provider is only documented if it was specified when PHP was built, see https://github.com/php/php-src/commit/7f1b3606a834b6e9103c44d359eb6e0c42e2e1dd - it is not always available. We should document that
* it is not always available
* [maybe, not sure if this should be included] by default (meaning if PHP is built directly from source without configuration) it is not present
* if the provider of the PHP package specified a name to show here, it is available as the build provider, but it can be any random string

DanielEScherzer (changes requested):

appendices/reserved.constants.core.xml
jordikroon: I added a note. As mentioned in the docs also PHP_MANDIR is not always available. Only on non-windows builds. I included a note for that as well.

DanielEScherzer (approved):

appendices/reserved.constants.core.xml
```diff
@@ -80,6 +80,34 @@
...
    </listitem>
   </varlistentry>
+  <varlistentry xml:id="constant.php-build-date">
+   <term>
+    <constant>PHP_BUILD_DATE</constant>
+    (<type>string</type>)
+   </term>
+   <listitem>
+    <simpara>
+     The date and time when PHP was built, in "M d Y H:i:s" format. Available as of PHP 8.5.0.
```
TimWolla:
```suggestion
     The date and time when PHP was built, in <literal>"M d Y H:i:s"</literal> format. Available as of PHP 8.5.0.
```

shivammathur (changes requested):

PR #5059, opened 2025-12-25
Add reserved words restrictions for class-alias in PHP 8.5

jordikroon:
Documentation updated for new reserved class names

 - Update reserved words documentation to include array and callable as reserved words as of PHP 8.5. 
 - Clarify in class_alias() function that aliases cannot use reserved words, with link to reserved words reference.

### DIFF
--- a/reference/classobj/functions/class-alias.xml
+++ b/reference/classobj/functions/class-alias.xml
@@ -18,6 +18,9 @@
    based on the user defined class <parameter>class</parameter>.
    The aliased class is exactly the same as the original class.
   </para>
+  <para>
+   The class alias cannot be any of the PHP <link linkend="reserved.other-reserved-words">reserved words</link>.
+  </para>
    <note>
    <simpara>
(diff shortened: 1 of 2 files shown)


reference/classobj/functions/class-alias.xml
```diff
@@ -17,6 +17,9 @@
    Creates an alias named <parameter>alias</parameter>
    based on the user defined class <parameter>class</parameter>.
    The aliased class is exactly the same as the original class.
+  </para>
+  <para>
```
lacatoire:
```suggestion
  <simpara>
```

reference/classobj/functions/class-alias.xml
```diff
@@ -17,6 +17,9 @@
...
+   The class alias cannot be any of the PHP <link linkend="reserved.other-reserved-words">reserved words</link>.
   </para>
```
lacatoire:
```suggestion
  </simpara>
```

kamil-tekiela (approved):

PR #5064, opened 2025-12-29
Refactor NULL and TRUE/FALSE to use their lowercase variant

jordikroon:
Within the documentation there are many inconsistencies. One of those is how null and booleans are written.
At some places it is written as NULL where in other places null is used.

Within parameters you generally want to use the lowercase variants. And also when mentioning the type null on its own.
At places like var dump results, you generally want to use the uppercase variant to make it in line with what PHP returns.

This change tries to tackle these inconsistencies. I did my very best not to touch values that are irrelevant, for example the value NULL within SQL. But since it we touch a lot of different files, have a good look.

The same has been done with TRUE/FALSE. Though in those situations it is almost always the lowercase variant.

### DIFF
--- a/reference/apcu/functions/apcu-add.xml
+++ b/reference/apcu/functions/apcu-add.xml
@@ -18,7 +18,7 @@
   <methodsynopsis>
    <type>array</type><methodname>apcu_add</methodname>
    <methodparam><type>array</type><parameter>values</parameter></methodparam>
-   <methodparam choice="opt"><type>mixed</type><parameter>unused</parameter><initializer>NULL</initializer></methodparam>
+   <methodparam choice="opt"><type>mixed</type><parameter>unused</parameter><initializer>null</initializer></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>ttl</parameter><initializer>0</initializer></methodparam>
   </methodsynopsis>
   <para>
@@ -86,7 +86,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns TRUE if something has effectively been added into the cache, FALSE otherwise.
+   Returns <type>true</type> if something has effectively been added into the cache, <type>false</type> otherwise.
    Second syntax returns array with error keys.
   </para>
  </refsect1>
--- a/reference/com/constants.xml
+++ b/reference/com/constants.xml
@@ -505,7 +505,7 @@
    </term>
    <listitem>
     <simpara>
-     Either expression is NULL.
+     Either expression is <type>null</type>.
     </simpara>
    </listitem>
   </varlistentry>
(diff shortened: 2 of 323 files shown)


kamil-tekiela:
What is the actual benefit? 

I'd rather not do it. Constants are usually written in all uppercase letters, whereas types are written in lowercase, which makes these three values somewhat ambiguous. https://3v4l.org/7NpFtB 

I'd prefer that PHP manual doesn't dictate the style as long as the language doesn't differentiate.

jordikroon:
On the technical side there are no benefits since the compiler will use the same Opcode regardless of using the uppercase, lowercase or a variant. E.g. `NuLL` would also work.

However there are some aspects of interest I believe. Within the stubs the lowercase variant is exclusively used. That means that all documented method/function signatures are all in lowercase already and changing it within examples would make it more consistent. 

Secondly internally these are considered special constants but technically they are considered literals. So in some way it remove ambiguity with regular constants or predefined flags.  

For example, this would be allowed:

```php
const BOOL = 123;

function test(NULL|BOOL $test ) {
    return BOOL;
}
    
echo test(true);
```

But this wouldn't because only null/true/false are considered special constants. 

```php
const NULL = 123;

function test(NULL|BOOL $test ) {
    return NULL;
}
    
echo test(true);
```

So while I do agree that this is very opinionated, I believe it aligns more with other resources and modern standards. I also agree that we should not dictate styles. But as you mentioned that we can generally accept constants to be in uppercase (though PHP doesn't differentiate either), we also generally accept literals to be in lowercase.

Though if this change is not desirable within the docs, I am happy to accept that fact.

reference/apcu/functions/apcu-add.xml
```diff
@@ -86,7 +86,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns TRUE if something has effectively been added into the cache, FALSE otherwise.
+   Returns <type>true</type> if something has effectively been added into the cache, <type>false</type> otherwise.
```
Girgias, line 89: This is not even how it should be written in the first place, it should use the XML entities.

reference/com/constants.xml
```diff
@@ -505,7 +505,7 @@
    </term>
    <listitem>
     <simpara>
-     Either expression is NULL.
+     Either expression is <type>null</type>.
```
Girgias, line 508: This is not a PHP null, but a COM NULL pointer.

Girgias (changes requested):
No, just no.

I can appreciate the thought of standardizing the coding style. But this is just going to flood translations with changes that need to be synced that do not provide a tangible benefit.

Especially as it seems the majority of changes aren't even correct, the pages in question are not bundled extensions so have even more problems, or is affecting output that is actually produced by PHP.

PR #5080, opened 2026-01-07
docs(printf): fix incorrect cutoff value in examples

democod:
Found a regression from 2019 where the comment says "8 characters" but the code uses .9s. Correcting it to "9 characters" for consistency with the actual output.


Girgias (approved):
Thank you!

PR #5083, opened 2026-01-08
Describe precedence for pipe and throw operators

vadimonus:
Pipe operator docs was introduced in 8.5 (https://github.com/php/doc-en/pull/4890), but currently missing in precedence table.

Throw operator was intorduced in 8.0 (https://wiki.php.net/rfc/throw_expression), but precedence still not documented.

### DIFF
--- a/language/operators/precedence.xml
+++ b/language/operators/precedence.xml
@@ -145,7 +145,7 @@
       </entry>
      </row>
      <row>
-      <entry>(n/a)</entry>
+      <entry>left</entry>
       <entry><literal>|&gt;</literal></entry>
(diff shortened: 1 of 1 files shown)


language/operators/precedence.xml
```diff
@@ -144,6 +144,13 @@
        <link linkend="language.operators.string">string</link> (as of PHP 8.0.0)
       </entry>
      </row>
+     <row>
+      <entry>(n/a)</entry>
```
Crell:
```suggestion
      <entry>left</entry>
```

Pipe is left associative.

Issue #4957, opened 2025-10-27
Alias link missing in SplObjectStorage::attach


Grummfy:
From manual page: https://php.net/splobjectstorage.attach

When you go to offsetset you see it's an alias of attach who is deprecated. But the other way around doesn't exist. Since it's a deprecated function, it would be nice to have some indication about the alias. 

Thanks

kamil-tekiela:
I think this issue can be closed.

PR #5089, opened 2026-01-09
Document offsetSet() as alias of deprecated SplObjectStorage::attach()

lacatoire:
`SplObjectStorage::offsetSet()` is an alias of `SplObjectStorage::attach()`,
which is deprecated as of PHP 8.5.

This change makes the alias relationship explicit in the documentation.
Issue: https://github.com/php/doc-en/issues/4957

### DIFF
--- a/reference/spl/splobjectstorage/attach.xml
+++ b/reference/spl/splobjectstorage/attach.xml
@@ -21,6 +21,10 @@
   <para>
    Adds an <type>object</type> inside the storage, and optionally associate it to some data.
   </para>
+  <para>
+   This method is an alias of <methodname>SplObjectStorage::offsetSet</methodname>.
+  </para>
+
  </refsect1>
 
  <refsect1 role="parameters">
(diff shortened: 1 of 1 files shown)


reference/spl/splobjectstorage/attach.xml
```diff
@@ -21,6 +21,10 @@
   <para>
    Adds an <type>object</type> inside the storage, and optionally associate it to some data.
   </para>
+  <para>
+   This method is an alias of <methodname>SplObjectStorage::offsetSet</methodname>.
+  </para>
```
Girgias:
```suggestion
  <simpara>
   This method is an alias of <methodname>SplObjectStorage::offsetSet</methodname>.
  </simpara>
```

Girgias (approved):
Normally when we mark a page as an alias we just link to the other page, but that's only for functions... I think this is the first time this happened with methods. So I guess this is fine.

kamil-tekiela:
> Normally when we mark a page as an alias we just link to the other page, but that's only for functions... I think this is the first time this happened with methods. So I guess this is fine.

We also do that with methods: https://www.php.net/manual/en/function.mysqli-escape-string.php

Issue #4950, opened 2025-10-24
PHP 8.5: User contributed notes for `setlocale` are deprecated


lat9:
### Affected page

https://www.php.net/manual/en/function.setlocale.php

### Current issue

Since `setlocale` no longer accepts an `(int)0` as its second argument, the majority of the User Contributed Notes (going back to 2002) are no longer valid and will result in a PHP error.

### Suggested improvement

Update the various notes to change `0` to `'0'`.

PR #5094, opened 2026-01-12
Document setlocale() incompatibility with integer 0 in PHP 8.5

lacatoire:
As of PHP 8.5, passing an integer `0` as the `locales` argument to `setlocale()` is no longer supported and triggers a `TypeError`.

This PR adds a short note to clarify this change and to warn that older examples or user contributed notes may still reference `setlocale(category, 0)`, which is no longer valid.

This helps reduce confusion for users upgrading to PHP 8.5 and copying legacy examples that were historically correct.

Issue https://github.com/php/doc-en/issues/4950

### DIFF
--- a/reference/strings/functions/setlocale.xml
+++ b/reference/strings/functions/setlocale.xml
@@ -165,7 +165,6 @@
    </para>
   </note>
  </refsect1>
- 
  <refsect1 role="changelog">
   &reftitle.changelog;
   <informaltable>
@@ -189,7 +188,6 @@
    </tgroup>
   </informaltable>
  </refsect1>
- 
  <refsect1 role="examples">
   &reftitle.examples;
   <para>


reference/strings/functions/setlocale.xml
```diff
@@ -112,6 +112,18 @@
...
+        As of PHP 8.5.0, passing an integer <literal>0</literal> as the
+        <parameter>locales</parameter> argument is no longer supported and
+        will result in a <classname>TypeError</classname>.
+       </para>
+       <para>
+        Older documentation and user contributed notes may still refer to
+        <code>setlocale(category, 0)</code>. Such usage should be updated to
+        use the string <literal>"0"</literal> instead.
+       </para>
+      </note>
```
Girgias: You should create a new refsection for the 8.5 changelog, also if you give me the note IDs I can either delete or fix them manually so we don't mention the user notes.

lacatoire: Thanks @Girgias ! I’ll add a dedicated PHP 8.5.0 changelog entry for this change and keep the description free of version-specific notes.

PR #5098, opened 2026-01-13
Clarify intval() behavior with scientific notation numeric strings

### LINKED ISSUE #5019: intval returns unxpected result for string with "." and "e" which is not documented.
```
echo intval('42.'); //42     ok
echo intval('42.42'); //42     ok
echo intval('42.42e'); //42    ok
echo intval('42.42e42'); //9223372036854775807
```

---

From manual page: https://php.net/function.intval

lacatoire:
This PR clarifies how intval() behaves when passed numeric strings using scientific notation.

While this behavior is documented in the Numeric strings section, it was not explicitly referenced from the intval() manual page,
which could lead to confusion.

A short note and example have been added to make this behavior easier to discover.

Issue: https://github.com/php/doc-en/issues/5019

### DIFF
--- a/reference/var/functions/intval.xml
+++ b/reference/var/functions/intval.xml
@@ -97,26 +97,25 @@
    apply.
   </para>
   <note>
-   <para>
+   <simpara>
     Numeric strings using scientific notation (for example containing
     the letter <literal>e</literal>) are first parsed as numbers before
     being converted to an integer.
-   </para>
-   <para>
+   </simpara>
+   <simpara>
     This may lead to unexpected results for large exponents:
-   </para>
-   <example>
-    <title>Example</title>
+   </simpara>
+   <informalexample>
     <programlisting role="php">
 <![CDATA[
 echo intval('42.42e42'); // 9223372036854775807 on 64-bit systems
 ]]>
     </programlisting>
-   </example>
-   <para>
+   </informalexample>
+   <simpara>
     See <link linkend="language.types.numeric-strings">Numeric strings</link>
     for details on how such strings are interpreted.
-   </para>
+   </simpara>
   </note>
  </refsect1>
 
--- a/reference/var/functions/intval.xml
+++ b/reference/var/functions/intval.xml
@@ -96,6 +96,27 @@
    <link linkend="language.types.integer.casting">integer casting</link> 
    apply.
   </para>
+  <note>
+   <para>
+    Numeric strings using scientific notation (for example containing 
+    the letter <literal>e</literal>) are first parsed as numbers before
+    being converted to an integer.
+   </para>
+   <para>
+    This may lead to unexpected results for large exponents:
+   </para>
+   <example>
+    <programlisting role="php">
+<![CDATA[
+echo intval('42.42e42'); // 9223372036854775807 on 64-bit systems
+]]>
+    </programlisting>
+   </example>
+   <para>
+    See <link linkend="language.types.numeric-strings">Numeric strings</link>
+    for details on how such strings are interpreted.
+   </para>
+  </note>
  </refsect1>
 
  <refsect1 role="changelog">
--- a/reference/var/functions/intval.xml
+++ b/reference/var/functions/intval.xml
@@ -98,7 +98,7 @@
   </para>
   <note>
    <para>
-    Numeric strings using scientific notation (for example containing 
+    Numeric strings using scientific notation (for example containing
     the letter <literal>e</literal>) are first parsed as numbers before
     being converted to an integer.
    </para>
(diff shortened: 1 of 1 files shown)


reference/var/functions/intval.xml
```diff
@@ -96,6 +96,28 @@
...
+<![CDATA[
+echo intval('42.42e42'); // 9223372036854775807 on 64-bit systems
+]]>
+    </programlisting>
+   </example>
+   <para>
+    See <link linkend="language.types.numeric-strings">Numeric strings</link>
+    for details on how such strings are interpreted.
+   </para>
+  </note>
```
Girgias, line 123:
```suggestion
  <note>
   <simpara>
    Numeric strings using scientific notation (for example containing
    the letter <literal>e</literal>) are first parsed as numbers before
    being converted to an integer.
   </simpara>
   <simpara>
    This may lead to unexpected results for large exponents:
   </simpara>
   <informalexample>
    <programlisting role="php">
<![CDATA[
echo intval('42.42e42'); // 9223372036854775807 on 64-bit systems
]]>
    </programlisting>
   </informalexample>
   <simpara>
    See <link linkend="language.types.numeric-strings">Numeric strings</link>
    for details on how such strings are interpreted.
   </simpara>
  </note>
```

Girgias (approved):
Outside markup nits, LGTM.

reference/var/functions/intval.xml
```diff
@@ -96,6 +96,27 @@
...
    apply.
   </para>
+  <note>
+   <simpara>
+    Numeric strings using scientific notation (for example containing
+    the letter <literal>e</literal>) are first parsed as numbers before
+    being converted to an integer.
+   </simpara>
+   <simpara>
+    This may lead to unexpected results for large exponents:
```
kamil-tekiela: Does it lead to expected results for small exponents?

kamil-tekiela:
I don't know if this change is really needed. I don't think scientific notation is that special that it requires an additional note here. It is already stated that:

> Strings will most likely return 0 although this depends on the leftmost characters of the string. The common rules of [integer casting](https://www.php.net/manual/en/language.types.integer.php#language.types.integer.casting) apply.

Even though I don't like the first sentence, it does a good job at pointing the user into understanding what numeric strings are and what rules apply. If anything I would just improve this sentence instead of adding a new note.

lacatoire:
Fair point @kamil-tekiela , my reasoning for keeping the note:
The surprise is _why doesn't intval('10e5') return 10?_, and someone expecting `intval` to stop at the first `non-digit` (like it does for '10abc') won't click through to Numeric strings unprompted. 
The note makes that specific case discoverable from here. I can drop it if you think it still isn't worth it.

reference/var/functions/intval.xml
```diff
@@ -96,6 +96,29 @@
    <link linkend="language.types.integer.casting">integer casting</link> 
    apply.
   </para>
+  <note>
+   <simpara>
+    Numeric strings using scientific notation (for example containing
```
kamil-tekiela: Are there other examples of scientific notation? I believe it's only with the letter E.

reference/var/functions/intval.xml
```diff
@@ -96,6 +96,29 @@
...
+   </simpara>
+   <simpara>
+    Because the whole numeric string is parsed, the result is not simply
+    the leading integer part. Large exponents can additionally overflow to
+    <constant>PHP_INT_MAX</constant>:
+   </simpara>
+   <informalexample>
+    <programlisting role="php">
+<![CDATA[
+echo intval('42.42e42'); // 9223372036854775807 on 64-bit systems
```
kamil-tekiela, line 114: Shouldn't it have an opening PHP tag?

reference/var/functions/intval.xml
```diff
@@ -96,6 +96,29 @@
...
    apply.
   </para>
+  <note>
+   <simpara>
+    Numeric strings using scientific notation (for example containing
+    the letter <literal>e</literal>) are first parsed as numbers before
+    being converted to an integer.
+   </simpara>
+   <simpara>
+    Because the whole numeric string is parsed, the result is not simply
```
kamil-tekiela: That's also not necessarily true. Only the number part of the numeric string is parsed. So, it still has to start with a number, sign, or whitespace.

kamil-tekiela:
> Fair point @kamil-tekiela , my reasoning for keeping the note: The surprise is _why doesn't intval('10e5') return 10?_, and someone expecting `intval` to stop at the first `non-digit` (like it does for '10abc') won't click through to Numeric strings unprompted. The note makes that specific case discoverable from here. I can drop it if you think it still isn't worth it.

I don't think it's going to hurt if we have this information highlighted here. I think it can be helpful, especially with the overflow part. After the edits, it's better than it was initially.

kamil-tekiela (approved):

PR #5100, opened 2026-01-13
Docs: document intval() behavior with scientific notation

### LINKED ISSUE #5019: intval returns unxpected result for string with "." and "e" which is not documented.

anshujain0:
This PR adds a documentation note explaining how intval() behaves when
converting strings containing scientific notation, including cases where
integer overflow results in PHP_INT_MAX.

Fixes #5019

### DIFF
--- a/reference/var/functions/is-numeric.xml
+++ b/reference/var/functions/is-numeric.xml
@@ -68,6 +68,17 @@
   </informaltable>
  </refsect1>
 
+  <refsect1 role="notes">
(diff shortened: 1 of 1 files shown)


Girgias:
Conflicts with #5098

PR #5101, opened 2026-01-13
Another description of proprietary fgetscv

alfsb:
While inspecting some `<parameter>` named `escaped` instead of `escape` in doc-pt_BR translation, I think the new text is still incorrect.

> Inside an enclosure,

Inside enclosed strings in the input stream

> the enclosure character can always be escaped by doubling it,

"escaped" is a textual word, not `<parameter>` here.

> resulting in a single enclosure character in the parsed result. The escaped character works differently: If it is followed by an enclosure character then that enclosure character will not be treated as one,

the sequence of escape and enclosure characters will be preserved in the parsed value.

Sugested here:
```xml
       In the input stream, the <parameter>enclosure</parameter> character
       can always be escaped by doubling it inside a quoted string,
       resulting in a single <parameter>enclosure</parameter> character
       in the parsed result.
       The <parameter>escape</parameter> character works differently:
       If a sequence of <parameter>escape</parameter> and
       <parameter>enclosure</parameter> characters appear in the stream,
       both characters will be present in the parsed value.
       So for the default parameters, <literal>""</literal>
       in quoted strings will be parsed as <literal>"</literal>, while
       <literal>\"</literal> parsed as <literal>\"</literal>.
```

I would like revisions, both about the English and function description.

### DIFF
--- a/reference/session/functions/session-cache-limiter.xml
+++ b/reference/session/functions/session-cache-limiter.xml
@@ -72,7 +72,7 @@
 <![CDATA[
 Expires: (sometime in the future, according session.cache_expire)
 Cache-Control: public, max-age=(sometime in the future, according to session.cache_expire)
-Last-Modified: (the timestamp of when the session was last saved)
+Last-Modified: (the timestamp of the current script)
 ]]>
            </programlisting>
           </entry>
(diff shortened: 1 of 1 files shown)


Girgias (approved):

Girgias:
Although realising that this should probably also be fixed for the SPL method....

Issue #4984, opened 2025-11-04
Inconsistency of descriptions in stats_rand_gen_funiform


Marqitos:
### Affected page

https://www.php.net/manual/en/function.stats-rand-gen-funiform.php

### Issue description

The function description says:
Generates uniform float between **low (exclusive)** and high (exclusive)

But in parameters description says:

- low: The **lower bound (inclusive)**
- high: The upper bound (exclusive)

Lower bound description is inconsistent (exclusive/inclusive).

Narkunan:
@Marqitos shall I work on the Fix please?

PR #5105, opened 2026-01-14
Fix inconsistent bounds description in stats_rand_gen_funiform()

lacatoire:
The function description stated that both bounds were exclusive, while the parameter documentation specifies that the lower bound is inclusive and the upper bound exclusive.

This change aligns the refpurpose with the documented parameter behavior and removes an internal inconsistency. 
Issue: https://github.com/php/doc-en/issues/4984


Girgias (changes requested):
The PECL source code seems to say that exclusive is correct: https://github.com/php/pecl-math-stats/blob/master/randlib.c#L702

Issue #4978, opened 2025-11-03
yaml_parse_file() return code not as given in the manual


rmfrankch:
From manual page: https://php.net/function.yaml-parse-file

---
The manual clains:
"Returns the value encoded in input in appropriate PHP type or false on failure. If pos is -1 an array will be returned with one entry for each document found in the stream."

When installed using apt install php-yaml, the versions 8.2 and 8.3 return either an array with the parsed data or a string with an error message. It never returns false.

Narkunan:
@rmfrankch Shall I work on the fix Please?

PR #5106, opened 2026-01-14
Fix incorrect return value description for yaml_parse_file()

lacatoire:
The manual stated that yaml_parse_file() returns false on failure.

In practice, the function returns either the parsed value (usually an array) or a string containing an error message, and does not return false.

This PR updates the return value documentation to reflect the actual behavior observed with ext-yaml on PHP 8.2 and 8.3.
Issue: https://github.com/php/doc-en/issues/4978

### DIFF
--- a/reference/yaml/functions/yaml-parse-file.xml
+++ b/reference/yaml/functions/yaml-parse-file.xml
@@ -69,10 +69,15 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns the value encoded in <parameter>filename</parameter> in appropriate
-   PHP type&return.falseforfailure;. If <parameter>pos</parameter> is <literal>-1</literal> an
-   <type>array</type> will be returned with one entry for each document found
-   in the stream.
+   Returns the value encoded in <parameter>filename</parameter> in the appropriate
+   PHP type.
+  </para>
+  <para>
+   On failure, a string containing an error message is returned.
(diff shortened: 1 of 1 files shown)


reference/yaml/functions/yaml-parse-file.xml
```diff
@@ -69,10 +69,15 @@
...
+   Returns the value encoded in <parameter>filename</parameter> in the appropriate
+   PHP type.
+  </para>
+  <para>
+   On failure, a string containing an error message is returned.
+  </para>
+  <para>
+   If <parameter>pos</parameter> is <literal>-1</literal>, an <type>array</type>
+   will be returned with one entry for each document found in the stream.
   </para>
```
Girgias:
```suggestion
  </simpara>
  <simpara>
   On failure, a string containing an error message is returned.
  </simpara>
  <simpara>
   If <parameter>pos</parameter> is <literal>-1</literal>, an <type>array</type>
   will be returned with one entry for each document found in the stream.
</simpara>
```

reference/yaml/functions/yaml-parse-file.xml
```diff
@@ -69,10 +69,15 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
```
Girgias:
```suggestion
  <simpara>
```

Girgias (approved):

PR #5132, opened 2026-01-18
gmagick: fix XML by converting para to simpara tags via script

Girgias:


reference/gmagick/book.xml
```diff
@@ -1,36 +1,34 @@
...
+<book xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="book.gmagick">
  <?phpdoc extension-membership="pecl" ?>
  <title>Gmagick</title>
  <titleabbrev>Gmagick</titleabbrev>
 
  <preface xml:id="intro.gmagick">
   &reftitle.intro;
-  <para>
+  <simpara>
    &warn.experimental;
```
jordikroon: This resolves to a `<warning>/<div>`. The script likely needs to consider `language-snippets.ent` also.

Girgias: Handling XML entities is a pain, but the simple solution is to just move the XML entity out of the para tag.

But thanks for locating it :)

Girgias:
Need to wait for #5133 to be applied across translations (which I will do when REVCHECK status updates)

PR #5175, opened 2026-01-20
fsockopen's error_code & error_message are nullable.

divinity76:
tested on 8.3.4.

### DIFF
--- a/reference/network/functions/fsockopen.xml
+++ b/reference/network/functions/fsockopen.xml
@@ -12,8 +12,8 @@
    <type class="union"><type>resource</type><type>false</type></type><methodname>fsockopen</methodname>
    <methodparam><type>string</type><parameter>hostname</parameter></methodparam>
    <methodparam choice="opt"><type>int</type><parameter>port</parameter><initializer>-1</initializer></methodparam>
-   <methodparam choice="opt"><type>int</type><parameter role="reference">error_code</parameter><initializer>&null;</initializer></methodparam>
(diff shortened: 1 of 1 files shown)


lacatoire (approved):

lacatoire:
LGTM. In php-src [`ext/standard/basic_functions.stub.php` line 3836](https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php#L3836), the signature is: 

```php                                                                                                                                                                                                  
  function fsockopen(string $hostname, int $port = -1, &$error_code = null, &$error_message = null, ?float $timeout = null) {}                                                                            

  The error_code and error_message parameters have no explicit type but default to null. The current documentation shows int and string with = null, which is inconsistent. Changing to int|null and string|null correctly reflects that these parameters accept null as their default value.

**Comment by jordikroon** [comment]
Could you give an example where `$error_code` or `$error_message` is in fact null? While the stub mentions `null`. The above DocBlock states they will not return `null`. 

Practically the only reason it is set to `null`, is because this parameter may be omitted. But there is no way we can pass null (or any other value other than a variable).

Also if we to change this function, we should also change it's companion -> `pfsockopen`.

**Comment by divinity76** [comment]
Whoa, when testing it now, it is in fact not nullable ❓ 
```
$ php -r 'fsockopen("www.example.com", 80, null, null, 30);'
PHP Fatal error:  Uncaught Error: fsockopen(): Argument #3 ($error_code) could not be passed by reference in Command line code:1
Stack trace:
#0 {main}
  thrown in Command line code on line 1

real	0m0.053s
user	0m0.020s
sys	0m0.008s
```

no idea what happened.. closing.

PR #5176, opened 2026-01-20
Add strict inequality check for Suit enumeration

kevindecapite:
The paragraph above the examples states:

> By default, cases are not intrinsically backed by a scalar value. That is, Suit::Hearts is not equal to "0". Instead, each case is backed by a singleton object of that name.

This new example demonstrates that more clearly than the others.


kevindecapite:
Do I need to do anything here?

PR #5179, opened 2026-01-21
curl_multi_close: add changelog for NOP and deprecation

DanielEScherzer:


lacatoire:
Thanks for adding the changelog entries! The 8.0.0 NOP entry looks good 👍                                                                                                                                                           For the 8.5.0 deprecation though, I couldn't find evidence that `curl_multi_close()` was deprecated. Looking at                                                                                         
[`ext/curl/curl.stub.php`](https://github.com/php/php-src/blob/PHP-8.5/ext/curl/curl.stub.php), `curl_close()` has the deprecation attribute:                                                           
                                                                                                                                                                                                          
  ```php                                                                                                                                                                                                  
  #[\Deprecated(since: '8.5', message: "as it has no effect since PHP 8.0")]                                                                                                                              
  function curl_close(CurlHandle $handle): void {}                                                                                                                                                        
                                                                                                                                                                                                          
  But https://github.com/php/php-src/blob/PHP-8.5/ext/curl/curl.stub.php#L109 doesn't seem to have one:                                                                                                   
                                                                                                                                                                                                          
  function curl_multi_close(CurlMultiHandle $multi_handle): void {}                                                                                                                                       
                                                                                                                                                                                                          
  Am I missing something? If it's intentionally not deprecated, maybe we should remove that entry from this PR.

DanielEScherzer:
Sent #5249 with the correct function

PR #5181, opened 2026-01-23
Document environment variable expansion in configuration

yariksheptykin:
I noticed that I can use variable expansion within `php-fpm.conf` files. The documentation page does not explicitly mentions this. Instead it only provides examples of passing environment variables on to the child process.

At the very top of the page there is a mention that `php.ini` syntax is used. This somewhat implies that variable expansion is supported just like in `php.ini`. I suggest though that we update the discussion on environment variables with an example of how it php-fpm can be configured using env vars to make the support for this more explicit.

### DIFF
--- a/install/fpm/configuration.xml
+++ b/install/fpm/configuration.xml
@@ -1018,6 +1018,7 @@ group = ${USER_NAME:-www-data}
         </programlisting>
        </example>
     </para>
+    <para>
      It's possible to pass additional environment variables and update PHP settings of a certain pool.
      To do this, you need to add the following options to the pool configuration file.
      <example>
--- a/install/fpm/configuration.xml
+++ b/install/fpm/configuration.xml
@@ -1004,6 +1004,20 @@
      </varlistentry>
     </variablelist>
     <para>
+     Configuration values can be set using environment variable expansion.
+     Setting default values when the environment variable is unset or null is also supported.
+     For example:
+        <example>
+        <title>Using environment variable expansion</title>
+        <programlisting role="ini">
+<![CDATA[
+listen = /var/run/php-fpm-$POOL_NAME.sock
+user = ${USER_NAME:-www-data}
+group = ${USER_NAME:-www-data}
+]]>
+        </programlisting>
+       </example>
+    </para>
      It's possible to pass additional environment variables and update PHP settings of a certain pool.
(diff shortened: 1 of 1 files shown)


install/fpm/configuration.xml
```diff
@@ -1004,6 +1004,20 @@
...
+        <title>Using environment variable expansion</title>
+        <programlisting role="ini">
+<![CDATA[
+listen = /var/run/php-fpm-$POOL_NAME.sock
+user = ${USER_NAME:-www-data}
+group = ${USER_NAME:-www-data}
+]]>
+        </programlisting>
+       </example>
+    </para>
```
yariksheptykin, line 1020:
```suggestion
    </para>
    <para>
```

yariksheptykin (commented):
balance

install/fpm/configuration.xml
```diff
@@ -1003,6 +1003,21 @@
...
     </variablelist>
+    <para>
+     Configuration values can be set using environment variable expansion.
+     Setting default values when the environment variable is unset or null is also supported.
+     For example:
+        <example>
+        <title>Using environment variable expansion</title>
+        <programlisting role="ini">
+<![CDATA[
+listen = /var/run/php-fpm-$POOL_NAME.sock
```
lacatoire:
```suggestion
listen = /var/run/php-fpm-${POOL_NAME}.sock
```

install/fpm/configuration.xml
```diff
@@ -1003,6 +1003,21 @@
       </listitem>
      </varlistentry>
     </variablelist>
+    <para>
+     Configuration values can be set using environment variable expansion.
+     Setting default values when the environment variable is unset or null is also supported.
```
lacatoire:
```suggestion
     Setting default values when the environment variable is unset or null is also supported as of PHP 8.3.0.
```

lacatoire: Since [bc8b9aedf](https://github.com/php/php-src/commit/bc8b9aedf)

lacatoire (approved):

yariksheptykin:
@lacatoire anything else you want me to do for this PR?

lacatoire:
LGTM !

yariksheptykin:
image

PR #5182, opened 2026-01-24
grapheme-str-split.xml Remove extra `locale` description

mmalferov:
…ature


youkidearitai (approved):
Yes, your right. LGTM.

PR #5204, opened 2026-01-26
Ps xml auto fixes

Girgias:


lacatoire (approved):

jordikroon:
I am going to close this PR. Not because I disagree with it, but because in the process of doc-extensions, we will do an all-in-one docbook-cs fix run. Which should address most issues without having to bother translators.

PR #5215, opened 2026-01-28
Update set.xml Specify the correct return type and values, CS

mmalferov:
Let's tell the reader the truth :)

```/**
 * Adds a new element, or updates the value if the key already is part of the hash
 *
 * Parameters:
 * - hash: A valid quickhash
 * - key: The key
 * - value: The value
 *
 * Returns:
 * - 1 if the element is part of the hash and was updated, 2 if the element was
 *   added or 0 if an error occurred.
 */
int qhi_hash_set(qhi *hash, qhv key, qhv value)
{...

    if (find_bucket_from_list(...)) {
        return qhi_update_value_in_bucket(...); // returns 1
    } else {
        return qhi_add_entry_to_list(...) ? 2 : 0; // returns 2 or 0
    }...
```

### DIFF
--- a/reference/quickhash/quickhashinthash/set.xml
+++ b/reference/quickhash/quickhashinthash/set.xml
@@ -10,7 +10,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <modifier>public</modifier> <type>bool</type><methodname>QuickHashIntHash::set</methodname>
+   <modifier>public</modifier> <type>int</type><methodname>QuickHashIntHash::set</methodname>
    <methodparam><type>int</type><parameter>key</parameter></methodparam>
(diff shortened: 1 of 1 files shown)


lacatoire:
LGTM! Verified against the source:

- Return type is int, not bool: The https://github.com/derickr/quickhash/blob/master/qh_inthash.c#L283 says proto int QuickHashIntHash::set(...) and the function uses RETURN_LONG() (not RETURN_BOOL()).

- Return values were swapped in the old docs: The C function https://github.com/derickr/quickhash/blob/master/lib/quickhash.c#L1008 documents itself as: "1  if the element is part of the hash and was updated, 2 if the element was added or 0 if an error occurred." The old documentation had 1 and 2 reversed.

kamil-tekiela (approved):

PR #5218, opened 2026-01-29
Fix ldap_count_entries() return value documentation

### LINKED ISSUE #4683: ldap_count_entries() return type should be int|false, not int
**Affected page**

_([Manual Page](https://www.php.net/manual/en/function.ldap-count-entries.php))_

**Issue description**

The function signature for ldap_count_entries() in the PHP stub and reflection API is currently declared as:
`
function ldap_count_entries(LDAP\Connection $ldap, LDAP\Result $result): int
`
However, the actual behaviour (and [manual documentation](https://www.php.net/manual/en/function.ldap-count-entries.php)) clearly state that the function returns false on failure.

**Steps to reproduce**

Example:
`
var_dump(ldap_count_entries($ldap, $invalidResult)); // bool(false) on failure
`
Reflection output:
`
$func = new ReflectionFunction('ldap_count_entries');
var_dump($func->getReturnType()->getName()); // "int"
`
This causes problems for strict types, static analysis, and runtime expectations.

**Suggested fix**

The function should be declared as:
`
function ldap_count_entries(LDAP\Connection $ldap, LDAP\Result $result): int|false
`
Please update ext/ldap/ldap.stub.php accordingly.

lacatoire:
The manual stated that `ldap_count_entries()` returns false on failure.
According to the OpenLDAP documentation and the PHP implementation, the function returns `-1` on error.
This PR corrects the return value documentation.

Reference: https://www.openldap.org/software/man.cgi?query=ldap_count_entries ("If an error occurs in ldap_count_entries(), -1 is returned, and ld_errno is set appropriately.")

Fixes https://github.com/php/doc-en/issues/4683

### DIFF
--- a/reference/ldap/functions/ldap-count-entries.xml
+++ b/reference/ldap/functions/ldap-count-entries.xml
@@ -46,7 +46,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   Returns number of entries in the result,&return.falseforfailure;.
+   Returns the number of entries in the result, or <literal>-1</literal> on error.
   </para>
  </refsect1>
 
--- a/reference/ldap/functions/ldap-count-entries.xml
+++ b/reference/ldap/functions/ldap-count-entries.xml
@@ -45,9 +45,9 @@
 
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
-  <para>
+  <simpara>
    Returns the number of entries in the result, or <literal>-1</literal> on error.
-  </para>
+  </simpara>
  </refsect1>
 
  <refsect1 role="changelog">


reference/ldap/functions/ldap-count-entries.xml
```diff
@@ -46,7 +46,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
```
lacatoire:
```suggestion
  <simpara>
```

reference/ldap/functions/ldap-count-entries.xml
```diff
@@ -46,7 +46,7 @@
...
-   Returns number of entries in the result,&return.falseforfailure;.
+   Returns the number of entries in the result, or <literal>-1</literal> on error.
   </para>
```
lacatoire:
```suggestion
  </simpara>
```

PR #5233, opened 2026-01-30
Document curl_errno/curl_error behavior with multi handles

### LINKED ISSUE #5047: curl_errno() docs should mention that it always returns 0 when the CurlHandle was executed via a CurlMultiHandle
### Affected page

https://php.net/function.curl-errno, https://php.net/function.curl-error

### Current issue

When executing an individual `CurlHandle` using `curl_exec()`, you can then retrieve potential errors and error codes using `curl_errno()` or `curl_error()`.

When a `CurlHandle` is executed via a `CurlMultiHandle` using `curl_multi_exec()`, these methods will simply return `0` or `""` respectively, as if no error had occured.

This behaviour is not documented anywhere, neither on the page for `curl_errno()`, nor for `curl_error()`, nor for `curl_multi_info_read()`.

### Suggested improvement

The pages should feature a clearly visible, visually distinct notice that mentions this behaviour. Another user has previously mentioned this in a [comment](https://www.php.net/manual/en/function.curl-errno.php#91518) 16 years ago, however this hint is currently buried beyond a list of error constants in the topmost comment. I believe it should be part of the documentation.

Alternatively, this could also be considered a bug, however, it would still need to be documented for past versions regardless.

Individual `CurlHandle` execution:
```php
$curlHandle = curl_init("thishostdoesnotresolve"); // Invalid host, will cause error number 6

curl_exec($curlHandle);

// Outputs "6"
echo curl_errno($curlHandle);
```

Executing a `CurlHandle` via `CurlMultiHandle`:
```php
$curlHandle = curl_init("thishostdoesnotresolve"); // Invalid host, will cause error number 6

$multiHandle = curl_multi_init();
curl_multi_add_handle($multiHandle, $curlHandle);

do {
    curl_multi_exec($multiHandle, $stillRunning);
} while ($stillRunning);

// The assumed behaviour is that the individual handles know their own errors,
// therefore, curl_errno() applied to an executed handle would return its error, if any.
// However, this outputs "0"
echo curl_errno($curlHandle);

// This is the only way to retrieve curl errors while using a CurlMultiHandle
// [result] => 6
print_r(curl_multi_info_read($multi));
```

lacatoire:
Add a note to `curl_errno()` and `curl_error()` documentation explaining that these functions always return 0 or empty string when the handle was executed via `curl_multi_exec()`. Users should use `curl_multi_info_read()` to retrieve errors for individual transfers.

Fixes #5047

### DIFF
--- a/reference/curl/functions/curl-errno.xml
+++ b/reference/curl/functions/curl-errno.xml
@@ -15,6 +15,15 @@
   <para>
    Returns the error number for the last cURL operation.
   </para>
+  <note>
+   <para>
+    When a handle has been added to a multi handle and executed via
+    <function>curl_multi_exec</function>, this function will always
+    return <literal>0</literal>. To retrieve errors for individual
+    transfers, use <function>curl_multi_info_read</function> instead
+    and check the <literal>result</literal> key of the returned array.
+   </para>
+  </note>
  </refsect1>
 
  <refsect1 role="parameters">
--- a/reference/curl/functions/curl-errno.xml
+++ b/reference/curl/functions/curl-errno.xml
@@ -15,15 +15,6 @@
   <para>
    Returns the error number for the last cURL operation.
   </para>
-  <note>
-   <para>
-    When a handle has been added to a multi handle and executed via
-    <function>curl_multi_exec</function>, this function will always
-    return <literal>0</literal>. To retrieve errors for individual
-    transfers, use <function>curl_multi_info_read</function> instead
-    and check the <literal>result</literal> key of the returned array.
-   </para>
-  </note>
  </refsect1>
 
  <refsect1 role="parameters">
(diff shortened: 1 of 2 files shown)


reference/curl/functions/curl-errno.xml
```diff
@@ -15,6 +15,15 @@
   <para>
    Returns the error number for the last cURL operation.
   </para>
+  <note>
+   <para>
```
lacatoire:
```suggestion
   <simpara>
```

reference/curl/functions/curl-errno.xml
```diff
@@ -15,6 +15,15 @@
...
+    When a handle has been added to a multi handle and executed via
+    <function>curl_multi_exec</function>, this function will always
+    return <literal>0</literal>. To retrieve errors for individual
+    transfers, use <function>curl_multi_info_read</function> instead
+    and check the <literal>result</literal> key of the returned array.
+   </para>
```
lacatoire:
```suggestion
   </simpara>
```

reference/curl/functions/curl-error.xml
```diff
@@ -15,6 +15,15 @@
   <para>
    Returns a clear text error message for the last cURL operation.
   </para>
+  <note>
+   <para>
```
lacatoire:
```suggestion
   <simpara>
```

reference/curl/functions/curl-error.xml
```diff
@@ -15,6 +15,15 @@
...
+    When a handle has been added to a multi handle and executed via
+    <function>curl_multi_exec</function>, this function will always
+    return an empty string. To retrieve errors for individual
+    transfers, use <function>curl_multi_info_read</function> instead
+    and check the <literal>result</literal> key of the returned array.
+   </para>
```
lacatoire:
```suggestion
   </simpara>
```

Issue #3339, opened 2024-04-22
Randomizer::getFloat() unexpected results in extreme edge cases


ausi:
### Description

While playing around with `Random\Randomizer::getFloat()` I noticed that it does return unexpected results in some extreme edge cases.

The following code:

```php
<?php
$r = new \Random\Randomizer();

echo "\ninterval: (-5.0E-324, +5.0E-324)";
echo "\nexpected: float(0)\nactual:   ";
var_dump($r->getFloat(-5.0E-324, +5.0E-324, \Random\IntervalBoundary::OpenOpen));
// float(-5.0E-324)

echo "\ninterval: (0.0, 1.0E-323)";
echo "\nexpected: float(5.0E-324)\nactual:   ";
var_dump($r->getFloat(0.0, 1.0E-323, \Random\IntervalBoundary::OpenOpen));
// float(-5.0E-324)

echo "\ninterval: (5.0E-324, 1.5E-323)";
echo "\nexpected: float(1.0E-323)\nactual:   ";
var_dump($r->getFloat(5.0E-324, 1.5E-323, \Random\IntervalBoundary::OpenOpen));
// float(1.5E-323)

echo "\ninterval: (0.0, 5.0E-324]";
echo "\nexpected: float(5.0E-324)\nactual:   ";
var_dump($r->getFloat(0.0, 5.0E-324, \Random\IntervalBoundary::OpenClosed));
// float(0)

echo "\ninterval: (5.0E-324, 1.0E-323]";
echo "\nexpected: float(1.0E-323)\nactual:   ";
var_dump($r->getFloat(5.0E-324, 1.0E-323, \Random\IntervalBoundary::OpenClosed));
// float(0)
```

Resulted in this output:
```
interval: (-5.0E-324, +5.0E-324)
expected: float(0)
actual:   float(-5.0E-324)

interval: (0.0, 1.0E-323)
expected: float(5.0E-324)
actual:   float(-5.0E-324)

interval: (5.0E-324, 1.5E-323)
expected: float(1.0E-323)
actual:   float(1.5E-323)

interval: (0.0, 5.0E-324]
expected: float(5.0E-324)
actual:   float(0)

interval: (5.0E-324, 1.0E-323]
expected: float(1.0E-323)
actual:   float(0)
```

These are all cases where only a single float value is possible to be drawn, but the results are consistently wrong and up to two values apart from the correct result.

[...]

PR #5240, opened 2026-02-01
Document underflow limitation in Randomizer::getFloat()

lacatoire:
## Summary                                                                                                                                                                                              
                                                                                                                                                                                                          
  Add a caution note documenting that the γ-section algorithm intentionally leaves underflow unhandled. This may result in incorrect values for intervals with boundaries in the subnormal range of       
  floating point numbers (values smaller than approximately 2^-1020).                                                                                                                                     
                                                                                                                                                                                                          
  This limitation is inherent to the algorithm as stated in the reference paper: "all our results hold in the absence of underflow."                                                                      
                                                                                                                                                                                                          
  Refs #3339

### DIFF
--- a/reference/random/random/randomizer/getfloat.xml
+++ b/reference/random/random/randomizer/getfloat.xml
@@ -346,7 +346,7 @@ Lat: +69.244304 Lng: -53.548951
     Underflow is intentionally left unhandled in the γ-section algorithm.
     This may result in incorrect values being returned for intervals with
     boundaries in the subnormal range of floating point numbers, i.e.
-    for values smaller than approximately
+    for boundaries with an absolute value smaller than approximately
     <literal>2<superscript>-1020</superscript></literal> (about <literal>8.9e-308</literal>).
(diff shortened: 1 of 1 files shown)


reference/random/random/randomizer/getfloat.xml
```diff
@@ -341,6 +341,15 @@ Lat: +69.244304 Lng: -53.548951
     to obtain the desired behavioral properties.
    </para>
   </note>
+  <caution>
+   <para>
+    Underflow is intentionally left unhandled in the γ-section algorithm.
+    This may result in incorrect values being returned for intervals with
+    boundaries in the subnormal range of floating point numbers, i.e.
+    for values smaller than approximately
```
TimWolla:
```suggestion
    for boundaries with an absolute value smaller than approximately
```
Perhaps this to make it clear that negative numbers are equally affected?

cweiske:
Thank you!

PR #5241, opened 2026-02-01
Fix forward_static_call_array() can be used outside a class

### LINKED ISSUE #3282: forward_static_call_array can be used outside of a class too
https://www.php.net/manual/en/function.forward-static-call.php
https://www.php.net/manual/en/function.forward-static-call-array.php

>This function must be called within a method context, it can't be used outside a class.

This isn't true for `forward_static_call_array` which can be used outside of a class. I'm not sure whether this is a bug/feature of PHP itself, but at least the docs aren't correct for how it works atm.
e.g. https://3v4l.org/Cb8AZ

lacatoire:
Fixes #3282

Remove incorrect statement that `forward_static_call_array()` must be called within a method context. See https://3v4l.org/Cb8AZ

### DIFF
--- a/reference/funchand/functions/forward-static-call-array.xml
+++ b/reference/funchand/functions/forward-static-call-array.xml
@@ -15,12 +15,10 @@
   </methodsynopsis>
   <para>
    Calls a user defined function or method given by the <parameter>callback</parameter>
-   parameter. This function must be called within a method context, it can't be 
-   used outside a class.
-   It uses the <link linkend="language.oop5.late-static-bindings">late static
-   binding</link>.
-   All arguments of the forwarded method are passed as values,
-   and as an array, similarly to <function>call_user_func_array</function>.
+   parameter, with arguments passed as an array, similarly to
+   <function>call_user_func_array</function>.
+   When called from within a method context, it uses the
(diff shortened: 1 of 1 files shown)


reference/funchand/functions/forward-static-call-array.xml
```diff
@@ -15,12 +15,10 @@
   </methodsynopsis>
   <para>
```
lacatoire:
```suggestion
  <simpara>
```

reference/funchand/functions/forward-static-call-array.xml
```diff
@@ -15,12 +15,10 @@
...
-   used outside a class.
-   It uses the <link linkend="language.oop5.late-static-bindings">late static
-   binding</link>.
-   All arguments of the forwarded method are passed as values,
-   and as an array, similarly to <function>call_user_func_array</function>.
+   parameter, with arguments passed as an array, similarly to
+   <function>call_user_func_array</function>.
+   When called from within a method context, it uses the
+   <link linkend="language.oop5.late-static-bindings">late static binding</link>.
   </para>
```
lacatoire:
```suggestion
  </simpara>
```

PR #5246, opened 2026-02-03
Remove descriptions of obsolete parameters not in synopsis

chschneider:
The function `imap_headerinfo` had an empty description for parameter `defaulthost`.
The function `mb_decode_numericentity` had an obsolete description for parameter `is_hex`.

Both were removed to make the parameter descriptions match the synopsis of the function.

### DIFF
--- a/reference/mbstring/functions/mb-decode-numericentity.xml
+++ b/reference/mbstring/functions/mb-decode-numericentity.xml
@@ -47,14 +47,6 @@
       &mbstring.encoding.parameter;
      </listitem>
     </varlistentry>
-    <varlistentry>
-     <term><parameter>is_hex</parameter></term>
-     <listitem>
-      <para>
-       This parameter is not used.
-      </para>
-     </listitem>
-    </varlistentry>
    </variablelist>
   </para>
  </refsect1>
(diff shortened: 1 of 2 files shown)


reference/mbstring/functions/mb-decode-numericentity.xml
```diff
@@ -47,14 +47,6 @@
...
      </listitem>
     </varlistentry>
-    <varlistentry>
-     <term><parameter>is_hex</parameter></term>
-     <listitem>
-      <para>
-       This parameter is not used.
-      </para>
-     </listitem>
-    </varlistentry>
```
Girgias, line 57: Do you know since which version? As the documentation covers PHP 7.0 to PHP 8.5.

chschneider: According to the migration guide from 7.4 to 8.0:
> The is_hex parameter, which was not used internally, has been removed from mb_decode_numericentity().

My understanding is that the parameter was a implementation-relic/copy-paste from mb_encode_numericentity and never really used, PHP 8.x complains about ArgumentCountError and I'm not sure we should keep the documentation around.

Are you thinking of someone looking at an old code base trying to figure out what the additional parameter (which not mentioned in the synopsis and leading to an error for 8.x) does?
I'm not sure this is reason enough to the above part in the documentation.

Girgias: It's more is it worth it to add a changelog entry, but considering it was never useful removing it is fine.

PR #5271, opened 2026-02-05
Fix hash functions changelog to mention E_WARNING

### LINKED ISSUE #3480: Error in Changelog for the function "hash" (php.net)
Function:
hash
Changelog:
8.0.0 hash() now throws a ValueError exception if algo is unknown; previously, false was returned instead.

It is not true. Before version 8.0.0, it issues
"Warning: hash(): Unknown hashing algorithm..."

lacatoire:
Fixes #3480

Add missing E_WARNING mention to changelog entries for hash(), hash_hmac(), and hash_hmac_file().

### DIFF
--- a/reference/hash/functions/hash.xml
+++ b/reference/hash/functions/hash.xml
@@ -89,7 +89,8 @@
        <entry>8.0.0</entry>
        <entry>
         <function>hash</function> now throws a <classname>ValueError</classname> exception
-        if <parameter>algo</parameter> is unknown; previously, &false; was returned instead.
+        if <parameter>algo</parameter> is unknown; previously, &false; was returned
+        and an <constant>E_WARNING</constant> message was emitted.
        </entry>
       </row>
      </tbody>
--- a/reference/hash/functions/hash.xml
+++ b/reference/hash/functions/hash.xml
@@ -71,10 +71,10 @@
 
  <refsect1 role="errors">
   &reftitle.errors;
-  <para>
+  <simpara>
    Throws a <classname>ValueError</classname> exception if
    <parameter>algo</parameter> is unknown.
-  </para>
+  </simpara>
  </refsect1>
 
  <refsect1 role="changelog">
--- a/reference/hash/functions/hash.xml
+++ b/reference/hash/functions/hash.xml
@@ -69,6 +69,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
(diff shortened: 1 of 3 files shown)


reference/hash/functions/hash.xml
```diff
@@ -89,7 +89,8 @@
        <entry>8.0.0</entry>
        <entry>
         <function>hash</function> now throws a <classname>ValueError</classname> exception
```
kamil-tekiela: Why is this one different than the other two? Also, the `Errors/Exceptions` section is missing.

kamil-tekiela: I think a snippet would be a good idea if it's copied 3 times exactly the same. Edit: I see that one of them is different, so you can leave it as it is.

reference/hash/functions/hash.xml
```diff
@@ -69,6 +69,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   Throws a <classname>ValueError</classname> exception if
+   <parameter>algo</parameter> is unknown.
+  </para>
```
lacatoire:
```suggestion
  <simpara>
   Throws a <classname>ValueError</classname> exception if
   <parameter>algo</parameter> is unknown.
  </simpara>
```

kamil-tekiela (approved):

reference/hash/functions/hash.xml
```diff
@@ -69,6 +69,14 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   Throws a <classname>ValueError</classname> exception if
```
jordikroon:
```suggestion
   Throws a <exceptionname>ValueError</exceptionname> exception if
```

reference/hash/functions/hash.xml
```diff
@@ -88,8 +96,9 @@
       <row>
        <entry>8.0.0</entry>
        <entry>
-        <function>hash</function> now throws a <classname>ValueError</classname> exception
-        if <parameter>algo</parameter> is unknown; previously, &false; was returned instead.
+        Now throws a <classname>ValueError</classname> exception
```
jordikroon:
```suggestion
        Now throws a <exceptionname>ValueError</exceptionname> exception
```

jordikroon (commented):
Just some minor nits. Other than that LGTM.

PR #5275, opened 2026-02-05
Clarify ldap_connect/ldap_bind/ldap_set_option behavior

### LINKED ISSUE #3526: LDAP: `ldap_connect()`/`ldap_bind()`/`ldap_set_option()` is unclear
I'm using PHP 8.2 and there are several inconsistencies in this all. First, `ldap_connect` does not actually perform connection. It basically just inits the whole thing. First connection is done by `ldap_bind`. This is actually stated [here](https://www.php.net/manual/en/function.ldap-connect.php) and turning on debugging perfectly shows this behavior. 
Said, so, it would all make sense to be able to call `ldap_set_option` on a "connection" object, however it wouldn't work (for sure it doesn't as of PHP 8.2), especially with TLS related options. But it seems many other options you could not be set if not globally. And the actual issue with documentation is there is no reference about which parameters _have to_ be specified together with a connection and which _haven't_.

_Originally posted by @QcFe in https://github.com/php/doc-en/issues/1722#issuecomment-2196703545_

lacatoire:
Clarifies that `ldap_connect()` only validates the URI and initializes parameters, while `ldap_bind()` establishes the actual TCP connection. Documents that TLS options must be set globally with `null` before bind.

Fixes #3526

### DIFF
--- a/reference/ldap/functions/ldap-connect.xml
+++ b/reference/ldap/functions/ldap-connect.xml
@@ -23,14 +23,26 @@
    <methodparam choice="opt"><type>int</type><parameter>port</parameter><initializer>389</initializer></methodparam>
   </methodsynopsis>
   <para>
-   Creates an <classname>LDAP\Connection</classname> connection and checks whether the given
-   <parameter>uri</parameter> is plausible.
+   Creates an <classname>LDAP\Connection</classname> instance and checks
+   whether the given <parameter>uri</parameter> is plausible.
   </para>
   <note>
    <simpara>
-    This function does <emphasis>not</emphasis> open a connection.
-    It checks whether the given parameters are plausible and can be used
-    to open a connection as soon as one is needed.
+    This function does <emphasis>not</emphasis> open a network connection.
+    It only validates the URI syntax and initializes connection parameters.
+    The actual TCP connection is established when
+    <function>ldap_bind</function> is called.
+   </simpara>
+  </note>
+  <note>
+   <simpara>
+    Options that affect the connection setup, such as
+    <constant>LDAP_OPT_PROTOCOL_VERSION</constant> and TLS-related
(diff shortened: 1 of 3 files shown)


reference/ldap/functions/ldap-connect.xml
```diff
@@ -23,14 +23,26 @@
...
+    It only validates the URI syntax and initializes connection parameters.
+    The actual TCP connection is established when
+    <function>ldap_bind</function> is called.
+   </simpara>
+  </note>
+  <note>
+   <simpara>
+    Options that affect the connection setup, such as
+    <constant>LDAP_OPT_PROTOCOL_VERSION</constant> and TLS-related
+    options (<constant>LDAP_OPT_X_TLS_*</constant>), must be set
```
jordikroon: This should be a literal instead.

reference/ldap/functions/ldap-set-option.xml
```diff
@@ -17,6 +17,15 @@
   <para>
    Sets the value of the specified option to be <parameter>value</parameter>.
   </para>
+  <note>
+   <simpara>
+    TLS-related options (<constant>LDAP_OPT_X_TLS_*</constant>) must be
```
jordikroon: This should be a literal.

reference/ldap/functions/ldap-connect.xml
```diff
@@ -23,14 +23,26 @@
    <methodparam choice="opt"><type>int</type><parameter>port</parameter><initializer>389</initializer></methodparam>
   </methodsynopsis>
   <para>
```
jordikroon: This should be `<simpara>`

jordikroon (changes requested):

jordikroon:
@lacatoire could you go over your (older) PRs and alter to match with the style guide? So `<para>` to `<simpara>`.

PR #5291, opened 2026-02-06
Add missing 8.5.0 changelog entry for imagecopyresampled return type

### LINKED ISSUE #5426: imagecopyresampled(): from which version does it always return true?
### Affected page

https://www.php.net/manual/en/function.imagecopyresampled.php

### Issue description

PR #5291 claims it added an entry in the changelog, but it didn’t.

/cc @lacatoire 

### Steps to reproduce

_No response_

### Suggested fix

_No response_

lacatoire:
Add missing `&return.type.true;` changelog entry for 8.5.0.

Fixes https://github.com/php/doc-en/issues/5426

### DIFF
--- a/reference/image/functions/imagecopyresampled.xml
+++ b/reference/image/functions/imagecopyresampled.xml
@@ -142,7 +142,12 @@
      </row>
     </thead>
     <tbody>
-     &return.type.true;
+     <row>
+      <entry>8.5.0</entry>
+      <entry>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
This was done only as of PHP 8.5, correct? This should be mentioned there too.

kamil-tekiela (approved):

PR #5312, opened 2026-02-06
Fix imagestring return type: bool -> true

lacatoire:
Since PHP 8.2, `imagestring()` always returns `true` instead of `bool`.

Changes:
- Update `<type>bool</type>` to `<type>true</type>` in methodsynopsis

**Reference:** [`ext/gd/gd.stub.php` line 693](https://github.com/php/php-src/blob/7fed075ba6b0431195795a7f3cc9a114a102a2e8/ext/gd/gd.stub.php#L693)

```php
function imagestring(GdImage $image, GdFont|int $font, int $x, int $y, string $string, int $color): true {}
```

### DIFF
--- a/reference/image/functions/imagestring.xml
+++ b/reference/image/functions/imagestring.xml
@@ -8,7 +8,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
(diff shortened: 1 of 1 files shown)


Girgias:

PR #5314, opened 2026-02-06
Fix finfo_close return type: bool -> true

lacatoire:
Since PHP 8.2, `finfo_close()` always returns `true` instead of `bool`.

Changes:
- Update `<type>bool</type>` to `<type>true</type>` in methodsynopsis

**Reference:** [`ext/fileinfo/fileinfo.stub.php` line 96](https://github.com/php/php-src/blob/7fed075ba6b0431195795a7f3cc9a114a102a2e8/ext/fileinfo/fileinfo.stub.php#L96)

```php
function finfo_close(finfo $finfo): true {}
```


Girgias:

PR #5321, opened 2026-02-06
Fix pg_escape_literal signature to match php-src

lacatoire:
Sync `pg_escape_literal()` documentation with php-src stub:

- Return type: `string` -> `string|false` (can return false on failure)
- Parameter name: `$data` -> `$string` (matches php-src pgsql.stub.php)

**Reference:** [`ext/pgsql/pgsql.stub.php` line 874](https://github.com/php/php-src/blob/7fed075ba6b0431195795a7f3cc9a114a102a2e8/ext/pgsql/pgsql.stub.php#L874)

```php
function pg_escape_literal($connection, string $string = UNKNOWN): string|false {}
```

### DIFF
--- a/reference/pgsql/functions/pg-escape-literal.xml
+++ b/reference/pgsql/functions/pg-escape-literal.xml
@@ -12,9 +12,9 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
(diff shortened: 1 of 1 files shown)


Girgias:

PR #5324, opened 2026-02-06
Fix pg_execute parameter name: $stmtname -> $statement_name

lacatoire:
Sync `pg_execute()` parameter name `$stmtname` -> `$statement_name` to match php-src pgsql.stub.php.

**Reference:** [`ext/pgsql/pgsql.stub.php` line 552](https://github.com/php/php-src/blob/7fed075ba6b0431195795a7f3cc9a114a102a2e8/ext/pgsql/pgsql.stub.php#L552)

```php
function pg_execute($connection, $statement_name, array $params = UNKNOWN): PgSql\Result|false {}
```


Girgias:

PR #5329, opened 2026-02-06
Fix pg_set_error_verbosity return type: int -> int|false

lacatoire:
Sync `pg_set_error_verbosity()` return type with php-src: `int` -> `int|false`.

**Reference:** [`ext/pgsql/pgsql.stub.php` line 822](https://github.com/php/php-src/blob/7fed075ba6b0431195795a7f3cc9a114a102a2e8/ext/pgsql/pgsql.stub.php#L822)

```php
function pg_set_error_verbosity($connection, int $verbosity = UNKNOWN): int|false {}
```


Girgias:

PR #5334, opened 2026-02-06
Fix PDOStatement::setFetchMode return type: bool -> true

lacatoire:
Since PHP 8.2, `PDOStatement::setFetchMode()` always returns `true` instead of `bool`.

Changes:
- Update `<type>bool</type>` to `<type>true</type>` in all 4 methodsynopsis variants

**Reference:** [`ext/pdo/pdo_stmt.stub.php` line 65](https://github.com/php/php-src/blob/7fed075ba6b0431195795a7f3cc9a114a102a2e8/ext/pdo/pdo_stmt.stub.php#L65)

```php
public function setFetchMode(int $mode, mixed ...$args): true {}
```

### DIFF
--- a/reference/pdo/pdostatement/setfetchmode.xml
+++ b/reference/pdo/pdostatement/setfetchmode.xml
@@ -11,25 +11,25 @@
   &reftitle.description;
 
   <methodsynopsis role="PDOStatement">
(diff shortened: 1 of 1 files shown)


Girgias:

PR #5350, opened 2026-02-06
Add documentation for 6 new pcntl functions (PHP 8.4)

lacatoire:
Add documentation for 6 new pcntl functions available since PHP 8.4: `pcntl_wifcontinued`, `pcntl_getcpu`, `pcntl_forkx`, `pcntl_setns`, `pcntl_getqos_class`, and `pcntl_setqos_class`.

### DIFF
--- a/reference/pcntl/functions/pcntl-getqos-class.xml
+++ b/reference/pcntl/functions/pcntl-getqos-class.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8"?>
+<refentry xml:id="function.pcntl-getqos-class" xmlns="http://docbook.org/ns/docbook">
+ <refnamediv>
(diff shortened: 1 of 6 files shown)


jordikroon (changes requested):
`<simplelist>` and  `<variablelist>` shouldn't be wrapped in a `<para>`.

reference/pcntl/functions/pcntl-getqos-class.xml
```diff
@@ -13,11 +12,10 @@
    <void/>
   </methodsynopsis>
   <simpara>
-   Retrieve the QoS class.
+   Returns the current Quality of Service (QoS) class of the calling process.
```
jordikroon: It's probably better to make `QoS` an acronym which needs to be added to doc-base.

PR #5355, opened 2026-02-06
Fix pgsql function signatures

lacatoire:
Fix return types and parameter names for pgsql functions to match php-src stubs.

### DIFF
--- a/reference/pgsql/functions/pg-escape-literal.xml
+++ b/reference/pgsql/functions/pg-escape-literal.xml
@@ -12,9 +12,9 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>string</type><methodname>pg_escape_literal</methodname>
+   <type class="union"><type>string</type><type>false</type></type><methodname>pg_escape_literal</methodname>
    <methodparam choice="opt"><type>PgSql\Connection</type><parameter>connection</parameter></methodparam>
-   <methodparam><type>string</type><parameter>data</parameter></methodparam>
+   <methodparam><type>string</type><parameter>string</parameter></methodparam>
   </methodsynopsis>
   <para>
    <function>pg_escape_literal</function> escapes a literal for
@@ -46,7 +46,7 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><parameter>data</parameter></term>
+     <term><parameter>string</parameter></term>
      <listitem>
       <para>
        A <type>string</type> containing text to be escaped.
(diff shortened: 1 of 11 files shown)


reference/pgsql/functions/pg-escape-literal.xml
```diff
@@ -12,9 +12,9 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>string</type><methodname>pg_escape_literal</methodname>
+   <type class="union"><type>string</type><type>false</type></type><methodname>pg_escape_literal</methodname>
```
kamil-tekiela, line 15: or false on failure is missing

reference/pgsql/functions/pg-lo-import.xml
```diff
@@ -12,8 +12,8 @@
   <methodsynopsis>
    <type>int|string|false</type><methodname>pg_lo_import</methodname>
```
kamil-tekiela: I think this needs to be fixed.

reference/pgsql/functions/pg-lo-import.xml
```diff
@@ -12,8 +12,8 @@
...
    <methodparam choice="opt"><type>PgSql\Connection</type><parameter>connection</parameter></methodparam>
-   <methodparam><type>string</type><parameter>pathname</parameter></methodparam>
-   <methodparam choice="opt"><type>mixed</type><parameter>object_id</parameter></methodparam>
+   <methodparam><type>string</type><parameter>filename</parameter></methodparam>
+   <methodparam choice="opt"><type>mixed</type><parameter>oid</parameter></methodparam>
```
kamil-tekiela: I don't know about these types. I don't really understand what is going on here nor what the expected type is. Perhaps, you should make a TODO GH issue to look into it.

kamil-tekiela: After being confused for a few minutes, I came to the conclusion that the type here should be a union of int and string.

reference/pgsql/functions/pg-set-error-verbosity.xml
```diff
@@ -13,7 +13,7 @@
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
-   <type>int</type><methodname>pg_set_error_verbosity</methodname>
+   <type class="union"><type>int</type><type>false</type></type><methodname>pg_set_error_verbosity</methodname>
```
kamil-tekiela, line 16: The return section is missing false.

kamil-tekiela (changes requested):

kamil-tekiela (approved):

PR #5357, opened 2026-02-06
Fix return type for 5 misc functions: bool -> true

lacatoire:
Fix return type from `bool` to `true` for finfo_close, ftp_set_option, libxml_set_external_entity_loader, shm_detach, and array_multisort to match php-src stubs.

Happy to adjust if needed.

### DIFF
--- a/reference/array/functions/array-multisort.xml
+++ b/reference/array/functions/array-multisort.xml
@@ -133,7 +133,12 @@
      </row>
     </thead>
     <tbody>
-     &return.type.true;
+     <row>
+      <entry>8.5.0</entry>
+      <entry>
+       The return type is &true; now; previously, it was <type>bool</type>.
+      </entry>
+     </row>
     </tbody>
    </tgroup>
   </informaltable>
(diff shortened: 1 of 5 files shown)


ndossche:
Don't you also need a changelog entry for this?

ndossche (changes requested):
Take for example the libxml_set_external_entity_loader() function: the changelog says it changed in PHP 8.2 but that's not true. You may have the same issue on other places too

reference/array/functions/array-multisort.xml
```diff
@@ -118,10 +118,27 @@
...
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     &return.type.true;
```
ndossche: Is the version right here? Same questino for other entries

ndossche (approved):
Thanks

PR #5360, opened 2026-02-06
Fix class method signatures

lacatoire:
Fix method signatures to match php-src stubs

### DIFF
--- a/reference/intl/intlbreakiterator/getpartsiterator.xml
+++ b/reference/intl/intlbreakiterator/getpartsiterator.xml
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis role="IntlBreakIterator">
    <modifier>public</modifier> <type>IntlPartsIterator</type><methodname>IntlBreakIterator::getPartsIterator</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>type</parameter><initializer><constant>IntlPartsIterator::KEY_SEQUENTIAL</constant></initializer></methodparam>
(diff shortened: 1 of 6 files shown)


reference/pdo/pdostatement/setfetchmode.xml
```diff
@@ -88,10 +88,27 @@
...
+  <informaltable>
+   <tgroup cols="2">
+    <thead>
+     <row>
+      <entry>&Version;</entry>
+      <entry>&Description;</entry>
+     </row>
+    </thead>
+    <tbody>
+     &return.type.true;
```
kamil-tekiela: Wasn't it all changed in PHP 8.4 though?

kamil-tekiela: You could also add a new snippet with the version number in the name.

reference/intl/intlbreakiterator/getpartsiterator.xml
```diff
@@ -10,7 +10,7 @@
   &reftitle.description;
   <methodsynopsis role="IntlBreakIterator">
    <modifier>public</modifier> <type>IntlPartsIterator</type><methodname>IntlBreakIterator::getPartsIterator</methodname>
-   <methodparam choice="opt"><type>string</type><parameter>type</parameter><initializer><constant>IntlPartsIterator::KEY_SEQUENTIAL</constant></initializer></methodparam>
+   <methodparam choice="opt"><type>int</type><parameter>type</parameter><initializer><constant>IntlPartsIterator::KEY_SEQUENTIAL</constant></initializer></methodparam>
```
jordikroon: While it's true this is fixed. This change is targeting master and will likely not be backported downstream. So for the time being  this should be a string.

jordikroon (changes requested):

PR #5361, opened 2026-02-07
constants.fetch-modes.xml Fix the call sequence

mmalferov:


kamil-tekiela:
Why? Is the output wrong as it is now?

mmalferov:
In the current version, the fetch mode is set before the `$stmt` variable is declared

kamil-tekiela (approved):
Ahh ok, I didn't see that. It's difficult to see all the details on mobile.

lacatoire (approved):

PR #5365, opened 2026-02-10
Add PHP 8.4.0 changelog entries for Intl and MBString pages

lacatoire:
Relates to #3872

Add missing 8.4.0 changelog entries for mb_substr, mb_strcut, idn_to_ascii, and idn_to_utf8.

### DIFF
--- a/reference/intl/idn/idn-to-ascii.xml
+++ b/reference/intl/idn/idn-to-ascii.xml
@@ -91,6 +91,21 @@
      </thead>
      <tbody>
       <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Now throws a <exceptionname>ValueError</exceptionname>
+        if the <parameter>domain</parameter> name is empty.
+       </entry>
+      </row>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Now throws a <exceptionname>ValueError</exceptionname>
+        if the <parameter>variant</parameter> parameter is not
+        <constant>INTL_IDNA_VARIANT_UTS46</constant>.
+       </entry>
+      </row>
+      <row>
        <entry>7.4.0</entry>
        <entry>
         The default value of <parameter>variant</parameter> is now
--- a/reference/intl/idn/idn-to-ascii.xml
+++ b/reference/intl/idn/idn-to-ascii.xml
@@ -94,7 +94,7 @@
        <entry>8.4.0</entry>
        <entry>
         Now throws a <exceptionname>ValueError</exceptionname>
-        if the <parameter>domain</parameter> name is empty.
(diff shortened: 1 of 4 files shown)


reference/intl/idn/idn-to-ascii.xml
```diff
@@ -90,6 +90,21 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Now throws a <exceptionname>ValueError</exceptionname>
+        if the <parameter>domain</parameter> name is empty.
```
jordikroon:
```suggestion
        if the <parameter>domain</parameter> parameter is empty.
```

reference/intl/idn/idn-to-utf8.xml
```diff
@@ -90,6 +90,21 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Now throws a <exceptionname>ValueError</exceptionname>
+        if the <parameter>domain</parameter> name is empty.
```
jordikroon:
```suggestion
        if the <parameter>domain</parameter> parameter is empty.
```

reference/mbstring/functions/mb-strcut.xml
```diff
@@ -107,6 +107,13 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       The behavior of <function>mb_strcut</function> is more
```
jordikroon: We don't need to mention `mb_strcut` again here. Additionally UTF-8 and UTF-16 should become literals.

```xml
The behavior is more consistent now on invalid <literal>UTF-8</literal> and <literal>UTF-16</literal> strings.
```

reference/mbstring/functions/mb-substr.xml
```diff
@@ -97,6 +97,17 @@
...
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       On invalid strings (those with encoding errors),
+       <function>mb_substr</function> now interprets character
+       indices in the same manner as most other mbstring functions.
+       This means that character indices returned by
+       <function>mb_strpos</function> can be passed to
+       <function>mb_substr</function>.
+      </entry>
```
jordikroon, line 107: Same here. It's better to just use:

```xml
<entry>
 On invalid strings (those with encoding errors),
 character indices are now interpreted in the same manner as most
 other mbstring functions. This means that character indices returned
 by <function>mb_strpos</function> can be passed directly.
</entry>
```

jordikroon (changes requested):

PR #5366, opened 2026-02-10
Add PHP 8.4.0 changelog entries for Session and LDAP pages

lacatoire:
Relates to #3872

Add missing 8.4.0 changelog entries for session_set_save_handler and ldap_connect.

### DIFF
--- a/reference/ldap/functions/ldap-connect.xml
+++ b/reference/ldap/functions/ldap-connect.xml
@@ -104,6 +104,14 @@
     </thead>
     <tbody>
      <row>
+      <entry>8.4.0</entry>
+      <entry>
+       Calling <function>ldap_connect</function> with more than
+       two arguments is now deprecated.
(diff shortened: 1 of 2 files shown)


reference/ldap/functions/ldap-connect.xml
```diff
@@ -103,6 +103,14 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.4.0</entry>
+      <entry>
+       Calling <function>ldap_connect</function> with more than
+       two arguments is now deprecated.
+       Use <function>ldap_connect_wallet</function> instead.
```
jordikroon: `ldap_connect_wallet` is deprecated since PHP 8.5. So I suggest we drop this suggestion.

jordikroon (changes requested):

PR #5367, opened 2026-02-10
Add PHP 8.4.0 changelog entries for SPL, Standard, and Filesystem pages

lacatoire:
Helps with #3872

Add missing 8.4.0 changelog entries for SplFixedArray, debug_zval_dump, tempnam, and unserialize.

### DIFF
--- a/reference/var/functions/unserialize.xml
+++ b/reference/var/functions/unserialize.xml
@@ -173,6 +173,13 @@
        </entry>
       </row>
       <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Unserializing strings using the uppercase <literal>S</literal> tag
+        is now deprecated.
+       </entry>
+      </row>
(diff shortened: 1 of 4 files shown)


reference/var/functions/unserialize.xml
```diff
@@ -172,6 +172,13 @@
         element of <parameter>options</parameter> is not an <type>array</type> of class names.
        </entry>
       </row>
+      <row>
+       <entry>8.4.0</entry>
+       <entry>
+        Unserializing strings using the uppercase <literal>S</literal> tag
+        is now deprecated.
```
jordikroon: We can suggest to use lowercase `s` instead. Also I feel we need to quote `"S"` here.

jordikroon (changes requested):
Just a small nit to improve developer experience.

PR #5371, opened 2026-02-11
Since PHP 8.5 semicolon is deprecated to terminate case in switch

janmazanek:
According to https://www.php.net/releases/8.5/en.php#features: Terminating case statements with a semicolon instead of a colon has been deprecated.


lacatoire:
Thanks @janmazanek for the contribution! However, this appears to be a duplicate of #4895 which covers the same change with proper DocBook formatting and a changelog entry. I'd suggest closing this in favor of that PR.

PR #5377, opened 2026-02-17
FAQ: Mention named arguments and discourage manual escaping

Dustaboy3:
Updates the FAQ:

- Mentions named arguments in the parameter order entry.
- Adds a note discouraging manual escaping in the addslashes entry, recommending prepared statements and parameter binding instead.


Dustaboy3:
Thanks for the review! Good catch - updated the wording

PR #5380, opened 2026-02-17
Clarify how multi_query works

kamil-tekiela:
Supersedes #4782


kamil-tekiela:
> This is a good improvement. I just wouldn't remove the only use of "asynchronous" in this article. How about...
> 
> "MySQL server will continue processing the remaining queries asynchronously of PHP"

I decided to remove it because it is ambiguous. It could imply several things and I think it's just not needed here at all. But I guess I could add it as you suggest.

PR #5383, opened 2026-02-18
Fix outdated content on the HTTP Basic auth page

narfbg:
It's quite self-explanatory, but feedback welcome on better formatting or whatever.

### DIFF
--- a/features/http-auth.xml
+++ b/features/http-auth.xml
@@ -29,8 +29,8 @@
 <![CDATA[
 <?php
 if (!isset($_SERVER['PHP_AUTH_USER'])) {
+    header('HTTP/1.1 401 Unauthorized');
     header('WWW-Authenticate: Basic realm="My Realm"');
-    header('HTTP/1.0 401 Unauthorized');
     echo 'Text to send if user hits Cancel button';
     exit;
 } else {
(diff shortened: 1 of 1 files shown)


features/http-auth.xml
```diff
@@ -29,8 +29,8 @@
 <![CDATA[
 <?php
 if (!isset($_SERVER['PHP_AUTH_USER'])) {
+    header('HTTP/1.1 401 Unauthorized');
```
cweiske, line 32: I don't think it's necessary to send 1.1 instead of 1.0.

narfbg: HTTP 1.1 is the first formally standardized version of the protocol, while 1.0 is largely obsolete and not used in any serious capacity today. Gaps and ambiguities in 1.0 are largely what led to the wider problems I'm trying to address with this pull request.

With that in mind, there are very valid reasons to avoid 1.0 and defaulting to version 1.1 is more than appropriate. Not just here, but in all code examples. It's softly nudging people in the right direction.

cweiske: I'd say that an [RFC 1945](https://www.rfc-editor.org/rfc/rfc1945) is a formal standardization.

narfbg: In a more conversational sense, sure, but it is categorized as "Informational" instead of "Standards Track" like [RFC2068](https://datatracker.ietf.org/doc/html/rfc2068) and its further updates. MDN also [explicitly](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Evolution_of_HTTP#http1.1_%E2%80%93_the_standardized_protocol) refers to 1.1 as the standardized protocol (the article outlines important context quite well too).

But even if you were to discount that as just semantics, my remaining arguments still stand.

PR #5394, opened 2026-02-27
Declare the namespace of ext/random classes

kocsismate:
Based on https://github.com/php/phd/issues/194. Generated with https://github.com/php/php-src/pull/21313


kocsismate:
Maybe I got the syntax wrong? The build fails (https://github.com/php/doc-en/actions/runs/22478766074/job/65111529831?pr=5394).

PR #5409, opened 2026-03-06
ext/pgsql: Update outdated PostgreSQL version requirements

KentarouTakeda:
## Summary

The pgsql extension documentation references PostgreSQL versions that have been outdated for years. The requirements page (`reference/pgsql/setup.xml`) states:

https://www.php.net/manual/en/pgsql.requirements.php

> To use PostgreSQL support, you need PostgreSQL 6.5 or later,
> PostgreSQL 8.0 or later to enable all PostgreSQL module features.

However, the actual minimum libpq versions enforced by php-src are:

| PHP version | Minimum libpq | Commit |
|---|---|---|
| PHP 8.0+ | libpq 9.1 | php/php-src@ce668c0ec6 |
| PHP 8.4+ | libpq 10.0 | php/php-src#14628 |

The "PostgreSQL 6.5" requirement has been incorrect since at least PHP 8.0.0 (released 2020-11-26), which requires libpq 9.1 via `PQlibVersion` check in `ext/pgsql/config.m4`. As of PHP 8.4, the minimum was further raised to libpq 10.0 via `PQencryptPasswordConn` check in `build/php.m4`.

## Changes

- **setup.xml**: Update requirements to match php-src, following the same pattern used by ext/curl and ext/openssl (per-PHP-version listing).
- **14 function pages**: Remove notes referencing PostgreSQL versions 6.3–9.0 that are no longer relevant given the libpq 10.0 minimum. These include version-gated notes for `pg_prepare`, `pg_query_params`, `pg_execute`, `pg_escape_string`, `pg_escape_identifier`, `pg_escape_literal`, `pg_client_encoding`, `pg_unescape_bytea`, `pg_result_error_field`, `pg_parameter_status`, `pg_lo_create`, `pg_version`, `pg_affected_rows`, and the general `reference.xml`.

### DIFF
--- a/reference/pgsql/functions/pg-affected-rows.xml
+++ b/reference/pgsql/functions/pg-affected-rows.xml
@@ -20,8 +20,7 @@
    queries. 
   </para>
   <para>
-   Since PostgreSQL 9.0 and above, the server returns the number of 
-   SELECTed rows. Older PostgreSQL return 0 for SELECT.
+   The server also returns the number of SELECTed rows.
   </para>
   <note>
    <para>
--- a/reference/pgsql/functions/pg-client-encoding.xml
+++ b/reference/pgsql/functions/pg-client-encoding.xml
@@ -24,8 +24,7 @@
   </para>
   <note>
    <para>
-    This function requires PostgreSQL 7.0 or
-    higher. If libpq is compiled without multibyte encoding support,
+    If libpq is compiled without multibyte encoding support,
     <function>pg_client_encoding</function> always returns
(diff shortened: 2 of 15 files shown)


KentarouTakeda:
PHP 7.x had no explicit minimum libpq version — it used feature detection via `AC_CHECK_LIB` rather than version enforcement, so I documented only the explicit requirements introduced in PHP 8.0.

I also considered whether the per-function libpq version notes should be retained, but concluded they are unnecessary — the most recent one (pg_affected_rows, PostgreSQL 9.0) dates back to 2010, well below the minimum for any supported PHP version.

PR #5412, opened 2026-03-06
Fix gh-5270: Clarify request_parse_body behavior with consumed body

### LINKED ISSUE #5270: Request_parse_body returns nothing when something has accessed php://input stream before it
From manual page: https://php.net/function.request-parse-body

---

$_POST array is always empty even it there is correct payload in request body:

```<?php
if ($_SERVER['REQUEST_METHOD'] === 'PUT' && intval($_SERVER['CONTENT_LENGTH']) > 0 && str_starts_with($_SERVER['CONTENT_TYPE'], 'application/x-www-form-urlencoded')) {
    $backup = file_get_contents('php://input');
    [$_POST,$_FILES] = request_parse_body();
}
```

jordikroon:
Fixes #5270

### DIFF
--- a/reference/network/functions/request-parse-body.xml
+++ b/reference/network/functions/request-parse-body.xml
@@ -38,7 +38,8 @@
   <caution>
    <simpara>
     <function>request_parse_body</function> consumes the request body without
-    buffering it to the <literal>php://input</literal> stream.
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
@lacatoire Could you please open an alternative PR and we will merge that instead. Thanks.

PR #5418, opened 2026-03-09
Fix GH-1132: Document mb_strlen invalid encoding behavior change in P…

### LINKED ISSUE #1132: Providing an incorrect encoding to mb_strlen returns false before PHP 8.0
From manual page: https://php.net/function.mb-strlen

```
<?php var_dump(@mb_strlen('', ''));
```

The documentation should either note what the behavior was in versions prior to PHP 8.0 or how it changed starting with PHP 8.0.

GabrielCordeiroBarrosoTeles:
Fixes #1132

Documents that passing an invalid encoding to `mb_strlen()` had undefined behavior before PHP 8.0 (could return `false`) and that as of PHP 8.0 a `ValueError` is thrown instead, as suggested in the issue discussion.

### DIFF
--- a/reference/mbstring/functions/mb-strlen.xml
+++ b/reference/mbstring/functions/mb-strlen.xml
@@ -70,6 +70,14 @@
     </thead>
     <tbody>
      &mbstring.changelog.encoding-nullable;
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       Passing an invalid encoding now throws a
+       <classname>ValueError</classname>. Previously, the return value was
+       undefined (could return &false;).
+      </entry>
+     </row>
     </tbody>
    </tgroup>
   </informaltable>
--- a/reference/mbstring/functions/mb-strlen.xml
+++ b/reference/mbstring/functions/mb-strlen.xml
@@ -74,8 +74,8 @@
       <entry>8.0.0</entry>
       <entry>
        Passing an invalid encoding now throws a
-       <classname>ValueError</classname>. Previously, the return value was
-       undefined (could return &false;).
+       <exceptionname>ValueError</exceptionname>. Previously, &false; was
+       returned.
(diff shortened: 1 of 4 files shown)


reference/mbstring/functions/mb-strlen.xml
```diff
@@ -70,6 +70,14 @@
     </thead>
     <tbody>
      &mbstring.changelog.encoding-nullable;
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       Passing an invalid encoding now throws a
+       <classname>ValueError</classname>. Previously, the return value was
```
jordikroon: Please use `<exceptionname>ValueError</exceptionname>` instead. 

And `Previously, &false; was returned.` makes more sense here since prior to 8.0, false was always returned on error.

https://github.com/php/php-src/blob/PHP-7.4.33/ext/mbstring/mbstring.c#L2503

GabrielCordeiroBarrosoTeles: Done, @jordikroon . Switched to `<exceptionname>ValueError</exceptionname>` and updated to "`Previously, &false; was returned.`" to match the PHP 7.4 source. Thanks.

kamil-tekiela:
@lacatoire is right. This change was done to multiple functions and such such should be documented consistently. https://github.com/php/php-src/commit/90eeca2531296648a264f3fdc9a3e7e8e281b24e#diff-28e30bf65ac518b331e0315af123e060f35e9053d7c05b68b5bcbcd1be080ba4 

Since the change was already documented in one place, all you need to do is take it out and put it into a new snippet (language-snippets.ent) and then use that snippet for all the changed functions. Both the `Errors/Exceptions` and `Changelog` sections need to be adjusted.

kamil-tekiela (approved):

PR #5420, opened 2026-03-10
openssl_sign()'s new $padding parameter needs to be OPENSSL_PKCS1_PSS_PADDING for PSS padding to work

terrafrost:
Per https://www.php.net/manual/en/function.openssl-sign.php PHP 8.5.0 saw the addition of a new parameter to `openssl_sign()` - `$padding`. The documentation says, simply, "_RSA PSS padding to use_". From that one might think simply setting `$padding` to 1 would be sufficient but it isn't - it needs to be `OPENSSL_PKCS1_PSS_PADDING` for PSS signatures to work.

That said, `OPENSSL_PKCS1_PSS_PADDING` does not appear to be mentioned anywhere else. https://www.php.net/manual/en/openssl.padding.php doesn't seem appropriate because that's "Padding flags for asymmetric encryption". For that I suppose there are two things that could be done. Either that could be renamed "Padding flags" or a whole new page - "Padding flags for signatures" - could be created. I suppose a third option would be to add `OPENSSL_PKCS1_PSS_PADDING` to https://www.php.net/manual/en/openssl.constants.other.php . That page already has `OPENSSL_ZERO_PADDING` after all...

### DIFF
--- a/reference/openssl/functions/openssl-sign.xml
+++ b/reference/openssl/functions/openssl-sign.xml
@@ -71,7 +71,10 @@
     <varlistentry>
      <term><parameter>padding</parameter></term>
      <listitem>
-      <simpara>RSA PSS padding to use.</simpara>
+      <para>
+       <parameter>padding</parameter> only supports one option:
+       <constant>OPENSSL_PKCS1_PSS_PADDING</constant>.
+      </para>
      </listitem>
     </varlistentry>
    </variablelist>


reference/openssl/functions/openssl-sign.xml
```diff
@@ -71,7 +71,10 @@
     <varlistentry>
      <term><parameter>padding</parameter></term>
      <listitem>
-      <simpara>RSA PSS padding to use.</simpara>
+      <para>
```
jordikroon, line 74:
```suggestion
      <simpara>
```

reference/openssl/functions/openssl-sign.xml
```diff
@@ -71,7 +71,10 @@
...
+       <parameter>padding</parameter> only supports one option:
+       <constant>OPENSSL_PKCS1_PSS_PADDING</constant>.
+      </para>
```
jordikroon, line 77:
```suggestion
      </simpara>
```

jordikroon (commented):
It's not necessary to change to `<para>` here.

reference/openssl/functions/openssl-sign.xml
```diff
@@ -71,7 +71,10 @@
     <varlistentry>
      <term><parameter>padding</parameter></term>
      <listitem>
-      <simpara>RSA PSS padding to use.</simpara>
+      <para>
+       <parameter>padding</parameter> only supports one option:
+       <constant>OPENSSL_PKCS1_PSS_PADDING</constant>.
```
lacatoire, line 76:
```suggestion
       Controls RSA padding. If unspecified (<literal>0</literal>), the default RSA PKCS#1 v1.5 padding is used.
       Set to <constant>OPENSSL_PKCS1_PSS_PADDING</constant> to enable PSS.
```

lacatoire: `php_openssl_setup_rsa_padding()` accept 3 values (0 / `RSA_PKCS1_PADDING `/ `RSA_PKCS1_PSS_PADDING`). [Here](https://github.com/php/php-src/blob/c5061420444001237bfd4d8cd921fb7ff0a9fcc8/ext/openssl/openssl.c#L4045-L4067)

lacatoire:
Covered by #5758

PR #5428, opened 2026-03-18
stream_select: Fix explanation of the microseconds behavior

divinity76:
it can be 0 or null or unspecified without blocking.

### DIFF
--- a/reference/stream/functions/stream-select.xml
+++ b/reference/stream/functions/stream-select.xml
@@ -77,7 +77,7 @@
        The <parameter>timeout</parameter> is an upper bound on the amount of time
        that <function>stream_select</function> will wait before it returns.
        If <parameter>seconds</parameter> is set to <literal>0</literal> and
-       <parameter>microseconds</parameter> is omitted, <literal>0</literal>, or &null;,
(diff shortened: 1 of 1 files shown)


reference/stream/functions/stream-select.xml
```diff
@@ -76,10 +76,10 @@
        <parameter>microseconds</parameter> the number of microseconds.
        The <parameter>timeout</parameter> is an upper bound on the amount of time
        that <function>stream_select</function> will wait before it returns.
-       If <parameter>seconds</parameter> and <parameter>microseconds</parameter> are
-       both set to <literal>0</literal>, <function>stream_select</function> will
-       not wait for data - instead it will return immediately, indicating the
-       current status of the streams.
+       If <parameter>seconds</parameter> is set to <literal>0</literal> and
+       <parameter>microseconds</parameter> is omitted, <literal>0</literal>, or &null;,
```
TimWolla: I don't think mentioning “omitted” here is useful, because that effectively means we are explaining how default values work (because “omitted” is equivalent to `null`).


```suggestion
       <parameter>microseconds</parameter> is <literal>0</literal> or &null;,
```

TimWolla (approved):

PR #5429, opened 2026-03-18
Include warnings for disable_functions and disable_classes

iluuu1994:
Added warnings about security measures for shared hosting environments.

### DIFF
--- a/appendices/ini.core.xml
+++ b/appendices/ini.core.xml
@@ -221,10 +221,10 @@
         It cannot be set in &httpd.conf;.
        </simpara>
        <warning>
-        <para>
+        <simpara>
          This directive can be circumvented and should not be considered a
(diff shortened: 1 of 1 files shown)


appendices/ini.core.xml
```diff
@@ -220,6 +220,12 @@
         This directive must be set in &php.ini;.
         It cannot be set in &httpd.conf;.
        </simpara>
+       <warning>
+        <para>
```
TimWolla: My understanding is that this should be `<simpara>` for all new additions.

```suggestion
        <simpara>
```
ditto below (and for the closing tag).

iluuu1994:
I find this message a bit more explicit that the directive is entirely ineffective. But I don't object to using the other one.

PR #5434, opened 2026-03-21
Document each global ext/dom constant on its own row

kocsismate:
Before these changes, the `gen_stub.php --verify-manual` command used to fail with an exception due to ext/DOM constants being wrongly formatted: the old and new constant counterparts were documented in the same row (e.g. `DOM_VALIDATION_ERR` / `Dom\VALIDATION_ERR`) which format cannot be recognized by gen_stub.php. Instead, let's document each constant on its own row.

Related to https://github.com/php/php-src/pull/21484

### DIFF
--- a/reference/dom/constants.xml
+++ b/reference/dom/constants.xml
@@ -251,7 +251,7 @@
      </row>
     </thead>
     <tbody>
-     <row xml:id="constant.dom-html-no-default-ns">
+     <row xml:id="constant.-dom-html-no-default-ns">
       <entry>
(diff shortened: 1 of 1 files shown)


jordikroon:
Just a quick note. This seems to be a duplicate of #4296

kocsismate:
Oh shit, I completely forgot about the original one. I'm closing both of them now based on our conversation in https://github.com/php/doc-en/pull/5516#issuecomment-4398149617.

PR #5438, opened 2026-03-23
Clarify php.ini search order and restructure Windows registry details

joshtrichards:
This  revises the "configuration file" section of `install/ini.xml` to make the `php.ini` search logic clearer:

- Replace the flat unordered list with a numbered list reflecting the actual two-phase logic
- Move the `php-SAPI.ini` vs `php.ini` fallback into the search order where it logically belongs
- Extract Windows registry key details into a dedicated `<sect2>` to declutter the main flow
- Wrap the existing environment variable / syntax content in a new `<sect2>` to fix mixed sectioned/non-sectioned siblings
- Move the Apache root directory `<note>` into the CWD list item it relates to
- Add a brief mention of `PHP_INI_SCAN_DIR` / `--with-config-file-scan-dir` in the high-level flow
- Fix `WOW6421Node` / `WOW6432Node` typo and malformed brackets in registry key references

Related: php/php-src#21503

### DIFF
--- a/install/ini.xml
+++ b/install/ini.xml
@@ -6,81 +6,76 @@
  <sect1 xml:id="configuration.file">
   <title>The configuration file</title>
 
-  <simpara>
-   The configuration file (&php.ini;)
-   is read when PHP starts up. For the server module versions of PHP,
-   this happens only once when the web server is started. For the
-   <acronym>CGI</acronym> and <acronym>CLI</acronym> versions, it happens on
-   every invocation.
-  </simpara>
   <para>
-   &php.ini; is searched for in these locations (in order):
-   <itemizedlist>
-    <listitem>
-     <simpara>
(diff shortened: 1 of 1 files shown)


install/ini.xml
```diff
@@ -6,119 +6,192 @@
...
-   defined.
-  </simpara>
-  <example>
-   <title>&php.ini; Environment Variables</title>
-    <programlisting role="ini">
+
+   <para>
+    The exact registry keys examined depend on the PHP version and whether
+    PHP and Windows are 32-bit or 64-bit.
+   </para>
```
lacatoire, line 97:
```suggestion
   <simpara>
    The exact registry keys examined depend on the PHP version and whether
    PHP and Windows are 32-bit or 64-bit.
   </simpara>
```

Issue #5155, opened 2026-01-18
QA: Meta-issue: Bulk XML markup fixes (wrapping `<para>` and conversion to `<simpara>` tags)


Girgias:
The PHP documentation XML sources have long suffered from inconsistencies and differing way of using DocBook markup.

We have established recent-ish guidelines on how things should be marked in the style guide but large parts of the manual don't follow it. This makes it harder for new contributors to actually get PRs approved from the get-go as they most likely will copy and paste a nearby XML file.

However, just blanket fixes to doc-en are a no-go as this would affect translations and their ability to track which changes are actually important.

I have created some scripts (that I will at some point add to doc-base) that can perform some XML updates and automatically apply it on translations, and update the REVCHECK hash.
The primary downside of this approach is that this can only be done when **all** translations for a specific file are up-to-date. The state of translations can be determined by going to https://doc.php.net/ and checking each-language individually.

The purpose of this issue is for me to be able to track which folders have been checked and migrated and if not, what is blocking them.

- [ ] `appendices`: Need to examine translation status
- [ ] `chapters`: Need to examine translation status
- [ ] `chmonly`: Need to examine translation status
- [ ] `faq`: Need to examine translation status
- [ ] `features`: Need to examine translation status
- [ ] `install`: Need to examine translation status
- [ ] `language`: Need to examine translation status
- [ ] `reference`:
  - [ ] `apache`: Blocked by IT/RO/TR translations not up to date
  - [x] `apcu`: Performed wrapping para & conversion to simpara scripts
  - [ ] `array`: Blocked by DE/IT/PL/RO/RU/TR/UK translations not up to date
  - [ ] `bc`: Blocked by FR/IT/RU translations not up to date
  - [ ] `bzip2`: Blocked by IT translations not up to date
  - [ ] `calendar`: Blocked by IT/TR translations not up to date
  - [ ] `classobj`: Blocked by IT/PL/TR translations not up to date
  - [x] `cmark`: Performed wrapping para & conversion to simpara scripts
  - [ ] `com` Blocked by IT/TR translations not up to date
  - [x] `componere`: Performed wrapping para & conversion to simpara scripts
  - [ ] `ctype`: Blocked by IT/RO translations not up to date
  - [x] `cubrid`: Performed wrapping para & conversion to simpara scripts
  - [ ] `curl`: Blocked by DE/IT/PL/RU/TR/UK/ZH translations not up to date
- [ ] `security`: Need to examine translation status

PR #5445, opened 2026-03-29
Add CI warning for xml:id before xmlns attribute order

lacatoire:
Following the great work tracked in #5155 to clean up XML markup inconsistencies, I thought it might be helpful to add a small CI check that gently warns contributors when `xml:id` is placed before `xmlns` declarations, which is the most common pattern to fix.
Just wanted to propose it as a small step forward.

### DIFF
--- a/.github/workflows/check-attr-order.yml
+++ b/.github/workflows/check-attr-order.yml
@@ -34,6 +34,6 @@ jobs:
             [ -z "$file" ] && continue
             [ ! -f "$file" ] && continue
             grep -n 'xml:id="[^"]*".*xmlns=' "$file" | while IFS=: read -r line _; do
-              echo "::warning file=$file,line=$line::xml:id should be placed after xmlns declarations (see docs/style.md)"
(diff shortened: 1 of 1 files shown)


.github/workflows/check-attr-order.yml
```diff
@@ -0,0 +1,39 @@
...
+          echo "EOF" >> "$GITHUB_OUTPUT"
+
+      - name: Check xml:id before xmlns
+        if: steps.changed.outputs.files != ''
+        run: |
+          while IFS= read -r file; do
+            [ -z "$file" ] && continue
+            [ ! -f "$file" ] && continue
+            grep -n 'xml:id="[^"]*".*xmlns=' "$file" | while IFS=: read -r line _; do
+              echo "::warning file=$file,line=$line::xml:id should be placed after xmlns declarations (see docs/style.md)"
```
jordikroon: docs/style.md lives inside doc-base. Since not every contributor know where these docs live, it would be better to link to https://doc.php.net/guide/style.md instead.

PR #5447, opened 2026-03-31
json-last-error.xml Move `Available as of` to *Availability* column t…

mmalferov:
…o consistency

### DIFF
--- a/reference/json/functions/json-last-error.xml
+++ b/reference/json/functions/json-last-error.xml
@@ -31,7 +31,7 @@
   </para>
   <table>
    <title>JSON error codes</title>
-   <tgroup cols="2">
+   <tgroup cols="3">
     <thead>
      <row>
       <entry>Constant</entry>
--- a/reference/json/functions/json-last-error.xml
+++ b/reference/json/functions/json-last-error.xml
@@ -102,8 +102,8 @@
      </row>
      <row>
       <entry><constant>JSON_ERROR_NON_BACKED_ENUM</constant></entry>
-      <entry>Value contains a non-backed enum which cannot be serialized. Available as of PHP 8.1.0.</entry>
-      <entry></entry>
+      <entry>Value contains a non-backed enum which cannot be serialized</entry>
+      <entry>Available as of PHP 8.1.0</entry>
      </row>
     </tbody>
    </tgroup>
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Wouldn't it be better to just remove the last column? It should be documented with only the version number as per https://github.com/php/doc-en/commit/c65729614f8c01245e58a1141633f94843ebdef9, but IMHO it is better without the last column.

reference/json/functions/json-last-error.xml
lacatoire:
```suggestion
   <tgroup cols="3">
```

jordikroon:
I agree with Kamil. We should instead create a Changelog entry. While we are doing this I suggest to change this into a `<variablelist>` as well.

reference/json/functions/json-last-error.xml
lacatoire, line 110:
```suggestion
 </refsect1>

 <refsect1 role="changelog">
  &reftitle.changelog;
  <informaltable>
   <tgroup cols="2">
    <thead>
     <row>
      <entry>&Version;</entry>
      <entry>&Description;</entry>
     </row>
    </thead>
    <tbody>
     <row>
      <entry>8.1.0</entry>
      <entry>
       <constant>JSON_ERROR_NON_BACKED_ENUM</constant> was added.
      </entry>
     </row>
    </tbody>
   </tgroup>
  </informaltable>
 </refsect1>
```

reference/json/functions/json-last-error.xml
```diff
@@ -31,7 +31,7 @@
   </para>
   <table>
    <title>JSON error codes</title>
-   <tgroup cols="2">
+   <tgroup cols="3">
```
lacatoire:
```suggestion
   <tgroup cols="2">
```

reference/json/functions/json-last-error.xml
```diff
@@ -102,8 +102,8 @@
      </row>
      <row>
       <entry><constant>JSON_ERROR_NON_BACKED_ENUM</constant></entry>
-      <entry>Value contains a non-backed enum which cannot be serialized. Available as of PHP 8.1.0.</entry>
-      <entry></entry>
+      <entry>Value contains a non-backed enum which cannot be serialized</entry>
+      <entry>Available as of PHP 8.1.0</entry>
```
lacatoire:
```suggestion
```

PR #5452, opened 2026-03-31
ini.fatal-error-backtraces

hakre:
The ini setting so far was only mentioned in the migration guide (new in 8.5).

refs https://github.com/php/doc-en/issues/4886

### DIFF
--- a/reference/errorfunc/ini.xml
+++ b/reference/errorfunc/ini.xml
@@ -23,6 +23,14 @@
      <entry></entry>
     </row>
     <row>
+     <entry><link linkend="ini.fatal-error-backtraces">fatal_error_backtraces</link></entry>
+     <entry>"1"</entry>
+     <entry><constant>INI_ALL</constant></entry>
+     <entry>
+      Available as of PHP 8.5.0
+     </entry>
+    </row>
+    <row>
      <entry><link linkend="ini.display-errors">display_errors</link></entry>
      <entry>"1"</entry>
      <entry><constant>INI_ALL</constant></entry>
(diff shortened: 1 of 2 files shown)


DanielEScherzer:
No idea if this should be simpara or para, the entries above and below it use para - @Girgias since I know you have been active with the xml, is there documentation about when to use which?

DanielEScherzer (approved):
Thanks - I'll merge this in a few days if there are no objections (please ping me if I forget)

TimWolla (approved):

reference/errorfunc/ini.xml
```diff
@@ -198,6 +206,26 @@
...
+      Fatal errors are <constant>E_ERROR</constant>,
+      <constant>E_CORE_ERROR</constant>, <constant>E_COMPILE_ERROR</constant>,
+      <constant>E_USER_ERROR</constant>,
+      <constant>E_RECOVERABLE_ERROR</constant>, and
+      <constant>E_PARSE</constant>.
+     </simpara>
+     <simpara>
+      The default value is <literal>"1"</literal>.
+     </simpara>
+    </listitem>
```
TimWolla:
```suggestion
    </listitem>
```
It seems this `error_reporting` is the only one that mentions it in the text, because the handling of its default value is somewhat special. The others just use the table above.

TimWolla: I would commit the change myself, but it seems you didn't enable the “allow maintainers to push into the branch” option.

hakre: Thank you for your feedback, I'm not sure I understand. How I'm receiving it, the default value should not be documented under the entry of the directive because it is not special enough (albeit being default), and despite documenting the value range?

That would appear lazy to me, and in my view collides with what I'd consider a readers expectation under the pretext that PHP directives have default values, to find it documented at the most detailed place in the reference, without the sportive activity jumping across a page to consult a table, then through rows and cells.

And isn't it most likely that the rest of the documentation is also on a best-effort basis and therefore the table is most complete, furthermore with default values, but the detailed entries are of varying degrees and that is most likely the reason, and not a seemingly special case of the one directive you mentioned?

Or is it that it would make the other entries look bad, if seemingly two entries stand out, and therefore you want to prevent that?

Please clarify and also provide the best reference to the in-tree docs you know about, so that I can orient myself on that, too. @TimWolla

hakre: Thanks, applied.

DanielEScherzer:
@TimWolla does this look okay now?

hakre:
rebased to resolve the conflict, no content changes.

lacatoire (approved):

jordikroon (approved):

jordikroon:
Thank you @hakre! Sorry this took long to merge.

hakre:
> Thank you @hakre! Sorry this took long to merge.

No sweat, thank you for taking care! @jordikroon

PR #5455, opened 2026-04-02
Enable runnable (WASM) examples for language.variables section

AllenJB:
Related: #4799 

Line endings, `var_dump()` or function calls have been added as appropriate to actually run the code and provide useful output.

Some example output code was rearranged so the WASM output replaces it as expected.

Additional changes:

* Removing `?>` where not required
* Added a note to the invalid reference assignment examples to describe actual behavior 

I'm wondering if there are some alternative [examples of invalid assignment](https://www.php.net/manual/en/language.variables.basics.php#:~:text=only%20variables%20may%20be%20assigned%20by%20reference) that aren't just outright (and "unrelated") syntax errors.

### DIFF
--- a/language/variables.xml
+++ b/language/variables.xml
@@ -499,7 +499,7 @@ function test()
     static $count = 0;
 
     $count++;
-    echo $count . "\n";
+    echo $count . PHP_EOL;
     if ($count < 10) {
         test();
     }
--- a/language/variables.xml
+++ b/language/variables.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
- <chapter xml:id="language.variables" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
+ <chapter xml:id="language.variables" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink"
+  annotations="interactive">
   <title>Variables</title>
   
   <sect1 xml:id="language.variables.basics">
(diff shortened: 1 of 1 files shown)


language/variables.xml
```diff
@@ -499,13 +499,14 @@ function test()
     static $count = 0;
 
     $count++;
-    echo $count;
+    echo $count . "\n";
```
jordikroon: This should be PHP_EOL in order to be consistent with the other examples.

language/variables.xml
```diff
@@ -43,7 +44,7 @@
 
    <example>
     <title>Valid variable names</title>
-    <programlisting role="php">
+    <programlisting role="php" annotations="non-interactive">
```
jordikroon, line 47: Is there a specific reason why this is marked as `non-interactive`? To me a valid example seems more useful than examples that generate a syntax error.

language/variables.xml
```diff
@@ -139,17 +137,17 @@ echo $foo;                 // $foo is altered too.
...
 
 function test()
 {
    return 25;
 }
 
-$bar = &test();    // Invalid because test() doesn't return a variable by reference.
-?>
+// Note: A notice is issued, but the non-reference value is assigned
+$ref3 = &test();    // Invalid because test() doesn't return a variable by reference.
```
jordikroon: This function never executed because the above expression gives a syntax error. However this will return a notice. It's best to move this to its own example.

Notice: Only variables should be assigned by reference in script on line 9

jordikroon (changes requested):

language/variables.xml
AllenJB: The example is demonstrating valid variable names. At least half the example gives no useful output and I thought adding extra output here might dilute the purpose of the example.

language/variables.xml
AllenJB: I've split the examples.

Your mention of the expected error message here (and the actual error this example gives) led me to realize there's 2 ways to get this wrong, so I've also expanded the example to cover both of them.

jordikroon (approved):

PR #5456, opened 2026-04-02
Enable runnable (WASM) examples for language sections: constants, exceptions, expressions

AllenJB:
Related: #4799 

These sections are all basically one page so I decided to tackle them all in one PR.

Additionally:

* Removed unnecessary `?>`
* Added some links to Exceptions section help people who are looking for the Exception class definition page

I looked at the Fibers section, but attempting to run the example causes an (browser console) error apparently from wasm regarding a missing getcontext function. This error also happens when trying to run the same code on 3v4l.org's live preview.

### DIFF
--- a/language/exceptions.xml
+++ b/language/exceptions.xml
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<chapter xml:id="language.exceptions" xmlns="http://docbook.org/ns/docbook">
+<chapter xml:id="language.exceptions" xmlns="http://docbook.org/ns/docbook" annotations="interactive">
  <title>Exceptions</title>
+ <note>
+  <simpara>See also the <classname>Exception</classname> class</simpara>
+ </note>
  <para>
   PHP has an exception model similar to that of other programming
   languages. An exception can be &throw;n, and caught ("&catch;ed") within
@@ -112,7 +115,7 @@
     </para>
     <example>
      <title>Converting error reporting to exceptions</title>
-     <programlisting role="php">
+     <programlisting role="php" annotations="non-interactive">
 <![CDATA[
 <?php
 function exceptions_error_handler($severity, $message, $filename, $lineno) {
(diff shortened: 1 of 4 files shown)


language/exceptions.xml
```diff
@@ -406,12 +399,12 @@ string(6) "Fourth"
   <title>Extending Exceptions</title>
   <para>
    A User defined Exception class can be defined by extending the built-in
-   Exception class. The members and properties below, show what is accessible
+   <classname>Exception</classname> class. The members and properties below, show what is accessible
```
jordikroon: `<exceptionname>` is preferred over `<classname>` for exceptions.

```suggestion
   <exceptionname>Exception</exceptionname> class. The members and properties below, show what is accessible
```

language/exceptions.xml
```diff
@@ -439,12 +432,11 @@ class Exception implements Throwable
     // Overrideable
     public function __toString();               // formatted string for display
 }
-?>
 ]]>
    </programlisting>
   </example>
   <para>
-   If a class extends the built-in Exception class and re-defines the <link
+   If a class extends the built-in <classname>Exception</classname> class and re-defines the <link
```
jordikroon:
```suggestion
   If a class extends the built-in <exceptionname>Exception</exceptionname> class and re-defines the <link
```

language/exceptions.xml
```diff
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
-<chapter xml:id="language.exceptions" xmlns="http://docbook.org/ns/docbook">
+<chapter xml:id="language.exceptions" xmlns="http://docbook.org/ns/docbook" annotations="interactive">
  <title>Exceptions</title>
+ <note>
+  <simpara>See also the <classname>Exception</classname> class</simpara>
```
jordikroon:
```suggestion
  <simpara>See also the <exceptionname>Exception</exceptionname> class</simpara>
```

language/fibers.xml
```diff
@@ -57,7 +57,6 @@ $value = $fiber->start();
 echo "Value from fiber suspending: ", $value, PHP_EOL;
 
 $fiber->resume('test');
-?>
```
jordikroon, line 60: I am not against removing `?>`. But I don't really see it as a necessary change. At least not if that is the only change in this file.

jordikroon (approved):

PR #5457, opened 2026-04-02
Document DelayedTargetValidation attribute (PHP 8.5)

jordikroon:
Document DelayedTargetValidation attribute that was introduced in PHP 8.5.

### DIFF
--- a/language/predefined/attributes/delayedtargetvalidation.xml
+++ b/language/predefined/attributes/delayedtargetvalidation.xml
@@ -9,7 +9,7 @@
    &reftitle.intro;
    <simpara>
     This attribute delays target validation errors for internal attributes
-    from compile time to runtime.
+    from compile time to when the attribute is instantiated via the Reflection API.
    </simpara>
    <simpara>
     When applied to a declaration, any invalid usage of internal attributes
@@ -37,7 +37,7 @@
 
   </section>
 
-  <section>
+  <section xml:id="delayedtargetvalidation.examples">
    &reftitle.examples;
 
    <example>
--- a/language/predefined/attributes/delayedtargetvalidation.xml
+++ b/language/predefined/attributes/delayedtargetvalidation.xml
@@ -79,7 +79,7 @@ foreach ($reflection->getAttributes() as $attribute) {
 ]]></programlisting>
 
     <simpara>
-     When any attribute applied to the same target (other than 
+     When any attribute applied to the same target (other than
      DelayedTargetValidation itself) is instantiated via reflection using
      <link linkend="reflectionattribute.newinstance">ReflectionAttribute::newInstance()</link>,
      target validation is performed and an exception may be thrown if the attribute
--- a/language/predefined/attributes/delayedtargetvalidation.xml
+++ b/language/predefined/attributes/delayedtargetvalidation.xml
@@ -1,0 +1,132 @@
+<?xml version="1.0" encoding="utf-8"?>
+<reference xml:id="class.delayedtargetvalidation" role="class" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xi="http://www.w3.org/2001/XInclude">
+ <title>The DelayedTargetValidation attribute</title>
+ <titleabbrev>DelayedTargetValidation</titleabbrev>
+
+ <partintro>
+
+  <section xml:id="delayedtargetvalidation.intro">
+   &reftitle.intro;
+   <simpara>
+    This attribute delays target validation errors for internal attributes
+    from compile time to runtime.
+   </simpara>
+   <simpara>
+    When applied to a declaration, any invalid usage of internal attributes
+    on the same target will not trigger a compile time error. Instead, the
+    validation is deferred and performed when the attribute is accessed via
+    the Reflection API.
+   </simpara>
+   <simpara>
+    This is primarily intended for forward compatibility, allowing code to
+    use attributes that may gain additional valid targets in future PHP
+    versions without breaking on older versions.
+   </simpara>
+  </section>
+
+  <section xml:id="delayedtargetvalidation.synopsis">
+   &reftitle.classsynopsis;
+
+   <classsynopsis class="class">
+    <ooclass>
+     <modifier role="attribute">#[\Attribute]</modifier>
+     <modifier>final</modifier>
+     <classname>DelayedTargetValidation</classname>
+    </ooclass>
+   </classsynopsis>
+
+  </section>
+
+  <section>
+   &reftitle.examples;
+
+   <example>
+    <title>Delaying validation of an invalid target</title>
+    <programlisting role="php"><![CDATA[
+<?php
+
+class Base {
+    protected function foo(): void {}
+}
+
+class Child extends Base {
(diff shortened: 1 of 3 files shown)


language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,132 @@
...
+<?php
+
+class Base {
+    protected function foo(): void {}
+}
+
+class Child extends Base {
+
+    #[\DelayedTargetValidation]
+    #[\Override]
```
DanielEScherzer, line 55: given that I have a current RFC for supporting override on constants, great example - if that RFC passes then this would be the perfect demonstration of using the delayed target validation feature

language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,132 @@
...
+$reflection = new ReflectionClassConstant(Child::class, 'NAME');
+
+foreach ($reflection->getAttributes() as $attribute) {
+    $attribute->newInstance(); // May throw if invalid
+}
+]]></programlisting>
+
+    <simpara>
+     When the attribute is instantiated via reflection, target validation is
+     performed and an error may be thrown if the attribute is used on an
```
DanielEScherzer:
```suggestion
     performed and an exception may be thrown if the attribute is used on an
```

and maybe link to ReflectionAttribute::newInstance() (which could also use some improvements, but that is separate)

language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,132 @@
...
+    For example, <classname>Override</classname> will still emit an error if a
+    method does not actually override a parent method.
+   </simpara>
+  </section>
+
+  <section xml:id="delayedtargetvalidation.seealso">
+   &reftitle.seealso;
+   <simplelist>
+    <member><link linkend="language.attributes">Attributes overview</link></member>
+    <member><link linkend="class.override">Override</link></member>
```
DanielEScherzer, line 108: not sure what the policy is about linking to external resources, but I have a write up at https://scherzer.dev/Blog/20250820-delayed-target-validation that may be a useful reference to link to

DanielEScherzer (commented):
looks correct to me, a few suggestions

language/predefined/attributes/delayedtargetvalidation.xml
jordikroon: While it's a great resource. I don't think it's the right thing to do. I prefer we put our best effort into the docs so that external resources are not necessary. If we fail to do, we should fix it in the docs. 

That said we currently don't link to external resources in the see also sections.

language/predefined/versions.xml
```diff
@@ -181,6 +181,7 @@
  <function name="SensitiveParameterValue::getValue" from="PHP 8 &gt;= 8.2.0"/>
  <function name="Override" from="PHP 8 &gt;= 8.3.0"/>
  <function name="Override::__construct" from="PHP 8 &gt;= 8.3.0"/>
+ <function name="DelayedTargetValidation" from="PHP 8 &gt;= 8.5.0"/>
  <function name="Deprecated" from="PHP 8 &gt;= 8.4.0"/>
  <function name="Deprecated::__construct" from="PHP 8 &gt;= 8.4.0"/>
```
lacatoire:
```suggestion
 <function name="Deprecated" from="PHP 8 &gt;= 8.4.0"/>
 <function name="Deprecated::__construct" from="PHP 8 &gt;= 8.4.0"/>
 <function name="DelayedTargetValidation" from="PHP 8 &gt;= 8.5.0"/>
```

language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,133 @@
...
+    <ooclass>
+     <modifier role="attribute">#[\Attribute]</modifier>
+     <modifier>final</modifier>
+     <classname>DelayedTargetValidation</classname>
+    </ooclass>
+   </classsynopsis>
+
+  </section>
+
+  <section>
```
lacatoire:
```suggestion
  <section xml:id="delayedtargetvalidation.examples">
```

language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,133 @@
...
+ <title>The DelayedTargetValidation attribute</title>
+ <titleabbrev>DelayedTargetValidation</titleabbrev>
+
+ <partintro>
+
+  <section xml:id="delayedtargetvalidation.intro">
+   &reftitle.intro;
+   <simpara>
+    This attribute delays target validation errors for internal attributes
+    from compile time to runtime.
```
lacatoire:
```suggestion
    from compile time to when the attribute is instantiated via the Reflection API.
```

language/predefined/attributes/delayedtargetvalidation.xml
jordikroon: Good one!

language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,133 @@
...
+
+$reflection = new ReflectionClassConstant(Child::class, 'NAME');
+
+foreach ($reflection->getAttributes() as $attribute) {
+    $attribute->newInstance(); // May throw if invalid
+}
+]]></programlisting>
+
+    <simpara>
+     When the attribute is instantiated via reflection using
```
DanielEScherzer: we should somehow clarify that "the attribute" here means the other attribute applied in the same location

language/predefined/attributes/delayedtargetvalidation.xml
```diff
@@ -0,0 +1,133 @@
...
+   &reftitle.intro;
+   <simpara>
+    This attribute delays target validation errors for internal attributes
+    from compile time to when the attribute is instantiated via the Reflection API.
+   </simpara>
+   <simpara>
+    When applied to a declaration, any invalid usage of internal attributes
+    on the same target will not trigger a compile time error. Instead, the
+    validation is deferred and performed when the attribute is accessed via
+    the Reflection API.
```
lacatoire:
```suggestion
    When applied to a declaration, any invalid usage of internal attributes
    on the same target will not trigger a compile time error. Instead, the
    validation is deferred and performed when the attribute is instantiated
    via <link linkend="reflectionattribute.newinstance">ReflectionAttribute::newInstance()</link>.
```

lacatoire: `ReflectionClass::getAttributes()` alone does not trigger validation: the deferred check only fires on `ReflectionAttribute::newInstance()` (see [`Zend/zend_compile.c:7874-7930`](https://github.com/php/php-src/blob/master/Zend/zend_compile.c#L7874-L7930), where the error is stored on `attr->validation_error` and only rethrown by `newInstance()`).

jordikroon: Aha. Thank you for clarifying.

jordikroon:
@DanielEScherzer Could you have one final look to see if you are satisfied?

PR #5461, opened 2026-04-04
Add missing 8.0.0 changelog for datetime functions no longer returning false

KentarouTakeda:
In PHP 7.x, several datetime functions returned `false` on parameter parsing failure. PHP 8.0 changed this to throw `TypeError` instead (per RFC: Consistent type errors for internal functions).

This changelog entry was already documented for
`DateTimeZone::listIdentifiers`/`timezone_identifiers_list`, but was missing from the following functions:

- `date()`
- `gmdate()`
- `localtime()`
- `getdate()`
- `date_parse()`
- `date_parse_from_format()`
- `date_sun_info()`

Functions that still return `false` from their own logic (e.g. `strtotime()`, `mktime()`, `date_create()`) are not included, as the parameter parsing change is not observable in their return types.

The issue also requests reverting the return value description to mention `false`, but the old text was inaccurate (non-numeric strings were coerced to `0`, not rejected) and was intentionally removed in #2130.

Closes #5460

### DIFF
--- a/reference/datetime/functions/date-parse-from-format.xml
+++ b/reference/datetime/functions/date-parse-from-format.xml
@@ -124,6 +124,13 @@
       </entry>
      </row>
      <row>
+      <entry>8.0.0</entry>
+      <entry>
+       This function no longer returns &false; on failure, but throws a
+       <exceptionname>TypeError</exceptionname> instead.
+      </entry>
+     </row>
+     <row>
       <entry>7.2.0</entry>
       <entry>
        The <literal>zone</literal> element of the returned array represents
--- a/reference/datetime/functions/date-parse.xml
+++ b/reference/datetime/functions/date-parse.xml
@@ -125,6 +125,13 @@
     </thead>
     <tbody>
      <row>
+      <entry>8.0.0</entry>
+      <entry>
+       This function no longer returns &false; on failure, but throws a
+       <exceptionname>TypeError</exceptionname> instead.
+      </entry>
+     </row>
+     <row>
       <entry>7.2.0</entry>
       <entry>
        The <literal>zone</literal> element of the returned array represents
--- a/reference/datetime/functions/date-sun-info.xml
+++ b/reference/datetime/functions/date-sun-info.xml
@@ -155,6 +155,13 @@
      </thead>
      <tbody>
       <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function no longer returns &false; on failure, but throws a
+        <exceptionname>TypeError</exceptionname> instead.
+       </entry>
+      </row>
+      <row>
        <entry>7.2.0</entry>
        <entry>
         The calculation was fixed with regards to local midnight instead of
--- a/reference/datetime/functions/date.xml
+++ b/reference/datetime/functions/date.xml
@@ -84,6 +84,13 @@
       <row>
        <entry>8.0.0</entry>
        <entry>
+        This function no longer returns &false; on failure, but throws a
+        <exceptionname>TypeError</exceptionname> instead.
+       </entry>
+      </row>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
         <parameter>timestamp</parameter> is nullable now.
        </entry>
       </row>
--- a/reference/datetime/functions/getdate.xml
+++ b/reference/datetime/functions/getdate.xml
@@ -131,6 +131,13 @@
       <row>
        <entry>8.0.0</entry>
        <entry>
+        This function no longer returns &false; on failure, but throws a
+        <exceptionname>TypeError</exceptionname> instead.
+       </entry>
+      </row>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
         <parameter>timestamp</parameter> is nullable now.
        </entry>
       </row>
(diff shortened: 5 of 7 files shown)


reference/datetime/functions/date-parse-from-format.xml
```diff
@@ -123,6 +123,13 @@
        ignored.
       </entry>
      </row>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       This function no longer returns &false; on failure, but throws a
+       <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
       <exceptionname>TypeError</exceptionname> instead.
```

reference/datetime/functions/date-parse.xml
```diff
@@ -124,6 +124,13 @@
      </row>
     </thead>
     <tbody>
+     <row>
+      <entry>8.0.0</entry>
+      <entry>
+       This function no longer returns &false; on failure, but throws a
+       <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
       <exceptionname>TypeError</exceptionname> instead.
```

reference/datetime/functions/date-sun-info.xml
```diff
@@ -154,6 +154,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function no longer returns &false; on failure, but throws a
+        <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
        <exceptionname>TypeError</exceptionname> instead.
```

reference/datetime/functions/date.xml
```diff
@@ -81,6 +81,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function no longer returns &false; on failure, but throws a
+        <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
        <exceptionname>TypeError</exceptionname> instead.
```

reference/datetime/functions/getdate.xml
```diff
@@ -128,6 +128,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function no longer returns &false; on failure, but throws a
+        <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
        <exceptionname>TypeError</exceptionname> instead.
```

reference/datetime/functions/gmdate.xml
```diff
@@ -58,6 +58,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function no longer returns &false; on failure, but throws a
+        <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
        <exceptionname>TypeError</exceptionname> instead.
```

reference/datetime/functions/localtime.xml
```diff
@@ -124,6 +124,13 @@
       </row>
      </thead>
      <tbody>
+      <row>
+       <entry>8.0.0</entry>
+       <entry>
+        This function no longer returns &false; on failure, but throws a
+        <classname>TypeError</classname> instead.
```
jordikroon:
```suggestion
        <exceptionname>TypeError</exceptionname> instead.
```

jordikroon (commented):
This looks good to me, just some minor nits since `exceptionname` is preferred for throwables.

kamil-tekiela:
1. If the same text should be pasted in multiple places, a snippet should be used instead. 
2. As far as I know, these functions should have never been documented to return false as they never did. They only returned false in case the argument parsing failed, which to my knowledge was an undefined behaviour and was never documented in the function's signatures. So in this regard, the changelog is correct that there was no change. See https://github.com/php/doc-en/commit/4754397753fd79f1c846868b66a2448babab1c54

As such, I have to decline this PR as it would repeat the same mistake that was done 6 years ago. 

P.S. `timezone_identifiers_list` used to return `false` until PHP 8, so this is also correct.

KentarouTakeda:
@kamil-tekiela
Thanks for the review — I'm happy to have this PR closed.

For reference, though, it does appear that in PHP 7.4 these functions returned `false` under certain conditions:

- https://php-play.dev/?c=DwfgDgFmBQBuCGAnA%2BgEwK4FswApXwBcBTHASlIG4g&v=7.4&f=html
- https://github.com/php/php-src/blob/php-7.4.33/ext/date/php_date.c#L1288

kamil-tekiela:
> @kamil-tekiela Thanks for the review — I'm happy to have this PR closed.
> 
> For reference, though, it does appear that in PHP 7.4 these functions returned `false` under certain conditions:
> 
> * https://php-play.dev/?c=DwfgDgFmBQBuCGAnA%2BgEwK4FswApXwBcBTHASlIG4g&v=7.4&f=html
> * https://github.com/php/php-src/blob/php-7.4.33/ext/date/php_date.c#L1288

Or `null` https://php-play.dev/?c=DwfgDgFmBQBuCGAnA%2BgEwK4FswAoDOALogDYCmAdjgJRUDcQA&v=7.4&f=html 
It wasn't defined anywhere what they should return, that's why it is undefined behaviour.

PR #5462, opened 2026-04-04
Enable runnable (WASM) examples for language.control-structures section

AllenJB:
Related: #4799


language/control-structures/match.xml
```diff
@@ -219,21 +213,20 @@ try {
...
-?>
 ]]>
   </programlisting>
   &example.outputs;
   <screen>
 <![CDATA[
 object(UnhandledMatchError)#1 (7) {
   ["message":protected]=>
-  string(33) "Unhandled match value of type int"
+  string(33) "Unhandled match case 5"
```
lacatoire:
```suggestion
  string(22) "Unhandled match case 5"
```

language/control-structures/match.xml
```diff
@@ -219,21 +213,20 @@ try {
...
   ["string":"Error":private]=>
   string(0) ""
   ["code":protected]=>
   int(0)
   ["file":protected]=>
-  string(9) "/in/ICgGK"
+  string(9) "script"
```
lacatoire:
```suggestion
  string(6) "script"
```

language/control-structures/switch.xml
```diff
@@ -263,17 +266,18 @@ switch ($i):
...
         echo "i is not equal to 0, 1 or 2";
 endswitch;
-?>
 ]]>
    </programlisting>
   </informalexample>
  </para>
  <para>
-  It's possible to use a semicolon instead of a colon after a case like:
+  It's possible to use a semicolon instead of a colon after a case. This syntax is deprecated.
```
lacatoire:
```suggestion
  It's possible to use a semicolon instead of a colon after a case.
    As of PHP 8.4.0, this syntax is deprecated.
```

language/control-structures/for.xml
```diff
@@ -127,9 +122,9 @@ $people = array(
 );
 
 for($i = 0; $i < count($people); ++$i) {
-    $people[$i]['salt'] = mt_rand(000000, 999999);
+    $people[$i]['salt'] = random_int(100000, 999999);
```
lacatoire, line 163:
```suggestion
    $people[$i]['salt'] = random_int(0, 999999);
```

AllenJB: The 100k minimum was chosen intentionally for both examples here because it results in the output not varying in length, which I think reduces noise in the example.

language/control-structures/break.xml
```diff
@@ -26,25 +26,24 @@ foreach ($arr as $val) {
     if ($val == 'stop') {
         break;    /* You could also write 'break 1;' here. */
     }
-    echo "$val<br />\n";
+    echo "$val\n";
 }
 
-/* Using the optional argument. */
-
+echo "\nUsing the optional argument:\n";
```
jordikroon: Could you split this example up in two?

language/control-structures/for.xml
```diff
@@ -54,36 +54,31 @@ for (expr1; expr2; expr3)
...
-
+echo "Example 1:\n";
 for ($i = 1; $i <= 10; $i++) {
-    echo $i;
+    echo $i . " ";
 }
 
-/* example 2 */
-
+echo "\nExample 2:\n";
```
jordikroon: Could you split this example up?

language/control-structures/while.xml
```diff
@@ -57,19 +57,18 @@ endwhile;
...
 while ($i <= 10) {
-    echo $i++;  /* the printed value would be
-                   $i before the increment
-                   (post-increment) */
+    echo $i++ . "\n";  /* the printed value would be
+                          $i before the increment
+                          (post-increment) */
 }
 
 /* example 2 */
```
jordikroon: Could you split this example up in two?

jordikroon (changes requested):
Most of these examples would really benefit from having a `<screen>` to see the expected output directly. Would you be able to add those?

language/control-structures/switch.xml
```diff
@@ -155,9 +179,14 @@ switch ($i) {
     default:
        echo "i is not equal to 0, 1 or 2";
 }
-?>
 ]]>
    </programlisting>
+   &example.outputs;
+   <screen>
+<![CDATA[
+i is equal to 0, 1 or 2
```
lacatoire:
```suggestion
i is not equal to 0, 1 or 2
```

language/control-structures/continue.xml
```diff
@@ -103,10 +101,9 @@ Inner
 <?php
 for ($i = 0; $i < 5; ++$i) {
     if ($i == 2)
-        continue
+        continue;
```
lacatoire: The surrounding text says "Omitting the semicolon after continue… an example of what you shouldn't do", so this example must keep continue without a semicolon to demonstrate the pitfall (continue print "$i\n";). Changing it to continue; makes the code do exactly the "expected" thing, which contradicts the prose. Since the no-semicolon form is a compile error on modern PHP, this snippet should be marked annotations="non-interactive" and reverted to continue (no ;), rather than "fixed".

AllenJB: I've removed this entire example.

The current content is the result of partial-removal of the section of this example talking about prior behavior in https://github.com/php/doc-en/commit/cdaea0421544885f02ff3d36bd203dc01b78299e#diff-f87a62b6db7766dfa2ad438527a741fe865926718fbe33e13460344f8865bb7f

The behavior the example is trying to describe hasn't been an issue since PHP <5.4 - it now results in an explicit error.

language/control-structures/switch.xml
```diff
@@ -263,17 +300,25 @@ switch ($i):
...
+<![CDATA[
+i equals 1
+]]>
+   </screen>
   </informalexample>
  </para>
  <para>
-  It's possible to use a semicolon instead of a colon after a case like:
+  It's possible to use a semicolon instead of a colon after a case.
+    As of PHP 8.4.0, this syntax is deprecated.
```
lacatoire:
```suggestion
    As of PHP 8.5.0, this syntax is deprecated.
```

language/control-structures/switch.xml
```diff
@@ -286,9 +331,24 @@ switch($beer)
...
+
+Deprecated: Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in script on line 6
+
+Deprecated: Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in script on line 7
+
+Deprecated: Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in script on line 8
+
+Deprecated: Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in script on line 9
+
+Deprecated: Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in script on line 12
```
lacatoire:
```suggestion
Deprecated: Case statements followed by a semicolon (;) are deprecated, use a colon (:) instead in script on line 11
```

PR #5463, opened 2026-04-05
Enable runnable (WASM) examples for language.functions section

AllenJB:
Related: #4799

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -950,6 +950,13 @@ $variable = "Variable";
 Foo::$variable();  // This calls $foo->Variable() reading $variable in this scope.
 ]]>
      </programlisting>
+     &example.outputs;
+     <screen>
+<![CDATA[
+static property
+Method Variable called
+]]>
+     </screen>
     </example>
    </para>
    <para>
@@ -1666,9 +1673,8 @@ $privateMethod();
      </programlisting>
     </example>
    </para>
-   <para>
-    <informalexample>
-     <programlisting role="php">
+   <informalexample>
+    <programlisting role="php">
 <![CDATA[
 <?php
 class Foo1 {
@@ -1686,10 +1692,8 @@ $foo1 = new Foo1;
 $privateMethod = $foo1->getPrivateMethod();
 $privateMethod();  // Foo1::privateMethod
 ]]>
-     </programlisting>
-    </informalexample>
-
-   </para>
+    </programlisting>
+   </informalexample>
 
    <note>
     <para>
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
- <chapter xml:id="language.functions" xmlns="http://docbook.org/ns/docbook">
+ <chapter xml:id="language.functions" xmlns="http://docbook.org/ns/docbook" annotations="interactive">
   <title>Functions</title>
 
   <sect1 xml:id="functions.user-defined">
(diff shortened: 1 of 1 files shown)


language/functions.xml
```diff
@@ -14,27 +14,25 @@
...
-    <programlisting role="php">
+    <programlisting role="php" annotations="non-interactive">
 <![CDATA[
 <?php
 function foo($arg_1, $arg_2, /* ..., */ $arg_n)
 {
     echo "Example function.\n";
     return $retval;
 }
-?>
```
TimWolla, line 25: Removing the closing `?>` is needless churn. Given how the examples sit within the documentation, I personally also feel that the `?>` is a nice “example ends here” indicator.

language/functions.xml
```diff
@@ -1700,7 +1662,15 @@ $privateMethod = $foo->getPrivateMethod();
 $privateMethod();
 // Fatal error: Call to private method Foo::privateMethod() from global scope
 // This is because call is performed outside from Foo and visibility will be checked from this point.
-
+]]>
+     </programlisting>
+    </example>
+   </para>
+   <para>
+    <informalexample>
```
Girgias: You don't need a para tag to wrap an example/informalexample tag

language/functions.xml
```diff
@@ -958,15 +941,13 @@ class Foo
...
 
-echo Foo::$variable; // This prints 'static property'. It does need a $variable in this scope.
+echo Foo::$variable ."\n"; // This prints 'static property'. It does need a $variable in this scope.
 $variable = "Variable";
 Foo::$variable();  // This calls $foo->Variable() reading $variable in this scope.
-
-?>
 ]]>
      </programlisting>
     </example>
```
Girgias, line 960: Could this have a screen element and showcase the output?

Girgias (changes requested):
I don't think removing the `?>` is worthwhile churn, especially as this is adding kinda unnecessary work for translations.

language/functions.xml
AllenJB: I've been removing these because I've seen new developers always including them (and occasionally running into the subsequent issues) and I think one reason they may do this is because the manual examples do.

Additionally I would note that the coding style for examples in doc-base says to follow PER-CS, which also says that `?>` MUST be omitted from files containing only PHP (I generally consider each example to be a "file" for this purpose, particularly where the examples is runnable)

TimWolla: > Additionally I would note that the coding style for examples in doc-base says to follow PER-CS, which also says that `?>` MUST be omitted from files containing only PHP

The documentation naturally cannot follow any coding style in its entirety, because it needs to showcase everything that's possible in PHP. I would consider it a soft-target of “use the brace placement” or “use the whitespace” as defined in PER-CS, which is what does the heavy lifting of making code look uniform.

In fact almost all of the examples in the documentation already violate a SHOULD from PSR-1 because they mix declarations and side-effects. Due to the `echo`s I would also consider them to not be a file “containing only PHP”.

In any case, PRs should be focused on one thing. A PR titled “enable runnable examples” should not make changes that have no effect on whether the example is runnable or not.

AllenJB: Violating mixed declarations and side-effects is obviously unavoidable (particularly with runnable examples). I don't think this is a good counter-example here. Trailing `?>` is not unavoidable. 

Many examples obviously don't follow PER-CS in many ways for legacy reasons, but if it's the documented style standard then surely examples should be updated where this is noticed and reasonable, in the same way I've been asked to correct `simpara` vs `para` (and other) xml style violations, including outside the scope of changes I was otherwise making.

If the manual code style is to ignore this PER-CS rule / always include trailing `?>`, I would suggest this is noted in the cs-for-examples docs. I think this is a significant and not necessarily expected departure from PER-CS and other contributors are likely to run into the same issue.

PR #5471, opened 2026-04-06
uri: Unify example URL for RFC 3986 getters

TimWolla:
This new example URL includes all possible components and each of the component includes a meaningful difference between the “normalized” and “raw” version to make it easy for a reader to see what each of the getters refers to and what kind of normalization to expect.

### DIFF
--- a/reference/uri/uri/rfc3986/uri/getrawuserinfo.xml
+++ b/reference/uri/uri/rfc3986/uri/getrawuserinfo.xml
@@ -35,16 +35,18 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-$uri = new \Uri\Rfc3986\Uri("https://user:password@example.com");
 
-echo $uri->getRawUserInfo();
+$uri = new \Uri\Rfc3986\Uri('HTTPS://p%68p:elephp%61nt@WWW.PHP.NET:443/manual/ch%61nge.php?p%61ge=en/book.uri.php#%62ook.uri');
+
+var_dump($uri->getRawUserinfo());
+
 ?>
 ]]>
    </programlisting>
    &example.outputs;
    <screen>
 <![CDATA[
-user:password
+string(17) "p%68p:elephp%61nt"
 ]]>
    </screen>
   </example>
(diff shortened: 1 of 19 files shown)


TimWolla:
> I'm not sure that people less familiar with URI topic will have an easy time to understand what's going on

I think those folks in particular can benefit when all the examples use the same URL, since they can then compare the different getters to see which of the getters relates to which part of the URL. Having different URLs where some of the parts are left out can result in ambiguity. For example using `http://example.com/foo/bar` hides that the path will end when encountering `?` and `#`. This might mislead users to believe that the path is everything after the third slash.

Basically when I look at the documentation page for a specific getter, then I'm interested in the specifics of that getter.

I think simpler URLs are better suited for the documentation of the constructor or `::parse()` where users care more about “how do I get this to work” rather than extracting specific stuff.

reference/uri/uri/rfc3986/uri/getrawuserinfo.xml
```diff
@@ -35,16 +35,18 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-$uri = new \Uri\Rfc3986\Uri("https://user:password@example.com");
 
-echo $uri->getRawUserInfo();
+$uri = new \Uri\Rfc3986\Uri('HTTPS://p%68p:elephp%61nt@WWW.PHP.NET:443/manual/ch%61nge.php?p%61ge=en/book.uri.php#%62ook.uri');
+
+var_dump($uri->getRawUserinfo());
```
lacatoire, line 41:
```suggestion
var_dump($uri->getRawUserInfo());
```

reference/uri/uri/rfc3986/uri/getuserinfo.xml
```diff
@@ -35,16 +35,18 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-$uri = new \Uri\Rfc3986\Uri("https://user:password@example.com");
 
-echo $uri->getUserInfo();
+$uri = new \Uri\Rfc3986\Uri('HTTPS://p%68p:elephp%61nt@WWW.PHP.NET:443/manual/ch%61nge.php?p%61ge=en/book.uri.php#%62ook.uri');
+
+var_dump($uri->getUserinfo());
```
lacatoire, line 41:
```suggestion
var_dump($uri->getUserInfo());
```

PR #5482, opened 2026-04-07
proc_close: "waits", not "closes". Also grammar fix in Returns.

SinusPi:
The function waits (waitpid) for a process to terminate, it does not do any closing. The description is a little misleading.

### DIFF
--- a/reference/exec/functions/proc-close.xml
+++ b/reference/exec/functions/proc-close.xml
@@ -4,7 +4,7 @@
 <refentry xml:id='function.proc-close' xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>proc_close</refname>
-  <refpurpose>Close a process opened by <function>proc_open</function> and return the exit code of that process</refpurpose>
+  <refpurpose>Wait for a process opened by <function>proc_open</function> to close and return the exit code of that process</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
@@ -46,7 +46,7 @@
   &reftitle.returnvalues;
   <para>
    Returns the termination status of the process that was run. In case of 
-   an error then <literal>-1</literal> is returned.
+   an error <literal>-1</literal> is returned.
   </para>
   &note.sigchild;
  </refsect1>
--- a/reference/exec/functions/proc-close.xml
+++ b/reference/exec/functions/proc-close.xml
@@ -4,7 +4,7 @@
 <refentry xml:id='function.proc-close' xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>proc_close</refname>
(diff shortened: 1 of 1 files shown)


reference/exec/functions/proc-close.xml
```diff
@@ -4,7 +4,7 @@
 <refentry xml:id='function.proc-close' xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>proc_close</refname>
-  <refpurpose>Close a process opened by <function>proc_open</function> and return the exit code of that process</refpurpose>
+  <refpurpose>Wait for a process opened by <function>proc_open</function> to close and return the exit code of that process</refpurpose>
```
lacatoire:
```suggestion
  <refpurpose>Close pipes to a process opened by <function>proc_open</function>, wait for it to terminate, and return its exit code</refpurpose>
```

lacatoire: For context: `proc_close()` actually does perform closing, just not on the process itself. It closes the open pipes to the child before waiting on it, which is the whole point of the "avoid a deadlock" comment in the resource [destructor](https://github.com/php/php-src/blob/master/ext/standard/proc_open.c#L270-L277).

That is why the current "Close a process" phrasing, although imprecise about the process, is not entirely wrong about the closing. A refpurpose that names both actions (pipe close + process wait) would be the most faithful to the implementation.

reference/exec/functions/proc-close.xml
```diff
@@ -46,7 +46,7 @@
   &reftitle.returnvalues;
   <para>
```
jordikroon:
```suggestion
  <simpara>
```

reference/exec/functions/proc-close.xml
```diff
@@ -46,7 +46,7 @@
...
    Returns the termination status of the process that was run. In case of 
-   an error then <literal>-1</literal> is returned.
+   an error <literal>-1</literal> is returned.
   </para>
```
jordikroon:
```suggestion
  </simpara>
```

jordikroon (changes requested):
All good, just one small nit; when changing copy that is within a `<para>`. We should also change to `<simpara>`. See it as a change from `<div>` to `<p>`.

reference/exec/functions/proc-close.xml
```diff
@@ -46,7 +46,7 @@
   &reftitle.returnvalues;
   <para>
    Returns the termination status of the process that was run. In case of 
-   an error then <literal>-1</literal> is returned.
+   an error <literal>-1</literal> is returned.
```
lacatoire:
```suggestion
   an error, <literal>-1</literal> is returned.
```

jordikroon:
To not keep these nits wait longer, I have applied the suggestions on your behalf. Thank you for your contribution.

jordikroon (approved):

PR #5484, opened 2026-04-08
Enable runnable (WASM) examples for language sections: attributes, generators and references

AllenJB:
Related: #4799

### DIFF
--- a/language/references.xml
+++ b/language/references.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- $Revision$ -->
- <chapter xml:id="language.references" xmlns="http://docbook.org/ns/docbook">
+ <chapter xml:id="language.references" xmlns="http://docbook.org/ns/docbook" annotations="interactive">
   <title>References Explained</title>
   
   <sect1 xml:id="language.references.whatare">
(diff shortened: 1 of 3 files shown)


language/references.xml
```diff
@@ -185,7 +171,6 @@ echo $ref; // 3 - last element of the iterated array
       <programlisting role="php">
```
lacatoire, line 171:
```suggestion
      <programlisting role="php" annotations="non-interactive">
```

language/references.xml
```diff
@@ -495,14 +468,17 @@ $collection = &collector();
...
-
-?> 
 ]]>
      </programlisting>
+     &example.outputs;
+     <screen>
+<![CDATA[
+Array
+(
+   [0] => foo
```
lacatoire:
```suggestion
    [0] => foo
```

language/references.xml
jordikroon: This should stay as interactive. The program/example is executable.

jordikroon (changes requested):

language/references.xml
AllenJB: As jordi says, I see no reason to mark this example as non-interactive.

jordikroon (approved):

Issue #5488, opened 2026-04-13
Wrong information in comment at example #1 (Implicit Commit Example)


abdelrahman-gado:
From manual page: https://php.net/pdo.transactions

---

in the Example 1 Implicit Commit Example 
```php
<?php
$pdo->beginTransaction();
$pdo->exec("INSERT INTO users (name) VALUES ('Rasmus')");
$pdo->exec("CREATE TABLE test (id INT PRIMARY KEY)"); // Implicit COMMIT occurs here
$pdo->rollBack(); // This will NOT undo the INSERT/CREATE for MySQL or Oracle
?>
```
The comment in the last line `$pdo->rollback();` explains that we cannot undo the INSERT and this is wrong.
we can undo the INSERT but we cannot undo the create or drop (DDL) statements, So it should be like this 

```php
<?php
$pdo->beginTransaction();
$pdo->exec("INSERT INTO users (name) VALUES ('Rasmus')");
$pdo->exec("CREATE TABLE test (id INT PRIMARY KEY)"); // Implicit COMMIT occurs here
$pdo->rollBack(); // This will NOT undo the DROP/CREATE for MySQL or Oracle
?>
```

kamil-tekiela:
> The comment in the last line $pdo->rollback(); explains that we cannot undo the INSERT and this is wrong.
we can undo the INSERT but we cannot undo the create or drop (DDL) statements

It's not wrong. We cannot undo THIS `INSERT` because it has been implicitly committed by the DDL. If you think it can be written more clearly, feel free to submit a PR to improve the wording, but as it stands, this issue report is invalid, so I am going to close it.

PR #5489, opened 2026-04-13
fix: change INSERT to be DROP in implicit commit example 1

abdelrahman-gado:
Issue link => https://github.com/php/doc-en/issues/5488

From manual page: https://php.net/pdo.transactions

---

in the Example 1 Implicit Commit Example 
```php
<?php
$pdo->beginTransaction();
$pdo->exec("INSERT INTO users (name) VALUES ('Rasmus')");
$pdo->exec("CREATE TABLE test (id INT PRIMARY KEY)"); // Implicit COMMIT occurs here
$pdo->rollBack(); // This will NOT undo the INSERT/CREATE for MySQL or Oracle
?>
```
The comment in the last line `$pdo->rollback();` explains that we cannot undo the INSERT and this is wrong.
we can undo the INSERT but we cannot undo the create or drop (DDL) statements, So it should be like this 

```php
<?php
$pdo->beginTransaction();
$pdo->exec("INSERT INTO users (name) VALUES ('Rasmus')");
$pdo->exec("CREATE TABLE test (id INT PRIMARY KEY)"); // Implicit COMMIT occurs here
$pdo->rollBack(); // This will NOT undo the DROP/CREATE for MySQL or Oracle
?>
```

### DIFF
--- a/reference/pdo/transactions.xml
+++ b/reference/pdo/transactions.xml
@@ -72,7 +72,7 @@
 $pdo->beginTransaction();
 $pdo->exec("INSERT INTO users (name) VALUES ('Rasmus')");
 $pdo->exec("CREATE TABLE test (id INT PRIMARY KEY)"); // Implicit COMMIT occurs here
-$pdo->rollBack(); // This will NOT undo the INSERT/CREATE for MySQL or Oracle
+$pdo->rollBack(); // This will NOT undo the DROP/CREATE for MySQL or Oracle
 ?>
 ]]>
    </programlisting>


kamil-tekiela:
There is no `DROP` in that example. The comment is correct https://phpize.online/sql/mysql80/2521aa89a15e10f3e4cecaac6065540a/php/php84/385caba4a7d1ffaa8c6d497709ff188c/

Maybe it's a bit confusing that the slash is used there, but it means that neither the INSERT nor the CREATE can be rolled back.

PR #5491, opened 2026-04-14
Tutorial: Improved What's Next section

AllenJB:
I present a revised "What's Next?" tutorial section, with the aim of better pointing new users towards the manual sections for specific things they may want to achieve, and the range of resources available in the PHP ecosystem more generally.

I am aware the external links need to be pulled out to doc-base, but wanted to get initial feedback before I do that.

All feedback welcome.

### DIFF
--- a/chapters/tutorial.xml
+++ b/chapters/tutorial.xml
@@ -305,15 +305,15 @@ You are using Firefox.
 <![CDATA[
 <?php
 if (str_contains($_SERVER['HTTP_USER_AGENT'], 'Firefox')) {
-?>
-<h3>str_contains() returned true</h3>
-<p>You are using Firefox</p>
-<?php
+    ?>
+    <h3>str_contains() returned true</h3>
+    <p>You are using Firefox</p>
+    <?php
 } else {
-?>
-<h3>str_contains() returned false</h3>
-<p>You are not using Firefox</p>
-<?php
+    ?>
+    <h3>str_contains() returned false</h3>
+    <p>You are not using Firefox</p>
+    <?php
 }
 ?>
 ]]>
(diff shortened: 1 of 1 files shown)


chapters/tutorial.xml
```diff
@@ -420,14 +420,81 @@ Hi Joe. You are 22 years old.
...
+   </para>
+   <para>
+    In particular you may want to explore the following features:
+    <simplelist>
+     <member>Reading and writing files with the <link linkend="book.filesystem">filesystem functions</link></member>
+     <member><link linkend="features.file-upload">Handling file uploads</link></member>
+     <member>Fetching remote pages and files with <link linkend="book.curl">Curl</link></member>
+     <member>
+      Storing and analyzing data in a database with <link linkend="book.pdo">PDO</link> (and
+      <link linkend="ref.pdo-sqlite">SQLite</link>)
```
adrian-enspired: why call out sqlite specifically?
if you want more than "database," i'd mention several. maybe

> Storing and analyzing data in a database with PDO - supporting MySQL, PostgreSQL, SQLite and more

AllenJB: I wanted to specifically point readers towards SQLite as it's something they can easily run with just the built-in webserver (which the tutorial already has them using). 

It's something they can dive into immediately without worrying about extra software, DB-server accounts or networking issues (and while I don't have any recent experience with shared hosting to say definitively, something they should be able to run on a hosted site). If / when readers want to move to server-based DBs, everything they'll have learned with SQLite will be transferable.

TimWolla: SQLite is usable with PDO, so I would just refer them to PDO rather than specifically the SQLite section (where they would need to switch APIs when they want a different RDBMS). Perhaps something like “SQLite can be used without running a database server” (https://www.php.net/manual/en/class.pdo-sqlite.php).

chapters/tutorial.xml
```diff
@@ -420,14 +420,81 @@ Hi Joe. You are 22 years old.
...
+    the <link linkend="install.composer.intro">Composer package manager</link>.
+   </para>
+   <para>
+    For community help and advice, check out the <link xlink:href="&url.php.support;">Help page</link>.
+   </para>
+   <para>
+    For a variety of podcasts, presentations and other videos, check out the
+    <link xlink:href="https://phpc.tv">community PeerTube</link>.
+   </para>
+   <para>
```
TimWolla: All these should be `<simpara>`.

TimWolla (commented):
I think this is a good improvement overall.

Girgias (commented):
LGTM, outside of Tim's comment about using `<simpara>` where possible (as it gets rendered to a standard HTML `<p>` tag rather than a `<div>`

chapters/tutorial.xml
```diff
@@ -419,16 +419,83 @@ Hi Joe. You are 22 years old.
...
+    for every occasion on the <link xlink:href="&url.packagist;">Packagist repository</link>, all installable via
+    the <link linkend="install.composer.intro">Composer package manager</link>.
+   </simpara>
+   <simpara>
+    For community help and advice, check out the <link xlink:href="&url.php.support;">Help page</link>.
+   </simpara>
+   <simpara>
+    For a variety of podcasts, presentations and other videos, check out the
+    <link xlink:href="&url.phpctv;">community PeerTube</link>.
+   </simpara>
```
TimWolla, line 449: Seeing this one being the source of the build failure: This might require some discussion on Internals whether it's acceptable to specifically call out this instance (over any other or YouTube).

If you drop this paragraph, I'm happy to merge.

AllenJB: I would like to keep this if possible, I think it's a useful resource.

With regards to YouTube, I don't know of any way to reliably link to a variety of up-to-date PHP community content in the same way phpc.tv allows here. 

Based on socials (phpc.tv's and Anna Filina's mastodon posts) and information on phpc.tv itself, anyone can submit their existing channel content to be sync'd but accounts are moderated. Subsequently I believe the likelihood of phpc.tv showing users undesirable content is low.

Mailing list discussion: https://news-web.php.net/php.internals/130668

TimWolla: > I would like to keep this if possible, I think it's a useful resource.

Depending on the results of the discussion, we can add it back it later and get the other improvements shipped already?

chapters/tutorial.xml
```diff
@@ -419,16 +419,83 @@ Hi Joe. You are 22 years old.
...
+   <simpara>
+    Other community resources that will help you include "awesome lists" (curated directories of links) and "developer
+    roadmaps" (lists of related topics)
```
lacatoire:
```suggestion
    roadmaps" (lists of related topics).
```

chapters/tutorial.xml
```diff
@@ -419,16 +419,83 @@ Hi Joe. You are 22 years old.
...
+       <listitem>
+        <simpara>Persisting data (user login) across requests / pages (sessions)</simpara>
+       </listitem>
+      </itemizedlist>
+     </listitem>
+    </itemizedlist>
    </para>
+   <simpara>
+    If there's nothing particular you want to build, you can try searching for coding exercises such as katas,
+    challenges and "code golf". Even when not specifically aimed at PHP, most should be completeable and will likely
```
lacatoire:
```suggestion
    challenges and "code golf". Even when not specifically aimed at PHP, most should be completable and will likely
```

AllenJB:
It's now been 2 weeks since I posted on internals and there's been no messages against and 1 in support of the inclusion of phpc.tv

The doc-base PR has also been merged.

As far as I'm aware there's no outstanding issues on this PR

jordikroon:
As there haven't been any objections regarding the new external resource over the mailing list. This is good to go. 
Thank you for checking, and if desired we can always make alterations on this matter.

PR #5494, opened 2026-04-15
Document warning emitted by openssl_x509_read

jawira:
An error is emitted when the certificate is not found:

```php
<?php
openssl_x509_read('qsdfqsdfqsdf');
// PHP Warning:  openssl_x509_read(): X.509 Certificate cannot be retrieved in...
```

### DIFF
--- a/reference/openssl/functions/openssl-x509-read.xml
+++ b/reference/openssl/functions/openssl-x509-read.xml
@@ -45,9 +45,9 @@
 
  <refsect1 role="errors">
   &reftitle.errors;
-  <para>
+  <simpara>
    An <constant>E_WARNING</constant> is emitted if the X.509 certificate cannot be retrieved.
-  </para>
+  </simpara>
  </refsect1>
 
  <refsect1 role="changelog">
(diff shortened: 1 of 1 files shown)


reference/openssl/functions/openssl-x509-read.xml
```diff
@@ -43,6 +43,13 @@
   </para>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <para>
+   An <constant>E_WARNING</constant> is emitted if the X.509 certificate cannot be retrieved.
+  </para>
```
lacatoire:
```suggestion
  <simpara>
   An <constant>E_WARNING</constant> is emitted if the X.509 certificate cannot be retrieved.
  </simpara>
```

lacatoire:
Checked against the [implementation](https://github.com/php/php-src/blob/c5061420444001237bfd4d8cd921fb7ff0a9fcc8/ext/openssl/openssl.c#L1263). 
LGTM, one nit: `<para>` with only inline content should be a `<simpara>`.

jawira (commented):
changes accepted

lacatoire (approved):

jordikroon (approved):

PR #5496, opened 2026-04-16
MongoDB: document readonly properties and IWM URI options

alcaeus:
This PR adds documentation for `public readonly` properties being added in ext-mongodb 2.3.0 as well as the new URI options for Intelligent Workload Management.

### DIFF
--- a/reference/mongodb/mongodb/driver/exception/bulkwriteexception.xml
+++ b/reference/mongodb/mongodb/driver/exception/bulkwriteexception.xml
@@ -42,7 +42,8 @@
 <!-- }}} -->
     <classsynopsisinfo role="comment">&Properties;</classsynopsisinfo>
     <fieldsynopsis>
-     <modifier>protected</modifier>
+     <modifier>public</modifier>
+     <modifier>readonly</modifier>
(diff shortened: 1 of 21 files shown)


jordikroon:
Since this is still in draft, I will just point these 2 bullets as this will save reviewers a lot of time.

`simpara` is preferred over `para`. Could you change this where possible (most of the times).
Also I noticed that `informaltable` is wrapped in a `para`. It's fine to remove this outer `para`.

alcaeus:
> Could you change this where possible (most of the times).

Can you clarify "most of the time"? I have a draft PR up for the changes (see #5517), but I'd like to avoid having to look at every single documentation page to make sure I didn't break anything. I'll update this PR to remove the `para` wrapping the `informaltable` elements used for change logs and keep the remaining updates in a separate PR.

jordikroon:
@alcaeus `simpara`  will pretty much render as `<p>` tag whereas `para` will render as `<div>`.
The technical details can be found here in the [docbook manual](https://tdg.docbook.org/tdg/5.2/simpara.html) (see children for what it may contain). But like I said, mostly paragraphs only contain text.

GromNaN (approved):

PR #5500, opened 2026-04-20
Add new bpftrace section for DTrace document

egmc:
This PR adds new section which explains how to use DTrace with bpdtrace command, related to my previous PR #4456 .

Since this PR adds completely new section, I want to explain why it should be added, why it is useful for PHP users. Hoping it could be merged.

##  Background / Motivation

To encourage PHP users to use useful, stable DTrace  based tracepoints easier way.

DTrace is basically an option, though most of PHP packages provided by Linux distributions(aside from Debian) are enabled in reality.

Ubuntu 24.04 Noble is also disabled because of build error from upstream, though enabled again from next version. Here are related issues:
	
- https://bugs.launchpad.net/ubuntu/+source/php8.3/+bug/2088977
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1110885

Although the existing SystemTap section requires users to install SystemTap which is  a bit difficult on modern Linux distributions.

For instance, I have installed SytemTap( And gcc-12 because of the missing dependency) with Ubuntu 22.04 Jammy. I got bunch of errors like below.

Systemtap version 4.6/0.186
https://gist.github.com/egmc/1667b12fcf58eb4fdfbddce256f35084

It may be fixed, though it could be hard for users who tried the functionality for first time.

In contrast, eBPF is now supported natively in mainline Linux kernels, and bpftrace (a high-level eBPF tracing frontend) is available as a standard package on
major distributions (Fedora, RHEL, Oracle Linux, Debian, Ubuntu). The examples in the new section work reliably out of the box with a simple package install,
making them a more practical starting point for users who want to trace PHP on Linux.

bpftrace v0.21.2
https://gist.github.com/egmc/467678a1b6b845bd0a5605dcf3aaa906

So adding bpftrace version would provide more straightforward, easy way to uses.

Notes

- The new section is placed after the existing SystemTap section
- Probe names and arguments match the existing DTrace/SystemTap documentation
- The all_probes.bt script mirrors the existing all_probes.stp example for consistency

### DIFF
--- a/features/dtrace.xml
+++ b/features/dtrace.xml
@@ -552,6 +552,194 @@ probe process("sapi/cli/php").provider("php").mark("request__startup") {
    </para>
   </sect2>
  </sect1>
+ <sect1 xml:id="features.dtrace.bpftrace">
+  <title>Using bpftrace with PHP DTrace Static Probes</title>
+    <para>
+     On Linux distributions with a kernel that supports eBPF, the
+     bpftrace utility can attach to PHP's DTrace USDT probes directly,
+     without requiring SystemTap.
+    </para>
+  <sect2 xml:id="features.dtrace.bpftrace-install">
+   <title>Installing bpftrace</title>
+   <para>
+    Install bpftrace using the distribution's package manager. For
+    example, on Oracle Linux, RHEL, or Fedora:
+    <informalexample>
+     <programlisting role="shell">
+<![CDATA[
+# dnf install bpftrace
+]]>
+     </programlisting>
+    </informalexample>
+    Or, on Debian or Ubuntu:
+    <informalexample>
+     <programlisting role="shell">
+<![CDATA[
+# apt install bpftrace
+]]>
+     </programlisting>
+    </informalexample>
+   </para>
+   <para>
+    The examples below assume the target PHP binary is installed at
+    <filename>/usr/bin/php</filename>.
+   </para>
+   <para>The same USDT probes are also
+    exposed by other SAPIs built from the same source tree, so the
+    probe target may instead be the Apache module
+    (<filename>libphp.so</filename>) or the FastCGI Process Manager
+    binary (<filename>php-fpm</filename>); substitute the appropriate
+    path or attach by PID with <literal>-p</literal> as needed.
+   </para>
+   <para>
+    Make sure target binary is built with DTrace and configured environment properly.
+    <link linkend="features.dtrace.install">Configuring PHP for DTrace Static Probes</link> for details.
+    </para>
+   <para>
+    The static probes in PHP can be listed using
+    <filename>bpftrace</filename>:
+    <informalexample>
+     <programlisting>
+<![CDATA[
+# bpftrace -l 'usdt:/usr/bin/php:php:*'
+]]>
+     </programlisting>
+    </informalexample>
+   </para>
+
+   <para>
+    This outputs:
+    <informalexample>
+     <programlisting>
+<![CDATA[
+usdt:/usr/bin/php:php:compile__file__entry
+usdt:/usr/bin/php:php:compile__file__return
+usdt:/usr/bin/php:php:error
+usdt:/usr/bin/php:php:exception__caught
+usdt:/usr/bin/php:php:exception__thrown
+usdt:/usr/bin/php:php:execute__entry
+usdt:/usr/bin/php:php:execute__return
+usdt:/usr/bin/php:php:function__entry
+usdt:/usr/bin/php:php:function__return
+usdt:/usr/bin/php:php:request__shutdown
+usdt:/usr/bin/php:php:request__startup
+]]>
+     </programlisting>
+    </informalexample>
+   </para>
+
+   <para>
+    <example>
+     <title><filename>all_probes.bt</filename> for tracing all PHP Static Probes with bpftrace</title>
+     <programlisting role="shell">
+<![CDATA[
+#!/usr/bin/env bpftrace
+
+usdt:/usr/bin/php:php:compile__file__entry
+{
+    printf("Probe compile__file__entry\n");
+    printf("  compile_file %s\n", str(arg0));
+    printf("  compile_file_translated %s\n", str(arg1));
+}
+usdt:/usr/bin/php:php:compile__file__return
+{
+    printf("Probe compile__file__return\n");
+    printf("  compile_file %s\n", str(arg0));
+    printf("  compile_file_translated %s\n", str(arg1));
+}
+usdt:/usr/bin/php:php:error
+{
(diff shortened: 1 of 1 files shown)


features/dtrace.xml
```diff
@@ -546,6 +546,194 @@ probe process("sapi/cli/php").provider("php").mark("request__startup") {
...
+   </para>
+  </sect2>
+ </sect1>
+ <sect1 xml:id="features.dtrace.bpftrace">
+  <title>Using bpftrace with PHP DTrace Static Probes</title>
+    <para>
+     On Linux distributions with a kernel that supports eBPF, the
+     bpftrace utility can attach to PHP's DTrace USDT probes directly,
+     without requiring SystemTap.
+    </para>
```
lacatoire:
```suggestion
    <simpara>
     On Linux distributions with a kernel that supports eBPF, the
     bpftrace utility can attach to PHP's DTrace USDT probes directly,
     without requiring SystemTap.
    </simpara>
```

features/dtrace.xml
```diff
@@ -546,6 +546,194 @@ probe process("sapi/cli/php").provider("php").mark("request__startup") {
...
+    exposed by other SAPIs built from the same source tree, so the
+    probe target may instead be the Apache module
+    (<filename>libphp.so</filename>) or the FastCGI Process Manager
+    binary (<filename>php-fpm</filename>); substitute the appropriate
+    path or attach by PID with <literal>-p</literal> as needed.
+   </para>
+   <para>
+    Make sure target binary is built with DTrace and configured environment properly.
+    <link linkend="features.dtrace.install">Configuring PHP for DTrace Static Probes</link> for details.
+    </para>
```
lacatoire:
```suggestion
   <simpara>
    Make sure the target binary is built with DTrace and that the environment is configured properly. See 
    <link linkend="features.dtrace.install">Configuring PHP for DTrace Static Probes</link> for details.
   </simpara>
```

lacatoire:
Nit: indentation inside the new `<sect1 xml:id="features.dtrace.bpftrace">` uses mixed levels (1/4/6 spaces) whereas the rest of the file uses a regular 1/2/3 hierarchy, matching the SystemTap `<sect1>` above. Could you realign to match? For example `<title>` at 2 spaces, `<sect2>` at 2, `<para>` inside at 3.

lacatoire (changes requested):

egmc:
@lacatoire Thank you for the quick comment! I'll add commits as suggested and fix indent levels as well.

egmc:
@lacatoire I think I've fixed it up to this point. Should I push everything back using sqush?

lacatoire:
@egmc simple push, no squash

egmc:
The issues pointed out have been addressed, and I have completed the local build verification.

If any further corrections or explanations regarding the background are needed, I would happy to do it.
Thanks.

features/dtrace.xml
```diff
@@ -546,6 +546,194 @@ probe process("sapi/cli/php").provider("php").mark("request__startup") {
...
+<![CDATA[
+# apt install bpftrace
+]]>
+     </programlisting>
+    </informalexample>
+   </para>
+   <para>
+    The examples below assume the target PHP binary is installed at
+    <filename>/usr/bin/php</filename>.
+   </para>
```
jordikroon:
```suggestion
   <simpara>
    The examples below assume the target PHP binary is installed at
    <filename>/usr/bin/php</filename>.
   </simpara>
```

features/dtrace.xml
```diff
@@ -546,6 +546,194 @@ probe process("sapi/cli/php").provider("php").mark("request__startup") {
...
+   <para>
+    The same USDT probes are also exposed by other SAPIs built from the same source tree, so the
+    probe target may instead be the Apache module
+    (<filename>libphp.so</filename>) or the FastCGI Process Manager
+    binary (<filename>php-fpm</filename>); substitute the appropriate
+    path or attach by PID with <literal>-p</literal> as needed.
+   </para>
```
jordikroon:
```suggestion
   <simpara>
    The same USDT probes are also exposed by other SAPIs built from the same source tree, so the
    probe target may instead be the Apache module
    (<filename>libphp.so</filename>) or the FastCGI Process Manager
    binary (<filename>php-fpm</filename>); substitute the appropriate
    path or attach by PID with <literal>-p</literal> as needed.
   </simpara>
```

jordikroon (commented):
Just a few para's that should be simpara as well (because they only contain parsed text).

egmc:
@jordikroon Certainly. I've incorporated your suggestions and checked the rendering again; it looks good.Thanks!

features/dtrace.xml
```diff
@@ -546,6 +546,194 @@ probe process("sapi/cli/php").provider("php").mark("request__startup") {
...
+    binary (<filename>php-fpm</filename>); substitute the appropriate
+    path or attach by PID with <literal>-p</literal> as needed.
+   </simpara>
+   <simpara>
+    Make sure the target binary is built with DTrace and that the environment is configured properly.
+    See <link linkend="features.dtrace.install">Configuring PHP for DTrace Static Probes</link> for details.
+   </simpara>
+   <para>
+    The static probes in PHP can be listed using
+    <filename>bpftrace</filename>:
```
lacatoire:
```suggestion
    <command>bpftrace</command>:
```

features/dtrace.xml
lacatoire: The `-p $PID` example reuses `all_probes.bt`, whose probes are hardcoded to `usdt:/usr/bin/php`. Since USDT probes are tied to a specific binary, this won't match a `php-fpm` (`/usr/sbin/php-fpm`) or Apache (`libphp.so`) target unless the path is updated:

```suggestion
</informalexample>
The <literal>usdt:</literal> target path in the script must match the binary of
the running process; adjust <literal>usdt:/usr/bin/php</literal> to the
<filename>php-fpm</filename> binary or <filename>libphp.so</filename> as appropriate.
</para>
```

egmc: Ah, yes, need to match the path of each binary.

I have tested in my environment(with version 8.1, ubuntu), both `usdt:/usr/sbin/php-fpm8.1:` with php-fpm and `usdt:/usr/lib/apache2/modules/libphp8.1` with apache+libphp work.

https://gist.github.com/egmc/c1fd74f04bf555a98d4f241c9d63e046

Note:file path with wild card (e.g. `usdt:/usr/lib/apache2/modules/libphp*.so`) also works using bpftrace v0.21.2 without -p option, though it doesn't work with -p option(need be exact path to the binary).

I'll apply changes as suggested. Thank you for the review!

jordikroon:
Thank you very much. Great work!

PR #5505, opened 2026-04-22
DATE_RFC7231 constant has been deprecated in PHP 8.5

jawira:
I updated the DateTimeInterface changelog since the DATE_RFC721 constant has been deprecated.

Source: <https://www.php.net/manual/en/migration85.deprecated.php#migration85.deprecated.date>


cweiske:
Thank you!

PR #5507, opened 2026-04-22
Fix several issues in versions.xml files

kamil-tekiela:
These files are manually maintained, and so they suffer from several issues: missing entries, unnecessary entries, typos, etc. 

- The first few commits fix typos and add a few missing entries that I was able to verify
- The rest removes entries that are definitely no longer useful to anyone

However, this is just a small patch. The number of missing entries is much larger. There are also hundreds of entries that don't serve any purpose. They can be grouped mostly into two types: inherited but not overridden methods, and aliases. None of them have a matching page in the documentation, which means that these entries aren't used by PHP manual at least. However, I was cautious about removing them as they can still offer some insight into other tools or human eyes.

### DIFF
--- a/reference/simplexml/versions.xml
+++ b/reference/simplexml/versions.xml
@@ -24,7 +24,8 @@
  <function name="simplexmlelement::savexml" from="PHP 5 &gt;= 5.2.0, PHP 7, PHP 8"/>
  <function name="simplexmlelement::xpath" from="PHP 5, PHP 7, PHP 8"/>
 
- <function name="simplexmlelement" from="PHP 8"/>
+ <function name="simplexmliterator" from="PHP 5, PHP 7, PHP 8"/>
+
  <function name="simplexmlelement::count" from="PHP 8"/>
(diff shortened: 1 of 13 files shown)


reference/simplexml/versions.xml
lacatoire:
```suggestion
<function name="simplexmliterator::count" from="PHP 8"/>
<function name="simplexmliterator::current" from="PHP 8"/>
<function name="simplexmliterator::getchildren" from="PHP 8"/>
<function name="simplexmliterator::haschildren" from="PHP 8"/>
<function name="simplexmliterator::key" from="PHP 8"/>
<function name="simplexmliterator::next" from="PHP 8"/>
<function name="simplexmliterator::rewind" from="PHP 8"/>
<function name="simplexmliterator::valid" from="PHP 8"/>
```

kamil-tekiela: I guess that makes sense. Do you know why we do that?

PR #5516, opened 2026-04-27
[idea] Merge constant aliases by using multiple term tags for a varlistentry

Girgias:
As a reply to https://github.com/php/doc-en/pull/5434#issuecomment-4142370139 I think improving the markup of aliased constants to something like this probably makes the most sense, as we don't duplicate wording nor need to specify X is an alias of Y. :)

### DIFF
--- a/appendices/reserved.constants.core.xml
+++ b/appendices/reserved.constants.core.xml
@@ -108,45 +108,29 @@
     </note>
    </listitem>
   </varlistentry>
-  <varlistentry xml:id="constant.zend-thread-safe">
-   <term>
+  <varlistentry>
+   <term xml:id="constant.zend-thread-safe">
     <constant>ZEND_THREAD_SAFE</constant>
     (<type>bool</type>)
    </term>
+   <term xml:id="constant.php-zts">
+    <constant>PHP_ZTS</constant>
+    (<type>bool</type>)
+   </term>
    <listitem>
     <simpara>
      Indicates whether the current build of PHP is thread safe.
     </simpara>
    </listitem>
   </varlistentry>
-  <varlistentry xml:id="constant.zend-debug-build">
-   <term>
+  <varlistentry>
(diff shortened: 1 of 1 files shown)


jordikroon:
From a reader perspective I am not sure if this is clear enough. Because the extra enter could also indicate that the first constant may not have a description. Even though if that were the case it would have had an extra enter.

Additionally `Error reporting constants` are also formatted this way, but they are not considered aliases.

My suggestion (this would keep them on the same line):

```xml
<varlistentry>
 <term>
  <constant xml:id="constant.zend-thread-safe">ZEND_THREAD_SAFE</constant>,
  <constant xml:id="constant.php-zts">PHP_ZTS</constant>
  (<type>bool</type>)
 </term>
 <listitem>
  <simpara>
   Indicates whether the current build of PHP is thread safe.
  </simpara>
 </listitem>
</varlistentry>
  ```

Girgias:
> From a reader perspective I am not sure if this is clear enough. Because the extra enter could also indicate that the first constant may not have a description. Even though if that were the case it would have had an extra enter.
> 
> Additionally `Error reporting constants` are also formatted this way, but they are not considered aliases.
> 
> My suggestion (this would keep them on the same line):
> 
> ```
> <varlistentry>
>  <term>
>   <constant xml:id="constant.zend-thread-safe">ZEND_THREAD_SAFE</constant>,
>   <constant xml:id="constant.php-zts">PHP_ZTS</constant>
>   (<type>bool</type>)
>  </term>
>  <listitem>
>   <simpara>
>    Indicates whether the current build of PHP is thread safe.
>   </simpara>
>  </listitem>
> </varlistentry>
> ```

Sorry for the late reply, that could also work and happy either way :)

jordikroon:
@kocsismate is this something you could try for #5434?

kocsismate:
@jordikroon Absolutely, that's what I would also prefer!

PR #5519, opened 2026-04-28
Clarify in_array() strict comparison behavior for PHP 8.0+

joshtrichards:
The current note for the `strict` parameter focuses primarily on a legacy string-to-integer comparison bug that was partially resolved in PHP 8.0.0. This can give users the false impression that non-strict mode is now "safe" in modern versions.

This PR updates the documentation to:
1. Clarify the specific change introduced in PHP 8.0.0 (restriction to numeric strings).
2. Explicitly mention the persistent "truthy" trap where boolean `true` loosely matches any non-empty string.
3. Strongly reinforce the recommendation for strict comparison to ensure type safety.

### DIFF
--- a/reference/array/functions/in-array.xml
+++ b/reference/array/functions/in-array.xml
@@ -55,12 +55,16 @@
       </para>
       <note>
        <para>
-        Prior to PHP 8.0.0, a <literal>string</literal> <parameter>needle</parameter> will match an array
-        value of <literal>0</literal> in non-strict mode, and vice versa.  That may lead to undesireable
-        results.  Similar edge cases exist for other types, as well.  If not absolutely certain of the
-        types of values involved, always use the <parameter>strict</parameter> flag to avoid unexpected behavior.
-       </para>
-      </note>
+        Prior to PHP 8.0.0, a <literal>string</literal> <parameter>needle</parameter> would loosely match
+        a numeric <parameter>haystack</parameter> value of <literal>0</literal>, and vice versa.
+        As of PHP 8.0.0, this behavior is restricted to numeric strings. However, non-strict mode
+        still employs loose comparison (<literal>==</literal>), which allows for matches across
+        unrelated types. For example, a boolean <constant>&true;</constant> value in the
+        <parameter>haystack</parameter> will match any non-empty <literal>string</literal>
+        <parameter>needle</parameter>. It is recommended to use the <parameter>strict</parameter>
+        flag to ensure type-safe results.
+        </para>
+       </note>
      </listitem>
     </varlistentry>
    </variablelist>
--- a/reference/array/functions/in-array.xml
+++ b/reference/array/functions/in-array.xml
@@ -55,12 +55,16 @@
       </para>
       <note>
        <para>
-        Prior to PHP 8.0.0, a <literal>string</literal> <parameter>needle</parameter> will match an array
-        value of <literal>0</literal> in non-strict mode, and vice versa.  That may lead to undesireable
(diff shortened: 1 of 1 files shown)


reference/array/functions/in-array.xml
```diff
@@ -55,12 +55,16 @@
...
-        value of <literal>0</literal> in non-strict mode, and vice versa.  That may lead to undesireable
-        results.  Similar edge cases exist for other types, as well.  If not absolutely certain of the
-        types of values involved, always use the <parameter>strict</parameter> flag to avoid unexpected behavior.
-       </para>
-      </note>
+        Prior to PHP 8.0.0, a <literal>string</literal> <parameter>needle</parameter> would loosely match
+        a numeric <parameter>haystack</parameter> value of <literal>0</literal>, and vice versa.
+        As of PHP 8.0.0, this behavior is restricted to numeric strings. However, non-strict mode
+        still employs loose comparison (<literal>==</literal>), which allows for matches across
+        unrelated types. For example, a boolean <constant>&true;</constant> value in the
```
lacatoire, line 62:
```suggestion
        unrelated types. For example, a boolean &true; value in the
```

reference/array/functions/in-array.xml
```diff
@@ -55,12 +55,16 @@
...
+        <parameter>haystack</parameter> will match any non-empty <literal>string</literal>
+        <parameter>needle</parameter>. It is recommended to use the <parameter>strict</parameter>
+        flag to ensure type-safe results.
+        </para>
+       </note>
```
lacatoire, line 67:
```suggestion
       </para>
      </note>
```

reference/array/functions/in-array.xml
```diff
@@ -55,12 +55,16 @@
...
-        results.  Similar edge cases exist for other types, as well.  If not absolutely certain of the
-        types of values involved, always use the <parameter>strict</parameter> flag to avoid unexpected behavior.
-       </para>
-      </note>
+        Prior to PHP 8.0.0, a <literal>string</literal> <parameter>needle</parameter> would loosely match
+        a numeric <parameter>haystack</parameter> value of <literal>0</literal>, and vice versa.
+        As of PHP 8.0.0, this behavior is restricted to numeric strings. However, non-strict mode
+        still employs loose comparison (<literal>==</literal>), which allows for matches across
+        unrelated types. For example, a boolean <constant>&true;</constant> value in the
+        <parameter>haystack</parameter> will match any non-empty <literal>string</literal>
```
lacatoire, line 63:
```suggestion
        <parameter>haystack</parameter> will match any truthy <literal>string</literal>
```

lacatoire: Tiny accuracy nit: `"0"` is technically a non-empty string but evaluates to `false` under PHP's loose comparison, so `in_array("0", [true])` returns `false` (see the [comparison table](https://www.php.net/manual/en/types.comparisons.php#types.comparisons.loose)). "Any truthy string" would be more precise.

reference/array/functions/in-array.xml
jordikroon: This should become `<simpara>` as well (instead of `<para>`

reference/array/functions/in-array.xml
jordikroon: FYI; This is the cause of why CI tests fail. Because `&true;` is an entity that already contains a `<constant>`

reference/array/functions/in-array.xml
kamil-tekiela, line 63:
```suggestion
        <parameter>haystack</parameter> will match any non-empty <type>string</type>
```

PR #5510, opened 2026-04-23
ci test: add sniffer for docbook (Do not merge!)

jordikroon:
Please ignore this. Testing to see how everything runs in CI.

Since this is an external tool (authored by me), it's probably better to discuss the use over the mailing list.

-> https://news-web.php.net/php.doc/969388652

PR #5524, opened 2026-04-29
CI: Add docbook-cs linting

jordikroon:
As announced here: https://news-web.php.net/php.doc/969388652
See for test implementation: https://github.com/php/doc-en/pull/5510

The sniffer includes:
 - 
This also makes check-whitespaces.xml obsolete, as the sniffer will check for that as well.

The sniffer also covers https://github.com/php/doc-en/pull/5445
 - `<simpara>` should be used instead of `<para>`
 - Exception or error class names wrapped in `<classname>` that should use `<exceptionname>` instead.
 - Whitespace checks (.github/workflows/check-whitespace.yml became obsolete)
 - Ensuring that when an element has both xml and xmlns attributes or xmlns variants, xml appears first (covers #5445)


.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,79 @@
+name: DocBook Lint
+
+on:
+  push:
+    branches: [master]
```
TimWolla: Use yaml syntax (see integrate.yml):

```suggestion
    branches:
      - master
```

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,79 @@
...
+
+jobs:
+  docbook-cs:
+    name: DocBook Style Check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout doc-en
+        uses: actions/checkout@v4
```
TimWolla: Latest versions, please.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: en
+          fetch-depth: 0
```
TimWolla: Do we need the full history?

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Upload report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: docbook-report
+          path: en/docbook-report.xml
+
+      - name: Setup reviewdog
+        if: always()
```
TimWolla:
```suggestion
        if: ${{ ! cancelled() }}
```
Same for the other `always()`.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: reviewdog/action-setup@v1
+        with:
+          reviewdog_version: latest
+
+      - name: Report violations via reviewdog
```
TimWolla: Do we need another external service here?

.github/workflows/docbook-cs.yaml
```diff
@@ -0,0 +1,79 @@
+name: DocBook Lint
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
```
TimWolla, line 17: Missing workflow permissions.

TimWolla (commented):
Some remarks for the workflow definition. Didn't look at the tool itself, but the premise sounds good to me.

.github/workflows/docbook-cs.yml
jordikroon: We need at least enough for to retrieve PR commits. I have set the limit now to 50 for now. Should be plenty enough for 99% of the PRs.

.github/workflows/docbook-cs.yml
jordikroon: Honestly I was expecting this to be a lot more complicated, but it turned out it was as simple as using workflow commands in the output. So the external service has been ditched.

We can still tweak the output to whatever we desire.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,135 @@
...
+        uses: "actions/upload-artifact@v7"
+        with:
+          name: "docbook-report"
+          path: "${{ matrix.language }}/docbook-report.xml"
+
+      - name: "Annotate PR with violations"
+        if: ${{ ! cancelled() }}
+        working-directory: "${{ matrix.language }}"
+        run: |
+          php -r '
```
TimWolla: You can likely avoid the extra quoting by setting PHP as a custom shell: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#custom-shell

Though can't your tool emit the GitHub Action metadata directly by itself? e.g. `--report=github`? Alternatively, if it's standard checkstyle output, the `setup-php` action can also install `cs2pr`.

.github/workflows/docbook-cs.yaml
```diff
@@ -0,0 +1,135 @@
+# https://docs.github.com/en/actions
+
```
TimWolla, line 2:
```suggestion
```

TimWolla (commented):
Some more quick GHA comments. Will need to think about the git / checkout usage more at a later point.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,135 @@
...
+        run: |
+          php doc-base/configure.php \
+            --disable-libxml-check \
+            --enable-xml-details \
+            --redirect-stderr-to-stdout \
+            --with-lang=${{ matrix.language }}
+
+      - name: "Install docbook-cs"
+        working-directory: "${{ matrix.language }}"
+        run: composer require jordikroon/docbook-cs
```
DanielEScherzer: should this be pinned to a version? otherwise new changes to the package can lead to breakage of CI here

.github/workflows/docbook-cs.yml
DanielEScherzer: I was going to suggest cs2pr - please use that, or put this PHP in a script that is invoked, rather than having it inlined

docbookcs.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<docbookcs xmlns="https://docbookcs.org/config">
```
DanielEScherzer: `docbookcs.org` is not currently registered (DNS_PROBE_FINISHED_NXDOMAIN)

.github/workflows/docbook-cs.yaml
jordikroon: I purposely added a newline here as this is consistent with `integrate.yaml`. If still preferred I can apply this change though.

docbookcs.xml
jordikroon: Changed this to a proper GitHub pages with schema validation.

.github/workflows/docbook-cs.yml
jordikroon: cs2pr is fine. The tool supports Checkstyle output, so I don't see a reason to also support Github annotations.

jordikroon:
I am planning to merge Docbook-cs in a few days as all feedback has been addressed. Please let me know if there are comments, objections or concerns regarding this new tool. So to say any feedback is welcome.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: |
+          php doc-base/configure.php \
+            --disable-libxml-check \
+            --enable-xml-details \
+            --redirect-stderr-to-stdout \
+            --with-lang=${{ matrix.language }}
+
+      - name: "Install docbook-cs"
+        working-directory: "${{ matrix.language }}"
+        run: composer require jordikroon/docbook-cs:^1.1 staabm/annotate-pull-request-from-checkstyle:^1.8
```
TimWolla: cs2pr can be installed with setup-php (which might have it cached) and doesn't require a manual composer installation. I recommend moving to that.

See: https://github.com/shivammathur/setup-php#wrench-tools-support

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: |
+          php doc-base/configure.php \
+            --disable-libxml-check \
+            --enable-xml-details \
+            --redirect-stderr-to-stdout \
+            --with-lang=${{ matrix.language }}
+
+      - name: "Install docbook-cs"
+        working-directory: "${{ matrix.language }}"
+        run: composer require jordikroon/docbook-cs:^1.1
```
lacatoire: `^1.1` means any future 1.x release lands on the next CI run. Pinning to an exact version (1.1.2) plus a Dependabot config could route bumps through reviewed PRs. Same suggestion for `cs2pr:^1.8`

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: "Run docbook-cs with diff"
+        working-directory: "${{ matrix.language }}"
+        run: |
+          set -o pipefail
+          git diff origin/${{ github.base_ref }}...HEAD | vendor/bin/docbook-cs \
+            --report=checkstyle \
+            --diff \
+            --no-colors > docbook-report.xml || true
```
lacatoire: The trailing `|| true` after docbook-cs swallows exit code 1 (violations) and 2 (runtime error) the same way. The tool returns 2 if the config fails to load or if a sniff throws, and in that case the report stays empty so cs2pr would silently report "no violations". Could we let exit 2 fail the job?

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,94 @@
...
+            fi
+            git fetch --deepen=100 origin ${{ github.base_ref }}
+          done
+
+      - name: "Setup PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "8.5"
+          extensions: "dom, libxml, simplexml"
+          tools: composer:v2, cs2pr:^1.8
```
TimWolla:
```suggestion
          tools: composer, cs2pr
```
Defaults are fine.

.github/workflows/docbook-cs.yml
TimWolla: Running on the latest version of docbook-cs is fine. It's a tool we build for our own use case - we control its behavior.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          set -o pipefail
+          code=0
+          git diff origin/${{ github.base_ref }}...HEAD | php ../docbook-cs/bin/docbook-cs \
+            --report=checkstyle \
+            --diff \
+            --no-colors > docbook-report.xml || code=$?
+          if [ "$code" -gt 1 ]; then
+            exit "$code"
+          fi
```
TimWolla: Is the special handling for the exit code even necessary? The annotation step runs on `! cancelled()`, so will run even if this step fails.

jordikroon: Yes you are right. The previous feedback comment confused me + that GH Actions is not my strongest. 
And since that is true, `|| true` was also wrong. So it will return the pure status code now at all times.

.github/workflows/docbook-cs.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: "Checkout php/docbook-cs"
+        uses: "actions/checkout@v6"
+        with:
+          path: "docbook-cs"
+          repository: "php/docbook-cs"
+
+      - name: "Fetch base branch with merge-base"
+        working-directory: "${{ matrix.language }}"
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth=50
```
TimWolla: I wanted to double-check what `github.base_ref` does. It is documented as:

> The base_ref or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

This means that it doesn't do the right thing for the `push:` trigger. Can you double-check on your fork with by removing the branch restriction on the `push:` trigger and observing the CI for the `feature/docbook-cs` branch *in your fork*?

jordikroon: I have been doing some tested and that variable is indeed empty on push.

Any thoughts on this commit before I implement it in this PR?
https://github.com/jordikroon/doc-en/pull/1/changes/ba77f57aa72e1d851a4a3ad485f969a23186f13e

In short. It fallbacks to `github.event.before`. And `github.event.pull_request.base.sha` also seems a little bit better as it will just pick the hash instead of the branch name. On push it will compare against the previous commit hash. On PR it will compare against the base sha.

TimWolla: Can you try `${{ case(github.event_name == 'pull_request', github.event.pull_request.base.sha, github.event_name == 'push', github.event.before, 'INVALID') }}` or similar for clarity? Other than that it looks good to me, I particularly add the added shell quoting (that would've been a follow-up request, depending on how your solution would've looked).

jordikroon: Thanks again. I have added this. Did some extra testing and it gave the expected results on both push and PR.

TimWolla (approved):

PR #5525, opened 2026-04-29
Change 'Old adult' to 'Adult' in match.xml

sprky0:
Removes lightly irritating ageist language from match syntax documentation.


TimWolla (approved):
The inconsistent use of comparison operators makes this example super weird anyways. I'd be in favor of replacing it entirely with something that is easier to follow.

Issue #5526, opened 2026-05-01
Incomplete statement about the Null Coalescing Operator


Krause-a:
From manual page: https://php.net/language.operators.comparison

---

In the Null Coalescing Operator section, the documentation states "The expression (expr1) ?? (expr2) evaluates to expr2 if expr1 is null, and expr1 otherwise." which is not fully correct. The next sentence states it suppresses notices and warning for non existent expr1, just like isset. The code block above has an "identical" example given of an array access situation with ?? and with if/else block using isset. This led me to believe that I could substitute that array access for any expression and it would still be identical. This is not the case. This next code block appears to be a proper equivalence example.
```PHP
<?php
// Example usage for: Null Coalesce Operator
$result = _expr_ ?? 'default';

// The above is identical to this if/else statement
if (isset(_expr_)) {
	$tmp_var = _expr_;
	if ($tmp_var !== null) {
		$result = $tmp_var;
	}
	else {
		$result = 'default';
	}
} else {
    $result = 'default';
}
?>
```

PR #5527, opened 2026-05-01
Specify Null Coalesce Operator isset Relation

Krause-a:
This PR addresses issue #5526

Make the `isset` relationship with `??` explicit to reduce potential confusion.

### DIFF
--- a/language/operators/comparison.xml
+++ b/language/operators/comparison.xml
@@ -462,7 +462,7 @@ echo 0 ?: 0 ?: 0 ?: 3, PHP_EOL; //3
 // Example usage for: Null Coalesce Operator
 $action = $_POST['action'] ?? 'default';
 
-// The above is identical to this if/else statement
+// The above is equivalent to this if/else statement
 if (isset($_POST['action'])) {
     $action = $_POST['action'];
 } else {
@@ -473,13 +473,12 @@ if (isset($_POST['action'])) {
     </programlisting>
    </example>
    The expression <literal>(expr1) ?? (expr2)</literal> evaluates to
-   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is
-   &null;, and <replaceable>expr1</replaceable> otherwise.
+   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is not set,
+   as determined by <function>isset</function>, or is &null;.
+   Otherwise, it evaluates to <replaceable>expr1</replaceable>.
   </para>
   <para>
-   In particular, this operator does not emit a notice or warning if the left-hand side
-   value does not exist, just like <function>isset</function>. This is especially
-   useful on array keys.
+   This operator is especially useful for safe array access.
   </para>
   <note>
    <simpara>
--- a/language/operators/comparison.xml
+++ b/language/operators/comparison.xml
@@ -473,12 +473,12 @@ if (isset($_POST['action'])) {
     </programlisting>
    </example>
    The expression <literal>(expr1) ?? (expr2)</literal> evaluates to
-   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is not set,
-   as determined by <function>isset</function>, or is &null;.
-   Otherwise, it evaluates to <replaceable>expr1</replaceable>.
+   <replaceable>expr1</replaceable> if <replaceable>expr1</replaceable> is set
+   and not &null;, otherwise <replaceable>expr2</replaceable>.
   </para>
   <simpara>
-   This operator is especially useful for safe array access.
+   In particular, this operator does not emit a notice or warning if the left-hand side
+   value is not set. This is especially useful on array keys.
   </simpara>
   <note>
    <simpara>
--- a/language/operators/comparison.xml
+++ b/language/operators/comparison.xml
@@ -477,9 +477,9 @@ if (isset($_POST['action'])) {
    as determined by <function>isset</function>, or is &null;.
    Otherwise, it evaluates to <replaceable>expr1</replaceable>.
   </para>
-  <para>
+  <simpara>
    This operator is especially useful for safe array access.
-  </para>
+  </simpara>
   <note>
    <simpara>
     Please note that the null coalescing operator is an expression, and that it


language/operators/comparison.xml
```diff
@@ -473,13 +473,12 @@ if (isset($_POST['action'])) {
     </programlisting>
    </example>
    The expression <literal>(expr1) ?? (expr2)</literal> evaluates to
-   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is
-   &null;, and <replaceable>expr1</replaceable> otherwise.
+   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is not set,
+   as determined by <function>isset</function>, or is &null;.
+   Otherwise, it evaluates to <replaceable>expr1</replaceable>.
   </para>
   <para>
```
lacatoire:
```suggestion
  <simpara>
```

language/operators/comparison.xml
```diff
@@ -473,13 +473,12 @@ if (isset($_POST['action'])) {
...
-   In particular, this operator does not emit a notice or warning if the left-hand side
-   value does not exist, just like <function>isset</function>. This is especially
-   useful on array keys.
+   This operator is especially useful for safe array access.
   </para>
```
lacatoire:
```suggestion
  </simpara>
```

language/operators/comparison.xml
```diff
@@ -473,13 +473,12 @@ if (isset($_POST['action'])) {
...
-   &null;, and <replaceable>expr1</replaceable> otherwise.
+   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is not set,
+   as determined by <function>isset</function>, or is &null;.
+   Otherwise, it evaluates to <replaceable>expr1</replaceable>.
   </para>
   <para>
-   In particular, this operator does not emit a notice or warning if the left-hand side
-   value does not exist, just like <function>isset</function>. This is especially
-   useful on array keys.
+   This operator is especially useful for safe array access.
```
AllenJB: I don't think the note about warnings / notices should be removed. This is a behavior that users should be aware of.

language/operators/comparison.xml
```diff
@@ -473,13 +473,12 @@ if (isset($_POST['action'])) {
     </programlisting>
    </example>
    The expression <literal>(expr1) ?? (expr2)</literal> evaluates to
-   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is
-   &null;, and <replaceable>expr1</replaceable> otherwise.
+   <replaceable>expr2</replaceable> if <replaceable>expr1</replaceable> is not set,
+   as determined by <function>isset</function>, or is &null;.
+   Otherwise, it evaluates to <replaceable>expr1</replaceable>.
```
AllenJB: It should be noted that `isset()` returns `false` is the value is defined but set to `null`.

While I think the wording here could be improved to make the behavior clearer, I think the suggested change here potentially causes confusion about the behavior of isset.

It might be better to omit mention of isset here entirely and just say:

>  if expr1 is not defined or is &null;.```

Krause-a: Thank you for the feedback!

The driving factor for the inclusion of `isset()` is that both the `isset` check and the `null` check appear to be performed separately. I understand that this is redundant for array checks and variable checks. However, the seemingly redundant null check matters for object properties when using the `__isset` magic method.
```PHP
class MyClass { # Minimal Example Class
    public function __isset($name) {
        return true;
    }
    public function __get($name) {
        return null;
    }
}
$my_obj = new MyClass;
# Ex.1
var_dump($my_obj->fake ?? 'rhs'); # rhs
# Ex.2
var_dump(isset($my_obj->fake) ? $my_obj->fake : 'rhs'); # null
# Ex.3
var_dump(isset($my_obj->fake) && $my_obj->fake !== null ? $my_obj->fake : 'rhs'); # rhs
```
You can try the different combinations with `__isset` returning `true` and `false` and `__get` returning `null` or `123`.

> if expr1 is not defined or is &null;

My worry is that this will be confused with _only_ an `isset` check, since, in normal circumstances, `isset` will do the null check, making the second null check redundant. The documentation on `isset` starts with "Determine if a variable is declared and is different than null." The [`isset`](https://www.php.net/manual/en/function.isset.php) documentation sounds very similar. It could be believed that it is, in fact, exactly the same. The [PHP 7.0 new features announcement](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op) includes the "... if it exists and is not null..." wording and references `??` as syntax sugar for an `isset` ternary. Do you have an alternative wording that would still have the `isset` and `null` check as distinct steps in the `??` operator?

language/operators/comparison.xml
Krause-a: Happy to do it. Thanks!

language/operators/comparison.xml
AllenJB: So it turns out this is a known oddity of `isset()` and the magic methods.

My personal feeling is that if this behavior is documented, it belongs under either (or both) the `isset()` or magic methods documentation.

The only change I would suggest to the null coalesce documentation, if any, is to say:

> if expr1 is not set or is null

Related:

* https://github.com/php/php-src/issues/12695
* https://externals.io/message/130725 / https://wiki.php.net/rfc/exists-magic-method

Krause-a: Oh, wow! Thank you for the links! In light of this information, I agree with a less explicit reference to `isset`. Let me know what you think :)

language/operators/comparison.xml
Krause-a: Okay, note restored. Thanks!

lacatoire (approved):

kamil-tekiela:
I'm sorry, but I disagree with this change. An expression cannot be set! An expression must evaluate, and if it evaluates to NULL, then the right-hand side expression is used. Yes, it is difficult to represent this clearly, but in this regard, your PR doesn't improve the situation. 

The fact that it doesn't emit notices or warnings is already mentioned in the next sentence. So I don't think anything is missing in the way it is currently presented. 

As you are the same person who raised the linked issue, could you explain clearly what is it that's missing from that page?

kamil-tekiela:
@lacatoire What are we trying to improve here? I still don't see an improvement. It's just unnecessary change in my opinion.

lacatoire:
> @lacatoire What are we trying to improve here? I still don't see an improvement. It's just unnecessary change in my opinion.

OK closing it so

PR #5529, opened 2026-05-02
Clarify http_build_query() behavior with objects and __toString()

0xRashad:
Addresses legacy bug 66966 by adding a note and an example demonstrating the need for explicit string casting.

https://bugs.php.net/bug.php?id=66966

### DIFF
--- a/reference/url/functions/http-build-query.xml
+++ b/reference/url/functions/http-build-query.xml
@@ -40,6 +40,13 @@
        If <parameter>data</parameter> is an object, then only public
        properties will be incorporated into the result.
       </para>
+      <note>
+       <para>
+        The <link linkend="language.oop5.magic.tostring">__toString()</link> magic method is
+        not called when an object is evaluated. If you want the query string to use the
+        string representation of an object, you must explicitly cast the object to a string.
+       </para>
+      </note>
      </listitem>
     </varlistentry>
     <varlistentry>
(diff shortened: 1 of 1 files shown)


reference/url/functions/http-build-query.xml
```diff
@@ -40,6 +40,13 @@
        If <parameter>data</parameter> is an object, then only public
        properties will be incorporated into the result.
       </para>
+      <note>
+       <para>
+        The <link linkend="language.oop5.magic.tostring">__toString()</link> magic method is
+        not called when an object is evaluated. If you want the query string to use the
+        string representation of an object, you must explicitly cast the object to a string.
+       </para>
+      </note>
```
jordikroon, line 49:
```suggestion
      <note>
       <simpara>
        The <link linkend="language.oop5.magic.tostring">__toString()</link> magic method is
        not called when an object is evaluated. To use the string representation of an
        object in the query string, the object must be explicitly cast to a string.
       </simpara>
      </note>
```

jordikroon:
There has been an attempt to address this bug in https://github.com/php/php-src/pull/10235 but it has gone stale.

So documenting this behaviour seems fine to me.

reference/url/functions/http-build-query.xml
jordikroon: Direct address (the use of "you"), should be avoided in the manual because it is a technical document.

0xRashad:
Thanks for the helpful context! I completely agree, given the BC implications of changing the long standing behaviour the best way forward is to clarify the documentation to reflect this behaviour. I've applied your formatting suggestions. Let me know if there is anything else needed!

reference/url/functions/http-build-query.xml
```diff
@@ -40,6 +40,13 @@
        If <parameter>data</parameter> is an object, then only public
        properties will be incorporated into the result.
       </para>
+      <note>
+       <simpara>
+        The <link linkend="language.oop5.magic.tostring">__toString()</link> magic method is
```
jordikroon:
```suggestion
        The <link linkend="object.tostring">__toString()</link> magic method is
```

reference/url/functions/http-build-query.xml
```diff
@@ -260,6 +267,44 @@ echo http_build_query($parent);
    <screen>
 <![CDATA[
 pub=publicParent&pub_bar%5Bpub%5D=publicChild
+]]>
+   </screen>
+  </example>
+
+  <example>
+   <title>Using <function>http_build_query</function> with objects containing
+    <link linkend="language.oop5.magic.tostring">__toString()</link>
```
jordikroon:
```suggestion
    <link linkend="object.tostring">__toString()</link>
```

jordikroon (changes requested):
Thank you for addressing the reported issues. I just noticed isn't `language.oop5.magic.tostring` really used for linking to `__toString()`. In that case `object.tostring` is preferred.

0xRashad:
Thanks for the tip! I have updated as suggested.

jordikroon (approved):

jordikroon:
Good work. Thank you!

PR #5528, opened 2026-05-02
`gc_collect_cycles()`: add changelog for PHP 8.5

DanielEScherzer:

PR #5535, opened 2026-05-04
Sockets: add changelog entries for PHP 8.5 changes

lacatoire:
PHP 8.5 brings several Sockets changes already documented in
\`appendices/migration85\` but missing from the per-function changelogs.

BC breaks:
- socket_bind, socket_create_listen, socket_sendto: ValueError on out-of-range port
- socket_addrinfo_lookup: TypeError/ValueError on hints array values
- socket_set_option: exception on multicast misuse + ValueError on non AF_INET/AF_INET6

New features:
- socket_create: AF_PACKET family support
- socket_getsockname: interface index for AF_PACKET sockets

Same spirit as #5528.

### DIFF
--- a/reference/sockets/constants.xml
+++ b/reference/sockets/constants.xml
@@ -44,7 +44,22 @@
    </term>
    <listitem>
     <simpara>
-     Available as of PHP 8.3.0 (FreeBSD only)
+     Socket address family for the FreeBSD <literal>divert(4)</literal>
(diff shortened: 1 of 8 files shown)


jordikroon:
Besides adding a changelog entry we should document `AF_PACKET` as well (Linux only). Would you mind also adding`AF_DIVERT` which was added in PHP 8.3 (FreeBSD only) while were are revising this.

PR #1630, opened 2022-06-16
Add note explaining the bitmap combination possiblities

shaedrich:

PR #5537, opened 2026-05-05
Use constants for int mask combinations of `debug_backtrace()`

shaedrich:
Follow-up to #1630

grafik

### DIFF
--- a/reference/errorfunc/functions/debug-backtrace.xml
+++ b/reference/errorfunc/functions/debug-backtrace.xml
@@ -67,13 +67,13 @@
              <entry><code>debug_backtrace(1)</code></entry>
             </row>
             <row>
-             <entry><code>debug_backtrace(0)</code></entry>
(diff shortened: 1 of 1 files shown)


reference/errorfunc/functions/debug-backtrace.xml
```diff
@@ -67,17 +67,23 @@
...
+             <entry morerows="2" valign="middle">
               Omits index <literal>"object"</literal> <emphasis>and</emphasis> index <literal>"args"</literal>.
              </entry>
             </row>
+            <row>
+             <entry><code>debug_backtrace(~DEBUG_BACKTRACE_PROVIDE_OBJECT | DEBUG_BACKTRACE_IGNORE_ARGS)</code></entry>
+            </row>
             <row>
              <entry><code>debug_backtrace(2)</code></entry>
             </row>
```
TimWolla: These numeric entries (except for `0`) should be removed. Folks must use the constants, not their value.

shaedrich: Will do

TimWolla (approved):

PR #5539, opened 2026-05-05
Change <para> to <simpara> in phar extension

lacatoire:
Same conversion as #5536 (zip) and #5522 (language-snippets), now on the `phar` extension. Replaces `<para>` with `<simpara>` everywhere the body is purely inline; the 248 remaining `<para>` are kept because they wrap block children (`<example>`, `<programlisting>`, `<itemizedlist>`, `<note>`, `<variablelist>`, etc.) which `<simpara>` cannot contain. Inner paras nested inside a kept outer para are converted when eligible.

104 files touched, 521 conversions, no semantic change.

### DIFF
--- a/reference/phar/Phar.xml
+++ b/reference/phar/Phar.xml
@@ -9,10 +9,10 @@
 <!-- {{{ Phar intro -->
   <section xml:id="phar.intro">
    &reftitle.intro;
-   <para>
+   <simpara>
     The Phar class provides a high-level interface to accessing and creating
     phar archives.
-   </para>
+   </simpara>
   </section>
 <!-- }}} -->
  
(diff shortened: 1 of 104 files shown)


jordikroon:
Converting `<para>` to `<simpara>` throughout would require a lot of energy which gives little benefit. Additionally this would add a lot of extra work for the translators of other supported languages. That also needs to be reviewed. 

The reason #5522 was merged into the master is because it would otherwise give a lot of extra violations when we the Docbook sniffer will be used for checking Merge Requests. 

I would therefore like to ask that <para> only be changed to <simpara> when the surrounding copy is already being updated. In other words, please avoid making standalone changes whose only purpose is replacing <para> with <simpara>.

PR #5544, opened 2026-05-06
Add omission hints to return values

shaedrich:
image

### DIFF
--- a/reference/errorfunc/functions/debug-backtrace.xml
+++ b/reference/errorfunc/functions/debug-backtrace.xml
@@ -167,7 +167,7 @@
       </row>
       <row>
        <entry>type</entry>
-       <entry><<type>string</type></entry>
+       <entry><type>string</type></entry>
        <entry>
         The current call type. If a method call, "-&gt;" is returned. If a static
         method call, "::" is returned. If a function call, nothing is returned.
--- a/reference/errorfunc/functions/debug-backtrace.xml
+++ b/reference/errorfunc/functions/debug-backtrace.xml
@@ -162,7 +162,7 @@
        <entry>object</entry>
        <entry><type>object</type></entry>
        <entry>
-        The current <link linkend="language.oop5">object</link>.
+        The current <link linkend="language.oop5">object</link> if <code>DEBUG_BACKTRACE_PROVIDE_OBJECT</code> or equivalent is passed.
        </entry>
       </row>
       <row>
@@ -178,7 +178,7 @@
        <entry><type>array</type></entry>
        <entry>
         If inside a function, this lists the functions arguments.  If
-        inside an included file, this lists the included file name(s).
+        inside an included file, this lists the included file name(s). Unless <code>DEBUG_BACKTRACE_PROVIDE_OBJECT</code> or equivalent is passed.
        </entry>
       </row>
      </tbody>
--- a/reference/errorfunc/functions/debug-backtrace.xml
+++ b/reference/errorfunc/functions/debug-backtrace.xml
@@ -160,14 +160,14 @@
       </row>
       <row>
        <entry>object</entry>
-       <entry><type class="union"><type>array</type><type>null</type></type></entry>
+       <entry><type>object</type></entry>
        <entry>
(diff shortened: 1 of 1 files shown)


reference/errorfunc/functions/debug-backtrace.xml
```diff
@@ -162,7 +162,7 @@
        <entry>object</entry>
        <entry><type>object</type></entry>
        <entry>
-        The current <link linkend="language.oop5">object</link>.
+        The current <link linkend="language.oop5">object</link> if <code>DEBUG_BACKTRACE_PROVIDE_OBJECT</code> or equivalent is passed.
```
jordikroon:
```suggestion
        The current <link linkend="language.oop5">object</link> if <constant>DEBUG_BACKTRACE_PROVIDE_OBJECT</constant> is given.
```

reference/errorfunc/functions/debug-backtrace.xml
```diff
@@ -178,7 +178,7 @@
        <entry><type>array</type></entry>
        <entry>
         If inside a function, this lists the functions arguments.  If
-        inside an included file, this lists the included file name(s).
+        inside an included file, this lists the included file name(s). Unless <code>DEBUG_BACKTRACE_IGNORE_ARGS</code> or equivalent is passed.
```
jordikroon:
```suggestion
        inside an included file, this lists the included file name(s). Unless <constant>DEBUG_BACKTRACE_IGNORE_ARGS</constant> is given.
```

jordikroon (changes requested):
There isn't really an equivalent. So we can omit that.

reference/errorfunc/functions/debug-backtrace.xml
shaedrich: My train of thought was, there are several ways to get the `object` index populated:
* Function is called without arguments
* `DEBUG_BACKTRACE_PROVIDE_OBJECT` is passed
* `1` is passed

Same for `args`:
* Function is called without arguments
* `DEBUG_BACKTRACE_PROVIDE_OBJECT` or `~DEBUG_BACKTRACE_PROVIDE_OBJECT` is passed
* `1` or `~2` is passed
* `~DEBUG_BACKTRACE_IGNORE_ARGS` or `-3` passed

But I can remove that if you want

shaedrich:
> Do you have a working example where object is null? My assumption is that if it would be null it would be omitted.

Yeah, it's not exactly right, I'd be omitted, which, aside from emitting a warning if trying to access it, is _treated as_ `null` as well as being considered `null` by `is_null()` and even in strict comparison with `null`—how would you describe that as a type?

_(in reply to https://github.com/php/doc-en/pull/5545#issuecomment-4392282517)_

reference/errorfunc/functions/debug-backtrace.xml
jordikroon: Yes however all of these state that DEBUG_BACKTRACE_PROVIDE_OBJECT are given or not given. As by default DEBUG_BACKTRACE_PROVIDE_OBJECT is given.

We should not suggest passing integers directly. We should always suggest the use of constants.

jordikroon:
> > Do you have a working example where object is null? My assumption is that if it would be null it would be omitted.
> 
> Yeah, it's not exactly right, I'd be omitted, which, aside from emitting a warning if trying to access it, is _treated as_ `null` as well as being considered `null` by `is_null()` and even in strict comparison with `null`—how would you describe that as a type?
> 
> _(in reply to [#5545 (comment)](https://github.com/php/doc-en/pull/5545#issuecomment-4392282517))_

It wouldn't be right to change the signature because of that. PHP doesn't have an undefined type like javascript. So if it isn't there, it isn't there. The table already explicitly mentions that `Possible returned elements from`. Indicating that if may or may not return all values.

The types should stay `object` and `string`. They have also wrongfully been changed to `array` (+ `null`).

jordikroon (approved):
Looks good. Thank you!

shaedrich:
You're welcome! Thanks for the review!

PR #5545, opened 2026-05-06
Mark optional array properties as such

shaedrich:
image

### DIFF
--- a/reference/errorfunc/functions/debug-backtrace.xml
+++ b/reference/errorfunc/functions/debug-backtrace.xml
@@ -162,7 +162,7 @@
        <entry>object</entry>
        <entry><type>object</type></entry>
        <entry>
-        The current <link linkend="language.oop5">object</link>.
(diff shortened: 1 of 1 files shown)


jordikroon:
Do you have a working example where object is `null`? My assumption is that if it would be null it would be omitted.

Given that you created 3 MRs for the same function call. It would be much more helpful to bundle them into 1 single MR. The contents of the commits are rather small.

shaedrich:
> Do you have a working example where object is null? My assumption is that if it would be null it would be omitted.

Yeah, it's not exactly right, I'd be omitted, which, aside from emitting a warning if trying to access it, is _treated as_ `null` as well as being considered `null` by `is_null()` and even in strict comparison with `null`—how would you describe that as a type?

> Given that you created 3 MRs for the same function call. It would be much more helpful to bundle them into 1 single MR. The contents of the commits are rather small.

I'm used to making encapsulated PRs in other projects because that's what some maintainers want but I can merge them together. Sorry if that caused noise.

shaedrich:
Closed in favor of #5544

PR #5038, opened 2025-12-13
[PHP 8.5] Add new tokens, `T_VOID_CAST` `T_PIPE`

sasezaki:
<strike>WIP - (void) cast needs `#[\NoDiscard]` documentation in order to link<</strike>

PR #5546, opened 2026-05-07
Add (void) casting syntax for PHP 8.5

jordikroon:
Added the (void) cast to indicate that not using a value is intentional. The (void) cast has no effect on the program's execution by itself, but it can be used to suppress warnings emitted by #[\NoDiscard] and possibly also diagnostics emitted by external IDEs or static analysis tools. RFC: https://wiki.php.net/rfc/marking_return_value_as_important

https://github.com/php/doc-en/pull/5038/ depends on this.

### DIFF
--- a/language/types/type-juggling.xml
+++ b/language/types/type-juggling.xml
@@ -317,6 +317,12 @@ var_dump($bar);
    <member><literal>(unset)</literal> - cast to <type>NULL</type></member>
   </simplelist>
 
+  <simpara>
+   The <link linkend="language.types.void.casting"><literal>(void)</literal></link>
+   cast is also available as of PHP 8.5.0, but it is not a value conversion.
+   It is used as a statement to explicitly discard the result of an expression.
+  </simpara>
+
   <warning>
    <para>
     <literal>(integer)</literal> is an alias of the <literal>(int)</literal> cast.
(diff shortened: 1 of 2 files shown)


language/types/void.xml
```diff
@@ -17,6 +17,39 @@
...
  </note>
 
+ <sect2 xml:id="language.types.void.casting">
+  <title>Discarding a value with <literal>(void)</literal></title>
+
+  <simpara>
+   The <literal>(void)</literal> syntax may be used to explicitly discard the
+   result of an expression. This is useful to indicate that ignoring a return
+   value is intentional, especially when calling a function or method marked
+   with the <link linkend="class.nodiscard"><classname>NoDiscard</classname></link>
```
TimWolla:
```suggestion
   with the <classname>NoDiscard</classname>
```

I don't think the explicit link is necessary.

language/types/type-juggling.xml
```diff
@@ -422,10 +428,17 @@ if ($fst === $str) {
...
+    <member><link linkend="language.types.void.casting">Discarding a value with <literal>(void)</literal></link></member>
     <member><link linkend="types.comparisons">The type comparison tables</link></member>
    </simplelist>
   </para>
 
+  <simpara>
+   The <literal>(void)</literal> syntax is also available as of PHP 8.5.0,
+   but it is not a value conversion. It is used as a statement to explicitly
+   discard the result of an expression.
+  </simpara>
```
TimWolla: This is redundant with the one above.

TimWolla (changes requested):

language/types/void.xml
jordikroon: You are right. It even breaks official HTML syntax (a within an a tag).

TimWolla (approved):

PR #5552, opened 2026-05-12
Improve Docker build process, remove  --disable-segfault-error

jordikroon:
https://github.com/php/doc-base/pull/285 removed `--disable-segfault-error`. Dockerfile was still passing this argument resulting in a warning. Thus we should remove it as well.

Also altered `Makefile` to write a file hash to .docker/built instead. If we don't this warning will still pop unless we manually remove that file.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -18,5 +18,13 @@ RUN echo 'memory_limit = 512M' >> /usr/local/etc/php/conf.d/local.ini
 
 ENV FORMAT=xhtml
 
-CMD php doc-base/configure.php --disable-segfault-error && \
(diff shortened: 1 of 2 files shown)


kamil-tekiela:
I'm not familiar with Docker, so I can't review it.

jordikroon:
> I'm not familiar with Docker, so I can't review it.

No worries. Much appreciated to let me know.

cweiske:
The CMD change is fine, but I don't like the Makefile change. make lets you add file names as dependencies to tasks, and when the file changes, it will re-run the task. Adding the Dockerfile as dependency should fix that without manually comparing hashes.

jordikroon:
Thank you for checking @cweiske. I didn't know it was that simple. But it seems to do the trick.

cweiske:
Thanks.

PR #5553, opened 2026-05-14
Remove SimpleXML from bool false examples

prateekbhujel:
The bool casting page currently says that SimpleXML objects created from empty elements without attributes are false.

That does not match the existing SimpleXML behavior: an existing empty element casts to true, while a missing element casts to false. The php-src discussion for GH-21583 settled on keeping the current behavior and documenting it instead of changing SimpleXML's bool cast.

This keeps the internal-object note, but leaves GMP zero as the example.

Related: php/php-src#21583, php/php-src#21590, php/php-src#22042

Validation:

```
git diff --check
```

### DIFF
--- a/language/types/boolean.xml
+++ b/language/types/boolean.xml
@@ -113,9 +113,7 @@ if ($show_separators) {
    <listitem>
     <simpara>
      Internal objects that overload their casting behaviour to bool.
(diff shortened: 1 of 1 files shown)


devnexen (approved):

devnexen:
Thanks !

PR #5563, opened 2026-05-16
Update canbepassedbyvalue.xml

uuf6429:
Fix inaccurate return type description as per:
https://github.com/php/php-src/blob/master/ext/reflection/php_reflection.c#L2872 (master)

It seems it might have returned null in the past:
https://github.com/php/php-src/blob/PHP-7.3.19/ext/reflection/php_reflection.c#L2690 (PHP 7.3).

How do we best handle this?

### DIFF
--- a/reference/reflection/reflectionclass/gettraits.xml
+++ b/reference/reflection/reflectionclass/gettraits.xml
@@ -12,9 +12,9 @@
    <modifier>public</modifier> <type>array</type><methodname>ReflectionClass::getTraits</methodname>
    <void/>
   </methodsynopsis>
-  <para>
+  <simpara>
    Get the array of traits used by this class.
-  </para>
+  </simpara>
 
  </refsect1>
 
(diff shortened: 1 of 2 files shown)


jordikroon:
Removing this line would be correct, as this method never officially returned `null`. And for the reasons it would have returned `null`, it would also have returned a warning.

While you're at it, could you also convert this `<para>` to a `<simpara>` and remove the same line from [ReflectionClass::getTraits()](https://www.php.net/manual/en/reflectionclass.gettraits.php)?

Before PHP 8, passing an argument when the signature did not allow one resulted in a warning and a null return value. Since PHP 8, an `ArgumentCountError` is thrown instead.

jordikroon:
Thank you!

uuf6429:
@jordikroon quick question - are other languages handled in a similar manner, with a PR in their respective repo? (or is there some quicker way to mass-updated them?)

jordikroon:
You could create a PR for the other languages if you are willing to do so. Those are more than welcome.

Generally speaking the translations follow the upstream (en) and are managed specialised contributors.

PR #5568, opened 2026-05-18
The grapheme_strlen function returns null on encoding failures

### LINKED ISSUE #5554: grapheme_strlen(): Return Values section does not mention null
### Affected page

https://www.php.net/manual/en/function.grapheme-strlen.php

### Issue description

The `grapheme_strlen()` manual page currently shows the signature as:

```php
grapheme_strlen(string $string): int|false|null
```

However, the Return Values section says:

```text
The length of the string on success, or false on failure.
```

This appears to be incomplete because grapheme_strlen() can return null when the input string is not valid UTF-8.

The Parameters section already says that the argument:

```text
must be a valid UTF-8 string.
```

So the Return Values section should also explain what happens when this precondition is not met.

### Steps to reproduce

Code:
```php
<?php

$text = "\xFF";

$count = grapheme_strlen($text);

if ($count === null || $count === false) {
    $code = intl_get_error_code();

    throw new RuntimeException(sprintf(
        'grapheme_strlen() failed: %s (%d)',
        intl_error_name($code),
        $code
    ));
}

echo $count;
```

Result: 
```bash
PHP Fatal error:  Uncaught RuntimeException: grapheme_strlen() failed: U_INVALID_CHAR_FOUND (10) in ...
```

### Suggested fix

The Return Values section should mention both null and false.

A minimal wording change could be:

```text
- The length of the string on success, or false on failure.
+ The length of the string on success, null if string is not a valid UTF-8 string, or false on failure.
```

chschneider:
The documentation only mentions false for grapheme_strlen failures but in reality it also returns null for encoding failures (which is the more common failure case).
Mention this additional return value in the corresponding section.

Fixes #5554

### DIFF
--- a/reference/intl/grapheme/grapheme-strlen.xml
+++ b/reference/intl/grapheme/grapheme-strlen.xml
@@ -37,7 +37,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   The length of the string on success, &return.falseforfailure;. 
+   The length of the string on success,  &null; or &false; on failure.
   </para>
  </refsect1>
  
--- a/reference/intl/grapheme/grapheme-strlen.xml
+++ b/reference/intl/grapheme/grapheme-strlen.xml
@@ -36,9 +36,9 @@
  
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
-  <para>
-   The length of the string on success,  &null; or &false; on failure.
-  </para>
+  <simpara>
+   The length of the string on success, &null; or &false; on failure.
+  </simpara>
  </refsect1>
  
  <refsect1 role="examples">


reference/intl/grapheme/grapheme-strlen.xml
```diff
@@ -37,7 +37,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
-   The length of the string on success, &return.falseforfailure;. 
+   The length of the string on success,  &null; or &false; on failure.
```
lacatoire:
```suggestion
   The length of the string on success, &null; if <parameter>string</parameter> is not valid UTF-8, or &false; on failure.
```

reference/intl/grapheme/grapheme-strlen.xml
```diff
@@ -37,7 +37,7 @@
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
   <para>
```
lacatoire:
```suggestion
  <simpara>
```

reference/intl/grapheme/grapheme-strlen.xml
```diff
@@ -37,7 +37,7 @@
...
-   The length of the string on success, &return.falseforfailure;. 
+   The length of the string on success,  &null; or &false; on failure.
   </para>
```
lacatoire:
```suggestion
  </simpara>
```

reference/intl/grapheme/grapheme-strlen.xml
lacatoire: It seems php-src returns null only when [the input isn't valid UTF-8 (the UTF-8 to UTF-16 conversion fails)](https://github.com/php/php-src/blob/master/ext/intl/grapheme/grapheme_string.cpp#L71), and false when [grapheme splitting fails](https://github.com/php/php-src/blob/master/ext/intl/grapheme/grapheme_string.cpp#L83).

chschneider: While this is technically correct I'm not sure this is guaranteed as there is currently (as far as I can see) no test in php-src which ensures this behaviour and the implementation could change.
That's why I opted to not specify the cause of the different return values.

Do you think we should still add it to the documentation?

reference/intl/grapheme/grapheme-strlen.xml
chschneider: I assume this is just a general cleanup, not directly related to the text change, correct?

reference/intl/grapheme/grapheme-strlen.xml
lacatoire: good point, no test guarantees it and it could change. Let's keep it generic, your version works

reference/intl/grapheme/grapheme-strlen.xml
jordikroon: Yes that is correct. The general rule of thumb is to change the outer `<para>` to `<simpara>` if possible.

jordikroon:
LGTM. Thank you!

PR #5574, opened 2026-05-22
Fix spelling typos across the manual

lacatoire:
This fixes a batch of small spelling and grammar typos across the manual (its/it's, latter, layer, equivalent, localhost, etc.).

Text-only, no behavioral changes.


ndossche (approved):

jordikroon (approved):
Just a minor nit that we don't need to use `&apos;`. But since it was already there it's fine to leave it as.

PR #5577, opened 2026-05-23
Outdated note in example 10 on types arrays page

### LINKED ISSUE #4788: Outdated note in Example \#10 on `Types > Arrays` page
**Affected page**

[https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.modifying](https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.modifying)

**Problem**

The Note directly beneath **Example \#10** still says:

> As mentioned above, if no key is specified, the maximum of the existing \[int] indices is taken, and the new key will be that maximum value plus 1 **(but at least 0)**.

That parenthetical describes behaviour **prior to PHP 8.3.0**.
Since 8.3 the next automatic key after a negative one is *n + 1*, exactly as Example \#7 and its warning already state.

**Suggested fix**

Remove `(but at least 0)` from the sentence and replace Note in Example \#10 with this:

```
As mentioned above, if no key is specified, PHP uses the next integer key: the
maximum existing integer index + 1. If the array currently has no integer
indices, the key will be 0 (zero). For the historical behaviour prior to
PHP 8.3.0—where appending after a negative key could yield 0 (see Example #7).
```

Narkunan:
Fix: [https://github.com/php/doc-en/issues/4788](https://github.com/php/doc-en/issues/4788)


Narkunan:
Branch not updated

PR #5578, opened 2026-05-23
Updated notes for example 10 in arrays page

### LINKED ISSUE #4788: Outdated note in Example \#10 on `Types > Arrays` page

Narkunan:
Fix: [https://github.com/php/doc-en/issues/4788](https://github.com/php/doc-en/issues/4788)

### DIFF
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -534,10 +534,10 @@ var_dump($arr);
 
    <note>
     <para>
-     As mentioned above, if no key is specified, PHP uses the next integer key: the
-     maximum existing integer index + <literal>1</literal>. If the array currently has no integer
-     indices, the key will be <literal>0</literal> (zero). For the historical behaviour prior to
-     <literal>PHP 8.3.0</literal> — where appending after a negative key could yield <literal>0</literal> — see Example #7.
+      If no key is specified, the key will be the
+      maximum existing integer index + <literal>1</literal>. If the array has no positive
+      integer indices, the key will be <literal>0</literal>.
+      As of PHP 8.3.0, it can also be a negative integer.
     </para>
 
     <para>
--- a/language/types/array.xml
+++ b/language/types/array.xml
@@ -534,10 +534,10 @@ var_dump($arr);
 
    <note>
     <para>
-     As mentioned above, if no key is specified, the maximum of the existing
-     <type>int</type> indices is taken, and the new key will be that maximum
-     value plus 1 (but at least 0). If no <type>int</type> indices exist yet, the key will
-     be <literal>0</literal> (zero).
+     As mentioned above, if no key is specified, PHP uses the next integer key: the
+     maximum existing integer index + <literal>1</literal>. If the array currently has no integer
+     indices, the key will be <literal>0</literal> (zero). For the historical behaviour prior to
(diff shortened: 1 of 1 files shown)


language/types/array.xml
```diff
@@ -534,10 +534,10 @@ var_dump($arr);
...
    <note>
     <para>
-     As mentioned above, if no key is specified, the maximum of the existing
-     <type>int</type> indices is taken, and the new key will be that maximum
-     value plus 1 (but at least 0). If no <type>int</type> indices exist yet, the key will
-     be <literal>0</literal> (zero).
+     As mentioned above, if no key is specified, PHP uses the next integer key: the
+     maximum existing integer index + <literal>1</literal>. If the array currently has no integer
+     indices, the key will be <literal>0</literal> (zero). For the historical behaviour prior to
+     <literal>PHP 8.3.0</literal> — where appending after a negative key could yield <literal>0</literal> — see Example #7.
```
kamil-tekiela: Why is PHP a literal value?

kamil-tekiela:
Was this written with AI?

language/types/array.xml
kamil-tekiela:
```suggestion
     If no key is specified, the key will be the
     maximum existing integer index + <literal>1</literal>. If the array has no positive
     integer indices, the key will be <literal>0</literal>.
     As of PHP 8.3.0, it can also be a negative integer.
```

jordikroon: We shouldn't mention PHP at all. It falls under the same category as using "you". Instead I prefer to use the following type of phrasing. As Kamil also suggested.

> As mentioned above, if no key is specified, the next integer key will be used: the maximum existing integer index + <literal>1</literal>.

kamil-tekiela: Yes, correct, but also the old "as mentioned above" is super ambiguous.

Narkunan:
> Was this written with AI?

No. I used existing piece as reference they used literal

Narkunan:
@kamil-tekiela and @jordikroon Ready for re-review

kamil-tekiela (approved):
It's ok for me, but I think you have one too many spaces in indentation.

jordikroon:
Thank you @Narkunan

Ps. Besides removing the extra space. I changed `para` to `simpara` as well.

PR #5580, opened 2026-05-24
Clarify negative -1 return value for sscanf/fscanf

jordikroon:
Also did some maintenance work for these files.
See commits in order. Specifically this commit https://github.com/php/doc-en/commit/7af7c98e4f7c983581dc05c58b958bb6c0c1267b is what it's all about.

Ref: https://externals.io/message/131005

### DIFF
--- a/reference/filesystem/functions/fscanf.xml
+++ b/reference/filesystem/functions/fscanf.xml
@@ -14,67 +14,64 @@
    <methodparam><type>string</type><parameter>format</parameter></methodparam>
    <methodparam rep="repeat"><type>mixed</type><parameter role="reference">vars</parameter></methodparam>
   </methodsynopsis>
-  <para>
+  <simpara>
    The function <function>fscanf</function> is similar to
    <function>sscanf</function>, but it takes its input from a file
    associated with <parameter>stream</parameter> and interprets the
    input according to the specified <parameter>format</parameter>. 
-  </para>
-  <para>
+  </simpara>
+  <simpara>
    Any whitespace in the format string matches any whitespace in the input
    stream. This means that even a tab (<literal>\t</literal>) in the format
    string can match a single space character in the input stream.
-  </para>
-  <para>
+  </simpara>
+  <simpara>
    Each call to <function>fscanf</function> reads one line from the file.
-  </para>
+  </simpara>
  </refsect1>
 
  <refsect1 role="parameters">
   &reftitle.parameters;
-  <para>
-   <variablelist>
-    <varlistentry>
-     <term><parameter>stream</parameter></term>
-     <listitem>
-      &fs.file.pointer;
-     </listitem>
-    </varlistentry>
-    &strings.scanf.parameter.format;
-    <varlistentry>
-     <term><parameter>vars</parameter></term>
-     <listitem>
-      <para>
-       The optional assigned values.
-      </para>
-     </listitem>
-    </varlistentry>
-   </variablelist>
-  </para>
+  <variablelist>
+   <varlistentry>
+    <term><parameter>stream</parameter></term>
+    <listitem>
+     &fs.file.pointer;
+    </listitem>
+   </varlistentry>
+   &strings.scanf.parameter.format;
+   <varlistentry>
+    <term><parameter>vars</parameter></term>
+    <listitem>
+     <para>
+      The optional assigned values.
+     </para>
+    </listitem>
+   </varlistentry>
+  </variablelist>
  </refsect1>
 
  <refsect1 role="returnvalues">
   &reftitle.returnvalues;
-  <para>
+  <simpara>
    If only two parameters were passed to this function, the values parsed will be
-   returned as an array. Otherwise, if optional parameters are passed, the
-   function will return the number of assigned values. The optional
+   returned as an <type>array</type>. Otherwise, if optional parameters are passed,
+   the function will return the number of assigned values. The optional
    parameters must be passed by reference. 
-  </para>
-  <para>
+  </simpara>
+  <simpara>
    If there are more substrings expected in the <parameter>format</parameter>
    than there are available within <parameter>string</parameter>,
    &null; will be returned. On other errors, &false; will be returned.
-  </para>
+  </simpara>
  </refsect1>
 
  <refsect1 role="examples">
   &reftitle.examples;
-  <para>
-   <example>
-    <title><function>fscanf</function> Example</title>
-    <programlisting role="php">
-<![CDATA[
+  <example>
+   <title><function>fscanf</function> Example</title>
+   <programlisting role="php">
+    <![CDATA[
 <?php
 $handle = fopen("users.txt", "r");
 while ($userinfo = fscanf($handle, "%s\t%s\t%s\n")) {
@@ -84,36 +81,31 @@ while ($userinfo = fscanf($handle, "%s\t%s\t%s\n")) {
 fclose($handle);
 ?>
 ]]>
-    </programlisting>
-   </example>
-  </para>
-  <para>
-   <example>
-    <title>Contents of users.txt</title>
-    <programlisting role="txt">
-<![CDATA[
+   </programlisting>
+  </example>
+  <example>
+   <title>Contents of users.txt</title>
+   <programlisting role="txt">
+    <![CDATA[
 javier  argonaut        pe
 hiroshi sculptor        jp
 robert  slacker us
 luigi   florist it
 ]]>
-    </programlisting>
-   </example>
-  </para>
+   </programlisting>
+  </example>
  </refsect1>
 
  <refsect1 role="seealso">
   &reftitle.seealso;
-  <para>
-   <simplelist>
-    <member><function>fread</function></member>
-    <member><function>fgets</function></member>
-    <member><function>fgetss</function></member>
-    <member><function>sscanf</function></member>
-    <member><function>printf</function></member>
-    <member><function>sprintf</function></member>
-   </simplelist>
-  </para>
+  <simplelist>
+   <member><function>fread</function></member>
+   <member><function>fgets</function></member>
+   <member><function>fgetss</function></member>
+   <member><function>sscanf</function></member>
+   <member><function>printf</function></member>
+   <member><function>sprintf</function></member>
+  </simplelist>
  </refsect1>
 
 </refentry>
--- a/reference/filesystem/functions/fscanf.xml
+++ b/reference/filesystem/functions/fscanf.xml
@@ -64,6 +64,11 @@
    than there are available within <parameter>string</parameter>,
    &null; will be returned. On other errors, &false; will be returned.
   </simpara>
+  <simpara>
+   When optional parameters are used and the end of the input
+   <parameter>string</parameter> is reached before any value has been
+   parsed, <literal>-1</literal> is returned.
+  </simpara>
  </refsect1>
 
  <refsect1 role="examples">
--- a/reference/filesystem/functions/fscanf.xml
+++ b/reference/filesystem/functions/fscanf.xml
@@ -76,7 +76,7 @@
   <example>
    <title><function>fscanf</function> Example</title>
    <programlisting role="php">
-    <![CDATA[
+<![CDATA[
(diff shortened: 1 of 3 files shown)


reference/filesystem/functions/fscanf.xml
```diff
@@ -14,67 +14,69 @@
...
   &reftitle.examples;
-  <para>
-   <example>
-    <title><function>fscanf</function> Example</title>
-    <programlisting role="php">
-<![CDATA[
+  <example>
+   <title><function>fscanf</function> Example</title>
+   <programlisting role="php">
+    <![CDATA[
```
kamil-tekiela:
```suggestion
<![CDATA[
```

hakre:
@jordikroon thanks for the initiative, I spotted the mixed forms of "were" and "are" (time /E: [Active Voice](https://github.com/php/doc-en/pull/5580)) across the paragraphs. I have to admit that I don't know which rule we apply/use specifically, probably present is better because it is more active, less passive language (for which IIRC [and as I remembered incorrectly] we have [not] a rule in the docs folder), but as English is not my first language I might be a bit off here, so please take it with a grain of salt. This btw. is across the two functions' files, and was already different previously.

/E: And my concrete line comments didn't make it to the server and then back.

jordikroon:
@hakre you bring up a good and important point. Present tense feels more natural in the documentation. However, this has not always been consistent across the existing files, so I think it would be good to standardize on one style going forward. Thank you for bringing this up. Perhaps it would be good to mark this more explicitly in the style guide. Since it contributes to a professional documentation.

hakre:
@jordikroon Thank for your thoughful feedback and considering my thoughts. I've taken it as an opportunity to scan the doc-base/docs folder once more (phpdoc doc-base docs in one of my forks), specifically the _Style guidelines_  in the file entry named `style.md`.

A [_General Grammar_ HTML section exists](https://doc.php.net/guide/style.md), but it does not contain a hint of [Active Voice](https://github.com/php/doc-en/pull/5580) as I think this is named more appropriately (not past / present tense as the word _time_ may have suggested in my earlier comment).

Prolonging too much on this could derail yours beautiful PR a bit, so it should go without writing, that I would not suggest to water it down with this discussion too much, just from what I know it is normally suggested to use _active_, not passive, voice for documentation and the already existing sentence bullet-pointed in the _General Grammar_ section *may* hint towards that:

> If a statement includes a conditional conjunction, the condition being met should come before the independent clause.

Conditions are normally formulated in active voice I'd say (probably what _"conditional conjunction"_ pins down in that sentence), but this sentence is only what caught my eye while scanning.

And I spared the _Personalization_ section that is more leading but may only be contextually related. 

Henceforth, I'm thinking that your suggestion to align ("standardize") it within your PR's changed XML fragments in the *scanf() function family, e.g. to "are" not "were", could be a good example - if you feel that's right for your changes. Doing so may also help us to find a useful guideline for our future edits.

reference/spl/splfileobject/fscanf.xml
```diff
@@ -13,81 +13,85 @@
...
+   If only two parameters were passed to this function, the values parsed will be
+   returned as an <type>array</type>. Otherwise, if optional parameters are passed,
+   the function will return the number of assigned values. The optional
    parameters must be passed by reference.
-  </para>
+  </simpara>
+  <simpara>
+   If there are more substrings expected in the <parameter>format</parameter>
+   than there are available within <parameter>string</parameter>,
+   &null; will be returned. On other errors, &false; will be returned.
```
lacatoire:
```suggestion
   &null; will be returned.
```

hakre:
@jordikroon okay, I stand corrected after consulting more resources, incl. grammatical onces: my earlier _time_ was more correct (tense is different), but this is not active or passive voice as I had it in mind (which could also be a rule for a style-guide, but it's not what caught my eyes). Both the "were" and "are" that I spotted are passive constructions. Even if a style-guide would have a rule for active voice to prefer (which it doesn't), this is not a binary preference. E.g. if the focus is on the return value, passive voice can put that on focus, not the actor (e.g. the _caller_ of a function - as we can't use "you" per style-guide :D).

So this is not easy to find some rule-of thumb that helps editing/reviewing I guess _and_ "are" and "were" are not in a real "conflict" within the same simpara as I now understand it.

reference/strings/functions/sscanf.xml
```diff
@@ -83,18 +85,16 @@ list($month, $day, $year) = sscanf($mandate, "%s %d %d");
...
-  <para>
-   <example>
-    <title><function>sscanf</function> - using optional parameters</title>
-    <programlisting role="php">
-<![CDATA[
+  </simpara>
+  <example>
+   <title><function>sscanf</function> - using optional parameters</title>
+   <programlisting role="php">
+    <![CDATA[
```
lacatoire:
```suggestion
<![CDATA[
```

reference/strings/functions/sscanf.xml
```diff
@@ -14,66 +14,68 @@
...
   &reftitle.examples;
-  <para>
-   <example>
-    <title><function>sscanf</function> Example</title>
-    <programlisting role="php">
-<![CDATA[
+  <example>
+   <title><function>sscanf</function> Example</title>
+   <programlisting role="php">
+    <![CDATA[
```
lacatoire:
```suggestion
<![CDATA[
```

reference/spl/splfileobject/fscanf.xml
```diff
@@ -13,81 +13,85 @@
...
+  </simpara>
+  <simpara>
+   When optional parameters are used and the end of the input
+   <parameter>string</parameter> is reached before any value has been
+   parsed, <literal>-1</literal> is returned.
```
lacatoire:
```suggestion
   When optional parameters are used and the end of the line read from the
   file is reached before any value has been parsed, <literal>-1</literal>
   is returned.
```

reference/filesystem/functions/fscanf.xml
```diff
@@ -14,66 +14,68 @@
...
+  <simpara>
    If there are more substrings expected in the <parameter>format</parameter>
    than there are available within <parameter>string</parameter>,
    &null; will be returned. On other errors, &false; will be returned.
-  </para>
+  </simpara>
+  <simpara>
+   When optional parameters are used and the end of the input
+   <parameter>string</parameter> is reached before any value has been
+   parsed, <literal>-1</literal> is returned.
```
lacatoire:
```suggestion
   When optional parameters are used and the end of the input read from
   <parameter>stream</parameter> is reached before any value has been
   parsed, <literal>-1</literal> is returned.
```

reference/filesystem/functions/fscanf.xml
```diff
@@ -84,36 +86,31 @@ while ($userinfo = fscanf($handle, "%s\t%s\t%s\n")) {
...
-   <example>
-    <title>Contents of users.txt</title>
-    <programlisting role="txt">
-<![CDATA[
+   </programlisting>
+  </example>
+  <example>
+   <title>Contents of users.txt</title>
+   <programlisting role="txt">
+    <![CDATA[
```
lacatoire:
```suggestion
<![CDATA[
```

reference/spl/splfileobject/fscanf.xml
```diff
@@ -13,81 +13,85 @@
...
+  <simpara>
+   If only two parameters were passed to this function, the values parsed will be
+   returned as an <type>array</type>. Otherwise, if optional parameters are passed,
+   the function will return the number of assigned values. The optional
    parameters must be passed by reference.
-  </para>
+  </simpara>
+  <simpara>
+   If there are more substrings expected in the <parameter>format</parameter>
+   than there are available within <parameter>string</parameter>,
```
lacatoire:
```suggestion
   than there are available in the line read from the file,
```

reference/spl/splfileobject/fscanf.xml
```diff
@@ -13,81 +13,85 @@
...
   &reftitle.returnvalues;
-  <para>
-   If only one parameter is passed to this method, the values parsed will be
-   returned as an array. Otherwise, if optional parameters are passed, the
-   function will return the number of assigned values. The optional
+  <simpara>
+   If only two parameters were passed to this function, the values parsed will be
+   returned as an <type>array</type>. Otherwise, if optional parameters are passed,
+   the function will return the number of assigned values. The optional
    parameters must be passed by reference.
```
lacatoire:
```suggestion
   If only one parameter is passed to this method, the values parsed will be
   returned as an <type>array</type>. Otherwise, if optional parameters are passed,
   the method will return the number of assigned values. The optional
   parameters must be passed by reference.
```

reference/spl/splfileobject/fscanf.xml
hakre: ~~on other errors fscanf() still returns false, e.g. scan on `/dev/null` makes fscanf() return bool(false).~~

~~/E: https://3v4l.org/ZuL0k~~

/E: SplFileObject::fscanf() here, not fscanf()

reference/spl/splfileobject/fscanf.xml
hakre: ~~... unless the line is empty (string zero value), then bool(false) is returned (with optional parameters).~~

/E: SplFileObject::fscanf() here, not fscanf()

lacatoire: I double-checked, for `SplFileObject::fscanf()`, an empty line read with optional parameters returns -1, not false (same on 8.3, 8.4 and 8.5):
```
file_put_contents('/tmp/empty.txt', "\n");
$f = new SplFileObject('/tmp/empty.txt');
var_dump($f->fscanf('%d', $a)); // int(-1)
```
false only appears for the function `fscanf()` when no line can be read at all (EOF, for example `/dev/null`), and even then the method stays within its array|int|null type (null or -1): See [here](https://github.com/php/php-src/blob/14624997bbb5dffa6b47dc5e9b2e16b22e1f5b54/ext/spl/spl_directory.stub.php#L281)

reference/spl/splfileobject/fscanf.xml
lacatoire: snippet does print `bool(false)`. This page documents the method SplFileObject::fscanf() though, whose declared return type is narrower and has no false:
- SplFileObject::fscanf(string $format, mixed &...$vars): array|int|null -> [here](https://github.com/php/php-src/blob/14624997bbb5dffa6b47dc5e9b2e16b22e1f5b54/ext/spl/spl_directory.stub.php#L281)
- fscanf($stream, string $format, mixed &...$vars): array|int|false|null -> [here](https://github.com/php/php-src/blob/14624997bbb5dffa6b47dc5e9b2e16b22e1f5b54/ext/standard/basic_functions.stub.php#L2801)

On the same /dev/null input the method returns null (or -1 with optional parameters), never false, while the function returns false (consistent on 8.3, 8.4 and 8.5):
```
// method (this page)
var_dump((new SplFileObject('/dev/null'))->fscanf('%d'));  // NULL
$f = new SplFileObject('/dev/null');
var_dump($f->fscanf('%d', $a));                            // int(-1)
```

```
// function
var_dump(fscanf(fopen('/dev/null', 'r'), '%d'));          // bool(false)
```
So removing false here keeps the page in line with the method's declared return type. The false wording stays fully correct on the function page (reference/filesystem/functions/fscanf.xml), which this PR also updates.

hakre: Right, this is SplFileObject, I stand corrected. My mistake.

reference/spl/splfileobject/fscanf.xml
hakre: @lacatoire It seems I've mistaken those two (fscanf() function with the SplFileObject method) with each other, my bad.

jordikroon:
Thank you all for checking. I must say this escalated quickly (in a good sense). While I was very focused on fixing the initial issue. I didn't expect those 3 functions to be so similar yet historically so wrongfully implemented.

That said it's good to see that corrected as well.

I will leave it open for a few days, so if anyone has any additional notes. Please drop those here.

PR #5585, opened 2026-05-26
unserialization: Documented allowed-classes option with enums

edorian:
Following the discussion in https://github.com/php/php-src/issues/22059 

As enums are technically [classes](https://www.php.net/manual/en/language.types.enumerations.php), and the allowed-classes option set to "false" does not affect them this piece of clarification should help document the expected (and current) behavior.

### DIFF
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -816,10 +816,11 @@ print serialize(Suit::Hearts);
    On deserialization, if an enum and case cannot be found to match a serialized
    value a warning will be issued and &false; returned.</para>
 
-  <para>
+  <simpara>
    The <literal>allowed_classes</literal> option of
-   <function>unserialize</function> does not affect enums.
-  </para>
+   <function>unserialize</function> does not affect 
+   <link linkend="language.enumerations">Enumerations</link>.
+  </simpara>
 
   <para>
    If a Pure Enum is serialized to JSON, an error will be thrown. If a Backed Enum
--- a/language/enumerations.xml
+++ b/language/enumerations.xml
@@ -817,6 +817,11 @@ print serialize(Suit::Hearts);
    value a warning will be issued and &false; returned.</para>
 
   <para>
+   The <literal>allowed_classes</literal> option of
+   <function>unserialize</function> does not affect enums.
+  </para>
+
+  <para>
    If a Pure Enum is serialized to JSON, an error will be thrown. If a Backed Enum
    is serialized to JSON, it will be represented by its scalar value only, in the
    appropriate type. The behavior of both may be overridden by implementing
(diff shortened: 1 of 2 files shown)


language/enumerations.xml
```diff
@@ -816,6 +816,11 @@ print serialize(Suit::Hearts);
    On deserialization, if an enum and case cannot be found to match a serialized
    value a warning will be issued and &false; returned.</para>
 
+  <para>
+   The <literal>allowed_classes</literal> option of
+   <function>unserialize</function> does not affect enums.
```
TimWolla:
```suggestion
   <function>unserialize</function> does not affect <link linkend="language.enumerations">Enumerations</link>.
```
And ditto in the other file.

language/enumerations.xml
```diff
@@ -816,6 +816,11 @@ print serialize(Suit::Hearts);
    On deserialization, if an enum and case cannot be found to match a serialized
    value a warning will be issued and &false; returned.</para>
 
+  <para>
+   The <literal>allowed_classes</literal> option of
```
TimWolla, line 820:
```suggestion
   The <literal>"allowed_classes"</literal> option of
```
Not sure about this one, but `<literal>` feels like it should be combined with `""`. Perhaps check if there's precedent.

TimWolla (changes requested):

language/enumerations.xml
jordikroon: There isn't really a policy for this. It is done, but sometimes.. and mostly for cases like "0" not really for options or arguments. So I am more leaning towards omitting in this specific case.

language/enumerations.xml
```diff
@@ -816,6 +816,11 @@ print serialize(Suit::Hearts);
    On deserialization, if an enum and case cannot be found to match a serialized
    value a warning will be issued and &false; returned.</para>
 
+  <para>
```
jordikroon:
```suggestion
  <simpara>
```

language/enumerations.xml
```diff
@@ -816,6 +816,11 @@ print serialize(Suit::Hearts);
...
+   The <literal>allowed_classes</literal> option of
+   <function>unserialize</function> does not affect enums.
+  </para>
```
jordikroon:
```suggestion
  </simpara>
```

jordikroon (changes requested):
Just a minor nit; Simpara should be used here too.

reference/var/functions/unserialize.xml
```diff
@@ -103,6 +103,9 @@
             Omitting this option is the same as defining it as &true;: PHP
             will attempt to instantiate objects of any class.
            </simpara>
+           <simpara>
+            This option does not affect enums.
```
lacatoire:
```suggestion
            This option does not affect <link linkend="language.enumerations">enums</link>.
```

edorian:
All review feedback should be incorporated now. Sorry for the delay.

lacatoire (approved):

jordikroon:
Thank you!

PR #5587, opened 2026-05-29
Update ibase_trans method parameters in XML

MartinKoeditz:
Transaction args have to be in first place. Removed wrong description. Removed optional parameter 'trans_args' from ibase_trans method documentation.


MartinKoeditz:
See also: https://github.com/php/doc-de/pull/201

jordikroon (approved):
LGTM. Given you are an active contributor of php-firebird, I trust its accuracy.

PR #5588, opened 2026-06-01
Fix non-runnable code examples (undefined variables, syntax errors)

lacatoire:
This fixes a batch of code examples that cannot run as written:

- undefined variables (a missing `$` sigil, or a variable name that does not
  match the one declared just above);
- a few syntax errors: a missing `;`, a stray `.` that turns a subtraction
  into string concatenation, a `=` instead of `=>`, an unterminated string;
- a Cyrillic homoglyph (`С`, U+0421) used in a type hint instead of the Latin
  `C` class;
- a couple of examples instantiating or calling the wrong class/method name.

Each change only touches the example so it parses and runs as intended.

### DIFF
--- a/appendices/migration74/new-features.xml
+++ b/appendices/migration74/new-features.xml
@@ -94,8 +94,8 @@ class A
 class B extends A
 {
     // Fatal error: Could not check compatibility between B::method():C and
-    // A::method(): A, because class С is not available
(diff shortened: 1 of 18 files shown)


jordikroon (approved):
Thanks!

PR #5591, opened 2026-06-05
reflection: ReflectionFunction::__construct() example expects bool param

rbairwell:
The example on https://www.php.net/manual/en/reflectionfunction.construct.php uses `var_export` with the second argument as `1` but `var_export` accepts a boolean as its second argument. Using the code as provided causes
`TypeError: var_export(): Argument #2 ($return) must be of type bool, int given`
if strict_types is enabled.

Changed the second argument to the two `var_export` calls to "true".


jordikroon:
Thank you!

PR #5593, opened 2026-06-06
PDO: update constants type from int to bool for PHP 8.4.0

jordikroon:


reference/pdo/constants.xml
```diff
@@ -484,11 +488,12 @@
    <varlistentry xml:id="pdo.constants.attr-emulate-prepares">
     <term>
      <constant>PDO::ATTR_EMULATE_PREPARES</constant>
-     (<type>int</type>)
+     (<type>bool</type>)
     </term>
     <listitem>
      <simpara>
```
lacatoire, line 501:
```suggestion
      <simpara>
        Whether to enable or disable emulation of prepared statements. Some drivers
        either do not support native prepared statements or have only limited
        support for them. When set to &true;, PDO always emulates prepared
        statements; when set to &false;, it uses the driver's native prepared
        statements.
     </simpara>
     <simpara>
```

reference/pdo/constants.xml
```diff
@@ -484,11 +488,19 @@
    <varlistentry xml:id="pdo.constants.attr-emulate-prepares">
     <term>
      <constant>PDO::ATTR_EMULATE_PREPARES</constant>
-     (<type>int</type>)
+     (<type>bool</type>)
     </term>
     <listitem>
+      <simpara>
```
jordikroon:
```suggestion
     <simpara>
```

vrana:
This is wrong, the constants are still int.

```php
<?php
echo PHP_VERSION . "\n";
var_dump(PDO::ATTR_AUTOCOMMIT);
?>
8.4.25
int(0)
```

https://github.com/php/php-src/blob/php-8.5.10/ext/pdo_dblib/pdo_dblib.c#L200

What you probably meant is that they return bool from `getAttribute`?

PR #5595, opened 2026-06-06
Add deprecation notice for mysqli_execute that was added in PHP 8.5

jordikroon:
According to Git blame the old notice existed since 2004. But it is formally deprecated in PHP 8.5.


kamil-tekiela (approved):
LGTM

PR #5599, opened 2026-06-08
Update doc-en/language/functions.xml

Abrifq:
Updated the part explaining when a function becomes available.

It used to check for the boolean, which was always true since the code defines it in the first line, but the boolean itself does not guarantee the function was available.

The example code now checks if the function exists again, so it should make more sense.

### DIFF
--- a/language/functions.xml
+++ b/language/functions.xml
@@ -89,9 +89,9 @@ if ($makefoo) {
 }
 
 /* Now we can safely call foo()
-   since $makefoo evaluated to true */
+   since function_exists('foo') evaluated to true */
 
-if ($makefoo) foo();
+if (function_exists('foo')) foo();
 
 function bar() 
 {


jordikroon:
Hi @Abrifq

The given example tries to explain conditional functions. So to specify, functions that are within a conditional statement. 
Changing `$makefoo` to `function_exists('foo')` will break the purpose of this example all together. 

The example above is almost identical to this change. Therefor I don't think anything is wrong with the current example.

```php
/* We can't call foo() from here 
   since it doesn't exist yet,
   but we can call bar() */
if (function_exists('foo')) {
    foo();
} else {
    echo "Function foo does not exist (yet).\n";
}
```

Abrifq:
> Hi @Abrifq
> 
> The given example tries to explain conditional functions. So to specify, functions that are within a conditional statement. 

Technically a more correct term could be "conditionally **defined** functions", as the condition has to be met _only_ on where the definition is being made.

> Changing `$makefoo` to `function_exists('foo')` will break the purpose of this example all together.

After defining `foo`, changing `$makefoo` to `false` shouldn't affect the already existing definition. Also there is no absolute need for the variable to be exposed or have the condition stored in a variable. It could even be `rand(0,1)` and still work with my current fix.

(Though, I would reject a PR in heartbeat that randomly defines a function without saving the result of the random condition. This is about syntatically valid behavior vs. morally valid behavior, like the condition being a constant)

```php
$makefoo=false;

if($makefoo){
    function foo(){ return "foo fighters!!"; }
}

$makefoo=true;

if(function_exists('foo') /* == false */) foo();

if($makefoo /* == true */) foo();
```

> The example above is almost identical to this change. Therefore I don't think anything is wrong with the current example.

It is identical in the sense that the "inner" functions get defined in another block scope, whether in an if condition or function body. However I'd like to believe the proper use case for conditionally defined functions to be something like compiler definitions and the use case for functions defined in another functions to be something like lazy evaluated modules.

In both cases though, the consumer code shouldn't check the environment variable or call the `init_module_xyz()` every time it needs to call the functions. Instead, they should check for the function's existence, and if it exists, then do call it.

---

TL;DR: the condition being met at call time itself doesn't guarantee the function is available, therefore the user should ALWAYS check before attempting to call.

Abrifq:
Hello again,

Is there anything blocking this PR?

lacatoire (approved):

kamil-tekiela:
I am closing this PR as it seems to me like you misunderstood the purpose of the example. There is nothing wrong with how the example is written right now. Your change would make the example invalid so we cannot accept it.

PR #5600, opened 2026-06-08
Fix example output (screen) inconsistent with code

lacatoire:
Several example output blocks did not match what the code produces (found during a doc-fr vs doc-en comparison). Each change was verified against the actual code, most of them run empirically.

Covers: oop5 inheritance, mktime, dom construct, Ds (map/pair/vector), memcached fetch, mongodb bson/cursor, spl rewind, uri getQuery, wincache delete, yaf config-ini, yar examples.

### DIFF
--- a/reference/yar/examples.xml
+++ b/reference/yar/examples.xml
@@ -78,7 +78,7 @@ var_dump($client->add(1, 2));
 var_dump($client->call("add", array(3, 2)));
 
 
-/* _add can not be called */
+/* _add cannot be called */
 var_dump($client->_add(1, 2));
 ?>
 ]]>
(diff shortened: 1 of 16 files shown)


reference/yar/examples.xml
```diff
@@ -78,7 +78,7 @@ var_dump($client->add(1, 2));
 var_dump($client->call("add", array(3, 2)));
 
 
-/* __add can not be called */
+/* _add can not be called */
```
jordikroon:
```suggestion
/* _add cannot be called */
```

jordikroon (changes requested):

PR #5602, opened 2026-06-09
Fix: bpftrace note about environment variable

egmc:
Thank you for merging https://github.com/php/doc-en/pull/5500 !

I have noticed that there was a minor error in some of the notes regarding tracing setup, so please check this as well 🙇‍♂️


jordikroon:
You are welcome. And thank you for this change as well.

PR #5608, opened 2026-06-12
returntypewillchange.xml Add warning about future strict return type …

mmalferov:
…enforcement

Should we not explicitly mention that internal methods will strictly enforce return types in future PHP versions?

Currently, the documentation might give developers a false sense of security, making it look like `#[ReturnTypeWillChange]` is a permanent fix. In reality, it only works during the tentative phase. Once PHP 9.0 drops, any remaining signature mismatches in child classes will cause a fatal error, as far as I understand


language/predefined/attributes/returntypewillchange.xml
```diff
@@ -16,6 +16,17 @@
...
     the deprecation notice.
    </para>
+
+   <warning>
+    <simpara>
+     The <classname>ReturnTypeWillChange</classname> attribute suppresses
+     deprecation warnings during the tentative return type phase <emphasis>only</emphasis>.
+     Any remaining signatures mismatches will trigger an <constant>E_COMPILE_ERROR</constant> once internal methods
+     transition to standard strict return types. This breaking change is scheduled for PHP 9.0.0,
+     at which point a fatal error is raised and this attribute will no longer have any effect.
```
jordikroon: We generally avoid mentioning planned future release behaviour in the docs, since release plans can change and the information tends to become outdated quickly. 

I think it would be enough to state it only suppresses deprecation notices during the tentative return type phase.

mmalferov: I completely agree that mentioning specific versions and error constants was overkill. I've removed them to keep the focus purely on the core behavior.

The term *tentative return type* was used in the warning without being defined anywhere else in the text, which could be confusing. So I briefly defined what the tentative phase actually means.

Thank you for the guidance! Ready for another look

language/predefined/attributes/returntypewillchange.xml
```diff
@@ -10,12 +10,25 @@
    <para>
```
jordikroon:
```suggestion
   <simpara>
```

jordikroon (changes requested):
Since we touch `<para>` it should become `<simpara>`. Could you do a rebase with master as well, so that checks will run correctly.

language/predefined/attributes/returntypewillchange.xml
```diff
@@ -10,12 +10,25 @@
...
     a compatible return type, otherwise a deprecated notice is emitted during
-    inheritance validation.
+    inheritance validation. This introduces a tentative return type phase:
+    the engine emits a deprecation notice instead of a fatal error when return
+    types are incompatible, before they become enforced in a future version.
     In case the return type cannot be declared for an overriding method due to
     PHP cross-version compatibility concerns,
     a <code>#[\ReturnTypeWillChange]</code> attribute can be added to silence
     the deprecation notice.
    </para>
```
jordikroon:
```suggestion
   </simpara>
```

PR #5506, opened 2026-04-22
Document pcntl_getqos_class and pcntl_setqos_class

kamil-tekiela:
This is the bare minimum as I have no idea what these functions do or how they work.

PR #5609, opened 2026-06-13
Fix duplicate documentation for pcntl_getqos_class and pcntl_setqos_class

KentarouTakeda:
`pcntl_getqos_class` and `pcntl_setqos_class` were each documented in two files sharing the same `xml:id`. Because the build normalizes `_` to `-` in file-entity names, the hyphen-named files (#5350) collided with the underscore-named ones (#5506) and were silently dropped, so they were never rendered.

Fix this by merging both versions into the hyphen-named files (matching the `xml:id` and the directory's naming convention) and removing the underscore-named ones.


reference/pcntl/functions/pcntl-getqos-class.xml
```diff
@@ -26,15 +28,24 @@
...
-   the current <acronym>QoS</acronym> class.
+   Returns the current <acronym>QoS</acronym> class as a
+   <enumname>Pcntl\QosClass</enumname>.
+  </simpara>
+ </refsect1>
+
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   Throws an <classname>Error</classname> if the underlying call to
```
jordikroon:
```suggestion
   Throws an <exceptionname>Error</exceptionname> if the underlying call to
```

reference/pcntl/functions/pcntl-setqos-class.xml
```diff
@@ -47,10 +98,19 @@
   </simpara>
  </refsect1>
 
+ <refsect1 role="errors">
+  &reftitle.errors;
+  <simpara>
+   Throws an <classname>Error</classname> if the underlying call to
```
jordikroon:
```suggestion
   Throws an <exceptionname>Error</exceptionname> if the underlying call to
```

reference/pcntl/functions/pcntl-setqos-class.xml
```diff
@@ -24,20 +23,72 @@
...
-      <member><literal>Pcntl\QosClass::Default</literal></member>
-      <member><literal>Pcntl\QosClass::Utility</literal></member>
-      <member><literal>Pcntl\QosClass::Background</literal></member>
-     </simplelist>
+     <para>
+      The Quality of Service class to assign to the current thread. The
+      operating system uses this as a hint to schedule CPU time, I/O priority,
+      and energy usage, with higher classes preempting lower ones. See
+      <enumname>Pcntl\QosClass</enumname> for the available cases.
+     </para>
```
jordikroon:
```suggestion
     <simpara>
      The Quality of Service class to assign to the current thread. The
      operating system uses this as a hint to schedule CPU time, I/O priority,
      and energy usage, with higher classes preempting lower ones. See
      <enumname>Pcntl\QosClass</enumname> for the available cases.
     </simpara>
```

reference/pcntl/functions/pcntl-setqos-class.xml
```diff
@@ -24,20 +23,72 @@
...
+     <para>
```
jordikroon: `<variablelist>` doesn't need to be wrapped in a `<para>`.

jordikroon (changes requested):
Thank you for improving these pages. Just a few styling comments.

KentarouTakeda:
Thanks for the review! Applied all four suggestions.

PR #5613, opened 2026-06-14
returntypewillchange.xml Add warning about future strict return type enforcement

mmalferov:
Should we not explicitly mention that internal methods will strictly enforce return types in future PHP versions?

Currently, the documentation might give developers a false sense of security, making it look like #[ReturnTypeWillChange] is a permanent fix. In reality, it only works during the tentative phase. Once PHP 9.0 drops, any remaining signature mismatches in child classes will cause a fatal error, as far as I understand.

P. S. The checks did not pass in the previous PR. And something went wrong with rebase. So I created a new Pull Request, and closed the old one:

https://github.com/php/doc-en/pull/5608

But even now the check doesn 't work 🤷

### DIFF
--- a/language/predefined/attributes/returntypewillchange.xml
+++ b/language/predefined/attributes/returntypewillchange.xml
@@ -7,15 +7,28 @@
 
   <section xml:id="returntypewillchange.intro">
    &reftitle.intro;
-   <para>
+   <simpara>
     Most non-final internal methods now require overriding methods to declare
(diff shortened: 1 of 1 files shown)


jordikroon:
The failed was caused by docbook-cs (that was introduced yesterday), that had a small issue specifically related to GH actions. It has been fixed in https://github.com/php/docbook-cs/pull/15

PR #5614, opened 2026-06-15
Update `ReturnTypeWillChange` intro for clarity

mmalferov:
I know I'm revisiting [this topic](https://github.com/php/doc-en/pull/5613) so soon, but the clarity of the current phrasing leaves a lot to be desired. What if we structure and present the page the way this PR suggests?

### DIFF
--- a/language/predefined/attributes/returntypewillchange.xml
+++ b/language/predefined/attributes/returntypewillchange.xml
@@ -8,27 +8,44 @@
   <section xml:id="returntypewillchange.intro">
    &reftitle.intro;
    <simpara>
+    As of PHP 8.1.0, a tentative phase of transitioning to return type declarations
+    for internal class methods began.
+   </simpara>
+
+   <simpara>
     Most non-final internal methods now require overriding methods to declare
-    a compatible return type, otherwise a deprecated notice is emitted during
-    inheritance validation. This introduces a tentative return type phase:
-    the engine emits a deprecation notice instead of a fatal error when return
-    types are incompatible, before they become enforced in a future version.
+    a compatible return type. Otherwise, a deprecation notice is emitted during
+    inheritance validation, warning that the signature violates covariance rules.
+   </simpara>
+
+   <simpara>
+    Return type compatibility checks for internal class extensions remain in a soft mode
+    until a future PHP version enforces strict checking.
+    At that point, mismatches will cause a fatal error instead of a deprecation notice.
+   </simpara>
+
+   <simpara>
     In case the return type cannot be declared for an overriding method due to
-    PHP cross-version compatibility concerns,
-    a <code>#[\ReturnTypeWillChange]</code> attribute can be added to silence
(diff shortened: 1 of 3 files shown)


lacatoire (approved):

language/predefined/attributes/returntypewillchange.xml
```diff
@@ -7,28 +7,45 @@
...
+   </simpara>
+
    <simpara>
     Most non-final internal methods now require overriding methods to declare
-    a compatible return type, otherwise a deprecated notice is emitted during
-    inheritance validation. This introduces a tentative return type phase:
-    the engine emits a deprecation notice instead of a fatal error when return
-    types are incompatible, before they become enforced in a future version.
+    a compatible return type. Otherwise, a deprecation notice is emitted during
+    inheritance validation, warning that the signature violates covariance rules.
```
kamil-tekiela:
```suggestion
    inheritance validation warning that the signature violates covariance rules.
```

language/predefined/attributes/returntypewillchange.xml
```diff
@@ -7,28 +7,45 @@
...
+     strictly. A fatal error is emitted on mismatch, even if the overriding method
+     is marked with the attribute.
     </simpara>
-   </warning>
 
+    <simpara>
+     The <classname>ReturnTypeWillChange</classname> attribute suppresses deprecation
+     warnings <emphasis>only</emphasis> during the tentative return type phase.
+     After strict type checking is enforced, the attribute will stop working,
+     and a mismatch in an overriding method signature of an internal class
```
kamil-tekiela:
```suggestion
     After strict type checking is enforced, the attribute will stop working
     and a mismatch in the overriding method signature of an internal class
```

language/predefined/attributes/returntypewillchange.xml
```diff
@@ -7,28 +7,45 @@
...
-    the engine emits a deprecation notice instead of a fatal error when return
-    types are incompatible, before they become enforced in a future version.
+    a compatible return type. Otherwise, a deprecation notice is emitted during
+    inheritance validation, warning that the signature violates covariance rules.
+   </simpara>
+
+   <simpara>
+    Return type compatibility checks for internal class extensions remain in a soft mode
+    until a future PHP version enforces strict checking.
+    At that point, mismatches will cause a fatal error instead of a deprecation notice.
```
kamil-tekiela: It's not very clear what the soft mode is. I assume it means that a deprecation is thrown instead of the fatal error, but that's what the paragraph above says. So is this paragraph entirely redundant?

mmalferov: You're right, the paragraph looks overloaded, but at the same time, we must warn the reader that in a future version of PHP, the signature compatibility check will become strict. I have fixed the "lame" paragraph. Please check if everything is in order now

kamil-tekiela (approved):

PR #5615, opened 2026-06-15
Follow up with Dev Container improvements

jordikroon:
- Clone web-php (required by `build.sh php`).
 - Drop `_JAVA_OPTIONS`, fixed in doc-base.
 - Override `HTTP_HOST` since Dev Container uses a proxy.
 - Run `php` built-in server in the background.


jordikroon:
~When https://github.com/php/web-php/pull/1927 is merged, we can ditch `server-prepend.php`~

Actually we can't since HTTP_HOST is localhost. So we need to use HTTP_X_FORWARDED_X headers.

PR #5620, opened 2026-06-16
Fix radius_get_vendor_attr return type hint

tfaller:
As the return description says, `radius_get_vendor_attr` returns `false` on error. This PR fixes the return type hint by changing it to an union of an array and false.


lacatoire (approved):

reference/radius/functions/radius-get-vendor-attr.xml
jordikroon:
```suggestion
<refentry xml:id="function.radius-get-vendor-attr" xmlns="http://docbook.org/ns/docbook">
```

jordikroon (commented):
Not directly related. But sniffer failed because the direct parent is `<refentry>`.

jordikroon (approved):

jordikroon:
LGTM. Thank you for your contribution!

PR #5625, opened 2026-06-18
Remove 72 unused entities from language-snippets.ent

lacatoire:
Auditing `language-snippets.ent` surfaced unused entity declarations that are not referenced anywhere, directly or transitively: deprecation banners for old PHP cycles, snippets for removed extensions (mcrypt, sqlite, t1lib, safemode, ...) and stale helpers.

This removes the 72 that no translation references either. The 11 entities still referenced by some translations are split into a separate PR.

Pure cleanup, no rendered output changes.

### DIFF
--- a/language-snippets.ent
+++ b/language-snippets.ent
@@ -3,20 +3,8 @@
 
 <!ENTITY installation.enabled.disable 'This extension is enabled by default. It may be disabled by using the following option at compile time: '>
 
-<!-- Not used in EN anymore -->
 <!ENTITY changelog.randomseed '<row xmlns="http://docbook.org/ns/docbook"><entry>4.2.0</entry><entry>The random
 number generator is seeded automatically.</entry></row>'>
-
-<!ENTITY warn.deprecated.feature-5-3-0.removed-6-0-0 '<warning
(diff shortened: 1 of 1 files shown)


alfsb:
Also checked non usage on translations?

But even if some outdated translations still use some of then, keep the deletion. Anything that shrinks `language-snippets.ent` is a big **yes** for me.

And after https://github.com/php/doc-base/pull/301 is merged, any deleted entity on doc-en still used for translations can now be moved to a temporary file, so full remotions on `doc-en` can be done in one step, without breaking translations.

lacatoire:
@alfsb i've removed those used in translations in another PR https://github.com/php/doc-en/pull/5628

alfsb:
That is impressive. You have a comprehensive view of dependencies between these DTD entities and languages. Hope that doc-base's 301 will help make this work easier. In fact, https://github.com/php/doc-en/pull/5628 may be the first use/user of an `doc-en/entities/entities-remove.ent` file.

sy-records (approved):

kamil-tekiela (approved):

PR #5628, opened 2026-06-19
Remove 11 unused entities still referenced by translations

lacatoire:
Follow-up to the unused-entities cleanup, kept separate because these 11 entities are still referenced by some translations (it/ro/uk/tr/ru/es, depending on the entity).

They are unused in doc-en (directly or transitively), so removing them changes no rendered EN output. Instead of a plain deletion, they are now recorded as tombstones in `entities/entities-remove.ent` (`translate="remove"`), so revcheck gives the translations a heads up to drop them on their side.

Stacked on #5634, which introduces the `entities/` mechanism. Also waiting for https://github.com/php/doc-base/pull/301 to be merged.


sy-records (approved):

kamil-tekiela:
Is this PR still needed?

alfsb:
I think it is still good, at least the content changes of language-snippets.ent and entities.remove.ent.

I have not tested it, but in principle it shrinks language-snippets.ent, without breaking translations, so manual splitting would be easier in the end.

alfsb:
Tested today, all languages build. Plan to merge this week.

kamil-tekiela (approved):

PR #5626, opened 2026-06-18
pdo_pgsql: document 8.5 deprecations, add examples, fix inconsistencies

philip:
Deprecate legacy methods and constants (8.5), add tested examples, and fill in missing details.

PR #5629, opened 2026-06-21
pdo_pgsql: document ATTR_PREFETCH lazy fetch, fix copyFromArray sig

philip:
Document the PHP 8.5 lazy (single-row) fetch mode: add a `PDO::ATTR_PREFETCH` entry to the PDO_PGSQL constants describing the =0 toggle, its connection vs per-statement scope, and the single-active-statement caveat. Note the driver-specific behavior on the generic PDO::ATTR_PREFETCH entry.

Update `Pdo\Pgsql::copyFromArray`'s rows parameter to `array|Traversable` in the synopsis, matching the changelog entry already added in #5626 and closes tasks noted in the https://github.com/orgs/php/projects/13/views

Behavior verified against PostgreSQL on PHP 8.5.7.

### DIFF
--- a/reference/pdo/constants.xml
+++ b/reference/pdo/constants.xml
@@ -251,6 +251,13 @@
       setting of the prefetch size.  A larger prefetch size results in
       increased performance at the cost of higher memory usage.
      </simpara>
+     <simpara>
+      The PDO_PGSQL driver instead treats this attribute as a toggle: as of
+      PHP 8.5.0, a value of <literal>0</literal> enables lazy (single-row)
+      fetching. See
+      <link linkend="pdo-pgsql.constants.attr-prefetch"><constant>PDO::ATTR_PREFETCH</constant></link>
+      in the PDO_PGSQL driver.
+     </simpara>
     </listitem>
    </varlistentry>
    <varlistentry xml:id="pdo.constants.attr-timeout">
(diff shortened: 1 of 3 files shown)


reference/pdo/constants.xml
```diff
@@ -251,6 +251,13 @@
       setting of the prefetch size.  A larger prefetch size results in
       increased performance at the cost of higher memory usage.
      </simpara>
+     <simpara>
+      The PDO_PGSQL driver instead treats this attribute as a toggle: as of
+      PHP 8.5.0, a value of <literal>0</literal> enables lazy (single-row)
+      fetching. See
+      <link linkend="pdo-pgsql.constants.attr-prefetch"><constant>PDO::ATTR_PREFETCH</constant></link>
```
jordikroon: I believe this doesn't need a `<link>`. In fact I feel this will render a duplicate anchor tag.

jordikroon (changes requested):

reference/pdo_pgsql/pdo/pgsql/copyfromarray.xml
lacatoire:
```suggestion
An indexed <type>array</type> (or <type>Traversable</type>) of
<type>string</type>s with fields separated by
<parameter>separator</parameter>.
```

philip: Thanks!

reference/pdo/constants.xml
philip: fixed, thanks

PR #5641, opened 2026-06-30
Use &true;/&false;/&null; entities for boolean/null constants

lacatoire:
Replaces `<literal>true/false/null</literal>` with the standard `&true;`/`&false;`/`&null;` entities where they denote the PHP boolean/null constant (26 occurrences, 21 files).

Quoted strings, JSON/INI literals, SQL `NULL` and type notations were reviewed and intentionally left untouched. Inline-only `<para>` touched by the change are converted to `<simpara>` (DocBook style check).

### DIFF
--- a/reference/win32service/rightinfo/get-full-username.xml
+++ b/reference/win32service/rightinfo/get-full-username.xml
@@ -26,7 +26,7 @@
  <refsect1 role="returnvalues"><!-- {{{ -->
   &reftitle.returnvalues;
   <para>
-   Return the domain and username or username if domain is <literal>null</literal> or &null; if no username found.
+   Return the domain and username or username if domain is &null; or &null; if no username found.
   </para>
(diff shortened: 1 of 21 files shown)


reference/win32service/rightinfo/get-full-username.xml
```diff
@@ -25,9 +25,9 @@
 
  <refsect1 role="returnvalues"><!-- {{{ -->
   &reftitle.returnvalues;
-  <para>
-   Return the domain and username or username if domain is <literal>null</literal> or &null; if no username found.
-  </para>
+  <simpara>
+   Return the domain and username or username if domain is &null; or &null; if no username found.
```
jordikroon: I think we should rewrite this. I agree this should be `&null;`, but now it's more confusing to read then it already was.

PR #5647, opened 2026-07-02
Add a warning about the usage of `LIBXML_PARSEHUGE`

alexandre-daubois:


ndossche (approved):

jordikroon:
Thanks!

alfsb:
I do not think this configuration even works, as the build of PHP manual [uses it](https://github.com/php/doc-base/blob/cd53050a60c65e3cb72180c9a368d815498c91ea/configure.php#L658), and yet this script easy hits [two](https://github.com/php/doc-base/blob/cd53050a60c65e3cb72180c9a368d815498c91ea/scripts/file-entities.php#L39) different hard limits of libxml.

ndossche:
The option works: libxml raises the limits compared to the default when not using PARSEHUGE, but it doesn't set the limit to infinite.

alfsb:
It's good that the option works. But it is sad that the manual compilation is arbitrarily near of the breaking point. Alas, I was toying with a userland XML parser that does entity expansion. I will try to complete it, as a spare.

alfsb:
In fact, a suggestion. From:

> Sets XML_PARSE_HUGE flag, which relaxes any hardcoded limit from the parser.

to

> Sets XML_PARSE_HUGE flag, which increases some hardcoded limits from the parser.

English is not my main language, but "relaxes" plus "any" has some connotations to "remove" (for me), whereas "increases" is more certain that the limits still exist.

PR #5650, opened 2026-07-03
info: fix assert/dl/getrusage return-value docs, E_ALL value, and versions.xml duplicates

lacatoire:
- **assert**: drop `assert.exception=1`, `assert.bail=1` and "custom exception" from the *always returns `true`* list, since those cases throw or bail instead of returning `true`.
- **dl**: with `enable_dl` off (or DL unsupported) an `E_WARNING` is emitted and `false` is returned, not an `E_ERROR` + stop.
- **getrusage**: `mode` `1` (`RUSAGE_CHILDREN`) collects child-process usage, not threads.
- **get_defined_constants**: `E_ALL` is `30719` (was `2047`, an outdated value).
- **versions.xml**: remove duplicate `get_required_files` and `get_defined_functions` (belongs to funchand, no page in `reference/info/`).

Touched files were also brought up to the DocBook style enforced by CI (inline-only `<para>` to `<simpara>`, exception classes to `<exceptionname>`).

### DIFF
--- a/reference/info/functions/assert.xml
+++ b/reference/info/functions/assert.xml
@@ -13,32 +13,32 @@
    <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
    <methodparam choice="opt"><type class="union"><type>Throwable</type><type>string</type><type>null</type></type><parameter>description</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
-  <para>
+  <simpara>
    <function>assert</function>
    allows for the definition of expectations: assertions that take effect
    in development and testing environments, but are optimised away to have
    zero cost in production.
-  </para>
-  <para>
+  </simpara>
+  <simpara>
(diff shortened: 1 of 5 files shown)


reference/info/functions/assert.xml
```diff
@@ -13,32 +13,32 @@
    <methodparam><type>mixed</type><parameter>assertion</parameter></methodparam>
    <methodparam choice="opt"><type class="union"><type>Throwable</type><type>string</type><type>null</type></type><parameter>description</parameter><initializer>&null;</initializer></methodparam>
   </methodsynopsis>
-  <para>
+  <simpara>
```
jordikroon, line 16: Please don't change para to simpara if the inners don't change. It makes reviewing changes more difficult and it doesn't add much. Planning to do bulk changes when doc-extensions is ready.

jordikroon (changes requested):

PR #5657, opened 2026-07-06
Mark do-not-translate files

alfsb:
Apply the new `<?do-not-translate?>` in `doc-en` files that are auto-generated, expected not translated, ot that not contains any translatable text.

In special, I would like to confirm/know:

* `docbookcs.xml` should not be copied on translatoins?
* en/appendices/license.xml is expected never translated.
* where is the code that generates en/reference/datetime/timezones.xml, this the generator outputs the `<?do-not-translate?>` mark.

Things I considered:

1. To remove the vi/vim comments and eof very small files, and from all version.xml files;

2. To remove the top comment, and leave only at start:
```xml
<?do-not-translate?>

```

There still about 200 `/*/versions.xml` files, that I will change, directly, if and after this PR is merged. Plan to merge it in one week, if there is no hard oppositions.

### DIFF
--- a/appendices/license.xml
+++ b/appendices/license.xml
@@ -1,7 +1,6 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
+<?xml version="1.0" encoding="utf-8"?><?do-not-translate?>
 
-<!-- DO NOT TRANSLATE THIS FILE -->
+<!-- DO NOT TRANSLATE THIS FILE. -->
 
 <appendix xml:id="cc.license" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
  <title>Creative Commons Attribution 3.0</title>
@@ -390,4 +389,3 @@
   </listitem>
  </orderedlist>
 </appendix>
-
--- a/appendices/license.xml
+++ b/appendices/license.xml
@@ -1,6 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?><?do-not-translate?>
-
-<!-- DO NOT TRANSLATE THIS FILE. -->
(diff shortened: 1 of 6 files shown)


alfsb:
Also, are there some other .xml files, besides versions.xml files, that I missed? I got this "list" from file exclusions of revcheck, that I expect is correct, but to be sure.

appendices/license.xml
```diff
@@ -1,7 +1,6 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
+<?xml version="1.0" encoding="utf-8"?><?do-not-translate?>
```
jordikroon: `extensions.xml` is on it's own line. 
I think we should do that for this and the following markers as well.

jordikroon:
docbookcs.xml may be copied. It's up to the translator if they want this in as it could be useful for them.
And I added this marker to https://github.com/php/doc-en/pull/5658 as well.

appendices/license.xml
alfsb: The style on `extensions.xml` was by necessity, but I also think it is visually pleasant.

alfsb:
Plan to merge this by friday, having a small cooldown from 5634.

alfsb:
> docbookcs.xml may be copied. It's up to the translator if they want this in as it could be useful for them. And I added this marker to #5658 as well.

Then it needs a `<!-- $Revision$ -->` line after XML decl, so translations have a place where to replace for the revtag, without disturbing line counts or creating long lines.

Also, can you add some other XML comment, poining for documentation or a fast/small tutorial, of where the file is used and how it could be changed in translations?

alfsb:
Plan to merge this by next monday, and then after that, to mark all `version.xml` files. Last is to remove the ignore-by-name conventions in revcheck,

PR #5658, opened 2026-07-06
Add redirect.xml map for manual pages on php.net

jordikroon:
The goal for this PR is to migrate all doc related redirects out of web-php.
Instead we generate a `redirects.json` from `redirects.xml` which is then loaded in-place.

Redirects not related to docs are NOT migrated. However shortcuts are.

Depends on:

https://github.com/php/phd/pull/260

### DIFF
--- a/redirects.xml
+++ b/redirects.xml
@@ -1,0 +1,454 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Redirect map for manual pages on php.net.
+
+  Rendered to redirects.php by phd (Redirects package) and included by
+  web-php/error.php. Only redirects that concern manual pages live here;
+  site-level aliases (downloads, support, git, etc.) and external URLs
+  stay in web-php/error.php.
+
+  Scopes:
+    shortcut   match a bare URI  (php.net/{from})
+                  -> /manual/{lang}/{to}.php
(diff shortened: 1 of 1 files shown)


redirects.xml
```diff
@@ -0,0 +1,776 @@
+<?xml version="1.0" encoding="utf-8"?>
+<?do-not-translate?>
+<!--
+  Redirect map for manual pages on php.net.
+
+  Rendered to redirects.php by phd and included by web-php/error.php.
+  Only redirects that concern manual pages live here;
+
+  Every redirect here is served with HTTP 301.
```
derickr, line 9: Right now, not all of these do a 301 redirect, but instead serve the content directly with the "legacy" URL. I think for many, that is the right goal. (As example `https://www.php.net/class.oci-lob`)

For others, such as `https://www.php.net/@`, there is a 301.

alfsb:
The question in another PR, about the `phd/Redirects` class only doing validations, is in the context that I would not oppose having a .json file in `doc-en`, already in the final format that `web-php/error.php` uses. Something like:
```json
[{"comment": "See web-php/error.php for details."}

,{"comment": "   OCI-Collection class renamed   "}
,{"redirect":"manual", "from":"class.oci-lob", "to":"class.ocilob"}
,{"redirect":"manual", "from":"oci-lob.append", "to":"ocilob.append"}

]
```

PR #5661, opened 2026-07-08
Remove return.falseproblem banner for pcntl_setcpuaffinity

SteenSchutt:
The function can only return a boolean value, so the banner does not apply.


lacatoire (approved):

jordikroon:
LGTM. Thank you!

PR #5664, opened 2026-07-10
Fix dead external links in the English manual

lacatoire:
Fixes dead external links, verified against the resolved targets:

* `reference/ldap/constants.xml`: `&url.rfc;45282` had a stray digit (link text reads "RFC 4528").
* `reference/mongodb/architecture.xml`: removed the double slash after `&url.packagist.package;` (already ends in `/`).
* `install/windows/commandline.xml`: Microsoft KB 321788 and the retired MSDN forum post (both 404, no archive) are unlinked, keeping the reference text.


install/windows/commandline.xml
```diff
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
```
jordikroon, line 2: Why the revision drop?

PR #5675, opened 2026-07-11
Document ReflectionConstant::getAttributes() (PHP 8.5)

lacatoire:
PHP 8.5 adds `ReflectionConstant::getAttributes()` (RFC: attributes on constants), which had no documentation. Adds the method page.

Verified against php-src: `ext/reflection/php_reflection.stub.php` declares `getAttributes(?string $name = null, int $flags = 0): array`, matching the documented signature.

### DIFF
--- a/reference/reflection/reflectionconstant/getattributes.xml
+++ b/reference/reflection/reflectionconstant/getattributes.xml
@@ -57,6 +57,46 @@
   </para>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title>Basic usage</title>
+    <programlisting role="php">
+<![CDATA[
+<?php
+#[Attribute]
+class Fruit {
+}
+
+#[Attribute]
+class Red {
+}
+
+#[Fruit]
+#[Red]
+const APPLE = 'apple';
+
+$constant = new ReflectionConstant('APPLE');
+$attributes = $constant->getAttributes();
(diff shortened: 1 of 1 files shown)


reference/reflection/reflectionconstant/getattributes.xml
```diff
@@ -0,0 +1,92 @@
...
+     </tbody>
+    </tgroup>
+   </informaltable>
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
+   <simplelist>
```
jordikroon: `<simplelist>` doesn't need to be wrapped in a `<para>`.

reference/reflection/reflectionconstant/getattributes.xml
```diff
@@ -0,0 +1,92 @@
...
+    <member><methodname>ReflectionClass::getAttributes</methodname></member>
+    <member><methodname>ReflectionClassConstant::getAttributes</methodname></member>
+    <member><methodname>ReflectionFunctionAbstract::getAttributes</methodname></member>
+    <member><methodname>ReflectionProperty::getAttributes</methodname></member>
+   </simplelist>
+  </para>
+ </refsect1>
+
```
jordikroon, line 109: Could we add an example? The other `getAttributes` have one.

jordikroon (changes requested):

jordikroon (approved):

PR #5676, opened 2026-07-11
Document Dom\Element::$outerHTML property (PHP 8.5)

lacatoire:
PHP 8.5 adds the `Dom\Element::$outerHTML` virtual property. Documents it in the class synopsis and property list.

Verified against php-src: `ext/dom/php_dom.stub.php` declares `public string $outerHTML;` (`@virtual`, no `readonly`), and `ext/dom/php_dom.c` registers both `dom_element_outer_html_read` and `dom_element_outer_html_write` handlers, so the property is writable (mirrors `innerHTML`).


ndossche (approved):
Thank you

PR #5677, opened 2026-07-12
Document the new Dom\ParentNode $children property (PHP 8.5)

lacatoire:
Documents the new read-only `$children` property (`Dom\HTMLCollection`) added to `Dom\ParentNode` and its implementations `Dom\Element`, `Dom\Document` and `Dom\DocumentFragment` in PHP 8.5.

The property is defined once on `Dom\ParentNode` and referenced through a local anchor in each implementing class, matching how the other ParentNode properties are documented.


ndossche (approved):
thanks

PR #5678, opened 2026-07-14
Document more curl deprecations

Sjord:
I compiled against cURL 8.18 with enabled deprecation warnings, and documented the results.

### DIFF
--- a/reference/curl/constants_curl_getinfo.xml
+++ b/reference/curl/constants_curl_getinfo.xml
@@ -740,7 +740,7 @@
   </term>
   <listitem>
    <simpara>
-    Average upload speed. Deprecated since 7.55.0. Use <constant>CURLINFO_SPEED_UPLOAD_T</constant> instead.
+    Average upload speed. Deprecated since cURL 7.55.0. Use <constant>CURLINFO_SPEED_UPLOAD_T</constant> instead.
    </simpara>
   </listitem>
  </varlistentry>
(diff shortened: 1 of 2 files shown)


reference/curl/constants_curl_getinfo.xml
```diff
@@ -740,7 +740,7 @@
   </term>
   <listitem>
    <simpara>
-    Average upload speed
+    Average upload speed. Deprecated since 7.55.0. Use <constant>CURLINFO_SPEED_UPLOAD_T</constant> instead.
```
lacatoire:
```suggestion
    Average upload speed. Deprecated since cURL 7.55.0. Use <constant>CURLINFO_SPEED_UPLOAD_T</constant> instead.
```

reference/curl/constants_curl_setopt.xml
```diff
@@ -1528,7 +1528,7 @@
      <literal>private</literal> is used.
      Setting this option to &null; will disable kerberos support for FTP.
      Defaults to &null;.
-     Available as of cURL 7.16.4.
+     Available as of cURL 7.16.4 and deprecated as of cURL 8.17.0.
```
lacatoire, line 1531:
```suggestion
     Available as of cURL 7.16.4 and removed as of cURL 8.17.0.
```

Sjord: I don't think this is correct. The *functionality* was removed, but the constant is still available but deprecated.

Sjord: https://curl.se/libcurl/c/CURLOPT_KRBLEVEL.html

lacatoire: my bad, had missed that

lacatoire (approved):

PR #5679, opened 2026-07-14
[skip-revcheck] Add missing <!-- $Revision$ --> in relative-class-types.xml

DavidePastore:
Add missing `<!-- $Revision$ -->` in `relative-class-types.xml`.

### DIFF
--- a/language/types/relative-class-types.xml
+++ b/language/types/relative-class-types.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
 <sect1 xml:id="language.types.relative-class-types">
  <title>Relative class types</title>
 


alfsb:
Prefix the PR title with [skip-revcheck], to avoid changing the hash on all translations. See Translating documentation in https://doc.php.net/guide/editing.md .

DavidePastore:
@alfsb, done 😄

alfsb (approved):
I'm somewhat very fearful of merging things in doc-en, but I will try to do the merging of small PRs like this, in the future, if you or someone else doesn't do it first.

jordikroon:
@alfsb Davide doesn't have write access to doc-en. But feel free to merge. There isn't really much that can go wrong.
But if you prefer, I can do. Given how much you are involved in doc-base and doc-en no need to be hesitant.

alfsb:
> But if you prefer, I can do.

Yes, please. It was only by chance that I glanced these PRs.

> Given how much you are involved in doc-base and doc-en no need to be hesitant.

Thanks, and yet, I prefer to restrict myself to doc-base most of the time. While the work there is foundational, like doc-en, it's only code, not text. English is not my first language.

PR #5680, opened 2026-07-14
[skip-revcheck] Add missing <!-- $Revision$ --> to singleton.xml

DavidePastore:
Add missing `<!-- $Revision$ -->` to `singleton.xml`.

### DIFF
--- a/language/types/singleton.xml
+++ b/language/types/singleton.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- $Revision$ -->
 <sect1 xml:id="language.types.singleton">
  <title>Singleton types</title>
 


alfsb:
Prefix the PR title with `[skip-revcheck]`, to avoid changing the hash on all translations. See Translating documentation in https://doc.php.net/guide/editing.md .

DavidePastore:
@alfsb, done. In the documentation that you linked, I read the following:

> Changes in the English version are eventually picked up by the translators. If a change doesn't affect translations (e.g. fixing a typo in English) then the commit message should start with [skip-revcheck].

Maybe I should also change the commit message. Do you confirm that?

alfsb:
> Maybe I should also change the commit message. Do you confirm that?

Yes. In fact, I just merged a PR, and noted the commit message ignored any changes in PR name. So yes, the skip-revcheck need to be added in the commit message. Sorry about the wrong instruction.

alfsb (approved):

jordikroon:
@alfsb usually before merging we change the commit message. But thanks for informing to add `[skip-revcheck]` there.

PR #5681, opened 2026-07-14
Clarify `forward_static_call()` documentation

zonuexe:
The `forward_static_call()` and `forward_static_call_array()` descriptions implied that these functions simply call static methods.

This clarifies that they call a callback while preserving the current called class for method callbacks, making them useful for static method callbacks that should behave like forwarding calls such as `parent::` or `self::`.

The `forward_static_call_array()` parameter note is also updated to match `call_user_func_array()`: parameters can be passed by reference when the corresponding element in the `args` array is a reference.

I am not a native English speaker, so I do not have strong preferences about the exact wording or nuance.

### DIFF
--- a/reference/funchand/functions/forward-static-call-array.xml
+++ b/reference/funchand/functions/forward-static-call-array.xml
@@ -3,7 +3,8 @@
 <refentry xml:id="function.forward-static-call-array" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>forward_static_call_array</refname>
-  <refpurpose>Call a static method and pass the arguments as array</refpurpose>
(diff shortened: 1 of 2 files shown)


lacatoire (approved):

lacatoire:
Thanks for the PR @zonuexe ! A commit landed on master (846dfb87) shortly after opening that covered the same ground. I'm closing to avoid conflict with it.

zonuexe:
Rebased on latest master and resolved the conflict with 846dfb87. I kept that commit's correction that forward_static_call_array() does not require an active class scope, and adjusted this PR to focus on clarifying late static binding/current called class behavior and the args-by-reference note consistent with call_user_func_array().

PR #5682, opened 2026-07-14
Third-party manual, doc-en side

alfsb:
This PR is the first step in preparation for https://wiki.php.net/rfc/third_party_ext_documentation . The main objective is to "move" files that are specific to the main manual into `doc-en`, so `doc-base` can build each manual with the minimum of conventions.

Also, it intends to stop cross-repositories PRs between manuals and `doc-base`, even more so in the many piecewise changes that the splitting of manuals may cause.

Many changes in manual.xml already included here:
- Root `<set>` is reorganized;
- Some xmlns re-declarations removed;
- funcindex.xml is embedded, so `&global.function-index;` is not necessary anymore;
- Auto generated files are to be loaded from a temp/ dir, sibling of doc-en (to keep git clean and manual build idempotent).

global.ent is moved as is, to be manually  converted into XML entities later.

If everything goes well, the preamble of manual.xml eventually will contain only two parameter entities, one for text-entities, other for file-entities.

**Questions:**

1. This is ok?
There are *many* changes to be made in doc-base/configure.php (and probably in PhD) to start building the new external extensions/projects manual. Moving these files here, in particular manual.xml, has the potential of making many of these changes unnecessary.

2. The root `<set>` is not missing a xmlns for XInclude?
Even if the final XML should *not* include any XInclude in the end, there many intermediary steps there xi:include tags are valid and expected content of manual building.

3. Is the version="5.2" in root `<set>` necessary?

4. funcindex.xml not moved
This file is small, has no translatable text, so I decided to merge it in manual.xml. But it is referred to by name in PhD and `file-entities.php`, so it's necessary to open an issue on PhD, noting its future remotion, I presume. This file used/mentioned elsewhere?

5. acronyms.xml, idem
I do not plan to move this file, as it appears not to be used *directly* anywhere. I presume it is used in PhD, to patch all acronym elements. I would like to hear some PhD maintainers or translations, in about "converting" this file into translatable XML entities, keyed by acronym.lower. That will need changes on PhD, of course, but will make acronyms translable. The very first acronym, AKA, "Also Known As", is a good example of the necessity of this part of the manual to be translatable.

---

As third-party efforts already started, I would like to merge this as soon as possible, but I know this a big change, and a controversial one to boot, so I plan to leave this open at least two weeks. Comments and reviews are *very* welcome.


alfsb:
There is an alternative plan. To first move only `manual.xml` with the minimal changes necessary, so output changes would be minimal or nonexistent at first, and then to start converting or erasing per-manual exclusive files from doc-en. I may take this route, while working on the doc-base side of this PR.

alfsb:
Took the path for minimal changes. I planned to merge both sides in about two weeks. Splitted manuals can be fully functional by early August.

alfsb:
The force pushes are to re-run Checks against master HEAD instead master past, pure rebases, or for another reason?

jordikroon:
The rebase + force push were made to apply the latest docbook-cs configuration that came with a few fixes that affected this PR. One of those was the absence of `LANG`. And to re-run the checks after those changes were made.

The violations that remain are actual violations that would be good to address. No code has been changed here during the force push.

alfsb:
> The rebase + force push were made to apply the latest docbook-cs configuration that came with a few fixes that affected this PR.

So you are directly observing the GH checks being stuck in the past, and/or ignoring posterior changes on master?

> One of those was the absence of `LANG`. And to re-run the checks after those changes were made.

This got me curious. docbookcs loads all DTD and XML .ent files, to merge them and use them in each file validation? Something like this?

> The violations that remain are actual violations that would be good to address.

Planned after manual split settles down.

> No code has been changed here during the force push.

I recently came across the `git commit --allow-empty -m "Trigger build"` trick.

jordikroon:
> So you are directly observing the GH checks being stuck in the past, and/or ignoring posterior changes on master?

No I just stumbled upon this PR and saw that there was a bug in docbook-cs that caused GH checks to fail. Since I was working on it anyway, I decided to patch the bug and re-run the checks. To give more accurate feedback, and to have yet another bug fixed in a recently introduced tool. And docbook-cs loads with the setup that exists in this branch. Now that I think of it, perhaps we could always load the master XML file, as that's always the most accurate version. It would have saved the rebase. Will put it on my list.

> One of those was the absence of `LANG`. And to re-run the checks after those changes were made. 
> This got me curious. docbookcs loads all DTD and XML .ent files, to merge them and use them in each file validation? Something like this?

Docbook-cs first loads a list of all entities (.ent or .dtd) files with regexes, and keeps them in-memory through a simple PHP array. Once that's done it will iterate through all the files that are within it's scope) and it will replace the references that exist in a file with the once we found in memory. Overall this results in massive performance benefits. A full checker run is just a few seconds long. Faster if the scope is more small, e.g a diff check.

> I recently came across the `git commit --allow-empty -m "Trigger build"` trick.

You can also re-run CS from here, but since we had to change a few files in master (docbookcs.xml) I had to do a rebase + force push in order to re-run CS with the latest master.

alfsb:
> > So you are directly observing the GH checks being stuck in the past, and/or ignoring posterior changes on master?
> 
> No I just stumbled upon this PR and saw that there was a bug in docbook-cs that caused GH checks to fail. Since I was working on it anyway, I decided to patch the bug and re-run the checks. To give more accurate feedback, and to have yet another bug fixed in a recently introduced tool. And docbook-cs loads with the setup that exists in this branch.

So, the rebase was necessary to code to observe posterior changes in the repository. Simple re-runs were not sufficient, right?

> Now that I think of it, perhaps we could always load the master XML file, as that's always the most accurate version. It would have saved the rebase. Will put it on my list.

By master XML file do you mean the file some file stored into the default branch of some repository, instead the same file in the feature branch? Or another thing?

Related, I think I have a fix for this, to re-runs to always observe posterior changes in PRs, without rebases. But in the end, it may never land in doc-en or doc-base, as no one else appears to observe this (d)effect.

> Docbook-cs first loads a list of all entities (.ent or .dtd) files with regexes, and keeps them in-memory through a simple PHP array. Once that's done it will iterate through all the files that are within it's scope) and it will replace the references that exist in a file with the once we found in memory. Overall this results in massive performance benefits. A full checker run is just a few seconds long. Faster if the scope is more small, e.g a diff check.

I think I can change the XML trim PR, as to simplify this part of loading all entities, down to a single file. I will only work on it after manual split settles down, but may avoid chasing all .ent/.dtd files around many repos.

> You can also re-run CS from here, but since we had to change a few files in master (docbookcs.xml) I had to do a rebase + force push in order to re-run CS with the latest master.

So simple re-runs were not sufficient. Re-runs here were stuck into the past, ignoring posterior changes in the master... Sorry to sound repetitive, but I'm trying to really be sure we are talking about the same problem, or if it is something that I am not seeing.

alfsb:
Merged. It will not cause any changes until the doc-base side is merged, tomorrow. I may further try to reduce the `DOCTYPE` node to a single instruction. The dream is to make cross-repo PRs disappear completely, at least between doc-base/configure and main manual.xml of each repo.

PR #5051, opened 2025-12-22
Add partitioned for setcookie and setrawcookie

jordikroon:
Add documentation for the new "partitioned" cookie option introduced in PHP 8.5.0 (CHIPS support). Updates include:

- Added "partitioned" to the list of supported cookie options in setcookie() and setrawcookie() functions
- Documented the new session.cookie_partitioned INI directive with description of CHIPS functionality
- Added changelog entries for PHP 8.5.0 in affected function documentation
- Updated session cookie parameter function references to include the new partitioned option

Ref: https://github.com/php/doc-en/issues/4886

PR #5689, opened 2026-07-16
ext/standard: Document partitioned cookie option (8.5.0)

lacatoire:
PHP 8.5.0 adds the `partitioned` key to the `options` array of `setcookie()` and `setrawcookie()` (CHIPS). The option requires `secure` to be set, or a `ValueError` is thrown.

Documents the new option and adds changelog entries to both function pages.


jordikroon:
This has been covered now in #5051

PR #5700, opened 2026-07-23
imagick: fix broken methodsynopsis xpointers in exception classes

lacatoire:
The five Imagick exception classes had broken `<xi:include>` xpointers masked
by empty `<xi:fallback/>`: an own-methods include pointing at a non-existent
`refentry`, and an inherited-methods include using the class name as
`methodsynopsis` role instead of `Exception`. Both selected an empty node set,
so the pages rendered with their inherited methods missing.

Fixed by mirroring the working `jsonexception.xml` pattern.

Related:
- php/doc-en#5699 removes the empty fallbacks that hid this
- php/doc-base#326 recovers this class of failure on translations

### DIFF
--- a/reference/imagick/imagickdrawexception.xml
+++ b/reference/imagick/imagickdrawexception.xml
@@ -38,20 +38,12 @@
      </oointerface>
     </classsynopsisinfo>
     <classsynopsisinfo role="comment">&InheritedProperties;</classsynopsisinfo>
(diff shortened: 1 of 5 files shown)


alfsb (approved):
Plan to merge this after the bigger one, that also removes xi:fallback.

kamil-tekiela (approved):
LGTM

PR #5705, opened 2026-07-24
Remove dead code in exec.xml example

kamil-tekiela:


alfsb:
That is to say, passing by reference *declares* undefined variables. I do not think this is mentioned anywhere in the manual.

kamil-tekiela:
It's explained here https://www.php.net/manual/en/language.references.whatdo.php

alfsb:
Oh. I missed that. That is the principal reason the doc-base/configure.php "pre declares" passed by reference output variables.

jordikroon (approved):
Thanks for this follow up.

PR #5707, opened 2026-07-24
Fix localeconv() function description

### LINKED ISSUE #5666: localconv: parameters documented as bool while example shows int
From manual page: https://php.net/function.localeconv

---

see

Image

which defines bool parameters, while the example below shows numeric values

Image

kamil-tekiela:
Fixes #5666

### DIFF
--- a/reference/strings/functions/localeconv.xml
+++ b/reference/strings/functions/localeconv.xml
@@ -89,29 +89,29 @@
       <row>
        <entry>p_cs_precedes</entry>
        <entry>
-        &true; if currency_symbol precedes a positive value, &false;
+        1 if currency_symbol precedes a positive value, 0
         if it succeeds one
        </entry>
       </row>
       <row>
        <entry>p_sep_by_space</entry>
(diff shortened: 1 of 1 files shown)


reference/strings/functions/localeconv.xml
```diff
@@ -89,29 +89,29 @@
...
-        value, &false; otherwise
+        1 if a space separates currency_symbol from a positive
+        value, 0 otherwise
        </entry>
       </row>
       <row>
        <entry>n_cs_precedes</entry>
        <entry>
-        &true; if currency_symbol precedes a negative value, &false;
+        1 if currency_symbol precedes a negative value, 0;
```
jordikroon: The `;` is a left over from `&false;` I think.

kamil-tekiela:
```suggestion
        1 if currency_symbol precedes a negative value, 0
```

jordikroon (approved):

PR #5718, opened 2026-07-26
Fix examples whose documented output doesn't match Run code

KentarouTakeda:
Every example that has a Run code button and declares its output with `&example.outputs;` (832 of them) was run and compared against its `<screen>`.

This PR is the subset where the fix needs no judgement: each hunk can be confirmed by clicking Run code on the page.

One commit per group:

- Broken examples: two have no `<?php` so Run code prints the source instead of running it, one calls getName() on the getShortName() page.
- DOM XML declaration: saveXML() emits it, the screen drops it or drops the encoding the document declares.
- Formatting `var_dump()`/`print_r()` never produce: indentation, `[0] =>` for `[0]=>`, hand alignment, spaces where the data holds tabs.
- Wrong values: stale line numbers and offsets, a wrong ordering, a dropped semicolon, output split across lines that is really printed on one.
- One example read `date('T')` with no timestamp, so the page was right only half the year; the timestamp is now pinned and the screen is unchanged.

<details>
<summary>Pages to reproduce this on</summary>

Click any page and press Run code: the output that appears is what this PR puts in the `<screen>`. The static `<screen>` still shown on these pages is the "before" side of the diff, until the manual is rebuilt.

**Fix two broken examples**

- https://www.php.net/manual/en/uri-rfc3986-uri.withfragment.php
- https://www.php.net/manual/en/reflectionconstant.getshortname.php

**Fix the XML declaration in DOM <screen> blocks**

- https://www.php.net/manual/en/domdocument.append.php
- https://www.php.net/manual/en/domelement.before.php
- https://www.php.net/manual/en/dom-characterdata.before.php
- https://www.php.net/manual/en/dom-characterdata.remove.php
- https://www.php.net/manual/en/dom-characterdata.replacewith.php

**Use the formatting var_dump() and print_r() actually produce**

- https://www.php.net/manual/en/language.oop5.basic.php
- https://www.php.net/manual/en/language.oop5.lazy-objects.php
- https://www.php.net/manual/en/domelement.getattributenames.php
- https://www.php.net/manual/en/reflectionclass.getdefaultproperties.php
- https://www.php.net/manual/en/simplexmlelement.getdocnamespaces.php
- https://www.php.net/manual/en/function.simplexml-load-string.php
- https://www.php.net/manual/en/function.strtok.php
- https://www.php.net/manual/en/function.array-unshift.php
- https://www.php.net/manual/en/datetime.formats.php
- https://www.php.net/manual/en/function.array-unique.php
- https://www.php.net/manual/en/function.explode.php
- https://www.php.net/manual/en/function.libxml-set-external-entity-loader.php
- https://www.php.net/manual/en/function.ltrim.php
- https://www.php.net/manual/en/function.rtrim.php
- https://www.php.net/manual/en/function.trim.php
- https://www.php.net/manual/en/function.array-fill.php

**Fix <screen> values that do not match the actual output**

- https://www.php.net/manual/en/function.str-word-count.php
- https://www.php.net/manual/en/function.htmlentities.php
- https://www.php.net/manual/en/function.natsort.php
- https://www.php.net/manual/en/function.array-udiff-assoc.php
- https://www.php.net/manual/en/function.array-map.php
- https://www.php.net/manual/en/domnode.getlineno.php
- https://www.php.net/manual/en/domdocumentfragment.replacechildren.php
- https://www.php.net/manual/en/dom-htmldocument.createfromstring.php
- https://www.php.net/manual/en/language.oop5.basic.php
- https://www.php.net/manual/en/language.oop5.overloading.php
- https://www.php.net/manual/en/language.types.array.php
- https://www.php.net/manual/en/language.types.callable.php
- https://www.php.net/manual/en/control-structures.foreach.php
- https://www.php.net/manual/en/control-structures.goto.php
- https://www.php.net/manual/en/control-structures.switch.php
- https://www.php.net/manual/en/control-structures.match.php
- https://www.php.net/manual/en/function.preg-replace.php
- https://www.php.net/manual/en/simplexmlelement.children.php
- https://www.php.net/manual/en/reflectionclass.getdoccomment.php
- https://www.php.net/manual/en/xmlwriter.writecdata.php
- https://www.php.net/manual/en/function.date-sun-info.php
- https://www.php.net/manual/en/function.libxml-get-errors.php

**Pin the timestamp in the date('T') example**

- https://www.php.net/manual/en/function.date-default-timezone-get.php

</details>


language/control-structures/goto.xml
```diff
@@ -95,6 +95,8 @@ echo "$i = $i";
    <screen>
 <![CDATA[
 Fatal error: 'goto' into loop or switch statement is disallowed in script on line 2
+Stack trace:
+#0 {main}
```
kamil-tekiela: I don't know if we should do that. Maybe just change the &example.outputsl to &example.outputs.similar;? Otherwise we would have to split this into two sections with one for PHP 8.5+.

language/types/array.xml
```diff
@@ -1124,10 +1124,10 @@ var_dump((array) new B());
...
 array(3) {
-  ["BA"]=>
-  NULL
   ["AA"]=>
   NULL
+  ["BA"]=>
+  NULL
   ["AA"]=>
   NULL
 }
```
kamil-tekiela, line 1146: Please split into two sections: before PHP 8.1 and as of PHP 8.1.

reference/datetime/functions/date-sun-info.xml
```diff
@@ -282,7 +282,7 @@ echo "Length of day: ",
     &example.outputs;
     <screen>
 <![CDATA[
-Length of day: 13h 56s
+Length of day: 13h 53s
```
kamil-tekiela, line 285: 56s was the incorrect value, as I understand. The 53s was the PHP 7 value and the current value. https://www.php.net/ChangeLog-8.php#8.3.22 In that case 53s is ok. Good catch.

This is not the first time this value was broken and adjusted.

reference/dom/dom/htmldocument/createfromstring.xml
```diff
@@ -85,7 +85,7 @@ echo $dom->saveHtml();
    <screen>
 <![CDATA[
 <!DOCTYPE html><html><head></head><body>
-    <p>Hello, world!</p>
+   <p>Hello, world!</p>
```
kamil-tekiela: This is a nit, but I would rather add a space in the code than remove in the output.

reference/libxml/functions/libxml-get-errors.xml
```diff
@@ -103,10 +103,10 @@ function display_xml_error($error, $xml)
...
   <titles>PHP: Behind the Parser</title>
-----------------------------------------------^
+-----------------------------------------^
 Fatal Error 76: Opening and ending tag mismatch: titles line 4 and title
   Line: 4
-  Column: 46
+  Column: 41
 
 --------------------------------------------
 ]]>
```
kamil-tekiela, line 112: This is truly odd. Both outputs are correct. I can't explain why that is. https://3v4l.org/GtBvD#veol
Maybe let's leave this one as is for now.

reference/pcre/functions/preg-replace.xml
```diff
@@ -282,8 +282,7 @@ echo $count; //3
     &example.outputs;
     <screen>
 <![CDATA[
-xp***to
-3
+xp***to3
```
kamil-tekiela: Add `echo "\n";` in the code.

kamil-tekiela (changes requested):

language/control-structures/goto.xml
KentarouTakeda: Dropped the stack trace and switched to `&example.outputs.similar;`. Confirmed 8.2–8.4 print no trace and 8.5 does: it comes from the new `fatal_error_backtraces` INI, which can also be switched off, so it doesn't belong in a version-agnostic screen. The manual has plenty of other `Fatal error:` screens in the same position, so that's probably worth deciding manual-wide rather than here.

language/types/array.xml
KentarouTakeda: Split as asked, using `&example.outputs.80;` / `&example.outputs.81;` to match the pattern in `appendices/migration81/incompatible.xml`. Checked the boundary: 7.4.33 and 8.0.30 give the old order, 8.1.0 onwards the new one.

reference/dom/dom/htmldocument/createfromstring.xml
KentarouTakeda: Done — indented the heredoc by one more space, so the screen goes back to what it was.

reference/libxml/functions/libxml-get-errors.xml
KentarouTakeda: It's a libxml difference, not a PHP one: libxml 2.9.10 reports column 46, while 2.9.13, 2.9.14 and 2.13.9 all report 41. Hence 3v4l showing 46 for every version from 5.1 to 8.5.9 — its builds share one old libxml.

The current 46 came from 3fbb1438bd (2009), "Update the output to use recent libxml2 version which includes better column support", replacing `Column: 0`. The value has always tracked current libxml, so 41 continues that — but happy to drop the hunk if you'd rather leave it.

reference/pcre/functions/preg-replace.xml
KentarouTakeda: Done, and the screen goes back to two lines.

kamil-tekiela (approved):
@jordikroon Is the CI failure something we should be concerned with?

KentarouTakeda:
Mine, and it's fixed — green after a rebase onto master.

Any branch based before #5682 fails the style check. That's most of the open PRs right now, so they'll go red on their next run until rebased.

PR #5719, opened 2026-07-27
Add note to shell_exec that the backtick operator is deprecated

cookieguru:
See https://github.com/php/doc-en/pull/5017

### DIFF
--- a/reference/exec/functions/shell-exec.xml
+++ b/reference/exec/functions/shell-exec.xml
@@ -14,7 +14,7 @@
    <methodparam><type>string</type><parameter>command</parameter></methodparam>
   </methodsynopsis>
   <para>
-   This function is identical to the <link linkend="language.operators.execution">backtick operator</link>.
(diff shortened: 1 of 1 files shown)


reference/exec/functions/shell-exec.xml
```diff
@@ -14,7 +14,7 @@
    <methodparam><type>string</type><parameter>command</parameter></methodparam>
   </methodsynopsis>
   <para>
```
jordikroon:
```suggestion
  <simpara>
```

reference/exec/functions/shell-exec.xml
```diff
@@ -14,7 +14,7 @@
...
-   This function is identical to the <link linkend="language.operators.execution">backtick operator</link>.
+   This function is identical to the deprecated <link linkend="language.operators.execution">backtick operator</link>.
   </para>
```
jordikroon:
```suggestion
  </simpara>
```

jordikroon:
Thank you!

PR #5720, opened 2026-07-28
Update screens for output that changed with PHP itself

KentarouTakeda:
Follow-up to #5718, from the same audit: the examples where PHP's own behaviour moved and the `<screen>` was never updated. One commit per kind of change.

- `+`/`-` vs `.` precedence ([8.0](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.other)), constants left of `instanceof` ([7.3](https://www.php.net/manual/en/migration73.new-features.php#migration73.new-features.core.instanceof-literals))
- reading a missing array key is a warning ([8.0](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.other)); the `BackedEnum::from()` ValueError no longer quotes the enum name (8.2)
- internal functions declare return types (8.0)
- anonymous class naming ([8.0](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.other)), `var_export()` leading backslash ([8.2](https://www.php.net/manual/en/migration82.incompatible.php#migration82.incompatible.standard))
- `mb_detect_encoding()` non-strict detection ([8.3](https://www.php.net/manual/en/migration83.other-changes.php#migration83.other-changes.functions.mbstring))
- `ReflectionClass::__toString()` regenerated rather than attributed to one change; the `Exception` dump has drifted across several releases

Where a corrected screen would leave the page contradicting itself, the prose or comment moves with it: the `foreach` "notice" sentence, two comments in the precedence example, the `mb_detect_encoding()` comment, and the anonymous class note. The two reflection examples needed replacing outright — they used `array_merge()` to show that internals report no return type, which it now does, so a method with a tentative return type takes its place.

<details>
<summary>Pages to check with Run code</summary>

Press Run code and the output that appears is what this PR puts in the `<screen>`. The static one still on these pages is the "before" side of the diff, until the manual is rebuilt.

- https://www.php.net/manual/en/language.operators.precedence.php
- https://www.php.net/manual/en/language.operators.type.php
- https://www.php.net/manual/en/control-structures.foreach.php
- https://www.php.net/manual/en/backedenum.from.php
- https://www.php.net/manual/en/reflectionfunctionabstract.getreturntype.php
- https://www.php.net/manual/en/reflectionfunctionabstract.hasreturntype.php
- https://www.php.net/manual/en/reflectionclass.tostring.php
- https://www.php.net/manual/en/language.oop5.anonymous.php
- https://www.php.net/manual/en/language.oop5.magic.php
- https://www.php.net/manual/en/function.mb-detect-encoding.php

</details>
---

**P.S.** — after the review on #5718, every example in this PR was re-run on 8.2, 8.3, 8.4 and 8.5 rather than trusted to Run code, which is one wasm build of one version.

That turned up one case: `mb_detect_encoding()`'s non-strict change landed in 8.3.0, so 8.2 still prints the old result and a single screen cannot cover both. Both examples on that page are now split with `&example.outputs.82;` / `&example.outputs.83;` — the basic example was already showing only the 8.3 behaviour, so it is split too rather than leaving one page inconsistent with itself.

Everything else agrees across 8.2–8.5, including the line numbers in the two `Fatal error:` screens and the `ReflectionClass::__toString()` dump of `Exception`.

### DIFF
--- a/language/control-structures/foreach.xml
+++ b/language/control-structures/foreach.xml
@@ -221,7 +221,7 @@ foreach ($array as [, , $c]) {
   </para>
 
   <para>
-   A notice will be generated if there aren't enough array elements to fill
+   A warning will be generated if there aren't enough array elements to fill
    the <function>list</function>:
 
    <informalexample>
@@ -241,10 +241,10 @@ foreach ($array as [$a, $b, $c]) {
     &example.outputs;
     <screen>
 <![CDATA[
-Notice: Undefined offset: 2 in example.php on line 7
+Warning: Undefined array key 2 in script on line 7
 A: 1; B: 2; C:
 
-Notice: Undefined offset: 2 in example.php on line 7
+Warning: Undefined array key 2 in script on line 7
 A: 3; B: 4; C:
 ]]>
     </screen>
--- a/language/control-structures/foreach.xml
+++ b/language/control-structures/foreach.xml
@@ -222,10 +222,11 @@ foreach ($array as [, , $c]) {
 
   <para>
    A warning will be generated if there aren't enough array elements to fill
-   the <function>list</function>:
+   the <function>list</function>.
+   Prior to PHP 8.0.0 a notice was generated instead:
 
    <informalexample>
-    <programlisting role="php">
+    <programlisting role="php" annotations="non-interactive">
 <![CDATA[
 <?php
 $array = [
@@ -238,7 +239,17 @@ foreach ($array as [$a, $b, $c]) {
 }
 ]]>
     </programlisting>
-    &example.outputs;
+    &example.outputs.71;
+    <screen>
+<![CDATA[
+Notice: Undefined offset: 2 in script on line 7
+A: 1; B: 2; C:
+
+Notice: Undefined offset: 2 in script on line 7
+A: 3; B: 4; C:
+]]>
+    </screen>
+    &example.outputs.8;
     <screen>
 <![CDATA[
 Warning: Undefined array key 2 in script on line 7
--- a/language/control-structures/foreach.xml
+++ b/language/control-structures/foreach.xml
@@ -221,9 +221,8 @@ foreach ($array as [, , $c]) {
   </para>
 
   <para>
-   A warning will be generated if there aren't enough array elements to fill
-   the <function>list</function>.
-   Prior to PHP 8.0.0 a notice was generated instead:
+   If there are not enough array elements to populate the variables,
+   a warning is emitted; prior to PHP 8.0.0, a notice was emitted instead:
 
    <informalexample>
     <programlisting role="php" annotations="non-interactive">
--- a/language/oop5/anonymous.xml
+++ b/language/oop5/anonymous.xml
@@ -63,8 +63,8 @@ var_dump(new class(10) extends SomeClass implements SomeInterface {
   &example.outputs;
   <screen>
 <![CDATA[
-object(class@anonymous)#1 (1) {
-  ["Command line code0x104c5b612":"class@anonymous":private]=>
+object(SomeClass@anonymous)#1 (1) {
+  ["num":"SomeClass@anonymous":private]=>
   int(10)
 }
 ]]>
@@ -156,7 +156,8 @@ same class
   <simpara>
    Note that anonymous classes are assigned a name by the engine, as
    demonstrated in the following example. This name has to be regarded as an
-   implementation detail, which should not be relied upon.
+   implementation detail, which should not be relied upon. The generated name
+   contains a NUL byte, which is not visible in the output below.
   </simpara>
   <informalexample>
    <programlisting role="php">
@@ -168,7 +169,7 @@ echo get_class(new class {});
   &example.outputs.similar;
   <screen>
 <![CDATA[
-class@anonymous/in/oNi1A0x7f8636ad2021
+class@anonymousscript:2$0
 ]]>
    </screen>
   </informalexample>
--- a/language/oop5/anonymous.xml
+++ b/language/oop5/anonymous.xml
@@ -168,7 +168,7 @@ echo get_class(new class {});
   &example.outputs.similar;
   <screen>
 <![CDATA[
-class@anonymousscript:2$0
+class@anonymous/in/oNi1A0x7f8636ad2021
 ]]>
    </screen>
   </informalexample>
--- a/language/oop5/anonymous.xml
+++ b/language/oop5/anonymous.xml
@@ -156,8 +156,7 @@ same class
   <simpara>
    Note that anonymous classes are assigned a name by the engine, as
    demonstrated in the following example. This name has to be regarded as an
-   implementation detail, which should not be relied upon. The generated name
-   contains a NUL byte, which is not visible in the output below.
+   implementation detail, which should not be relied upon.
   </simpara>
   <informalexample>
    <programlisting role="php">
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -527,7 +527,7 @@ var_dump($c);
     &example.outputs;
     <screen>
 <![CDATA[
-string(60) "A::__set_state(array(
+string(61) "\A::__set_state(array(
    'var1' => 5,
    'var2' => 'foo',
 ))"
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -495,7 +495,7 @@ Array
    </para>
    <example>
     <title>Using <link linkend="object.set-state">__set_state()</link></title>
-    <programlisting role="php" annotations="non-interactive">
+    <programlisting role="php">
 <![CDATA[
 <?php
 
@@ -524,22 +524,7 @@ var_dump($c);
 ?>
 ]]>
     </programlisting>
-    &example.outputs.81;
-    <screen>
-<![CDATA[
-string(60) "A::__set_state(array(
-   'var1' => 5,
-   'var2' => 'foo',
-))"
-object(A)#2 (2) {
-  ["var1"]=>
-  int(5)
-  ["var2"]=>
-  string(3) "foo"
-}
-]]>
-    </screen>
-    &example.outputs.82;
+    &example.outputs;
     <screen>
 <![CDATA[
 string(61) "\A::__set_state(array(
--- a/language/oop5/magic.xml
+++ b/language/oop5/magic.xml
@@ -495,7 +495,7 @@ Array
    </para>
    <example>
     <title>Using <link linkend="object.set-state">__set_state()</link></title>
-    <programlisting role="php">
+    <programlisting role="php" annotations="non-interactive">
 <![CDATA[
 <?php
 
@@ -524,7 +524,22 @@ var_dump($c);
 ?>
 ]]>
     </programlisting>
-    &example.outputs;
+    &example.outputs.81;
+    <screen>
+<![CDATA[
+string(60) "A::__set_state(array(
+   'var1' => 5,
+   'var2' => 'foo',
+))"
+object(A)#2 (2) {
+  ["var1"]=>
+  int(5)
+  ["var2"]=>
+  string(3) "foo"
+}
+]]>
+    </screen>
+    &example.outputs.82;
     <screen>
 <![CDATA[
 string(61) "\A::__set_state(array(
--- a/language/operators/precedence.xml
+++ b/language/operators/precedence.xml
@@ -372,36 +372,6 @@ $array[$i] = $i++; // may set either index 1 or 2
    </programlisting>
   </example>
   <example>
-   <title><literal>+</literal>, <literal>-</literal> and <literal>.</literal> precedence</title>
-   <programlisting role="php">
-<![CDATA[
-<?php
-$x = 4;
-// "-" binds tighter than ".", so $x-1 is one operand of the concatenation:
-echo "x minus one equals " . $x-1 . ", or so I hope\n";
-
-// the same grouping, made explicit:
-echo "x minus one equals " . ($x-1) . ", or so I hope\n";
-
-// this is not allowed, and throws a TypeError:
-echo (("x minus one equals " . $x) - 1) . ", or so I hope\n";
-?>
-]]>
-   </programlisting>
-   &example.outputs;
-   <screen>
-<![CDATA[
-x minus one equals 3, or so I hope
-x minus one equals 3, or so I hope
-
-Fatal error: Uncaught TypeError: Unsupported operand types: string - int in script:10
-Stack trace:
-#0 {main}
-  thrown in script on line 10
-]]>
-   </screen>
-  </example>
-  <example>
    <title>Prior to PHP 8, <literal>+</literal>, <literal>-</literal> and <literal>.</literal> had the same precedence</title>
    <programlisting role="php" annotations="non-interactive">
 <![CDATA[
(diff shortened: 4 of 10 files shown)


language/operators/precedence.xml
```diff
@@ -391,9 +391,13 @@ echo (("x minus one equals " . $x) - 1) . ", or so I hope\n";
...
--1, or so I hope
-Fatal error: Uncaught TypeError: Unsupported operand types: string - int
+x minus one equals 3, or so I hope
+x minus one equals 3, or so I hope
+
+Fatal error: Uncaught TypeError: Unsupported operand types: string - int in script:10
+Stack trace:
+#0 {main}
+  thrown in script on line 10
 ]]>
```
kamil-tekiela: I have not looked at the rest of the changes in this PR yet. If the behaviour changed, then we need to maintain both outputs. In this case, the example is only applicable to PHP 7 (it looks like the output was incorrect even with PHP 7), but we need to keep the example until PHP 9 is released. So if you want to show that this behaviour changed in PHP 8, first you need to fix the example so that it makes sense for PHP 7, and then add a note that as of PHP 8, this has been fixed and is no longer an issue. You should not change the code comments as they are correct and point out the pitfall. The output should include the warnings too. There should also be a note about the deprecation in PHP 7.4. 

The correct output:
```
Warning: A non-numeric value encountered in /in/mFYgu on line 4
-1, or so I hope
x minus one equals 3, or so I hope

Warning: A non-numeric value encountered in /in/mFYgu on line 10
-1, or so I hope
```
Although this warning has been there only since PHP 7.1. 

---

We could make the example represent the current state, but that would be incorrect in this case. We do not want to preserve this example going forward. It only exists to warn people of this historical quirk. So in my opinion, the most beneficial is to show what it was like between 7.1 and 7.4 and add a note for 7.4 and 8.0 changes.

language/operators/type.xml
```diff
@@ -188,7 +188,7 @@ var_dump(FALSE instanceof stdClass);
 bool(false)
 bool(false)
 bool(false)
-PHP Fatal error:  instanceof expects an object instance, constant given
+bool(false)
```
kamil-tekiela, line 194: Ok, but we need to retain the error message. Perhaps split it into two outputs per version? Or just add a code comment on `var_dump(FALSE instanceof stdClass);` saying it would throw with this error message?

language/operators/precedence.xml
kamil-tekiela: I just saw that there is another example underneath which shows what it was before. I don't think that's right. Both code blocks are the same. This needs to be merged as I said above.

KentarouTakeda: Merged — the current-behaviour example is gone, and this one keeps its original comments untouched.

The output includes the warnings now. Measured: 7.1, 7.2 and 7.3 all print the two `A non-numeric value encountered` warnings, and the screen that was here matched PHP 7.0 exactly, the one 7.x that doesn't warn. Hence `&example.outputs.71;` rather than `&example.outputs.7;`. The 7.4 deprecation is in a note below, as you asked.

One thing I did differently: the 8.0 change is a second screen rather than only a note, following your "if the behaviour changed, then we need to maintain both outputs". It also shows that the spelled-out grouping now throws, so the third line never runs. Say the word and it goes back to prose.

I kept `annotations="non-interactive"` on the example — Run code overwrites the first screen in place, caption included, so on a split example the older screen disappears on the first click. More on that in a comment on the PR.

language/operators/type.xml
KentarouTakeda: Took the code comment. A split would mislead here: prior to 7.3 this is a compile-time error, so 7.2 prints the fatal line and nothing else — none of the three `bool(false)` above it ever run. The screen that was here, three bools followed by the fatal error, was never any version's output.

That's why the comment says "so nothing above ran". The 7.3 boundary itself is already covered by the example directly below, which uses `&example.outputs.73;`.

KentarouTakeda:
Following up on "if the behaviour changed, then we need to maintain both outputs" — since you hadn't looked at the rest of the PR yet, I went back over every hunk with one question: was the screen that was there an actual capture of a version the manual still covers? Every answer below is measured, not guessed.

**Split, because there was a real output to keep**

- `foreach` — the old screen matches 7.4 exactly, so it is now `&example.outputs.71;` plus `&example.outputs.8;`. Not `&example.outputs.7;`, because the short list syntax doesn't parse on 7.0.
- `__set_state()` — the old screen matches 8.1 exactly, so it is now `&example.outputs.81;` plus `&example.outputs.82;`.

**Not split, because the old screen was no version's output**

- `ReflectionClass::__toString()` — the old screen has ten methods and no `__wakeup`, so it predates 5.6. Nothing in the ten files this PR touches refers to anything older than 7.1, and a faithful PHP 5 screen would run to sixty lines.
- anonymous classes — 7.4 prints `["num":"class@anonymous":private]`, not the `["Command line code0x104c5b612":...]` that was there. The 8.0 naming change is real, but there is no old output to preserve, and the note on that example already says the generated name must not be relied upon.
- `BackedEnum::from()` — the message quoting did change in 8.2, but the old screen also dropped the stack trace that every version prints, and had the line number wrong.

**One thing worth knowing about the split pattern**

Run code replaces the first screen in place, caption included (`interactive-examples.js`), and leaves later screens alone. On an example with two screens that destroys the older one on the first click — the very screen the split exists to preserve.

It is visible today on language.types.array.php: press Run code and the "PHP 8.0" screen becomes "PHP 8.5.0" carrying the 8.1 output, leaving two identical outputs under different labels. The markup there is fine; it is the runner that assumes one screen per example.

Every example I split in this PR is therefore marked `annotations="non-interactive"`. I can send the same one-line change for array.xml, though the better fix may be in web-php: when an example has more than one screen, append the live output rather than overwrite, or target the screen whose caption matches the running version.

language/operators/precedence.xml
kamil-tekiela: BTW are doing all of it with Claude? If so it would be nice if you gave it the proper attribution like in https://github.com/php/doc-en/commit/0009ebd09f938e8f066779bb9bb774baed10bd8a

KentarouTakeda: Yes. I'll add the trailer from here on — thanks for the pointer.

language/control-structures/foreach.xml
```diff
@@ -221,11 +221,12 @@ foreach ($array as [, , $c]) {
   </para>
 
   <para>
-   A notice will be generated if there aren't enough array elements to fill
-   the <function>list</function>:
+   A warning will be generated if there aren't enough array elements to fill
+   the <function>list</function>.
+   Prior to PHP 8.0.0 a notice was generated instead:
```
kamil-tekiela:
```suggestion
   If there are not enough array elements to populate the variables,
   a warning is emitted; prior to PHP 8.0.0, a notice was emitted instead:
```

language/oop5/magic.xml
```diff
@@ -537,6 +537,21 @@ object(A)#2 (2) {
   ["var2"]=>
   string(3) "foo"
 }
+]]>
+    </screen>
+    &example.outputs.82;
+    <screen>
+<![CDATA[
+string(61) "\A::__set_state(array(
```
kamil-tekiela, line 530: I can't even remember what this change was. Is this something that we document or is this an implementation detail?

language/oop5/anonymous.xml
```diff
@@ -168,7 +169,7 @@ echo get_class(new class {});
   &example.outputs.similar;
   <screen>
 <![CDATA[
-class@anonymous/in/oNi1A0x7f8636ad2021
+class@anonymousscript:2$0
```
kamil-tekiela: What's the difference? Isn't it the same thing?

language/oop5/anonymous.xml
```diff
@@ -153,11 +153,12 @@ same class
...
 
  <note>
-  <para>
+  <simpara>
    Note that anonymous classes are assigned a name by the engine, as
    demonstrated in the following example. This name has to be regarded an
-   implementation detail, which should not be relied upon.
-  </para>
+   implementation detail, which should not be relied upon. The generated name
+   contains a NUL byte, which is not visible in the output below.
```
kamil-tekiela: Where is the NUL byte? Is it always there? What's the significance of it?

language/oop5/anonymous.xml
```diff
@@ -63,8 +63,8 @@ var_dump(new class(10) extends SomeClass implements SomeInterface {
   &example.outputs;
   <screen>
 <![CDATA[
-object(class@anonymous)#1 (1) {
-  ["Command line code0x104c5b612":"class@anonymous":private]=>
+object(SomeClass@anonymous)#1 (1) {
+  ["num":"SomeClass@anonymous":private]=>
```
kamil-tekiela, line 67: Same here, I am not actually sure what's changed. I can see that PHP 8 uses the parent class name, but is this important? Is this intentional or just an implementation detail?

language/operators/type.xml
```diff
@@ -178,6 +178,9 @@ $c = fopen('/tmp/', 'r');
 var_dump($a instanceof stdClass); // $a is an integer
 var_dump($b instanceof stdClass); // $b is NULL
 var_dump($c instanceof stdClass); // $c is a resource
+
+// Prior to PHP 7.3.0 this was a compile-time error, so nothing above ran:
```
kamil-tekiela:
```suggestion
// Prior to PHP 7.3.0, using a constant would trigger a compile-time fatal error with the following message:
```

language/predefined/backedenum/from.xml
```diff
@@ -71,7 +71,11 @@ $b = Suit::from('B');
 <![CDATA[
 enum(Suit::Hearts)
 
-Fatal error: Uncaught ValueError: "B" is not a valid backing value for enum "Suit" in /file.php:15
+Fatal error: Uncaught ValueError: "B" is not a valid backing value for enum Suit in script:14
+Stack trace:
+#0 script(14): Suit::from('B')
+#1 {main}
+  thrown in script on line 14
```
kamil-tekiela: Just like before, I don't think we should be documenting these stack traces. Just keep the first line and use `&example.outputs.similar;`

reference/mbstring/functions/mb-detect-encoding.xml
```diff
@@ -158,7 +158,16 @@ var_dump(mb_detect_encoding($str, $encodings));
 ?>
 ]]>
     </programlisting>
-    &example.outputs;
+    &example.outputs.82;
```
kamil-tekiela: Why 8.2? What changed in 8.2 that we document the difference in output? I can see no related information on this page that would explain this. 

I would lean towards rejecting this change. Unless we have a basis to add the split example, we should only document the current behaviour. For all I know, the change in behaviour came from a bug fix or some internal refactoring.

reference/reflection/reflectionclass/tostring.xml
```diff
@@ -46,7 +46,7 @@ echo $reflectionClass->__toString();
     &example.outputs;
```
kamil-tekiela: Yeah ok, we can update the output to match the current state, but it's not really important. The example should use `&example.outputs.similar;` IMHO because the output can change from version to version.

reference/reflection/reflectionfunctionabstract/getreturntype.xml
```diff
@@ -56,35 +56,38 @@ int
    </example>
 
    <example>
-    <title>Usage on built-in functions</title>
+    <title>Methods with a tentative return type</title>
```
kamil-tekiela: I don't know about the last two reflection changes. They seem to be unrelated to this PR. They should go in their own dedicated PR.

language/oop5/magic.xml
KentarouTakeda: Documented and intentional — migration82 lists it under incompatible changes ("var_export() no longer omits the leading backslash for exported classes").

I dropped the split anyway. This example is about `__set_state()`; the backslash is a `var_export()` detail. 8.1 is EOL, so the current output covers every supported version.

language/oop5/anonymous.xml
KentarouTakeda: Not quite. The old string ends in `0x7f8636ad2021`; anonymous class names only had that shape before 7.4. Measured: 7.0 and 7.2 print `...php0xffff9202a039`, 7.4 through 8.5 print `...php:2$0`. That screen cannot occur on any supported version.

`&example.outputs.similar;` covers the file path, which is environment-dependent. The format is not.

language/oop5/anonymous.xml
KentarouTakeda: Dropped the sentence. It sits right after `class@anonymous` and is always there — migration80 already says so ("The name shown above is still followed by a NUL byte and a unique suffix") — but that is exactly the implementation detail the note above tells people not to rely on, and it would go into every translation.

language/oop5/anonymous.xml
KentarouTakeda: Intentional — migration80 lists it under incompatible changes, with examples for `ParentClass@anonymous`, `FirstInterface@anonymous` and `class@anonymous`.

But the naming is not really why this hunk exists. The old screen printed `["Command line code0x104c5b612":"class@anonymous":private]`, where the property-name slot holds part of the class name. 7.4 prints `["num":"class@anonymous":private]`. It was a broken capture rather than an outdated one.

language/predefined/backedenum/from.xml
KentarouTakeda: Done — first line only, with `&example.outputs.similar;`.

reference/mbstring/functions/mb-detect-encoding.xml
KentarouTakeda: You guessed right — withdrawn. migration83 says the non-strict mode "now behaves as described in the documentation", so the old behaviour contradicted the documented contract and the change was a bug fix. Labelling it as the 8.2 output would enshrine a bug.

Same reasoning you applied to the `date_sun_info()` screen in #5718.

reference/reflection/reflectionclass/tostring.xml
KentarouTakeda: Done. You are right that it moves: 8.2 through 8.5 are identical, but 7.4, 7.0 and 5.6 each differ. The screen that was here matches 5.6 minus `__wakeup`, so it predates that.

reference/reflection/reflectionfunctionabstract/getreturntype.xml
KentarouTakeda: Half right. The trigger is in scope: `array_merge` got `: array` in its stub in 8.0, so `getReturnType()` stopped returning null and the prose under the example ("many internal functions do not have types ... best to avoid using this method on built-in functions") became false. The screen could not be fixed on its own.

Your point lands on my replacement, which pulled in `getTentativeReturnType()` — I removed that. The example uses `fopen()` now, which returns null on 7.0 through 8.5 alike, so no version split either. 24 internal functions still have no declared return type in 8.5, nearly all of them returning a resource, which cannot be written as a type declaration.

language/oop5/anonymous.xml
kamil-tekiela: Yeah, but it doesn't matter what the format is, right? What makes you think that "That screen cannot occur on any supported version."? Where in the documentation does it say that this is the value to be expected? If it's an implementation detail, then we do not document it and we don't care if the output shows something different.

KentarouTakeda:
Thanks for the review — it made this PR smaller and more accurate, and tracing each screen back to a version taught me a lot of PHP history I did not know.

kamil-tekiela:
I appreciate your PR, but I have to be honest, it is not an easy one to review. For each change, I have to check what changed, why it changed, whether it is documented on the page or not, and which versions it covers. And it doesn't help that I am talking to Claude instead of a human being. 

I would recommend this course of action: for each file open a separate PR and in the description put a short 1-paragraph justification for why the change is made. It would help not only with the review, but also with future git bisect.

language/oop5/anonymous.xml
KentarouTakeda: You are right — the format does not matter for a value we refuse to document. Reverted.

KentarouTakeda:
Fair. Putting the reasoning in the PR doesn't fix it by itself, and neither does trimming the comments. Ten changes with their own version histories are ten things for you to check, wherever I write them down. I'll cut the PRs down, and keep the justification to something you can check in a click, the changelog or a migration guide, rather than my own test runs.

On the last point: the judgement calls here are mine. I gave you too much to argue with and not enough that you could just check.

reference/mbstring/functions/mb-detect-encoding.xml
lacatoire, line 115:
```suggestion
    <tbody>
     <row>
      <entry>8.3.0</entry>
      <entry>
       In non-strict mode, <function>mb_detect_encoding</function> now returns
       the closest matching encoding. Previously, it could return &false; or
       an encoding eliminated later in the string.
      </entry>
     </row>
```

PR #5721, opened 2026-07-28
uri: Fix Uri\WhatWg\Url screens that ignore URL normalization

KentarouTakeda:
Third batch from the audit behind #5718: screens that do not match what their example prints.

`Uri\WhatWg\Url::getPort()` and `withPort()` echoed a port that WHATWG normalization removes, so both printed nothing while the screens claimed 443. They use `var_dump()` now, and show a port that survives normalization and one that does not. `Url::parse()` gains the `/` that normalization adds to an empty path.

Three more screens were here and have been dropped. Two had wrong values that nothing in the examples depends on, and the third is now #5740.

Press Run code on these pages and the output that appears is what this PR puts in the `<screen>`:

- https://www.php.net/manual/en/uri-whatwg-url.getport.php
- https://www.php.net/manual/en/uri-whatwg-url.withport.php
- https://www.php.net/manual/en/uri-whatwg-url.parse.php

### DIFF
--- a/reference/uri/uri/whatwg/url/getport.xml
+++ b/reference/uri/uri/whatwg/url/getport.xml
@@ -35,16 +35,20 @@
    <programlisting role="php">
 <![CDATA[
 <?php
-$url = new \Uri\WhatWg\Url("https://example.com:443");
+$url = new \Uri\WhatWg\Url("https://example.com:8080");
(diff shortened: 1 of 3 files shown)


TimWolla (approved):
Thank you!

PR #5722, opened 2026-07-28
Convert and move global.ent into doc-en

alfsb:
Move global.ent entirely into doc-en, all of it, including URLs, at least while manual split is not complete.

The only thing that is truly global resting in doc-base/entities/global.ent was URLs, but these are also "moved" here as to allow changes to be more easily done, without depending on PRs that cross repositories.

Tested on all languages. As an incidental first result, text-entitites.php found a duplication (url.libevent), already fixed. After, is to change all residual files in doc-base/entities to note these files are to be erased, when doc-ext recreates any (hopefully few) entities they need.

Comments and reviews are welcome. Plan to merge this in one week.


jordikroon:
I was asking you if this was all, but then the full list of links came in.
This looks great. I don't have much to say about the comment, let's see if we get any feedback from translators positive or negative if any.

I have a strong feeling a big chunk of these links are in third-party extensions. So it would do good to see this file become shorter over time when we migrate those to doc-ext.

alfsb:
After parts of manuals start moving into doc-ext, a tool that lists all entities, and then a tool that tries to build the manual with one entity erased, can be used to check unused entities in doc-en, and also translations, to shrink the language-snippets.ent.

I may create an entities/entities.remove.dtd file, to push these cases here, so no conversion into XML would be necessary.

lacatoire:
LGTM

PR #5724, opened 2026-07-29
Fix reference to non-existent "example three"

### LINKED ISSUE #5685: Refer to absent example
From manual page: https://php.net/language.basic-syntax.phptags

---

"Short tags" block refers to "example three", which is absent from the page.

GabrielCordeiroBarrosoTeles:
## Summary
- The short tags paragraph in `language.basic-syntax.phptags` referred to "example three", but the page only has a single `<example>` block containing three manually numbered code snippets (1., 2., 3.), not three separate DocBook examples.
- Reworded the reference to point to the third line of that example instead, so it doesn't refer to a non-existent "example three".

Closes #5685

### DIFF
--- a/language/basic-syntax.xml
+++ b/language/basic-syntax.xml
@@ -40,12 +40,12 @@
     </example>
    </para>
 
-   <para>
+   <simpara>
     Short tags (the third line in the example above) are available by default but can be disabled
     either via the <link linkend="ini.short-open-tag">short_open_tag</link>
     &php.ini; configuration file directive, or are disabled by default if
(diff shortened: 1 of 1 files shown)


language/basic-syntax.xml
```diff
@@ -41,7 +41,7 @@
    </para>
 
    <para>
```
jordikroon:
```suggestion
   <simpara>
```

language/basic-syntax.xml
jordikroon:
```suggestion
   </simpara>
```

jordikroon:
Thank you for your contribution!

PR #5725, opened 2026-07-30
Update LuaSandbox installation method

tstarling:
New releases will be in PIE and Debian.

MongoDB seemed to be the nearest analogy, but what we need to say here is a lot simpler, so I pared it down to a single section. I used `programlisting` for the shell commands, as in MongoDB.

### DIFF
--- a/reference/luasandbox/setup.xml
+++ b/reference/luasandbox/setup.xml
@@ -25,23 +25,35 @@
 
  <section xml:id="luasandbox.installation">
   &reftitle.install;
-  <simpara>
-   &pecl.moved;
-  </simpara>
-  <simpara>
-   &pecl.info;
-   <link xlink:href="&url.pecl.package;luasandbox">&url.pecl.package;luasandbox</link>.
-  </simpara>
+
   <para>
    If the operating system is Debian 10 or later, or Ubuntu 18.04 or later, then
    LuaSandbox should typically be installed from the package
    <literal>php-luasandbox</literal>:
-   <screen>
+
+   <programlisting role="shell">
 <![CDATA[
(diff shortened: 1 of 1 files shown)


reference/luasandbox/setup.xml
```diff
@@ -25,23 +25,33 @@
...
+   <programlisting role="shell">
 <![CDATA[
 sudo apt-get install php-luasandbox
 ]]>
-   </screen>
+   </programlisting>
   </para>
+
+  <para>
+   To install LuaSandbox using <link xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="https://github.com/php/pie">PIE</link>,
```
jordikroon: Please use the `&url.pie;` entity reference for href.

reference/luasandbox/setup.xml
```diff
@@ -25,23 +25,33 @@
...
+   run the following command:
+   <programlisting role="shell">
+<![CDATA[
+pie install wikimedia/luasandbox
+]]>
+   </programlisting>
+  </para>
+
+  <simpara>
```
jordikroon: Should we make this a note instead?

Also I think it's more consistent with the docs if we don't include "the" in the url.

jordikroon (changes requested):

reference/luasandbox/setup.xml
tstarling: Done

reference/luasandbox/setup.xml
tstarling: Done.

jordikroon (approved):

PR #5728, opened 2026-07-31
Fix grammar in oop5

kamil-tekiela:
I am somewhat less sure about these, so I appreciate any improvements.


kamil-tekiela:
The simpara check is super annoying. This should have been done in one step mechanically.

lacatoire (approved):

PR #5737, opened 2026-08-03
ext/simplexml: document that get methods and casts no longer implicitly rewind (PHP 8.4)

lacatoire:
Resolves the following issue:

https://github.com/orgs/php/projects/11/views/1?pane=issue&itemId=199826662


jordikroon:
Could you do a rebase with master?

PR #5739, opened 2026-08-04
ext/pcre: document variable-length lookbehind and longer subpattern names (PHP 8.4)

lacatoire:
Resolves the following issues:

https://github.com/orgs/php/projects/11/views/1?pane=issue&itemId=199826167
https://github.com/orgs/php/projects/11/views/1?pane=issue&itemId=199826180

**Sources**

- variable-length lookbehind and the 128-character subpattern name limit come from the PCRE2 10.44 bundled in PHP 8.4.0: php/php-src@d1f14a46095571b1b6363fa7f83c6a681d1d02de (`ext/pcre: update to PCRE2 v10.44`, php/php-src#14498)

Checked on `php:8.3-cli` (PCRE 10.42) and `php:8.4-cli` (PCRE 10.44):

- `(?<!dogs?|cats?)bar` and `(?<=ab(c|de))x` fail to compile on 8.3 and compile and match on 8.4
- the longest accepted subpattern name is 32 characters on 8.3 and 128 on 8.4

### DIFF
--- a/reference/pcre/pattern.syntax.xml
+++ b/reference/pcre/pattern.syntax.xml
@@ -1510,6 +1510,10 @@
    also by name. There are two alternative syntaxes
    <literal>(?&lt;name&gt;pattern)</literal> and <literal>(?'name'pattern)</literal>.
   </para>
+  <simpara>
+   As of PHP 8.4.0, which bundles PCRE2 10.44, the maximum length of a
+   subpattern name is 128 characters; prior to PHP 8.4.0 it was 32 characters.
+  </simpara>
 
   <para>
    Sometimes it is necessary to have multiple matching, but alternating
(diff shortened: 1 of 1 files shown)


reference/pcre/pattern.syntax.xml
```diff
@@ -1510,6 +1510,10 @@
    also by name. There are two alternative syntaxes
    <literal>(?&lt;name&gt;pattern)</literal> and <literal>(?'name'pattern)</literal>.
   </para>
+  <simpara>
+   As of PHP 8.4.0, which bundles PCRE2 10.44, the maximum length of a
+   subpattern name is 128 characters; prior to PHP 8.4.0 it was 32 characters.
```
jordikroon:
```suggestion
   subpattern name is 128 characters; previously it was 32 characters.
```

reference/pcre/pattern.syntax.xml
```diff
@@ -1915,6 +1919,14 @@
    at the ends of strings; an example is given at the end
    of the section on once-only subpatterns.
   </para>
+  <simpara>
+   The fixed-length restriction described above applies prior to PHP 8.4.0.
```
jordikroon: We shouldn't mention that the behavior described above applies to older versions. Instead, we should move the older behaviour below and state that the current description applies to PHP 8.4.0 and later. The documentation should reflect the latest version by default, whereas behaviour from older versions should be documented separately as historical information.

jordikroon (changes requested):

jordikroon (approved):

PR #5745, opened 2026-08-07
Fix naming, paths, and add php-fpm to install example in dnf.xml

rbowen:
Use "Apache httpd" per project naming convention. Replace incorrect Debian-style /etc/php/8.3/ paths with the actual DNF-based distro layout (/etc/php.d/). Add php-fpm to the install example since it is the default SAPI on these systems. Note PHP-FPM as the service to restart.

### DIFF
--- a/install/unix/dnf.xml
+++ b/install/unix/dnf.xml
@@ -28,7 +28,7 @@
    <title>DNF Install Example</title>
    <programlisting role="shell">
 <![CDATA[
-# dnf install php php-common
+# dnf install php php-common php-fpm
 ]]>
    </programlisting>
   </example>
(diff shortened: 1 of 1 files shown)


install/unix/dnf.xml
```diff
@@ -86,10 +86,9 @@
   <simpara>
    DNF will automatically add the appropriate lines to the
    different &php.ini; related files like
-   <filename>/etc/php/8.3/php.ini</filename>,
-   <filename>/etc/php/8.3/conf.d/*.ini</filename>, etc. and depending on
+   <filename>/etc/php.d/*.ini</filename>, etc. and depending on
```
lacatoire, line 90:
```suggestion
   <filename>/etc/php.ini</filename>,
   <filename>/etc/php.d/*.ini</filename>, etc. and depending on
```

PR #5746, opened 2026-08-07
Fix grammar in language/types

kamil-tekiela:
The sentence in iterable.xml seems wrong or maybe I cannot understand it, but I don't know how to fix it.

### DIFF
--- a/language/types/mixed.xml
+++ b/language/types/mixed.xml
@@ -16,8 +16,8 @@
  </para>
 
  <para>
-  <type>mixed</type> is, in type theory parlance, the top type.
-  Meaning every other type is a subtype of it.
+  <type>mixed</type> is, in type theory parlance, the top type,
+  meaning every other type is a subtype of it.
  </para>
 
 </sect1>
--- a/language/types/mixed.xml
+++ b/language/types/mixed.xml
@@ -15,10 +15,10 @@
   Available as of PHP 8.0.0.
  </para>
 
- <para>
+ <simpara>
   <type>mixed</type> is, in type theory parlance, the top type,
   meaning every other type is a subtype of it.
- </para>
+ </simpara>
 
 </sect1>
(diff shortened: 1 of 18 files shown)


language/types/string.xml
```diff
@@ -13,7 +13,7 @@
 
  <note>
   <simpara>
-   On 32-bit builds, a <type>string</type> can be as large as up to 2GB
+   On 32-bit builds, a <type>string</type> can be as large as 2GB
```
jordikroon: How about this:

`On 32-bit builds, a <type>string</type> can be up to 2GB (2147483647 bytes).`.

language/types/mixed.xml
```diff
@@ -15,10 +15,10 @@
...
  </para>
 
- <para>
-  <type>mixed</type> is, in type theory parlance, the top type.
-  Meaning every other type is a subtype of it.
- </para>
+ <simpara>
+  <type>mixed</type> is, in type theory parlance, the top type,
+  meaning every other type is a subtype of it.
+ </simpara>
```
jordikroon, line 21: While I agree with the changes. This phrase is that advanced that I don't believe the majority of the people that read the documentation will understand. Read: Non-native speakers.

My suggestion:
`<type>mixed</type> is a special type that accepts values of any other type.`

jordikroon (commented):
Some minor (non blocking) nits. Feel free to choose whatever you feel right with.

language/types/mixed.xml
kamil-tekiela: That's already on the page a couple of lines above it. I agree, it's difficult to understand, but not sure that dumbing it down is a good idea. It says exactly what it was supposed to say for people who are interested in type theory. An average reader will stop on the first paragraph.

jordikroon (approved):

PR #5747, opened 2026-08-07
Example emits warning unrelated to content

### LINKED ISSUE #5729: Example emits warning unrelated to content
### Affected page

https://www.php.net/manual/en/language.basic-syntax.phpmode.php

### Issue description

Here is the code provided for *Example #&NoBreak;2 Advanced escaping using conditions*:
```php
<?php if ($expression == true): ?>
  This will show if the expression is true.
<?php else: ?>
  Otherwise this will show.
<?php endif; ?>
```
When running this code, the following warning is printed (along with the text output "Otherwise this will show."):
```
Warning: Undefined variable $expression in script on line 1
```
This is correct behavior for the example code, since `$expression` is indeed undefined. But that isn't related to the behavior the example is supposed to be showing. The purpose of the section is to demonstrate conditional text skipping. This bug in the script only distracts from the intended content.

It's possible that this is intentional (maybe to show the interaction of the text output and the PHP engine's error output?), but the surrounding documentation makes no reference to it, so I'm guessing that it isn't.

I'm very new to PHP, so my apologies if I'm just misunderstanding the documentation and/or script.

### Steps to reproduce

1. Go to the page: [with the example's header highlighted](https://www.php.net/manual/en/language.basic-syntax.phpmode.php#:~:text=Example%20#2%20Advanced%20escaping%20using%20conditions)
2. Press `Run code` underneath Example #&NoBreak;2

### Suggested fix

Here's my suggestion for how to make this example more helpful:
```php
<?php if (1 == 2): ?>
  This will show if the expression is true.
<?php else: ?>
  Otherwise this will show.
<?php endif; ?>
```
This has the same text output as the original and it still demonstrates that an evaluable expression can be used as the predicate of an if-then-else statement to conditionally skip HTML text. At the same time, it doesn't involve the unrelated semantics of comparing values to undefined variables.

Of course, this is by no means the only solution.

kamil-tekiela:
Closes #5729

### DIFF
--- a/language/basic-syntax.xml
+++ b/language/basic-syntax.xml
@@ -117,7 +117,7 @@ echo "Last statement\n";
      <title>Advanced escaping using conditions</title>
      <programlisting role="php">
 <![CDATA[
-<?php if ($expression == true): ?>
+<?php if (true): ?>
   This will show if the expression is true.
 <?php else: ?>
   Otherwise this will show.


jordikroon (approved):
I agree that `true` is a better choice than the suggested example.

PR #5751, opened 2026-08-09
sigmoidalcontrastimage.xml Amend parameter names

mmalferov:
Integer values in the description are confusing, whereas numbers in the x.xx format seem to be perceived more easily, and the type remains consistent

### DIFF
--- a/reference/imagick/imagick/sigmoidalcontrastimage.xml
+++ b/reference/imagick/imagick/sigmoidalcontrastimage.xml
@@ -24,7 +24,7 @@
    <parameter>beta</parameter> indicates where midtones fall in the resultant
    image, as a fraction of the quantum range: <literal>0.0</literal> is white,
    <literal>0.5</literal> is middle-gray, <literal>1.0</literal> is black
-   (multiply by <code>getQuantumRange()['quantumRangeLong']</code> to get the value to 
+   (multiply by <code>getQuantumRange()['quantumRangeLong']</code> to get the value to
    pass).
   </simpara>
   <simpara>
@@ -88,7 +88,7 @@
    &imagick.imagick.throws;
   </para>
  </refsect1>
- 
+
  <refsect1 role="examples">
   &reftitle.examples;
   <example>
@@ -108,7 +108,7 @@ function generateBlendImage($width, $height, $alpha = 10, $beta = 0.5)
     $quanta = $imagick->getQuantumRange();
     $imagick->sigmoidalContrastImage(true, $alpha, $beta * $quanta["quantumRangeLong"]);
 
-    return $imagick; 
+    return $imagick;
 }
 ]]>
    </programlisting>
--- a/reference/imagick/imagick/sigmoidalcontrastimage.xml
+++ b/reference/imagick/imagick/sigmoidalcontrastimage.xml
@@ -18,10 +18,10 @@
   <para>
    Adjusts the contrast of an image with a non-linear sigmoidal contrast
    algorithm. Increase the contrast of the image using a sigmoidal transfer
-   function without saturating highlights or shadows. Contrast indicates
-   how much to increase the contrast (0 is none; 3 is typical; 20 is
-   pushing it); mid-point indicates where midtones fall in the resultant
-   image (0 is white; 50 is middle-gray; 100 is black). Set sharpen to
+   function without saturating highlights or shadows. <parameter>alpha</parameter> indicates
+   how much to increase the contrast: 0.00 is none; 3.00 is typical; 20.00 is pushing it.
+   <parameter>beta</parameter> indicates where midtones fall in the resultant
+   image: 0.00 is white; 0.50 is middle-gray; 1.00 is black. Set <parameter>sharpen</parameter> to
    &true; to increase the image contrast otherwise the contrast is reduced.
   </para>
   <para>
@@ -46,7 +46,7 @@
      <term><parameter>alpha</parameter></term>
      <listitem>
       <para>
-       The amount of contrast to apply. 1 is very little, 5 is a significant amount, 20 is extreme.
+       The amount of contrast to apply. 1.00 is very little, 5.00 is a significant amount, 20.00 is extreme.
       </para>
      </listitem>
     </varlistentry>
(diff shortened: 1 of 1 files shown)


reference/imagick/imagick/sigmoidalcontrastimage.xml
```diff
@@ -37,33 +39,35 @@
...
+      </simpara>
      </listitem>
     </varlistentry>
     <varlistentry>
      <term><parameter>beta</parameter></term>
      <listitem>
       <simpara>
-       Where the midpoint of the gradient will be. This value should be in the range 0 to 1 - multiplied by the quantum value for ImageMagick.
+       Where the midpoint of the gradient will be. This value should be in the range <literal>0.00</literal>
+       to <literal>1.00</literal> - multiplied by the quantum value for ImageMagick.
```
kamil-tekiela, line 62: This is still super confusing to me, but I have no idea how to write it better.

reference/imagick/imagick/sigmoidalcontrastimage.xml
```diff
@@ -15,19 +15,21 @@
...
-   function without saturating highlights or shadows. Contrast indicates
-   how much to increase the contrast (0 is none; 3 is typical; 20 is
-   pushing it); mid-point indicates where midtones fall in the resultant
-   image (0 is white; 50 is middle-gray; 100 is black). Set sharpen to
+   function without saturating highlights or shadows. <parameter>alpha</parameter> indicates
+   how much to increase the contrast: <literal>0.00</literal> is none; <literal>3.00</literal> is
+   typical; <literal>20.00</literal> is pushing it.
+   <parameter>beta</parameter> indicates where midtones fall in the resultant
+   image: <literal>0.00</literal> is white; <literal>0.50</literal> is
+   middle-gray; <literal>1.00</literal> is black. Set <parameter>sharpen</parameter> to
```
kamil-tekiela: I don't think dividing by 100 is correct here. I don't know anything about Imagick, so I think keeping it 0, 50, 100 is safer.

reference/imagick/imagick/sigmoidalcontrastimage.xml
```diff
@@ -86,31 +90,28 @@
...
-     Create a gradient image using <function>Imagick::sigmoidalContrastImage</function>
-     suitable for blending two images together smoothly, with the blending
-     defined by $contrast and $the midpoint
-    </title>
-    <programlisting role="php">
+  <example>
+   <title>
+    Create a gradient image using <function>Imagick::sigmoidalContrastImage</function>
+    suitable for blending two images together smoothly, with the blending
+    defined by $alpha and $beta
```
kamil-tekiela:
```suggestion
    defined by <parameter>alpha</parameter> and <parameter>beta</parameter>
```

kamil-tekiela:
This page was based on https://phpimagick.com/Imagick/sigmoidalContrastImage?midpoint=0&sigmoidal_contrast=0.5&sharpening=Increase&image_path=Lorikeet so I think we should try to keep it the same as they document it in the official docs. 

It looks like we have the contrast and midpoint reversed.

kamil-tekiela:
Honestly, I think this user note is way better at explaining it than our whole manual page: https://www.php.net/manual/en/imagick.sigmoidalcontrastimage.php#113920

The whole page needs to be rewritten.

reference/imagick/imagick/sigmoidalcontrastimage.xml
```diff
@@ -15,19 +15,21 @@
...
-   how much to increase the contrast (0 is none; 3 is typical; 20 is
-   pushing it); mid-point indicates where midtones fall in the resultant
-   image (0 is white; 50 is middle-gray; 100 is black). Set sharpen to
+   function without saturating highlights or shadows. <parameter>alpha</parameter> indicates
+   how much to increase the contrast: <literal>0.00</literal> is none; <literal>3.00</literal> is
+   typical; <literal>20.00</literal> is pushing it.
+   <parameter>beta</parameter> indicates where midtones fall in the resultant
+   image: <literal>0.00</literal> is white; <literal>0.50</literal> is
+   middle-gray; <literal>1.00</literal> is black. Set <parameter>sharpen</parameter> to
    &true; to increase the image contrast otherwise the contrast is reduced.
```
lacatoire:
```suggestion
   image, as a fraction of the quantum range: <literal>0.0</literal> is white,
   <literal>0.5</literal> is middle-gray, <literal>1.0</literal> is black
   (multiply by <code>getQuantumRange()['quantumRangeLong']</code> to get the value to 
   pass).
```

lacatoire (approved):
The 0/50/100 scale comes from the ImageMagick command line, where the mid-point is written as a percentage. `Imagick::sigmoidalContrastImage()` takes an absolute quantum value instead, so those numbers do not carry over.

### Measurement
ImageMagick 7.1.1-43 Q16, Imagick 3.8.1, PHP 8.4. The filter is applied to a `gradient:black-white`, and the pixel that does not move — the midpoint of the transfer curve — is read back. `quantumRangeLong` is 65535.

| `beta` passed | reading | midtone |
| --- | --- | --- |
| `0` | raw | 0.9870 |
| `50` | raw | 0.9869 |
| `100` | raw | 0.9869 |
| `0.0 * QR` = 0 | fraction | 0.9870 |
| `0.5 * QR` = 32767.5 | fraction | 0.5050 |
| `1.0 * QR` = 65535 | fraction | 0.0136 |

Raw 0, 50 and 100 are indistinguishable: against a quantum range of 65535 they all mean white. The fraction form gives white, middle-gray and black as documented, and it is the only build-independent form, since `getQuantumRange()` returns 255 on a Q8 build and 65535 on Q16 — the same literal number would denote a different midtone.

```php
$im = new Imagick();
$im->newPseudoImage(256, 1, 'gradient:black-white');
$im->sigmoidalContrastImage(true, 10.0, $beta);
echo $im->getImagePixelColor(128, 0)->getColorValue(Imagick::COLOR_RED);
```

### The page contradicts itself today
The description says "0 is white; 50 is middle-gray; 100 is black", while the `beta` entry ten lines below says "range 0 to 1 - multiplied by the quantum value" — which is also the reading the example on the page relies on. Only one of the two can be right, and the measurement above says it is the second.

### Parameter order

The signature is `sigmoidalContrastImage(bool $sharpen, float $alpha, float $beta, int $channel)`. Holding `beta` at `0.5 * QR` and raising `alpha` widens the gradient around that midpoint:

| `alpha` | p25 | p75 |
| --- | --- | --- |
| `0` | 0.2510 | 0.7529 |
| `3` | 0.2189 | 0.7851 |
| `20` | 0.0068 | 0.9937 |

So `alpha` is the contrast and `beta` the mid-point, which is the order both the current page and this PR use.

This PR does not attempt the full rewrite mentioned above; it only corrects the scale and the parameter names. The rewrite is worth a separate issue.

PR #5752, opened 2026-08-09
Rewrite example on for page

kamil-tekiela:
I saw this example on `for` page, which appeared to me as misleading. I kept thinking about it all day and I decided that it needs to change. `count` was used as an example of an expensive operation, which it is not when used with arrays. It can be expensive on objects. What's more important, though, is that the example failed to explain why it's a bad idea to do it this way (it was much worse in the original version years ago). So the example needed modernizing and I took the inspiration from the user note on that page, which is now folded into the new example. 

User note: https://www.php.net/manual/en/control-structures.for.php#13463  
My explanation on Stack Overflow: https://stackoverflow.com/a/79991793/1839439

### DIFF
--- a/language/control-structures/for.xml
+++ b/language/control-structures/for.xml
@@ -171,21 +171,22 @@ for($i = 0, $peopleCount = getIterationCount($people); $i < $peopleCount; ++$i) 
   </informalexample>
  </para>
  <simpara>
-  In the example above, the invoked function always returns the same value.
-  Since the size never changes, the loop can be easily
-  optimized by using an intermediate variable to store the size instead
-  of repeatedly calling the same function. An added benefit is that the
-  number of iterations is now constant, which helps avoid bugs when the
-  array is modified during the loop.
+  In the example above, the function is called on every iteration of the
+  first loop, even though it always returns the same value. Storing that
+  value in a variable, as the second loop does, calls the function once.
+  Note that the number of iterations is then fixed: if the array is
+  modified inside the loop, the stored value no longer reflects its size.
  </simpara>
 
  <note>
   <simpara>
    The size of an <type>array</type> is stored with the array, which means that
    calling the built-in function <function>count</function> does not require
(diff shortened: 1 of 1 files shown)


language/control-structures/for.xml
```diff
@@ -142,97 +142,60 @@ endfor;
...
  <simpara>
-  The above code can be slow, because the array size is fetched on
-  every iteration. Since the size never changes, the loop can be easily
+  In the example above, the invoked function always returns the same value.
+  Since the size never changes, the loop can be easily
   optimized by using an intermediate variable to store the size instead
-  of repeatedly calling <function>count</function>:
+  of repeatedly calling the same function. An added benefit is that the
+  number of iterations is now constant, which helps avoid bugs when the
+  array is modified during the loop.
```
lacatoire:
```suggestion
  In the example above, the function is called on every iteration of the
  first loop, even though it always returns the same value. Storing that
  value in a variable, as the second loop does, calls the function once.
  Note that the number of iterations is then fixed: if the array is
  modified inside the loop, the stored value no longer reflects its size.
```

language/control-structures/for.xml
```diff
@@ -142,97 +142,60 @@ endfor;
...
-   </screen>
-  </informalexample>
- </para>
+ <note>
+  <simpara>
+   The size of an <type>array</type> is stored with the array, which means that
+   calling the built-in function <function>count</function> does not require
+   counting the elements and does not cause performance issues. However, care
+   should be taken when using <function>count</function> on objects implementing
+   <interfacename>Countable</interfacename>, as such calls can be more expensive.
```
lacatoire:
```suggestion
   counting the elements and does not cause performance issues. This does not
   apply to <constant>COUNT_RECURSIVE</constant>, which walks the whole array.
   Care should also be taken when using <function>count</function> on objects
   implementing <interfacename>Countable</interfacename>, as such calls can be
   more expensive.
```

PR #5754, opened 2026-08-10
strtr.xml Fix the confusing example, CS

mmalferov:
The first example — with multi-byte characters — seems confusing. The function replaces only the first byte of a multi-byte character `å`, and the output shows a black diamond with a question mark (U+FFFD).

Instead of a real example, the page actually offers pseudocode, which an unprepared reader has to slog through to finally understand the substitution algorithm: ä → a, å → a, ö → o. I don't know why 🤷‍♂️

### DIFF
--- a/reference/strings/functions/strtr.xml
+++ b/reference/strings/functions/strtr.xml
@@ -119,9 +119,15 @@ $original = "He1Lo, w0rld!";
 
 // With three arguments, strtr() performs byte-by-byte translation.
 $result = strtr($original, "1L0", "llo");
-echo $result, PHP_EOL;
+echo $result;
 ]]>
     </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+Hello, world!
+]]>
+    </screen>
    </example>
   </para>
   <para>
@@ -158,10 +164,25 @@ hello all, I said hi
 <![CDATA[
 <?php
 
-echo strtr("baab", "ab", "01"), "\n";
+/**
+ * Single-byte
+ */
+$singlebyte_string = "baab";
+echo strtr($singlebyte_string, "ab", "01"), PHP_EOL; // Byte-by-byte translation
 
 $replace_pairs = ["ab" => "01"];
-echo strtr("baab", $replace_pairs);
+echo strtr($singlebyte_string, $replace_pairs), PHP_EOL; // Substring translation
+
+/**
+ * Multi-byte
+ */
+$multibyte_string = "Ä"; // In UTF-8, encoded as two bytes: "\xC3\x84"
+
+$result = strtr($multibyte_string, 'Ä', 'A');
+echo bin2hex($result), PHP_EOL; // "4184", where "c3" becomes "41" ("A"), while "84" is kept and breaks UTF-8
+
+$replace_pairs = ['Ä' => 'A'];
+echo strtr($multibyte_string, $replace_pairs); // A correct multibyte substring translation
 ]]>
     </programlisting>
     &example.outputs;
@@ -169,6 +190,8 @@ echo strtr("baab", $replace_pairs);
 <![CDATA[
 1001
 ba01
+4184
+A
 ]]>
     </screen>
   </example>
--- a/reference/strings/functions/strtr.xml
+++ b/reference/strings/functions/strtr.xml
@@ -164,22 +164,18 @@ hello all, I said hi
 <![CDATA[
 <?php
 
-/**
- * Single-byte
- */
+// Single-byte characters
 $singlebyte_string = "baab";
-echo strtr($singlebyte_string, "ab", "01"), PHP_EOL; // Byte-by-byte translation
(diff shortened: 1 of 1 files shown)


reference/strings/functions/strtr.xml
```diff
@@ -114,13 +114,12 @@
...
-// In this form, strtr() does byte-by-byte translation
-// Therefore, we are assuming a single-byte encoding here:
-$addr = strtr($addr, "äåö", "aao");
-echo $addr, PHP_EOL;
-?>
+$original = "He1Lo, w0rld!";
+
+// With three arguments, strtr() performs byte-by-byte translation.
+$result = strtr($original, "1L0", "llo");
+echo $result, PHP_EOL;
```
kamil-tekiela: The outputs section is missing.

reference/strings/functions/strtr.xml
```diff
@@ -158,11 +157,11 @@ hello all, I said hi
...
 <?php
-echo strtr("baab", "ab", "01"),"\n";
 
-$trans = array("ab" => "01");
-echo strtr("baab", $trans);
-?>
+echo strtr("baab", "ab", "01"), "\n";
+
+$replace_pairs = ["ab" => "01"];
+echo strtr("baab", $replace_pairs);
```
kamil-tekiela: Since you removed the broken example from the first one, maybe it would be a good idea to add it here, so that readers see why it doesn't work with multibyte strings and 3-parameter mode. See https://www.php.net/manual/en/function.strtr.php#111270

kamil-tekiela (changes requested):

reference/strings/functions/strtr.xml
mmalferov: Done

reference/strings/functions/strtr.xml
mmalferov: Please take a look now. Is the example of replacement in a multi‑byte string clear?

reference/strings/functions/strtr.xml
```diff
@@ -158,18 +163,35 @@ hello all, I said hi
...
 <![CDATA[
 <?php
-echo strtr("baab", "ab", "01"),"\n";
 
-$trans = array("ab" => "01");
-echo strtr("baab", $trans);
-?>
+/**
+ * Single-byte
+ */
```
kamil-tekiela:
```suggestion
// Single-byte characters
```

reference/strings/functions/strtr.xml
```diff
@@ -158,18 +163,35 @@ hello all, I said hi
...
+$singlebyte_string = "baab";
+echo strtr($singlebyte_string, "ab", "01"), PHP_EOL; // Byte-by-byte translation
+
+$replace_pairs = ["ab" => "01"];
+echo strtr($singlebyte_string, $replace_pairs), PHP_EOL; // Substring translation
+
+/**
+ * Multi-byte
+ */
```
kamil-tekiela:
```suggestion
// Multi-byte characters
```

reference/strings/functions/strtr.xml
```diff
@@ -158,18 +163,35 @@ hello all, I said hi
...
-echo strtr("baab", "ab", "01"),"\n";
 
-$trans = array("ab" => "01");
-echo strtr("baab", $trans);
-?>
+/**
+ * Single-byte
+ */
+$singlebyte_string = "baab";
+echo strtr($singlebyte_string, "ab", "01"), PHP_EOL; // Byte-by-byte translation
```
kamil-tekiela:
```suggestion
echo strtr($singlebyte_string, "ab", "01"), "\n"; // Byte-by-byte translation
```

reference/strings/functions/strtr.xml
```diff
@@ -158,18 +163,35 @@ hello all, I said hi
...
+
+$replace_pairs = ["ab" => "01"];
+echo strtr($singlebyte_string, $replace_pairs), PHP_EOL; // Substring translation
```
kamil-tekiela:
```suggestion
echo strtr($singlebyte_string, $replace_pairs), "\n"; // Substring translation
```

reference/strings/functions/strtr.xml
```diff
@@ -158,18 +163,35 @@ hello all, I said hi
...
+
+/**
+ * Multi-byte
+ */
+$multibyte_string = "Ä"; // In UTF-8, encoded as two bytes: "\xC3\x84"
+
+$result = strtr($multibyte_string, 'Ä', 'A');
+echo bin2hex($result), PHP_EOL; // "4184", where "c3" becomes "41" ("A"), while "84" is kept and breaks UTF-8
```
kamil-tekiela:
```suggestion
echo bin2hex($result), "\n"; // "4184", where "c3" becomes "41" ("A"), while "84" is kept and breaks UTF-8
```

kamil-tekiela (changes requested):
Just some nits

kamil-tekiela (approved):

PR #5756, opened 2026-08-10
HaY Rename parseerror.xml to parseerror

HaYGiton:


kamil-tekiela:
I don't know what you were trying to do, or maybe you opened this PR by mistake, but we cannot accept this.

PR #5758, opened 2026-08-11
Document padding constants for OpenSSL functions.

### LINKED ISSUE #5757: openssl_sign padding parameter
### Affected page

https://www.php.net/manual/en/function.openssl-sign.php

### Current issue

> RSA PSS padding to use.

This is not actionable to select a padding.

`$padding` parameter can be either 0, OPENSSL_PKCS1_PSS_PADDING or OPENSSL_PKCS1_PADDING. I think the documentation should refer to those, because it is currently not clear how the parameter should be used.

Ideally this should be an enum instead of a int, but I guess it's too late now to change that.

### Suggested improvement

Point to OPENSSL_PKCS1_PSS_PADDING and OPENSSL_PKCS1_PADDING in $padding documentation.

https://github.com/php/php-src/pull/19432

kamil-tekiela:
Closes #5757

### DIFF
--- a/reference/openssl/functions/openssl-sign.xml
+++ b/reference/openssl/functions/openssl-sign.xml
@@ -72,7 +72,7 @@
      <term><parameter>padding</parameter></term>
      <listitem>
       <simpara>
-       RSA PSS padding to use. It can be either
+       RSA-PSS padding to use. It can be either
        <constant>OPENSSL_PKCS1_PADDING</constant>,
        <constant>OPENSSL_PKCS1_PSS_PADDING</constant>, or <literal>0</literal>.
       </simpara>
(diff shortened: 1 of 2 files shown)


reference/openssl/functions/openssl-sign.xml
```diff
@@ -71,7 +71,11 @@
     <varlistentry>
      <term><parameter>padding</parameter></term>
      <listitem>
-      <simpara>RSA PSS padding to use.</simpara>
+      <simpara>
+       RSA PSS padding to use. It can be either
```
jordikroon: [reference/openssl/constants.xml#L167](https://github.com/php/doc-en/blob/master/reference/openssl/constants.xml#L167) documents RSA-PSS (hyphenated). I suggest we go with that notation. Same for openssl_verify

jordikroon (approved):

PR #5760, opened 2026-08-13
Add warning for header_register_callback usage

beryllium:
Added a warning about callback de-registration in header_register_callback.

### DIFF
--- a/reference/network/functions/header-register-callback.xml
+++ b/reference/network/functions/header-register-callback.xml
@@ -22,12 +22,12 @@
    to manipulate the outgoing headers before being sent.
   </para>
   <warning>
-    <simplepara>
+    <simpara>
       Callbacks are not stacked. Subsequent calls to <function>header_register_callback</function>
       will silently de-register any existing <parameter>callback</parameter>, regardless of where
       it was set. Be aware that dependencies may overwrite the desired callback.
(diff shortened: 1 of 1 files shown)


reference/network/functions/header-register-callback.xml
```diff
@@ -21,6 +21,14 @@
    headers to be sent, and before any other output is sent, creating a window
    to manipulate the outgoing headers before being sent.
   </para>
+  <warning>
+    <para>
+      Callbacks are not stacked. Subsequent calls to <function>header_register_callback</function>
+      will silently de-register any existing <parameter>callback</parameter>, regardless of where
+      it was set. Be aware that dependencies could overwrite your callback.
```
jordikroon: “your callback” should simply become “the callback” as we don’t allow personalization according to the style guidelines. I am also thinking whether could is the correct choice. Shouldn’t we should may instead as it’s conditionally.

And by docbook CS, the outer should become simpara.

jordikroon (changes requested):

reference/network/functions/header-register-callback.xml
beryllium: I've applied these suggestions. Thanks!

jordikroon (approved):

jordikroon:
Thank you!

PR #5762, opened 2026-08-14
Clarify heredoc indentation rules and examples in PHP

AlwaysVictorious:


AlwaysVictorious:
Some of the information occurred later than I expected in the docs. My editor then changed some tabs into whitespaces automatically and I found it easier to start over from scratch. New pull request is #5763

PR #5764, opened 2026-08-15
Iterator::rewind: link to reset() function vs rewind() function

terrafrost:


lacatoire (changes requested):
The change replaces `rewind` with `reset`, but these are different things: `reset()` is a built-in function that rewinds plain arrays, while `Iterator::rewind()` is the method `foreach` actually calls on objects implementing `Iterator`.

You can follow the call chain in the engine: `foreach` triggers `zend_fe_reset_iterator()`, which calls `iter->funcs->rewind(iter)`, wired to `zend_user_it_rewind()` — which dispatches to the PHP-level `rewind` method (resolved by name [here](https://github.com/php/php-src/blob/master/Zend/zend_interfaces.c#L344), called [here](https://github.com/php/php-src/blob/master/Zend/zend_interfaces.c#L180)).

So the original wording was correct. Happy to be wrong if I'm missing something!

terrafrost:
Well then maybe `As &foreach; always calls <methodname>rewind</methodname> before starting` should be replaced with `As &foreach; always calls rewind before starting` because, as is, at https://www.php.net/manual/en/iterator.rewind.php, in the Note box, where it says, "_As [foreach](https://www.php.net/manual/en/control-structures.foreach.php) always calls [rewind()](https://www.php.net/manual/en/function.rewind.php) before starting iteration, manually advancing the iterator position_" it links to the function that "_rewind_[s] _the position of a file pointer_".

For that matter, maybe it could even say `As &foreach; always calls this method before starting`.

But, as is, it's hyperlinking to a function that affects _file pointers_.

language/predefined/iterator/rewind.xml
```diff
@@ -22,7 +22,7 @@
     executed <emphasis>after</emphasis> &foreach; loops.
    </para>
    <simpara>
-    As &foreach; always calls <methodname>rewind</methodname> before starting
+    As &foreach; always calls <methodname>reset</methodname> before starting
```
lacatoire:
```suggestion
    As &foreach; always calls <methodname>Iterator::rewind</methodname> before starting
```

terrafrost: You're proposing the page link to itself? Is there precedent for that on php.net? In general I think that that's kinda silly. Like making every instance of "World War "2 on the "World War 2" wikipedia article link to itself...  I'm not sure that's a very useful change.

I feel like `As &foreach; always calls this method before starting` would be better.

jordikroon: @terrafrost I agree with your suggestion. Using `this method` makes more sense.

jordikroon:
Thank you @terrafrost!

PR #5768, opened 2026-08-16
ext/odbc: document the row default value change to null in fetch_object/array/into (PHP 8.4)

lacatoire:
The changelog rows of `odbc_fetch_object()`, `odbc_fetch_array()` and `odbc_fetch_into()` state only that `row` became nullable in PHP 8.4.0. They omit the accompanying change of default value, which is the part that changes behaviour for existing calls: the previous defaults `-1` and `0` were row positions, whereas `null` means "fetch the next row", the semantics `odbc_fetch_row()` already had.

**Sources**

`ext/odbc/odbc.stub.php`, comparing the `php-8.3.0` and `php-8.4.0` tags:

```
8.3.0  odbc_fetch_object($statement, int $row = -1): stdClass|false
8.3.0  odbc_fetch_array($statement, int $row = -1): array|false
8.3.0  odbc_fetch_into($statement, &$array, int $row = 0): int|false
8.3.0  odbc_fetch_row($statement, ?int $row = null): bool

8.4.0  odbc_fetch_object($statement, ?int $row = null): stdClass|false
8.4.0  odbc_fetch_array($statement, ?int $row = null): array|false
8.4.0  odbc_fetch_into($statement, &$array, ?int $row = null): int|false
8.4.0  odbc_fetch_row(Odbc\Result $statement, ?int $row = null): bool
```

### DIFF
--- a/reference/uodbc/functions/odbc-fetch-array.xml
+++ b/reference/uodbc/functions/odbc-fetch-array.xml
@@ -65,7 +65,9 @@
      <row>
       <entry>8.4.0</entry>
       <entry>
-       <parameter>row</parameter> is now nullable.
+       <parameter>row</parameter> is now nullable, and its default value
+       changed from <literal>-1</literal> to &null;, consistent with
+       <function>odbc_fetch_row</function>.
       </entry>
      </row>
     </tbody>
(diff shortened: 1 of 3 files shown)


kamil-tekiela:
The parameter descriptions need to be aligned as well

kamil-tekiela (approved):

kamil-tekiela:
Looks good to me, but I think you have to fix the simparas.

jordikroon (approved):

PR #5772, opened 2026-08-18
Fix method names

mmalferov:


language/predefined/serializable.xml
```diff
@@ -17,7 +17,7 @@
     Classes that implement this interface no longer support
     <link linkend="object.sleep">__sleep()</link> and
     <link linkend="object.wakeup">__wakeup()</link>.
-    The <methodname>serialize</methodname> method is
+    The <methodname>__serialize</methodname> method is
```
jordikroon, line 20: `<methodname>serialize</methodname>` is actually correct, but it conflicts with the function serialize.

So instead of linking to https://www.php.net/manual/en/function.serialize.php, it should link to https://www.php.net/manual/en/serializable.serialize.php.

This should be fixed in phd.

PR #5774, opened 2026-08-19
Reserved Keywords: minor improvements

jrfnl:
1. Mention enums where applicable.
2. Fix formatting of the compile-time constants table.

The below screenshot shows what I intend to fix with [2]:
image


kamil-tekiela (approved):
Thanks!

PR #5776, opened 2026-08-19
Fix unresolved exception ZMQ entity references

jordikroon:
A few more follow-up MRs to come. The main goal is to bring docbook-cs to a state where it reports no unresolved entities.
These errors are caused by incorrect or outdated references.

I have not verified whether the synopsis is correct; that is out of scope.

### DIFF
--- a/reference/zmq/book.xml
+++ b/reference/zmq/book.xml
@@ -19,12 +19,22 @@
  </preface>
 
  &reference.zmq.setup;
- 
+
  &reference.zmq.zmq;
+ &reference.zmq.zmqexception;
+
  &reference.zmq.zmqcontext;
+ &reference.zmq.zmqcontextexception;
+
  &reference.zmq.zmqsocket;
+ &reference.zmq.zmqsocketexception;
+
  &reference.zmq.zmqpoll;
+ &reference.zmq.zmqpollexception;
+
  &reference.zmq.zmqdevice;
+ &reference.zmq.zmqdeviceexception;
+
 
 </book>
 
--- a/reference/zmq/zmqcontextexception.xml
+++ b/reference/zmq/zmqcontextexception.xml
@@ -29,7 +29,7 @@
      <ooclass>
       <classname>ZMQContextException</classname>
      </ooclass>
-     
+
      <ooclass>
       <modifier>extends</modifier>
       <classname>ZMQException</classname>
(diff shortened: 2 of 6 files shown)


alfsb:
What entities is docbook-cs not detecting? DTD entity references, or the targets of XPointers, or other cases?

It's important to note that changing DTD entity references around will change the final XML structure, so XPointers need to be also changed, and that is ok. But changing DTD entities around to fix unresolved entities appear to be a form of loading order mismatch. So to be sure:

1. DTD entity references are resolved on a "first" pass, before anything else;
2. Any resolved DTD entity referenced restart the process; or
3. Each entity content is further expanded before being replaced in text.

This is the "macro order" or "tree building algorithm" of HTML, that *I think* that xmllib does, while assembling the manual. But the more general question is why docbook-cs is hitting undefined entities that libxml does not.

jordikroon:
So the exact errors for this PR are:
```
FILE: reference/zmq/zmqcontextexception.xml
-------------------------------------------
   89 | ERROR   | DocbookCS.Internal | XML parse error: Entity 'reference.zmq.entities.zmqcontextexception' not defined
-------------------------------------------

FILE: reference/zmq/zmqdeviceexception.xml
------------------------------------------
   89 | ERROR   | DocbookCS.Internal | XML parse error: Entity 'reference.zmq.entities.zmqdeviceexception' not defined
------------------------------------------

FILE: reference/zmq/zmqexception.xml
------------------------------------
   89 | ERROR   | DocbookCS.Internal | XML parse error: Entity 'reference.zmq.entities.zmqexception' not defined
------------------------------------

FILE: reference/zmq/zmqpollexception.xml
----------------------------------------
   89 | ERROR   | DocbookCS.Internal | XML parse error: Entity 'reference.zmq.entities.zmqpollexception' not defined
----------------------------------------

FILE: reference/zmq/zmqsocketexception.xml
------------------------------------------
   89 | ERROR   | DocbookCS.Internal | XML parse error: Entity 'reference.zmq.entities.zmqsocketexception' not defined
------------------------------------------
```

The real reason libxml doesn't catch these errors is because these files are never loaded by the manual. They were absent in `reference/zmq/book.xml` which had to be registered. Once done it would have given similar results.

The XPointer changes are a side effect of the files being loaded. Docbook-cs doesn't check. configure.php does.
In total there were 35 violations of undefined entities across the manual, which I have covered in this and follow up PRs.

As a side note, I am planning to load `doc-base/temp/doctype.dtd` instead (it contains all entities as you mentioned in php/doc-base#321). In order to be 100% sure about the order. But the errors reported here would still be triggered.

alfsb:
> The real reason libxml doesn't catch these errors is because these files are never loaded by the manual. They were absent in `reference/zmq/book.xml` which had to be registered. Once done it would have given similar results.

Ah, you found out files not included in XML assembly. That was unexpected.

> As a side note, I am planning to load `doc-base/temp/doctype.dtd` instead (it contains all entities as you mentioned in [php/doc-base#321](https://github.com/php/doc-base/pull/321)). In order to be 100% sure about the order. But the errors reported here would still be triggered.

Please, do. doc-base 321 is now merged. I will wait for the manual to build and publish before erasing the PR branch, but I expect the change will build ok. After this settles, then *all* loading of DTD files can be replaced by the loading of doc-base/temp/doctype.dtd.

jordikroon:
@alfsb thank you for the work done in 321. https://github.com/php/doc-en/commit/f3aa7a8fff15164808bdabb0281d5e88f8a5e504 simplified a lot.

PR #5779, opened 2026-08-19
Fix unresolved entity references

jordikroon:
This is the last MR related to this. The main goal is to bring docbook-cs to a state where it reports no unresolved entities.
These errors are caused by incorrect or outdated references.

I have not verified whether the synopsis is correct; that is out of scope.


lacatoire:
LGTM. The 4 PR cleans and remove some files and links, couldn't find any `linkend` in the manual pointing at the removed ids.

One note, not a blocker and not the subject of this PR: registering these files also makes them render for the first time, and most of them have an empty `&reftitle.intro;` section, the built page shows an "Introduction" heading with nothing under it. Worth a follow-up at some point.

PR #5783, opened 2026-08-20
mysqlnd: document wait-timeout error code change 2006 -> 4031 (8.4.0)

lacatoire:
Document that as of PHP 8.4.0, the error code for MySQL server wait timeouts changed from 2006 (CR_SERVER_GONE_ERROR) to 4031 (ER_CLIENT_INTERACTION_TIMEOUT) for MySQL 8.0.24+.

https://github.com/orgs/php/projects/11/views/4

### DIFF
--- a/reference/mysqli/mysqli/query.xml
+++ b/reference/mysqli/mysqli/query.xml
@@ -131,6 +131,33 @@
   &mysqli.conditionalexception;
  </refsect1>
 
+ <refsect1 role="changelog">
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
I'm not sure about this one. It would be documented in a pretty obscure place and the change isn't something out of the ordinary. While it makes sense to have it in UPGRADING, I don't know that we need to put it in the PHP manual.

PR #5784, opened 2026-08-20
Clarify forward_static_call docs

zonuexe:
This PR replaces #5681 and has been rebased on the latest `master` after 846dfb87.

Commit 846dfb87 clarified that `forward_static_call_array()` does not require an active class scope and behaves like `call_user_func_array()` when called outside a class context.

This PR keeps that correction and focuses on the remaining parts not covered by that commit:

- clarifying that `forward_static_call()` is not limited to calling static methods, but preserves the current called class for method callbacks
- explaining how this relates to late static binding, where `static::` resolves to the current called class
- clarifying that this is useful for static method callbacks that should behave like forwarding calls such as `parent::` or `self::`
- updating the `forward_static_call_array()` parameter note to match `call_user_func_array()`: parameters can be passed by reference when the corresponding element in the `args` array is a reference

### DIFF
--- a/reference/funchand/functions/forward-static-call-array.xml
+++ b/reference/funchand/functions/forward-static-call-array.xml
@@ -3,8 +3,8 @@
 <refentry xml:id="function.forward-static-call-array" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>forward_static_call_array</refname>
-  <refpurpose>Call a callback with parameters as an array, forwarding late
-   static binding when available</refpurpose>
+  <refpurpose>Call a callback with an array of parameters, preserving the 
+  current called class</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
--- a/reference/funchand/functions/forward-static-call-array.xml
+++ b/reference/funchand/functions/forward-static-call-array.xml
@@ -3,7 +3,8 @@
 <refentry xml:id="function.forward-static-call-array" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>forward_static_call_array</refname>
-  <refpurpose>Call a static method and pass the arguments as array</refpurpose>
+  <refpurpose>Call a callback with parameters as an array, forwarding late
+   static binding when available</refpurpose>
  </refnamediv>
 
  <refsect1 role="description">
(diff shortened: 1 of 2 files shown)


reference/funchand/functions/forward-static-call-array.xml
```diff
@@ -3,7 +3,8 @@
 <refentry xml:id="function.forward-static-call-array" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>forward_static_call_array</refname>
-  <refpurpose>Call a static method and pass the arguments as array</refpurpose>
+  <refpurpose>Call a callback with parameters as an array, forwarding late
+   static binding when available</refpurpose>
```
lacatoire:
```suggestion
  <refpurpose>Call a callback with an array of parameters, preserving the 
  current called class</refpurpose>
```

reference/funchand/functions/forward-static-call-array.xml
```diff
@@ -17,10 +18,13 @@
...
-   <link linkend="language.oop5.late-static-bindings">late static binding</link>.
-   When called outside a class context, it behaves like
-   <function>call_user_func_array</function>, without late static binding.
+   When called from within a method context and calling a method callback,
+   it forwards the current
+   <link linkend="language.oop5.late-static-bindings">late static binding</link>,
+   so <literal>static::</literal> inside the method resolves to the current
+   called class. This is useful for static method callbacks that should behave
+   like forwarding calls such as <literal>parent::</literal> or
+   <literal>self::</literal>.
```
lacatoire:
```suggestion
   When called from within a method context and calling a method callback,
   it makes a
   <link linkend="language.oop5.late-static-bindings">forwarding call</link>,
   so <literal>static::</literal> inside the method resolves to the same called
   class as in the caller. This is useful for static method callbacks that
   should behave like <literal>self::</literal> or <literal>parent::</literal>.
```

reference/funchand/functions/forward-static-call.xml
```diff
@@ -13,13 +13,17 @@
...
    context, it can't be used outside a class.
-   It uses the <link linkend="language.oop5.late-static-bindings">late static
-   binding</link>.
-  </para>
+   When calling a method callback, it forwards the current
+   <link linkend="language.oop5.late-static-bindings">late static binding</link>,
+   so <literal>static::</literal> inside the method resolves to the current
+   called class. This is useful for static method callbacks that should behave
+   like forwarding calls such as <literal>parent::</literal> or
+   <literal>self::</literal>.
```
lacatoire:
```suggestion
   When calling a method callback, it makes a
   <link linkend="language.oop5.late-static-bindings">forwarding call</link>,
   so <literal>static::</literal> inside the method resolves to the same called
   class as in the caller. This is useful for static method callbacks that
   should behave like <literal>self::</literal> or <literal>parent::</literal>.
```

reference/funchand/functions/forward-static-call-array.xml
```diff
@@ -3,7 +3,8 @@
 <refentry xml:id="function.forward-static-call-array" xmlns="http://docbook.org/ns/docbook">
  <refnamediv>
   <refname>forward_static_call_array</refname>
-  <refpurpose>Call a static method and pass the arguments as array</refpurpose>
+  <refpurpose>Call a callback with an array of parameters, preserving the
```
lacatoire:
```suggestion
  <refpurpose>Call a callback with an array of parameters, preserving the
```

PR #5785, opened 2026-08-20
Update tables to fix mobile labels

pwright64:
Moved "properties" and "methods" to table titles from table entries, splitting tables at these points; removed class names from table entries; and updated table id's to match new tables. This fixes the following issue: On mobile, table is collapsed and table head entries are used as :before label elements for table body entries, in a cycling fashion (see web-php/public/js/common.js line 835-851). Table had class names as head entries and "methods" and "properties" as body entries, resulting in incorrect labels for body entries.

### DIFF
--- a/reference/mysqli/summary.xml
+++ b/reference/mysqli/summary.xml
@@ -114,7 +114,7 @@
    </tbody>
   </tgroup>
  </table>
- 
+
  <table xml:id="mysqli.summary.classmethods">
   <title>Summary of <classname>mysqli</classname> methods</title>
   <tgroup cols="4">
@@ -424,8 +424,8 @@
   </tgroup>
  </table>
 
- <table xml:id="mysqli.summary.stmtmethods"> 
-  <title>Summary of <classname>mysqli_stmt</classname> methods</title>    
+ <table xml:id="mysqli.summary.stmtmethods">
+  <title>Summary of <classname>mysqli_stmt</classname> methods</title>
   <tgroup cols="4">
    <thead>
     <row>
@@ -593,7 +593,7 @@
    </tbody>
   </tgroup>
  </table>
- 
+
  <table xml:id="mysqli.summary.resultmethods">
   <title>Summary of <classname>mysqli_result</classname> methods</title>
   <tgroup cols="4">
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Thank you for this PR. However, I propose an alternative: remove this page altogether. I don't know why we have it. It's often out of sync with the rest of the manual. The individual pages for each class are better. I suppose it was for people using only the procedural style functions, but I don't think even they would find this page very useful. The DateTime extension has this done a bit differently https://www.php.net/manual/en/book.datetime.php. Perhaps we should do something similar in mysqli: list all functions as aliases of the methods and link to their pages without a table or description.

kamil-tekiela:
On top of that, we have this page https://www.php.net/manual/en/ref.mysqli.php confusingly named and equally useless. Are they all the aliases available? Are they all deprecated? No, it's neither.

pwright64:
I see what you're saying, and that explains why I had trouble finding an analogous page to compare with. So: getting rid of this page, splitting the alias and deprecated section into functions and deprecated, and adding the missing functions pages, seems like a sensible course of action. If it's helpful, I could work on creating pages for the procedural functions as aliases with links to the OOP pages, like the date/time page. Although that practice seems wasteful. It would be nice if we could use symlinks or redirects instead.

pwright64:
I don't know if there is a way to implement a redirect as I mentioned above, but for an end user, it would be nicer. A way of doing something similar is to copy the OOP page to the functions folder, change the xml:id to "function.[function_name]" and move the function name to the first refname. Then move deprecated functions/aliases to their own section, if we need to maintain a section for them, with xml:id like "deprecated.[function_name]".

kamil-tekiela:
> I see what you're saying, and that explains why I had trouble finding an analogous page to compare with. So: getting rid of this page, splitting the alias and deprecated section into functions and deprecated, and adding the missing functions pages, seems like a sensible course of action. If it's helpful, I could work on creating pages for the procedural functions as aliases with links to the OOP pages, like the date/time page. Although that practice seems wasteful. It would be nice if we could use symlinks or redirects instead.

This is a lot of work and you would be opening up a pandora's box. We need a simpler solution. Perhaps, it's not even worth it to split it up like that anymore.

kamil-tekiela:
Thank you for bringing this to our attention. I have decided to solve this problem by removing the page completely. We no longer have a list of functions, but I don't think that's going to be an issue. To make it like they do in DateTime, we would have to split every page, which is just too much work.

pwright64:
You're welcome, thanks for following up quickly throughout!

PR #5791, opened 2026-08-25
How to flush cookies as of PHP8

### LINKED ISSUE #2239: can we make curl_close() write CURLOPT_COOKIEJAR again?
### Description

The following code:

```php
<?php
declare(strict_types=1);
$ch = curl_init();
$cookiefile_handle = tmpfile();
$cookiefile = stream_get_meta_data($cookiefile_handle)['uri'];
curl_setopt_array($ch, array(
    CURLOPT_COOKIEFILE => $cookiefile,
    CURLOPT_COOKIEJAR => $cookiefile,
    CURLOPT_WRITEFUNCTION => function ($ch, $str) {
        // echo $str;
        return strlen($str);
    },
));
curl_setopt($ch, CURLOPT_URL, "https://example.com/");
curl_exec($ch);
var_dump(file_get_contents($cookiefile));
curl_close($ch);
var_dump(file_get_contents($cookiefile));
```
will, prior to PHP8.0.0 result in:
```
$ php7.4 test.php
string(0) ""
string(131) "# Netscape HTTP Cookie File
# https://curl.se/docs/http-cookies.html
# This file was generated by libcurl! Edit at your own risk.


```

and in >=8.0.0 result in
```
$ php8.2 test.php
string(0) ""
string(0) ""

```
which breaks scripts relying on curl_close() writing to CURLOPT_COOKIEJAR

- either we should make curl_close write to COOKIEJAR again, or we should at least add a note to https://www.php.net/manual/en/migration80.incompatible.php that you can no longer rely on curl_close to write to COOKIEJAR.
random
- related stackoverflow post: https://stackoverflow.com/a/13800822/1067003

### PHP Version

PHP 8.0.0

### Operating System

_No response_

divinity76:
Prior to php8, curl_close($ch) would flush the cookie list. Code that depend on the cookie list to be flushed on curl_close() breaks on PHP8. Document an alternative way to flush the cookie list: curl_setopt($ch, CURLOPT_COOKIELIST, "FLUSH");

Resolves https://github.com/php/doc-en/issues/2239

### DIFF
--- a/reference/curl/functions/curl-close.xml
+++ b/reference/curl/functions/curl-close.xml
@@ -19,11 +19,11 @@
   </methodsynopsis>
   &note.resource-migration-8.0-dead-function;
    <note>
-    <para>
+    <simpara>
      If cookies need to be written before the <parameter>handle</parameter>
      is destroyed, call <function>curl_setopt</function> with
      <constant>CURLOPT_COOKIELIST</constant> set to <literal>"FLUSH"</literal>.
-    </para>
+    </simpara>
    </note>
 
   <para>
--- a/reference/curl/functions/curl-close.xml
+++ b/reference/curl/functions/curl-close.xml
@@ -19,13 +19,12 @@
   </methodsynopsis>
   &note.resource-migration-8.0-dead-function;
    <note>
-   <para>
-    If cookies need to be written before the handle is automatically
-    destroyed, call
-    <function>curl_setopt</function>(<parameter>handle</parameter>,
-    <constant>CURLOPT_COOKIELIST</constant>, <literal>"FLUSH"</literal>).
-   </para>
-  </note>
+    <para>
+     If cookies need to be written before the <parameter>handle</parameter>
+     is destroyed, call <function>curl_setopt</function> with
+     <constant>CURLOPT_COOKIELIST</constant> set to <literal>"FLUSH"</literal>.
+    </para>
+   </note>
 
   <para>
    Closes a cURL session and frees all resources.  The cURL handle,
--- a/reference/curl/functions/curl-close.xml
+++ b/reference/curl/functions/curl-close.xml
@@ -9,7 +9,7 @@
  <refsynopsisdiv>
   &warn.deprecated.function-8-5-0;
  </refsynopsisdiv>
- 
+
  <refsect1 role="description">
   &reftitle.description;
   <methodsynopsis>
@@ -18,6 +18,15 @@
    <methodparam><type>CurlHandle</type><parameter>handle</parameter></methodparam>
   </methodsynopsis>
   &note.resource-migration-8.0-dead-function;
+   <note>
+   <para>
+    If cookies need to be written before the handle is automatically
+    destroyed, call
+    <function>curl_setopt</function>(<parameter>handle</parameter>,
+    <constant>CURLOPT_COOKIELIST</constant>, <literal>"FLUSH"</literal>).
+   </para>
+  </note>
+
   <para>
    Closes a cURL session and frees all resources.  The cURL handle,
    <parameter>handle</parameter>, is also deleted.
--- a/reference/curl/functions/curl-close.xml
+++ b/reference/curl/functions/curl-close.xml
@@ -21,8 +21,8 @@
    <note>
     <simpara>
      If cookies need to be written before the <parameter>handle</parameter>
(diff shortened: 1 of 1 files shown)


reference/curl/functions/curl-close.xml
```diff
@@ -18,6 +18,15 @@
...
   </methodsynopsis>
   &note.resource-migration-8.0-dead-function;
+   <note>
+   <para>
+    If cookies need to be written before the handle is automatically
+    destroyed, call
+    <function>curl_setopt</function>(<parameter>handle</parameter>,
+    <constant>CURLOPT_COOKIELIST</constant>, <literal>"FLUSH"</literal>).
+   </para>
+  </note>
```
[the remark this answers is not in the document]
divinity76: > added note shows curl_setopt(handle, ...) using <parameter>handle</parameter>, which renders as handle (missing $) and isn’t valid PHP code. 

Not quite, the parameter tag adds the $ automatically..

>It also omits that FLUSH writes cookies to the file specified by CURLOPT_COOKIEJAR, which is necessary context for what gets “written”

pretty sure anyone who needs to know how to flush the cookie jar in php8, already knows where the jar gets flushed to. Target audience already knows.

reference/curl/functions/curl-close.xml
```diff
@@ -18,6 +18,15 @@
    <methodparam><type>CurlHandle</type><parameter>handle</parameter></methodparam>
   </methodsynopsis>
   &note.resource-migration-8.0-dead-function;
+   <note>
```
jordikroon: `<note>` is wrongfully indented here. Besides, `<para>` should become a `simpara` as Docbook Lint suggested.

reference/curl/functions/curl-close.xml
```diff
@@ -18,6 +18,14 @@
...
+    <para>
+     If cookies need to be written before the <parameter>handle</parameter>
+     is destroyed, call <function>curl_setopt</function> with
+     <constant>CURLOPT_COOKIELIST</constant> set to <literal>"FLUSH"</literal>.
+    </para>
```
lacatoire:
```suggestion
    <simpara>
     If cookies need to be written before the <parameter>handle</parameter>
     is destroyed, call <function>curl_setopt</function> with
     <constant>CURLOPT_COOKIELIST</constant> set to <literal>"FLUSH"</literal>.
    </simpara>
```

divinity76:
related to https://github.com/php/doc-en/issues/2239

lacatoire:
@divinity76 if you write Fixes: #IssueNumber in your PR's description, then once merged it will close the issue automatically ;)

nrps9909 (commented):
Approved exact head `884c31ad27aa3c8ffddc1c46d0f6cba93e6fbccd`.

The added migration note matches the underlying libcurl contract: `CURLOPT_COOKIELIST` with `"FLUSH"` writes all known cookies to the file configured through `CURLOPT_COOKIEJAR`. It also puts that recovery path on the `curl_close()` page, where PHP 8 users who discover that the function is now a no-op are most likely to need it; the more detailed option reference already documents the same operation.

I checked the exact diff against the current option documentation and libcurl's official `CURLOPT_COOKIELIST` reference. The current exact-head `Build (en)` and `DocBook Style Check (en)` jobs are both green, and `git diff --check` passes locally.

AI assistance disclosure: Codex was used to inspect the exact-head XML and repository guidance, compare the related PHP documentation, verify the libcurl semantics, audit live CI, and draft this review. I verified the cited SHA and results before approving.

nrps9909:
Clarification on my review above: GitHub does not allow me to submit an APPROVED review without explicit repository access, so its actual state is COMMENTED. “Approved exact head” there means I found no blocking issue in the cited SHA; it is not a maintainer approval. The semantic and CI evidence in the review is unchanged.

kamil-tekiela:
IMO this is the wrong place to add this information. The documentation page will removed at some point when that function gets removed from the language. The information you are proposing to add should outlive it. You'd need to do something like https://github.com/php/doc-en/pull/2949/changes

kamil-tekiela:
Correction: this seems to already be documented at https://www.php.net/manual/en/curl.constants.php#constant.curlopt-cookiejar, so this PR is redundant. The linked issue should have been closed already. Thank you for opening it anyway.

PR #5806, opened 2026-08-26
ext/intl: document PHP 8.5 changes

lacatoire:
Documents the following PHP 8.5 changes to the Intl extension:

- Add `Locale::addLikelySubtags()` method
- Add `Locale::minimizeSubtags()` method
- Add `grapheme_levenshtein()` function
- `IntlChar`, `Locale`, and related functions now throw `IntlException` instead of triggering warnings/notices on invalid input

PHP 8.5 tracker: https://github.com/orgs/php/projects/13

### DIFF
--- a/reference/intl/locale.xml
+++ b/reference/intl/locale.xml
@@ -35,7 +35,9 @@
    </simpara>
    <simpara>
     Locales cannot be instantiated as objects. All of the functions/methods
-    provided are static.
+    provided are static. Use <methodname>Locale::getDefault</methodname> and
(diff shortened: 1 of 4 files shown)


reference/intl/locale.xml
```diff
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- $Revision$ -->
```
jordikroon, line 2: This shouldn't be removed and the other files should have this as well.

PR #5811, opened 2026-08-27
Add examples for mb_scrub()

### LINKED ISSUE #5562: Add basic examples for mb_scrub()
### Affected page

https://www.php.net/manual/en/function.mb-scrub.php

### Current issue

The `mb_scrub()` manual page explains that the function replaces ill-formed byte sequences with the substitute character, but it currently has no examples.

Without an example, it is difficult to see that the replacement happens inside the PHP string itself. Visual output alone can also be misleading, because terminals and browsers may display malformed byte sequences using a replacement character.

As the original author of `mb_scrub()`, I apologize that the manual page has remained without examples for a long time.

### Suggested improvement

Add a basic example using bin2hex() to show the byte-level replacement performed by `mb_scrub()`.

Also add a small example showing that the scrubbed string can be passed to a UTF-8 aware PCRE pattern using the u modifier.

**Proposed example: byte-level replacement**

```php
<?php

$input = "A\xFFB";

echo bin2hex($input), "\n";

$clean = mb_scrub($input, 'UTF-8');

echo bin2hex($clean), "\n";
```

Expected output:

```text
41ff42
41efbfbd42
```

Suggested explanation:

The byte ff is not valid in UTF-8. After `mb_scrub()`, it is replaced with efbfbd, the UTF-8 encoding of U+FFFD REPLACEMENT CHARACTER.

This example uses `bin2hex()` so that the result does not depend on how a terminal, browser, or font displays malformed byte sequences.

**Proposed example: use before UTF-8 aware PCRE processing**

```php
<?php

$input = "A\xFFB";

$clean = mb_scrub($input, 'UTF-8');

$count = preg_match_all('/./us', $clean);

if ($count === false) {
    echo preg_last_error_msg(), "\n";
    exit;
}

echo $count, "\n";
```

Expected output:

```text
3
```

Suggested explanation:

`mb_scrub()` can be useful before passing external input to UTF-8 aware text processing functions. In this example, the malformed byte sequence is replaced before the string is passed to a PCRE pattern using the u modifier.

The result of `preg_match_all()` is the number of matches for the given pattern. In this example, `/./us` matches each UTF-8 code point, including the replacement character.

This issue intentionally focuses only on simple examples for `mb_scrub()`.

Related topics, such as how PCRE functions and grapheme_* functions handle invalid UTF-8 input, how streamed input should be handled, and how substitution differs from validation, escaping, or database storage behavior, can be discussed separately.

lacatoire:
The `mb_scrub()` page had no example, which made it hard to see that the replacement happens in the string itself. Terminals, browsers and fonts render an ill-formed byte sequence with a replacement character of their own, so visual output alone is misleading. Both examples therefore use `bin2hex()` or `var_dump()`, so the reader sees the bytes rather than the rendering.

**Byte-level replacement.** The first example also shows that the result depends on the substitute character. `mbstring.substitute_character` defaults to `"?"` (`0x3F`), so scrubbing `"A\xFFB"` yields `41 3f 42`; `41 ef bf bd 42` is only produced after calling `mb_substitute_character(0xFFFD)`. This differs from the output given in the issue, which assumed U+FFFD by default.

**Use before UTF-8 aware processing.** The second example shows the practical case: a PCRE pattern with the `u` modifier fails on the raw input with `Malformed UTF-8 characters, possibly incorrectly encoded`, and succeeds once the string has been scrubbed.

A see also section is added, since the page had none and the substitute character is a prerequisite for reading the first example.

Both `<screen>` blocks were extracted from the file and compared byte for byte against the real output on PHP 8.1, 8.4 and 8.5.

Fixes: #5562

### DIFF
--- a/reference/mbstring/functions/mb-scrub.xml
+++ b/reference/mbstring/functions/mb-scrub.xml
@@ -71,6 +71,90 @@
   </informaltable>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
+   <example>
+    <title>Byte-level replacement performed by <function>mb_scrub</function></title>
+    <simpara>
+     <function>bin2hex</function> is used here because terminals, browsers and
+     fonts may render an ill-formed byte sequence with a replacement character
+     of their own, which hides what the string actually contains.
+    </simpara>
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+// The byte 0xFF cannot appear in a valid UTF-8 string.
+$input = "A\xFFB";
+echo bin2hex($input), "\n";
+
+// The default substitute character is "?" (0x3F).
+echo bin2hex(mb_scrub($input, 'UTF-8')), "\n";
+
+// U+FFFD REPLACEMENT CHARACTER is encoded as EF BF BD in UTF-8.
+mb_substitute_character(0xFFFD);
+echo bin2hex(mb_scrub($input, 'UTF-8')), "\n";
+
+?>
+]]>
+    </programlisting>
+    &example.outputs;
+    <screen>
+<![CDATA[
+41ff42
+413f42
+41efbfbd42
+]]>
+    </screen>
+   </example>
+  </para>
+  <para>
+   <example>
+    <title>Using <function>mb_scrub</function> before UTF-8 aware processing</title>
+    <simpara>
+     PCRE patterns using the <literal>u</literal> modifier reject subjects that
+     are not well-formed UTF-8. Scrubbing the input first makes it acceptable.
+    </simpara>
+    <programlisting role="php">
+<![CDATA[
+<?php
+
+$input = "A\xFFB";
+
+var_dump(preg_match_all('/./us', $input));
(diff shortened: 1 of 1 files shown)


reference/mbstring/functions/mb-scrub.xml
```diff
@@ -71,6 +71,90 @@
   </informaltable>
  </refsect1>
 
+ <refsect1 role="examples">
+  &reftitle.examples;
+  <para>
```
jordikroon: Examples don't need to be wrapped in a para. It's on my backlog to be added to the sniffer.

reference/mbstring/functions/mb-scrub.xml
```diff
@@ -71,6 +71,90 @@
...
+int(3)
+]]>
+    </screen>
+   </example>
+  </para>
+ </refsect1>
+
+ <refsect1 role="seealso">
+  &reftitle.seealso;
+  <para>
```
jordikroon: Same goes on for lists

PR #5836, opened 2026-09-05
Aligns Git config

NickSdot:
Updates `.gitattributes` so Git normalises text files to LF.


jordikroon (approved):
I am fine with this change. Will leave this open for a few days in order for anyone to share their concerns.

PR #5840, opened 2026-09-07
Fail the CI on changes to migrated extensions

lacatoire:
Fails the CI when a pull request adds to or modifies a `reference/` folder of an extension that is migrating to php/doc-extensions, so that the notices merged in #5828 are enforced rather than only stated in prose.

The list is not hardcoded: the job takes the extensions carrying the `MOVING_TO_DOC_EXTENSIONS_DO_NOT_CHANGE_HERE.md` notice, read at the base commit, and matches them against the changed paths. Deletions are left alone, so removing those folders in batches is not blocked, and the check disappears on its own once the notices are gone.

Two limits worth knowing before merging this:

- It only sees pull requests, so it does not cover direct pushes to master.
- Repository-wide sweeps do touch these folders. A `skip-migrated-check` label lets one through, which only a maintainer can add. Seven currently open pull requests would go red until rebased or labelled.

### DIFF
--- a/.github/workflows/migrated-extensions.yaml
+++ b/.github/workflows/migrated-extensions.yaml
@@ -32,25 +32,20 @@ jobs:
       NOTICE: "MOVING_TO_DOC_EXTENSIONS_DO_NOT_CHANGE_HERE.md"
 
     steps:
+      # The check only reads path names, never file contents, so a tree:0
+      # partial clone brings the whole commit history for about the cost
+      # of a shallow one. The base commit is then always present, which
+      # is all the diff below needs.
       - name: "Checkout php/doc-en"
         uses: "actions/checkout@v7"
         with:
           ref: "${{ github.event.pull_request.head.sha }}"
-          fetch-depth: 50
+          fetch-depth: 0
+          filter: "tree:0"
 
-      - name: "Fetch diff base"
+      - name: "Check the diff base is reachable"
         run: |
           set -euo pipefail
-
-          if ! git fetch origin "$BASE_SHA" --depth=50; then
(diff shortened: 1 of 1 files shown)


.github/workflows/migrated-extensions.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - name: "Fetch diff base"
+        run: |
+          set -euo pipefail
+
+          if ! git fetch origin "$BASE_SHA" --depth=50; then
+            echo "::error::Cannot fetch the base commit $BASE_SHA, it may have been force-pushed."
+            exit 1
+          fi
+          # Deepen until the merge-base is reachable (long-lived branches)
+          for i in 1 2 3 4 5; do
```
alfsb: Having a for  `for i in 10 100 1000 ; do`, and using `$i` as the parameter for --deepen would cover most of the cases.

For the following `if`, per https://github.com/php/doc-pt_br/blob/master/.github/workflows/buildxml.yaml#L19 , a clone/fetch with ` --filter=tree:0 ` will download all commit history without any blobs, that is fast enough, so the only case when `git merge-base` would fail is if the commit itself is detached from the git tree.

alfsb:
Too late as the code is done, but a simpler test would be cloning the repo with `--filter=blob:none `, listing all dirs with the file marker, and then git log or ls-tree --name-only from current to after date of creation of files to HEAD, or from SHA after file marker creation. So the test would fail if any filename line (or any line, really) matches the prefix of file marks.

This test not need to be the most optimized one, as it will only necessary while the first mass migration is not completed.

PR #5842, opened 2026-09-07
Fixes the "Display full permissions" example using `fileperms`.

brcontainer:
In the *regular file* example, you are using `r`, when you should be using `-`:

```php
case 0x8000: // regular
    $info = 'r';
```

This is according to `ls -l`:

| Character | Description |
| --- | --- |
| `-` | regular |
| `d` | directory |
| `l` | symbolic link |
| `c` | character device |
| `b` | block device |
| `p` | FIFO |
| `s` | socket |

The page states that the output of the code for `/etc/passwd` is:

```
-rw-r--r--
```

However, I obtained this:

Image

Another problem is that the use of `fileperms()` will never utilize this:

```php
case 0xA000: // symbolic link
    $info = 'l';
```

Test:

```bash
$ ln -s sample.txt samplelnk.txt
$ ls -l samplelnk.txt
lrwxrwxrwx 1 Brcontainer Brcontainer 10 Sep  7 16:46 samplelnk.txt -> sample.txt
```
Whereas in PHP (`$perms = fileperms('samplelnk.txt');`), I got this: `-rw-r--r--`.

Therefore, the snippet for the example is essentially _dead code_. I don't think this section needs to be removed, but it would be important to document it to avoid confusion.

> Note: In the code, I kept the parentheses solely to aid readability and understanding.

### DIFF
--- a/reference/filesystem/functions/fileperms.xml
+++ b/reference/filesystem/functions/fileperms.xml
@@ -103,19 +103,19 @@ switch ($perms & 0xF000) {
 }
 
 // Owner
-$setuid  = $perms & 0x0800;
+$setuid = $perms & 0x0800;
 $info .= (($perms & 0x0100) ? 'r' : '-');
 $info .= (($perms & 0x0080) ? 'w' : '-');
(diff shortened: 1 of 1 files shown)


lacatoire:
Well done @brcontainer ;)

PR #5854, opened 2026-09-10
cloning: visibility applies to all properties, readonly is orthogonal

lacatoire:
- `language/oop5/cloning.xml`: visibility is checked for all properties (not just `readonly`); `readonly` is a separate orthogonal constraint

### DIFF
--- a/language/oop5/cloning.xml
+++ b/language/oop5/cloning.xml
@@ -168,10 +168,11 @@ echo (clone $dateTime)->format('Y');
    <simpara>
     The overrides are applied after
     <link linkend="object.clone">__clone()</link> has run, and they honour the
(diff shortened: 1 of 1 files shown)


TimWolla (changes requested):
This is no longer wrong, but I still find it confusing: Why would it mention `readonly` for visibility? There is no reason for users to believe that `readonly` behave any different from other properties visibility-wise (for cloning).

The “Additionally” sentence just repeats the same visibility part again. Of course readonly can only be overridden when you are in the right scope.

The correct claims are:

1. Visibility is checked. No ifs.
2. Updating readonly properties is legal, even if they are already set on the original object.

TimWolla (approved):

PR #5861, opened 2026-09-17
`FILTER_VALIDATE_URL`: document some limitations in the validation

DanielEScherzer:
This filter accepts a lot of things that might be unexpected, which led to some Mythos reports. While the limited validation isn't a security issue in PHP itself, it is worth calling out for anyone using `FILTER_VALIDATE_URL` for security-sensitive validation in applications.


jordikroon (approved):
LGTM. Though not related to this change, we should probably split this manual page into sub pages:
 - Input Constants
 - Generic Filter Flags
 - Sanitization Filter Flags
 - Validation Filters
 - User Defined Filter

Kind of writing this for myself, but if anyone wants to take the job. Feel free.

DanielEScherzer:
I'll leave this open for a bit before merging in case there are any objections

PR #5862, opened 2026-09-17
`DOMDocument::xinclude()`: Add a warning about untrusted input

DanielEScherzer:


reference/dom/domdocument/xinclude.xml
```diff
@@ -22,6 +22,14 @@
     unexpected results if the included XML file have an attached DTD.
    </para>
   </note>
+  <warning>
+    <simpara>
```
jordikroon, line 26: The parentheses make it harder to read IMO. So, how about:

> Avoid calling this method on untrusted XML, as XIncludes can include files from external websites or arbitrary local files, depending on whether the href attribute is a URL or a file:// path.

DanielEScherzer: The "depending on" part reads a bit confusingly, since it suggests that those are the only two options

I switched to saying "when" instead of "where" after reading it with fresh eyes, but I think the parentheses are needed for clarity

ndossche (approved):

jordikroon (approved):

PR #5865, opened 2026-09-18
`XSLTProcessor::registerPHPFunctions()`: document and warn against `null`

DanielEScherzer:


lacatoire (approved):

DanielEScherzer:
@lacatoire generally we let contributors with merge access merge their own PRs

PR #5868, opened 2026-09-22
Document printf precision fix in PHP 8.5

lacatoire:
A formatter without a specified precision now correctly treats it as precision 0 instead of resetting it. Added 8.5.0 changelog entry to both `strings.changelog.sprintf` and `strings.changelog.vsprintf` entities, covering printf, sprintf, fprintf, vprintf, vsprintf and vfprintf.

See php/php-src#18897.

[PHP 8.5 Documentation Tracker](https://github.com/orgs/php/projects/13?pane=issue&itemId=199834854)


B198603 (commented):
The

PR #5870, opened 2026-09-22
Remove unlink from SQLite3::prepare example

brcontainer:
The `unlink` was likely a leftover from some behavior test. Its use in the example doesn't seem to serve any purpose in the context of the explanation regarding `SQLite3::prepare`.


jordikroon:
Removing `unlink` is a good idea in my opinion , however by passing a filename it will create a file on initialisation. That's something we need to be aware of.

If we change the filename to `:memory:` it will be more clear for a user that this database will be stored in-memory and be that data will be gone when the connection closes.

brcontainer:
> If we change the filename to `:memory:` it will be more clear for a user that this database will be stored in-memory and be that data will be gone when the connection closes.

Thanks for the reply. While the idea isn't bad, I should point out that other examples pointing to .db files don't use `unlink`, and that automatic creation depends on the flags. I think removing `unlink` is sufficient for now, and in the future, I might submit a pull request with improvements to all the examples (including error handling for open failures). Thanks!

jordikroon:
Fair enough, and thank you.

PR #5872, opened 2026-09-23
`session.upload_progress.enabled`: Add a warning about potential exploits

DanielEScherzer:
Document that if upload progress bars are not going to be used, the feature should be disabled, since it makes exploiting any local file inclusion vulnerability very easy.


reference/session/ini.xml
```diff
@@ -970,6 +970,16 @@
...
     </simpara>
    </listitem>
+   <warning>
+     <simpara>
+        Upload progress writes the client-supplied file name and field name to the
+        session store while the upload in progress, to allow polling the process
+        from a separate request. In applications that do not use upload progress
+        bars, consider disabling this option; file names and field names are not
+        validated and may contain PHP code, making any local file inclusion
+        vulnerability trivial to exploit.
```
jordikroon, line 980: Perhaps we should add a link to https://en.wikipedia.org/wiki/File_inclusion_vulnerability as well? WDYT?

reference/session/ini.xml
```diff
@@ -970,6 +970,16 @@
      Defaults to 1, enabled.
     </simpara>
    </listitem>
+   <warning>
```
jordikroon, line 973: Warning is not allowed here. It should be a child of `<listitem>`.

reference/session/ini.xml
DanielEScherzer: done

PR #5873, opened 2026-09-23
ext/pgsql: Remove the reference to `pg_service()` which is unavailable with any released libpq

KentarouTakeda:
The `pg_service()` function, added to the documentation in #5820, is never available with the released version of libpq. This feature was removed upstream during the PostgreSQL 18 beta phase.

Actual behavior:
```php
php > var_dump(PHP_VERSION, PGSQL_LIBPQ_VERSION);
string(6) "8.5.10"
string(4) "18.6"
php > var_dump(function_exists('pg_close_stmt'), function_exists('pg_service'));
bool(true)
bool(false)
```

See also:
- libpq: Remove PQservice()
  https://github.com/postgres/postgres/commit/fc3edb52fbb9b773442ce0a89116f893aaa766af
- Discussion
  https://www.postgresql.org/message-id/20250706161319.c1.nmisch@google.com


lacatoire (approved):

kamil-tekiela (approved):
I see this has already been removed from PHP-src.

